PR #7103, opened 2017-12-11
Do not indent right-side of pipelines in the formatter

### LINKED ISSUE #7102: Formatter: Indentation is broken when |> case do
I'm testing around the formatter and looking at my code. I just found that some indentation is broken in some cases.

An issue is when you have a pipeline to a `case`, the line after `do`, i'm not sure it it needs an extra space or there is an extra space already.

screen shot 2017-12-11 at 13 58 14

If you have a case inside a case, the inner case clauses indentation is just fine:

screen shot 2017-12-11 at 14 02 28

### Environment

```
Erlang/OTP 20 [erts-9.1.5] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:10] [hipe] [kernel-poll:false] [dtrace]

Elixir 1.6.0-dev (840c995) (compiled with OTP 20)
```

* Operating system: MacOS

josevalim:
Closes #7102

Please use the :+1: and :-1: reaction accordingly.

### DIFF
--- a/lib/elixir/lib/base.ex
+++ b/lib/elixir/lib/base.ex
@@ -187,12 +187,12 @@ defmodule Base do
     alphabet
     |> Stream.with_index()
     |> Enum.flat_map(fn {encoding, value} = pair ->
-         if encoding in ?A..?Z do
-           [pair, {encoding - ?A + ?a, value}]
-         else
-           [pair]
-         end
-       end)
+      if encoding in ?A..?Z do
+        [pair, {encoding - ?A + ?a, value}]
+      else
+        [pair]
+      end
+    end)
     |> decode_clauses()
   end
 
--- a/lib/elixir/lib/code/formatter.ex
+++ b/lib/elixir/lib/code/formatter.ex
@@ -699,10 +699,7 @@ defmodule Code.Formatter do
           # we cannot group the left side yet.
           left = if op_info == parent_info, do: left, else: group(left)
 
-          # The right side is never the same operator, so we can group and nest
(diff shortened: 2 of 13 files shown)


eksperimental:
I would go with option 1 mentioned in the referenced issue

NobbZ:
Can you please explain where exactly we are supposed to up or downvote?

OvermindDL1:
> Can you please explain where exactly we are supposed to up or downvote?

@NobbZ In the parent post.

yordis:
Indent nothing

```
Account.create_user(user_params)
|> case do
  {:ok, %User{} = user} ->
    user

  {:error, other} ->
    other
end
```

yordis:
Indent by the operator

```elixir
Account.create_user(user_params)
|> case do
     {:ok, %User{} = user} ->
       user

     {:error, other} ->
       other
   end
```

eksperimental:
@NobbZ reactions are not allowed in the mobile interfelace, that's probably why you cannot see the option

NobbZ:
@eksperimental I'm aware of that, but from the linked issue I know about 3 variations, and now I want to know where I need to cast my vote for option 2, since in the other issue all 3 are presented in a single post and this PR does implement one and in the huge diff which is displayed clunky on my mobile device I'm not even sure which option it is.

josevalim:
@NobbZ option 2 is what this PR is implementing, so basically a 👍 .

amrnt:
❤️💜💚

Issue #7104, opened 2017-12-11
Introduce @deprecated module attribute


josevalim:
The plan is to use it alongside xref and print compile time warnings to deprecated functions. Docs can use it in the future too.

ggcampinho:
Should this be stored in the defs table on the key `{def, {module, arity}}`?
I think I can do it

josevalim:
@ggcampinho please go ahead! Some notes:

1. During compilation we can store it as we store documentation using a `{{deprecated, {name, arity}}, deprecation_reason}` entry or similar (I can't quite recall how we store docs).

2. We need to validate the reason is a string (similar to how we validate `@since`)

3. When emitting the Erlang bytecode, we will add an extra chunk called "ExDp" that will return `{{name, arity}, reason}`. We will also store it under `__info__(:deprecated)`.

4. We will change xref to invoke `__info__(:deprecated)` or read from the chunk and emit the proper warnings.

We can probably do this over multiple PRs. One PR could do 1 to 3, another could do 4.

whatyouhide:
@josevalim did we decide we want to store the deprecation reason in the chunk? I still think storing the version it was deprecated in in the chunk and just mentioning the reason in the documentation might be a better idea. However I wonder if this would mean that we would not do `@doc false` anymore on deprecated functions.

josevalim:
@whatyouhide I think storing the reason is better because that is more immediately useful for developers for warnings coming from xref.

Compare:

    foo.ex:3: String.contain?/2 was deprecated. Use String.contains?/2 instead.

with:

    foo.ex:3: String.contain?/2 was deprecated (on version v1.3.2)

And I would say the version it was deprecated is not very important. The version the replacement was introduced and the version it will be removed (if one is scheduled) are more important.

whatyouhide:
@josevalim okay, fair enough. 👍

michalmuskala:
One thing to consider is that both this attribute and `@since` might be problematic to libraries that want to be compatible with multiple elixir versions because of the unused attribute warnings. I'm not sure what would be the best course of action here.

whatyouhide:
@michalmuskala they can have a `_ = @since` to be compatible without warnings?

josevalim:
I have also pushed a commit to v1.5 so it doesn't warn there. We should have a new release today or tomorrow.

ggcampinho:
[...]

PR #7113, opened 2017-12-14
Introduce @deprecated

ggcampinho:
Add deprecated to `ExDp` beam chunk and to `Module.__info__(:deprecated)` as discussed on #7104

### DIFF
--- a/lib/elixir/lib/module.ex
+++ b/lib/elixir/lib/module.ex
@@ -104,6 +104,17 @@ defmodule Module do
   Multiple uses of `@compile` will accumulate instead of overriding
   previous ones. See the "Compile options" section below.
 
+  ### `@deprecated`
+
+  Provides the deprecation reason for a function. For example:
+
+      defmodule MyModule do
+        @deprecated "Use MyModule.hello_world/0 instead"
+        def hello, do: "world"
+
+        def hello_world, do: "Hello world!"
+      end
+
   ### `@doc` (and `@since`)
 
   Provides documentation for the function or macro that follows the
@@ -1218,6 +1229,22 @@ defmodule Module do
   end
 
   @doc false
+  # Used internally to compile deprecated.
+  # This function is private and must be used only internally.
+  def compile_deprecated(env, _kind, name, args, _guards, _body) do
+    module = env.module
+    table = data_table_for(module)
+    arity = length(args)
+    pair = {name, arity}
+
+    if reason = get_deprecated_info(table) do
+      :ets.insert(table, {{:deprecated, pair}, reason})
+    end
+
+    :ok
+  end
+
+  @doc false
   # Used internally to check the validity of arguments to @impl.
   # This function is private and must be used only internally.
   def compile_impl(env, kind, name, args, _guards, _body) do
@@ -1686,6 +1713,12 @@ defmodule Module do
             "the version a function, macro, type or callback was added, got: #{inspect(value)}"
   end
 
+  defp preprocess_attribute(:deprecated, value) when not is_binary(value) do
+    raise ArgumentError,
+          "@deprecated expects a string with the reason for the deprecation, " <>
+            "got: #{inspect(value)}"
+  end
+
   defp preprocess_attribute(_key, value) do
     value
   end
@@ -1705,6 +1738,13 @@ defmodule Module do
 
   defp get_since_info(table) do
     :ets.take(table, :since)
+  end
+
+  defp get_deprecated_info(table) do
+    case :ets.take(table, :deprecated) do
+      [{:deprecated, reason, _, _}] -> reason
+      [] -> nil
+    end
   end
 
   defp data_table_for(module) do
--- a/lib/elixir/lib/module.ex
+++ b/lib/elixir/lib/module.ex
@@ -108,11 +108,11 @@ defmodule Module do
 
   Provides the deprecation reason for a function. For example:
 
-      defmodule MyModule do
-        @deprecated "Use MyModule.hello_world/0 instead"
-        def hello, do: "world"
-
-        def hello_world, do: "Hello world!"
+      defmodule Keyword do
+        @deprecated "Use Kernel.length/1 instead"
+        def size(keyword) do
+          length(keyword)
+        end
       end
 
   ### `@doc` (and `@since`)
--- a/lib/elixir/lib/module.ex
+++ b/lib/elixir/lib/module.ex
@@ -1188,7 +1188,20 @@ defmodule Module do
   @doc false
   # Used internally to compile documentation.
   # This function is private and must be used only internally.
-  def compile_doc(table, pair, env, kind, args) do
+  def compile_definition_attributes(env, kind, name, args, _guards, _body) do
+    module = env.module
+    table = data_table_for(module)
+    arity = length(args)
+    pair = {name, arity}
+
+    # Docs must come first as they read the impl callback
+    compile_doc(table, pair, env, kind, args)
+    compile_deprecated(table, pair)
+    compile_impl(table, name, env, kind, args)
+    :ok
+  end
+
+  defp compile_doc(table, pair, env, kind, args) do
     {line, doc} = get_doc_info(table, env)
 
     # TODO: Store @since alongside the docs
@@ -1222,19 +1235,13 @@ defmodule Module do
     end
   end
 
-  @doc false
-  # Used internally to compile deprecated.
-  # This function is private and must be used only internally.
-  def compile_deprecated(table, pair) do
+  defp compile_deprecated(table, pair) do
     if reason = get_deprecated_info(table) do
       :ets.insert(table, {{:deprecated, pair}, reason})
     end
   end
 
-  @doc false
-  # Used internally to check the validity of arguments to @impl.
-  # This function is private and must be used only internally.
-  def compile_impl(table, name, env, kind, args) do
+  defp compile_impl(table, name, env, kind, args) do
     %{line: line, file: file} = env
 
     case :ets.take(table, :impl) do
--- a/lib/elixir/lib/module.ex
+++ b/lib/elixir/lib/module.ex
@@ -1188,22 +1188,16 @@ defmodule Module do
   @doc false
   # Used internally to compile documentation.
   # This function is private and must be used only internally.
-  def compile_doc(env, kind, name, args, _guards, _body) do
-    module = env.module
-    table = data_table_for(module)
-    arity = length(args)
-    pair = {name, arity}
-
+  def compile_doc(table, pair, env, kind, args) do
     {line, doc} = get_doc_info(table, env)
 
     # TODO: Store @since alongside the docs
     _ = get_since_info(table)
 
     add_doc(table, line, kind, pair, args, doc, env)
-    :ok
   end
 
-  defp add_doc(_module, line, kind, {name, arity}, _args, doc, env)
+  defp add_doc(_table, line, kind, {name, arity}, _args, doc, env)
        when kind in [:defp, :defmacrop] do
     if doc do
       error_message =
@@ -1231,25 +1225,17 @@ defmodule Module do
   @doc false
   # Used internally to compile deprecated.
   # This function is private and must be used only internally.
-  def compile_deprecated(env, _kind, name, args, _guards, _body) do
-    module = env.module
-    table = data_table_for(module)
-    arity = length(args)
-    pair = {name, arity}
-
+  def compile_deprecated(table, pair) do
     if reason = get_deprecated_info(table) do
       :ets.insert(table, {{:deprecated, pair}, reason})
     end
-
-    :ok
   end
 
   @doc false
   # Used internally to check the validity of arguments to @impl.
   # This function is private and must be used only internally.
-  def compile_impl(env, kind, name, args, _guards, _body) do
-    %{module: module, line: line, file: file} = env
-    table = data_table_for(module)
+  def compile_impl(table, name, env, kind, args) do
+    %{line: line, file: file} = env
 
     case :ets.take(table, :impl) do
       [{:impl, value, _, _}] ->
@@ -1263,8 +1249,6 @@ defmodule Module do
       [] ->
         :ok
     end
-
-    :ok
   end
 
   defp args_count([{:\\, _, _} | tail], total, defaults) do
--- a/lib/elixir/lib/protocol.ex
+++ b/lib/elixir/lib/protocol.ex
@@ -309,7 +309,7 @@ defmodule Protocol do
   end
 
   defp beam_protocol(protocol) do
-    chunk_ids = [:abstract_code, :attributes, :compile_info, 'ExDc']
+    chunk_ids = [:abstract_code, :attributes, :compile_info, 'ExDc', 'ExDp']
     opts = [:allow_missing_chunks]
 
     case :beam_lib.chunks(beam_file(protocol), chunk_ids, opts) do
@@ -318,12 +318,13 @@ defmodule Protocol do
           {:abstract_code, {_raw, abstract_code}},
           {:attributes, attributes},
           {:compile_info, compile_info},
-          {'ExDc', docs}
+          {'ExDc', docs},
+          {'ExDp', deprecated}
         ] = entries
 
         case attributes[:protocol] do
           [fallback_to_any: any] ->
-            {:ok, {protocol, any, abstract_code}, {compile_info, docs}}
+            {:ok, {protocol, any, abstract_code}, {compile_info, docs, deprecated}}
 
           _ ->
             {:error, :not_a_protocol}
@@ -497,14 +498,14 @@ defmodule Protocol do
   end
 
   # Finally compile the module and emit its bytecode.
-  defp compile(protocol, code, {compile_info, docs}) do
+  defp compile(protocol, code, {compile_info, docs, deprecated}) do
     opts = Keyword.take(compile_info, [:source])
     opts = if Code.compiler_options()[:debug_info], do: [:debug_info | opts], else: opts
     {:ok, ^protocol, binary, _warnings} = :compile.forms(code, [:return | opts])
 
     case docs do
       :missing_chunk -> {:ok, binary}
-      _ -> {:ok, :elixir_erl.add_beam_chunks(binary, [{"ExDc", docs}])}
+      _ -> {:ok, :elixir_erl.add_beam_chunks(binary, [{"ExDc", docs}, {"ExDp", deprecated}])}
     end
   end
 
--- a/lib/elixir/lib/protocol.ex
+++ b/lib/elixir/lib/protocol.ex
@@ -309,7 +309,7 @@ defmodule Protocol do
   end
 
   defp beam_protocol(protocol) do
-    chunk_ids = [:abstract_code, :attributes, :compile_info, 'ExDc', 'ExDp']
+    chunk_ids = [:abstract_code, :attributes, :compile_info, 'ExDc']
     opts = [:allow_missing_chunks]
 
     case :beam_lib.chunks(beam_file(protocol), chunk_ids, opts) do
@@ -318,13 +318,12 @@ defmodule Protocol do
           {:abstract_code, {_raw, abstract_code}},
           {:attributes, attributes},
           {:compile_info, compile_info},
-          {'ExDc', docs},
-          {'ExDp', deprecated}
+          {'ExDc', docs}
         ] = entries
 
         case attributes[:protocol] do
           [fallback_to_any: any] ->
-            {:ok, {protocol, any, abstract_code}, {compile_info, docs, deprecated}}
+            {:ok, {protocol, any, abstract_code}, {compile_info, docs}}
 
           _ ->
             {:error, :not_a_protocol}
@@ -498,14 +497,14 @@ defmodule Protocol do
   end
 
   # Finally compile the module and emit its bytecode.
-  defp compile(protocol, code, {compile_info, docs, deprecated}) do
+  defp compile(protocol, code, {compile_info, docs}) do
     opts = Keyword.take(compile_info, [:source])
     opts = if Code.compiler_options()[:debug_info], do: [:debug_info | opts], else: opts
     {:ok, ^protocol, binary, _warnings} = :compile.forms(code, [:return | opts])
 
     case docs do
       :missing_chunk -> {:ok, binary}
-      _ -> {:ok, :elixir_erl.add_beam_chunks(binary, [{"ExDc", docs}, {"ExDp", deprecated}])}
+      _ -> {:ok, :elixir_erl.add_beam_chunks(binary, [{"ExDc", docs}])}
     end
   end
 
--- a/lib/elixir/src/elixir_erl.erl
+++ b/lib/elixir/src/elixir_erl.erl
@@ -22,10 +22,10 @@ debug_info(elixir_v1, _Module, none, _Opts) ->
 debug_info(elixir_v1, _Module, {elixir_v1, Map, _Specs}, _Opts) ->
   {ok, Map};
 debug_info(erlang_v1, _Module, {elixir_v1, Map, Specs}, _Opts) ->
-  {Prefix, Forms, _, _} = dynamic_form(Map, []),
+  {Prefix, Forms, _, _} = dynamic_form(Map),
   {ok, Prefix ++ Specs ++ Forms};
 debug_info(core_v1, _Module, {elixir_v1, Map, Specs}, Opts) ->
-  {Prefix, Forms, _, _} = dynamic_form(Map, []),
+  {Prefix, Forms, _, _} = dynamic_form(Map),
   #{compile_opts := CompileOpts} = Map,
 
   %% Do not rely on elixir_erl_compiler because we don't
@@ -149,8 +149,7 @@ definition_scope(Meta, File) ->
 
 compile(#{module := Module} = Map) ->
   Data = elixir_module:data_table(Module),
-  Deprecated = get_deprecated(Data),
-  {Prefix, Forms, Defmacro, Unreachable} = dynamic_form(Map, Deprecated),
+  {Prefix, Forms, Defmacro, Unreachable} = dynamic_form(Map),
   Specs =
     case elixir_config:get(bootstrap) of
       true -> [];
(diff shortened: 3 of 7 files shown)


lib/elixir/src/elixir_erl.erl
```diff
@@ -21,11 +21,13 @@ debug_info(elixir_v1, _Module, none, _Opts) ->
...
-debug_info(erlang_v1, _Module, {elixir_v1, Map, Specs}, _Opts) ->
-  {Prefix, Forms, _, _} = dynamic_form(Map),
+debug_info(erlang_v1, Module, {elixir_v1, Map, Specs}, _Opts) ->
+  Data = elixir_module:data_table(Module),
+  {Prefix, Forms, _, _} = dynamic_form(Map, Data),
   {ok, Prefix ++ Specs ++ Forms};
-debug_info(core_v1, _Module, {elixir_v1, Map, Specs}, Opts) ->
-  {Prefix, Forms, _, _} = dynamic_form(Map),
+debug_info(core_v1, Module, {elixir_v1, Map, Specs}, Opts) ->
+  Data = elixir_module:data_table(Module),
```
ggcampinho: I'm having some problems with the chunks in protocols, I believe is related to the way that the `debug_info` is saved. I'm checking this

lib/elixir/src/elixir_module.erl
```diff
@@ -176,20 +180,15 @@ build(Line, File, Module, Lexical) ->
...
+  Bootstrap = elixir_config:get(bootstrap),
+  CompileDocs = elixir_compiler:get_opt(docs),
 
   %% Docs must come first as they read the impl callback.
-  OnDefinition = DocsOnDefinition ++ ImplOnDefinition,
+  OnDefinition = [
+    deprecated_on_definition(not Bootstrap),
+    docs_on_definition(CompileDocs),
+    impl_on_definition(not Bootstrap)
+  ],
```
josevalim: Let's have a general `compile_definition_attributes` function and from there we can call `compile_deprecated` and `compile_impl` but sharing something things, such as the arity and the data table.

josevalim:
@ggcampinho the test is failing because we are introducing a new atom in the module and that changes the atom hash. Updating the hash is OK and should fix the build.

lib/elixir/src/elixir_erl.erl
```diff
@@ -22,10 +22,10 @@ debug_info(elixir_v1, _Module, none, _Opts) ->
 debug_info(elixir_v1, _Module, {elixir_v1, Map, _Specs}, _Opts) ->
   {ok, Map};
 debug_info(erlang_v1, _Module, {elixir_v1, Map, Specs}, _Opts) ->
-  {Prefix, Forms, _, _} = dynamic_form(Map),
+  {Prefix, Forms, _, _} = dynamic_form(Map, []),
```
ggcampinho: There are some problems with the protocol consolidation due to this. I'm still exploring how to solve this, but I think we need to add something to the tuple `{elixir_v1, Map, Specs}`, like `{elixir_v1, Map, Specs, Deprecated}` (and maybe bump the version), or add them to the `Map` itself. Otherwise, after consolidation `Module.__info__(:deprecated)` always return `[]`.

Another possibility is to send the extra chunks via `Opts` and recover the deprecated from them, but I'm not sure if this will work. I'll try out the different things that I have in mind to check.

lib/elixir/test/elixir/kernel/deprecated_test.exs
```diff
@@ -0,0 +1,38 @@
...
+
+defmodule Kernel.DeprecatedTest do
+  use ExUnit.Case
+
+  import PathHelpers
+
+  test "raises on invalid @deprecated" do
+    assert_raise ArgumentError, ~r"@deprecated expects a string with the reason", fn ->
+      defmodule InvalidDeprecated do
+        @deprecated 1.2
```
ggcampinho: One extra comment here, I would expect some people only using `@deprecated` without any argument and this was my first test case, but I think the warning is not very good/clear in the case:

```
warning: module attribute @deprecated in code block has no effect as it is never returned (remove the attribute or assign it to _ to avoid warnings)
  lib/elixir/test/elixir/kernel/deprecated_test.exs:11

warning: undefined module attribute @deprecated, please remove access to @deprecated or explicitly set it before access
  lib/elixir/test/elixir/kernel/deprecated_test.exs:11: Kernel.DeprecatedTest.InvalidDeprecated (module)
```

I don't if it makes sense to improve this message specifically, to something like:

```
warning: @deprecated expects a string with the reason for the deprecation, got: nil
```

lib/elixir/lib/protocol.ex
```diff
@@ -309,7 +309,7 @@ defmodule Protocol do
   end
 
   defp beam_protocol(protocol) do
-    chunk_ids = [:abstract_code, :attributes, :compile_info, 'ExDc']
+    chunk_ids = [:abstract_code, :attributes, :compile_info, 'ExDc', 'ExDp']
```
josevalim: Yup, this is the way to go!

lib/elixir/src/elixir_module.erl
```diff
@@ -229,6 +219,28 @@ build(Line, File, Module, Lexical) ->
 
   {Data, Defs, Ref}.
 
+compile_definition_attributes(#{module := Module} = Env, Kind, Name, Args, _Guards, _Body) ->
```
josevalim: We can move this logic to the `Module` module.

lib/elixir/src/elixir_module.erl
```diff
@@ -229,6 +219,28 @@ build(Line, File, Module, Lexical) ->
...
+  Data = data_table(Module),
+  Arity = length(Args),
+  Tuple = {Name, Arity},
+
+  %% Docs must come first as they read the impl callback
+  case elixir_compiler:get_opt(docs) of
```
josevalim: Let's skip this check. We already skip adding the chunk later in `elixir_erl`, this way we can hide the compilation of all attributes behind the bootstrap flag.

lib/elixir/test/elixir/kernel/deprecated_test.exs
josevalim: > I don't if it makes sense to improve this message specifically, to something like:

I don't think it makes sense because differently from other languages `@foobar` in Elixir is always a read operation, never a write one.

josevalim:
I am not a big fan of storing the deprecated info in the debug info map because it means the deprecated annotation is in three places in the chunk:

  1. Inside the __info__ function
  2. Inside the ExDp chunk
  3. Inside the debug_info

At the same time, I don't understand why it is being discarded when we consolidate. Don't we keep all of the `__info__` clauses as is when we consolidate?

lib/elixir/src/elixir_module.erl
```diff
@@ -32,13 +31,17 @@ defs_table(Module) ->
...
+  ets:delete(Data, doc).
 
-delete_impl(#{module := Module}, _, _, _, _, _) ->
-  ets:delete(data_table(Module), impl),
-  ok.
+delete_deprecated(Data) ->
+  ets:delete(Data, deprecated).
+
+get_deprecated(Data) ->
+  lists:usort(ets:select(Data, [{{{deprecated, '$1'}, '$2'}, [], [{{'$1', '$2'}}]}])).
```
josevalim: So to clarify the comment in the main thread, I prefer this as it was, happening in `elixir_erl`, because we don't want to store the deprecated stuff in three different places in the chunk.

lib/elixir/test/elixir/kernel/deprecated_test.exs
ggcampinho: Yeah, this is why I didn't change. But if you are not used to the language, you can spend some time trying to understand what is exactly that. And to be fair, it's written on the second warning, it's just that you maybe get confused with it saying that the attribute is undefined and them talking about access. For me, maybe it would more clear if the warning was something like that:
```
warning: you are trying to use @deprecated, but it was never explicitly set before. If you are trying to set the attribute, please define its value after the attribute. Otherwise, remove the access to it.
```

But, in this case specifically, I was thinking that I did something wrong with the reading of the attribute, so I was reading "undefined module attribute @deprecated" like a different error. Maybe the warning is clear enough..

lib/elixir/src/elixir_module.erl
josevalim: So in a nutshell, we will have `Module.compile_definition_attributes` and `:elixir_module.delete_definition_attributes`. We will specify which one to call based on the bootstrap value when the module is defined ([the previous OnDefinition field here](https://github.com/elixir-lang/elixir/pull/7113/files#diff-b11934843d48f08ba2bdb4d4965b812bL204)).

lib/elixir/test/elixir/kernel/deprecated_test.exs
josevalim: @ggcampinho we can definitely rewrite the warning although I still think we should start with "undefined module attribute @deprecated". The initial reason should be compact, so those familiar with the warning can know what happened without reading the whole message. Then we expand the message.

So maybe something like:

```
undefined module attribute @deprecated. If you are trying to set the attribute, a value must be explicitly given as argument, otherwise remove the use of @deprecated
```

ggcampinho:
I have the same issue with the debug_info, I don't understand exactly what is happening because I don't understand the compilation so deeply, but the module is purged before the consolidation. If I try to print the value of `__info__(:deprecated)` it will show the reason inside of it, but if I print what is inside of the code in the [consolidation](https://github.com/elixir-lang/elixir/blob/master/lib/elixir/lib/protocol.ex#L318) it will be:

```
[
  {:abstract_code,
   {:raw_abstract_v1,
    [
      {:attribute, 267, :file,
       {'lib/elixir/test/elixir/protocol_test.exs', 267}},
      {:attribute, 267, :module, Protocol.ConsolidationTest.Sample},
      {:attribute, 267, :compile, :no_auto_import},
      {:attribute, 271, :callback,
       {{:ok, 1},
        [
          {:type, 271, :fun,
           [
             {:type, 271, :product, [{:user_type, 271, :t, []}]},
             {:type, 271, :boolean, []}
           ]}
        ]}},
      {:attribute, 271, :spec,
       {{:ok, 1},
        [
          {:type, 271, :fun,
           [
             {:type, 271, :product, [{:user_type, 271, :t, []}]},
             {:type, 271, :boolean, []}
           ]}
        ]}},
      {:attribute, 267, :spec,
       {{:impl_for!, 1},
        [
          {:type, 267, :fun,
           [
             {:type, 267, :product, [{:type, 267, :term, []}]},
             {:type, 267, :union,
              [{:type, 267, :atom, []}, {:type, 267, :no_return, []}]}
           ]}
        ]}},
      {:attribute, 267, :spec,
       {{:impl_for, 1},
        [
          {:type, 267, :fun,
           [
             {:type, 267, :product, [{:type, 267, :term, []}]},
             {:type, 267, :union, [{:type, 267, :atom, []}, {:atom, 0, nil}]}
           ]}
        ]}},
      {:attribute, 267, :spec,
       {{:__protocol__, 1},
        [
          {:type, 267, :fun,
           [
             {:type, 267, :product, [{:atom, 0, :module}]},
             {:atom, 0, Protocol.ConsolidationTest.Sample}
           ]},
          {:type, 267, :fun,
           [
             {:type, 267, :product, [{:atom, 0, :functions}]},
             {:type, 0, :nonempty_list,
              [{:type, 0, :tuple, [{:atom, 0, :ok}, {:integer, 0, 1}]}]}
           ]},
          {:type, 267, :fun,
           [
             {:type, 267, :product, [{:atom, 0, :consolidated?}]},
             {:atom, 0, false}
           ]},
          {:type, 267, :fun,
           [
             {:type, 267, :product, [{:atom, 0, :impls}]},
             {:atom, 0, :not_consolidated}
           ]}
        ]}},
      {:attribute, 268, :export_type, [t: 0]},
      {:attribute, 268, :type, {:t, {:type, 268, :any, []}, []}},
      {:attribute, 267, :compile, :debug_info},
      {:attribute, 267, :compile,
       {:inline, [any_impl_for: 0, struct_impl_for: 1, impl_for?: 1]}},
      {:attribute, 267, :protocol, [fallback_to_any: false]},
      {:attribute, 267, :export,
       [__info__: 1, __protocol__: 1, impl_for: 1, impl_for!: 1, ok: 1]},
      {:attribute, 267, :spec,
       {{:__info__, 1},
        [
          {:type, 267, :fun,
           [
             {:type, 267, :product,
              [
                {:type, 267, :union,
                 [
                   {:atom, 267, :attributes},
                   {:atom, 267, :compile},
                   {:atom, 267, :functions},
                   {:atom, 267, :macros},
                   {:atom, 267, :md5},
                   {:atom, 267, :module}
                 ]}
              ]},
             {:type, 267, :union,
              [
                {:type, 267, :atom, []},
                {:type, 267, :list,
                 [{:type, 267, :union, [{:type, ...}, {...}]}]}
              ]}
           ]}
        ]}},
      {:function, 0, :__info__, 1,
       [
         {:clause, 0, [{:atom, 0, :module}], [],
          [{:atom, 0, Protocol.ConsolidationTest.Sample}]},
         {:clause, 0, [{:atom, 0, :functions}], [],
          [
            {:cons, 0,
             {:tuple, 0, [{:atom, 0, :__protocol__}, {:integer, 0, 1}]},
             {:cons, 0, {:tuple, 0, [{:atom, 0, :impl_for}, {:integer, 0, 1}]},
              {:cons, 0, {:tuple, 0, [{:atom, ...}, {...}]},
               {:cons, 0, {:tuple, ...}, {...}}}}}
          ]},
         {:clause, 0, [{:atom, 0, :macros}], [], [nil: 0]},
         {:clause, 0, [{:atom, 0, :attributes}], [],
          [
            {:call, 0,
             {:remote, 0, {:atom, 0, :erlang}, {:atom, 0, :get_module_info}},
             [
               {:atom, 0, Protocol.ConsolidationTest.Sample},
               {:atom, 0, :attributes}
             ]}
          ]},
         {:clause, 0, [{:atom, 0, :compile}], [],
          [
            {:call, 0,
             {:remote, 0, {:atom, 0, :erlang}, {:atom, 0, :get_module_info}},
             [
               {:atom, 0, Protocol.ConsolidationTest.Sample},
               {:atom, 0, :compile}
             ]}
          ]},
         {:clause, 0, [{:atom, 0, :md5}], [],
          [
            {:call, 0,
             {:remote, 0, {:atom, 0, :erlang}, {:atom, 0, :get_module_info}},
             [{:atom, 0, Protocol.ConsolidationTest.Sample}, {:atom, 0, :md5}]}
          ]},
         {:clause, 0, [{:atom, 0, :deprecated}], [], [nil: 0]}
       ]},
      {:function, 267, :__protocol__, 1,
       [
         {:clause, 267, [{:atom, 0, :module}], [],
          [{:atom, 0, Protocol.ConsolidationTest.Sample}]},
         {:clause, 267, [{:atom, 0, :functions}], [],
          [
            {:cons, 0, {:tuple, 0, [{:atom, 0, :ok}, {:integer, 0, 1}]},
             {nil, 0}}
          ]},
         {:clause, 267, [{:atom, 0, :consolidated?}], [], [{:atom, 0, false}]},
         {:clause, 267, [{:atom, 0, :impls}], [],
          [{:atom, 0, :not_consolidated}]}
       ]},
      {:function, 267, :any_impl_for, 0,
       [{:clause, 267, [], [], [{:atom, 0, nil}]}]},
      {:function, 267, :impl_for, 1,
       [
         {:clause, 267,
          [
            {:map, 267,
             [
               {:map_field_exact, 267, {:atom, 0, :__struct__},
                {:var, 267, :_@1}}
             ]}
          ],
          [
            [
              {:call, 267,
               {:remote, 267, {:atom, 0, :erlang}, {:atom, 267, :is_atom}},
               [{:var, 267, :_@1}]}
            ]
          ],
          [{:call, 267, {:atom, 267, :struct_impl_for}, [{:var, 267, :_@1}]}]},
         {:clause, 267, [{:var, 267, :_@1}],
          [
            [
              {:call, 267,
               {:remote, 267, {:atom, 0, :erlang}, {:atom, 267, :is_tuple}},
               [{:var, 267, :_@1}]}
            ]
          ],
          [
            {:case, 267,
             {:call, 267, {:atom, 267, :impl_for?},
              [{:atom, 0, Protocol.ConsolidationTest.Sample.Tuple}]},
             [
               {:clause, 267, [{:atom, 0, true}], [], [{:call, 267, ...}]},
               {:clause, 267, [{:atom, 0, false}], [], [{:call, ...}]}
             ]}
          ]},
         {:clause, 267, [{:var, 267, :_@1}],
          [
            [
              {:call, 267,
               {:remote, 267, {:atom, 0, :erlang}, {:atom, 267, :is_atom}},
               [{:var, 267, :_@1}]}
            ]
          ],
          [
            {:case, 267,
             {:call, 267, {:atom, 267, :impl_for?},
              [{:atom, 0, Protocol.ConsolidationTest.Sample.Atom}]},
             [
               {:clause, 267, [{:atom, 0, true}], [], [{:call, ...}]},
               {:clause, 267, [{:atom, 0, ...}], [], [{...}]}
             ]}
          ]},
         {:clause, 267, [{:var, 267, :_@1}],
          [
            [
              {:call, 267,
               {:remote, 267, {:atom, 0, :erlang}, {:atom, 267, :is_list}},
               [{:var, 267, :_@1}]}
            ]
          ],
          [
            {:case, 267,
             {:call, 267, {:atom, 267, :impl_for?},
              [{:atom, 0, Protocol.ConsolidationTest.Sample.List}]},
             [
               {:clause, 267, [{:atom, 0, ...}], [], [{...}]},
               {:clause, 267, [{:atom, ...}], [], [...]}
             ]}
          ]},
         {:clause, 267, [{:var, 267, :_@1}],
          [
            [
              {:call, 267,
               {:remote, 267, {:atom, 0, :erlang}, {:atom, 267, :is_map}},
               [{:var, 267, :_@1}]}
            ]
          ],
          [
            {:case, 267,
             {:call, 267, {:atom, 267, :impl_for?}, [{:atom, 0, ...}]},
             [
               {:clause, 267, [{:atom, ...}], [], [...]},
               {:clause, 267, [{...}], [], ...}
             ]}
          ]},
         {:clause, 267, [{:var, 267, :_@1}],
          [
            [
              {:call, 267,
               {:remote, 267, {:atom, 0, :erlang}, {:atom, 267, ...}},
               [{:var, 267, :_@1}]}
            ]
          ],
          [
            {:case, 267, {:call, 267, {:atom, 267, :impl_for?}, [{:atom, ...}]},
             [{:clause, 267, [{...}], [], ...}, {:clause, 267, [...], ...}]}
          ]},
         {:clause, 267, [{:var, 267, :_@1}],
          [
            [
              {:call, 267, {:remote, 267, {:atom, 0, ...}, {:atom, ...}},
               [{:var, 267, :_@1}]}
            ]
          ],
          [
            {:case, 267, {:call, 267, {:atom, 267, ...}, [{...}]},
             [{:clause, 267, [...], ...}, {:clause, 267, ...}]}
          ]},
         {:clause, 267, [{:var, 267, :_@1}],
          [
            [
              {:call, 267, {:remote, 267, {:atom, ...}, {...}},
               [{:var, 267, ...}]}
            ]
          ],
          [
            {:case, 267, {:call, 267, {:atom, ...}, [...]},
             [{:clause, 267, ...}, {:clause, ...}]}
          ]},
         {:clause, 267, [{:var, 267, :_@1}],
          [[{:call, 267, {:remote, 267, {...}, ...}, [{:var, ...}]}]],
          [{:case, 267, {:call, 267, {...}, ...}, [{:clause, ...}, {...}]}]},
         {:clause, 267, [{:var, 267, :_@1}],
          [[{:call, 267, {:remote, 267, ...}, [{...}]}]],
          [{:case, 267, {:call, 267, ...}, [{...}, ...]}]},
         {:clause, 267, [{:var, 267, :_@1}],
          [[{:call, 267, {:remote, ...}, [...]}]],
          [{:case, 267, {:call, ...}, [...]}]},
         {:clause, 267, [{:var, 267, :_@1}], [[{:call, 267, {...}, ...}]],
          [{:case, 267, {...}, ...}]},
         {:clause, 267, [{:var, 267, :_}], [], [{:call, 267, ...}]}
       ]},
      {:function, 267, :impl_for!, 1,
       [
         {:clause, 267, [{:var, 267, :_@1}], [],
          [
            {:case, 267,
             {:call, 267, {:atom, 267, :impl_for}, [{:var, 267, :_@1}]},
             [
               {:clause, 267, [{:var, 267, :_@2}], [[{:op, 267, ...}]],
                [{:call, 267, ...}]},
               {:clause, 267, [{:var, 267, :_@3}], [], [{:var, ...}]}
             ]}
          ]}
       ]},
      {:function, 267, :impl_for?, 1,
       [
         {:clause, 267, [{:var, 267, :_@1}], [],
          [
            {:case, 267,
             {:call, 267, {:remote, 267, {:atom, 0, Code}, {:atom, 267, ...}},
              [{:var, 267, :_@1}]},
             [
               {:clause, [generated: true, location: 267], [{:atom, 0, true}],
                [], [{:call, ...}]},
               {:clause, [generated: true, location: 267], [{:atom, 0, ...}],
                [], [{...}]},
               {:clause, [generated: true, location: 267], [{:var, ...}], [],
                [...]}
             ]}
          ]}
       ]},
      {:function, 272, :ok, 1,
       [
         {:clause, 272, [{:var, 272, :_@1}], [],
          [
            {:call, 272,
             {:remote, 272, {:call, 272, {:atom, 272, ...}, [{...}]},
              {:atom, 272, :ok}}, [{:var, 272, :_@1}]}
          ]}
       ]},
      {:function, 267, :struct_impl_for, 1,
       [
         {:clause, 267, [{:var, 267, :_@1}], [],
          [
            {:match, 267, {:var, 267, :_@2},
             {:call, 267, {:remote, 267, {...}, ...}, [{:atom, ...}, {...}]}},
            {:case, 267, {:call, 267, {:atom, 267, :impl_for?}, [{:var, ...}]},
             [{:clause, 267, [{...}], [], ...}, {:clause, 267, [...], ...}]}
          ]}
       ]}
    ]}},
  {:attributes,
   [
     protocol: [fallback_to_any: false],
     vsn: [250587113122219927348857560523310482569]
   ]},
  {:compile_info,
   [
     options: [],
     version: '7.1',
     source: '/Users/glaubercampinho/work/elixir/lib/elixir/test/elixir/protocol_test.exs'
   ]},
  {'ExDc',
   <<131, 80, 0, 0, 1, 111, 120, 156, 149, 144, 77, 14, 194, 32, 16, 133, 135,
     254, 24, 141, 173, 119, 240, 8, 30, 194, 181, 71, 32, 20, 104, 160, 29,
     164, 161, 104, 244, 246, 242, 35, 155, 38, 46, 220, ...>>},
  {'ExDp',
   <<131, 104, 2, 100, 0, 20, 101, 108, 105, 120, 105, 114, 95, 100, 101, 112,
     114, 101, 99, 97, 116, 101, 100, 95, 118, 49, 108, 0, 0, 0, 1, 104, 2, 104,
     2, 100, 0, 2, 111, 107, 97, 1, 109, ...>>}
]
```

If you check, inside of the `__info__(:deprecated)` we have `[]`, even that `ExDp` is set. In some point [this callback](https://github.com/elixir-lang/elixir/blob/master/lib/elixir/src/elixir_erl.erl#L25) is called and the `__info__` is reconstructed, this is the part that I got a bit lost.

ggcampinho:
I also can inject the `__info__(:deprecated)` in the module `Protocol`, but I don't think this will be nice

josevalim:
@ggcampinho I see. That makes sense. Let me think about it more and I will follow up later.

ggcampinho:
I think it can be done in the `Protocol` if `consolidate` get the debug info and call `form`, I'll try.

ggcampinho:
@josevalim can you take another look?

lib/elixir/src/elixir_erl.erl
```diff
@@ -21,11 +21,11 @@ debug_info(elixir_v1, _Module, none, _Opts) ->
   {error, missing};
 debug_info(elixir_v1, _Module, {elixir_v1, Map, _Specs}, _Opts) ->
   {ok, Map};
-debug_info(erlang_v1, _Module, {elixir_v1, Map, Specs}, _Opts) ->
-  {Prefix, Forms, _, _} = dynamic_form(Map),
+debug_info(erlang_v1, Module, {elixir_v1, Map, Specs}, _Opts) ->
+  {Prefix, Forms, _, _} = dynamic_form(Map, deprecated_chunk_from_beam(Module)),
```
ggcampinho: In the worst case, it will read twice the `ExDp` chunk, but this should not happen unless the module is not loaded and the user is trying to read the `:abstract_code` and `ExDp` chunks

lib/elixir/test/elixir/kernel/deprecated_test.exs
ggcampinho: It's better for me, should I change here or in other PR?

lib/elixir/lib/module.ex
```diff
@@ -104,6 +104,17 @@ defmodule Module do
...
 
+  ### `@deprecated`
+
+  Provides the deprecation reason for a function. For example:
+
+      defmodule MyModule do
+        @deprecated "Use MyModule.hello_world/0 instead"
+        def hello, do: "world"
+
+        def hello_world, do: "Hello world!"
```
eksperimental: Could we use a real example like any function that has been deprecated in elixir?

ggcampinho: Should I already add the @deprecated to the functions deprecated on `elixir_disptach.erl`

josevalim: > Should I already add the @deprecated to the functions deprecated on elixir_disptach.erl?

No. :) Those are a bit trickier because this proposal will only warn for Mix projects. And we want the Elixir ones to always warn (in scripts, etc). We still need to find a way to solve it.

lib/elixir/test/elixir/kernel/deprecated_test.exs
josevalim: Let's do it in another PR please!

lib/elixir/test/elixir/protocol_test.exs
```diff
@@ -375,6 +381,10 @@ defmodule Protocol.ConsolidationTest do
     assert {{:ok, 1}, _, :def, [{:term, _, nil}], "Ok"} = List.keyfind(docs, {:ok, 1}, 0)
   end
 
+  test "consolidation keeps deprecated" do
+    assert [{{:ok, 1}, "Reason"}] = Sample.__info__(:deprecated)
```
josevalim: We need to check that it also keeps the deprecated chunk. My impression is that it does not, given the discussion on #7117.

ggcampinho: Sorry, I see my error now. I was using the `:code.get_object_code(Sample)` and it was returning the old binary I think. When I checked that inside of the tests for the docs the same thing happens in `Code.get_docs`, I thought it somehow the compiler was preserving the binary.

josevalim:

josevalim:
Thanks for all the work here @ggcampinho, especially with all the back and forth!

PR #7114, opened 2017-12-14
Fix Stream.chunk_every/4 odd results when step > count

### LINKED ISSUE #7096: Stream.chunk_every/4 gives odd results
I am *very* new to Elixir, and so looking at `chunk_every` when argument `step` is more than `count`.

    Erlang/OTP 20 [erts-9.0] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:10]
    
    Interactive Elixir (1.5.2) - press Ctrl+C to exit (type h() ENTER for help)
    iex(17)> Enum.chunk_every('ABCDEFGHIJK', 3, 4, [])
    ['ABC', 'EFG', 'IJK', 'IJK']
    iex(18)> Enum.chunk_every('ABCDEFGHIJKL', 3, 4, [])
    ['ABC', 'EFG', 'IJK']
    iex(19)> 

The output with 'IJK' repeated does not seem of much use?

Thanks.

(Copied from Reddit [here](https://www.reddit.com/r/elixir/comments/7i9ssa/messing_with_chunk_every_gives_odd_results/))

albertoalmagro:
# Summary
`Stream.chunk_every/4` was giving odd results when step was greater than
count and the last element of the enumerable was the same as last
chunk's last element.

This patch fixes this issue and adds tests for different scenarios when
`step > count`. It also adds a `doctest` example to reflect this in the
documentation and test it simultaneously.

Fixes #7096

### DIFF
--- a/lib/elixir/lib/stream/reducers.ex
+++ b/lib/elixir/lib/stream/reducers.ex
@@ -25,7 +25,7 @@ defmodule Stream.Reducers do
     end
 
     after_fun = fn {acc_buffer, acc_count} ->
-      if leftover == :discard or acc_count == 0 do
+      if leftover == :discard or acc_count == 0 or (step > count and count == acc_count) do
         {:cont, []}
       else
         {:cont, :lists.reverse(acc_buffer, Enum.take(leftover, count - acc_count)), []}
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -412,6 +412,9 @@ defmodule Enum do
       iex> Enum.chunk_every([1, 2, 3, 4], 10)
       [[1, 2, 3, 4]]
 
+      iex> Enum.chunk_every([1, 2, 3, 4, 5], 2, 3, [])
+      [[1, 2], [4, 5]]
+
   """
   @spec chunk_every(t, pos_integer, pos_integer, t | :discard) :: [list]
   def chunk_every(enumerable, count, step, leftover \\ [])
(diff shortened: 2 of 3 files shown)


lib/elixir/lib/stream/reducers.ex
```diff
@@ -25,7 +25,7 @@ defmodule Stream.Reducers do
     end
 
     after_fun = fn {acc_buffer, acc_count} ->
-      if leftover == :discard or acc_count == 0 do
+      if leftover == :discard or acc_count == 0 or (step > count and count == acc_count) do
```
ggcampinho: I think `count == acc_count` is a problem, you can test with:
```
assert Enum.chunk_every([1, 2, 3, 4, 5, 6, 7], 2, 4, []) == [[1, 2], [5, 6]]
```

albertoalmagro: Thanks @ggcampinho I will return to this after finishing my working hours 👍

josevalim:
I went ahead and merged it because we want to release v1.5.x and this will be backported. :)

josevalim:

albertoalmagro:
@josevalim I will try to make a fix this afternoon. Shall I push a new commit on this PR or create a new one?

josevalim:
@albertoalmagro i already pushed a fix to master. :)

albertoalmagro:
Haha @josevalim thanks 😊  I came to the same solution... I should have read this before 😆 

I also had this extra example on my branch at `enum_test.exs`

`assert Enum.chunk_every([1, 2, 3, 4, 5, 6, 7, 8], 2, 4, []) == [[1, 2], [5, 6]]`

May I commit it?

albertoalmagro:
@josevalim I realized there is still a bug here, current code isn't working for:

`assert Enum.chunk_every([1, 2, 3, 4, 5, 6, 7, 8], 2, 4, [9]) == [[1, 2], [5, 6], [9]]`

I will commit a new fix now.

PR #7117, opened 2017-12-16
Remove unnecessary processing of the ExDc chunk

ggcampinho:
The extra chuncks are already preserved during consolidation, no extra work is needed.
Notice this when I was working on #7113, and I think it will speed up protocol consolidation.

### DIFF
--- a/lib/elixir/lib/protocol.ex
+++ b/lib/elixir/lib/protocol.ex
@@ -309,7 +309,7 @@ defmodule Protocol do
   end
 
   defp beam_protocol(protocol) do
-    chunk_ids = [:abstract_code, :attributes, :compile_info, 'ExDc']
+    chunk_ids = [:abstract_code, :attributes, :compile_info]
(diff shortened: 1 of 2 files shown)


josevalim:
@ggcampinho sorry but I can't see how the chunk would be preserved. Could it be that in your tests the chunk was not overwritten yet? I merged this commit manually and then I tried fetching the docs for Enumerable in the plug project and I got this:

```
> h Enumerable.reduce

                              Enumerable.reduce/3

    @spec reduce(t(), acc(), reducer()) :: result()

Module was compiled without docs. Showing only specs.
```

ggcampinho:
Due to what I explained at https://github.com/elixir-lang/elixir/pull/7113#discussion_r157339099, maybe I would be nice to check the ExDc chunk directly on the tests and not `Code.get_docs`. If you agree I can do it.

josevalim:
Sounds good!

PR #7118, opened 2017-12-16
Stream.Reducers.chunk_every/4 Include leftover when acc_count = 0

albertoalmagro:
# Summary
In case acc_count = 0, if there is a leftover, it should be added to the result.

## Additional information
Continues #7114 (already merged)


josevalim:
I am not sure I agree. We didn't start a new chunk, so I don't think we should include the leftover. Same as this:

```iex
iex(1)> Enum.chunk_every([], 2, 2, [1])
[]
iex(2)> Enum.chunk_every([1, 2], 2, 2, [1])
[[1, 2]]
```

albertoalmagro:
You are right @josevalim I will close this one.

PR #7123, opened 2017-12-17
Minor docs updates

wojtekmach:


lib/elixir/lib/kernel.ex
```diff
@@ -3557,7 +3556,7 @@ defmodule Kernel do
   If the `Foo.Bar` module is moved somewhere else, the references to `Bar` in
   the `Foo` module need to be updated to the fully-qualified name (`Foo.Bar`) or
   an alias has to be explicitly set in the `Foo` module with the help of
-  `Kernel.SpecialForms.alias/2`.
```
josevalim: I am not a fan of those changes because when we see `alias/2` it usually means it is in the same module. I would prefer to be explicit in our docs. :)

lib/elixir/lib/kernel.ex
```diff
@@ -3819,7 +3818,7 @@ defmodule Kernel do
 
   ## rescue/catch/after
 
-  Function bodies support `rescue`, `catch` and `after` as `SpecialForms.try/1`
```
josevalim: Good catch!

lib/elixir/lib/kernel.ex
```diff
@@ -3507,7 +3506,7 @@ defmodule Kernel do
   be hygienized. This means the alias will be expanded when
   the macro is expanded.
 
-  Check `Kernel.SpecialForms.quote/2` for more information.
```
josevalim: I would prefer as it was as well for clarity!

lib/elixir/lib/kernel.ex
wojtekmach: Fair enough, reverted!

I just thought that based on https://github.com/elixir-lang/ex_doc/issues/679 this is the way to go. Or maybe for Kernel docs themselves we're really explicit, but everywhere else it's ok to link to Kernel/Kernel.SpecialForms? Or we just do it on case by case basis? ;-)

josevalim: In Elixir itself, I would prefer to always be explicit. Not sure about other projects thoughts we maintain - although I completely understand the expectation users may have that it should autolink.

josevalim:

lib/logger/lib/logger.ex
```diff
@@ -548,8 +548,8 @@ defmodule Logger do
   ## Options
 
     * `:flush` - when `true`, guarantees all messages currently sent
-      to both Logger and Erlang's `error_logger` are processed before
-      the backend is removed
+      to both Logger and OTP's [`:error_logger`](http://erlang.org/doc/man/error_logger.html)
```
wojtekmach: zrzut ekranu 2017-12-20 o 13 00 53

oops, is it expected that in IEx the url `http://erlang.org/doc/man/error_logger.html` is shown? If not, I can look into it.

josevalim: @wojtekmach it is expected, no worries.

PR #7127, opened 2017-12-20
Fix typo in String doc

gusaiani:
Hi all,

`as this options specifies` → `as this option specifies`.

Cheers!


whatyouhide:
Thanks @gusaiani! 💟

PR #7128, opened 2017-12-20
[WIP] Check deprecated calls on task xref warnings

ggcampinho:
Check deprecated using `__info__` or `ExDp` chunk if the module is not loaded,
it doesn't check local deprecations #7104

### DIFF
--- a/lib/mix/test/mix/tasks/xref_test.exs
+++ b/lib/mix/test/mix/tasks/xref_test.exs
@@ -145,9 +145,7 @@ defmodule Mix.Tasks.XrefTest do
 
   defp assert_all_calls(files, expected) do
     in_fixture "no_mixfile", fn ->
-      for {file, contents} <- files do
-        File.write!(file, contents)
-      end
+      generate_files(files)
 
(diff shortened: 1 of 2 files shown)


lib/mix/test/mix/tasks/xref_test.exs
```diff
@@ -253,59 +261,96 @@ defmodule Mix.Tasks.XrefTest do
...
-      else
-        List.old_flatten([1, 2, 3])
+  test "warnings: reports deprecated calls" do
+    files = %{
+      "lib/a.ex" => """
+      defmodule A do
+        @deprecated "Use A.c/0 instead"
+        def a, do: c()
+
+        def b, do: a()
```
ggcampinho: There is a problem here, that I can't warn for local files since they aren't in the manifest. Should I transverse the definitions or add a warning on `elixir_locals`?

josevalim: Let's document we won't warn for local calls... only the qualified ones.

josevalim:

PR #7129, opened 2017-12-20
Fix documentation for dependency format import

binaryseed:
The atom `:export_locals_without_parens` doesn't actually appear in source - the option is nested inside `:export`

### DIFF
--- a/lib/mix/lib/mix/tasks/format.ex
+++ b/lib/mix/lib/mix/tasks/format.ex
@@ -79,7 +79,7 @@ defmodule Mix.Tasks.Format do
   A dependency that wants to export formatter configuration needs to have a `.formatter.exs` file
   at the root of the project. In this file, the dependency can export a `:export` option with
   configuration to export. For now, only one option is supported under `:export`:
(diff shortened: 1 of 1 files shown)


binaryseed:
Somewhat related question: With this definition of `function: arity`, does that mean that we might over-include function calls?

Say a dependency like Plug includes `get: :*`, won't that mean that the rule will apply anywhere in source that we have a function named `get`? Even if it's not actually the Plug macro but rather a user-defined function?

Would it be better to define them with full MFA like `{Plug.Router, :get, :*}`

whatyouhide:
Thanks @binaryseed! 💟 

> Would it be better to define them with full MFA like {Plug.Router, :get, :*}

The formatter operates on the AST level so it doesn't know that `get("/foo")` is `Plug.Router.get("/foo")`. Yes, exporting `get: :*` from Plug would mean every `get/*` local function would be formatted without parens unless parens are manually added by the user. We discourage `:*` as an arity for this reason. Also, dependencies need to be explicitly imported so you will never stumble upon something like this without having control over disabling it. :)

PR #7133, opened 2017-12-22
Fix typo in Registry doc

yct21:


josevalim:
Both are correct. :) Although there is probably a better placement for the word "though" in that sentence.

yct21:
OK, I'm not a native English speaker and didn't know that was correct.

I'm closing it now :）

eksperimental:
IMO This PR was correct. I cannot think of a proper use of "though" in that sentence

josevalim:
The "though" is meant to be a contrast to the previous example, that uses start_link. It could also be written as:

> However, the registry is started as part of a supervision tree:

eksperimental:
Then your last wording is way more clear @josevalim

PR #7134, opened 2017-12-23
Introduce __STACKTRACE__

### LINKED ISSUE #7097: Introduce __STACKTRACE__
**TL;DR** Replace the use of `System.get_stacktrace()` with a pseudo-variable `__STACKTRACE__` similar to `__MODULE__` or `__CALLER__` available only inside `catch` and `rescue` clauses of the `try` expression.

## Reasons
Erlang recently set out to overhaul how stack traces are handled. The `:erlang.get_stacktrace/0` call (and thus it's Elixir's equivalent `System.get_stacktrace/0`) are problematic - they may cause the emulator to hold on to the stack trace for a very long time. This can be harmful, since the stack trace can be very big - in case of `FunctionClauseError` it includes all the arguments. The emulator has to hold on the stack trace for the last exception in general until the next exception is raised - which can be a long time.
To rectify this potential performance issue, OTP 20 started warning on the use of `:erlang.get_stacktrace()` outside of the `catch` clause of a `try` expression (in Elixir this would be outside of `catch` and `rescue`). The plan was to make the call return an empty list in some future Erlang version when called outside of `catch`.
OTP 21 will introduce an even more restrictive mechanism, to understand it we need to understand the syntax for `catch` in Erlang:
```erlang
try
  ...
catch
  Kind:Reason ->
    ...
end
```
Where `Kind` can be one of `:exit`, `:error` and `:throw`. The proposed (and merged) enhancement replaces this with:
```erlang
try
  ...
catch
  Kind:Reason:Stack ->
    ...
end
```
Where the `Stack` variable would contain the stack-trace for the just-rescued exception. It is forbidden to pattern match on this variable or use an already-bound variable (which would be equivalent to pinning it in Elixir). It's now planned that some future Erlang release will always return empty lists from `:erlang.get_stacktrace()` and the only way to acquire the stacktrace will be with this syntax. This means we need a compatible Elixir solution.

## Proposal
To solve this, we propose introduction of a new pseudo-variable `__STACKTRACE__` that would be available inside `catch` and `rescue` and would return the stack trace of the exception being currently handled.
This means that this code:
```elixir
try do
  ...
rescue
  e ->
    ...
    reraise(e, System.get_stacktrace())
end
```
Would become:
```elixir
try do
  ...
rescue
  e ->
    ...
    reraise(e, __STACKTRACE__)
end
```

The `__STACKTRACE__` pseudo-variable would be only available inside `catch` and `resuce` parts of `try` - it should fail to compile if present in any other place, similar to how `__CALLER__` is not available outside of macros. 

Having this as this kind of pseudo-variable is also compatible with how this works in Erlang - since neither pattern matching nor pinning of the stack variable is allowed, we don't loose any features.

## Implementation

Since OTP 21 is not stable yet, and we also want to support multiple Erlang versions, a conditional compilation of this feature is required. 

### On VMs without new syntax

When compiling on systems without support for the new syntax, the `__STACKTRACE__` pseudo-variable should compile into a call to `:erlang.get_stacktrace/0`.
Additionally, the call to `:erlang.get_stacktrace/0` should happen as the very first thing in the block, assigning the result to some internal variable, which could be later substituted for the `__STACKTRACE__` expression. This provides a very important advantage compared to the current `System.get_stacktrace/0` call by protecting against the "wrong stack trace" bug, which is very easy to introduce accidentally:
```elixir
try do
  ...
catch
  kind, reason ->
    cleanup()
    report_error(kind, reason, System.get_stacktrace())
end

defp cleanup() do
  {:ok, this_may_throw()}
catch
  throw -> {:error, throw}
end
```
If the `this_may_throw()` operation does indeed throw, the `System.get_stacktrace/0` call would return the value for that throw, instead of the exception we've just caught, since it always returns the stack trace of the last exception. With the use of the `__STACKTRACE__` pseudo-variable, this would compile into something conceptually like:
```elixir
try do
  ...
catch
  kind, reason ->
    stracktrace = :erlag.get_stacktrace()
    cleanup()
    report_error(kind, reason, stacktrace)
end
```
Which does not suffer from the issue. This means not only is the proposed feature giving us compatibility with Erlang, it is also a significant improvement in terms of preventing strange and hard to find bugs.

### On VMs with new syntax

When compiling on a VM that supports the new syntax, the `__STACKTRACE__` call should just use it, the previous example would compile into the following equivalent Erlang code:
```erlang
try do
  ...
catch
  Kind:Reason:Stacktrace ->
    cleanup()
    report_error(Kind, Reason, Stacktrace)
end
```

## Alternatives

An alternative would be to also support something like:
```elixir
try do
  ...
catch
  kind, reason, stack ->
    ...
end
```
There are couple of problems with that approach:
* similar to Erlang, it introduces this special variable that can't be pattern matched
* there's no clear way of using this syntax in combination with `rescue` or `catch` with just one pattern (which catches only throws).
* the compilation for VMs that don't support the new syntax is more awkward.

## Compatibility
Because of the conditional-compilation, the feature becomes compatible with all the Erlang releases. There is the constraint that code compiled on VM that supports the feature wouldn't work on older VMs, but this is also true today - code compiled on OTP 20 wouldn't work on OTP 18, even though Elixir in itself supports running on both VMs.

## Roadmap
The `System.get_stacktrace()` call should be soft-deprecated in the next release that includes the `__STACKTRACE__` feature. Additionally a loud deprecation should happen after 2 releases or as soon as Elixir declares compatibility for an Erlang VM that would always return an empty list for `:erlang.get_stacktrace()`. This might break the general deprecation policy approach of minimum 2 releases, but this is somewhat out of our hands.

michalmuskala:
This still needs probably more tests and additional docs, but I wanted to see if the approach for the implementation makes sense.

Closes #7097

### DIFF
--- a/lib/elixir/lib/kernel/special_forms.ex
+++ b/lib/elixir/lib/kernel/special_forms.ex
@@ -671,6 +671,14 @@ defmodule Kernel.SpecialForms do
   defmacro __CALLER__, do: error!([])
 
   @doc """
+  Returns the stacktrace for the curently handled exception.
+
+  It is available only in the `catch` and `rescue` clauses of `try/1`
+  expressions.
+  """
+  defmacro __STACKTRACE__, do: error!([])
+
+  @doc """
   Accesses an already bound variable in match clauses. Also known as the pin operator.
 
   ## Examples
--- a/lib/elixir/src/elixir_erl_pass.erl
+++ b/lib/elixir/src/elixir_erl_pass.erl
@@ -66,6 +66,9 @@ translate({'__block__', Meta, Args}, S) when is_list(Args) ->
 translate({'__CALLER__', Meta, Atom}, S) when is_atom(Atom) ->
   {{var, ?ann(Meta), '__CALLER__'}, S#elixir_erl{caller=true}};
 
+translate({'__STACKTRACE__', Meta, Atom}, S = #elixir_erl{stacktrace={Var, _}}) when is_atom(Atom) ->
+  {{var, ?ann(Meta), Var}, S#elixir_erl{stacktrace={Var, true}}};
+
 translate({'super', Meta, [{Kind, Name} | Args]}, S) ->
   %% In the expanded AST, super is used to invoke a function
   %% in the current module originated from a default clause
--- a/lib/elixir/lib/macro/env.ex
+++ b/lib/elixir/lib/macro/env.ex
@@ -71,6 +71,7 @@ defmodule Macro.Env do
   @typep unused_vars :: %{{var, var_version} => non_neg_integer | false}
   @typep current_vars :: %{var => {var_version, var_type}}
   @typep prematch_vars :: current_vars | :warn | :raise | :pin | :apply
+  @typep contextual_vars :: [atom]
 
   @type t :: %{
           __struct__: __MODULE__,
@@ -89,7 +90,8 @@ defmodule Macro.Env do
           unused_vars: unused_vars,
           current_vars: current_vars,
           prematch_vars: prematch_vars,
-          lexical_tracker: lexical_tracker
+          lexical_tracker: lexical_tracker,
+          contextual_vars: contextual_vars
         }
 
   # TODO: Remove :vars field
@@ -111,7 +113,8 @@ defmodule Macro.Env do
       unused_vars: %{},
       current_vars: %{},
       prematch_vars: :warn,
-      lexical_tracker: nil
+      lexical_tracker: nil,
+      contextual_vars: []
     }
   end
 
--- a/lib/elixir/src/elixir.hrl
+++ b/lib/elixir/src/elixir.hrl
@@ -11,7 +11,8 @@
   vars=#{},                %% a map of defined variables and their alias
   backup_vars=nil,         %% a copy of vars to be used on ^var
   extra_guards=[],         %% extra guards from args expansion
-  counter=#{}              %% a map counting the variables defined
+  counter=#{},             %% a map counting the variables defined
+  stacktrace=false         %% holds information about the stacktrace variable
 }).
 
 -record(elixir_quote, {
--- a/lib/elixir/src/elixir_clauses.erl
+++ b/lib/elixir/src/elixir_clauses.erl
@@ -213,11 +213,17 @@ expand_try(Meta, {'else', _} = Else, E) ->
   Fun = expand_one(Meta, 'try', 'else', fun head/2),
   expand_clauses(Meta, 'try', Fun, Else, E);
 expand_try(Meta, {'catch', _} = Catch, E) ->
-  expand_clauses(Meta, 'try', fun expand_catch/3, Catch, E);
+  expand_clauses_with_stacktrace(Meta, fun expand_catch/3, Catch, E);
 expand_try(Meta, {'rescue', _} = Rescue, E) ->
-  expand_clauses(Meta, 'try', fun expand_rescue/3, Rescue, E);
+  expand_clauses_with_stacktrace(Meta, fun expand_rescue/3, Rescue, E);
 expand_try(Meta, {Key, _}, E) ->
   form_error(Meta, ?key(E, file), ?MODULE, {unexpected_option, 'try', Key}).
+
+expand_clauses_with_stacktrace(Meta, Fun, Clauses, E) ->
+  OldContextualVars = ?key(E, contextual_vars),
+  ES = E#{contextual_vars := ['__STACKTRACE__' | OldContextualVars]},
+  {Ret, EE} = expand_clauses(Meta, 'try', Fun, Clauses, ES),
+  {Ret, EE#{contextual_vars := OldContextualVars}}.
 
 expand_catch(_Meta, [_] = Args, E) ->
   head(Args, E);
--- a/lib/elixir/src/elixir_env.erl
+++ b/lib/elixir/src/elixir_env.erl
@@ -24,7 +24,8 @@ new() ->
     unused_vars => #{},                    %% a map with unused variables
     current_vars => #{},                   %% a map with current variables
     prematch_vars => warn,                 %% behaviour outside and inside matches
-    lexical_tracker => nil}.               %% holds the lexical tracker PID}.
+    lexical_tracker => nil,                %% holds the lexical tracker PID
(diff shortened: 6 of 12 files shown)


lib/elixir/lib/kernel/special_forms.ex
```diff
@@ -670,6 +670,14 @@ defmodule Kernel.SpecialForms do
   """
   defmacro __CALLER__, do: error!([])
 
+  @doc """
+  Returns the stacktrace for the curently handled exception.
+
+  It is available only in the `catch` and `rescue` clauses of `try/1`
+  expression.
```
whatyouhide: expression**s**

lib/elixir/src/elixir_erl_pass.erl
```diff
@@ -629,3 +634,6 @@ generate_struct_name_guard([{map_field_exact, Ann, {atom, _, '__struct__'} = Key
   {lists:reverse(Acc, [{map_field_exact, Ann, Key, Match} | Rest]), S2};
 generate_struct_name_guard([Field | Rest], Acc, S) ->
   generate_struct_name_guard(Rest, [Field | Acc], S).
+
+format_error(invalid_stacktrace) ->
+  "__STACKTRACE__ is available only inside catch and rescue clauses of the try expression".
```
whatyouhide: "inside **the** catch and rescue clauses", or "inside catch and rescue clauses of the try expression**s**"?

lib/elixir/src/elixir_erl_pass.erl
```diff
@@ -66,6 +66,11 @@ translate({'__block__', Meta, Args}, S) when is_list(Args) ->
 translate({'__CALLER__', Meta, Atom}, S) when is_atom(Atom) ->
   {{var, ?ann(Meta), '__CALLER__'}, S#elixir_erl{caller=true}};
 
+translate({'__STACKTRACE__', Meta, Atom}, S = #elixir_erl{stacktrace={Var, _}}) when is_atom(Atom) ->
+  {{var, ?ann(Meta), Var}, S#elixir_erl{stacktrace={Var, true}}};
+translate({'__STACKTRACE__', Meta, Atom}, S) when is_atom(Atom) ->
+  elixir_errors:form_error(Meta, S#elixir_erl.file, ?MODULE, invalid_stacktrace);
```
josevalim: We cannot raise errors inside elixir_erl pass. All validation should happen during expansion. We even removed the `.file` field in master. :)

josevalim:
@michalmuskala sorry for the delay but I finally got to review this. :)

michalmuskala:
@josevalim any ideas how to do this during the expansion pass? Is it ok to add another field to the `Macro.Env` struct? Or maybe we could reuse some place to store this information?

josevalim:
@michalmuskala I would introduce a field called contextual_vars in Macro.Env that is a list of fields allowed in a given place. So once you enter catch/rescue, you would append `'__STACKTRACE__'`. When the variable is accessed, you check if the entry exists in the field.

This will also allow us to properly validate `__CALLER__`. Today it raises an awkward `{:unbound_var, var}` Erlang error.

michalmuskala:
@josevalim My current idea is to generate new variable name before entering try and save it in the environment struct as you proposed. Later when expanding the try, we can use the saved variable (and mark it as used) and then annotate the `try` metadata that we use the stacktrace, so code generation in the erlang pass can do the right thing (we could do it in Elixir right now, but we won't be able when we start using the new `catch` argument).  I'm not clear, though, how could I generate a new variable name. In the erlang pass I could use the `elixir_erl_var:build/2`, but I don't see anything similar to use during the elixir pass.

josevalim:
@michalmuskala I can think of two approaches.

In both cases, I don't think we should generate a new variable name in the Elixir pass.

### contextual_vars as a list

The `contextual_vars` simply says the given variable is allowed at that moment.

The generating of var would happen exclusively on the Erlang side, especially because we will ned to generate different types of code depending on the Erlang version.

### contextual_vars as a map

The `contextual_vars` is a map with the variable name as keys and if it was used or not as a boolean. If the key does not exist in the map, it means the variable is not available in that scope.

In this case, we could propagate the handling of the variable or not as a third argument in catch and a second argument in rescue. The con, in my opinion, is that we are changing the AST representation between passes and I would like to avoid those as much as possible.

Similar trade-offs happen in the metadata. Adding stuff to metadata makes it harder to write future passes. Remember we want the expanded AST to be translatable to JavaScript too.

### Wrapping up

I would go with keeping the `contextual_vars` as a list and simply validate the AST in the Elixir pass. The Erlang pass can be simpler because it will work on the validate AST, so it doesn't need to track context, only if stacktrace was used or not.

michalmuskala:
Updated. I took the second approach making `contextual_vars` a list. Indeed handling all the variable generation on Erlang side makes much more sense.

lib/elixir/src/elixir_erl_pass.erl
```diff
@@ -66,6 +66,9 @@ translate({'__block__', Meta, Args}, S) when is_list(Args) ->
 translate({'__CALLER__', Meta, Atom}, S) when is_atom(Atom) ->
   {{var, ?ann(Meta), '__CALLER__'}, S#elixir_erl{caller=true}};
 
+translate({'__STACKTRACE__', Meta, Atom}, S = #elixir_erl{stacktrace={Var, _}}) when is_atom(Atom) ->
+  {{var, ?ann(Meta), Var}, S#elixir_erl{stacktrace={Var, true}}};
```
josevalim: I think we can do like caller and simply use the `__STACKTRACE__` variable. This way the stacktrace field only keeps a boolean.

josevalim:
Do you want to add `__CALLER__` handling here too or should we do it separately?

lib/elixir/src/elixir_erl_pass.erl
michalmuskala: We can't since `try` (and stacktraces) could be nested, unlike `__CALLER__`.

michalmuskala:
I wanted to do `__CALLER__` separately

michalmuskala:
It seems there's some error on OTP 19, locally running `make test_elixir` just hangs...

josevalim:
The 19 error has been fixed on master. :)
-- 


josevalim:

PR #7153, opened 2017-12-27
Fix Application start and start_phase spec

angelikatyborska:
I think there was a mix up of application type (`:permanent | :transient | :temporary`) and application start type (`:normal | {:takeover, node} | {:failover, node}`) in the specs. The latter was missing from specs altogether. Callbacks `start` and `start_phase` were described as expecting as type one of `:permanent | :transient | :temporary`, wheres Erlang documentation says they expect one of `:normal | {:takeover, node} | {:failover, node}`. I fixed their specs to match Erlang documentation.

Related Erlang documentation:
http://erlang.org/doc/apps/kernel/application.html#Module:start-2
http://erlang.org/doc/apps/kernel/application.html#Module:start_phase-3
http://erlang.org/doc/apps/kernel/application.html#type-start_type


whatyouhide:
Great catch @angelikatyborska! I am not sure if this would cause a breaking change. I mean it would definitely but if we consider what we have now as a bug (which I think we should) then it would not be breaking, so 👍 from me. Wdyt about calling `app_type` `restart_type` instead, like Erlang does?

josevalim:
IMO We should go with start type and restart type.
-- 


whatyouhide:
> IMO We should go with start type and restart type.

Agreed. So let's go with `start_type` as the one given to `start/2` (`:normal | {:takeover, _} | ...`) and `restart_type` for what is called `app_type` in this PR. :)

angelikatyborska:
I've renamed `app_type` to `restart_type`.

josevalim:

PR #6347, opened 2017-07-17
Validate variable struct name

lexmag:

PR #7160, opened 2017-12-30
return right side expression value on struct matching

### LINKED ISSUE #7156: Strange result of pattern matching on structs
### Environment

* Elixir & Erlang versions (elixir --version): Elixir 1.6.0-rc.0 (182c730) (compiled with OTP 20)
* Operating system: macOS 10.13.2 - High Sierra

### Current behavior

```elixir
defmodule Foo do
  defstruct [bar: :baz]
end

iex(1)> %{} = %Foo{}
%Foo{bar: :baz}
iex(2)> %struct{} = %Foo{}
true
iex(3)> struct
Foo
```
Second line returned `true`. In previous elixir versions same code returns struct from right side of `=`
### Expected behavior
```elixir
iex(1)> %{} = %Foo{}
%Foo{bar: :baz}
iex(2)> %struct{} = %Foo{}
%Foo{bar: :baz}
iex(3)> struct
Foo
```

dmlittle:
There was a regression introduced in #6347 where `true` was returned as the value of an expression match instead of the right side of the expression. See #7156 for more info.

Fixes #7156

### DIFF
--- a/lib/elixir/src/elixir_erl_pass.erl
+++ b/lib/elixir/src/elixir_erl_pass.erl
@@ -25,7 +25,7 @@ translate({'=', Meta, [Left, Right]}, S) ->
       Reason = {tuple, Generated, [{atom, Generated, badmatch}, ResultVar]},
       RaiseExpr = elixir_erl:remote(Generated, erlang, error, [Reason]),
       GuardsExp = {'if', Generated, [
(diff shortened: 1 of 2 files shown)


dmlittle:
I'm not sure if this should be a PR against _master_ or _v1.6_ since it's in the roadmap for Elixir v1.6

PR #7166, opened 2018-01-01
Don't mention deprecated Map,Keyword.replace/3 in docs

wojtekmach:


lib/elixir/lib/keyword.ex
```diff
@@ -606,8 +606,8 @@ defmodule Keyword do
   end
 
   @doc """
-  Similar to `replace/3`, but will raise a `KeyError`
-  if the entry `key` does not exist.
+  Alters the value stored under `key` to `value`, but only
+  if the entry `key` already exists in `keywords`.
```
whatyouhide: I think we should mention that this raises a `KeyError` in the documentation and not just in the examples. Can you add it either to the summary or as a separate sentence below? Same for `Map` :)

wojtekmach: good call, fixed!

whatyouhide:
Thanks @wojtekmach! 💟

PR #7167, opened 2018-01-01
Add `@impl true` to StringIO & IEx.Pry

wojtekmach:
This way we'll no longer see `init/1` callback in docs:
https://hexdocs.pm/elixir/1.6.0-rc.0/StringIO.html#init/1

### DIFF
--- a/lib/elixir/lib/string_io.ex
+++ b/lib/elixir/lib/string_io.ex
@@ -99,11 +99,13 @@ defmodule StringIO do
 
   ## callbacks
 
+  @impl true
   def init({string, options}) do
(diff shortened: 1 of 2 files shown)


whatyouhide:
@wojtekmach I don't remember the behavior and #lazyweb, but does using `@impl true` makes the callback disappear from the docs or does it just not repeat the documentation?

wojtekmach:
@whatyouhide `@impl true` will essentially add `@doc false` so it will disappear from docs and won't autocomplete in IEx.

not sure why it wasn't present in 1.5.3 docs (https://hexdocs.pm/elixir/1.5.3/StringIO.html)

Issue #7171, opened 2018-01-03
Default value for System.get_env


hylkealons:
Hi there!

I'd like to supply a default value for `System.get_env` like in `Application.get_env/3`, so a default value can be specified if a system var is not specified.

Since this is a rather small change I already made the PR for this:
https://github.com/elixir-lang/elixir/pull/7172

josevalim:
However, because we always return strings, you can easily get the behaviour of a default value by doing `System.get_env(...) || "default"`. In any case, thanks for the PR!

hylkealons:
that will work and is my current solution, but that'd be also the case in any function with an optional value right?

I am now having this several times in my config file:
```
 (System.get_env("some_var") || "default_value") |> String.to_atom
```

Which I'd like to replace with:
```
System.get_env("some_var", "default_value) |> String.to_atom
```

Thanks for your time reviewing this :)

whatyouhide:
> but that'd be also the case in any function with an optional value right?

@hylkealons a function with an optional value can still return `nil` as its "proper" value and return something else as the default value, so this is not generally true.

zachdaniel:
It differs from something like `Map.get/3` because an env var cannot have a key with no value.

PR #7172, opened 2018-01-03
Default value for System.get_env

hylkealons:
## Issue
https://github.com/elixir-lang/elixir/issues/7171

## Problem
There is no default option for `System.get_env`

## Solution
Make it possible to supply a default value, like in `Application.get_env`


josevalim:
Because we always return strings, you can easily get the behaviour of a default value by doing `System.get_env(...) || "default"`.

PR #7181, opened 2018-01-05
Fix inspect for non-decimal negative integers

fhunleth:
Before:

iex> inspect(-1, base: :hex)
"0x-1"

After:

iex> inspect(-1, base: :hex)
"-0x1"

This change also applies to the octal and binary base options.

### DIFF
--- a/lib/elixir/lib/inspect.ex
+++ b/lib/elixir/lib/inspect.ex
@@ -262,6 +262,10 @@ defimpl Inspect, for: Integer do
 
   defp prepend_prefix(value, :decimal), do: value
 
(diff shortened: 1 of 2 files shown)


NobbZ:
Negative non-decimals with an unary minus to display signedness feels odd in general.

If bitwidth is known use 2-complement or if not either use decimal output or fail.

This is just an opinion though...

wojtekmach:
> Negative non-decimals with an unary minus to display signedness feels odd in general.

one way to look at it is that `-0x1` is valid Elixir syntax and `0x-1` is not.

josevalim:

fhunleth:
All of you are quick with reviews! Thanks!

Issue #7184, opened 2018-01-07
v1.6.0-rc.0 new documentation groups for Modules do not add any clarity


ianrumford:
I don't believe the new documentation groups for Modules add clarity
and can be rather confusing.

For example when I saw IO&System I immediately thought it would have
the date & time functions: Finding the time is *the* system feature I
use most in every language I've ever used. But of course DateTime gets
the time and formats it as well. So maybe DateTime should live in
IO&System and Data&Behaviours?

Also OptionParser lives in IO&System. Well yes the data comes from the
system but its surely a Data transformation package?

Also, one of the most important feature of any languages are its
collections. So why not group called Collections with Enum, Stream,
Enumerable and Collectable in it?

The point I'm highlighting, as I'm sure your see, is that any grouping
is essentially a personal view and may be confusing to other.

I note Erlang still, and just, uses an alphabetic list.  I would
commend alphabetic as the best, future-proof solution for Elixir.

josevalim:
We completely understand this change may be an improvement to some and a regression for others. We will only know the net outcome once enough users try it out and give their opinion, so we really appreciate you are doing so.

I also agree that groups are a personal view and there are multiple interpretations. Our hope, however, is that the grouping will help newcomers, who need to make sense of what the Elixir standard library provides and an alphabetical ordering is of absolutely no help here. From this perspective, an eagle view is more important than being accurate according to everyone's interpretation (which is most likely be impossible).

Therefore a discussion on how to improve the groups is definitely welcome. For example, I like the suggestion of having a separate group for collections. You would include Keyword, Map, and MapSet and in there as well?

On the other hand, if the ordering makes it harder to navigate, we need to consider improvements. For example, we could provide search by modules (similar to pressing t on GitHub to find files) or have an option that removes the group and keeps the alphabetical ordering.

If you would like to gather more feedback on the usefulness of groups as a whole,  then it is probably best to start a wider discussion on the topic, maybe by starting a thread in the mailing list and/or Elixir Forum.

josevalim:
To move the discussion forward I have opened #7185.

josevalim:
Closing in favor of the PR. Feel free to open more discussions about specific changes on the grouping functionality. The plan for now is to keep it around! Thanks!

Issue #6571, opened 2017-09-14
v1.5.1 Kernel & Kernel.SpecialForms documentation issue


ianrumford:
### Environment

Erlang/OTP 20 [erts-9.0] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:10] [hipe] [kernel-poll:false]

Elixir 1.5.1

Ubuntu 16.04

### Current behavior

The v1.5.1 documentation does not disambiguate functions and macros; everything is listed as a function.

So you might be surpised, as I was :-), if you do this

`Kernel.SpecialForms.__info__(:functions)`

and get an empty list.

### Expected behavior

Unless my memory is playing up, I feel sure Kernel docs once did separate into functions and macros?

Pragmatically it makes no difference but the docs should be correct IMHO.

josevalim:
The division between functions and macros in the docs caused difficulty when learning new APIs. You can see that on the right side, close to the source link, we still say they are a "macro". Please open up an issue in ExDoc if you would like to discuss how those should be exhibited. Thank you!

PR #7185, opened 2018-01-07
Introduce a group for collections

josevalim:
Currently the docs have a large group called "Data & Behaviour". The goal of this PR is to extract "Collections" (lists, sets, etc) and "Enumerables" (such as ranges) into their own group. Feedback is welcome!

### DIFF
--- a/lib/elixir/docs.exs
+++ b/lib/elixir/docs.exs
@@ -4,34 +4,37 @@
   groups_for_modules: [
     # [Kernel, Kernel.SpecialForms],
 
-    "Data & Behaviours": [
-      Access,
+    "Basic Types": [
       Atom,
       Base,
       Bitwise,
       Calendar,
       Calendar.ISO,
       Date,
-      Date.Range,
       DateTime,
-      Enum,
       Exception,
       Float,
       Integer,
-      Keyword,
-      List,
-      Map,
-      MapSet,
       NaiveDateTime,
-      Range,
       Record,
       Regex,
-      Stream,
       String,
       Time,
       Tuple,
       URI,
       Version,
+    ],
+
+    "Collections & Enumerables": [
+      Access,
+      Date.Range,
+      Enum,
+      Keyword,
+      List,
+      Map,
+      MapSet,
+      Range,
+      Stream,
     ],
 
     "IO & System": [


lib/elixir/docs.exs
```diff
@@ -4,36 +4,39 @@
...
-      Map,
-      MapSet,
       NaiveDateTime,
-      Range,
       Record,
       Regex,
-      Stream,
       String,
       Time,
       Tuple,
```
michalmuskala: Should `Tuple` also live in collections?

michalmuskala:
I think this is a move in the right direction, I'm not particularly convinced by the name "Collections & Enumerables" - I think "Collections" or "Collection Types" would be enough. You could easily view range as an optimised collection of sequential integers.

lib/elixir/docs.exs
josevalim: I would say Tuple is a composite type but not a collection.

josevalim:
> You could easily view range as an optimised collection of sequential integers.

I like Collections if we can agree with the above.

ianrumford:
The above discussion just highlights my point :-)  There is no "solution" and all groupings are essentially arbitrary.

A while ago I opened [#6571](https://github.com/elixir-lang/elixir/pull/6571) regarding the removal of the distinction between "functions" from "macros" in Kernel.SpecialForms.  Jose made the point that newcomers found that confusing.  I accepted that.

But adding these groups are even more confusing for newcomers in my view  and I can't reconcile keeping it simple for newcomers v adding these groups which really only make sense to people who already know their way around the the language.

Which goes to the heart of who is the target audience for the reference documentation? If its newcomers (and I think that's best) then the simpler the better.

josevalim:
> But adding these groups are even more confusing for newcomers in my view

As I said in #7184, we need to validate this. Otherwise, we will only be guessing which one is better.

Awenimpagaritse:
Cool

lib/elixir/docs.exs
Awenimpagaritse: Cool

Awenimpagaritse: Cool

josevalim:

lib/elixir/docs.exs
fertapric: The first paragraph of the `Tuple` documentation says: [`Tuples are ordered collections of elements`](https://github.com/elixir-lang/elixir/blob/master/lib/elixir/lib/tuple.ex#L5). Should we update the documentation expressing is a composite type?

The [Erlang documentation](http://erlang.org/doc/reference_manual/data_types.html) says:

> A tuple is a compound data type with a fixed number of terms:
>
> `{Term1,...,TermN}`
>
> Each term `Term` in the tuple is called an *element*. The number of elements is said to be the size of the tuple.

josevalim:
Good call, we should update it.
-- 

PR #7189, opened 2018-01-08
Change `assert_in_delta` better handle a delta of 0

devonestes:
Previously if the two values that we were checking were within a given
delta were equal, they would fail if the expected delta was `0`.


josevalim:

PR #7187, opened 2018-01-07
Add symmetric closure

crabonature:
In `do_digits()` there are two symmetric closures in recursion, so probably for `do_undigits()` there can be two too.

PR #7190, opened 2018-01-09
Simplify recursion

crabonature:
Relates in some way to: https://github.com/elixir-lang/elixir/pull/7187


josevalim:
I pushed a commit that formats the code, otherwise the build will fail. You can do it locally by running `mix format`. :) Thanks!

crabonature:
Nah man :) I forget about formatter, there is longer line length that 80 what I got in my Emacs. Thanks!

josevalim:

PR #7191, opened 2018-01-09
Add additional documentation for `assert_in_delta` and `refute_in_delta`

devonestes:
In a previous commit I changed the behavior of `assert_in_delta`, but I
forgot to document that change so it's clear to users in the future. I
also made it clear that `refute_in_delta` is exclusive, so folks
shouldn't have questions about that anymore.


josevalim:

PR #7193, opened 2018-01-09
Add discard_threshold to Logger

### LINKED ISSUE #7192: Improved Logger behaviour
As discussed on the Elixir Forum and also on a blog post I wrote, the current Logger behaviour of switching to sync mode in order to try and put some back-pressure to the log producers can break some systems entirely.

When a system comes under unexpected load and the logging code might generate a huge amount of logs, a more safe behaviour is to drop log messages entirely (perhaps logging the fact!) — it seems that [Lager](https://github.com/erlang-lager/lager) is doing that for Erlang.

I am opening this issue at the request of @josevalim. I'd be happy to help but I'd require some guidance.

josevalim:
See https://github.com/josevalim/cascading_failures.

Closes #7192.


dbohdan:
Very cool to see this feature targeting the v1.6.0 milestone! 👍

lib/logger/lib/logger.ex
```diff
@@ -130,6 +130,13 @@ defmodule Logger do
       in the queue is reduced to `sync_threshold * 0.75` messages.
       Defaults to 20 messages.
 
+    * `:discard_threshold` - if the `Logger` manager has more than
+      `:discard_threshold` messages in its queue, `Logger` will change
+      to *discard mode* and messages will be discarded directly in the
+      clients. `Logger` will return to *sync mode* once the number of
+      messages in the queue is reduced to `discard_threshold * 0.75`
+      messages. Defaults to 10000 messages.
```
fishcakez: 10000 is a very big number, there could be a lot of reference counted binaries in there or other large data structures using up memory. The GC time on these processes will be significant as we have on heap message queues - we could switch to off heap for Logger.

Also the time to process 7500 messages is still likely to be a long time to be stuck in sync mode (which has infinity timeout!). If the sync mode is 20 and the discard is 10000 chances are your app is already broken and remains broken regardless because the sync mode did no provide back pressure to the application, so likely the logs are coming in from non-backpressure requests (e.g. the node is a server and logging based on external requests) and those are going to be very slow. I think we should default to a number closer to 500 like before.

Another issue with a big number is we can end up with "buffer bloat" and are more likely to have empty periods where there are no logs.

lib/logger/lib/logger/utils.ex
```diff
@@ -1,6 +1,37 @@
...
+        mode,
+        messages,
+        async_threshold,
+        sync_threshold,
+        keep_threshold,
+        discard_threshold
+      ) do
+    case mode do
+      _ when messages >= discard_threshold -> :discard
+      :discard when messages <= keep_threshold -> :sync
```
fishcakez: Could there be situation where messages >= keep_threshold and messages <= async threshold? This configuration could occur because someone never wants to sync.

lib/logger/lib/logger/config.ex
```diff
@@ -105,6 +105,16 @@ defmodule Logger.Config do
         {:ok, state}
 
       new_mode ->
+        if new_mode == :discard do
```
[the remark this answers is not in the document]
josevalim: The logs will be back by then, so we should be fine.

lib/logger/lib/logger/utils.ex
josevalim: Oh, I see, I will add some tests for that.

lib/logger/lib/logger.ex
josevalim: What do you suggest? 1000? 2000? 5000?

fishcakez: Staying with 500 is ok I think as we had that before.

lib/logger/lib/logger/config.ex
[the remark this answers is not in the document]
fishcakez: We likely want to inform someone reading the logs that we stopped discarding. The order of the discard log messages will be confusing as it may appear logging resumed when it didn't because logs can arrive after the discard log message as they read `:async` or `:sync` just before it changed, and the message is sent after the discard log message. The only way to determine the end of a discarding period is to log it I think.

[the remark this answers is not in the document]
josevalim: Done.

PR #7197, opened 2018-01-10
Add type union explanation to typespecs

mdelaossa:
A [question on StackOverflow](https://stackoverflow.com/questions/48196624/what-is-the-meaning-of-an-elixir-typespec-with-types-in-parenthesis-separated-by/48196807) brought to light that type unions are not called out anywhere in the docs, so here's a small edit to hopefully shed a little more light on them for newcomers :)


josevalim:

PR #7201, opened 2018-01-12
Replace Code.load_file/2 by Code.compile_file/2

josevalim:
Currently there is no way to compile a file without leaving
footprints on the system. Plus, the term load_file/2 is very
confusing as loaded is the notation used by modules.

This commit focuses the whole loaded aspect of files exclusively
on require_file/2 (and renames two functions accordingly) and
introduces compile_file/2 which is semantically simpler than
load_file/2.

### DIFF
--- a/lib/elixir/lib/code.ex
+++ b/lib/elixir/lib/code.ex
@@ -5,25 +5,55 @@ defmodule Code do
   This module complements Erlang's [`:code` module](http://www.erlang.org/doc/man/code.html)
   to add behaviour which is specific to Elixir. Almost all of the functions in this module
   have global side effects on the behaviour of Elixir.
+
+  ## Working with files
+
+  This module contains three functions for compiling and evaluating files.
+  Here is a summary of them and their behaviour:
+
+    * `require_file/2` - compiles a file and tracks its name. It does not
+      compile the file again if has been previously required (or loaded).
+
+    * `compile_file/2` - compiles a file without tracking its name. Compiles the
+      file multiple times when invoked multiple times.
+
+    * `eval_file/2` - evaluates the file contents without tracking its name. It
+      returns the result of the last expression in the file, instead of the modules
+      defined in it.
+
+  In a nutshell, the first must be used when you want to keep track of the files
+  handled by the system, to avoid the same file from being compiled multiple
+  times. This is common in scripts.
+
+  `compile_file/2` must be used when you are only interested in the compilation
+  results, without tracking. `eval_file/2` should be used when you are intested on
+  the result of evaluating the file rather than the modules it defines.
   """
 
   @doc """
-  Lists all loaded files.
+  Lists all required files.
 
   ## Examples
 
       Code.require_file("../eex/test/eex_test.exs")
-      List.first(Code.loaded_files()) =~ "eex_test.exs"
+      List.first(Code.required_files()) =~ "eex_test.exs"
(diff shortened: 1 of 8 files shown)


lib/elixir/lib/code.ex
```diff
@@ -32,16 +62,23 @@ defmodule Code do
...
+  @since "1.7.0"
+  def unrequire_files(files) do
+    :elixir_code_server.cast({:unrequire_files, files})
+  end
+
+  # TODO: Deprecate me on 1.9
+  @doc false
   def unload_files(files) do
-    :elixir_code_server.cast({:unload_files, files})
+    :elixir_code_server.cast({:unrequire_files, files})
```
fishcakez: As we no longer test this function can we call `unrequire_files/1` instead here, as a change to unrequire could break this unknowingly.

lib/elixir/lib/code.ex
```diff
@@ -5,25 +5,55 @@ defmodule Code do
...
+  @since "1.7.0"
+  @spec required_files() :: [binary]
+  def required_files do
+    :elixir_code_server.call(:loaded)
+  end
+
+  # TODO: Deprecate me on 1.9
+  @doc false
   def loaded_files do
     :elixir_code_server.call(:loaded)
```
fishcakez: Call `required_files/0`.

lib/elixir/lib/code.ex
```diff
@@ -5,25 +5,55 @@ defmodule Code do
...
-      List.first(Code.loaded_files()) =~ "eex_test.exs"
+      List.first(Code.required_files()) =~ "eex_test.exs"
       #=> true
 
   """
-  @spec loaded_files() :: [binary]
+  @since "1.7.0"
+  @spec required_files() :: [binary]
+  def required_files do
+    :elixir_code_server.call(:loaded)
```
fishcakez: Can we refactor to use `:required` to match naming?

lib/elixir/lib/code.ex
```diff
@@ -5,25 +5,55 @@ defmodule Code do
...
   to add behaviour which is specific to Elixir. Almost all of the functions in this module
   have global side effects on the behaviour of Elixir.
+
+  ## Working with files
+
+  This module contains three functions for compiling and evaluating files.
+  Here is a summary of them and their behaviour:
+
+    * `require_file/2` - compiles a file and tracks its name. It does not
+      compile the file again if has been previously required (or loaded).
```
eksperimental: if "it" has previuosly been required

lib/elixir/lib/code.ex
```diff
@@ -5,25 +5,55 @@ defmodule Code do
...
+    * `eval_file/2` - evaluates the file contents without tracking its name. It
+      returns the result of the last expression in the file, instead of the modules
+      defined in it.
+
+  In a nutshell, the first must be used when you want to keep track of the files
+  handled by the system, to avoid the same file from being compiled multiple
+  times. This is common in scripts.
+
+  `compile_file/2` must be used when you are only interested in the compilation
+  results, without tracking. `eval_file/2` should be used when you are intested on
```
eksperimental: Interested in the results

lib/elixir/lib/code.ex
eksperimental: I meant "previously"

josevalim:

PR #4295, opened 2016-02-12
ExUnit: Show skipped tests as "S" and invalid in red in test output

wojtekmach:
Lately I noticed that I kept forgetting to re-enable tests that I previously marked as skipped. The total count and skipped count shows up at the bottom but I keep missing it and I wouldn't be surprised if other developers have the same problem. Here I propose a simple change that would make it obvious that a test was skipped. The "S" is stolen from minitest; a yellow "*" from RSpec or anything else, really, would work too. I also changed invalid to red.

Since it's a pretty trivial change I wasn't sure if it was worth discussing in elixir-lang-core, so I just went ahead and tried it out, but of course I can open up a thread.

Example run:
example

Example run with trace:
with trace

When some tests are skipped:
skipped

PR #7203, opened 2018-01-13
Highlight skipped tests in yellow

devonestes:
This adds yellow highlighting for skipped tests in the ExUnit CLI output. This way, when folks have skipped tests that were only intended to be temporary, they will have less of a chance of accidentally thinking they're "all green" and pushing broken code (that would also pass automated CI/CD tests and break production, which I _may_ have done last week 😜 ).

Here's what the new output will look like with skipped tests.
ex_unit_colors

### DIFF
--- a/lib/ex_unit/lib/ex_unit/cli_formatter.ex
+++ b/lib/ex_unit/lib/ex_unit/cli_formatter.ex
@@ -55,11 +55,7 @@ defmodule ExUnit.CLIFormatter do
   end
 
   def handle_cast({:test_finished, %ExUnit.Test{state: {:skip, _}} = test}, config) do
-    if config.trace do
-      IO.puts(trace_test_skip(test))
-    else
-      IO.write(skipped("*", config))
-    end
+    if config.trace, do: IO.puts(trace_test_skip(test))
 
     test_counter = update_test_counter(config.test_counter, test)
     config = %{config | test_counter: test_counter, skipped_counter: config.skipped_counter + 1}
@@ -242,7 +238,10 @@ defmodule ExUnit.CLIFormatter do
 
     message =
       "#{test_type_counts}#{config.failure_counter} #{failure_pl}"
-      |> if_true(config.skipped_counter > 0, &(&1 <> ", #{config.skipped_counter} skipped"))
+      |> if_true(
+        config.skipped_counter > 0,
+        &(&1 <> "," <> skipped(" #{config.skipped_counter} skipped", config))
+      )
       |> if_true(config.invalid_counter > 0, &(&1 <> ", #{config.invalid_counter} invalid"))
 
     cond do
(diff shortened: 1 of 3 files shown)


wojtekmach:
@devonestes I proposed similar thing a while back, here's a reason it was rejected: https://github.com/elixir-lang/elixir/pull/4295#issuecomment-183417147

devonestes:
@wojtekmach Funny, I asked on the mailing list about this and [got the ok](https://groups.google.com/d/msg/elixir-lang-core/HCAM4QFN9a8/WCv1Z6nTAAAJ) from @josevalim, which is why I went ahead and implemented this. Sorry, didn't see your earlier PR! Looks like the issues that José mentioned there might still be valid (but maybe they're not?), so I'll wait to see what he says and if needed I'll find a way to deal with those issues. I actually think this is fairly important visual feedback since the problems raised by pushing accidentally broken code to CI can be significant.

josevalim:
Maybe I misunderstood the proposal but I agree with printing the "3 skipped" in yellow at the end. I don't think we should print skipped tests as `*` because skipping tests is very very very common.

What do you think?

wojtekmach:
just "3 skipped" in yellow at the bottom seems like a good middle ground to me 👍

devonestes:
Yeah, that's a solid compromise. At least something yellow in there to hopefully keep me from being silly.

devonestes:
Ok, updated. Here's the new output:

ex_unit_colors

josevalim:
Ok, let's wait for a bit more feedback and then we can merge it!

josevalim:
Folks, I have tried this PR for a while and I am not really happy with the result. Maybe we should track between tests skipped via command line and via the test suite? For suites where you expect tests to be excluded, printing the result in yellow is calling the attention to something that nothing can be done about.

@devonestes I assume that the issue that initially triggered this was a `@tag :skip`? Maybe we can say 5 excluded for the ones excluded via the command line and 3 skipped for the tag ones?

josevalim:
If we make a distinction between excluded and skipped, we can even print `*` during the test run, as originally proposed. For now, I will revert this PR. :)

@devonestes please let me know if you would like to try this new direction and if you have any questions! :heart:

devonestes:
You'll never believe this (or maybe you will), but I was going to propose something like this on the mailing list tomorrow!

I was running tests today by a given line number like this `mix test test/build_vs_buy/calculator_test.exs:66`, and it kind of bothered me that it still said `4 tests, 0 failures, 3 skipped`. I was thinking, "shouldn't that be `1 test, 0 failures`"? Is that along the lines of what you're thinking?

I think the difference between excluded and skipped is the big point there. That's a great way to put it. I'd be totally happy to give this new direction a peek and see what I can do.

josevalim:
Awesome, so please do try a new PR and let us know if you have any questions! Doing a search for `:skip` in ExUnit should reveal all of the places we need to change. Now would also be a good opportunity to change rename `:skip` to `:skipped`.

PR #7204, opened 2018-01-13
Don’t add parens for defrecord/3 and defrecordp/3

ejpcmac:
`defrecord/2` and `defrecordp/2` don’t get parentheses when running the code formatter, but their `/3` versions do, so I’ve added `defrecord/3` and `defrecordp/3` to the list of locals without parens.


josevalim:

PR #7205, opened 2018-01-13
Don’t add parens to on_exit/1

ejpcmac:
The `on_exit/1` function currently get parentheses when formatting the code. This pull request aims to add it to the locals without parens list.

As this is a function and not a macro, I’m not sure wether it’s intended, so I propose it here in a different pull request.


josevalim:
Thanks! I am OK with having parens here since it is not a module function nor part of the assertions. But let's see what others think. :)

whatyouhide:
It's not part of the DSL as I see it, I see it as a regular function, so I am very :+1: on keeping parens here.

josevalim:
Alright then. :)

PR #7206, opened 2018-01-14
Add clarity to IEx docs for OS X

9mm:


josevalim:
Are you sure this is necessary? The history does not use the `~/.iex_history` file, it doesn't even exist on my machine, and history works fine. :)

josevalim:
Actually, I think we used `~/.iex_history` for a brief period of time while v1.6 was still being developed but that's no longer the case. Please let us know. In any case, thanks for the PR!

9mm:
Hmmm. I’ll verify to make sure. I found it from a SO thread because that single line above still wasn’t working until I touched that history file. I’ll double check 

Sent from my iToaster

9mm:
Ok you're right I am crazy I guess. I found the SO thread and followed the directions and it didnt work, so I read further about `.iex_history`, touched that file, then it worked. Maybe it was my imagination.

PR #7207, opened 2018-01-14
Re-format codebase

devonestes:
I noticed yesterday while working on a different PR that when I ran
`mix format` on the Elixir codebase from the root directly, there were
changes that I didn't expect to see. I'm not sure how these snuck by
(maybe they came as a result of some recent updates to the formatter?),
but on my machine they are consistently applied by the formatter after
updating from the current master branch.

I was especially surprised to see this since the formating is checked
in CI, right? Wouldn't these changes being outstanding cause CI to
fail? Maybe there's some bug hidden in there that's giving false
positives in CI?

### DIFF
--- a/lib/elixir/lib/calendar/iso.ex
+++ b/lib/elixir/lib/calendar/iso.ex
@@ -74,15 +74,16 @@ defmodule Calendar.ISO do
       {2000, 1, 1, 12, 0, 0, {0, 6}}
 
   """
-  @spec naive_datetime_from_iso_days(Calendar.iso_days()) :: {
-          Calendar.year(),
-          Calendar.month(),
-          Calendar.day(),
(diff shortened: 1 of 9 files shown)


wojtekmach:
Could you run: make clean compile and try firmatting again on latest master?

josevalim:
Yes, our build fails if files are not formatted so as a general rule, if
Travis is green, all files are formatted. :)

So you are likely running “mix format” with an elixir version that is not
latest master. Either because you didn’t recompile or it is picking up
another elixir install on your machine.
-- 


devonestes:
Oh, yeah, my bad. I forgot I needed to run `bin/elixir bin/mix format`. I was running `mix format`, which was using my installed version of 1.6.0-rc.0.

Issue #821, opened 2013-02-03
Warn on unused imports


orenbenkiki:
This is a (very) nice to have - warn if importing a module, but not actually using anything from it (macros or functions). Haskell does this and it is great to know that if I see an import, it is actually used and not a leftover from some evolution of the code.

PR #7210, opened 2018-01-15
Add documentation for string keys in Keyword's colon syntax

unnawut:
Following ex_doc's [#821](https://github.com/elixir-lang/ex_doc/pull/821) discussion...

This is the example provided in [`Keyword`](https://hexdocs.pm/elixir/Keyword.html) docs:

```ex
iex> [{:exit_on_close, true}]
[exit_on_close: true]
```

And because below, `elem(0)` is a string, it is not a `Keyword`. So it remains a list of tuples:

```ex
iex> [{"exit_on_close", true}]
[{"exit_on_close", true}]
```

But the code below, which is currently not documented, is actually a valid keyword syntax, and the string key automatically gets converted to an atom:

```ex
iex> ["key": "value"]
[key: "value"]
```

So I thought it would be nice to document this in the docs. :heart:

### DIFF
--- a/lib/elixir/lib/keyword.ex
+++ b/lib/elixir/lib/keyword.ex
@@ -20,12 +20,13 @@ defmodule Keyword do
       iex> [{:active, :once}]
       [active: :once]
 
-  The two syntaxes are completely equivalent. 
-  
-  Using the colon syntax will automatically convert string keys to atoms:
+  The two syntaxes are completely equivalent. When using the colon syntax,
+  the key is always an atom, even when wrapped in quotes:
   
       iex> ["exit_on_close": true]
       [exit_on_close: true]
+  
+  This mirrors the quoted atom syntax such as `:"exit_on_close"`.
   
   Note that when keyword lists are passed as the last argument to a function,
   if the short-hand syntax is used then the square brackets around the keyword list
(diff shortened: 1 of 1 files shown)


josevalim:

mememori:
Just a note, on the following code

```elixir
["key": "value"]
```
`"key":` is not a string but an atom. For example

```elixir
:"key" = :key
```

because `:"key"` is equivalent to erlang's `'key'`; the quotes in atoms allow you to escape more complex atom names, eg: `:"atom with spaces"` etc, so, `["foo": :bar]` means `[foo: :bar]` because both are literally the same thing (there is no conversion)

whatyouhide:
@mememori yep, that's what the PR fixed right?

unnawut:
@mememori I believe what you are pointing out is mainly to do with my understanding/writing (sorry!) in the PR. Jose made an update to the PR before the merge, which probably already addressed your point. :)

> The two syntaxes are completely equivalent. When using the colon syntax,
>   the key is always an atom, even when wrapped in quotes:
>   
>       iex> ["exit_on_close": true]
>       [exit_on_close: true]
>   
>   This mirrors the quoted atom syntax such as `:"exit_on_close"`.

mememori:
@unnawut yup, i was just explaining that detail, it has nothing to do with the PR.

Thanks for the contrib :grin:

PR #7213, opened 2018-01-16
Adding tee/2 function to Kernel

kelvinst:
This is a basic implementation of what @josevalim suggested on https://groups.google.com/d/msgid/elixir-lang-core/CAGnRm4Ji7kAjv5i4_04AVp-hae4ghj1DgZx1zr%2B4LyGBYpj5_A%40mail.gmail.com?utm_medium=email&utm_source=footer

PS: Didn't dog food it yet because I don't if this is going to be the final place for it (or even if it would be accepted). Once decided, we can think some places to use it on elixir core.

### DIFF
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -3151,6 +3151,24 @@ defmodule Kernel do
   end
 
   @doc """
+  This function applies the given function to the subject, but
+  returns the unchanged subject.
+
+  It can be useful in situations like the following:
+
+      ["thing1", "thing2"]
+      |> generate_more_things()
+      |> tee(&IO.inspect(length(&1)))
+      |> tee(&File.write!("things", &1))
+      |> do_something_with_things()
+
+  """
+  def tee(subject, function) when is_function(function) do
+    function.(subject)
+    subject
+  end
+
+  @doc """
   Returns `true` if `module` is loaded and contains a
   public `function` with the given `arity`, otherwise `false`.
 
--- a/lib/elixir/lib/keyword.ex
+++ b/lib/elixir/lib/keyword.ex
@@ -22,12 +22,12 @@ defmodule Keyword do
 
   The two syntaxes are completely equivalent. When using the colon syntax,
   the key is always an atom, even when wrapped in quotes:
-  
+
       iex> ["exit_on_close": true]
       [exit_on_close: true]
-  
+
   This mirrors the quoted atom syntax such as `:"exit_on_close"`.
-  
+
   Note that when keyword lists are passed as the last argument to a function,
   if the short-hand syntax is used then the square brackets around the keyword list
   can be omitted as well. For example, the following:
--- a/lib/elixir/test/elixir/kernel_test.exs
+++ b/lib/elixir/test/elixir/kernel_test.exs
@@ -941,8 +941,8 @@ defmodule KernelTest do
 
   test "tee/2" do
     assert capture_io(fn ->
-      assert tee("This is a tee!", &IO.inspect(String.length(&1))) == "This is a tee!"
-    end) == "14\n"
+             assert tee("This is a tee!", &IO.inspect(String.length(&1))) == "This is a tee!"
+           end) == "14\n"
   end
 
   defp purge(module) do
--- a/lib/elixir/test/elixir/kernel_test.exs
+++ b/lib/elixir/test/elixir/kernel_test.exs
@@ -3,6 +3,8 @@ Code.require_file("test_helper.exs", __DIR__)
 defmodule KernelTest do
   use ExUnit.Case, async: true
 
+  import ExUnit.CaptureIO
+
   doctest Kernel
 
   defp empty_list(), do: []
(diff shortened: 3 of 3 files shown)


devonestes:
This looks like a really useful function!

However, am I alone in having no idea why it has the name “tee”? Is there some computer science thing I’m missing that would make this name relevant?

fxn:
What about removing “unchanged”? It returns “the subject” (or the first argument).

fertapric:
> However, am I alone in having no idea why it has the name “tee”? Is there some computer science thing I’m missing that would make this name relevant?

@devonestes I think it's named after the ["T" piece used in plumbing](https://en.wikipedia.org/wiki/Piping_and_plumbing_fitting#Tee).

josevalim:
The name comes from here: https://en.wikipedia.org/wiki/Tee_(command)

Since it is useful mostly inside pipelines, it probably makes sense to borrow the name from Unix pipes too.

josevalim:
> PS: Didn't dog food it yet because I don't if this is going to be the final place for it (or even if it would be accepted). Once decided, we can think some places to use it on elixir core.

The truth is I am not sure we would find places to use it in Elixir. This function is useful only during debugging or during side-effects. But if you are doing side-effects there is nothing stopping you from having a private function that does the job. If we take your example:

    ["thing1", "thing2"]
    |> generate_more_things()
    |> tee(&File.write!("things", &1))
    |> do_something_with_things()

I would write it as:

    ["thing1", "thing2"]
    |> generate_more_things()
    |> write_it_to_disk()
    |> do_something_with_things()

And to me that is clearer than a `tee`. So my concern is that by having those functions, we are encouraging people to write "sloppy pipelines".

In fact, I planned not long ago to send a proposal to elixir-core about adding two functions to Kernel, `tee`, which we discuss here, and `pipe`, which simply pipes the first argument to an anonymous function. So you could do:

    ["thing1", "thing2"]
    |> pipe(&generate_more_things("foo", &1))
    |> tee(&File.write!("things", &1))
    |> do_something_with_things()

And you probably see my point here. To me none of those replace writing proper private functions. I eventually gave up on writing the proposal but I supposed @kelvinst could create a project called `:pipe_tee` or `:plumbing` with those additions.

kelvinst:
Just realised my comparison is kind of unfair. Here is a fairer version with private functions:

```elixir
["thing1", "thing2"]
|> add_first_thing()
|> write_things_to_disk()
|> capitalize_things()
```

mememori:
I agree with @kelvinst and with @josevalim.

Albeit it might be an incentive for "sloppy pipelines", `tee` is great for debugging, scripts and simple side-effectful pipelines. I could see a lot of uses for `tee` to make the project i am currently working on a bit more readable on some bits.

So, i'd say it would be a good addition. LGTM

josevalim:
> I could see a lot of uses for tee to make the project i am currently working on a bit more readable on some bits.

That's a direction we can take. If users can post actual examples of pipelines before and after `tee`, it can really help the discussion move forward.

OvermindDL1:
> That's a direction we can take. If users can post actual examples of pipelines before and after tee, it can really help the discussion move forward.

For documentative purposes, the original use-case was:

```elixir
blah
|> do_something0()
|> tee(&IO.inspect(&1.vwoop))
|> do_something1()
```

yordis:
@josevalim I dont know if this would make sense but

I would use this with some async logger that I wouldn't care about the return value or any computation inside maybe?

```
data
|> do_something()
|> tee(&Logger.log(&1))
|> keep_going()
```

josevalim:
I understand there is a benefit in theory, but it is unclear if it would
actually be relevant or make the code better in practice. That’s why we are
asking for actual code samples. I am sure it makes pseudo code look fine. :)
-- 


josevalim:
Folks, since we haven't had actual code samples posted here, I will go ahead and close this. My suggestion is to move this discussion to a medium with more visibility, probably Elixir Forum, and see the use cases that pop out. Thanks!

josevalim:
Meanwhile, if someone wants to go ahead and create a project with the `tee/2` and `pipe/2` functions as I mentioned above, that's even better.

kelvinst:
> Meanwhile, if someone wants to go ahead and create a project with the tee/2 and pipe/2 functions as I mentioned above, that's even better.

Did it! https://hexdocs.pm/plumbing/

yordis:
@josevalim a couple of days after, I am in a position that I don't want to use @kelvinst package, not because it doesn't do what I need but because it will be another dependency to manage, so basically I will be copying this over to my package, so so far, the 3rd package on Hex that I know hat will include the exact same code.

To be honest,

I see the pattern of `|> tap( &log_error/1)` every where in my code because I combine it with other pipes so my code becomes cleaner.

The more piped style people use, the common the pattern of use `tap/tee` for logger is.


```elixir
some_value
|> My.business_do_something()
|> StrawHat.Response.tap_error(&log_error/1) # <-- this uses `tap/tee` under the hood
|> StrawHat.Response.tap(fn _ -> "my business did something" end) # <-- this uses `tap/tee` under the hood
```

I hope this actually gets into Elixir one day Jose, I really dont mind to copy over to the packages but so far has become part of my daily stack to use `tee` function.

PR #7225, opened 2018-01-18
Reorder keyword blocks in Macro.to_string

fishcakez:
Currently modules created with `Macro.to_string` will warn if rescue and catch because of the ordering. As we now get warnings if `catch` is before `rescue` I have changed the order to be more idiomatic.


josevalim:

PR #7229, opened 2018-01-19
Add :apply option to IO.inspect/3

mgartner:
This is an attempt to build functionality of IO.inspect/3 that was mentioned in the proposal here: https://groups.google.com/forum/#!topic/elixir-lang-core/TUkmNHI4IbI

The goal is to make something like this...

```elixir
["thing1", "thing2"]
|> generate_more_things()
|> (fn things ->
  things |> length() |> IO.inspect()
  things
end).()
|> do_something_with_things()
```

simpler to write, with something like this...

```elixir
["thing1", "thing2"]
|> generate_more_things()
|> IO.inspect(apply: &length/1)
|> do_something_with_things()
```

### DIFF
--- a/lib/elixir/lib/io.ex
+++ b/lib/elixir/lib/io.ex
@@ -259,6 +259,10 @@ defmodule IO do
   option to easily distinguish it from other `IO.inspect/2` calls.
   The label will be printed before the inspected `item`.
 
+  The `item` can be altered from its original form for inspection with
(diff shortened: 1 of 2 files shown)


josevalim:
Thanks @mgartner! There is no agreement yet if the feature will be added and, if so, in which shape, so I will go ahead and close this until there is a conclusion. If we agree this is the way to go, then we will reopen it, review it and merge it!

PR #7237, opened 2018-01-20
Set proper path of common options used in manpages

fertapric:
The path for the common options shared between `elixir` and `iex` was not changed when the `man/Makefile` was merged into the top-level `Makefile`.

Here is an screenshot of the change, "before" (left) and "after" (right):

screen shot 2018-01-20 at 12 55 27 pm

By the way, I installed Elixir v1.6.0 via Homebrew (Mac OS X) and it does not come with the manpages, is that intended?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -243,13 +243,13 @@ build_man: man/iex.1 man/elixir.1
 
 man/iex.1:
 	$(Q) cp man/iex.1.in man/iex.1
-	$(Q) sed -i.bak "/{COMMON}/r common" man/iex.1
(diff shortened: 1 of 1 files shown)


josevalim:
> By the way, I installed Elixir v1.6.0 via Homebrew (Mac OS X) and it does not come with the manpages, is that intended?

You need to ping whoever maintains the Homebrew bottle. :)

josevalim:

fertapric:
For further reference, here is the PR to update the Homebrew formulae https://github.com/Homebrew/homebrew-core/pull/23101

PR #7239, opened 2018-01-20
Add `@since` attributes for behavior introduced in 1.6

devonestes:


lib/elixir/lib/dynamic_supervisor.ex
```diff
@@ -463,6 +469,7 @@ defmodule DynamicSupervisor do
   ## Callbacks
 
   @impl true
+  @since "1.6.0"
```
josevalim: I don't think we should add @since to callbacks. :)

josevalim:

PR #7240, opened 2018-01-20
Perform unused variable checks during Elixir expansion

josevalim:
This gives us better control over when and how unused variables are printed.

As a result, we are able to emit unused variable warnings in situations
we could not before. This also opens up the way for us to remove a
dependency on erl_lint and track types information, which allows us to
speed up compilation times about 5% and allow us to emit more performant
code in some situations.

### DIFF
--- a/lib/elixir/src/elixir_env.erl
+++ b/lib/elixir/src/elixir_env.erl
@@ -1,7 +1,7 @@
 -module(elixir_env).
 -include("elixir.hrl").
--export([new/0, linify/1, env_to_scope/1, env_to_scope_with_vars/2]).
--export([mergea/2, mergev/2, merge_vars/2, merge_opt_vars/2]).
+-export([new/0, linify/1, env_to_scope/1, env_to_scope_with_vars/2, reset_vars/1]).
+-export([mergea/2, mergev/2]).
 
 new() ->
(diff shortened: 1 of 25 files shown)


lib/elixir/src/elixir_env.erl
```diff
@@ -16,50 +20,142 @@ new() ->
...
+
+check_unused_vars(#{unused_vars := Unused} = E) ->
+  [begin
+     case not_underscored(Name) of
+       true ->
+         elixir_errors:form_warn([{line, Line}], ?key(E, file), ?MODULE, {unused_var, Name, false});
+       false ->
+         ok
+     end
+   end || {{{Name, nil}, _}, Line} <- maps:to_list(Unused), Line /= false].
```
michalmuskala: Would it make sense to write this as:
```erl
check_unused_vars(#{unused_vaes := Unused, file := File}) ->
  [elixir_errors:form_warn([{line, Line}], File, ?MODULE, {unused_var, Name, false}) ||
    {{{Name, nil}, _}, Line} <- maps:to_list(Unused), Line /= false, not_underscored(Name)],
```

josevalim:
Merged manually, thanks for the review @michalmuskala!

PR #7241, opened 2018-01-21
Fix example of "Catching values of any kind"

ggpasqualino:
The code snippet in the section "Catching values of any kind" of the documentation of `try` had wrong indentation, because of that the snippet was wrongly formatted.

Before:
image

After:
image


josevalim:

PR #7242, opened 2018-01-21
Add typespec for Float.ratio/1

devonestes:
While going through and adding the `@since` attributes, I notice that
this function was missing a typespec, so I figured I'd add it.

I went with `{pos_integer, pos_integer} | {neg_integer, pos_integer}`
as the return type instead of `{integer, pos_integer}` to make it clear
that we would never have a numerator of 0.

### DIFF
--- a/lib/elixir/lib/float.ex
+++ b/lib/elixir/lib/float.ex
@@ -380,6 +380,7 @@ defmodule Float do
 
   """
   @since "1.4.0"
+  @spec ratio(float) :: {pos_integer, pos_integer} | {neg_integer, pos_integer}
   def ratio(float) when is_float(float) do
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/float.ex
```diff
@@ -380,6 +380,7 @@ defmodule Float do
 
   """
   @since "1.4.0"
+  @spec ratio(float) :: {pos_integer, pos_integer} | {neg_integer, pos_integer}
```
josevalim: `{pos_integer | neg_integer, pos_integer}` would be better then?

devonestes: Huh, didn't know you could do that. Totally better.

josevalim:

PR #7243, opened 2018-01-21
Change stdlib path to transparent for release/dev version

pprzetacznik:
If you build your otp version from sources you won't have stdlib path such as `"lib/stdlib-#{vsn}/src/"` but `"lib/stdlib/src/"`. In this case tests on development environment fails.

By using `{:code.lib_dir(:stdlib, :src)}` instead, you can get absolute path to the stdlib library so this solution is transparent for dev/release version.


josevalim:

PR #7244, opened 2018-01-22
Use the new child specification format in `DynamicSupervisor` docs since `Supervisor.Spec.worker/3` is deprecated.

benjamintanweihao:


josevalim:

lib/elixir/lib/dynamic_supervisor.ex
```diff
@@ -74,13 +74,14 @@ defmodule DynamicSupervisor do
...
 
-        def init(implicit_arg) do
+        def init(initial_arg) do
           children = [
-            worker(MyWorker, [implicit_arg])
+            # Or the deprecated: worker(MyWorker, [initial_arg])
+            %{id: MyWorker, start: {MyWorker, :start_link, [initial_arg]})
           ]
 
           supervise(children, strategy: :simple_one_for_one)
```
wojtekmach: I tried following the docs, with the code:

```elixir
  def init(initial_arg) do
    children = [
      %{id: MyWorker, start: {MyWorker, :start_link, [initial_arg]}}
    ]

    supervise(children, strategy: :simple_one_for_one)
  end
```

I got an error:

```
     ** (EXIT from #PID<0.173.0>) an exception was raised:
         ** (ArgumentError) invalid tuple specification given to supervise/2. If you are trying to use the map child specification that is part of the Elixir v1.5, use Supervisor.init/2 instead of Supervisor.Spec.supervise/2. See the Supervisor module for more information. Got: %{id: MyWorker, start: {MyWorker, :start_link, [[:one, :two, :three]]}}
             (elixir) lib/supervisor/spec.ex:186: Supervisor.Spec.get_id/1
             (elixir) lib/enum.ex:1294: Enum."-map/2-lists^map/1-0-"/2
             (elixir) lib/supervisor/spec.ex:177: Supervisor.Spec.supervise/2
             (stdlib) supervisor.erl:294: :supervisor.init/1
             (stdlib) gen_server.erl:365: :gen_server.init_it/2
             (stdlib) gen_server.erl:333: :gen_server.init_it/6
             (stdlib) proc_lib.erl:247: :proc_lib.init_p_do_apply/3
```

so maybe it's better to show just `worker(MyWorker, [implicit_arg])` and `supervise(children, strategy: :simple_one_for_one)`? On the other hand, this is just as an example for "old" way, so maybe it doesn't have to be 100% correct, just wondering.

lib/elixir/lib/dynamic_supervisor.ex
```diff
@@ -97,22 +98,22 @@ defmodule DynamicSupervisor do
         end
 
         def start_child(foo, bar, baz) do
-          # This will start child by calling MyWorker.start_link(implicit_arg, foo, bar, baz)
-          spec = Supervisor.Spec.worker(MyWorker, [foo, bar, baz])
+          # This will start child by calling MyWorker.start_link(initial_arg, foo, bar, baz)
+          spec = %{id: MyWorker, start: {MyWorker, :start_link, [foo, bar, baz]}}
```
wojtekmach: this could also be just: `spec = {MyWorker, [foo, bar, baz]}`

PR #7245, opened 2018-01-22
Update tests with new exclude/skipped behavior

devonestes:
I updated the tests with the new behavior that we're expecting by
excluding tests instead of skipping them. If any of this seems off, let
me know! It makes sense to me, but it's always good to get feedback
before I get too deep in the weeds.

### DIFF
--- a/lib/ex_unit/lib/ex_unit/cli_formatter.ex
+++ b/lib/ex_unit/lib/ex_unit/cli_formatter.ex
@@ -256,7 +256,10 @@ defmodule ExUnit.CLIFormatter do
 
     message =
       "#{test_type_counts}#{config.failure_counter} #{failure_pl}"
-      |> if_true(config.skipped_counter > 0, &(&1 <> ", " <> skipped("#{config.skipped_counter} skipped", config)))
+      |> if_true(
+        config.skipped_counter > 0,
+        &(&1 <> ", " <> skipped("#{config.skipped_counter} skipped", config))
+      )
       |> if_true(config.excluded_counter > 0, &(&1 <> ", #{config.excluded_counter} excluded"))
       |> if_true(config.invalid_counter > 0, &(&1 <> ", #{config.invalid_counter} invalid"))
 
--- a/lib/ex_unit/lib/ex_unit/cli_formatter.ex
+++ b/lib/ex_unit/lib/ex_unit/cli_formatter.ex
@@ -64,6 +64,19 @@ defmodule ExUnit.CLIFormatter do
     {:noreply, config}
   end
 
+  def handle_cast({:test_finished, %ExUnit.Test{state: {:skipped, _}} = test}, config) do
+    if config.trace do
+      IO.puts(trace_test_skipped(test))
+    else
+      IO.write(skipped("*", config))
+    end
+
+    test_counter = update_test_counter(config.test_counter, test)
+    config = %{config | test_counter: test_counter, skipped_counter: config.skipped_counter + 1}
+
+    {:noreply, config}
+  end
+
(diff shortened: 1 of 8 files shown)


lib/ex_unit/test/ex_unit_test.exs
```diff
@@ -158,20 +158,20 @@ defmodule ExUnitTest do
     assert output =~ "4 tests, 1 failure"
 
     {result, output} = run_with_filter([exclude: [even: true]], [ParityTest])
-    assert result == %{failures: 0, skipped: 1, total: 4}
```
josevalim: Let's add `:excluded` to the `result` map.

josevalim:
One tiny comment for now: I would add the `:excluded` key to the map and keep counting them towards total.

Maybe for the initial implementation, can simply replace all skip/skipped by excluded. Then you can introduce the new `skipped` functionality.

devonestes:
Ok, this is all updated with the addition of a the new concept of an "excluded" test.

Where do you think the best place to document this change is?

lib/ex_unit/lib/ex_unit/cli_formatter.ex
```diff
@@ -54,8 +55,21 @@ defmodule ExUnit.CLIFormatter do
...
+
+    test_counter = update_test_counter(config.test_counter, test)
+    config = %{config | test_counter: test_counter, excluded_counter: config.excluded_counter + 1}
+
+    {:noreply, config}
+  end
+
+  def handle_cast({:test_finished, %ExUnit.Test{state: {:skipped, _}} = test}, config) do
+    if config.trace do
+      IO.puts(trace_test_skipped(test))
```
josevalim: Should we color this yellow?

lib/ex_unit/lib/ex_unit/cli_formatter.ex
```diff
@@ -238,7 +256,8 @@ defmodule ExUnit.CLIFormatter do
 
     message =
       "#{test_type_counts}#{config.failure_counter} #{failure_pl}"
-      |> if_true(config.skipped_counter > 0, &(&1 <> ", #{config.skipped_counter} skipped"))
+      |> if_true(config.skipped_counter > 0, &(&1 <> ", " <> skipped("#{config.skipped_counter} skipped", config)))
```
josevalim: I think now we can color the whole thing in yellow? As you originally proposed?

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -140,7 +140,7 @@ defmodule ExUnit.Runner do
 
       case ExUnit.Filters.eval(include, exclude, tags, tests) do
         :ok -> %{test | tags: tags}
-        {:error, msg} -> %{test | state: {:skip, msg}}
+        excluded_or_skipped -> %{test | state: excluded_or_skipped}
```
josevalim: ❤️

josevalim:
I have added minor comments about style, everything else is perfect!

lib/ex_unit/lib/ex_unit/cli_formatter.ex
devonestes: Oh, I never really wanted to color the whole thing yellow, just the `1 skipped` part. Do you think the whole thing being yellow is better?

lib/ex_unit/lib/ex_unit/cli_formatter.ex
devonestes: Yeah, totally. I'll update that.

lib/ex_unit/lib/ex_unit/cli_formatter.ex
josevalim: I am fine with any approach.

josevalim:

PR #7247, opened 2018-01-23
Add IO.tee

Un3qual:
`IO.tee` inspects an item identically to `IO.inspect`, but it also returns the item. This is useful for inspecting in the middle of a chain of pipes.

### DIFF
--- a/lib/elixir/lib/io.ex
+++ b/lib/elixir/lib/io.ex
@@ -310,6 +310,29 @@ defmodule IO do
     puts(device, [label, chardata])
     item
   end
+  
+  @doc """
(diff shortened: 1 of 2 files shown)


yordis:
@josevalim second time I see a PR asking for `tee` thing in less than a week 😅

fbergero:
`IO.inspect` does exactly that (returns the item). 
I would push for a `Kernel.tee/2` which does
```
def tee(val, fun) do
   fun.(val)
   val
end
```

Un3qual:
@fbergero Whoops. I don't know how I missed this. For some reason, I thought it didn't return. Sorry. :shrug:

PR #7252, opened 2018-01-24
Add missing typespecs to `File` module

devonestes:
I noticed a few typespecs were missing for public functions in the
`File` module, so I went ahead and added them.

### DIFF
--- a/lib/elixir/lib/file.ex
+++ b/lib/elixir/lib/file.ex
@@ -528,6 +528,7 @@ defmodule File do
   `{:error, :enotsup}`.
   """
   @since "1.5.0"
+  @spec ln(Path.t(), Path.t()) :: :ok | {:error, posix}
   def ln(existing, new) do
     :file.make_link(IO.chardata_to_string(existing), IO.chardata_to_string(new))
   end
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/file.ex
```diff
@@ -1515,6 +1519,7 @@ defmodule File do
   See `Stream.run/1` for an example of streaming into a file.
 
   """
+  @spec stream!(Path.t(), [atom], atom | pos_integer) :: File.Stream.t()
```
AndrewDryga: I think there is an existing `t:mode/0`. Maybe it's better to reuse it and or define a new one that explicitly tells which atoms are allowed? Same for `line_or_bytes`, type spec could be something like `:line | pos_integer`.

devonestes: Do you think it's worth defining a type for `line_or_bytes` here even though that type isn't being used anywhere else other than this function?

The `mode` type as defined doesn't have `:trim_bom` as a possible mode, so we'll need to include that as an option.

Do you think `@spec stream!(Path.t(), [mode | :trim_bom], :line | pos_integer) :: File.Stream.t()` would be a helpful typespec?

josevalim:

lib/elixir/lib/file.ex
AndrewDryga: Jose already merged it, so it was ok. Generally speaking, I believe that having strict spec would allow static analysis tools (like Dyalizer) discover bugs where you, for eg., use mode which is not supported. It's not that important how you define them.

PR #7253, opened 2018-01-24
Illustrate specifying function argument names in Writing Documentation docs

michallepicki:
The previous example wasn't changing the name because `map` was inferred so I was confused about how this feature is supposed to work.


josevalim:

PR #7256, opened 2018-01-25
Add missing documentation, examples and specs to Calendar.ISO

devonestes:
I noticed that some public functions in the Calendar.ISO module
were missing typespecs, documentation and examples of use, so I added
them.

### DIFF
--- a/lib/elixir/lib/calendar/iso.ex
+++ b/lib/elixir/lib/calendar/iso.ex
@@ -264,7 +264,22 @@ defmodule Calendar.ISO do
 
   @doc """
   Converts the given time into a string.
+
+  ## Examples
+
+      iex> Calendar.ISO.time_to_string(2, 2, 2, {2, 6})
+      "02:02:02.000002"
+      iex> Calendar.ISO.time_to_string(2, 2, 2, {2, 2})
+      "02:02:02.00"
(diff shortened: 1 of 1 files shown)


josevalim:
Thanks! Note though we don’t need to add specs if they are implementation
of callbacks. Can you please remove the specs in those cases?
-- 


devonestes:
@josevalim The reason I added those is because there were many other implementations of callback that already had typespecs in that module so I thought they should be there. For example, [Calendar.ISO. naive_datetime_from_iso_days/7](https://github.com/devonestes/elixir/blob/5dfb9ce041f2aeba7fec8f063bd43743133dc052/lib/elixir/lib/calendar/iso.ex#L78). Want me to remove all of the typespecs for any function that's a callback implementation?

Personally I find seeing the typespecs really helpful in the documentation, and as it stands now if a function is an implementation of a callback the typespec just doesn't show up at all. Finding the proper usage for those functions is a little tricky. For example, check out the current docs for [Calendar.ISO.time_to_string/4](https://hexdocs.pm/elixir/Calendar.ISO.html#time_to_string/4).

Would it be possible for ExDoc to add the typespec information to callback implementations automatically?

josevalim:
Ah, good point. We actually tried to show callbacks on the implementations but we ran into issues because of local references to other types.

josevalim:

PR #7257, opened 2018-01-25
Flatten binary literals when optimizing concatenation

michalmuskala:
Before, code like following:

    "foo" <> "bar #{x} baz"

Would create two strings and then concatenate them, after the change
this creates just one final string. The previous optimization did not
account for binary constructors, just for binary literals.


josevalim:
@michalmuskala did you check the generated ASM? I am almost sure that we inline those either during the elixir expansion or the erlang conversion.

Something else to have in mind is that this should fail `<<0>> <> <<1::1>>` although I am open to make it the bitstring concatenation operator.

michalmuskala:
Yes, I did check the assembly, that's why I got an idea to make this.

And yes, you're also right that concatenating bitstrings creates problems here. Let me see, if I can fix it without relaxing the operator.

PR #7263, opened 2018-01-27
Add Enum.count/2 function for given element

crabonature:
What you think about such new function for simplify counting elements?

Right now I can count given element by calling count/2 by with defining function: 

```elixir
# when I just want to count 3
Enum.count([1, 3, 2, 3, 4, 3], &(&1 === 3)) 
# or
Enum.count([1, 3, 2, 3, 4, 3], fn(entry) -> entry === 3))

```
After chages code can be more understandable:

```elixir
Enum.count([1, 3, 2, 3, 4, 3], 3) # when I just want to count 3
```

Side note: Maybe better name for `count` with function as a second argument could be `count_by` some time in future?

### DIFF
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -576,11 +576,28 @@ defmodule Enum do
 
   """
   @spec count(t, (element -> as_boolean(term))) :: non_neg_integer
-  def count(enumerable, fun) do
+  def count(enumerable, fun) when is_function(fun) do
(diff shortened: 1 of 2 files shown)


josevalim:
Please use the mailing list for feature proposals. But i personally see no reason to support such entry as passing an anonymous function is more general.

crabonature:
Forget to read about contributing in here, got it mailing list first for features.

Issue #7246, opened 2018-01-22
Forcing Logger Sync


benwilson512:
Hey folks, in talking with @fishcakez it was unclear which precise set of logger configurations might force it to operate in sync mode. A forced sync mode can be useful for example in development when using it to correlate database access with other specific function calls.

This may be a simple matter of documentation, or if there isn't any such logger configuration, a change to enable this.

josevalim:
Probably setting sync_threshold to 0 is enough. But we would need to document it and write a test.

tontytwo:
Hi, I want to contribute to elixir, is this a good task for a beginner like me?

josevalim:
@tontytwo you will have to try it to be sure. :)

Here is the place where we would write a test to verify this behaviour: https://github.com/elixir-lang/elixir/blob/master/lib/logger/test/logger/utils_test.exs#L14

The documentation should be added to the `Logger` module.

Have fun!

PR #7264, opened 2018-01-27
Forcing Logger Sync

tontytwo:
Added tests and docs for #7246

### DIFF
--- a/lib/logger/test/logger/utils_test.exs
+++ b/lib/logger/test/logger/utils_test.exs
@@ -19,6 +19,9 @@ defmodule Logger.UtilsTest do
       assert compute_mode(:async, 20, 15, 20, 7500, 10000) == :sync
       assert compute_mode(:async, 30, 15, 20, 7500, 10000) == :sync
       assert compute_mode(:async, 8000, 15, 20, 7500, 10000) == :sync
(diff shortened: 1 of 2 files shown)


lib/logger/test/logger/utils_test.exs
```diff
@@ -19,6 +19,9 @@ defmodule Logger.UtilsTest do
       assert compute_mode(:async, 20, 15, 20, 7500, 10000) == :sync
       assert compute_mode(:async, 30, 15, 20, 7500, 10000) == :sync
       assert compute_mode(:async, 8000, 15, 20, 7500, 10000) == :sync
+      assert compute_mode(:async, 0, 15, 0, 7500, 10000) == :sync
```
josevalim: It should be:

    compute_mode(:async, 0, 0, 0, 7500, 10000)

That's because the number 15 is always computed in function of `sync_threshold`. We also need to change the examples below. Thanks!

tontytwo: okay, also the build is failing in OTP 19.1, is that something which can be avoided?

josevalim:
It seems to be a temporary failure, as it passes when it is re-run. I will
investigate it later.
-- 

PR #7274, opened 2018-01-28
Warn about comparison with calendar structs

michalmuskala:
People are frequently confused about results of comparison of structs. This is frequently experienced with calendar structs. This PR tries to catch common erroneous expressions in the compiler and issue warnings at compile-time.

### DIFF
--- a/lib/elixir/src/elixir_expand.erl
+++ b/lib/elixir/src/elixir_expand.erl
@@ -710,6 +710,7 @@ expand_remote(Receiver, DotMeta, Right, Meta, Args, #{context := Context} = E, E
                                  ?key(E, function), ?line(Meta), ?key(E, lexical_tracker)),
   {EArgs, EA} = expand_args(Args, E),
   Rewritten = elixir_rewrite:rewrite(Receiver, DotMeta, Right, Meta, EArgs),
+  maybe_warn_struct_comparison(Rewritten, E),
   case allowed_in_context(Rewritten, Arity, Context) of
     true ->
       {Rewritten, elixir_env:mergev(EL, EA)};
@@ -727,6 +728,40 @@ allowed_in_context(_, _Arity, guard) ->
   false;
 allowed_in_context(_, _, _) ->
   true.
+
+maybe_warn_struct_comparison({{'.', _, [erlang, Op]}, Meta, [Left, Right]}, E)
+    when Op =:= '>'; Op =:= '<'; Op =:= '=<'; Op =:= '>=' ->
+  Result =
+    case calendar_struct_expression(Left) of
+      false ->
+        case calendar_struct_expression(Right) of
+          false -> false;
+          Struct -> Struct
+        end;
+      Struct -> Struct
+    end,
+
+  case Result of
+    false -> ok;
+    _ -> elixir_errors:form_warn(Meta, ?key(E, file), ?MODULE, {struct_comparison, Result})
+  end;
+maybe_warn_struct_comparison(_Other, _E) ->
+  ok.
+
+calendar_struct_expression({'%{}', _, KVs}) ->
+  case lists:keyfind('__struct__', 1, KVs) of
+    {'__struct__', Struct} when Struct =:= 'Elixir.Time';
+                                Struct =:= 'Elixir.DateTime';
+                                Struct =:= 'Elixir.NaiveDateTime';
+                                Struct =:= 'Elixir.Date' ->
+      Struct;
+    false -> false
+  end;
+calendar_struct_expression({{'.', _, ['Elixir.Time', utc_now]}, _, _}) -> 'Elixir.Time';
+calendar_struct_expression({{'.', _, ['Elixir.DateTime', utc_now]}, _, _}) -> 'Elixir.DateTime';
+calendar_struct_expression({{'.', _, ['Elixir.NaiveDateTime', utc_now]}, _, _}) -> 'Elixir.NaiveDateTime';
+calendar_struct_expression({{'.', _, ['Elixir.Date', utc_today]}, _, _}) -> 'Elixir.Date';
+calendar_struct_expression(_Other) -> false.
 
 %% Lexical helpers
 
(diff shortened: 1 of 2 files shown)


josevalim:
@michalmuskala I really don't like hardcoding those structs in the compiler. Maybe we should have a proposal that generalizes this? Maybe something like `@comparable false`?

josevalim:
After some discussion, we have decided to warn every time there is a literal struct on the left or right side of `>`, `<`, `>=` and `<=`.

OvermindDL1:
Entirely agree with warning every time.  I'd also think it would be useful to create a `Comparable` protocol implemented for everything (with standard erlang term ordering) except for structs, which need to either opt-in to erlang term ordering or explicitly implement it.  (Thought that's for another discussion and PR.)

michalmuskala:
I started looking into this, but I'm not sure what message we could give here for an arbitrary map.

OvermindDL1:
> I started looking into this, but I'm not sure what message we could give here for an arbitrary map.

For an arbitrary map?  Why not just follow erlang term ordering for that?  I'd think the only unique parts should be around structs?

michalmuskala:
Ah yeah, sorry. Only a struct. But the question still stands. Since it's an arbitrary struct, we can't really suggest looking for a `compare/2` function because it might not be there.

OvermindDL1:
> Ah yeah, sorry. Only a struct. But the question still stands. Since it's an arbitrary struct, we can't really suggest looking for a compare/2 function because it might not be there.

Hmm, an idea, if there is a literal struct being compared to, perhaps delegating that to a compare function on the structs module (if it exists, else error saying that this comparison is ill-defined) might be useful?

michalmuskala:
Updated

lib/elixir/src/elixir_expand.erl
```diff
@@ -1020,4 +1039,8 @@ format_error({underscored_var_access, Name, Kind}) ->
   io_lib:format("the underscored variable \"~ts\"~ts is used after being set. "
                 "A leading underscore indicates that the value of the variable "
                 "should be ignored. If this is intended please rename the "
-                "variable to remove the underscore", [Name, context_info(Kind)]).
+                "variable to remove the underscore", [Name, context_info(Kind)]);
+format_error(struct_comparison) ->
+  "comparison operators (>, <, >=, <=) perform structural and not semantic comparison, "
+  "comparing with a struct literal is unlikely to give a meaningful result. "
+  "Many modules export a compare/2 or cmp/2 function that can be used for semantic comparison".
```
josevalim: I don't feel like we should talk about `cmp/2` since we have settled on `compare/2`. It is also nice to give the name of the struct and start the error message with a small sentence. Here is my suggestion:

> invalid comparison with struct literal ~ts. Comparison operators (>, <, >=, <=) perform structural
> and not semantic comparison. Comparing with a struct literal is unlikely to give a meaningful result.
> Modules typically define a compare/2 function that can be used for semantic comparison.

michalmuskala:
@josevalim I've updated the wording

josevalim:

lib/elixir/src/elixir_expand.erl
```diff
@@ -728,6 +729,30 @@ allowed_in_context(_, _Arity, guard) ->
...
+    when Op =:= '>'; Op =:= '<'; Op =:= '=<'; Op =:= '>=' ->
+  Result =
+    case struct_expression(Left) of
+      false -> struct_expression(Right);
+      Struct -> Struct
+    end,
+
+  case Result of
+    false -> ok;
+    _ -> elixir_errors:form_warn(Meta, ?key(E, file), ?MODULE, {struct_comparison, Result})
```
wojtekmach: currently the warning is:

```elixir
iex(1)> Date.utc_today() >= %Date{year: 2018, month: 1, day: 1}
warning: invalid comparison with struct literal Elixir.Date. Comparison operators (>, <, >=, <=) perform structural and not semantic comparison. Comparing with a struct literal is unlikely to give a meaningful result. Modules typically define a compare/2 function that can be used for semantic comparison
  iex:1

true
```

I'd be happy to send a patch with `elixir_aliases:inspect(Result)` to at least make it to `(...) with struct literal Date. (...)`.

With `'Elixir.Macro':to_string(Right)` (or Left) we could turn it to `(...) with struct literal %Date{calendar: Calendar.ISO, year: 2018, month: 1, day: 1}`. Maybe it would make sense to go all the way back to sigil?

wojtekmach: by trial and error :) I got to:

> warning: invalid comparison with struct literal ~D[2018-01-01]

with:

```erlang
      {M, _} = 'Elixir.Code':eval_quoted(Right),
      R = 'Elixir.Kernel':inspect(M),
```

not familar with internals; if it's worth pursuing what would be appropriate APIs to call here?

josevalim: Let's go with `elixir_aliases:inspect(Result)` and an updated test. Thanks!

josevalim: The second one seems a bit risky. :)

wojtekmach:
sounds good!

btw, I got it to work (still using eval, but I'm sure there's a better way):

```
iex(2)> 1 > ~D[2018-01-01]
warning: invalid comparison with struct literal in 1 > ~D[2018-01-01]. (...)
```

but hit a roadblock here:

```
iex(1)> 1 > %File.Stat{}
warning: invalid comparison with struct literal in 1 > %File.Stat{access: nil, atime: nil, ctime: nil, gid: nil, inode: nil, links: nil, major_device: nil, minor_device: nil, mode: nil, mtime: nil, size: nil, type: nil, uid: nil}. (...)
```

michalmuskala:
@wojtekmach for that to work, I think the simplest way would be to grab the operands before macro expansion - that way you get the original AST. After deciding which side is problematic, you should be able to print it using `'Elixir.Macro':to_string(Expr)`. The key is grabbing it before the expansion.

josevalim:
> The key is grabbing it before the expansion.

Good call. Let's do that and go with Macro.to_string.

sheharyarn:
Shouldn't we always warn when comparing structs directly, literal or not? For example, consider sorting a list of dates:

```elixir
Enum.sort(list_of_datetimes)
```

I also like @OvermindDL1's proposal to have a `Comparable` protocol with implementations for common things like `DateTime`, and letting it simply throw `Protocol.UndefinedError` for structs which do not have explicitly implemented it.

michalmuskala:
The problem is that it's not possible to both have custom logic and have comparison operators working in guards, since guards can't execute custom code. Additionally it could have a rather considerable runtime performance cost to inject such code in every comparison. A static compile-time check is the best we can do in current environment.

OvermindDL1:
> Additionally it could have a rather considerable runtime performance cost to inject such code in every comparison.

I certainly wouldn't propose such a thing, I just mean to make such a `cmp` call that calls into such a `Comparable` protocol to become standard enough that people get in the habit of always using it for structs.

PR #7276, opened 2018-01-28
Implement String.center/3, String.center/2

dunyakirkali:
Implemented String.center/3 and String.center/2

See:
https://groups.google.com/forum/#!topic/elixir-lang-core/Gr6Xxi7lsWo

### DIFF
--- a/lib/elixir/lib/string.ex
+++ b/lib/elixir/lib/string.ex
@@ -2271,6 +2271,7 @@ defmodule String do
   Centers str in width. If width is greater than the length of str, returns a new String of length width with str centered and padded with padstr; otherwise, returns str.
 
   ## Examples
+
       iex> String.center("fin", 9, "~")
       "~~~fin~~~"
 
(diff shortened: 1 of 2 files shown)


josevalim:
Thanks @dunyakirkali!

I see you have sent a proposal to the mailing list but we did not have enough discussion around the feature, which usually means it has limited application. For this reason, we will close the pull request but we can always reopen it in case there is a surge in interest.

In the future, in case you want to avoid working on something that may not be merged, it is best to wait for an explicit approval of maintainer, asking a PR to be sent. :)

dunyakirkali:
Thanx @josevalim.

Just wanted to be a little proactive :) Next time I'll wait until I get a clear go 👍

josevalim:
It is ok to be proactive and not a problem at all - as long as it is clear there is no expectation of merging! :)

dunyakirkali:
No, no. No expectations :) I'm just trying to find small tasks for my self in order to get familiar with elixir and help out whilst learning. So if u can think of anything else that I could help u out with, please let me know

Issue #7267, opened 2018-01-27
b() helper on IEx does not show optional callback for GenServer


victorolinasc:
### Environment

* Elixir & Erlang versions (elixir --version): 
erlang: 20.2 (asdf/kerl)
elixir: 1.6-otp-20 (asdf) 

* Operating system: 
Linux x64 (Fedora 27)

### Current behavior

```sh
Erlang/OTP 20 [erts-9.2] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:10] [hipe] [kernel-poll:false]

Interactive Elixir (1.6.0) - press Ctrl+C to exit (type h() ENTER for help)
iex(1)> b GenServer
@callback code_change(old_vsn, state :: term(), extra :: term()) ::
            {:ok, new_state :: term()}
            | {:error, reason :: term()}
            | {:down, term()}
          when old_vsn: term()

@callback format_status(reason, pdict_and_state :: list()) :: term()
          when reason: :normal | :terminate

@callback handle_call(request :: term(), from(), state :: term()) ::
            {:reply, reply, new_state}
            | {:reply, reply, new_state, timeout() | :hibernate}
            | {:noreply, new_state}
            | {:noreply, new_state, timeout() | :hibernate}
            | {:stop, reason, reply, new_state}
            | {:stop, reason, new_state}
          when reply: term(), new_state: term(), reason: term()

@callback handle_cast(request :: term(), state :: term()) ::
            {:noreply, new_state}
            | {:noreply, new_state, timeout() | :hibernate}
            | {:stop, reason :: term(), new_state}
          when new_state: term()

@callback handle_info(msg :: :timeout | term(), state :: term()) ::
            {:noreply, new_state}
            | {:noreply, new_state, timeout() | :hibernate}
            | {:stop, reason :: term(), new_state}
          when new_state: term()

@callback init(args :: term()) ::
            {:ok, state}
            | {:ok, state, timeout() | :hibernate}
            | :ignore
            | {:stop, reason :: any()}
          when state: any()

@callback terminate(reason, state :: term()) :: term()
          when reason: :normal | :shutdown | {:shutdown, term()}

iex(2)>
```

### Expected behavior

Information on which callbacks are optional. In the case of `GenServer` that would be `format_status/2`.

Not sure the best output though \o/

josevalim:
Thanks for the report. I think we should show all callbacks first and show the optional ones at the end. Something like this:

```elixir
@callback foo(...) :: ...

[...]

PR #7282, opened 2018-01-30
Add optional callbacks to IEx.Introspection.b/1 output

victorolinasc:
This is an attempt of solving #7267.

Output:
```
iex(1) b GenServer
@callback code_change(old_vsn, state :: term(), extra :: term()) ::
            {:ok, new_state :: term()}
            | {:error, reason :: term()}
            | {:down, term()}
          when old_vsn: term()

@callback format_status(reason, pdict_and_state :: list()) :: term()
          when reason: :normal | :terminate

@callback handle_call(request :: term(), from(), state :: term()) ::
            {:reply, reply, new_state}
            | {:reply, reply, new_state, timeout() | :hibernate}
            | {:noreply, new_state}
            | {:noreply, new_state, timeout() | :hibernate}
            | {:stop, reason, reply, new_state}
            | {:stop, reason, new_state}
          when reply: term(), new_state: term(), reason: term()

@callback handle_cast(request :: term(), state :: term()) ::
            {:noreply, new_state}
            | {:noreply, new_state, timeout() | :hibernate}
            | {:stop, reason :: term(), new_state}
          when new_state: term()

@callback handle_info(msg :: :timeout | term(), state :: term()) ::
            {:noreply, new_state}
            | {:noreply, new_state, timeout() | :hibernate}
            | {:stop, reason :: term(), new_state}
          when new_state: term()

@callback init(args :: term()) ::
            {:ok, state}
            | {:ok, state, timeout() | :hibernate}
            | :ignore
            | {:stop, reason :: any()}
          when state: any()

@callback terminate(reason, state :: term()) :: term()
          when reason: :normal | :shutdown | {:shutdown, term()}

@optional_callbacks [format_status: 2]

iex(2)> b Application
@callback start(start_type(), start_args :: term()) ::
            {:ok, pid()} | {:ok, pid(), state()} | {:error, reason :: term()}

@callback start_phase(phase :: term(), start_type(), phase_args :: term()) ::
            :ok | {:error, reason :: term()}

@callback stop(state()) :: term()

@optional_callbacks [start_phase: 3]
```

I've taken the info on optional_callbacks from `mod.behaviour_info(:optional_callbacks)`.

### DIFF
--- a/lib/iex/lib/iex/introspection.ex
+++ b/lib/iex/lib/iex/introspection.ex
@@ -457,10 +457,19 @@ defmodule IEx.Introspection do
   """
   def b(mod) when is_atom(mod) do
     case get_callback_docs(mod, fn _ -> true end) do
-      :no_beam -> no_beam(mod)
-      :no_docs -> no_docs(mod)
-      {:ok, []} -> puts_error("No callbacks for #{inspect(mod)} were found")
-      {:ok, docs} -> Enum.each(docs, fn {definition, _} -> IO.puts(definition) end)
+      :no_beam ->
+        no_beam(mod)
+
+      :no_docs ->
+        no_docs(mod)
+
+      {:ok, []} ->
+        puts_error("No callbacks for #{inspect(mod)} were found")
+
+      {:ok, docs} ->
+        docs
+        |> add_optional_callback_docs(mod)
+        |> Enum.each(fn {definition, _} -> IO.puts(definition) end)
     end
 
     dont_display_result()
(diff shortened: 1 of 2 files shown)


josevalim:
@victorolinasc thanks! :heart:

Can you please show the optional callbacks just after the `@optional_callbacks` entry?

lib/iex/lib/iex/introspection.ex
```diff
@@ -521,10 +521,33 @@ defmodule IEx.Introspection do
               {format_callback(kind, fun, {fun, arity}, callbacks), doc}
           end)
 
+        docs =
```
josevalim: We need to move this to one specific clause inside `b/1`, otherwise we will show optional callbacks information even when doing something like `b GenServer.handle_call`.

victorolinasc:
You mean, in the case of GenServer right after `format_status/2`? 

But what if there are more than one and they are out of order?

lib/iex/lib/iex/introspection.ex
victorolinasc: Okay! Will do it.

michalmuskala: I wonder if that's a bad thing. I think it would be nice to have `b GenServer.format_status` show that it's an optional callback.

josevalim:

PR #7287, opened 2018-01-31
Add missing & in port zombie script

fhunleth:
This is needed to background the processing being run so that control
could move on to the read loop.


adkron:
:+1:

josevalim:

PR #7289, opened 2018-01-31
Defines the prep_stop/1 optional callback for applications

fxn:
This PR defines the `prep_stop/1` optional callback for application callback modules. See the [original Erlang/OTP documentation](http://erlang.org/doc/apps/kernel/application.html#Module:prep_stop-1). As of this writing, the application master invokes that callback [here](https://github.com/erlang/otp/blob/1fe1c9a57d365c8203cb1380e4bdfe8f7f6ab8ce/lib/kernel/src/application_master.erl#L377).

In some places, the wording is deliberately a bit vague in identifying applications with application callback modules, to be coherent with the existing style (an application is a broader concept and may not have a callback module associated, you know).

The patch also complements the description of the argument of `stop/1` when `start/2` does not return a state.

The description mentions `kernel` in passing. I believe that way help build a mental model that is good enough. If you say nothing, the reader may be wondering who is doing that. There was a spot that said "the VM", maybe too generic. Some Erlang docs say "the application master", which is more specific and probably the best option, but this documentation does not introduce this concept by now, so you cannot use it. `kernel` seemed like a sweet spot to me in the current docs.

### DIFF
--- a/lib/elixir/lib/application.ex
+++ b/lib/elixir/lib/application.ex
@@ -79,10 +79,20 @@ defmodule Application do
     * [`:application` module](http://www.erlang.org/doc/man/application.html)
     * [Applications – OTP Design Principles](http://www.erlang.org/doc/design_principles/applications.html)
 
-  A developer may also implement the `stop/1` callback (automatically defined
-  by `use Application`) which does any application cleanup. It receives the
-  application state and can return any value. Note that shutting down the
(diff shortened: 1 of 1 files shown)


josevalim:

Thank you! notice i replaced the "stopped by `kernel`" by runtime because of potential confusion with Elixir's `Kernel`. plus we don't quite need to know which application does the shutdown. :)

PR #7290, opened 2018-01-31
Teach Mix erlang compiler alternative spelling for -behavior declaration

prefiks:
Erlang recognizes both -behaviour and -behavior, but mix was only able to
process one, which made erlang compiler complain about missing behaviour
when source file was using the other one


josevalim:

Issue #7294, opened 2018-01-31
`mix test –-only inexistent_tag` gives no warning/error


pablobcb:
Not sure if this is right place to talk about it, tried on the forum and slack but got no answer
### Environment

* Elixir & Erlang versions (elixir --version): 
Elixir 1.6.0 (compiled with OTP 19)

Erlang/OTP 20 [erts-9.2] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:10] [hipe] [kernel-poll:false]

* Operating system: Mac OS X 10.12.6

### Current behavior

`mix test –only inexistent_test` runs fine even if there is not test named `inexistent_test`

### Expected behavior

I think it would be better if some hint/warning/error to notify the user that the desired test is not going to run

josevalim:
We could indeed emit a error. We should likely error anytime no test runs at all.

josevalim:
Question: should we do this in `mix test` or inside ExUnit itself? Probably the former.

hlhr:
I checked the code of the `mix test`, it would be possible to emit a error if the only option is used and the excluded and total number of tests is equal.

If this is a possible solution I would give it a try.

josevalim:
That sounds good to me!
-- 


*José Valimwww.plataformatec.com.br
<http://www.plataformatec.com.br/>Founder and Director of R&D*

josevalim:
Closing in favor of the PR.

PR #7295, opened 2018-02-01
Mix test with only option raise a error if no tests matched

hlhr:
Fix for issue #7294 

`mix test` raise a error if the only option is used and the excluded and total number of tests is equal

### DIFF
--- a/lib/mix/lib/mix/tasks/test.ex
+++ b/lib/mix/lib/mix/tasks/test.ex
@@ -282,11 +282,11 @@ defmodule Mix.Tasks.Test do
             System.at_exit(fn _ -> exit({:shutdown, 1}) end)
 
           excluded == total and option_only_present and opts[:raise] ->
-              Mix.raise("no test matched only option")
-          
+            Mix.raise("no test matched only option")
+
           excluded == total and option_only_present ->
-              Mix.shell().error("no test matched only option")
-              System.at_exit(fn _ -> exit({:shutdown, 1}) end)
+            Mix.shell().error("no test matched only option")
+            System.at_exit(fn _ -> exit({:shutdown, 1}) end)
 
           true ->
             :ok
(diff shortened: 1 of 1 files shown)


lib/mix/lib/mix/tasks/test.ex
```diff
@@ -269,16 +269,25 @@ defmodule Mix.Tasks.Test do
     display_warn_test_pattern(matched_warn_test_files, test_pattern)
 
     case CT.require_and_run(files, matched_test_files, test_paths, opts) do
-      {:ok, %{failures: failures}} ->
+      {:ok, %{excluded: excluded, failures: failures, total: total}} ->
         cover && cover.()
 
+        option_only_present = Keyword.has_key?(opts, :only)
```
josevalim: You can name the variable with a question mark too `option_only_present?`.

lib/mix/lib/mix/tasks/test.ex
```diff
@@ -269,16 +269,25 @@ defmodule Mix.Tasks.Test do
...
+
         cond do
           failures > 0 and opts[:raise] ->
             Mix.raise("mix test failed")
 
           failures > 0 ->
             System.at_exit(fn _ -> exit({:shutdown, 1}) end)
 
+          excluded == total and option_only_present and opts[:raise] ->
+            Mix.raise("no test matched only option")
```
josevalim: We can clarify this message a bit, what do you think about: "The --only option was given to "mix test" but no test executed"

We would need to change it below too!

lib/mix/lib/mix/tasks/test.ex
hlhr: Thanks for the feedback, changed it.

lib/mix/lib/mix/tasks/test.ex
hlhr: Sounds definitely clearer, changed it.

josevalim:

PR #7296, opened 2018-02-01
Fix two Macro.to_string tests

gusaiani:
Fix two Macro.to_string tests that weren't really testing much.


josevalim:

PR #7297, opened 2018-02-01
Unify *.to_iso8601 functions for calendar types

michalmuskala:
* Fastlane structs already in the ISO calendar
* crash with `FunctionClauseError` on bad format. This should actually improve the error message in dev and test given exception blaming. Previously the "bad format" argument error would be also raised when the function was given not a calendar struct (but, for example an erlang style tuple) - very confusing!


josevalim:

PR #7299, opened 2018-02-01
Update doc for DynamicSupervisor.start_child

rossta:
The example makes it appear like the arguments to `def start_child`: `foo, bar, baz` are being ignored by the line `spec = MyWorker`. I *think* this would be the correct way to delegate the module and arguments to `DynamicSupervisor.start_child` in the implied case that MyWorker is using the new child specs. If not, I'd be curious about the preferred approach.


josevalim:

PR #7309, opened 2018-02-03
Use contextual vars for __CALLER__

michalmuskala:
I tried using the `contextual_vars` introduced along with `__STACKTRACE__` to give better error messages for `__CALLER__`. I think, I have the consumption part ready, but I can't really understand where I would set it, that the variable is available.

Any tips welcome.

### DIFF
--- a/lib/elixir/src/elixir_expand.erl
+++ b/lib/elixir/src/elixir_expand.erl
@@ -112,15 +112,12 @@ expand({'__MODULE__', _, Atom}, E) when is_atom(Atom) ->
   {?key(E, module), E};
 expand({'__DIR__', _, Atom}, E) when is_atom(Atom) ->
   {filename:dirname(?key(E, file)), E};
-expand({'__CALLER__', _, Atom} = Caller, E) when is_atom(Atom) ->
+expand({'__CALLER__', Meta, Atom} = Caller, E) when is_atom(Atom) ->
+  assert_contextual_var(Meta, '__CALLER__', E, caller_not_allowed),
   {Caller, E};
 expand({'__STACKTRACE__', Meta, Atom} = Stacktrace, E) when is_atom(Atom) ->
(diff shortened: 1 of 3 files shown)


lib/elixir/src/elixir_expand.erl
```diff
@@ -112,15 +112,12 @@ expand({'__MODULE__', _, Atom}, E) when is_atom(Atom) ->
...
+  assert_contextual_var(Meta, '__CALLER__', E, caller_not_allowed),
   {Caller, E};
 expand({'__STACKTRACE__', Meta, Atom} = Stacktrace, E) when is_atom(Atom) ->
-  case lists:member('__STACKTRACE__', ?key(E, contextual_vars)) of
-    true ->
-      {Stacktrace, E};
-    false ->
-      form_error(Meta, ?key(E, file), ?MODULE, stacktarce_not_allowed)
-  end;
+  assert_contextual_var(Meta, '__STACKTRACE__', E, stacktarce_not_allowed),
```
josevalim: btw, there is a typo here: `stacktarce_not_allowed`. It should be stacktrace.

lib/elixir/src/elixir_expand.erl
```diff
@@ -1066,5 +1071,7 @@ format_error({struct_comparison, StructExpr}) ->
                 "Comparing with a struct literal is unlikely to give a meaningful result. "
                 "Modules typically define a compare/2 function that can be used for "
                 "semantic comparison", [String]);
+format_error(caller_not_allowed) ->
+  "__CALLER__ is available only inside macros";
```
josevalim: "only inside defmacro/defmacrop".

josevalim:
@michalmuskala you would set it here based on the value in `Kind` (`defmacro` or `defmacrop`): https://github.com/elixir-lang/elixir/blob/master/lib/elixir/src/elixir_def.erl#L143

michalmuskala:
Updated.

josevalim:

PR #7314, opened 2018-02-05
Clarify documentation for URI.to_string/1

### LINKED ISSUE #7313: Clarify behavior of URI.to_string/1
Hello everyone 👋 

My team has been using the `URI` module to _cleanly_ modify URIs.

A simple example is to replace the **authority** of a URI.
However, the **authority** of a URI is the combination of three components: the **userinfo**, the **host**, and the **port**.

The `URI` struct thankfully has an authority field for convenience, however, looking at the code for `URI.to_string/1`, I see that the **host** field has precedence over the **authority** field.

https://github.com/elixir-lang/elixir/blob/2b588dfb3ebb2f3221bec3509cb1dbb8e08ef1af/lib/elixir/lib/uri.ex#L587-L601

Meaning the following doesn't work:
```elixir
iex> "https://hexdocs.pm/elixir/URI.html#to_string/1" |> URI.parse() |> Map.merge(%{authority: "example.com"}) |> URI.to_string()
"https://hexdocs.pm/elixir/URI.html#to_string/1"
```

But this does:
```elixir
iex> "https://hexdocs.pm/elixir/URI.html#to_string/1" |> URI.parse() |> Map.merge(%{userinfo: nil, host: "example.com", port: nil}) |> URI.to_string()
"https://example.com/elixir/URI.html#to_string/1"
```

I think this is acceptable behavior considering the **userinfo**, the **host**, and the **port** are more granular pieces of information but is this defined behavior?

If it is, should we make this more explicit in the documentation? (I would gladly contribute)

Is it that I am completely misusing the `URI` module?

Best 🤗

maxmellen:
Closes #7313

### DIFF
--- a/lib/elixir/lib/uri.ex
+++ b/lib/elixir/lib/uri.ex
@@ -471,6 +471,21 @@ defmodule URI do
       iex> URI.to_string(%URI{scheme: "foo", host: "bar.baz"})
       "foo://bar.baz"
 
+  Note that when creating this string representation, the `authority` will be
+  used if the host is `nil`. Otherwise, the `userinfo`, `host`, and `port` will
+  be used.
+
+      iex> URI.to_string(%URI{authority: "foo@example.com:80"})
+      "//foo@example.com:80"
+
+      iex> URI.to_string(%URI{userinfo: "bar", host: "example.org", port: 81})
+      "//bar@example.org:81"
+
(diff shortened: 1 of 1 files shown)


whatyouhide:
Thanks @maxmellen! 💟

PR #7315, opened 2018-02-05
is_integer guard for `n` argument in Stream.drop/2

Hinidu:
Before this change providing non-number produces harder to understand error "bad argument in arithmetic expression" like this:
```
[1,2,3] |> Stream.drop(nil) |> Enum.to_list
    (elixir) lib/stream.ex:293: anonymous fn/3 in Stream.drop/2
    (elixir) lib/enum.ex:3161: Enumerable.List.reduce/3
    (elixir) lib/stream.ex:1433: Enumerable.Stream.do_each/4
    (elixir) lib/enum.ex:1823: Enum.reverse/1
    (elixir) lib/enum.ex:2581: Enum.to_list/1
```
And floating point numbers give results like this:
```
[1,2,3] |> Stream.drop(1.1) |> Enum.to_list
[3]
```
All other similar functions that I've checked have these guards (i.e. Enum.drop, Stream.take).

### DIFF
--- a/lib/elixir/lib/stream.ex
+++ b/lib/elixir/lib/stream.ex
@@ -323,11 +323,11 @@ defmodule Stream do
 
   """
   @spec drop(Enumerable.t(), non_neg_integer) :: Enumerable.t()
-  def drop(enum, n) when n >= 0 do
+  def drop(enum, n) when is_integer(n) and n >= 0 do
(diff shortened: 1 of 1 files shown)


lexmag:
Thank you @Hinidu.

Hinidu:
Thank you @ericmj and @lexmag!

PR #7316, opened 2018-02-05
fix DynamicSupervisor error report

voltone:
DynamicSupervisor doesn't trigger SASL reports on child restarts due to a typo in the report type. Here's Erlang's supervisor for comparison:
https://github.com/erlang/otp/blob/master/lib/stdlib/src/supervisor.erl#L1394

I tried to add a relevant test, but it got messy very quickly. I'll be happy to look into that once I have a bit more time. For now I tested the fix and I was able to get SASL reports for my DynamicSupervisor.


josevalim:
Pushed to master with a test, thank you!

PR #7320, opened 2018-02-06
Use tags on windows/unix and allow failures on appveyor

josevalim:


lib/elixir/test/elixir/system_test.exs
```diff
@@ -132,12 +134,16 @@ defmodule SystemTest do
     after
       File.rm_rf!(Path.dirname(tmp_path(@echo)))
     end
-  else
-    test "cmd/2 unix" do
+  end
+
+  describe "UNIX" do
```
whatyouhide: You used "Unix" once and "UNIX" the other time :)

josevalim:

PR #7325, opened 2018-02-07
Conform the spec for String.slice/3 to its guard clauses

njonsson:
The guard clauses of [_String.slice/3_](https://github.com/elixir-lang/elixir/blob/e9dfa50c74488000d2c9de71e926cdd78609b3ac/lib/elixir/lib/string.ex#L1739) exclude negative numbers for the _length_ argument. The function’s spec does not agree, specifying it as `integer` rather than `non_neg_integer`.

### DIFF
--- a/lib/elixir/lib/string.ex
+++ b/lib/elixir/lib/string.ex
@@ -1736,7 +1736,7 @@ defmodule String do
       ""
 
   """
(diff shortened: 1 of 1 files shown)


josevalim:

PR #7328, opened 2018-02-08
Fix - Allow to use a dot_formatter config file in a different folder which imports dependencies

### LINKED ISSUE #7327: mix format raises File.Error when using .formatter.exs with import_deps in a different folder
### Precheck

* Do not use the issues tracker for help or support (try Elixir Forum, Stack Overflow, IRC, etc.) ✅ 
* For bugs, do a quick search and make sure the bug has not yet been reported ✅ 
* Finally, be nice and have fun! ✅ 

### Environment

- Elixir 1.6.1
- Erlang OTP20
- macOS High Sierra (10.13.2)

### Current behavior

Running `mix format` using `--dot-formatter` with a `.formatter.exs` in a different folder - containing a `import_deps` configuration - results in the following error:

```
** (File.Error) could not read file "<PATH_TO_CURRENT_FOLDER>/_build/dev/.mix/cached_formatter_deps": no such file or directory
    (elixir) lib/file.ex:310: File.read!/1
    (mix) lib/mix/tasks/format.ex:193: Mix.Tasks.Format.read_deps_manifest/1
    (mix) lib/mix/tasks/format.ex:173: Mix.Tasks.Format.fetch_deps_opts/1
    (mix) lib/mix/tasks/format.ex:129: Mix.Tasks.Format.run/1
    (mix) lib/mix/task.ex:314: Mix.Task.run_task/3
    (mix) lib/mix/cli.ex:80: Mix.CLI.run_task/2
```

### Expected behavior

`mix format` should format the file(s) using the configuration from `.formatter.exs` and respect `:import_deps`.

### Research

I've tried to track the cause of the issue and it seems the root cause might be this line:

https://github.com/elixir-lang/elixir/blob/ae8502d8cde107f30c170ef510ce2177d1c442d8/lib/mix/lib/mix/tasks/format.ex#L166

It seems that `Mix.Project.manifest_path/0` always takes the current folder to expand the path. This is fairly easy to reproduce in `iex`.

```elixir
# Let's assume we're in the root of an umbrella project
iex> Mix.Project.manifest_path
"<path-to-project>/_build/<env>"
iex> cd "apps"
iex> Mix.Project.manifest_path
"<path-to-project>/apps/_build/<env>"
```

This then results in `mix format` trying to read `<path-to-project>/apps/_build/<env>/cached_formatter_deps` which obviously can not be found and raises a `File.Error`.

Zeeker:
This PR changes the `format.ex` mix task and updates the corresponding test.

`eval_dot_formatter/1` now returns a two value tuple with the first value being the directory of the evaluated `dot_formatter` file. The second value are the actual `dot_formatter` options.

It then passes this `formatter_dir` to `fetch_deps_opts` which now has an arity two implementation. This implementation receives the `formatter_opts` and the `formatter_dir` as second argument. Take a look at the header.

```elixir
defp fetch_deps_opts(formatter_opts, in_directory: nil)
```

If the directory is non nil, it wraps the call to `fetch_deps_opts/1` in a `File.cd!/2` call. This ensures that the dependencies are actually loaded from the same directory in which the `dot_formatter` resides.

Fixes #7327

### DIFF
--- a/lib/mix/lib/mix/tasks/format.ex
+++ b/lib/mix/lib/mix/tasks/format.ex
@@ -138,8 +138,11 @@ defmodule Mix.Tasks.Format do
 
   defp eval_dot_formatter(opts) do
     case dot_formatter(opts) do
-      {:ok, dot_formatter} -> {Path.dirname(dot_formatter), eval_file_with_keyword_list(dot_formatter)}
-      :error -> {nil, []}
+      {:ok, dot_formatter} ->
+        {Path.dirname(dot_formatter), eval_file_with_keyword_list(dot_formatter)}
+
+      :error ->
+        {nil, []}
     end
   end
 
--- a/lib/mix/lib/mix/tasks/format.ex
+++ b/lib/mix/lib/mix/tasks/format.ex
@@ -126,8 +126,8 @@ defmodule Mix.Tasks.Format do
 
   def run(args) do
     {opts, args} = OptionParser.parse!(args, strict: @switches)
-    formatter_opts = eval_dot_formatter(opts)
-    formatter_opts = fetch_deps_opts(formatter_opts)
+    {formatter_dir, formatter_opts} = eval_dot_formatter(opts)
+    formatter_opts = fetch_deps_opts(formatter_opts, in_directory: formatter_dir)
 
     args
     |> expand_args(formatter_opts)
@@ -138,8 +138,8 @@ defmodule Mix.Tasks.Format do
 
   defp eval_dot_formatter(opts) do
     case dot_formatter(opts) do
-      {:ok, dot_formatter} -> eval_file_with_keyword_list(dot_formatter)
-      :error -> []
+      {:ok, dot_formatter} -> {Path.dirname(dot_formatter), eval_file_with_keyword_list(dot_formatter)}
+      :error -> {nil, []}
     end
   end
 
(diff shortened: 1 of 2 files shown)


josevalim:
I don't think this is going to work. Mix was designed to always run from the root. We shouldn't `cd` into the current directory because even your mix.exs may assume that it is running at the root directory.

In other words, this is not a formatter bug, but it is a consequence that you are not running in the same directory as the mix.exs file and using Mix specific features (such as dependency handling).

I will see if we can improve the error message though.

Zeeker:
The only logic which is executed in the dot_formatter folder is loading the dependency configuration for `mix format`. Nothing more, nothing less. It's not like this changes the behaviour of `mix` in general.

josevalim:
@Zeeker i have tried your test locally and i got this error:

```
** (Mix.Error) Unavailable dependency :my_dep given to :import_deps in the formatter configuration. The dependency cannot be found in the filesystem, please run mix deps.get and try again
```

It looks like correct to me.

josevalim:
> Nothing more, nothing less. It's not like this changes the behaviour of mix in general.

It changes, because it breaks the assumption Mix is always running at the root of the project. Once you `File.cd!` only in certain situations, it will lead to more failures somewhere down the road.

Zeeker:
I can see your point but as a user I was never aware of the fact that mix is meant to be run in the root.

On the other hand format **does** provide the option to use a dot format file from another folder, and it is possible to specify configuration which references data in it's folder.

What do we make of this directory bound configuration, if not load it?

josevalim:
Well, you can move around freely. But if you are using dependencies, then you need to be in the directory that defines the mix.exs with those dependencies, otherwise we indeed won't find them.

I don't think there is much for us to do here, so I will go ahead and make sure this is clear in the docs, thanks!

Zeeker:
What do you think of another option to either specify the dependency path or to enable the behaviour from this PR?

PR #7336, opened 2018-02-10
Use ISO8601 for System.build_info :date

mobileoverlord:
The current implementation of this field uses rfc1123. This will change it to use to_iso8601 so that it can easily be parsed using built in types later.

### DIFF
--- a/lib/elixir/lib/system.ex
+++ b/lib/elixir/lib/system.ex
@@ -139,7 +139,7 @@ defmodule System do
 
   # Get the date at compilation time.
   defmacrop get_date do
-    IO.iodata_to_binary(:httpd_util.rfc1123_date())
+    DateTime.utc_now() |> DateTime.to_iso8601()
   end
 
   @doc """


josevalim:
This will cause bootstrap issues. I think it is a good idea but the
implementation would be non trivial. I will investigate.
-- 


wojtekmach:
to overcome bootstraping issues we could use these or similar (e.g. if you need microsecond precision :)) erlang functions:

```elixir
iex> {{year, month, day}, {hour, minute, second}} = :erlang.universaltime()
iex> iolist = :io_lib.format("~4..0b-~2..0b-~2..0bT~2..0b:~2..0b:~2..0bZ", [year, month, day, hour, minute, second])
iex> :erlang.iolist_to_binary(iolist)
"2018-02-11T00:13:52Z"
```

josevalim:
@wojtekmach would you please send a PR? :)

PR #7338, opened 2018-02-11
Improve documentation for ExUnit.Callbacks

febeling:
The change refines and extends the documentation for the ExUnit
callbacks for test setup, `setup` and `setup_all`.

* refer to the introductory Context section in function documentation
* name various ways to define setup code using atom naming unary
  function, list of atoms naming unary functions and block.
* clarify that multiple setup callbacks can be used
* reduce two overly similar paragraphs about return values into a
  single one


lexmag:
Lovely improvements. 💛
Thanks @febeling.

PR #7347, opened 2018-02-14
Add some additional clarification to Float documentation

michalmuskala:


lib/elixir/lib/float.ex
```diff
@@ -30,6 +32,12 @@ defmodule Float do
 
   There are also other known problems like flooring or rounding numbers. See
   `round/2` and `floor/2` for more details about them.
+
+  To learn more about floting-point arithmetic visit:
+
+    * http://0.30000000000000004.com/
```
whatyouhide: Will this be a clickable link in the docs?

michalmuskala: I checked and there's no autolinking in ex_doc. I'll correct it

PR #7354, opened 2018-02-15
Add a subsection for function types in typespecs docs

ilya-kolpakov:
Makes it easier to find the syntax used for typing anonymous functions which is not mentioned explicitly in the Getting Started guide and is impossible to find by searching for "function types".

I would update the guide as well by adding a paragraph mentioning function types and providing an example.

### DIFF
--- a/lib/elixir/pages/Typespecs.md
+++ b/lib/elixir/pages/Typespecs.md
@@ -24,7 +24,8 @@ The notation to represent the union of types is the pipe `|`. For example, the t
 
 ### Basic types
 
-    type :: any()                   # the top type, the set of all terms
+    type ::
+          any()                     # the top type, the set of all terms
           | none()                  # the bottom type, contains no terms
           | atom()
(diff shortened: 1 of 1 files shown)


josevalim:
@ilya-kolpakov I have pushed a commit to move this to literals and then I realized that we do mention those. That said, I don't think we should move them to separate section because they are literals after all. For example, should we move maps to a separate section because folks cannot find them?

If the issue is discovery, then we can find ways to improve it, but organization wise this is correct. Thank you.

ilya-kolpakov:
Discovery is indeed the issue. I spent quite some time trying to figure this out and did not find what the syntax was until I posted on Elixir forum even if I had the right page open in the browser. If I google "typespecs function types" I do not get anything pointing at the right syntax.

I see your point if this documentation is targeted for developers. I would say it is probably quite difficult for elixir devs to imagine the experience of someone quite new to the language, so please do not underestimate that gap.

Issue #5851, opened 2017-03-08
Support Erlang 20 new features


josevalim:
The following are chores that be done in order to properly support 20. Those must be backported:

 * [x] The compiler has new warnings for repeated identical map keys
 * [x] No longer use regexes in Elixir as they are not portable from 19->20
 * [x] Fix logger messages
 * [x] Check for OTP version in Mix manifest

The following changes can be done as soon as the first RC is out:

 * [x] Support UTF-8 encoding for atoms and variables
 * [x] Use the new debug_info compiler options
 * [x] Use the new extra_chunks compiler options
 * [x] Support shell history

The following tasks requires us to support at least Erlang 20:

 * [x] Trim down the Unicode modules as Erlang provides NFC/NFD conversion and deprecate functions from String
 * [x] There are two new guard BIFs `floor/1` and `ceil/1`
 * [x] No longer use try/catch in Task (this requirement was dropped from proc_lib)
 * [x] No longer use outdated rand algorithms (exs1024 -> exs1024s)
 * [x] Use map instructions in comprehensions #7032 
 * [x] Optimize maps:put and maps:merge into instructions #7353 
 * [x] Consistently use :second, :nanosecond as time units and deprecate the old ones
 * [x] Update parse transforms warning to say they are no longer applied

Please do not send PR for those issues as they cannot be tackled right now but rather long term.

josevalim:
List updated with Erlang 20 RC1 notes.

c0b:
About unicode atoms, with erlang 20.0-rc1 I found only the quoted syntax is supported:

```erl
Erlang/OTP 20 [RELEASE CANDIDATE 1] [erts-9.0] [source-9ac8bdb] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:10] [hipe] [kernel-poll:false]

Eshell V9.0  (abort with ^G)
1> 'hello_юникод_世界'.
'hello_юникод_世界'
2> hello_юникод_世界.
* 1: illegal character
```

am not sure will the unquoted unicode atom be supported? when run iex on top of
it's similar that the quoted syntax is already supported, so wonder from here will the unquoted syntax be supported?

```elixir
Erlang/OTP 20 [RELEASE CANDIDATE 1] [erts-9.0] [source-9ac8bdb] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:10] [hipe] [kernel-poll:false]

[...]

PR #7355, opened 2018-02-15
No longer use outdated rand algorithms

TattdCodeMonkey:
Erlang 20 added new rand algorithms and marked others as outdated.
Replacing the usages of `:exs1024` to `:exs1024s` and `:exsplus` to
`:exsp` to support current random algorithms.

This addresses one of the TODO items in #5851

### DIFF
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -1774,7 +1774,7 @@ defmodule Enum do
   ## Examples
 
       # Although not necessary, let's seed the random algorithm
-      iex> :rand.seed(:exsplus, {101, 102, 103})
+      iex> :rand.seed(:exsp, {101, 102, 103})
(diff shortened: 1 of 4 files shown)


mememori:
OOC, why not use `exrop` as it became the new default algo ?

whatyouhide:
Don't we need to wait for strictly depending on Erlang >= 20 for this? Alternatively, we can use the right algorithm based on the OTP version.

josevalim:
@TattdCodeMonkey as the linked issue mentions:

> Please do not send PR for those issues as they cannot be tackled right now but rather long term.

Thanks but we can't do anything for now as we still need to support earlier Erlang versions. :)

TattdCodeMonkey:
Ah, sorry @josevalim I totally missed that.

PR #7358, opened 2018-02-16
Document how to programatically run profiling tasks

michalmuskala:
I didn't change how options are handled, so they are expected to be strings. Should it be amended, so that they can be passed as actual terms?

### DIFF
--- a/lib/mix/lib/mix/tasks/profile.cprof.ex
+++ b/lib/mix/lib/mix/tasks/profile.cprof.ex
@@ -122,13 +122,15 @@ defmodule Mix.Tasks.Profile.Cprof do
   end
 
   defp profile_code(code_string, opts) do
+    opts = Enum.map(opts, &parse_opt/1)
+
     content =
       quote do
         unquote(__MODULE__).profile(
           fn ->
             unquote(Code.string_to_quoted!(code_string))
           end,
-          unquote(opts)
+          unquote(Macro.escape(opts))
         )
       end
 
@@ -136,8 +138,31 @@ defmodule Mix.Tasks.Profile.Cprof do
     Code.compile_quoted(content)
   end
 
-  @doc false
-  def profile(fun, opts) do
+  defp parse_opt({:matching, matching}) do
+    case Mix.Utils.parse_mfa(matching) do
+      {:ok, [m, f, a]} -> {:matching, {m, f, a}}
+      {:ok, [m, f]} -> {:matching, {m, f, :_}}
+      {:ok, [m]} -> {:matching, {m, :_, :_}}
+      :error -> Mix.raise("Invalid matching pattern: #{matching}")
+    end
+  end
+
+  defp parse_opt({:module, module}), do: {:module, string_to_existing_module(module)}
+  defp parse_opt(other), do: other
+
+  @doc """
+  Allows to programatically run the `cprof` profiler on expression in `fun`.
+
+  ## Options
+
+    * `:matching` - only profile calls matching the given pattern in form of
+      `{module, function, arity}`, where each element may be replaced by `:_`
+      to allow any value
+    * `:limit` - filters out any results with a call count less than the limit
+    * `:module` - filters out any results not pertaining to the given module
+
+  """
+  def profile(fun, opts \\ []) when is_function(fun, 0) do
     fun
     |> profile_and_analyse(opts)
     |> print_output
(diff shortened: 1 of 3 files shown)


lib/mix/lib/mix/tasks/profile.cprof.ex
```diff
@@ -136,8 +136,17 @@ defmodule Mix.Tasks.Profile.Cprof do
     Code.compile_quoted(content)
   end
 
-  @doc false
-  def profile(fun, opts) do
+  @doc """
+  Allows programatically running the `cprof` profiler on expression in `fun`.
```
whatyouhide: "Allows to programmatically run", does it sound better? 🤔

lib/mix/lib/mix/tasks/profile.fprof.ex
```diff
@@ -154,7 +154,16 @@ defmodule Mix.Tasks.Profile.Fprof do
...
 
-  @doc false
+  @doc """
+  Allows programatically running the `fprof` profiler on expression in `fun`.
+
+  ## Options
+
+    * `:callers` - prints detailed information about immediate callers and called functions
+    * `:details` - includes profile data for each profiled process
+    * `:sort` - sorts the output by given key: `acc` (default) or `own`
```
whatyouhide: If we go with strings, we should use quotes around `"acc"` and `"own"`.

whatyouhide:
I am not sure if we should pass the arguments as strings, but I would `@spec` these functions to make it clear, whatever we end up doing.

josevalim:
> I didn't change how options are handled, so they are expected to be strings.

Which options?

michalmuskala:
@josevalim things like `matching: "Mod.fun/arity"` instead of `matching: {mod, fun, arity}` which I think would be more natural, etc. Or `sort: "time"` instead of `sort: :time`

josevalim:
@michalmuskala I see. Let's expect those as actual data structures. And do the parsing in `run/1`.

michalmuskala:
Updated. I changed the tasks to parse all parameters before passing them to `profile/2`.

lib/mix/lib/mix/tasks/profile.cprof.ex
```diff
@@ -128,16 +128,42 @@ defmodule Mix.Tasks.Profile.Cprof do
...
+
+    * `:matching` - only profile calls matching the given pattern
+    * `:limit` - filters out any results with a call count less than the limit
+    * `:module` - filters out any results not pertaining to the given module
+
+  ## Matching pattern
+
+  Matching pattern should be in the form of `{module, function, arity}`, where
+  each element may be replaced by `:_` to allow any value.
+
```
josevalim: No blank new line here?

lib/mix/lib/mix/tasks/profile.cprof.ex
```diff
@@ -128,16 +128,42 @@ defmodule Mix.Tasks.Profile.Cprof do
...
+  defp parse_opt({:module, module}), do: {:module, string_to_existing_module(module)}
+  defp parse_opt(other), do: other
+
+  @doc """
+  Allows to programatically run the `cprof` profiler on expression in `fun`.
+
+  ## Options
+
+    * `:matching` - only profile calls matching the given pattern
```
josevalim: Let's describe the format of this tuple.

lib/mix/lib/mix/tasks/profile.cprof.ex
```diff
@@ -128,16 +128,42 @@ defmodule Mix.Tasks.Profile.Cprof do
           fn ->
             unquote(Code.string_to_quoted!(code_string))
           end,
-          unquote(opts)
+          unquote(Macro.escape(Enum.map(opts, &parse_opt/1)))
```
josevalim: Let's do the Enum.map in a previous line to avoid nested calls?

lib/mix/lib/mix/tasks/profile.eprof.ex
```diff
@@ -140,16 +140,45 @@ defmodule Mix.Tasks.Profile.Eprof do
...
+  defp parse_opt({:sort, "calls"}), do: {:sort, :calls}
+  defp parse_opt({:sort, other}), do: Mix.raise("Invalid sort option: #{other}")
+  defp parse_opt(other), do: other
+
+  @doc """
+  Allows to programatically run the `eprof` profiler on expression in `fun`.
+
+  ## Options
+
+    * `:matching` - only profile calls matching the given pattern
```
josevalim: Same here.

michalmuskala:
@josevalim updated. I have indeed forgot about this one 😅

lib/mix/lib/mix/tasks/profile.cprof.ex
```diff
@@ -153,14 +178,8 @@ defmodule Mix.Tasks.Profile.Cprof do
...
-        nil ->
-          :cprof.start()
-
-        matching ->
-          case Mix.Utils.parse_mfa(matching) do
-            {:ok, args} -> apply(:cprof, :start, args)
-            :error -> Mix.raise("Invalid matching pattern: #{matching}")
-          end
+        nil -> :cprof.start()
+        matching -> :cprof.start(matching)
```
whatyouhide: We can replace this with

```elixir
case Keyword.fetch(opts, :matching) do
  {:ok, matching} -> :cprof.start(matching)
  :error -> :cprof.start()
end
```

josevalim:

PR #7359, opened 2018-02-16
Remove unecessery releases directory check

josevalim:
This check was added when Elixir itself was used as a release
but that's not the case anymore.


josevalim:

PR #7364, opened 2018-02-19
typespec declarations should return :ok like other attributes

webdeb:
@josevalim as discussed in #7363 will close it in favour of this one.


whatyouhide:
Thanks @webdeb! 💟

webdeb:
Thank you for giving me the opportunity to contribute ❤️

PR #7365, opened 2018-02-19
Revises the docs of Mix.Tasks.Run

fxn:
Some edits I did while reading this file. The main proposed changes are:

* The opening sentence said that applications "start and run". Applications can just be "started", really. The command allows the user to "run" _additional code_ if they wish.
* I found mixing configuration with `System.argv/0` a bit confusing. I believe the docs have to state 
 simply that the _command-line arguments_ for the script or expression are available in `System.argv/0`, whatever they are.
* The `-e` example had a `start/0` call. Given that all applications are started by default, didn't find this example very realistic. The patch has a more generic function call instead.


whatyouhide:
Thanks @fxn! 💟

PR #7366, opened 2018-02-19
Deprecate dynamic mode in OptionParser and promote strict

### LINKED ISSUE #7278: Deprecate dynamic mode in OptionParser and promote strict first

uesteibar:
Fixes #7278

### DIFF
--- a/lib/elixir/lib/option_parser.ex
+++ b/lib/elixir/lib/option_parser.ex
@@ -112,6 +112,8 @@ defmodule OptionParser do
 
   ### Parsing dynamic switches
 
+  WARNING: this option is deprecated.
+
   `OptionParser` also includes a dynamic mode where it will attempt to parse
   switches dynamically. Such can be done by not specifying the `:switches` or
   `:strict` option.
@@ -543,6 +545,7 @@ defmodule OptionParser do
           {strict, true}
 
         true ->
+          IO.warn("Not passing the :switches or :strict option to OptionParser is deprecated")
           {[], false}
       end
 
--- a/lib/elixir/test/elixir/option_parser_test.exs
+++ b/lib/elixir/test/elixir/option_parser_test.exs
@@ -400,3 +400,33 @@ defmodule OptionParserTest do
     assert original == OptionParser.to_argv(opts, switches: [counter: :count])
   end
 end
+
+defmodule OptionsParserDeprecationsTest do
+  use ExUnit.Case, async: false
+
+  test "dynamic mode is deprecated" do
+    warning = ~r[Not passing the :switches or :strict option to OptionParser is deprecated]
+
+    deprecated_cases = [
+      {["--d"], []},
+      {["--d"], [allow_nonexistent_atoms: true]}
+    ]
+
+    supported_cases = [
+      {["--d", "true"], [strict: [d: :boolean]]},
+      {["--d", "true"], [switches: [d: :boolean]]}
+    ]
+
+    Enum.each(deprecated_cases, fn {case, options} ->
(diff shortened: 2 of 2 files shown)


lib/elixir/lib/option_parser.ex
```diff
@@ -543,6 +545,10 @@ defmodule OptionParser do
           {strict, true}
 
         true ->
+          IO.warn(
+            "Dynamic mode in OptionParser is deprecated, please consider using `strict: [...]`"
```
josevalim: I think a better warning would be:

    "Not passing the :switches or :strict option to OptionParser is deprecated".

lib/elixir/test/elixir/option_parser_test.exs
```diff
@@ -400,3 +400,33 @@ defmodule OptionParserTest do
     assert original == OptionParser.to_argv(opts, switches: [counter: :count])
   end
 end
+
+defmodule OptionsParserDeprecationsTest do
```
josevalim: There are a bunch of warnings now in the OptionParser suite, we need to investigate those too.

lib/elixir/lib/option_parser.ex
```diff
@@ -112,6 +112,8 @@ defmodule OptionParser do
 
   ### Parsing dynamic switches
 
+  WARNING: this option is deprecated.
+
```
AndrewDryga: Maybe if the option is deprecated we should delete it from the doc?

lib/elixir/test/elixir/option_parser_test.exs
uesteibar: I'll check them and use either `switch` or `strict` as necessary

josevalim: One suggestion is to move them all to this module and wrap them with capture_io(:stderr).

josevalim:
Please take a look at the build. It is failing because of unformatted files.

lib/elixir/test/elixir/option_parser_test.exs
josevalim: We already have them in other places in the suite, I think we should rather move them here. :)

uesteibar:
@josevalim addressed your feedback

lib/elixir/lib/option_parser.ex
uesteibar: I saw it's done like that for `Behaviour`, should we just remove all this @josevalim?

lib/elixir/lib/option_parser.ex
```diff
@@ -543,6 +545,8 @@ defmodule OptionParser do
           {strict, true}
 
         true ->
+          IO.warn("Not passing the :switches or :strict option to OptionParser is deprecated")
+
```
whatyouhide: Nitpick: I think we can remove this line.

lib/elixir/test/elixir/option_parser_test.exs
```diff
@@ -400,3 +288,188 @@ defmodule OptionParserTest do
...
+  use ExUnit.Case, async: false
+
+  @warning ~r[Not passing the :switches or :strict option to OptionParser is deprecated]
+
+  defmacro assert_deprecated(clause) do
```
whatyouhide: I am not convinced we need a macro here. Could we get away with:

```elixir
defp assert_deprecated(fun) do
  assert ExUnit.CaptureIO.capture_io(:stderr, fun) =~ @warning
end
```

and

```elixir
assert_deprecated(fn ->
  assert OptionParser.parse(["--docs"]) == {[docs: true}, [], []}
end)
```

?

josevalim:

PR #7371, opened 2018-02-19
Fix DynamicSupervisor handling of extra arguments on child restart (fixes #7369)

### LINKED ISSUE #7369: DynamicSupervisor with extra_arguments crashes when restarting child
### Environment

* Elixir & Erlang versions (elixir --version): Elixir 1.6.1 (compiled with OTP 20)
* Operating system: macOS 10.13.2

### Current behavior

When using a DynamicSupervisor with `:extra_arguments`, a crash in one child kills all other children and restarts the supervisor.

Here is a slightly shortened example iex session to reproduce the behaviour, the corresponding code will be given below:

```
$ iex -S mix
iex(1)> Application.ensure_all_started(:sasl)
iex(2)> Demo.Supervisor.start_child(:foo, :bar, :baz)
iex(3)> Demo.Supervisor.start_child(:foo, :bar, :baz)
iex(4)> {:ok, pid} = Demo.Supervisor.start_child(:foo, :bar, :baz)
iex(5)> Supervisor.which_children(Demo.Application.Supervisor)
[{Demo.Supervisor, #PID<0.116.0>, :supervisor, [Demo.Supervisor]}]
iex(6)> Supervisor.which_children(Demo.Supervisor)
[
  {:undefined, #PID<0.128.0>, :worker, [Demo.Worker]},
  {:undefined, #PID<0.130.0>, :worker, [Demo.Worker]},
  {:undefined, #PID<0.132.0>, :worker, [Demo.Worker]}
]
iex(7)> Process.exit(pid, :kill)

=SUPERVISOR REPORT==== 19-Feb-2018::20:06:49 ===
     Supervisor: {local,'Elixir.Demo.Application.Supervisor'}
     Context:    child_terminated
     Reason:     shutdown
     Offender:   [{pid,<0.116.0>},
                  {id,'Elixir.Demo.Supervisor'},
                  {mfargs,
                      {'Elixir.Demo.Supervisor',start_link,[implicit_arg]}},
                  {restart_type,permanent},
                  {shutdown,infinity},
                  {child_type,supervisor}]
(...)
iex(8)> Supervisor.which_children(Demo.Application.Supervisor)
[{Demo.Supervisor, #PID<0.137.0>, :supervisor, [Demo.Supervisor]}]
iex(9)> Supervisor.which_children(Demo.Supervisor)
[]
iex(10)>
```

As you can see, killing one child crashes the Dynamic Supervisor and with it all of its children.

### Expected behavior

The Supervisor should restart the child without crashing.

### Example Code

The example code can also be found [here](https://github.com/jvf/dyn_sup_crash) as a ready-to-run example project. I tried to stick as close very closely to the example from the [DynamicSupervisor documentation](https://hexdocs.pm/elixir/DynamicSupervisor.html#module-migrating-from-supervisor-s-simple_one_for_one).

```elixir
defmodule Demo.Application do
  use Application

  def start(_type, _args) do
    children = [{Demo.Supervisor, :implicit_arg}]
    opts = [strategy: :one_for_one, name: __MODULE__.Supervisor]
    Supervisor.start_link(children, opts)
  end
end

defmodule Demo.Supervisor do
  use DynamicSupervisor

  def start_link(arg) do
    DynamicSupervisor.start_link(__MODULE__, arg, name: __MODULE__)
  end

  def start_child(foo, bar, baz) do
    # This will start child by calling Worker.start_link(implicit_arg, foo, bar, baz)
    spec = Supervisor.Spec.worker(Demo.Worker, [foo, bar, baz])
    DynamicSupervisor.start_child(__MODULE__, spec)
  end

  def init(implicit_arg) do
    DynamicSupervisor.init(
      strategy: :one_for_one,
      extra_arguments: [implicit_arg]
    )
  end
end

defmodule Demo.Worker do
  use GenServer

  require Logger

  def start_link(:implicit_arg, :foo, :bar, :baz) do
    GenServer.start_link(__MODULE__, [:implicit_arg, :foo, :bar, :baz])
  end

  def init(:implicit_arg, :foo, :bar, :baz) do
    {:ok, []}
  end

  def handle_call(request, from, state) do
    Logger.debug(fn -> "#{request} from #{from}" end)
    {:reply, :ok, state}
  end
end
```

jvf:
Fixes #7369 where a DynamicSupervisor using `extra_arguments` crashed when children were restarted.

### DIFF
--- a/lib/elixir/lib/dynamic_supervisor.ex
+++ b/lib/elixir/lib/dynamic_supervisor.ex
@@ -942,8 +942,9 @@ defmodule DynamicSupervisor do
 
   defp restart_child(:one_for_one, current_pid, child, state) do
     {{m, f, args} = mfa, restart, shutdown, type, modules} = child
+    %{extra_arguments: extra} = state
 
-    case start_child(m, f, args) do
+    case start_child(m, f, extra ++ args) do
       {:ok, pid, _} ->
         state = delete_child(current_pid, state)
         {:ok, save_child(pid, mfa, restart, shutdown, type, modules, state)}
--- a/lib/elixir/test/elixir/dynamic_supervisor_test.exs
+++ b/lib/elixir/test/elixir/dynamic_supervisor_test.exs
@@ -356,6 +356,27 @@ defmodule DynamicSupervisorTest do
       assert {:error, :max_children} = DynamicSupervisor.start_child(pid, child)
     end
 
+    # regression test for #7369
+    test "restarting a child with extra_args does not crash the supervisor" do
+      parent = self()
+
+      fun = fn ->
+        send(parent, :from_child)
+        :timer.sleep(:infinity)
+      end
+
+      {:ok, sup} = DynamicSupervisor.start_link(strategy: :one_for_one, extra_arguments: [fun])
+      child = %{id: Task, restart: :transient, start: {Task, :start_link, []}}
+
+      assert {:ok, child} = DynamicSupervisor.start_child(sup, child)
+      assert is_pid(child)
+      assert_receive :from_child
+      assert %{active: 1, workers: 1} = DynamicSupervisor.count_children(sup)
+      assert_kill(child, :oops)
+      assert_receive :from_child
+      assert %{workers: 1, active: 1} = DynamicSupervisor.count_children(sup)
+    end
+
     test "child is restarted when trying again" do
       child = current_module_worker([:try_again, self()], restart: :permanent)
       {:ok, pid} = DynamicSupervisor.start_link(strategy: :one_for_one, max_restarts: 2)
--- a/lib/elixir/test/elixir/dynamic_supervisor_test.exs
+++ b/lib/elixir/test/elixir/dynamic_supervisor_test.exs
@@ -357,7 +357,7 @@ defmodule DynamicSupervisorTest do
     end
 
     # regression test for #7369
-    test "restarting a child with extra_args does not crash the supervisor" do
+    test "restarting a child with extra_args successfully restarts child" do
       parent = self()
 
       fun = fn ->
--- a/lib/elixir/test/elixir/dynamic_supervisor_test.exs
+++ b/lib/elixir/test/elixir/dynamic_supervisor_test.exs
@@ -356,7 +356,6 @@ defmodule DynamicSupervisorTest do
       assert {:error, :max_children} = DynamicSupervisor.start_child(pid, child)
     end
 
-    # regression test for #7369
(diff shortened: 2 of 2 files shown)


lib/elixir/test/elixir/dynamic_supervisor_test.exs
```diff
@@ -356,6 +356,27 @@ defmodule DynamicSupervisorTest do
       assert {:error, :max_children} = DynamicSupervisor.start_child(pid, child)
     end
 
+    # regression test for #7369
+    test "restarting a child with extra_args does not crash the supervisor" do
```
fishcakez: Could we rename this test as `restarting a child with extra_args successfully restarts child`. The supervisor was not crashing, rather it would repeatedly try to restart a child with the wrong arity until it reached max intensity. The current name is a little confusing without the context of the issue.

lib/elixir/lib/dynamic_supervisor.ex
```diff
@@ -942,8 +942,9 @@ defmodule DynamicSupervisor do
 
   defp restart_child(:one_for_one, current_pid, child, state) do
     {{m, f, args} = mfa, restart, shutdown, type, modules} = child
+    %{extra_arguments: extra} = state
 
-    case start_child(m, f, args) do
+    case start_child(m, f, extra ++ args) do
```
fishcakez: If we generate a `:start_error` error report the child will not include the extra arguments in the report. We need to add those arguments in the report. In fact I think we have this bug ever time we report an error! As the extra args are always the same I think we can fetch it from the state in the reporting function, rather than than edit the child in the report.

fishcakez: This can be done in another PR :D

lib/elixir/test/elixir/dynamic_supervisor_test.exs
jvf: Done.

lib/elixir/lib/dynamic_supervisor.ex
jvf: I looked into it and wrote a fix (but no test yet) for the error reporting as well. I can push it here or create another PR, whatever you prefer.

josevalim: Please do send a new PR as I am merging this one. You can add a test in the translator_test file. Here is the last time we added a DynamicSupervisor test to it: https://github.com/elixir-lang/elixir/commit/f0ad7edf411de74fb0d2cf4384580a2e0672b01f

josevalim:

PR #7377, opened 2018-02-20
Fix DynamicSupervisor error report when restarting child

jvf:
As discussed in #7371, this PR Includes the `extra_arguments` in the error logging when restarting the child of a DynamicSupervisor.


fishcakez:
Thank you :heart:

PR #7383, opened 2018-02-22
Fix loading of DynamicSupervisor.init/1 docs

uesteibar:
Moving the `init/1` callback before the actual `init/1` function seems to fix the issue at #7382 for this specific case, but we think there's a bigger issue going on here.

We tried to track this down, and we got to the following:

- with the current code on master:
    ```elixir
      Code.get_docs(DynamicSupervisor, :docs) #=> [
        ...
        {{:init, 1}, 449, :def, [{:options, [], nil}], false},
        ...
      ]
    ```

- after changing the order of the functions:
    ```elixir
      Code.get_docs(DynamicSupervisor, :docs) #=> [
        ...
        {{:init, 1}, 452, :def, [{:options, [], nil}], "Receives a set of options that initializes a dynamic supervisor.\n\nThis is typically invoked at the end of the `c:init/1` callback of\nmodule-based supervisors. See the sections \"Module-based supervisors\"\nin the module documentation for more information.\n\nThe options received by this function are also supported by `start_link/2`.\n\nThis function returns a tuple containing the supervisor options.\n\n## Examples\n\n    def init(_arg) do\n      DynamicSupervisor.init(max_children: 1000, strategy: :one_for_one)\n    end\n\n## Options\n\n  * `:strategy` - the restart strategy option. The only supported\n    value is `:one_for_one` which means that no other child is\n    terminate if a child process terminates. You can learn more\n    about strategies in the `Supervisor` module docs.\n\n  * `:max_restarts` - the maximum number of restarts allowed in\n    a time frame. Defaults to `3`.\n\n  * `:max_seconds` - the time frame in which `:max_restarts` applies.\n    Defaults to `5`.\n\n  * `:max_children` - the maximum amount of children to be running\n    under this supervisor at the same time. When `:max_children` is\n    exceeded, `start_child/2` returns `{:error, :dynamic}`. Defaults\n    to `:infinity`.\n\n  * `:extra_arguments` - arguments that are prepended to the arguments\n    specified in the child spec given to `start_child/2`. Defaults to\n    an empty list.\n\n"},
        ...
      ]
    ```
    The weird part here is that in [dynamic_supervisor.ex#L452](https://github.com/uesteibar/elixir/blob/7382-docs-for-dynamic-supervisor/lib/elixir/lib/dynamic_supervisor.ex#L452) we find the *callback*, not the function that the documentation belongs to.

We also tracked down this to [module.ex#L1665](https://github.com/elixir-lang/elixir/blob/master/lib/elixir/lib/module.ex#L1665) and we checked it was being added to the correct *ets* table in both cases.

We got lost at this point, does this clarify anything about the issue? If someone could point us in the right direction here we'd love to tackle it :)

Thanks!

### DIFF
--- a/lib/elixir/lib/module.ex
+++ b/lib/elixir/lib/module.ex
@@ -1210,17 +1210,8 @@ defmodule Module do
     if kind in [:defp, :defmacrop, :typep] do
       if doc, do: {:error, :private_doc}, else: :ok
     else
-      compile_doc(
-        data_table_for(module),
-        line,
-        kind,
-        function_tuple,
-        signature,
-        doc,
-        __ENV__,
-        false
-      )
-
+      table = data_table_for(module)
+      compile_doc(table, line, kind, function_tuple, signature, doc, __ENV__, false)
       :ok
     end
   end
--- a/lib/elixir/lib/module.ex
+++ b/lib/elixir/lib/module.ex
@@ -1230,12 +1230,14 @@ defmodule Module do
 
     # TODO: Store @since and @deprecated alongside the docs
     {line, doc} = get_doc_info(table, env, impl)
-    compile_doc(table, line, kind, pair, args, doc, env)
+    compile_doc(table, line, kind, pair, args, doc, env, impl)
 
     :ok
   end
 
-  defp compile_doc(_table, line, kind, {name, arity}, _args, doc, env)
+  defp compile_doc(table, line, kind, pair, args, doc, env, impl \\ false)
+
+  defp compile_doc(_table, line, kind, {name, arity}, _args, doc, env, _impl)
        when kind in [:defp, :defmacrop] do
     if doc do
       error_message =
@@ -1246,7 +1248,7 @@ defmodule Module do
     end
   end
 
-  defp compile_doc(table, line, kind, pair, args, doc, env) do
+  defp compile_doc(table, line, kind, pair, args, doc, env, impl) do
     signature = build_signature(args, env)
 
     case :ets.lookup(table, {:doc, pair}) do
@@ -1255,7 +1257,7 @@ defmodule Module do
 
       [{doc_tuple, line, _current_kind, current_sign, current_doc}] ->
         signature = merge_signatures(current_sign, signature, 1)
-        doc = if is_nil(doc), do: current_doc, else: doc
+        doc = if is_nil(doc) || impl, do: current_doc, else: doc
         :ets.insert(table, {doc_tuple, line, kind, signature, doc})
     end
   end
--- a/lib/elixir/lib/module.ex
+++ b/lib/elixir/lib/module.ex
@@ -1210,7 +1210,17 @@ defmodule Module do
     if kind in [:defp, :defmacrop, :typep] do
       if doc, do: {:error, :private_doc}, else: :ok
     else
-      compile_doc(data_table_for(module), line, kind, function_tuple, signature, doc, __ENV__, false)
+      compile_doc(
+        data_table_for(module),
+        line,
+        kind,
+        function_tuple,
+        signature,
+        doc,
+        __ENV__,
+        false
+      )
+
       :ok
     end
   end
--- a/lib/elixir/lib/module.ex
+++ b/lib/elixir/lib/module.ex
@@ -1229,7 +1229,7 @@ defmodule Module do
     _since = compile_since(table)
 
     # TODO: Store @since and @deprecated alongside the docs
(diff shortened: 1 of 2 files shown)


josevalim:
Oh, for `@impl true` we see if the function was doc'ed but we don't check previous clauses. I think the fix is unfortunately elsewhere.

uesteibar:
Sure, we figured this would not be the proper fix, but we wanted to share the findings in case it could help. Should we close this one? or could you point us in the right direction?

josevalim:
@uesteibar I think the issue is [here](https://github.com/elixir-lang/elixir/blob/master/lib/elixir/lib/module.ex#L1231-L1233).

Note that we consider the doc to be false if there is no documentation for the current clause and if impl is set to true. However, we should consider the previous clause, so I am thinking we should move the check to those two clauses:

https://github.com/elixir-lang/elixir/blob/master/lib/elixir/lib/module.ex#L1252-L1260

At the last moment, if doc is nil and an implementation was given, then we set it to false, otherwise we keep it as is.

We would some unit tests around this, you can look at kernel/docs_test.exs.

uesteibar:
@josevalim followed your proposal, added a test case and tested on iex for `h(DynamicSupervisor.init)`. Seems fixed :)

josevalim:
@uesteibar however, we do want the doc to be set to false if `@impl true` and no documentation was written for it. It seems this test is missing from our suite.

Can you please also add a test for this scenario? Please add a new test case instead of changing exist ones. :) Thank you!

uesteibar:
Will do tomorrow, thanks for the support!

uesteibar:
Will do tomorrow, thanks for the support!

uesteibar:
@josevalim added the test case and changed some code to support it

lib/elixir/lib/module.ex
```diff
@@ -1230,12 +1230,14 @@ defmodule Module do
 
     # TODO: Store @since and @deprecated alongside the docs
     {line, doc} = get_doc_info(table, env, impl)
```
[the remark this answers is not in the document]
josevalim: Actually, I see what you did. :) I would prefer to remove the `impl` from `get_doc_info` and change both clauses in `compile_doc`, this way we have the logic in a single place and not in two different functions. Thoughts?

lib/elixir/test/elixir/kernel/docs_test.exs
```diff
@@ -227,4 +227,36 @@ defmodule Kernel.DocsTest do
              ] = docs[:callback_docs]
     end
   end
+
+  test "@impl true doesn't set @doc false if previous implementation has docs" do
```
josevalim: Beautiful! :heart:

uesteibar:
@josevalim addressed the feedback 😃

lib/elixir/lib/module.ex
```diff
@@ -1246,16 +1256,17 @@ defmodule Module do
     end
   end
 
-  defp compile_doc(table, line, kind, pair, args, doc, env) do
+  defp compile_doc(table, line, kind, pair, args, doc, env, impl) do
     signature = build_signature(args, env)
 
     case :ets.lookup(table, {:doc, pair}) do
       [] ->
+        doc = if impl, do: false, else: doc
```
josevalim: What happens if you do set `impl` true and you do set a doc though? The doc should still be there. Maybe we need a test for this too?

josevalim: I think we need something like:

```elixir
doc = if is_nil(doc) && impl, do: false, else: doc
```

lib/elixir/lib/module.ex
```diff
@@ -1246,16 +1256,17 @@ defmodule Module do
...
         :ets.insert(table, {{:doc, pair}, line, kind, signature, doc})
 
       [{doc_tuple, line, _current_kind, current_sign, current_doc}] ->
         signature = merge_signatures(current_sign, signature, 1)
-        doc = if is_nil(doc), do: current_doc, else: doc
+        doc = if is_nil(doc) || impl, do: current_doc, else: doc
```
josevalim: Same here. If there is a doc and impl then we keep the `doc`. I think we need to keep this line as is and then add the same line I suggested above to the line below:

```elixir
doc = if is_nil(doc) && impl, do: false, else: doc
```

uesteibar:
@josevalim fixed

josevalim:

uesteibar:
thanks for the support here @josevalim! ❤️

Issue #7381, opened 2018-02-21
h helper not work properly with imported functions


Eiji7:
### Environment

Both `Erlang` and `Elixir` are installed using [asdf](https://github.com/asdf-vm/asdf) version manager.

```
$ asdf current
elixir         1.6.1-otp-20
erlang         20.2.3
…
```
Funtoo Linux current x86-64bit on intel64-ivybridge

### Current behavior

```elixir
h Enum.take
# displays documentation
import Enum
# returns: Enum
h take
# displays error: No documentation for Kernel.take was found
```

### Expected behavior

I'm not sure if we even really need `h` helper working for imported functions. I just noticed this bug when trying to show documentation for `where/3` for module `Ecto.Query` (in documentation examples we are importing this module). The real problem is that `h` helper looks for this function in wrong module i.e. `Kernel`.

josevalim:
Yes, I believe we don't consider imports altogether in autocomplete but we can start doing so since autocomplete has enough context now.

dmlittle:
@josevalim I tried giving this a try. I think I might have the right approach but I want to confirm before I write tests and put up a PR. How does [this](https://github.com/dmlittle/elixir/commit/44d9c66e259c822ee97d2809f458285b8e976a97) look? 

As a side note, is it normal that `__ENV__` is set to whatever the `__ENV__` is when the module is created/loaded and not updated over time? I guess it makes sense but at the same time can't the environment change over time? I originally tried replacing `@default_files` in `IEx.Introspection.find_decompose_fun/1` with `__ENV__.requires` but those values were not updated when a new module was imported in the IEx shell. Because of this I resorted to passing through the `__CALLER__` enviroment

josevalim:
@dmlittle you should have the `__CALLER__` outside of the quote. Also we don't want the list of requires but rather the list of `functions` and `macros`. So I would recommend something like this:

  defmacro open(term) do
    quote do
      IEx.Introspection.open(unquote(IEx.Introspection.decompose(term, __CALLER__)))
    end
  end

And in decompose you look into functions and macros.

What do you think?

dmlittle:
@josevalim sounds good, I'll put up a PR!

PR #7391, opened 2018-02-24
pass through __CALLER__ context into IEx.Introspection.decompose

dmlittle:
As noted in #7381, helper methods currently don't take into account required functions. Since you're able to call the functions directly if the module has been required, ideally you should also able to invoke helpers on functions of required modules. This PR addresses this issue.

The following should now work:
```elixir
Erlang/OTP 20 [erts-9.2] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:10] [hipe] [kernel-poll:false] [dtrace]

Interactive Elixir (1.7.0-dev) - press Ctrl+C to exit (type h() ENTER for help)
iex(1)> h take
No documentation for Kernel.take was found
iex(2)> import Enum
Enum
iex(3)> h take

                          def take(enumerable, amount)

    @spec take(t(), integer()) :: list()

Takes the first amount items from the enumerable.

If a negative amount is given, the amount of last values will be taken. The
enumerable will be enumerated once to retrieve the proper index and the
remaining calculation is performed from the end.

## Examples

    iex> Enum.take([1, 2, 3], 2)
    [1, 2]

    iex> Enum.take([1, 2, 3], 10)
    [1, 2, 3]

    iex> Enum.take([1, 2, 3], 0)
    []

    iex> Enum.take([1, 2, 3], -1)
    [3]
```

I'm not sure on what the best way to write tests for this is. Any guidance would be great!

Solves #7381

### DIFF
--- a/lib/iex/lib/iex/introspection.ex
+++ b/lib/iex/lib/iex/introspection.ex
@@ -11,9 +11,8 @@ defmodule IEx.Introspection do
   Decomposes an introspection call into `{mod, fun, arity}`,
   `{mod, fun}` or `mod`.
   """
-  @default_modules [IEx.Helpers, Kernel, Kernel.SpecialForms]
 
-  def decompose({:/, _, [call, arity]} = term) do
+  def decompose({:/, _, [call, arity]} = term, context) do
     case Macro.decompose_call(call) do
       {_mod, :__info__, []} when arity == 1 ->
         {:{}, [], [Module, :__info__, 1]}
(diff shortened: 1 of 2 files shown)


lib/iex/lib/iex/introspection.ex
```diff
@@ -38,25 +37,28 @@ defmodule IEx.Introspection do
...
     end
   end
 
-  defp find_decompose_fun(fun) do
-    Enum.find(@default_modules, Kernel, fn mod ->
-      Keyword.has_key?(mod.__info__(:functions), fun) or
-        Keyword.has_key?(mod.__info__(:macros), fun)
+  defp find_decompose_fun(fun, context) do
+    context.functions
+    |> Keyword.merge(context.macros, fn _k, v1, v2 -> v1 ++ v2 end)
```
josevalim: I think this broke special forms because they are not really imported, so they are never listed in `__CALLER__`.

Also, the merge would make this unnecessarily expensive. Maybe we could do something like this:

    find_import(fun, context.functions) || find_import(fun, context.macros) || find_special_form(fun) || Kernel

And something similar for the function below.

dmlittle:
@josevalim nice catch! I updated the PR to address the issues mentioned.

josevalim:

PR #7396, opened 2018-02-26
Add cursor movement to IO.ANSI

mveytsman:
I added the cursor movement escape codes (from https://www.tldp.org/HOWTO/Bash-Prompt-HOWTO/x361.html) to `IO.ANSI`

### DIFF
--- a/lib/elixir/lib/io/ansi.ex
+++ b/lib/elixir/lib/io/ansi.ex
@@ -174,26 +174,35 @@ defmodule IO.ANSI do
   Sends cursor to the absolute position specified by `lines`, `columns`, 
   with 0,0 being the top left corner.
   """
+  @spec cursor(integer, integer) :: String.t()
   def cursor(line, column) when line >= 0 and column >= 0, do: "\e[#{line};#{column}H"
 
   @doc "Sends cursor `lines` up (defaults to 1)."
-  @spec cursor_up(integer) :: String.t()
+  @spec cursor_up :: String.t()
   def cursor_up(), do: "\e[A"
+
+  @spec cursor_up(integer) :: String.t()
   def cursor_up(lines) when lines > 0, do: "\e[#{lines}A"
 
   @doc "Sends cursor `lines` down (defaults to 1)."
-  @spec cursor_down(integer) :: String.t()
+  @spec cursor_down :: String.t()
   def cursor_down(), do: "\e[B"
+
+  @spec cursor_down(integer) :: String.t()
   def cursor_down(lines) when lines > 0, do: "\e[#{lines}B"
 
   @doc "Sends cursor `columns` to the left (defaults to 1)."
-  @spec cursor_left(integer) :: String.t()
+  @spec cursor_left :: String.t()
   def cursor_left(), do: "\e[C"
+
+  @spec cursor_left(integer) :: String.t()
   def cursor_left(columns) when columns > 0, do: "\e[#{columns}C"
 
   @doc "Sends cursor `columns` to the right (defaults to 1)."
-  @spec cursor_right(integer) :: String.t()
+  @spec cursor_right :: String.t()
   def cursor_right(), do: "\e[D"
+
+  @spec cursor_right(integer) :: String.t()
   def cursor_right(columns) when columns > 0, do: "\e[#{columns}D"
 
   @doc "Clears screen."
--- a/lib/elixir/lib/io/ansi.ex
+++ b/lib/elixir/lib/io/ansi.ex
@@ -171,27 +171,30 @@ defmodule IO.ANSI do
   defsequence(:home, "", "H")
 
   @doc """
-  Sends cursor to the absolute position specified by `lines`, `columns`, 
-  with 0,0 being the top left corner.
+  Sends cursor to the absolute position specified by `line` and `column`.
+
+  Line `0` and column `0` would mean the top left corner.
   """
   @spec cursor(integer, integer) :: String.t()
-  def cursor(line, column) when line >= 0 and column >= 0, do: "\e[#{line};#{column}H"
+  def cursor(line, column)
+      when is_integer(line) and line >= 0 and is_integer(column) and column >= 0,
+      do: "\e[#{line};#{column}H"
 
-  @doc "Sends cursor `lines` up. Defaults to 1"
+  @doc "Sends cursor `lines` up."
   @spec cursor_up(integer) :: String.t()
-  def cursor_up(lines \\ 1) when lines > 0, do: "\e[#{lines}A"
+  def cursor_up(lines \\ 1) when is_integer(lines) and lines > 0, do: "\e[#{lines}A"
 
-  @doc "Sends cursor `lines` down. Defaults to 1"
+  @doc "Sends cursor `lines` down."
   @spec cursor_down(integer) :: String.t()
-  def cursor_down(lines \\ 1) when lines > 0, do: "\e[#{lines}B"
+  def cursor_down(lines \\ 1) when is_integer(lines) and lines > 0, do: "\e[#{lines}B"
 
-  @doc "Sends cursor `columns` to the left. Defaults to 1."
+  @doc "Sends cursor `columns` to the left."
   @spec cursor_left(integer) :: String.t()
-  def cursor_left(columns \\ 1) when columns > 0, do: "\e[#{columns}C"
+  def cursor_left(columns \\ 1) when is_integer(columns) and columns > 0, do: "\e[#{columns}C"
 
-  @doc "Sends cursor `columns` to the right. Defaults to 1."
+  @doc "Sends cursor `columns` to the right."
   @spec cursor_right(integer) :: String.t()
-  def cursor_right(columns \\ 1) when columns > 0, do: "\e[#{columns}D"
+  def cursor_right(columns \\ 1) when is_integer(columns) and columns > 0, do: "\e[#{columns}D"
 
   @doc "Clears screen."
   defsequence(:clear, "2", "J")
--- a/lib/elixir/lib/io/ansi.ex
+++ b/lib/elixir/lib/io/ansi.ex
@@ -177,8 +177,9 @@ defmodule IO.ANSI do
   """
   @spec cursor(integer, integer) :: String.t()
   def cursor(line, column)
-      when is_integer(line) and line >= 0 and is_integer(column) and column >= 0,
-      do: "\e[#{line};#{column}H"
+      when is_integer(line) and line >= 0 and is_integer(column) and column >= 0 do
+    "\e[#{line};#{column}H"
+  end
 
   @doc "Sends cursor `lines` up."
   @spec cursor_up(integer) :: String.t()
--- a/lib/elixir/lib/io/ansi.ex
+++ b/lib/elixir/lib/io/ansi.ex
@@ -175,27 +175,27 @@ defmodule IO.ANSI do
 
   Line `0` and column `0` would mean the top left corner.
   """
-  @spec cursor(integer, integer) :: String.t()
+  @spec cursor(non_neg_integer, non_neg_integer) :: String.t()
   def cursor(line, column)
       when is_integer(line) and line >= 0 and is_integer(column) and column >= 0 do
     "\e[#{line};#{column}H"
   end
 
   @doc "Sends cursor `lines` up."
-  @spec cursor_up(integer) :: String.t()
-  def cursor_up(lines \\ 1) when is_integer(lines) and lines > 0, do: "\e[#{lines}A"
+  @spec cursor_up(pos_integer) :: String.t()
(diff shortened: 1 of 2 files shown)


eksperimental:
Or use left and right instead?

eksperimental:
Or backward / forward

mveytsman:
Good catch @eksperimental.

According to http://ascii-table.com/ansi-escape-sequences-vt-100.php what I called `cursor_forward`  is `CUF` and what I called `cursor_back` (oops!) is `CUB`. 

In [ECMA-48](http://www.ecma-international.org/publications/files/ECMA-ST/Ecma-048.pdf) `CUF` is called "CURSOR RIGHT" and `CUB` is "CURSOR LEFT", so calling the functions `cursor_left` and `cursor_right` would be more consistent

eksperimental:
Then I'm all for up, down, right, left
Thank you

eksperimental:
cursor uses an absolute or a relative position? I think it should be documented

fertapric:
What do you think about using `1` as default value for all the functions?

```elixir
IO.ANSI.cursor_up()
IO.ANSI.cursor_down()
IO.ANSI.cursor_left()
IO.ANSI.cursor_right()

IO.ANSI.cursor_up(2)
IO.ANSI.cursor_down(3)
IO.ANSI.cursor_left(4)
IO.ANSI.cursor_right(5)
```

eksperimental:
@fertapric it makes total sense

mveytsman:
ANSI will actually default to 1 for us if we don't pass an argument, i.e. "\e[A" moves 1 line up. I think that's the correct behavior with no argument for us as well.

mveytsman:
If you send the ANSI code for moving the cursor absolutely without any arguments it defaults to 0,0, but we already have this in `IO.ANSI.home/0` so I think it's okay to have `IO.ANSI.cursor/2` require 2 arguments.

lib/elixir/lib/io/ansi.ex
```diff
@@ -170,6 +170,30 @@ defmodule IO.ANSI do
   @doc "Sends cursor home."
   defsequence(:home, "", "H")
 
+  @doc "Sends cursor to the absolute position specified by `lines`, `columns`, with 0,0 being the top left corner."
```
fertapric: I would split this line:

```elixir
@doc """
Sends cursor to the absolute position specified by `lines`, `columns`, 
with 0,0 being the top left corner.
"""
```

fertapric: Btw, it seems the formatter leaves that `@doc` tag as is. Should the formatter try to convert `@doc "..."` into `@doc """ (new line) ... """`? Or at least emit a message?

lib/elixir/test/elixir/io/ansi_test.exs
```diff
@@ -155,4 +155,53 @@ defmodule IO.ANSITest do
...
+    assert_raise FunctionClauseError, fn ->
+      IO.ANSI.cursor(-1, 5)
+    end
+
+     assert_raise FunctionClauseError, fn ->
+      IO.ANSI.cursor(5, -1)
+    end
+  end
+
+  test "cursor_up/1" do
```
fertapric: This is also testing `cursor_up/0` :)

mveytsman: 😁 so it is

lib/elixir/lib/io/ansi.ex
```diff
@@ -170,6 +170,32 @@ defmodule IO.ANSI do
...
 
+  @doc """
+  Sends cursor to the absolute position specified by `lines`, `columns`, 
+  with 0,0 being the top left corner.
+  """
+  def cursor(line, column) when line >= 0 and column >= 0, do: "\e[#{line};#{column}H"
+
+  @doc "Sends cursor `lines` up (defaults to 1)."
+  @spec cursor_up(integer) :: String.t()
+  def cursor_up(), do: "\e[A"
```
fertapric: I think the `@spec` will be assigned to `cursor_up/0` and the specification won't be accurate. I see two options:

```elixir
# Option 1
@spec cursor_up :: String.t()
def cursor_up, do: "\e[A"

@spec cursor_up(integer) :: String.t()
def cursor_up(lines) when lines > 0, do: "\e[#{lines}A"

# Option 2
@spec cursor_up(integer) :: String.t()
def cursor_up(lines \\ 1)
def cursor_up(1), do: "\e[A"
def cursor_up(lines) when lines > 0, do: "\e[#{lines}A"
```

I would go for "Option 2" :)

mveytsman: I feel like option 1 is more faithful to the underlying implementation which might matter in the future since terminal emulators can behave in strange ways.

lib/elixir/lib/io/ansi.ex
```diff
@@ -170,6 +170,41 @@ defmodule IO.ANSI do
...
+  """
+  @spec cursor(integer, integer) :: String.t()
+  def cursor(line, column) when line >= 0 and column >= 0, do: "\e[#{line};#{column}H"
+
+  @doc "Sends cursor `lines` up (defaults to 1)."
+  @spec cursor_up :: String.t()
+  def cursor_up(), do: "\e[A"
+
+  @spec cursor_up(integer) :: String.t()
+  def cursor_up(lines) when lines > 0, do: "\e[#{lines}A"
```
fertapric: Last thing, promise 😅 

The docs are now assigned to `cursor_up/0`, leaving `cursor_up/1` without any documentation.

mveytsman: Actually thanks for all this feedback, as you can tell it's been a while and my elixir is a little rusty :)

is there a normal way to solve this problem?

fertapric: Multi-arity functions usually behave somehow different, so their documentation is usually not the same. There are some example in the Elixir codebase, [`Path.absname/1`](https://hexdocs.pm/elixir/Path.html#absname/1) and [`Path.absname/2`](https://hexdocs.pm/elixir/Path.html#absname/2) could be one of them.

lib/elixir/lib/io/ansi.ex
```diff
@@ -170,8 +170,10 @@ defmodule IO.ANSI do
   @doc "Sends cursor home."
   defsequence(:home, "", "H")
 
-  @doc "Sends cursor to the absolute position specified by `lines`, `columns`, with 0,0 being the top left corner."
-  @spec cursor(integer, integer) :: String.t()
```
mariusbutuc: No more `@spec`?

whatyouhide:
> ANSI will actually default to 1 for us if we don't pass an argument, i.e. "\e[A" moves 1 line up. I think that's the correct behavior with no argument for us as well.

@mveytsman is there any practical advantage in using `\e[A` instead of `\e1[A`, or any difference in general? If there isn't and it's just a "standard" think, personally I'd rather go with `cursor_up(columns \\ 1)` so the code and documentation are a bit more compact.

mveytsman:
@whatyouhide that's a fair point. Looking around, I don't see "\e[A" documented very much other than in the standard, so moving to `\\ 1` makes sense.

My thinking was that someone may be dealing with a "weird" terminal and want to send the "\e[A" directly but in that case they can just do it themselves...

whatyouhide:
> someone may be dealing with a "weird" terminal and want to send the "\e[A" 

Wouldn't the terminal be technically not working if `\e[A` wasn't the same as `\e1[A`? Anyways if something like this ever happens, we'll deal with it when the bug report comes in :)

lib/elixir/lib/io/ansi.ex
```diff
@@ -170,6 +170,29 @@ defmodule IO.ANSI do
...
   defsequence(:home, "", "H")
 
+  @doc """
+  Sends cursor to the absolute position specified by `lines`, `columns`, 
+  with 0,0 being the top left corner.
+  """
+  @spec cursor(integer, integer) :: String.t()
+  def cursor(line, column) when line >= 0 and column >= 0, do: "\e[#{line};#{column}H"
+
+  @doc "Sends cursor `lines` up. Defaults to 1"
```
whatyouhide: We don't need to mention the default, as `\\` is standard in Elixir. Let's just have 

```elixir
@doc "Sends cursor `lines` lines up.
```

lib/elixir/lib/io/ansi.ex
```diff
@@ -170,6 +170,29 @@ defmodule IO.ANSI do
   @doc "Sends cursor home."
   defsequence(:home, "", "H")
 
+  @doc """
+  Sends cursor to the absolute position specified by `lines`, `columns`, 
+  with 0,0 being the top left corner.
+  """
+  @spec cursor(integer, integer) :: String.t()
+  def cursor(line, column) when line >= 0 and column >= 0, do: "\e[#{line};#{column}H"
```
whatyouhide: If you want to check for the arguments being also integers, you should explicitly have a `is_integer/1` check. `>` and `>=` works for floats as well.

Please do this on all other functions as well :).

lib/elixir/lib/io/ansi.ex
```diff
@@ -170,6 +170,29 @@ defmodule IO.ANSI do
   @doc "Sends cursor home."
   defsequence(:home, "", "H")
 
+  @doc """
+  Sends cursor to the absolute position specified by `lines`, `columns`, 
+  with 0,0 being the top left corner.
```
whatyouhide: The first line of the documentation should be a summary of the function. Let's reshape this to be:

```elixir
@doc """
Sends cursor to the absolute position specified by `line` and `column`.

Line `0` and column `0` would mean the top left corner.
"""
```

lib/elixir/lib/io/ansi.ex
```diff
@@ -171,27 +171,30 @@ defmodule IO.ANSI do
...
-  Sends cursor to the absolute position specified by `lines`, `columns`, 
-  with 0,0 being the top left corner.
+  Sends cursor to the absolute position specified by `line` and `column`.
+
+  Line `0` and column `0` would mean the top left corner.
   """
   @spec cursor(integer, integer) :: String.t()
-  def cursor(line, column) when line >= 0 and column >= 0, do: "\e[#{line};#{column}H"
+  def cursor(line, column)
+      when is_integer(line) and line >= 0 and is_integer(column) and column >= 0,
```
mveytsman: @whatyouhide this is how `mix format` formatted it but it looks a little ugly. Think I should make a `is_positive_integer` guard / do we have one somewhere?

whatyouhide: We don't but this formatting is fine. When `when` gets split on its own line, we just use the `do`/`end` syntax:

```elixir
def cursor(line, column) 
    when is_integer(line) and line >= 0 and is_integer(column) and column >= 0 do
  ...
end
```

lib/elixir/test/elixir/io/ansi_test.exs
```diff
@@ -155,4 +155,53 @@ defmodule IO.ANSITest do
...
+    assert_raise FunctionClauseError, fn ->
+      IO.ANSI.cursor(5, -1)
+    end
+  end
+
+  test "cursor_up/1" do
+    assert IO.ANSI.cursor_up() == "\e[1A"
+    assert IO.ANSI.cursor_up(12) == "\e[12A"
+
+    assert_raise FunctionClauseError, fn ->
```
fertapric: Should we also test the value `0`? it seems a sensitive value because in an alternative implementation could be interpreted as `\e[A` or `\e[0A`. `\e[0A` works on my terminal and it's equivalent to `\e[A` or `\e[1A`.

josevalim: Good catch! Wouldn't it be best to have a guard that the value is `>= 1`? :)

michalmuskala: If we go that way, the typespec should use `pos_integer()`

mveytsman: I have the guard as `> 0`, but I didn't realize `pos_integer()` is a built-in. will change to that

mveytsman: I had the guard as `> 0` but `>= 1` is clearer. I changed that and the typespecs

lib/elixir/test/elixir/io/ansi_test.exs
```diff
@@ -155,4 +155,69 @@ defmodule IO.ANSITest do
...
+      IO.ANSI.cursor_left(0)
+    end
+
+    assert_raise FunctionClauseError, fn ->
+      IO.ANSI.cursor_left(-1)
+    end
+  end
+
+  test "cursor_right/0" do
```
fertapric: `cursor_right/1`

whatyouhide:
Thanks @mveytsman! 💟

PR #7398, opened 2018-02-26
Allow subdirectories in .formatter.exs

josevalim:


lib/mix/lib/mix/tasks/format.ex
```diff
@@ -142,71 +149,110 @@ defmodule Mix.Tasks.Format do
...
-    Mix.Utils.stale?([dot_formatter | Mix.Project.config_files()], [deps_manifest])
+  defp safe_binary_to_term(binary) do
+    {:ok, :erlang.binary_to_term(binary)}
+  rescue
+    _ -> :error
   end
 
-  defp read_deps_manifest(deps_manifest) do
-    deps_manifest |> File.read!() |> :erlang.binary_to_term()
+  defp write_manifest!(manifest, {entry, sources}) do
```
whatyouhide: This might be nitpicking, but I think it would make more sense from a code clarity point of view if `write_manifest!/2` would ignore what's being written. This would mean writing a manifest that looks like:

```elixir
{@manifest_vsn, term}
```

which I think makes more sense as the shape will stay the same even if we change the manifest version. It would simply return the term being written so this change wouldn't affect anything else in this PR I believe.

lib/mix/lib/mix/tasks/format.ex
```diff
@@ -267,12 +311,48 @@ defmodule Mix.Tasks.Format do
...
+      expand_dot_inputs(sub_formatter, [sub], formatter_opts_and_subs, acc)
+    end)
+  end
+
+  defp find_formatter_opts_for_file(split, {formatter_opts, subs}) do
+    Enum.find_value(subs, formatter_opts, fn {sub, formatter_opts_and_subs} ->
+      if List.starts_with?(split, Path.split(sub)) do
+        find_formatter_opts_for_file(split, formatter_opts_and_subs)
+      end
+    end)
```
whatyouhide: Can you elaborate on the reasoning behind doing this? I am having a bit of a hard time understanding it.

josevalim:
> If I got it correctly and it is like this, what is the reason why instead of doing the whole dance of finding the "closest" formatter file we don't just run mix format from the directory where each .formatter.exs is in? 

You are correct. The only reason we don't do so in the directory each .formatter.exs is in is because of the cache. For example, Phoenix plans to put a .formatter.exs inside priv/repo, and simply running in that directory means we have no sensible place to put the cache in.

We could devise a schema so we can still cache the inner .formatter.exs to complex paths in the manifest path but I thought keeping everything together is the simplest approach, especially because we still want to support loading dependencies from inner .formatter.exs files and things like `Mix.Project.deps_paths` can break if we are changing directories.

lib/mix/lib/mix/tasks/format.ex
josevalim: @whatyouhide the reason why I decided to not do so is because the manifest is now responsible to tell if it is out of date or not. We do so to avoid parsing the subdirectories in every command. So the manifest needs to know about the sources and I pushed the versioning concerns into it.

The `entry` part is completely opaque though.

lib/mix/lib/mix/tasks/format.ex
josevalim: This is the finding the closest `.formatter.exs` bit for when you invoke the formatter as `mix format path/to/some/file.ex`.

lib/mix/lib/mix/tasks/format.ex
```diff
@@ -22,6 +22,10 @@ defmodule Mix.Tasks.Format do
     * `:inputs` (a list of paths and patterns) - specifies the default inputs
       to be used by this task. For example, `["mix.exs", "{config,lib,test}/**/*.{ex,exs}"]`.
 
+    * `:subdirectories` (a list of paths and patterns) - specifies subdirectories
```
fertapric: > Overall it looks good. As far as I can tell, we don't merge the option in `a/b/.formatter.exs` with the options in `a/.formatter.exs` when formatting `a/b/foo.ex`, right? If I got it correctly and it is like this...

I also had the same doubts, and I had to look at the source code. I think a more detailed explanation about the inheritance rules between `.formatter.exs` and its `:subdirectories` could be helpful :)

fertapric:
> what is the reason why instead of doing the whole dance of finding the "closest" formatter file we don't just run mix format from the directory where each `.formatter.exs` is in?

I think this PR could be relevant https://github.com/elixir-lang/elixir/pull/7328 :)

lib/mix/lib/mix/tasks/format.ex
josevalim: I have clarified it here, thank you!

josevalim:
@whatyouhide thanks for the changes. You quoted one "lib/app" but not the other, so I went ahead and remove the quotes you added. We can change it later but i don't think it matters. :)

josevalim:

lib/mix/lib/mix/tasks/format.ex
```diff
@@ -26,7 +26,7 @@ defmodule Mix.Tasks.Format do
       that have their own formatting rules. Each subdirectory should have a
       `.formatter.exs` that configures how entries in that subdirectory should be
       formatted as. Configuration between `.formatter.exs` are not shared nor
-      inherited. If a `.formatter.exs` lists `"lib/app"` as a subdirectory, the rules
+      inherited. If a `.formatter.exs` lists "lib/app" as a subdirectory, the rules
```
whatyouhide: This will look horrible :P

whatyouhide:
@josevalim ah I see now, I didn't notice I missed one. It's fine for now yeah, no problem :) Great job!

PR #7407, opened 2018-03-01
Extend Macro.escape/2 docs to illustrate difference from quote/2

iamvery:
# Problem

Today I realized I couldn't remember the difference in Macro.escape/2 and Kernel.quote/2. The documented examples weren't quite enough to illustrate the difference as the results are the same from both.

# Solution

Extend documentation for Macro.escape/2 to illustrate the key difference in it from Kernel.quote/2. Namely: when working on a variable, Macro.escape/2 operates on the _value_.

### DIFF
--- a/lib/elixir/lib/macro.ex
+++ b/lib/elixir/lib/macro.ex
@@ -370,6 +370,31 @@ defmodule Macro do
       iex> Macro.escape({:unquote, [], [1]}, unquote: true)
       1
 
+  ## Comparison to `Kernel.quote/2`
+
+  The `escape/2` function is sometimes confused with `Kernel.SpecialForms.quote/2`,
+  because the above examples behave the same with both. The key difference is
+  best illustrated when the value to escape is stored in a variable.
+
+      iex> Macro.escape({:a, :b, :c})
+      {:{}, [], [:a, :b, :c]}
+      iex> quote do: {:a, :b, :c}
+      {:{}, [], [:a, :b, :c]}
+
+      iex> value = {:a, :b, :c}
+      iex> Macro.escape(value)
+      {:{}, [], [:a, :b, :c]}
+
+      iex> quote do: value
+      {:value, [], __MODULE__}
+
+      iex> value = {:a, :b, :c}
+      iex> quote do: unquote(value)
+      {:a, :b, :c}
+
+  `escape/2` is used to escape _values_ (either directly passed or variable
+  bound), while `Kernel.SpecialForms.quote/2` produces syntax trees for
+  expressions.
   """
   @spec escape(term, keyword) :: Macro.t()
   def escape(expr, opts \\ []) do
--- a/lib/elixir/lib/macro.ex
+++ b/lib/elixir/lib/macro.ex
@@ -392,7 +392,7 @@ defmodule Macro do
       iex> quote do: unquote(value)
       {:a, :b, :c}
 
-  `escape/2` is used to escape _values_ (either directly passed or variable
+  `escape/2` is used to escape *values* (either directly passed or variable
   bound), while `Kernel.SpecialForms.quote/2` produces syntax trees for
   expressions.
   """


lib/elixir/lib/macro.ex
```diff
@@ -370,6 +370,22 @@ defmodule Macro do
       iex> Macro.escape({:unquote, [], [1]}, unquote: true)
       1
 
+  The Macro.escape/2 function is sometimes confused with Kernel.quote/2,
+  because the above examples behave the same with both. The key difference is
+  best illustrated when the value to escape is stored in a variable.
+
+  ## Comparison to Kernel.quote/2
```
eksperimental: Same here

lib/elixir/lib/macro.ex
```diff
@@ -370,6 +370,25 @@ defmodule Macro do
...
+      iex> value = {:a, :b, :c}; Macro.escape(value)
+      {:{}, [], [:a, :b, :c]}
+
+      iex> quote do: value
+      {:value, [], __MODULE__}
+
+      iex> value = {:a, :b, :c}; quote do: unquote(value)
+      {:a, :b, :c}
+
+  Macro.escape/2 is used to escape _values_.
```
eksperimental: Same here

iamvery:
Updated with backticks 🙂

lib/elixir/lib/macro.ex
```diff
@@ -370,6 +370,25 @@ defmodule Macro do
       iex> Macro.escape({:unquote, [], [1]}, unquote: true)
       1
 
+  The `Macro.escape/2` function is sometimes confused with `Kernel.quote/2`,
```
whatyouhide: You don't need to use the module name when the function you're referring to is in the current module. Please use `escape/2` here and in other places instead of `Macro.escape/2`.

lib/elixir/lib/macro.ex
```diff
@@ -370,6 +370,25 @@ defmodule Macro do
...
+  because the above examples behave the same with both. The key difference is
+  best illustrated when the value to escape is stored in a variable.
+
+  ## Comparison to `Kernel.quote/2`
+
+      iex> quote do: {:a, :b, :c}
```
whatyouhide: Please use what the formatter would output which if I am not wrong should be

```elixir
iex> quote(do: {:a, :b, :c})
```

I don't have a computer with Elixir around, but you can verify that by doing `"quote do: {:a, :b, :c}" |> Code.format_string!() |> IO.puts()` in an IEx session. Do this for all `quote`s please :)

lib/elixir/lib/macro.ex
```diff
@@ -370,6 +370,25 @@ defmodule Macro do
...
+      {:{}, [], [:a, :b, :c]}
+
+      iex> value = {:a, :b, :c}; Macro.escape(value)
```
whatyouhide: Please do this instead of putting two expressions on the same line and using `;`:

```elixir
iex> value = {:a, :b, :c}
iex> Macro.escape(value)
```

here and in other places.

lib/elixir/lib/macro.ex
whatyouhide: Also, please move this paragraph to the following section "Comparison to Kernel.quote/2".

lib/elixir/lib/macro.ex
iamvery: I ran all the code in the examples through the formatter and got this:

```elixir
quote do: {:a, :b, :c}
value = {:a, :b, :c}
Macro.escape(value)
quote do: value
value = {:a, :b, :c}
quote do: unquote(value)
```

So the only change appears to be the variable bindings, however I'm not sure if I can split variable binding in doctests. Can you confirm?

iamvery: Even so, I'm happy to wrap the quote calls in parens! I don't have a strong opinion :)

iamvery: I'm mistaken. The separate variable binding works. Not sure what I did wrong. Will update!

lib/elixir/lib/macro.ex
iamvery: https://github.com/elixir-lang/elixir/pull/7407/files#r171648405

lib/elixir/lib/macro.ex
```diff
@@ -370,6 +370,27 @@ defmodule Macro do
       iex> Macro.escape({:unquote, [], [1]}, unquote: true)
       1
 
+  ## Comparison to `Kernel.quote/2`
+
+  The `escape/2` function is sometimes confused with `Kernel.quote/2`,
```
whatyouhide: You want `Kernel.SpecialForms.quote/2` in order for it to be linked, not `Kernel.quote/2` :)

lib/elixir/lib/macro.ex
```diff
@@ -370,6 +370,27 @@ defmodule Macro do
...
+      {:{}, [], [:a, :b, :c]}
+
+      iex> value = {:a, :b, :c}
+      iex> Macro.escape(value)
+      {:{}, [], [:a, :b, :c]}
+
+      iex> quote do: value
+      {:value, [], __MODULE__}
+
+      iex> value = {:a, :b, :c}
```
iamvery: @whatyouhide updated! I'm seeing missing variable errors when I don't bind the value for each example. I forget how doctests are isolated from one another, but perhaps it isn't surprising? Let me know what you think!

whatyouhide: Yeah bind it on every example. That's very much fine :)

lib/elixir/lib/macro.ex
iamvery: DOH!

lib/elixir/lib/macro.ex
```diff
@@ -370,6 +370,27 @@ defmodule Macro do
       iex> Macro.escape({:unquote, [], [1]}, unquote: true)
       1
 
+  ## Comparison to `Kernel.quote/2`
+
+  The `escape/2` function is sometimes confused with `Kernel.SpecialForms.quote/2`,
```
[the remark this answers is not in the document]
whatyouhide: Personally it's fine. :)

iamvery:
Thank you for all the help 🎉

lib/elixir/lib/macro.ex
```diff
@@ -370,6 +370,27 @@ defmodule Macro do
...
+  because the above examples behave the same with both. The key difference is
+  best illustrated when the value to escape is stored in a variable.
+
+      iex> quote do: {:a, :b, :c}
+      {:{}, [], [:a, :b, :c]}
+
```
eksperimental: I think we are missing the ` iex> Macro.escape({:a, :b, :c})` example

iamvery: I believe I left it out intentionally, since it's mentioned in the first set of examples. Do you feel like it's worth repeating here?

eksperimental: I dont think it will hurt being explicit in this particular case

eksperimental: ...but if you want to leave it out it's still fine

iamvery: I added it!

lib/elixir/lib/macro.ex
```diff
@@ -370,6 +370,30 @@ defmodule Macro do
...
+  The `escape/2` function is sometimes confused with `Kernel.SpecialForms.quote/2`,
+  because the above examples behave the same with both. The key difference is
+  best illustrated when the value to escape is stored in a variable.
+
+      iex> Macro.escape({:a, :b, :c})
+      {:{}, [], [:a, :b, :c]}
+
+      iex> quote do: {:a, :b, :c}
+      {:{}, [], [:a, :b, :c]}
```
eksperimental: I would put this in the same block above. Ie. Removing the emoty line

eksperimental: *empty line

lib/elixir/lib/macro.ex
```diff
@@ -370,6 +370,30 @@ defmodule Macro do
...
+
+      iex> value = {:a, :b, :c}
+      iex> Macro.escape(value)
```
eksperimental: `...>` instead of `iex>`

lib/elixir/lib/macro.ex
```diff
@@ -370,6 +370,30 @@ defmodule Macro do
...
+      {:{}, [], [:a, :b, :c]}
+
+      iex> quote do: value
+      {:value, [], __MODULE__}
+
+      iex> value = {:a, :b, :c}
+      iex> quote do: unquote(value)
```
eksperimental: `...>` instead of `iex>`

lib/elixir/lib/macro.ex
iamvery: TIL! Is there a functional difference, or does it just better communicate that the second line is a continuation?

lib/elixir/lib/macro.ex
iamvery: Like this?

```
iex> Macro.escape({:a, :b, :c})
{:{}, [], [:a, :b, :c]}
iex> quote do: {:a, :b, :c}
{:{}, [], [:a, :b, :c]}
```

lib/elixir/lib/macro.ex
whatyouhide: That is used for expressions that span over multiple lines. If you write the example here in an actual IEx session, the second prompt will not use ... so let's keep it like it is now.

iamvery:
Updated! I added a little more to the text after the examples too.

lib/elixir/lib/macro.ex
josevalim: No difference, just communication.

josevalim: But in this case `iex>` is better because it does not span over multiple lines.

josevalim:

PR #7408, opened 2018-03-01
Move `doctest Macro` to Macro module test file

iamvery:
This isn't a really big deal, but it was a little surprising to me that the doctest specification for the Macro module wasn't in the test file for the Macro module. Unless the decision to put it where it is, it might lower friction slightly to move it to the Macro module test file.

Thoughts?


josevalim:

PR #7413, opened 2018-03-02
Add a simple, concrete @spec example

nathanl:
I found this page a little too abstract on first encounter. I thought a concrete example at the top would help newcomers.

### DIFF
--- a/lib/elixir/pages/Typespecs.md
+++ b/lib/elixir/pages/Typespecs.md
@@ -19,16 +19,20 @@ See the "User-defined types" and "Defining a specification" sub-sections below f
 ## A simple example
 
     defmodule StringHelpers do
-      @spec long_word?(String.t()) :: boolean()
+      @type word() :: String.t()
+
+      @spec long_word?(word()) :: boolean()
       def long_word?(word) when is_binary(word) do
         String.length(word) > 8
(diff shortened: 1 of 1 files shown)


whatyouhide:
I took the liberty of slightly updating the code and the description to make them simpler (or at least what I believe is simpler 😃). However, I am not too much in favour of this change: I think it breaks the flow of the documentation page and that the examples on how to write specs at the end of the page are enough. If we want to keep this, I'd rather move it down at the end, but I'll let others decide whether to keep it or not :)

josevalim:
@whatyouhide I do like the introduction, as I think the examples at the bottom are too limited and they come too late. Giving some context for those first reading the page can be welcome. My only question is if we should also use `@type` in the example but I think you want to keep it simple. :)

whatyouhide:
@josevalim I thought about using `@type` and I agree then yeah. We can do

```elixir
@type word() :: String.t()
```

and then use `word/0` and mention it with another bullet point with something like "we're defining the `word()` type which is equivalent to `String.t()`". Thoughts?

josevalim:
Sounds good to me!

whatyouhide:
Thanks @nathanl! 💟

PR #7417, opened 2018-03-04
Add calendar callback months_in_year/1

kipcole9:
Not all calendars have a constant 12 months in
a year.  This function returns the number of
months for a given year. Calendar.ISO will return
12 for all years.

### DIFF
--- a/lib/elixir/lib/calendar/date.ex
+++ b/lib/elixir/lib/calendar/date.ex
@@ -177,6 +177,23 @@ defmodule Date do
   end
 
   @doc """
+  Returns the number of days in the given `date` month.
+
+  ## Example
+
+      iex> Date.months_in_year(~D[1900-01-13])
+      12
+
+  """
+  @since "1.7.0"
+  @spec months_in_year(Calendar.date()) :: Calendar.month()
+  def months_in_year(date)
+
+  def months_in_year(%{calendar: calendar, year: year}) do
+    calendar.months_in_year(year)
+  end
+
+  @doc """
   Builds a new ISO date.
 
   Expects all values to be integers. Returns `{:ok, date}` if each
--- a/lib/elixir/lib/calendar/iso.ex
+++ b/lib/elixir/lib/calendar/iso.ex
@@ -34,6 +34,8 @@ defmodule Calendar.ISO do
   @days_per_nonleap_year 365
   @days_per_leap_year 366
 
+  @months_in_year 12
+
   @doc """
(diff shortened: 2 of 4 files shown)


josevalim:
We should also add a function in Date that takes the year+calendar and invoke it. Similar to days_in_month: https://github.com/elixir-lang/elixir/blob/master/lib/elixir/lib/calendar/date.ex#L171-L177

Other than that, this looks beautiful.

kipcole9:
PR is updated to reflect the addition of `Date.months_in_year/1`

josevalim:

lib/elixir/lib/calendar/date.ex
```diff
@@ -176,6 +176,23 @@ defmodule Date do
     calendar.days_in_month(year, month)
   end
 
+  @doc """
+  Returns the number of days in the given `date` month.
```
David-Gil: Shouldn't it be "Returns the number of months in the given year"?

josevalim:
Yes, good catch. Please do send a PR!
-- 


lib/elixir/lib/calendar/date.ex
whatyouhide: Good catch @David-Gil, fixed in 7d621f60a9bd50a03f8eca8fc524aa7d041fa8f1 💟

lib/elixir/lib/calendar/iso.ex
```diff
@@ -232,6 +234,22 @@ defmodule Calendar.ISO do
   def days_in_month(_, month) when month in [4, 6, 9, 11], do: 30
   def days_in_month(_, month) when month in 1..12, do: 31
 
+  @doc """
```
fertapric: What do you think about using the module variable in the `@doc` and `@spec`.

```elixir
@doc """
Returns how many months there are in the given year.

## Example

    iex> Calendar.ISO.months_in_year(2004)
    #{@months_in_year}

"""
@impl true
@since "1.7.0"
@spec months_in_year(year) :: unquote(@months_in_year)
def months_in_year(_year) do
  @months_in_year
end
```

or maybe just remove the module variable and use 12 directly like it's done in `days_in_month/2`.

```elixir
@doc """
Returns how many months there are in the given year.

## Example

    iex> Calendar.ISO.months_in_year(2004)
    12

"""
@impl true
@since "1.7.0"
@spec months_in_year(year) :: 12
def months_in_year(_year), do: 12
```

josevalim: Looks good to me too!

Issue #5800, opened 2017-02-21
Perform erl_lint checks when building the AST


whatyouhide:
We aim to perform as many checks that `erl_lint` performs as we can when building the Elixir AST (during expansion). This issue exists to keep track of the checks that `erl_lint` performs; some of them we already perform, some of them are invalid for Elixir, and some of them we want to implement. [Reference](https://github.com/erlang/otp/blob/bd25898d3bf63316fcf116ba03f92c12500927fd/lib/stdlib/src/erl_lint.erl#L151-L407).

## General

- [x] `undefined_module`: no module definition
- [x] `redefine_module`
- [x] `pmod_unsupported`: parametrized modules are no longer supported
- [x] `{redefine_mod_import, M, P}`: module already imported from package `P`
- [x] `invalid_call`: invalid function call
- [x] `invalid_record`: invalid record expression
- [x] `{attribute,A}`: attribute after function definitions
- [x] `{missing_qlc_hrl,A}`: `qlc:q` called but `qlc.hrl` not included
- [x] `{redefine_import,{{F,A},M}}`: function already imported from another module
- [x] `{bad_inline,{F,A}}`: inlined function undefined
- [x] `{invalid_deprecated,D}`: badly formed deprecated attribute
- [x] `{bad_deprecated,{F,A}}`: depreacted function undefined or not exported
- [x] `{bad_nowarn_unused_function,{F,A}}`: function undefined
- [x] `{bad_nowarn_bif_clash,{F,A}}`: function undefined
- [x] `disallowed_nowarn_bif_clash`: `-compile` directive `nowarn_bif_clash` is no longer allowed
- [x] `{bad_nowarn_deprecated_function,{M,F,A}}`: not a depreacted function
- [x] `{bad_on_load,Term}`: badly formed `on_load` attribute
- [x] `multiple_on_loads`: multiple `on_load` attributes
- [x] `{bad_on_load_arity,{F,A}}`: arity must be `0`
- [x] `{undefined_on_load,{F,A}}`: function for `on_load` undefined
- [x] `export_all`: `export_all` flag enabled
- [x] `{duplicated_export, {F,A}}`: function already exported
- [x] `{unused_import,{{F,A},M}}`
- [x] `{undefined_function,{F,A}}`
- [x] `{redefine_function,{F,A}}`: function already defined
- [x] `{define_import,{F,A}}`: defining imported function
- [x] `{unused_function,{F,A}}`
[...]

PR #7422, opened 2018-03-06
Add a "erl_lint" check for behaviour_info/1 plus callback attributes

whatyouhide:
Part of https://github.com/elixir-lang/elixir/issues/5800.

@josevalim if we want to keep the shape of the `erl_lint` error (`{behaviour_info, {M, F, A}}`), then I don't think we can give more information than that and so we can't say `@callback` or `@macrocallback` with certainty, that's why I went with `@callback/@macrocallback` in the error message.

### DIFF
--- a/lib/elixir/src/elixir_erl.erl
+++ b/lib/elixir/src/elixir_erl.erl
@@ -375,9 +375,20 @@ specs_form(Map, Data, Defmacro, Unreachable, Forms) ->
   Optional = lists:flatten(take_type_spec(Data, optional_callbacks)),
   SpecsForms = specs_form(spec, Specs, Unreachable, [], Defmacro, Forms),
   AllCallbacks = Callbacks ++ Macrocallbacks,
+  validate_behaviour_info_and_attributes(Map, AllCallbacks),
   validate_optional_callbacks(Map, AllCallbacks, Optional),
   specs_form(callback, AllCallbacks, [], Optional,
              [NameArity || {_, NameArity, _, _} <- Macrocallbacks], SpecsForms).
+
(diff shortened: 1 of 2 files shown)


lib/elixir/src/elixir_erl.erl
```diff
@@ -375,10 +375,21 @@ specs_form(Map, Data, Defmacro, Unreachable, Forms) ->
...
   validate_optional_callbacks(Map, AllCallbacks, Optional),
   specs_form(callback, AllCallbacks, [], Optional,
              [NameArity || {_, NameArity, _, _} <- Macrocallbacks], SpecsForms).
 
+validate_behaviour_info_and_attributes(#{definitions := Defs, module := Mod} = Map, AllCallbacks) ->
+  case {lists:keyfind({behaviour_info, 1}, 1, Defs), AllCallbacks} of
+    {false, _} ->
+      ok;
+    {_, [{Kind, {Name, Arity}, _, _} | _]} when Kind == callback; Kind == macrocallback ->
+      form_error(Map, {behaviour_info, {Mod, Name, Arity}});
```
josevalim: We don't need to keep the same message shape. :)

whatyouhide: Oh cool, okay. I don't know why I thought we did. I'll change it then!

josevalim: Also, do we need to make sure to do nothing when we receive this message in `elixir_errors.erl`?

whatyouhide: @josevalim ah yes, probably. I saw that adding the error here made this error pop up and didn't think about doing nothing for the old error.

whatyouhide: @josevalim since ours pops up before, I think this PR is ready to go then!

PR #7427, opened 2018-03-07
Use relative paths for logger file when application is available

josevalim:
Today the Logger includes the full path to the source file in the `:file` metadata. With this PR, we use the relative path if the application is available, as that is enough to uniquely identify the path and it helps normalize and reduce the size of the metadata.

For example, before this patch, it would report:

     application=interpret_logger_call file=/Users/jose/OSS/interpret_logger_call/lib/interpret_logger_call.ex

and now it reports:

    application=interpret_logger_call file=lib/interpret_logger_call.ex

### DIFF
--- a/lib/logger/lib/logger.ex
+++ b/lib/logger/lib/logger.ex
@@ -718,7 +718,8 @@ defmodule Logger do
     %{module: module, function: fun, file: file, line: line} = caller
 
     caller =
-      compile_time_application(file) ++ [module: module, function: form_fa(fun), line: line]
+      compile_time_application_and_file(file) ++
+        [module: module, function: form_fa(fun), line: line]
 
     metadata =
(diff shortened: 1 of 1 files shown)


michalmuskala:
The formatter complained, btw

lib/logger/lib/logger.ex
```diff
@@ -718,8 +718,7 @@ defmodule Logger do
     %{module: module, function: fun, file: file, line: line} = caller
 
     caller =
-      compile_time_application() ++
-        [module: module, function: form_fa(fun), file: file, line: line]
+      compile_time_application(file) ++ [module: module, function: form_fa(fun), line: line]
```
whatyouhide: Maybe this should be called `compile_time_application_and_file/1`?

josevalim:

Issue #7428, opened 2018-03-07
Unexpected :timed_out message on failed compilation using Mix.Task.compile/2 with "--return-errors" option


allansson:
### Environment

* Elixir & Erlang versions (elixir --version): 
```
Erlang/OTP 20 [erts-9.0] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:10]

Elixir 1.6.2 (compiled with OTP 19)
```
* Operating system: Windows 10 x64

### Current behavior

When compiling an mix project with compilation errors using `Mix.Task.run("compile", ["--return-errors"])` a `:timed_out` message is sent to the process invoking the task.

To reproduce, create a new mix project and make sure the code does not compile, e.g. by trying to reference a variable/function which does not exist, then modify and paste the following in an `iex`-session:

```elixir
defmodule SSCCE do
  def run() do
    File.cd!("<DIRECTORY TO FAILING MIX PROJECT>")
    Mix.Task.clear()
    Mix.Task.run("compile", ["--return-errors"])

    receive do
      {:timed_out, pid} ->
        IO.puts("Unexpected timeout received from #{inspect(pid)}")
    after
      60_000 ->
        IO.puts("Works on my computer!")
    end
  end
end

SSCCE.run()

```

Eventually, the following message should be printed:

```
Unexpected timeout received from #PID<0.233.0>
```

### Expected behavior

All scheduled timeouts should be cancelled  after compilation has finished and no `:timed_out` messages should be sent.

### Investigation

The `:timed_out` message is sent from a `Process.send_after` scheduled by [`Kernel.ParallellCompiler`](https://github.com/elixir-lang/elixir/blob/a044559e8203e1a2b321fdcad4e66258c0d2ff5b/lib/elixir/lib/kernel/parallel_compiler.ex#L218). The timeout is cancelled in [`wait_for_messages`](https://github.com/elixir-lang/elixir/blob/a044559e8203e1a2b321fdcad4e66258c0d2ff5b/lib/elixir/lib/kernel/parallel_compiler.ex#L283), but one clause seem to be missing the call to cancel the timeout: [the final `{:file_done, _}` clause](https://github.com/elixir-lang/elixir/blob/a044559e8203e1a2b321fdcad4e66258c0d2ff5b/lib/elixir/lib/kernel/parallel_compiler.ex#L360). This clause appears to be in charge of handling compilation errors, which explains why the problem only appears to occur when compilation fails.

[...]

PR #7429, opened 2018-03-07
Properly cancel timeout on compilation error in ParallelCompiler #7428

allansson:
The `:timed_out` message was not cancelled properly when compilation failed in ParallelCompiler. This meant the timeout message would be delivered even after compilation completed. This could be problematic when invoking compilation programmatically via e.g. `Mix.Task.run("compile", ["--return-errors"])` since it means that the calling process, depending on the implemenation, might crash unexpectedly or get its inbox filled with messages.

The fix simply calls the `cancel_waiting_timer` when a compilation error has ocurred in the same way it is called in other similar cases.

This fixes issue #7428

### DIFF
--- a/lib/elixir/lib/kernel/parallel_compiler.ex
+++ b/lib/elixir/lib/kernel/parallel_compiler.ex
@@ -360,6 +360,7 @@ defmodule Kernel.ParallelCompiler do
       {:file_done, child_pid, file, {kind, reason, stack}} ->
         discard_down(child_pid)
         print_error(file, kind, reason, stack)
+        cancel_waiting_timer(queued, child_pid)
         terminate(queued)
         {:error, [to_error(file, kind, reason, stack)], warnings}
 


josevalim:
@allansson can you see if it is possible to check for this behaviour in a test? The [test suite is here](https://github.com/elixir-lang/elixir/blob/master/lib/elixir/test/elixir/kernel/parallel_compiler_test.exs) and we already have failing tests for the parallel compiler, maybe one of those have a leaked message that will be fixed with your test? It should be a matter of calling `refute_received {:timed_out, _}`. Thanks!

allansson:
Of course! I did look at the tests and considered implementing one, but decided not to because there can be a considerable delay between compilation finishing and the timeout message being received. I didn't want to create a test which would have to run for a at least 10 seconds or so. 

I was not aware of the `refute_received` assertion. It will make things really easy. 

I'll try to fix at soon as possible. While I am at it, I suppose could write some tests for when compilation is successful as well. ^.^

josevalim:
I see your point about the timeout. We could set it to one second but then
we can have races. With all that said, I think this is good to merge as is.
-- 


allansson:
Ok. I have almost completed writing two tests. 

I can get around the long wait by setting the `long_compilation_threshold` option to something low. But that means that if compilation was to stall for too long for whatever reason, the tests would fail.

If it was possible to "spy" on calls to `send_after` and `cancel_timer` I could of course verify that all references are cleared up after execution has finished, but I am not aware of any way of doing so?

josevalim:

PR #7431, opened 2018-03-08
Implement check for @spec attributes for undefined functions

whatyouhide:
This is part of https://github.com/elixir-lang/elixir/issues/5800. I implemented the check that checks that a `@spec` attribute is for a function/macro that is actually defined.

@josevalim this implementation might be very naive but it works for simple cases and for the tests that I did. I don't know how this is related to https://github.com/elixir-lang/elixir/blob/7bc0df82fa1c79063ce36d29a2bed90bfb71a7eb/lib/elixir/src/elixir_erl_compiler.erl#L82-L83, I can't see where `{spec_fun_undefined, {M, F, A}}` (with a module, not just `{F, A}`) is used or raised. Let me know how this looks 💟

### DIFF
--- a/lib/elixir/src/elixir_erl.erl
+++ b/lib/elixir/src/elixir_erl.erl
@@ -416,7 +416,7 @@ specs_form(Kind, Entries, Unreachable, Optional, Macros, Forms, Map) ->
   M =
     lists:foldl(fun({_, NameArity, Line, Spec}, Acc) ->
       case Kind of
-        spec -> validate_spec_for_existing_function(Map, NameArity);
+        spec -> validate_spec_for_existing_function(Map, NameArity, Line);
         _ -> ok
       end,
 
@@ -466,9 +466,9 @@ take_type_spec(Data, Key) ->
     [] -> []
   end.
 
-validate_spec_for_existing_function(#{definitions := Defs} = Map, NameAndArity) ->
+validate_spec_for_existing_function(#{definitions := Defs} = Map, NameAndArity, Line) ->
   case lists:keymember(NameAndArity, 1, Defs) of
-    false -> form_error(Map, {spec_for_undefined_function, NameAndArity});
+    false -> form_error(Map#{line := Line}, {spec_for_undefined_function, NameAndArity});
     true -> ok
   end.
 
--- a/lib/elixir/src/elixir_erl.erl
+++ b/lib/elixir/src/elixir_erl.erl
@@ -373,12 +373,12 @@ specs_form(Map, Data, Defmacro, Unreachable, Forms) ->
      {Kind, Expr, Caller} <- take_type_spec(Data, macrocallback)],
 
   Optional = lists:flatten(take_type_spec(Data, optional_callbacks)),
-  SpecsForms = specs_form(spec, Specs, Unreachable, [], Defmacro, Forms),
+  SpecsForms = specs_form(spec, Specs, Unreachable, [], Defmacro, Forms, Map),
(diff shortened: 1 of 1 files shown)


whatyouhide:
I also very naively fixed the line of the error in the last commit, there might be a better way but this Works™.

lib/elixir/src/elixir_erl.erl
```diff
@@ -410,11 +410,16 @@ validate_optional_callbacks(Map, AllCallbacks, Optional) ->
     maps:put(Callback, true, Acc)
   end, #{}, Optional).
 
-specs_form(_Kind, [], _Unreacheable, _Optional, _Macros, Forms) ->
+specs_form(_Kind, [], _Unreacheable, _Optional, _Macros, Forms, _Map) ->
   Forms;
-specs_form(Kind, Entries, Unreachable, Optional, Macros, Forms) ->
-  Map =
+specs_form(Kind, Entries, Unreachable, Optional, Macros, Forms, Map) ->
+  M =
```
josevalim: Let's give this a better name, like `SpecsMap`.

lib/elixir/src/elixir_erl.erl
```diff
@@ -410,11 +410,16 @@ validate_optional_callbacks(Map, AllCallbacks, Optional) ->
     maps:put(Callback, true, Acc)
   end, #{}, Optional).
 
-specs_form(_Kind, [], _Unreacheable, _Optional, _Macros, Forms) ->
+specs_form(_Kind, [], _Unreacheable, _Optional, _Macros, Forms, _Map) ->
   Forms;
-specs_form(Kind, Entries, Unreachable, Optional, Macros, Forms) ->
-  Map =
+specs_form(Kind, Entries, Unreachable, Optional, Macros, Forms, Map) ->
```
josevalim: Let's give this a better name, like `ModuleMap`.

lib/elixir/src/elixir_erl.erl
```diff
@@ -461,6 +466,12 @@ take_type_spec(Data, Key) ->
     [] -> []
   end.
 
+validate_spec_for_existing_function(#{definitions := Defs} = Map, NameAndArity, Line) ->
+  case lists:keymember(NameAndArity, 1, Defs) of
+    false -> form_error(Map#{line := Line}, {spec_for_undefined_function, NameAndArity});
```
josevalim: I would prefer to pass `#{line := line, file := ?m(Map, file)}` as it is more explicit since form_error was not really designed to handle the `ModuleMap`.

lib/elixir/src/elixir_erl.erl
```diff
@@ -373,12 +373,12 @@ specs_form(Map, Data, Defmacro, Unreachable, Forms) ->
      {Kind, Expr, Caller} <- take_type_spec(Data, macrocallback)],
 
   Optional = lists:flatten(take_type_spec(Data, optional_callbacks)),
-  SpecsForms = specs_form(spec, Specs, Unreachable, [], Defmacro, Forms),
+  SpecsForms = specs_form(spec, Specs, Unreachable, [], Defmacro, Forms, Map),
```
josevalim: Unreachable is also available in the `Map` and I think the only reason we pass it all the way down here is because we didn't pass the `Map`. But now that we do, I would change it so no arity of `specs_form` receive Unreachable and we also change `dynamic_form` to no longer return it.

whatyouhide:
@josevalim pushed an update that addressed all your comments. :)

Issue #7423, opened 2018-03-06
ExUnit @moduletag not working as described in the docs


bundacia:
### Environment

```
% elixir --version
Erlang/OTP 20 [erts-9.0.4] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:10] [hipe] [kernel-poll:false] [dtrace]

Elixir 1.6.1 (compiled with OTP 19)
% uname -v
Darwin Kernel Version 17.4.0: Sun Dec 17 09:19:54 PST 2017; root:xnu-4570.41.2~1/RELEASE_X86_64
```

### Current behavior

When I use a `@moduletag` as described in the docs [here](https://hexdocs.pm/ex_unit/ExUnit.Case.html#module-reporting-tags) I get the following error:

```
== Compilation error in file test/moduledoc_test.exs ==
** (FunctionClauseError) no function clause matching in anonymous fn/2 in ExUnit.Case.normalize_tags/1
```

See the test file and full error here:
https://gist.github.com/bundacia/a58db9fe5aa4f3ce46285071f16ac1da

If I change the `@moduletag` to this, things work as expected:

```
@moduletag [[report: [:foo]]]
```

### Expected behavior
I expected the tests to compile and run when I used the syntax specified in the docs.

I'm not sure if this is a bug in the docs or if the docs are "right" and the code should be changed to match.

bundacia:
Ahh, I see maybe this feature is going away? https://github.com/elixir-lang/elixir/commit/e0c112370ea1ca1c021ec69280faeda444690a2c#diff-da9b91caf1cd02e59ee50c01929a2058

josevalim:
@bundacia they are not going away. the module doc needs to be defined after use ExUnit.Case, i will make sure it emits a good error message.

bundacia:
Thanks!

m13m:
@josevalim I want to take this where do I have to make changes to add the error message?

whatyouhide:
@m13m I think you can start looking at where the exception is being raised now, that would be https://github.com/elixir-lang/elixir/blob/master/lib/ex_unit/lib/ex_unit/case.ex#L525.

PR #7432, opened 2018-03-09
ExUnit raise error for @moduletag and @tag when set before 'use ExUnit.Case'

aaruel:
Addresses #7423 

The original issue mentions only `@moduletag` but I found it happens with `@tag` as well.


josevalim:
Hi @aaruel! I am thinking a better fix would be to call `Module.get_attribute(__MODULE__, :moduletag)` and similar for `:tag` [around here](https://github.com/elixir-lang/elixir/blob/master/lib/ex_unit/lib/ex_unit/case.ex#L216) and then raise if they are not nil. What do you think?

aaruel:
@josevalim Ah yes, that's a much better solution! I've pushed changes implementing this idea.

josevalim:

PR #7436, opened 2018-03-10
Extract some functionality from elixir_erl to Kernel.Typespec

whatyouhide:


lib/elixir/lib/kernel/typespec.ex
```diff
@@ -1216,4 +1216,27 @@ defmodule Kernel.Typespec do
   defp unpack_typespec_kw(_, _acc) do
     :error
   end
+
+  # Called by elixir_erl.
+  def spec_related_info(table) do
```
josevalim: Beautiful implementation!

Although the name is too vague?

What about `translate_typespecs` or `translate_typespecs_for_module`?

lib/elixir/lib/kernel/typespec.ex
```diff
@@ -1216,4 +1216,27 @@ defmodule Kernel.Typespec do
...
+    translate_type = fn {kind, expr, caller} -> translate_type(kind, expr, caller) end
+    translate_spec = fn {kind, expr, caller} -> translate_spec(kind, expr, caller) end
```
josevalim: Isn't this just &translate_type/3 and &translate_spec/3?

whatyouhide: We are passing three-element tuples around, not three arguments :)

PR #7437, opened 2018-03-10
Modernize Kernel.Typespec by reducing usage of :lists functions

whatyouhide:


lib/elixir/lib/kernel/typespec.ex
```diff
@@ -174,11 +174,9 @@ defmodule Kernel.Typespec do
   @spec defines_spec?(module, atom, arity) :: boolean
   def defines_spec?(module, name, arity)
       when is_atom(module) and is_atom(name) and arity in 0..255 do
-    finder = fn {_kind, expr, _caller} ->
+    Enum.any?(Module.get_attribute(module, :spec), fn {_kind, expr, _caller} ->
       spec_to_signature(expr) == {name, arity}
-    end
-
-    :lists.any(finder, Module.get_attribute(module, :spec))
+    end)
```
whatyouhide: I was able to do this here and in `defines_callback?/3` below but I ran into bootstrapping problems doing this in https://github.com/elixir-lang/elixir/pull/7437/files#diff-6de4b356bc9d4e3d07f1e5731fbe7d23L167 (`defines_type?/3`), where it complained that `Enum.any?/2` was undefined 🤔

lib/elixir/lib/kernel/typespec.ex
```diff
@@ -577,21 +571,18 @@ defmodule Kernel.Typespec do
...
           [{:var, line, name}, typespec(type, vars, caller)]
         ]
 
-        type = {:type, line, :constraint, constraint}
-        [type | acc]
-    end
-
-    :lists.reverse(:lists.foldl(foldl_fun, [], guard))
+        [{:type, line, :constraint, constraint}]
+    end)
```
whatyouhide: This is really nice, was able to drop the reversing as well!

lib/elixir/lib/kernel/typespec.ex
```diff
@@ -1154,10 +1142,13 @@ defmodule Kernel.Typespec do
   end
 
   defp typespec(list, vars, caller) when is_list(list) do
-    [h | t] = :lists.reverse(list)
```
eksperimental: I think we should avoid single letter vars

lib/elixir/lib/kernel/typespec.ex
```diff
@@ -1154,10 +1142,13 @@ defmodule Kernel.Typespec do
...
+    [h | t] = Enum.reverse(list)
+
+    union =
+      Enum.reduce(t, validate_kw(h, list, caller), fn x, acc ->
```
eksperimental: Same here

lib/elixir/lib/kernel/typespec.ex
josevalim: defines_type? is used by Protocols and we need protocols to compile Enum. So I guess we can't do it for defines_type? although that's probably fine.

whatyouhide: Ah I see, cool!

PR #7443, opened 2018-03-11
Allow use of @describetag before a describe block.

ciaran:
My expectation for the use of the `@describetag` directive is to use it before the `describe` block, as demonstrated in [a recent addition to the docs](https://github.com/elixir-lang/elixir/blob/master/lib/ex_unit/lib/ex_unit/case.ex#L123).

However in practice when used this way the tag does not work, and is completely ignored unless used within the describe block (see [the test case in this repo](https://github.com/elixir-lang/elixir/blob/master/lib/ex_unit/test/ex_unit/describe_test.exs#L13)).

This PR assumes that current behaviour is a bug and fixes use of `@describetag` outside of the block, however if this is incorrect then using it outside of a block should be changed to display an error.

### DIFF
--- a/lib/ex_unit/lib/ex_unit/case.ex
+++ b/lib/ex_unit/lib/ex_unit/case.ex
@@ -391,7 +391,6 @@ defmodule ExUnit.Case do
 
       @ex_unit_describe {__ENV__.line, message}
       @ex_unit_used_describes message
-      Module.delete_attribute(__ENV__.module, :describetag)
 
       try do
         unquote(block)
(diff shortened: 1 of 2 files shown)


josevalim:
The current behaviour is intentional and meant to mirror @moduletag. If you would like to discuss this, please start a mailing list discussion. This has been discussed in the past too, so finding those previous discussions can be helpful too.

josevalim:
Oops, premature send. :)

I will update the docs and make sure we raise if @describetag is used before the describe block. Thanks for the PR!

ciaran:
Apologies, I only searched the issues on this repo.

Then if this is the desired behaviour I can make a separate PR which fixes the documentation and displays an informative error/warning when `@describetag` is used outside of a describe block.

josevalim:
Yeah, sorry for the confusion and the subtle close. My fingers got the best of me.

Issue #7434, opened 2018-03-09
ex_unit crash with invalid utf8 in doctests


michalmuskala:
### Environment

* Elixir & Erlang versions (elixir --version): 
```
Erlang/OTP 20 [erts-9.1] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:10] [hipe] [kernel-poll:false]

Elixir 1.7.0-dev (24790a80b) (compiled with OTP 20)
```
* Operating system: macOS

### Current behavior

Given a following doctest:

```elixir
      iex> Jason.encode("invalid utf8 \xFF")
      {:error, ""}
```

`mix test` crashes with:

```
=ERROR REPORT==== 9-Mar-2018::08:55:29 ===
** Generic server <0.164.0> terminating
** Last message in was {'$gen_cast',
                        {test_finished,
                         #{'__struct__' => 'Elixir.ExUnit.Test',
                           'case' => 'Elixir.JasonTest',logs => <<>>,
                           module => 'Elixir.JasonTest',
                           name => 'doctest Jason.encode/2 (6)',
                           state =>
                            {failed,
                             [{error,
                               #{'__exception__' => true,
                                 '__struct__' =>
                                  'Elixir.ExUnit.AssertionError',
                                 args => ex_unit_no_meaningful_value,
                                 expr =>
                                  <<"Jason.encode(\"invalid utf8 ÿ\")">>,
                                 left => ex_unit_no_meaningful_value,
                                 message =>
                                  <<"Doctest did not compile, got: (UnicodeConversionError) invalid encoding starting at <<255, 34, 41>>">>,
                                 right => ex_unit_no_meaningful_value},
                               [{'Elixir.Jason','__MODULE__',0,
                                 [{line,113},{file,<<"lib/jason.ex">>}]}]}]},
                           tags =>
                            #{async => true,'case' => 'Elixir.JasonTest',
                              describe => nil,describe_line => nil,
                              file =>
                               <<"/Users/michalmuskala/src/oss/jason/test/jason_test.exs">>,
                              line => 4,module => 'Elixir.JasonTest',
                              registered => #{},
                              test => 'doctest Jason.encode/2 (6)',
                              type => doctest},
                           time => 8}}}
** When Server state == #{colors => [{enabled,true}],
                          excluded_counter => 0,failure_counter => 1,
                          invalid_counter => 0,seed => 757800,
                          skipped_counter => 0,slowest => 0,
[...]

PR #7444, opened 2018-03-11
Implement illegal_bin_pattern check in Elixir pass

lexmag:


lib/elixir/src/elixir_expand.erl
```diff
@@ -8,7 +8,13 @@
 expand({'=', Meta, [Left, Right]}, E) ->
   assert_no_guard_scope(Meta, "=", E),
   {ERight, ER} = expand(Right, E),
-  {ELeft, EL}  = elixir_clauses:match(fun expand/2, Left, E),
+  {ELeft, EL} = elixir_clauses:match(fun expand/2, Left, E),
+  if
+    element(1, ELeft) == '<<>>' ->
+      refute_parallel_bitstring_match(ERight, E, ?key(E, context) == match);
```
josevalim: Wouldn't it be better to pass the context downstream instead of true/false?

josevalim: Actually, nevermind. :)

josevalim: Btw, ELeft can be a list, it is not always guaranteed to be a tuple. And we could also have entries like this:

    {:<<>>, ...} = ...

So I think we need to pattern match on a tuple with three elements.

josevalim: I think you only need to call `refute_parallel_bitstring_match` only if the context is match. That's because if the context is not a match, you would have already called it when expanding the inner pattern. Your tests would be able to say if this is true or not. :)

lexmag: Yeah, I tested this already, it does not work.

lib/elixir/src/elixir_expand.erl
```diff
@@ -8,7 +8,13 @@
 expand({'=', Meta, [Left, Right]}, E) ->
   assert_no_guard_scope(Meta, "=", E),
   {ERight, ER} = expand(Right, E),
-  {ELeft, EL}  = elixir_clauses:match(fun expand/2, Left, E),
+  {ELeft, EL} = elixir_clauses:match(fun expand/2, Left, E),
+  if
+    element(1, ELeft) == '<<>>' ->
```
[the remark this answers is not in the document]
lexmag: Right, this place will be updated soon with (unfortunately) more complicated approach anyways.

lib/elixir/src/elixir_expand.erl
```diff
@@ -891,6 +892,29 @@ assert_generator_start(Meta, _, E) ->
...
+refute_parallel_bitstring_match(_Expr, _Arg, _E, _Parallel) ->
+  ok.
+
+refute_parallel_bitstring_match_list([], _Arg, _E, _Parallel) ->
+  ok;
+refute_parallel_bitstring_match_list(_Expr, [], _E, _Parallel) ->
+  ok;
+refute_parallel_bitstring_match_list([Arg1 | Rest1], [Arg2 | Rest2], E, Parallel) ->
+  refute_parallel_bitstring_match(Arg1, Arg2, E, Parallel),
+  refute_parallel_bitstring_match_list(Rest1, Rest2, E, Parallel).
```
whatyouhide: Does all of this belong here, or should we move it to `elixir_bitstring`? Just wondering.

lib/elixir/src/elixir_expand.erl
```diff
@@ -891,6 +892,29 @@ assert_generator_start(Meta, _, E) ->
...
+
+
```
whatyouhide: Not important at this stage but ✂️

josevalim:
Interestingly, this also fails:

```elixir
defmodule Foo do
  def bar do
    %{foo: <<foo>>} = %{foo: <<baz>>} = Foo.baz()
  end

  def baz do
    %{foo: <<0>>}
  end
end
```

But erl_lint does not say anything. Another pass does say "pattern cannot possibly match". I don't think we need to do anything about it, but I thought I would mention for completeness.

devonestes:
Is this work related to #7434, or something else? At first glance it sort of looks related, so I wanted to ask before I started looking at that issue.

josevalim:
@devonestes no, this is part of #5800. :)

lib/elixir/src/elixir_expand.erl
lexmag: @whatyouhide considering that `=` is what triggers the error I think it belongs here.

lexmag:
Updated.

lib/elixir/src/elixir_expand.erl
```diff
@@ -921,6 +922,26 @@ assert_generator_start(Meta, _, E) ->
...
 %% Assertions
 
+refute_parallel_bitstring_match({'<<>>', _, _}, {'<<>>', Meta, _} = Arg, E, true) ->
+  form_error(Meta, ?key(E, file), ?MODULE, {parallel_bitstring_match, Arg});
+refute_parallel_bitstring_match(Expr, {'=', _Meta, [Left, Right]}, E, Parallel) ->
+  refute_parallel_bitstring_match(Expr, Left, E, true),
+  refute_parallel_bitstring_match(Expr, Right, E, Parallel);
+refute_parallel_bitstring_match(Expr, Arg, E, Parallel) when is_list(Expr), is_list(Arg) ->
+  refute_parallel_bitstring_match_each(Expr, Arg, E, Parallel);
+refute_parallel_bitstring_match(Expr, Arg, E, Parallel) when tuple_size(Expr) == 2, tuple_size(Arg) == 2 ->
```
josevalim: this is a bit uncommon in the Elixir codebase. Can we pattern match on the tuple and build the lists explicitly?

    refute_parallel_bitstring_match({Expr1, Expr2}, {Arg1, Arg2}, E, Parallel) ->
       ...

lib/elixir/src/elixir_expand.erl
```diff
@@ -921,6 +922,26 @@ assert_generator_start(Meta, _, E) ->
 
 %% Assertions
 
+refute_parallel_bitstring_match({'<<>>', _, _}, {'<<>>', Meta, _} = Arg, E, true) ->
+  form_error(Meta, ?key(E, file), ?MODULE, {parallel_bitstring_match, Arg});
+refute_parallel_bitstring_match(Expr, {'=', _Meta, [Left, Right]}, E, Parallel) ->
+  refute_parallel_bitstring_match(Expr, Left, E, true),
+  refute_parallel_bitstring_match(Expr, Right, E, Parallel);
+refute_parallel_bitstring_match(Expr, Arg, E, Parallel) when is_list(Expr), is_list(Arg) ->
```
josevalim: We can also pattern match here:

    refute_parallel_bitstring_match([_ | _] = Expr, [_ | _] = Arg, E, Parallel) ->

Especially because we don't care about empty lists.

lib/elixir/src/elixir_expand.erl
lexmag: Yeah, I went with `tuple_size/1` to not deal with naming: it is not really `Expr1`, `Expr2`, `Arg1`, `Arg2` here.

josevalim: `{Left1, Left2}, {Right1, Right2}` is used on parallel traversals from time to time. But I think in avoiding the naming issue we are making the code less clear so we should find a name that you are at least comfortable with.

lexmag: Agree, `{Left1, Left2}, {Right1, Right2}` is exactly what I switched to.

lib/elixir/src/elixir_expand.erl
```diff
@@ -921,6 +922,28 @@ assert_generator_start(Meta, _, E) ->
...
+  refute_parallel_bitstring_match(Expr, Left, E, true),
+  refute_parallel_bitstring_match(Expr, Right, E, Parallel);
+refute_parallel_bitstring_match([_ | _] = Expr, [_ | _] = Arg, E, Parallel) ->
+  refute_parallel_bitstring_match_each(Expr, Arg, E, Parallel);
+refute_parallel_bitstring_match({Left1, Left2}, {Right1, Right2}, E, Parallel) ->
+  refute_parallel_bitstring_match_each([Left1, Left2], [Right1, Right2], E, Parallel);
+refute_parallel_bitstring_match({'{}', _, Args1}, {'{}', _, Args2}, E, Parallel) ->
+  refute_parallel_bitstring_match_each(Args1, Args2, E, Parallel);
+refute_parallel_bitstring_match({'%{}', _, Args1}, {'%{}', _, Args2}, E, Parallel) ->
+  refute_parallel_bitstring_match(Args1, Args2, E, Parallel);
```
josevalim: We should add a test for structs and also make sure they are checked. At this point they have a special notation so we need to check for `{%, _, [Name, Map]}`.

michalmuskala:
I don't think the implementation for maps is correct.
```elixir
%{foo: <<foo>>} = %{bar: <<bar>>} = x
```
should not fail.

lexmag:
@michalmuskala why do you think it should not fail?

lib/elixir/src/elixir_expand.erl
```diff
@@ -934,9 +934,11 @@ refute_parallel_bitstring_match({Left1, Left2}, {Right1, Right2}, E, Parallel) -
 refute_parallel_bitstring_match({'{}', _, Args1}, {'{}', _, Args2}, E, Parallel) ->
   refute_parallel_bitstring_match_each(Args1, Args2, E, Parallel);
 refute_parallel_bitstring_match({'%{}', _, Args1}, {'%{}', _, Args2}, E, Parallel) ->
-  refute_parallel_bitstring_match(Args1, Args2, E, Parallel);
-refute_parallel_bitstring_match({'%', _, [_, Args1]}, {'%', _, [_, Args2]}, E, Parallel) ->
-  refute_parallel_bitstring_match(Args1, Args2, E, Parallel);
+  refute_parallel_bitstring_match_map_field(lists:sort(Args1), lists:sort(Args2), E, Parallel);
```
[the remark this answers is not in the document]
josevalim: @michalmuskala we only get this fair when comparing two maps side by side. I think we are fine.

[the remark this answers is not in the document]
lexmag: Considering we sort only if we get maps inside parallel matching, perhaps keeping here is alright. 🤔

PR #7447, opened 2018-03-13
Fix typo in NaiveDateTime moduledoc.

jung-hunsoo:
There's a typo in NaiveDateTime moduledoc.

There's no more new/7 which is then replaced with new/8 since some previous version.


jung-hunsoo:
I just found that there're other PR with same issue. So I close this PR.

PR #7449, opened 2018-03-13
Remove invalid UTF-8 code from assertion failures

### LINKED ISSUE #7434: ex_unit crash with invalid utf8 in doctests

devonestes:
We were having issues with trying to write invalid UTF-8 sequences to
the console in the CLI formatter because if it was possible, we would
try and print the code that was failing in the failure message. This is
totally helpful, but we unfortunately can't print invalid UTF-8, so if
the code failed because it was invalid UTF-8, we can't show the user
that code.

We do still get the sequence that failed, and the line number that it
was on, so I'm hoping that this is still going to be enough for folks
to fix whatever issues they have.

Resolves #7434

### DIFF
--- a/lib/ex_unit/lib/ex_unit/doc_test.ex
+++ b/lib/ex_unit/lib/ex_unit/doc_test.ex
@@ -177,7 +177,8 @@ defmodule ExUnit.DocTest do
     * `:import` - when `true`, one can test a function defined in the module
       without referring to the module name. However, this is not feasible when
       there is a clash with a module like Kernel. In these cases, `:import`
-      should be set to `false` and a full `M.f` construct should be used.
+      should be set to `false` and a full `Module.function` construct should be
+      used.
 
   ## Examples
 
--- a/lib/ex_unit/lib/ex_unit/doc_test.ex
+++ b/lib/ex_unit/lib/ex_unit/doc_test.ex
@@ -408,13 +408,17 @@ defmodule ExUnit.DocTest do
     rescue
       e ->
         ex_message = "(#{inspect(e.__struct__)}) #{Exception.message(e)}"
+        message = "Doctest did not compile, got: #{ex_message}"
+
+        opts =
+          if String.valid?(expr) do
+            [message: message, expr: String.trim(expr)]
+          else
+            [message: message]
+          end
 
         quote do
-          message = "Doctest did not compile, got: #{unquote(ex_message)}"
-
-          reraise ExUnit.AssertionError,
-                  [message: message, expr: unquote(String.trim(expr))],
-                  unquote(stack)
+          reraise ExUnit.AssertionError, unquote(opts), unquote(stack)
         end
     end
   end
--- a/lib/ex_unit/lib/ex_unit/formatter.ex
+++ b/lib/ex_unit/lib/ex_unit/formatter.ex
@@ -110,8 +110,11 @@ defmodule ExUnit.Formatter do
   end
 
   @doc """
-  Receives a test and formats its failure.
+  Receives a test and formats its failure. The returned formatted failure must
+  always be a valid UTF-8 string.
   """
+  @spec format_test_failure(ExUnit.Test.t(), list, non_neg_integer, pos_integer, fun) ::
+          String.t()
   def format_test_failure(test, failures, counter, width, formatter) do
     %ExUnit.Test{name: name, module: module, tags: tags} = test
 
@@ -144,6 +147,7 @@ defmodule ExUnit.Formatter do
       left: left,
       right: right
     ]
+    |> filter_out_invalid_code()
     |> format_meta(formatter, label_padding_size)
     |> make_into_lines(counter_padding)
   end
@@ -226,6 +230,13 @@ defmodule ExUnit.Formatter do
     do: formatter.(:blame_diff, Macro.to_string(node))
 
   defp blame_match(_, string, _formatter), do: string
+
+  defp filter_out_invalid_code(fields) do
+    Enum.filter(fields, fn
+      {:code, code_example} -> String.valid?(code_example)
+      _ -> true
+    end)
+  end
 
   defp format_meta(fields, formatter, padding_size) do
     for {label, value} <- fields, has_value?(value) do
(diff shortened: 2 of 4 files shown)


josevalim:
Thanks @devonestes!

I am wondering if this is the best fix. Generally speaking, the `:code` that comes from regular tests cannot be invalid because otherwise the test would not run in the first place.

So to seems to be an issue exclusive to doctests and I wonder if we should tackle it there. Probably around [here](https://github.com/elixir-lang/elixir/blob/master/lib/ex_unit/lib/ex_unit/doc_test.ex#L412-L416) by skipping the `expr: ...` key if the underlying String is not valid?

devonestes:
Ok, that works for me.

My thinking was that since that `expr` binary isn't technically "invalid", it's just not valid UTF-8, and so we can't print it to the console. That's why I thought we should only remove those code examples that we are unable to print to the console while we're formatting these errors for printing to the console.

I'll tack on another commit and we can see how this other solution looks.

devonestes:
Yeah, this is a nice and simple solution. I like it 😄

lib/ex_unit/lib/ex_unit/doc_test.ex
```diff
@@ -411,9 +412,13 @@ defmodule ExUnit.DocTest do
         quote do
           message = "Doctest did not compile, got: #{unquote(ex_message)}"
 
-          reraise ExUnit.AssertionError,
-                  [message: message, expr: unquote(String.trim(expr))],
-                  unquote(stack)
+          if unquote(String.valid?(expr)) do
```
josevalim: You can do this check at compilation time. So move the conditional out of the quote and then do something like:

    reraise ExUnit.AssertionError, unquote(opts), unquote(stack)

Please let me know if you have any questions. :)

devonestes: Got it 😉

michalmuskala:
Thank you! :heart:

PR #7455, opened 2018-03-16
Added otp_release: 20.3 to travis config

h4cc:
Lets see if it works :)


josevalim:
It seem that not yet, unfortunately. We also tried yesterday. Maybe tomorrow! :D

PR #6326, opened 2017-07-13
Registry.unregister_match/2

benwilson512:
Right now in a duplicate registry using unregister on a key unregisters all key / value pairs. The idea with this function is that it would only unregister particular values that match a given spec.

The primary issue at the moment is that the `true = :ets.delete_object(pid_ets, {self, key, key_ets})` deletes all instances of the key under self, but instead we want N instances of the key where N is the number of key value pairs remaining in the key_ets table.

PR #7456, opened 2018-03-16
Add the ability to count the number of keys in a Registry

gmodarelli:
Hello everyone,

I'd like to propose a new feature for the `Registry` module. Namely the `count/1` function that returns the number of registered keys in a `Registry`.

It works for both `:unique` and `:duplicate` registries (and the code is the same for both :) )

Do I need to propose this feature in the Mailing List or is if fine to provide it as a PR?

Thanks for making Elixir such a great and approachable language ❤️

### DIFF
--- a/lib/elixir/lib/registry.ex
+++ b/lib/elixir/lib/registry.ex
@@ -1017,9 +1017,7 @@ defmodule Registry do
     case key_info!(registry) do
       {_kind, partitions, nil} ->
         Enum.reduce(0..(partitions - 1), 0, fn partition_index, acc ->
-          [{^partition_index, partition_ets, _}] = :ets.lookup(registry, partition_index)
-          size = safe_size(partition_ets)
-          acc + size
+          acc + safe_size(key_ets!(registry, partition_index))
         end)
 
       {_kind, 1, key_ets} ->
--- a/lib/elixir/lib/registry.ex
+++ b/lib/elixir/lib/registry.ex
@@ -1025,6 +1025,71 @@ defmodule Registry do
     end
   end
 
+  @doc """
+  Returns the number of `{pid, value}` pairs under the given `key` in `registry`
+  that match `pattern`.
+
+  Pattern must be an atom or a tuple that will match the structure of the
+  value stored in the registry. The atom `:_` can be used to ignore a given
+  value or tuple element, while :"$1" can be used to temporarily assign part
+  of pattern to a variable for a subsequent comparison.
+
+  It is possible to pass list of guard conditions for more precise matching.
+  Each guard is a tuple, which describes check that should be passed by assigned part of pattern.
+  For example :"$1" > 1 guard condition would be expressed as {:>, :"$1", 1} tuple.
+  Please note that guard conditions will work only for assigned variables like :"$1", :"$2", etc.
+  Avoid usage of special match variables :"$_" and :"$$", because it might not work as expected.
+
+  Zero will be returned if there is no match.
+
+  For unique registries, a single partition lookup is necessary. For
+  duplicate registries, all partitions must be looked up.
+
+  ## Examples
+
+  In the example below we register the current process under the same
+  key in a duplicate registry but with different values:
+
+      iex> Registry.start_link(keys: :duplicate, name: Registry.MatchTest)
+      iex> {:ok, _} = Registry.register(Registry.MatchTest, "hello", {1, :atom, 1})
+      iex> {:ok, _} = Registry.register(Registry.MatchTest, "hello", {2, :atom, 2})
+      iex> Registry.count_match(Registry.MatchTest, "hello", {1, :_, :_})
+      1
+      iex> Registry.count_match(Registry.MatchTest, "hello", {2, :_, :_})
+      1
+      iex> Registry.count_match(Registry.MatchTest, "hello", {:_, :atom, :_})
+      2
+      iex> Registry.count_match(Registry.MatchTest, "hello", {:"$1", :_, :"$1"})
+      2
+      iex> Registry.count_match(Registry.MatchTest, "hello", {:_, :_, :"$1"}, [{:>, :"$1", 1}])
+      1
+      iex> Registry.count_match(Registry.MatchTest, "hello", {:_, :"$1", :_}, [{:is_atom, :"$1"}])
+      2
+
+  """
+  @since "1.7.0"
+  @spec count_match(registry, key, match_pattern :: term, guards :: list()) :: non_neg_integer()
+  def count_match(registry, key, pattern, guards \\ [])
+      when is_atom(registry) and is_list(guards) do
+    guards = [{:"=:=", {:element, 1, :"$_"}, {:const, key}} | guards]
+    spec = [{{:_, {:_, pattern}}, guards, [true]}]
+
+    case key_info!(registry) do
+      {:unique, partitions, key_ets} ->
+        key_ets = key_ets || key_ets!(registry, key, partitions)
+        :ets.select_count(key_ets, spec)
+
+      {:duplicate, 1, key_ets} ->
+        :ets.select_count(key_ets, spec)
+
+      {:duplicate, partitions, _key_ets} ->
+        Enum.reduce(0..(partitions - 1), 0, fn partition_index, acc ->
+          count = :ets.select_count(key_ets!(registry, partition_index), spec)
+          acc + count
+        end)
+    end
+  end
+
   ## Helpers
 
   @compile {:inline, hash: 2}
--- a/lib/elixir/lib/registry.ex
+++ b/lib/elixir/lib/registry.ex
@@ -977,11 +977,11 @@ defmodule Registry do
   end
 
   @doc """
-  Returns the size of the Registry (ie The number of registered keys).
+  Returns the size of the registry, that is, the number of registered keys.
 
   ## Examples
   In the example below we register the current process and ask for the
-  size of the Registry:
+  size of the registry:
 
       iex> Registry.start_link(keys: :unique, name: Registry.UniqueCountTest)
       iex> Registry.size(Registry.UniqueCountTest)
--- a/lib/elixir/lib/registry.ex
+++ b/lib/elixir/lib/registry.ex
@@ -171,6 +171,15 @@ defmodule Registry do
   @typedoc "The type of registry metadata values"
   @type meta_value :: term
 
+  @typedoc "A pattern to match on objects in a registry"
+  @type match_pattern :: atom | term
+
+  @typedoc "A guard to be evaluated when matching on objects in a registry"
+  @type guard :: {atom | term}
+
+  @typedoc "A list of guards to be evaluated when matching on objects in a registry"
+  @type guards :: [guard] | []
+
   ## Via callbacks
 
   @doc false
@@ -544,14 +553,14 @@ defmodule Registry do
 
   Pattern must be an atom or a tuple that will match the structure of the
   value stored in the registry. The atom `:_` can be used to ignore a given
-  value or tuple element, while :"$1" can be used to temporarily assign part
+  value or tuple element, while the atom `:"$1"` can be used to temporarily assign part
   of pattern to a variable for a subsequent comparison.
 
-  It is possible to pass list of guard conditions for more precise matching.
-  Each guard is a tuple, which describes check that should be passed by assigned part of pattern.
-  For example :"$1" > 1 guard condition would be expressed as {:>, :"$1", 1} tuple.
-  Please note that guard conditions will work only for assigned variables like :"$1", :"$2", etc.
-  Avoid usage of special match variables :"$_" and :"$$", because it might not work as expected.
+  Optionally, it is possible to pass a list of guard conditions for more precise matching.
+  Each guard is a tuple, which describes checks that should be passed by assigned part of pattern.
+  For example the `$1 > 1` guard condition would be expressed as the `{:>, :"$1", 1}` tuple.
+  Please note that guard conditions will work only for assigned variables like `:"$1"`, `:"$2"`, etc.
+  Avoid usage of special match variables `:"$_"` and `:"$$"`, because it might not work as expected.
 
   An empty list will be returned if there is no match.
 
@@ -581,7 +590,7 @@ defmodule Registry do
 
   """
   @since "1.4.0"
-  @spec match(registry, key, match_pattern :: term, guards :: list()) :: [{pid, term}]
+  @spec match(registry, key, match_pattern, guards) :: [{pid, term}]
   def match(registry, key, pattern, guards \\ []) when is_atom(registry) and is_list(guards) do
     guards = [{:"=:=", {:element, 1, :"$_"}, {:const, key}} | guards]
     spec = [{{:_, {:_, pattern}}, guards, [{:element, 2, :"$_"}]}]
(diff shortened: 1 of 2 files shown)


lib/elixir/lib/registry.ex
```diff
@@ -976,6 +976,54 @@ defmodule Registry do
...
+        end)
+
+      {_kind, 1, key_ets} ->
+        safe_count(key_ets)
+    end
+  end
+
+  defp safe_count(ets) do
+    try do
+      :ets.info(ets, :size)
```
josevalim: Thank you for the PR!

I could not find documentation about the performance of this operation. I am assuming it happens in constant time, so maybe we should call it `size`.

However, we may also want to introduce `count_match`, which then is definitely linear, so maybe sticking with the `count` vocabulary is the safest choice?

Thoughts @elixir-lang/elixir?

gmodarelli: Looking at the Erlang source it looks like it is a constant time operation

https://github.com/erlang/otp/blob/2960c7d08932b718dc93773e0718cd214277e09d/erts/emulator/beam/erl_db.c#L3999

I'm ok with renaming it to `size`

lib/elixir/lib/registry.ex
```diff
@@ -976,6 +976,54 @@ defmodule Registry do
...
+
+      iex> Registry.start_link(keys: :duplicate, name: Registry.DuplicateCountTest)
+      iex> Registry.count(Registry.DuplicateCountTest)
+      0
+      iex> {:ok, _} = Registry.register(Registry.DuplicateCountTest, "hello", :world)
+      iex> {:ok, _} = Registry.register(Registry.DuplicateCountTest, "hello", :world)
+      iex> Registry.count(Registry.DuplicateCountTest)
+      2
+  """
+  @spec count(registry) :: integer()
```
eksperimental: Return value should be `non_neg_integer()`

gmodarelli: Thanks you

lib/elixir/lib/registry.ex
```diff
@@ -976,6 +976,54 @@ defmodule Registry do
...
+
+      iex> Registry.start_link(keys: :duplicate, name: Registry.DuplicateCountTest)
+      iex> Registry.count(Registry.DuplicateCountTest)
+      0
+      iex> {:ok, _} = Registry.register(Registry.DuplicateCountTest, "hello", :world)
+      iex> {:ok, _} = Registry.register(Registry.DuplicateCountTest, "hello", :world)
+      iex> Registry.count(Registry.DuplicateCountTest)
+      2
+  """
+  @spec count(registry) :: non_neg_integer()
```
eksperimental: Missing `@since "1.7.0"`

gmodarelli: Thanks again

josevalim:
@gmodarelli btw, we may still go with `count`. But it should be a matter of reverting a commit. :)

In a nutshell, because this operation happens in constant time, maybe we should call it `size`.

However, we may also want to introduce `count_match`, which then is definitely linear, so maybe sticking with the count vocabulary is the safest choice?

/cc @elixir-lang/elixir

gmodarelli:
@josevalim oh I see, I will revert the commit then :) 

I had also thought about the `count_match` but I still don't know how I'd implement it so I've left it out. Maybe I'll tackle it as soon as a use case presents to me.

I will wait for a decision on the name and then revert the commit (if necessary).

Cheers

josevalim:
@gmodarelli `count_match` would probably look like the code in #6326 except it would call another ets operation underneath. :)

josevalim:
@gmodarelli actually, scratch that. It is probably simpler to port the code in `Registry.match` since `unregister_match` has to do a bunch of work to not corrupt the ets tables.

lib/elixir/lib/registry.ex
```diff
@@ -977,47 +977,47 @@ defmodule Registry do
   end
 
   @doc """
-  Counts registered keys.
+  Returns the size of the Registry (ie The number of registered keys).
```
eksperimental: It should read "ie."

lib/elixir/lib/registry.ex
```diff
@@ -976,6 +976,55 @@ defmodule Registry do
     end
   end
 
+  @doc """
+  Returns the size of the Registry (ie The number of registered keys).
```
eksperimental: It should read:
Returns the size of the registry (ie. the number of registered keys).

lib/elixir/lib/registry.ex
```diff
@@ -976,6 +976,55 @@ defmodule Registry do
...
+
+  ## Examples
+  In the example below we register the current process and ask for the
+  size of the Registry:
```
eksperimental: "registry" should be in lowercase

lib/elixir/test/elixir/registry_test.exs
```diff
@@ -21,6 +21,17 @@ defmodule RegistryTest do
         assert length(Supervisor.which_children(registry)) == partitions
       end
 
+      test "returns 0 as the size of an empty Registry", %{registry: registry} do
```
eksperimental: registry in lowercase

lib/elixir/test/elixir/registry_test.exs
```diff
@@ -21,6 +21,17 @@ defmodule RegistryTest do
...
+        assert 0 == Registry.size(registry)
+      end
+
+      test "returns the size of the Registry", %{registry: registry} do
```
eksperimental: registry in lowercase

lib/elixir/lib/registry.ex
whatyouhide: No use of "ie" or similar. Please use "that is, ..." instead.

lib/elixir/lib/registry.ex
whatyouhide: No use of "ie" or similar. Please use "that is, ..." instead.

gmodarelli:
@eksperimental @whatyouhide I've fixed the docs and the tests ;)

@josevalim thanks for the pointers. I'll try to port the `match` code.

gmodarelli:
I gave the `count_match` function a try. Right now it's almost an exact copy of the `match` function. I've tried to extract some of the common code into separate helpers but it didn't look right to me. I've tried extracting the creation of the match specification but it made the code harder to read and understand.

Before I go on and try to refactor it I'd like to have your thoughts about its implementation.

Cheers

josevalim:
@gmodarelli beautiful! Let's revert `size` to `count` then and add a documentation note that says it runs in constant time and we should be good to go. Thank you!

gmodarelli:
@josevalim great, thanks for the feedback ;)

lib/elixir/lib/registry.ex
```diff
@@ -976,6 +976,120 @@ defmodule Registry do
...
+  that match `pattern`.
+
+  Pattern must be an atom or a tuple that will match the structure of the
+  value stored in the registry. The atom `:_` can be used to ignore a given
+  value or tuple element, while :"$1" can be used to temporarily assign part
+  of pattern to a variable for a subsequent comparison.
+
+  It is possible to pass list of guard conditions for more precise matching.
+  Each guard is a tuple, which describes check that should be passed by assigned part of pattern.
+  For example :"$1" > 1 guard condition would be expressed as {:>, :"$1", 1} tuple.
```
eksperimental: > For example: `$1 > 1`
Remove space before colon. Use backticks and remove quote and volon from the pseudo guard

lib/elixir/lib/registry.ex
```diff
@@ -976,6 +976,121 @@ defmodule Registry do
...
+  Returns the number of `{pid, value}` pairs under the given `key` in `registry`
+  that match `pattern`.
+
+  Pattern must be an atom or a tuple that will match the structure of the
+  value stored in the registry. The atom `:_` can be used to ignore a given
+  value or tuple element, while :"$1" can be used to temporarily assign part
+  of pattern to a variable for a subsequent comparison.
+
+  It is possible to pass list of guard conditions for more precise matching.
+  Each guard is a tuple, which describes check that should be passed by assigned part of pattern.
```
eksperimental: Which describes **checks**

lib/elixir/lib/registry.ex
```diff
@@ -976,6 +976,121 @@ defmodule Registry do
...
+    end
+  end
+
+  @doc """
+  Returns the number of `{pid, value}` pairs under the given `key` in `registry`
+  that match `pattern`.
+
+  Pattern must be an atom or a tuple that will match the structure of the
+  value stored in the registry. The atom `:_` can be used to ignore a given
+  value or tuple element, while :"$1" can be used to temporarily assign part
```
eksperimental: We need to add backticks everywhere there's code
> `:"$1"`

lib/elixir/lib/registry.ex
eksperimental: Add backticks around tuple

lib/elixir/lib/registry.ex
```diff
@@ -976,6 +976,121 @@ defmodule Registry do
...
+  of pattern to a variable for a subsequent comparison.
+
+  It is possible to pass list of guard conditions for more precise matching.
+  Each guard is a tuple, which describes check that should be passed by assigned part of pattern.
+  For example :"$1" > 1 guard condition would be expressed as {:>, :"$1", 1} tuple.
+  Please note that guard conditions will work only for assigned variables like :"$1", :"$2", etc.
+  Avoid usage of special match variables :"$_" and :"$$", because it might not work as expected.
```
eksperimental: Add backticks around these values as well

lib/elixir/lib/registry.ex
eksperimental: Actually we need to rephrase this line a little bit. It's not clear

lib/elixir/lib/registry.ex
eksperimental: We need rephrasing here a bit I think
> For example, the condition `$1 > 1` would be expressed as a the guard `{:>, :"$1", 1}`.

lib/elixir/lib/registry.ex
```diff
@@ -976,6 +976,121 @@ defmodule Registry do
...
+  @doc """
+  Returns the number of `{pid, value}` pairs under the given `key` in `registry`
+  that match `pattern`.
+
+  Pattern must be an atom or a tuple that will match the structure of the
+  value stored in the registry. The atom `:_` can be used to ignore a given
+  value or tuple element, while :"$1" can be used to temporarily assign part
+  of pattern to a variable for a subsequent comparison.
+
+  It is possible to pass list of guard conditions for more precise matching.
```
eksperimental: We need to mention that if no list is passed an empty list is used or maybe simplified by saying "Optionally, it is possible to pass a list"

eksperimental: "**a** list"

lib/elixir/lib/registry.ex
eksperimental: It would be good to be explicit here as well that `:"$1"` is an atom, to avoid users using a string
> while the atom `:"$1"` can be ...

lib/elixir/test/elixir/registry_test.exs
```diff
@@ -231,6 +266,17 @@ defmodule RegistryTest do
...
+        assert 0 == Registry.count(registry)
+      end
+
+      test "counts the number of keys in a registry", %{registry: registry} do
+        {:ok, _} = Registry.register(registry, "hello", :value)
+        {:ok, _} = Registry.register(registry, "hello", :value)
+
+        assert 2 == Registry.count(registry)
+      end
+
```
eksperimental: This set ot count tests is duplicated

lib/elixir/lib/registry.ex
```diff
@@ -976,6 +976,121 @@ defmodule Registry do
...
+      iex> Registry.count_match(Registry.MatchTest, "hello", {:"$1", :_, :"$1"})
+      2
+      iex> Registry.count_match(Registry.MatchTest, "hello", {:_, :_, :"$1"}, [{:>, :"$1", 1}])
+      1
+      iex> Registry.count_match(Registry.MatchTest, "hello", {:_, :"$1", :_}, [{:is_atom, :"$1"}])
+      2
+
+  """
+  @since "1.7.0"
+  @spec count_match(registry, key, match_pattern :: term, guards :: list()) :: non_neg_integer()
```
eksperimental: I think we can be more specific in this spec. Pattern can be atom or tuple, and maybe we can specify what kind of tuple. Guards can be an empty list or a list of tuples with a certain format

eksperimental:
Sorry if some of my reviews are not accurate or blunt. It's kind of hard to review code.from my cell phone..but there you go

gmodarelli:
@eksperimental no problem at all, I'll make sure to fix them all ;)

lib/elixir/test/elixir/registry_test.exs
gmodarelli: I've duplicated to test for both `:unique` and `:duplicate` registries

gmodarelli:
Ok, I've updated the documentation and added more type specifications, although I'm not sure if that's how you meant them to be @eksperimental 😄

lib/elixir/lib/registry.ex
```diff
@@ -171,6 +171,15 @@ defmodule Registry do
...
   @type meta_value :: term
 
+  @typedoc "A pattern to match on objects in a registry"
+  @type match_pattern :: atom | term
+
+  @typedoc "A guard to be evaluated when matching on objects in a registry"
+  @type guard :: {atom | term}
+
+  @typedoc "A list of guards to be evaluated when matching on objects in a registry"
+  @type guards :: [guard] | []
```
eksperimental: No need for the empty list as it is implied in `[guard]`

lib/elixir/lib/registry.ex
```diff
@@ -171,6 +171,15 @@ defmodule Registry do
   @typedoc "The type of registry metadata values"
   @type meta_value :: term
 
+  @typedoc "A pattern to match on objects in a registry"
+  @type match_pattern :: atom | term
+
```
eksperimental: Here it should be `atom | tuple`. But we could be more detailed if tuple follows a specific format

lib/elixir/lib/registry.ex
```diff
@@ -171,6 +171,15 @@ defmodule Registry do
...
+  @typedoc "A guard to be evaluated when matching on objects in a registry"
+  @type guard :: {atom | term}
```
eksperimental: This should be `tuple`. But we could be more detailed if tuple follows a specific format

lib/elixir/lib/registry.ex
fishcakez: The pattern can be any term.

eksperimental: Then the documentation should be corrected accordingly. Now it says:
> Pattern must be an atom or a tuple that will match the structure of the value stored in the registry.

lib/elixir/lib/registry.ex
eksperimental: It is a tuple now. But the spec expresses a tuple of one element, and that is wrong

lib/elixir/lib/registry.ex
gmodarelli: I think the documentation is wrong. Since the client code is what ultimately defines what goes inside the registry as value for a key, I should probably change the type to be just `term` and update the documentation. Thoughts?

lib/elixir/lib/registry.ex
gmodarelli: I'll change it to tuple. I cannot find a type specification for guards in the Erlang documentation. What I've found is this informal grammar specification for the Match Specification which includes guards http://erlang.org/doc/apps/erts/match_spec.html and I think it's way to complex to be reduced to a type (at least for me) :)

lib/elixir/lib/registry.ex
gmodarelli: I was thinking about updating the documentation to read:

> Pattern must be any term that will match the structure of the value stored in the registry.

lib/elixir/lib/registry.ex
```diff
@@ -976,6 +985,121 @@ defmodule Registry do
...
+      iex> Registry.count(Registry.DuplicateCountTest)
+      2
+  """
+  @since "1.7.0"
+  @spec count(registry) :: non_neg_integer()
+  def count(registry) when is_atom(registry) do
+    case key_info!(registry) do
+      {_kind, partitions, nil} ->
+        Enum.reduce(0..(partitions - 1), 0, fn partition_index, acc ->
+          [{^partition_index, partition_ets, _}] = :ets.lookup(registry, partition_index)
```
michalmuskala: This should use `:ets.lookup_element` - it's slightly more efficient than `lookup` when we're interested in just one element of the tuple and always expect the key to be there.

josevalim:
@michalmuskala good call. We can just use `key_ets!`. I have pushed a commit.

josevalim:

PR #7457, opened 2018-03-17
Add function to Mix.Tasks.Format that returns options to be used for a given file

JakeBecker:
Someone recently filed an issue on ElixirLS pointing out that it hasn't kept up with the available options in `.formatter.exs`: https://github.com/JakeBecker/elixir-ls/issues/67 .  I had been loading them from `.formatter.exs` manually, but now that it supports having different config files in subdirectories and importing from dependencies, adding that functionality would increase code duplication more than I would like. I'd also like to have a more future-proof way to do it. The Language Server Protocol has the server respond to the format request with the changed file contents rather than changing the file on-disk, so simply running `mix format` on the file isn't ideal.

I took a quick stab at adding a function that exports the options to be used for a particular file. It doesn't change any of the existing code in `Mix.Tasks.Format`, and I tacked an assertion onto an existing test. It's a bit of a bolt-on, but at least not much code.

I'm open to other ideas for how to handle this, so let me know if there's another approach you'd recommend or prefer.

### DIFF
--- a/lib/mix/lib/mix/tasks/format.ex
+++ b/lib/mix/lib/mix/tasks/format.ex
@@ -142,6 +142,19 @@ defmodule Mix.Tasks.Format do
     |> check!()
   end
 
+  @doc """
+  Returns formatter options to be used for the given file.
+  """
+  def formatter_opts_for_file(path, opts \\ []) do
+    {dot_formatter, formatter_opts} = eval_dot_formatter(opts)
+
+    {formatter_opts_and_subs, _sources} =
+      eval_deps_and_subdirectories(dot_formatter, [], formatter_opts, [dot_formatter])
+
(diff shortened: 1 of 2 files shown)


josevalim:
Thanks! Is there a reason though why you don’t call the format task
directly instead?
-- 


JakeBecker:
The current state of the document might not be saved to disk, so the Language Server Protocol specifies that the server's response should contain any edits to the document text, and then the client makes those changes. I didn't see an easy way to make that work with the task as-is.

josevalim:
I see, thanks. Implementation wise we can call
find_formatter_opts_for_file, let’s also make sure to add more assertions
in the test that handles subdirectories.
-- 


JakeBecker:
I updated the implementation and added assertions to a couple more tests.

michalmuskala:
Thank you! :heart:

PR #7459, opened 2018-03-17
Mix.Utils.stale?: Add support for empty target list

michaelkschmidt:
Mix.Utils.stale? will crash if the target list is empty.  This can
easily happen if Path.wildcard() is used to generate it.

### DIFF
--- a/lib/mix/lib/mix/utils.ex
+++ b/lib/mix/lib/mix/utils.ex
@@ -138,6 +138,10 @@ defmodule Mix.Utils do
     stale_stream(sources, targets) |> Enum.to_list()
   end
 
+  defp stale_stream(sources, []) do
+    sources
+  end
+
   defp stale_stream(sources, targets) do
     modified_target = targets |> Enum.map(&last_modified/1) |> Enum.min()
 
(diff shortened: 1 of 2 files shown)


michalmuskala:
Thank you! :heart:

---

One thing to note is that `Mix.Utils` module is private and should not be used outside of mix itself - it can change or even disappear at any moment without deprecations.

michaelkschmidt:
@michalmuskala is there a better way for testing if artifacts are stale from a custom Mix.Task?  In this particular case we just have a wrapper that kicks off Angular 2 and runs phx.digest if the files have changed.  It takes a fair amount of time so the stale test is really handy.

michalmuskala:
Yeah, there's a lot of useful things in there for writing compiler-like extensions. I've checked and the module is actually used both in phoenix and gettext. Maybe that's a sign we should actually expose this module?

\cc @josevalim @whatyouhide

whatyouhide:
@michalmuskala as far as I can tell, both in Gettext and in Phoenix only stale-related functionality is used. Gettext uses only `stale?` and Phoenix also uses `extract_stale`. For this reason I am 👎 on making `Mix.Utils` public. If we want to provide these functions, I think they should be properly exposed (in, say, `Mix.Stale` or `Mix.Manifest`).

michalmuskala:
You're right @whatyouhide that it might not make sense to open it all.

I think `stale?`, `extract_stale` and `extract_files` would be good candidates for exposing.

PR #7461, opened 2018-03-18
Explicitly list all `__info__` spec clauses, closes #7460

### LINKED ISSUE #7460: In OTP/20, dialyzer seems to complain about every module's __info__ when run with -Woverspecs.
I'm not sure if this is related to https://github.com/elixir-lang/elixir/issues/6536.

### Environment

  ```
  Erlang/OTP 20 [erts-9.3] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:10] [hipe] [kernel-poll:false] [dtrace]
  ```
* Operating system: MacOS 10.13

### Current behavior

```
$ cat lib/hello.ex
defmodule Hello do
end
$ grep dialyxir mix.exs
      {:dialyxir, "~> 0.4", only: [:dev]}
$ mix dialyzer --overspecs
...
lib/hello.ex:1: Type specification
  'Elixir.Hello':'__info__'('attributes' | 'compile' | 'functions' | 'macros' | 'md5' | 'module' | 'deprecated') -> atom() | [{atom(),any()} | {atom(),byte(),integer()}]
is a subtype of the success typing:
  'Elixir.Hello':'__info__'('attributes' | 'compile' | 'deprecated' | 'functions' | 'macros' | 'md5' | 'module') -> any()
```

### Expected behavior

Dialyzer not complaining about every module in the project when using `--overspecs` / `-Woverspecs`.

josevalim:


michalmuskala:
Would it make sense to have a type like `elixir:info()` and use it here instead of listing everything there? I don't think it makes much sense to grow the module just for those types.

josevalim:
@michalmuskala we can't use a type for a spec though, can we?

michalmuskala:
I still wonder if it's worth it adding this overhead to compilation time and module size. Maybe instead we should add an annotation for dialyzer to ignore this function?

I tried looking through dialyzer and hipe to figure out what dialyzer does with `module_info`, but I wasn't able to find any relevant references.

josevalim:
@michalmuskala is it really that much code though? I think the code we generate in `__info__` is going to be much more than this by any reasonable module.

The other idea is to generate a very poor type: `__info__(:module | :md5 | ...) -> term`.

michalmuskala:
It's just annoying me when I decompile modules trying to look at the compiler internals and have a page of this to scroll through 😛

josevalim:
@michalmuskala haha, other than that, is this the way to go then? What do @seancribbs / @fishcakez think?

fishcakez:
As a policy I don't think we should make changes to satisfy `overspecs` warnings as people should not be running dialyzer with this warning, and likely it creates a burden of maintenance for us to manage multiple OTP versions. It does in this case. Officially we only support the intersection of the OTP versions we support.

Over specification warnings are when we spec a subset of the possible spec found via success typing. If OTP adds something then our spec may not include it the new value - and a warning will be generated. However we can only support the intersection of types for all OTP versions we support.

However for `underspecs` we could make changes because under specification warnings are when we spec a superset of the possible spec found via success typing. In this case we have types outside the intersection of all types for all the OTP versions we support.

Therefore if we are to test anything it should be under specification warnings, not over specification. If we are happy to take on this maintenance burden at the module level to support the tools in the OTP release used then we should do `specdiffs` to find any difference, i.e. test the spec matches success typing exactly to prevent any warnings for any cases.

Outside the module level, i.e. for functions and types documented in Elixir itself, we can not support `overspecs` as we only support the intersection of all OTP versions supported. However we could support `underspecs` - though it will lead to some false positives that require dialyzer attributes to explicitly disable warnings. Therefore if `elixir:info` or similarly types were documented we could only support `underspecs`.

@josevalim I pushed a fix so we the test works correctly - though this means the tests now fail.

josevalim:
@fishcakez thanks. I agree with you.

But then we still have the question which approach we would like to take here. I would personally go with the more complete approach.

fishcakez:
> But then we still have the question which approach we would like to take here. I would personally go with the more complete approach.

At the module level I am happy to support `specdiffs`, as we want users to take as much advantage of tooling as possible.

josevalim:
@fishcakez when I enable specsdiff, then dialyzer fails with:

```elixir
Type specification 'Elixir.Dialyzer.RemoteCall':'__info__'('attributes') -> [{atom(),any()}]
         ; ('compile') -> [{atom(),any()}]
         ; ('deprecated') -> [{{atom(),arity()},binary()}]
         ; ('functions') -> [{atom(),arity()}]
         ; ('macros') -> [{atom(),arity()}]
         ; ('md5') -> binary()
         ; ('module') -> 'Elixir.Dialyzer.RemoteCall' is a subtype of the success typing: 'Elixir.Dialyzer.RemoteCall':'__info__'('attributes' | 'compile' | 'deprecated' | 'functions' | 'macros' | 'md5' | 'module') -> any()
```

I really don't understand why Dialyzer wants me to use a worst version of the typespec.

fishcakez:
@josevalim probably the number of possible values caused a union to occur that contained too many types, so is simplified to `any`.

PR #7462, opened 2018-03-18
Split Kernel.Typespec into Kernel.Typespec and Code.Typespec

josevalim:
Kernel.Typespec will be responsible for handling typespec
logic during compilation, such as definitions, conversion to
Erlang and warnings. Its API shall never be public.

Code.Typespec will handle the runtime aspects, such as reading
from .beam files and conversion back to Elixir AST. Its API
may be made public in the future but it is currently private.

This PR also simplifies definition of typespecs internally by
avoiding multiple macro passes. There is still work to be done
on how typespecs are stored to avoid inneficient lookup operations.


josevalim:
Note for reviewers: please review only `Code.Typespec`, the rest is simply code and tests being moved around.

lib/elixir/lib/code/typespec.ex
```diff
@@ -0,0 +1,430 @@
...
+            end
+        end
+
+      :error ->
+        :error
+    end
+  end
+
+  defp get_module_and_beam(module) when is_atom(module) do
+    case :code.get_object_code(module) do
```
michalmuskala: I guess this could use `:code.whereis`. I'm not sure it would be actually simpler, though.

lib/elixir/lib/code/typespec.ex
```diff
@@ -0,0 +1,430 @@
...
+  # We should then be able to simplify this code and use `with`.
+  defp typespecs_abstract_code(module) do
+    case get_module_and_beam(module) do
+      {module, binary} ->
+        case :beam_lib.chunks(binary, [:debug_info]) do
+          {:ok, {_, [debug_info: {:debug_info_v1, backend, data}]}} ->
+            case data do
+              {:elixir_v1, %{}, specs} ->
+                # Fast path to avoid translation to Erlang from Elixir.
+                {:ok, specs}
```
michalmuskala: If we don't translate to erlang here and `fetch_callbacks` is documented to return in erlang format, where is the conversion happening?

josevalim: Those specs are already in Erlang. Btw, we are totally relying on the internal of debug_info here, which is generally a no-no, but at least the debug_info is ours.

This provides a massive speed-up for Elixir on OTP 20 because previously we were converting the whole module to erlang only to get the same specs later on.

lib/elixir/lib/code/typespec.ex
```diff
@@ -0,0 +1,430 @@
...
+    {dot, [line: line], args}
+  end
+
+  defp erl_to_ex_var(var) do
+    case Atom.to_string(var) do
+      <<"_", c::binary-1, rest::binary>> ->
+        String.to_atom("_#{String.downcase(c)}#{rest}")
+
+      <<c::binary-1, rest::binary>> ->
+        String.to_atom("#{String.downcase(c)}#{rest}")
```
michalmuskala: Isn't the first char the `V` we artificially add to variables? If so, we could just remove it here.

This will probably also crash on just `_` as the var name.

lib/elixir/lib/code/typespec.ex
```diff
@@ -0,0 +1,430 @@
...
+    end
+  end
+
+  defp unpack_typespec_kw([{:type, _, :tuple, [{:atom, _, atom}, type]} | t], acc) do
+    unpack_typespec_kw(t, [{atom, typespec_to_quoted(type)} | acc])
+  end
```
michalmuskala: I would write this as:
```elixir
  defp unpack_typespec_kw(types, acc) do
    try do
      Enum.map(fn
        {:type, _, :tuple, [{:atom, _, atom}, type]} -> {atom, typespec_to_quoted(type)}
    unpack_typespec_kw(t, [{atom, typespec_to_quoted(type)} | acc])
  end
```

josevalim: There is no reason to optimize it here and I would prefer to avoid non-local flow control.

lib/elixir/lib/code/typespec.ex
josevalim: Those are typed variables, so it is a whole other story. This code was just ported over but we probably need to change it to consider utf-8 variables.

lib/elixir/lib/code/typespec.ex
josevalim: @michalmuskala actually, you proposed it for speed or because you think it reads better?

michalmuskala: When I set out to write it, I thought it would read better, but now looking at it, I'm not so sure any more.

PR #7464, opened 2018-03-18
Add hist iex helper

FedericoEsparza:
Add `hist/0` iex helper
 
Prints iex session history, up to configured history size
 
as a convenience for outputting the commands all at once (akin to `history` or `fc` in `bash` and `zsh`, respectively)

 
`iex> h`:
 

 
`iex> h hist`:

### DIFF
--- a/lib/iex/lib/iex/helpers.ex
+++ b/lib/iex/lib/iex/helpers.ex
@@ -367,6 +367,22 @@ defmodule IEx.Helpers do
     IEx.History.nth(history(), n) |> elem(2)
   end
 
+  def hist do
+    %IEx.History{queue: {history_list, first_item }} = history()
+    write_hist_item(first_item)
+
+    history_list
+    |> Enum.reverse()
+    |>Enum.map(&write_hist_item/1)
+  end
+
+  defp write_hist_item([{ 1, item, _ }]) do
+    IO.write "iex(1)> #{item}"
+  end
+  defp write_hist_item(hist_item) do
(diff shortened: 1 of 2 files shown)


josevalim:
Thank you!

A couple things: the queue is an opaque data structure, we cannot match in
its internal tuple like that. The same thing for the history data type. So
we should add a function to IEx.History that returns the entries that we
want, using also the :queue module.

Also, we no longer abbreviate functions in IEx.Helpers, so we should call
this history() and rename the existing history() to history_data().

Also, do you have an use case? We already get to navigate history with the
up and down errors, so is there something new we can achieve with this?
-- 


josevalim:
Finally it is better to return :ok rather than nil. :)
-- 


FedericoEsparza:
Thanks for the review, @josevalim.
 
I'll gladly make the outlined adjustments.
 
As for a unique use case for the feature, I'll convey it in a personalized User Story:
 
**As an** Elixir Programmer
**So that I** can easily assess, or retrieve and document my iex session's commands
**I would like to** output my recent commands with one short helper command

josevalim:
Thanks @FedericoEsparza. That's not enough as a use case though, I am afraid. We have other ways to access the history, although not as compact, is probably enough to handle most cases. The IEx.Helpers API is already quite large, so we do need strong indication that those features are required. 

Other than that, the PR looks great, both in documentation and implementation. I will close it until we have a stronger indication this feature can be useful, thanks.

PR #7467, opened 2018-03-19
Place parse_transofm warning with other attribute validation

michalmuskala:


josevalim:
Let's not merge this one because my plan is move this out altogether when we skip the erl_lint pass and if we change this now we will need to do those changes later as we will no longer need to pass the stack down.

PR #7472, opened 2018-03-21
Move accumulate attributes to a separate table

josevalim:
Right now we support the new API only for internal attributes.
External attributes are being added next.

### DIFF
--- a/lib/elixir/lib/kernel/typespec.ex
+++ b/lib/elixir/lib/kernel/typespec.ex
@@ -166,13 +166,12 @@ defmodule Kernel.Typespec do
     # TODO: Add and merge this information to doc metadata
     _ = get_since_info(set)
     _ = get_deprecated_info(set)
-
     :ets.insert(set, {{kind, {name, arity}}, line, doc})
   end
 
   defp get_doc_info(set, attr, line) do
     case :ets.take(set, attr) do
-      [{^attr, {line, doc}, _, _}] -> {line, doc}
+      [{^attr, {line, doc}, _}] -> {line, doc}
       [] -> {line, nil}
     end
   end
@@ -209,8 +208,8 @@ defmodule Kernel.Typespec do
     {types, specs, callbacks, macrocallbacks, optional_callbacks}
   end
 
-  defp take_typespec(set, key) do
-    :ets.take(set, key)
+  defp take_typespec(bag, key) do
+    :ets.take(bag, key)
   end
 
   defp translate_type({_, {kind, {:::, _, [{name, _, args}, definition]}, pos}})
--- a/lib/elixir/lib/kernel/typespec.ex
+++ b/lib/elixir/lib/kernel/typespec.ex
@@ -200,21 +200,13 @@ defmodule Kernel.Typespec do
   ## Translation from Elixir AST to typespec AST
 
   @doc false
-  def translate_typespecs_for_module(set, bag) do
+  def translate_typespecs_for_module(_set, bag) do
     types = Enum.map(take_typespec(bag, :type), &translate_type/1)
     specs = Enum.map(take_typespec(bag, :spec), &translate_spec/1)
     callbacks = Enum.map(take_typespec(bag, :callback), &translate_spec/1)
     macrocallbacks = Enum.map(take_typespec(bag, :macrocallback), &translate_spec/1)
-    optional_callbacks = List.flatten(take_optional(set, :optional_callbacks))
+    optional_callbacks = List.flatten(get_typespec(bag, {:accumulate, :optional_callbacks}))
     {types, specs, callbacks, macrocallbacks, optional_callbacks}
-  end
-
-  # TODO: Remove this once we store accumulate attributes in the bag table.
-  defp take_optional(table, key) do
-    case :ets.take(table, key) do
-      [{^key, value, _, _}] -> value
-      [] -> []
-    end
   end
 
   defp take_typespec(set, key) do
--- a/lib/elixir/lib/module.ex
+++ b/lib/elixir/lib/module.ex
@@ -1623,13 +1623,13 @@ defmodule Module do
     {set, bag} = data_tables_for(module)
 
     case :ets.lookup(set, key) do
-      [{^key, _, :accumulate}] ->
+      [{_, _, :accumulate}] ->
         :lists.reverse(bag_lookup_element(bag, {:accumulate, key}, 2))
 
-      [{^key, val, :read}] ->
+      [{_, val, :read}] ->
         val
 
-      [{^key, val, _}] ->
+      [{_, val, _}] ->
         :ets.update_element(set, key, {3, :read})
         val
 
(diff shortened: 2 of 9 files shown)


lib/elixir/lib/module.ex
```diff
@@ -1174,16 +1178,15 @@ defmodule Module do
...
     if Keyword.get(options, :persist) do
-      attributes = :ets.lookup_element(table, {:elixir, :persisted_attributes}, 2)
-      :ets.insert(table, {{:elixir, :persisted_attributes}, [attribute | attributes]})
+      :ets.insert(bag, {:persisted_attributes, attribute})
     end
 
     if Keyword.get(options, :accumulate) do
-      :ets.insert_new(table, {attribute, [], _accumulated? = true, _unread_line = nil}) ||
-        :ets.update_element(table, attribute, {3, true})
+      :ets.insert_new(set, {attribute, [], :accumulate}) ||
```
michalmuskala: What happens if there was already some value before accumulate was registered?
Is the value going to be simply ignored?

lib/elixir/src/elixir_erl.erl
```diff
@@ -450,10 +450,8 @@ validate_spec_for_existing_function(ModuleMap, NameAndArity, Line) ->
 % Attributes
 
 attributes_form(Line, Attributes, Forms) ->
-  Fun = fun({Key, Value}, Acc) ->
-    [{attribute, Line, Key, Value} | Acc]
-  end,
-  lists:foldl(Fun, Forms, Attributes).
+  Fun = fun({Key, Value}, Acc) -> [{attribute, Line, Key, Value} | Acc] end,
+  lists:foldr(Fun, Forms, Attributes).
```
michalmuskala: Why the change to `foldr`?

lib/elixir/lib/kernel/typespec.ex
```diff
@@ -200,25 +199,17 @@ defmodule Kernel.Typespec do
...
 
   @doc false
-  def translate_typespecs_for_module(set, bag) do
+  def translate_typespecs_for_module(_set, bag) do
     types = Enum.map(take_typespec(bag, :type), &translate_type/1)
     specs = Enum.map(take_typespec(bag, :spec), &translate_spec/1)
     callbacks = Enum.map(take_typespec(bag, :callback), &translate_spec/1)
     macrocallbacks = Enum.map(take_typespec(bag, :macrocallback), &translate_spec/1)
-    optional_callbacks = List.flatten(take_optional(set, :optional_callbacks))
+    optional_callbacks = List.flatten(get_typespec(bag, {:accumulate, :optional_callbacks}))
```
michalmuskala: Is there a reason why we delete some attributes from the table, but keep `optional_callbacks`?

lib/elixir/lib/module.ex
josevalim: Yes. It is simply ignored.

lib/elixir/src/elixir_erl.erl
josevalim: Because persisted attributes were reversed before due to the way we collected them but not anymore.

lib/elixir/lib/kernel/typespec.ex
josevalim: Because the other ones are internal but `@optional_callbacks` can be accessed by the user in a `after_compile` callback, so we can't remove it.

lib/elixir/src/elixir_erl.erl
michalmuskala: Wouldn't `map` be a better choice in that case?

lib/elixir/lib/kernel/typespec.ex
michalmuskala: I was more asking, if we need to delete the others - if not, we could just read them, they would be removed with the whole table.

lib/elixir/src/elixir_erl.erl
josevalim: We would change the attributes order in the final erlang abstract format, which I am not sure if it may have consequences or not.

lib/elixir/lib/module.ex
```diff
@@ -1669,16 +1669,20 @@ defmodule Module do
...
         :ets.insert(bag, {{:accumulate, key}, value})
 
-      [{^key, {line, <<_::binary>>}, _unread_line}]
+      [{_, {line, <<_::binary>>}, _unread_line}]
       when key in [:doc, :typedoc, :moduledoc] and is_list(stack) ->
         IO.warn("redefining @#{key} attribute previously set at line #{line}", stack)
         :ets.insert(set, {key, value, unread_line})
 
-      _ ->
+      [_] ->
```
michalmuskala: What do you think about doing a full lookup only if the key is one of those when we care? Otherwise just do a `lookup_element` on the third element? If the value is large, this should probably improve performance significantly - no need to copy it each time it's overridden.

lib/elixir/src/elixir_erl.erl
michalmuskala:
```
10> lists:foldr(fun (X, Acc) -> [X * 2 | Acc] end, [], [1, 2, 3]).
[2,4,6]
11> lists:map(fun (X) -> X * 2 end, [1, 2, 3]).
[2,4,6]
```

lib/elixir/lib/kernel/typespec.ex
josevalim: Could you see a benefit for not removing? I thought I would release the memory available early and deletes are not expensive (unless they generate more garbage but unlikely?).

lib/elixir/lib/module.ex
josevalim: I like this. I will play with it.

lib/elixir/src/elixir_erl.erl
josevalim: Oh, a `map` operation instead of the data type. It depends. We would need to concatenate the map though to the existing forms.

lib/elixir/src/elixir_module.erl
```diff
@@ -292,14 +295,12 @@ lookup_attribute(DataSet, DataBag, Key) when is_atom(Key) ->
...
-
-  [case ets:lookup(DataSet, Key) of
-     [{_, _, Line}] when is_integer(Line) ->
-       elixir_errors:form_warn([{line, Line}], File, ?MODULE, {unused_attribute, Key});
-     _ ->
-       ok
-   end || Key <- lists:usort(Attrs)].
+  StoredAttrs = bag_lookup_element(DataBag, attributes, 2),
+  Attrs = [doc, typedoc, impl, since, deprecated | StoredAttrs -- PersistedAttrs],
+  Query = [{{Attr, '_', '$1'}, [{is_integer, '$1'}], [[Attr, '$1']]} || Attr <- lists:usort(Attrs)],
```
michalmuskala: Do we need to usort it now? I'm not sure if `ets:select` return duplicated documents if one document matches multiple times.

josevalim: I checked and indeed it doesn't return dupes.

lib/elixir/src/elixir_erl.erl
michalmuskala: Oh, I completely missed that we have attributes passed already in and it's not an empty list. I'm sorry.

lib/elixir/lib/module.ex
```diff
@@ -1644,38 +1650,48 @@ defmodule Module do
...
 
-    case :ets.lookup(table, key) do
-      [{^key, {line, <<_::binary>>}, accumulated?, _unread_line}]
-      when key in [:doc, :typedoc, :moduledoc] and is_list(stack) ->
-        IO.warn("redefining @#{key} attribute previously set at line #{line}", stack)
-        :ets.insert(table, {key, value, accumulated?, unread_line})
-
-      [{^key, current, _accumulated? = true, _read?}] ->
-        :ets.insert(table, {key, [value | current], true, unread_line})
+    :ets.insert(set, {key, value, line})
```
michalmuskala: Is there a reason we don't insert those to the bag table as regular attributes? We wouldn't need to manually prepend them in the `elixir_module` in that case.

lib/elixir/lib/kernel/typespec.ex
michalmuskala: Deleting attributes might resize the table and needs to grab a write lock instead of just the read lock. I'm not sure that would be important in that case, since there's just one process accessing the table anyway.

josevalim:
:heart: :green_heart: :blue_heart: :y...

PR #7475, opened 2018-03-22
Improve documentation and specs for Logger.Formatter

whatyouhide:


lib/logger/lib/logger/formatter.ex
```diff
@@ -62,20 +62,28 @@ defmodule Logger.Formatter do
...
-  @doc ~S"""
-  Compiles a format string into a data structure that the `format/5` can handle.
+  @doc """
+  Compiles a format string into a data structure that `format/5` can handle.
+
+  Check the module doc for documentation on the valid parameters that
+  will be interpolated in the pattern. If you pass `nil` as the pattern,
+  the pattern defaults to:
+
+      #{inspect(@default_pattern)}
```
whatyouhide: Doing this requires us to drop the `~S` sigil and use `\\n` below, but I think it's worth it so that we don't repeat the pattern here.

lib/logger/lib/logger/formatter.ex
```diff
@@ -62,20 +62,28 @@ defmodule Logger.Formatter do
...
-  `{module, function}` to `compile/1` and the rest is handled.
+      iex> Logger.Formatter.compile({MyLoggerFormatter, :format})
+      {MyLoggerFormatter, :format}
 
-      iex> Logger.Formatter.compile("$time $metadata [$level] $message\n")
-      [:time, " ", :metadata, " [", :level, "] ", :message, "\n"]
   """
   @spec compile(binary | nil) :: [pattern | binary]
-  @spec compile({atom, atom}) :: {atom, atom}
+  @spec compile(pattern) :: pattern when pattern: {module, function :: atom}
```
whatyouhide: This *should* make dialyzer smarter here, right? :)

lib/logger/lib/logger/formatter.ex
```diff
@@ -94,19 +102,21 @@ defmodule Logger.Formatter do
   defp compile_code(key) when key in @valid_patterns, do: key
 
   defp compile_code(key) when is_atom(key) do
-    raise ArgumentError, "$#{key} is an invalid format pattern."
+    raise ArgumentError, "$#{key} is an invalid format pattern"
```
whatyouhide: We don't use trailing periods for error messages.

lib/logger/lib/logger/formatter.ex
```diff
@@ -94,19 +102,21 @@ defmodule Logger.Formatter do
...
   end
 
   @doc """
   Formats time as chardata.
   """
+  @spec format_time({0..23, 0..59, 0..59, 0..999}) :: IO.chardata()
```
whatyouhide: This (and the spec for `format_time/1` are not optimal IMO. We should extract them to a `time/0` type and a `date/0` type, and have `format/5` take a `{date(), time()}` type, but we would change the meaning of the `time/0` type. This would be a breaking change I guess, that's why I didn't do it.

michalmuskala: Hmm... Could the seconds be 60 in case of a leap second or is it ignored in this format like in posix?

whatyouhide: I just copied the type from `t:Logger.Formatter.time/0`, which is what is essentially broken down and passed to these two formatting functions.

lib/logger/lib/logger/formatter.ex
josevalim: Yes.

lib/logger/lib/logger/formatter.ex
```diff
@@ -62,20 +62,28 @@ defmodule Logger.Formatter do
...
 
-  Check the module doc for documentation on the valid parameters. If you
-  pass `nil`, it defaults to: `$time $metadata [$level] $levelpad$message\n`
+      iex> Logger.Formatter.compile("$time $metadata [$level] $message\\n")
+      [:time, " ", :metadata, " [", :level, "] ", :message, "\\n"]
 
-  If you would like to make your own custom formatter simply pass
-  `{module, function}` to `compile/1` and the rest is handled.
+      iex> Logger.Formatter.compile({MyLoggerFormatter, :format})
+      {MyLoggerFormatter, :format}
```
josevalim: Are we going to comment here that the formatter should never fail or do we do it elsewhere?

whatyouhide: We do it elsewhere.

PR #7476, opened 2018-03-22
Validate @compile attribute in elixir

### LINKED ISSUE #7409: Unused functions given to @compile {:inline, ...} trigger compilation error
We already need to validate inline as part of #5800. So we need to make sure we also remove unreachable entries.

michalmuskala:
* error on undefined inlined functions
* warn on parse transforms

Closes #7409 
Addresses another part of #5800

### DIFF
--- a/lib/elixir/src/elixir_module.erl
+++ b/lib/elixir/src/elixir_module.erl
@@ -130,23 +130,26 @@ validate_compile_opt({parse_transform, Module} = Opt, _Defs, _Unreachable, File,
   elixir_errors:form_warn([{line, Line}], File, ?MODULE, {parse_transform, Module}),
   [Opt];
 validate_compile_opt({inline, Inlines}, Defs, Unreachable, File, Line) ->
-  try [Inline || Inline <- Inlines, validate_inline(Inline, Defs, Unreachable)] of
-    [] -> [];
-    FilteredInlines -> [{inline, FilteredInlines}]
-  catch
-    {bad_inline, _} = Error ->
-      elixir_errors:form_error([{line, Line}], File, ?MODULE, Error)
+  case validate_inlines(Inlines, Defs, Unreachable, []) of
+    {ok, []} -> [];
+    {ok, FilteredInlines} -> [{inline, FilteredInlines}];
+    {error, Def} -> elixir_errors:form_error([{line, Line}], File, ?MODULE, {bad_inline, Def})
   end;
 validate_compile_opt(Opt, Defs, Unreachable, File, Line) when is_list(Opt) ->
   validate_compile_opts(Opt, Defs, Unreachable, File, Line);
 validate_compile_opt(Opt, _Defs, _Unreachable, _File, _Line) ->
   [Opt].
 
-validate_inline(Inline, Defs, Unreachable) ->
+validate_inlines([Inline | Inlines], Defs, Unreachable, Acc) ->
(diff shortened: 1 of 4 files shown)


lib/elixir/src/elixir_module.erl
```diff
@@ -85,7 +85,9 @@ compile(Line, Module, Block, Vars, E) ->
     (not elixir_config:get(bootstrap)) andalso
      'Elixir.Module':check_behaviours_and_impls(E, DataSet, DataBag, AllDefinitions, OverridablePairs),
 
-    CompileOpts = lists:flatten(bag_lookup_element(DataBag, {accumulate, compile}, 2)),
+    RawCompileOpts = bag_lookup_element(DataBag, {accumulate, compile}, 2),
+    DefinitionTuples = [Name || {Name, _, _, _} <- AllDefinitions],
```
josevalim: I would prefer to not do this here and instead use `lists:keyfind` when validating inline.

The rationale is that inline is used very rarely, so it doesn't make sense to always do this pass. Instead we do a lookup only when the inline attribute is used. keyfind is still in C, it should be fast enough.

lib/elixir/src/elixir_module.erl
```diff
@@ -121,6 +123,24 @@ compile(Line, Module, Block, Vars, E) ->
...
+validate_compile_opts([{inline, Inlines} | Rest], Defs, Unreachable, File) ->
+  case Inlines -- Defs of
+    [] ->
+      [{inline, Inlines -- Unreachable} | validate_compile_opts(Rest, Defs, Unreachable, File)];
+    [{Name, Arity} | _] ->
+      elixir_errors:form_error([], File, ?MODULE, {bad_inline, Name, Arity})
+  end;
+%% TODO: Make this an error and skip parse transform processing on 2.0.
+validate_compile_opts([{parse_transform, Module} = Opt | Rest], Defs, Unreachable, File) ->
+  elixir_errors:form_warn([], File, ?MODULE, {parse_transform, Module}),
```
josevalim: Let's include the line as first argument: `[{line, Line}]`. You will need to pass the `Line` all the way in.

josevalim:
Beautiful! Two comments and that's it!

lib/elixir/src/elixir_module.erl
```diff
@@ -121,6 +122,32 @@ compile(Line, Module, Block, Vars, E) ->
...
+      elixir_errors:form_error([{line, Line}], File, ?MODULE, Error)
+  end;
+validate_compile_opt(Opt, Defs, Unreachable, File, Line) when is_list(Opt) ->
+  validate_compile_opts(Opt, Defs, Unreachable, File, Line);
+validate_compile_opt(Opt, _Defs, _Unreachable, _File, _Line) ->
+  [Opt].
+
+validate_inline(Inline, Defs, Unreachable) ->
+  case lists:keyfind(Inline, 1, Defs) of
+    false -> throw({bad_inline, Inline});
```
josevalim: Let's please not this style. :) We don't use it anywhere in the codebase and I personally think we should avoid non-local returns. If you think we should start using this, then let's have a discussion, but not in this PR.

If you want to do a single pass, then let's do recursion with `{ok, Var} | {error, Inline}` return.

michalmuskala: I'm pretty sure that style is a result of spending too much time in OTP internals 😅

PR #7477, opened 2018-03-22
Pass a function StringIO.open to allow the device to be closed automatically

easco:
Similar to the way you can pass a function to `File.open` this change allows you to pass a function to `StringIO.open`.  The function will receive the `StringIO` device as a parameter and when the function completes the device will be closed automatically.  The function may be passed in lieu of options to `StringIO.open/2` and a new `StringIO.open/3` function is available should you wish to pass both options and a function.

Examples are included in doc tests.

Please offer feedback or suggestions.  Reviewers should pay particular attention to `@spec` declarations as they are an area of weakness for me.

### DIFF
--- a/lib/elixir/lib/string_io.ex
+++ b/lib/elixir/lib/string_io.ex
@@ -49,16 +49,12 @@ defmodule StringIO do
   @spec open(binary, keyword, (pid -> res)) :: {:ok, res} when res: var
   def open(string, options, function)
       when is_binary(string) and is_list(options) and is_function(function, 1) do
-    case GenServer.start_link(__MODULE__, {string, options}, []) do
-      {:ok, pid} ->
-        try do
-          {:ok, function.(pid)}
-        after
-          {:ok, {_input, _output}} = close(pid)
-        end
+    {:ok, pid} = GenServer.start_link(__MODULE__, {string, options}, [])
 
-      error ->
-        error
+    try do
+      {:ok, function.(pid)}
+    after
+      {:ok, {_input, _output}} = close(pid)
     end
   end
 
--- a/lib/elixir/lib/string_io.ex
+++ b/lib/elixir/lib/string_io.ex
@@ -23,7 +23,58 @@ defmodule StringIO do
 
   If the `:capture_prompt` option is set to `true`,
   prompts (specified as arguments to `IO.get*` functions)
-  are captured.
+  are captured in the output.
+
+  The device will be created and sent to the function given.
+  When the function returns, the device will be closed. The final
+  result will be a tuple with `:ok` and the result of the function.
+
+  ## Examples
+      iex> {:ok, _contents} = StringIO.open("foo", [], fn(pid) ->
+      ...>     input = IO.gets(pid, ">")
+      ...>     IO.write(pid, "The input was \#{input}")
+      ...>     StringIO.contents(pid)
+      ...>   end)
+      {:ok, {"", "The input was foo"}}
+
+      iex> {:ok, _contents} = StringIO.open("foo", [capture_prompt: true], fn(pid) ->
+      ...>     input = IO.gets(pid, ">")
+      ...>     IO.write(pid, "The input was \#{input}")
+      ...>     StringIO.contents(pid)
+      ...>   end)
+      {:ok, {"", ">The input was foo"}}
+  """
+  @spec open(binary, keyword, (pid -> res)) :: {:ok, res} when res: var
+  def open(string, options, function)
+      when is_binary(string) and is_list(options) and is_function(function, 1) do
+    with {:ok, pid} <- GenServer.start_link(__MODULE__, {string, options}, []) do
+      try do
+        {:ok, function.(pid)}
+      after
+        {:ok, {_input, _output}} = close(pid)
+      end
+    else
+      error -> error
+    end
+  end
+
+  @doc """
+  Creates an IO device.
+
+  `string` will be the initial input of the newly created
+  device.
+
+  `options_or_function` can be a keyword list of options or
+  a function.
+
+  If options are provided, the result will be `{:ok, pid}`, returning the
+  IO device created. The option `:capture_prompt`, when set to `true`, causes
+  prompts (which are specified as arguments to `IO.get*` functions) to be
+  included in the device's output.
+
+  If a function is provided, the device will be created and sent to the
+  function. When the function returns, the device will be closed. The final
+  result will be a tuple with `:ok` and the result of the function.
 
   ## Examples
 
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/string_io.ex
```diff
@@ -23,7 +23,58 @@ defmodule StringIO do
...
+
+  The device will be created and sent to the function given.
+  When the function returns, the device will be closed. The final
+  result will be a tuple with `:ok` and the result of the function.
+
+  ## Examples
+      iex> {:ok, _contents} = StringIO.open("foo", [], fn(pid) ->
+      ...>     input = IO.gets(pid, ">")
+      ...>     IO.write(pid, "The input was \#{input}")
+      ...>     StringIO.contents(pid)
```
josevalim: Everything here and the example below is indented with one extra space. :)

lib/elixir/lib/string_io.ex
```diff
@@ -23,7 +23,58 @@ defmodule StringIO do
...
+      ...>   end)
+      {:ok, {"", ">The input was foo"}}
+  """
+  @spec open(binary, keyword, (pid -> res)) :: {:ok, res} when res: var
+  def open(string, options, function)
+      when is_binary(string) and is_list(options) and is_function(function, 1) do
+    with {:ok, pid} <- GenServer.start_link(__MODULE__, {string, options}, []) do
```
josevalim: Given it is a single clause, I think `case` would be clearer here.

josevalim:
@easco two comments and this looks good to go!

lib/elixir/lib/string_io.ex
easco: Oh noes! 🤪

lib/elixir/lib/string_io.ex
easco: OK I think I see what you mean.  I will change the "with" to a case

lib/elixir/lib/string_io.ex
```diff
@@ -96,6 +100,7 @@ defmodule StringIO do
       ...>     StringIO.contents(pid)
       ...>   end)
       {:ok, {"", "The input was foo"}}
+
```
easco: Extra line added to make this consistent with other examples in the file

lib/elixir/lib/string_io.ex
```diff
@@ -23,7 +23,62 @@ defmodule StringIO do
...
+    iex> {:ok, _contents} = StringIO.open("foo", [capture_prompt: true], fn(pid) ->
+    ...>    input = IO.gets(pid, ">")
+    ...>    IO.write(pid, "The input was \#{input}")
+    ...>    StringIO.contents(pid)
+    ...> end)
+    {:ok, {"", ">The input was foo"}}
+
+  """
+  @spec open(binary, keyword, (pid -> res)) :: {:ok, res} when res: var
```
josevalim: The spec says it returns `{:ok, res}` but the code says it may return the error from GenServer.start_link. Maybe we also don't need a case and instead we should do `{:ok, pid} = GenServer...`?

easco: I patterned my implementation after the implementation of `File.open` (https://github.com/elixir-lang/elixir/blob/master/lib/elixir/lib/file.ex#L1297) which does use a case statement.  The spec there also calls out the fact that the result may be `{:error, posix}`.  
If you agree that matching the behavior of that function is the right approach, I can revise the `@spec`.

The alternative, I suppose, is for the function to raise a match error if the `GenServer` can't start up.  That would make this an `open!` function, wouldn't it?

josevalim: We would use `{:ok, pid}` because there is no chance it can fail. If it can fail, then we should return `{:error, _}` and update the `@spec` but i don't think it can ever happen. Hence pattern matching. :)

easco: Ok.  I'll remove the case and add the pattern match.  Thanks!

easco: In a completely unrelated note, I also noticed some tests fail (on Mac OS X anyway) if there is a space in the path leading up to where you've cloned the elixir repo.

lib/elixir/lib/string_io.ex
```diff
@@ -23,7 +23,58 @@ defmodule StringIO do
...
-  are captured.
+  are captured in the output.
+
+  The device will be created and sent to the function given.
+  When the function returns, the device will be closed. The final
+  result will be a tuple with `:ok` and the result of the function.
+
+  ## Examples
+
+    iex> {:ok, _contents} = StringIO.open("foo", [], fn(pid) ->
```
whatyouhide: We don't need to match on `{:ok, _contents}` here since we show the output of the function below.

lib/elixir/lib/string_io.ex
```diff
@@ -23,7 +23,58 @@ defmodule StringIO do
...
+    ...>    input = IO.gets(pid, ">")
+    ...>    IO.write(pid, "The input was \#{input}")
+    ...>    StringIO.contents(pid)
+    ...> end)
+    {:ok, {"", "The input was foo"}}
+
+    iex> {:ok, _contents} = StringIO.open("foo", [capture_prompt: true], fn(pid) ->
```
whatyouhide: We don't need to match on `{:ok, _contents}` here since we show the output of the function below.

lib/elixir/lib/string_io.ex
```diff
@@ -39,10 +90,25 @@ defmodule StringIO do
       iex> StringIO.contents(pid)
       {"", ">"}
 
+      iex> {:ok, _contents} = StringIO.open("foo", fn(pid) ->
```
whatyouhide: We don't need to match on `{:ok, _contents}` here since we show the output of the function below.

lib/elixir/lib/string_io.ex
```diff
@@ -39,10 +90,25 @@ defmodule StringIO do
...
+      ...>     input = IO.gets(pid, ">")
+      ...>     IO.write(pid, "The input was \#{input}")
+      ...>     StringIO.contents(pid)
+      ...>   end)
```
whatyouhide: The body of the function is indented incorrectly.

lib/elixir/lib/string_io.ex
```diff
@@ -23,7 +23,58 @@ defmodule StringIO do
...
+
+  The device will be created and sent to the function given.
+  When the function returns, the device will be closed. The final
+  result will be a tuple with `:ok` and the result of the function.
+
+  ## Examples
+
+    iex> {:ok, _contents} = StringIO.open("foo", [], fn(pid) ->
+    ...>    input = IO.gets(pid, ">")
+    ...>    IO.write(pid, "The input was \#{input}")
```
whatyouhide: You can use `~S"""` to avoid having to escape things.

easco: OK.  I changed that in both places I believe.

lib/elixir/lib/string_io.ex
easco: I removed all of them.

josevalim:

Issue #7474, opened 2018-03-22
Ugly errors are raised when logging non-loggable stuff


whatyouhide:
I was going down the rabbit hole of `Logger` formatters, and I found a few interesting behaviours.

Let's start by logging something that doesn't look like a binary or a charlist. If we do this, we get a nice (and expected) `Protocol.UndefinedError` for `String.Chars`:

```
iex> Logger.info(%{})
** (Protocol.UndefinedError) protocol String.Chars not implemented for %{}
```

This is good™. Now, what actually calls `to_string/1` is `Logger.truncate/2`. However, `truncate/2` treats binaries and *what look like charlists* in a special way:

https://github.com/elixir-lang/elixir/blob/e703bb89f91dc51b731df330c7b35125b3b42488/lib/logger/lib/logger.ex#L771-L772

The problem is that `is_list/1` definitely doesn't tell if something is a charlist all the way down. Hence, it's easy to make `Logger` raise a (IMO) nasty `FunctionClauseError`:

```
iex> Logger.info([%{}])
** (FunctionClauseError) no function clause matching in Logger.Utils.truncate_n/2
```

Things get even worse when I trick Logger into bypassing `truncate/2` altogether, that is, by configuring `:truncate` to be `:infinity`. In this case, if I pass a list, Logger still mistakes it for chardata and either gives it to a formatter (like the default one in the console backend) that crashes on it:

```
iex> Logger.configure(truncate: :infinity)
iex> Logger.info([%{}])
02:06:14.509 [error] GenEvent handler Logger.Backends.Console installed in Logger terminating
** (ArgumentError) argument error
    (stdlib) :unicode.characters_to_binary([[[] | "\e[22m"], ["\n", [[48, "2"], 58, [48, "6"], 58, "14", 46, "502"], " ", "", "[", "info", "] ", " ", [%{}], "\n"] | "\e[0m"])
```

or passes it down to custom formatters (like `{module, function}`) which will receive a message that they cannot turn to chardata straight away.

This is somehow of a problem with metadata as well, as they're only tried to be turned into a string by each specific formatter. However, with metadata I guess that this is good because smarted backends or formatters can use non-string-friendly metadata for their own specialized purposes. However, it feels like we should ensure formatters and backends that the message they will get is a binary or chardata. Thoughts?

[...]

PR #7478, opened 2018-03-22
Handle bad chardata lookalikes in Logger

whatyouhide:
Related to #7474.

### DIFF
--- a/lib/logger/lib/logger.ex
+++ b/lib/logger/lib/logger.ex
@@ -275,9 +275,9 @@ defmodule Logger do
   The console backend allows you to customize the format of your log messages
   with the `:format` option.
 
-  You may set `:format` to either a string or a `{module, function}` tuple if you wish to provide
-  your own format function. Here is an example of how to configure the `:console` backend in a
-  `config/config.exs` file:
+  You may set `:format` to either a string or a `{module, function}` tuple if
+  you wish to provide your own format function. Here is an example of how to
+  configure the `:console` backend in a `config/config.exs` file:
 
       config :logger, :console,
         format: {MyConsoleLogger, :format}
@@ -307,9 +307,10 @@ defmodule Logger do
   The `{module, function}` will be invoked with four arguments:
 
     * the log level: an atom
-    * the message: this is usually chardata, but in some cases it may not be. Since the formatting
-      function should *never* fail, you need to prepare for the message being anything (and do
-      something like the `rescue` in the example above)
+    * the message: this is usually chardata, but in some cases it may not be.
+      Since the formatting function should *never* fail, you need to prepare for
+      the message being anything (and do something like the `rescue` in the example
+      above)
     * the current timestamp: a term of type `t:Logger.Formatter.time/0`
     * the medatata: a keyword list
 
--- a/lib/logger/lib/logger.ex
+++ b/lib/logger/lib/logger.ex
@@ -275,12 +275,8 @@ defmodule Logger do
   The console backend allows you to customize the format of your log messages
   with the `:format` option.
 
-  You may set `:format` to either a string or a `{module, function}` tuple if
-  you wish to provide your own format function. The `{module, function}` will be
-  invoked with the log level, the message, the current timestamp and the
-  metadata.
-
-  Here is an example of how to configure the `:console` backend in a
+  You may set `:format` to either a string or a `{module, function}` tuple if you wish to provide
+  your own format function. Here is an example of how to configure the `:console` backend in a
   `config/config.exs` file:
 
       config :logger, :console,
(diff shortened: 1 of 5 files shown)


lib/logger/lib/logger/utils.ex
```diff
@@ -74,6 +74,12 @@ defmodule Logger.Utils do
     truncate_n_list(list, n, [])
   end
 
+  defp truncate_n(other, _n) do
+    raise ArgumentError,
```
whatyouhide: I think this is enough. An alternative could be to call `List.to_string/1` on the whole list here in order to get an even nicer error, but I think this is fine.

lib/logger/lib/logger/backends/console.ex
```diff
@@ -236,7 +236,13 @@ defmodule Logger.Backends.Console do
   defp retry_log(error, %{device: device, ref: ref, output: dirty} = state) do
     Process.demonitor(ref, [:flush])
 
-    case :unicode.characters_to_binary(dirty) do
+    try do
+      :unicode.characters_to_binary(dirty)
+    rescue
+      ArgumentError ->
+        clean = ["failure while trying to output malformed data: ", inspect(dirty), ?\n]
+        %{state | ref: async_io(device, clean), output: clean}
```
whatyouhide: I am definitely not sure about this, and this is why I didn't add tests yet (would prefer to get feedback on the approach first). One thing that bugs me is that if we get here, the log will be weird in the console:

```
iex(1)> Logger.configure(truncate: :infinity)
:ok
iex(2)> Logger.bare_log(:info, [%{}])
:ok
iex(3)> failure while trying to output malformed data: [[[] | "\e[22m"], ["\n", [[48, "0"], 58, "40", 58, "57", 46, "292"], " ", "", "[", "info", "] ", " ", [%{}], "\n"] | "\e[0m"]
nil
```

Thoughts?

lib/logger/lib/logger/backends/console.ex
```diff
@@ -236,7 +236,13 @@ defmodule Logger.Backends.Console do
   defp retry_log(error, %{device: device, ref: ref, output: dirty} = state) do
     Process.demonitor(ref, [:flush])
 
-    case :unicode.characters_to_binary(dirty) do
+    try do
+      :unicode.characters_to_binary(dirty)
+    rescue
+      ArgumentError ->
+        clean = ["failure while trying to output malformed data: ", inspect(dirty), ?\n]
```
josevalim: "while trying to **log** malformed data"?

lib/logger/lib/logger/backends/console.ex
josevalim: Something else we could do is to perform an asynchronous call to the logger so it is processed as a regular log message.

lib/logger/lib/logger/utils.ex
```diff
@@ -74,6 +74,12 @@ defmodule Logger.Utils do
...
+          "cannot truncate chardata because it contains something that is not " <>
+            "valid chardata: #{inspect(other)}"
```
josevalim: I believe Logger works with iodata?

whatyouhide: @josevalim from these lines: https://github.com/elixir-lang/elixir/blob/5947ed751eb0276cb7fda666c319dba451a5e9e7/lib/logger/lib/logger/utils.ex#L68-L71

it looks to me like we're using chardata.

lib/logger/lib/logger/backends/console.ex
whatyouhide: @josevalim but then it would be broadcast to all backends, while there is a problem only with the console backend. Is it fine to log this to all backends?

lib/logger/lib/logger/utils.ex
josevalim: You are correct, apologies.

lib/logger/lib/logger.ex
```diff
@@ -275,12 +275,8 @@ defmodule Logger do
...
   with the `:format` option.
 
-  You may set `:format` to either a string or a `{module, function}` tuple if
-  you wish to provide your own format function. The `{module, function}` will be
-  invoked with the log level, the message, the current timestamp and the
-  metadata.
-
-  Here is an example of how to configure the `:console` backend in a
+  You may set `:format` to either a string or a `{module, function}` tuple if you wish to provide
+  your own format function. Here is an example of how to configure the `:console` backend in a
```
josevalim: Cosmetic: should we keep around the same line length as previous paragraphs?

lib/logger/lib/logger/utils.ex
josevalim: This is fine by me, yeah.

lib/logger/lib/logger.ex
whatyouhide: Yes, fixing it.

michalmuskala:
Thank you! :heart:

Issue #4586, opened 2016-05-05
Make mix test stop on nth failure


ibizaman:
Is it possible to add a flag asking mix test to stop on nth failure?

This is part of my test repairing workflow when using pytest, a python test executor.
I run it like this: [`pytest --maxfail=1`](https://pytest.org/latest/usage.html#stopping-after-the-first-or-n-failures), it stops on first error, I repair that error (or try at least), re-run the pytest command and repeat until there are errors.
I would use this only for manually repairing tests, not when running tests in a continuous integration fashion.

It would be awesome to have something similar in mix test task.

whatyouhide:
Is there value to making the maximum number of tests allowed to fail configurable instead of just using something like "stop on the first failing test"?

ibizaman:
I never set maxfail to something >1 so not for me at least.

eksperimental:
I guess >1 could be useful when introducing a breaking change in a project, that requires to modifiy a hell of a lot of tests.. so you want to see/fix them by chunks of n's

josevalim:
Why would such feature be required given we already print failures as they come? I would understand the importance of this feature if you had to wait for the whole suite to finish but you can already start acting as soon as the first failures show up.

eksperimental:
Because tests can be long, specially if you are on an slow computer, and you may want to be doing something else (concurrency in real life :-) rather than looking at the computer screen waiting for it to fail.
So maybe do something like:
`mix test --max-fail 3 && play_sound_sucess || play_sound_error`

I'm not saying it is required feature, I'm just giving a possible case scenario

josevalim:
I see. If someone would like to submit a pull request, we would likely accept it, but for us it is low priority. :)

ibizaman:
@josevalim to be honest, I wanted this because I didn't know at the time about the log capture option. I wanted this because in my case I have a 7000 line test debug output (all SQL requests are logged) when running all tests and I spend a stupid amount of time finding the printed failures. But as I said, log capture solves this!

PR #7480, opened 2018-03-23
Mix --max-fail opt

Kartstig:
- Relates to https://github.com/elixir-lang/elixir/issues/4586

### DIFF
--- a/lib/mix/lib/mix/tasks/test.ex
+++ b/lib/mix/lib/mix/tasks/test.ex
@@ -58,6 +58,7 @@ defmodule Mix.Tasks.Test do
     * `--listen-on-stdin` - runs tests, and then listens on stdin. Receiving a newline will
       result in the tests being run again. Very useful when combined with `--stale` and
       external commands which produce output on stdout upon file system modification.
+    * `--max-fail` - sets the maximum number of tests to allow to fail before stopping
     * `--max-cases` - sets the maximum number of tests running async. Only tests from
       different modules run in parallel. Defaults to twice the number of cores.
     * `--no-archives-check` - does not check archives
@@ -177,6 +178,7 @@ defmodule Mix.Tasks.Test do
     color: :boolean,
     cover: :boolean,
     trace: :boolean,
+    max_fail: :integer,
     max_cases: :integer,
     include: :keep,
     exclude: :keep,
@@ -277,6 +279,9 @@ defmodule Mix.Tasks.Test do
         option_only_present? = Keyword.has_key?(opts, :only)
 
         cond do
+          failures >= opts[:max_fail] ->
+            Mix.raise("Maximum number of test failures reached: #{failures}")
+
           failures > 0 and opts[:raise] ->
             Mix.raise("mix test failed")
 
@@ -330,6 +335,7 @@ defmodule Mix.Tasks.Test do
   @option_keys [
     :trace,
     :max_cases,
+    :max_fail,
     :include,
     :exclude,
     :seed,
@@ -368,7 +374,7 @@ defmodule Mix.Tasks.Test do
   defp default_opts(opts) do
     # Set autorun to false because Mix
     # automatically runs the test suite for us.
-    [autorun: false] ++ opts
+    [autorun: false, max_fail: :infinity] ++ opts
   end
 
   defp parse_files([], test_paths) do
--- a/lib/mix/lib/mix/tasks/test.ex
+++ b/lib/mix/lib/mix/tasks/test.ex
@@ -362,6 +362,7 @@ defmodule Mix.Tasks.Test do
       |> formatter_opts()
       |> color_opts()
       |> Keyword.take(@option_keys)
+      |> max_fail_opts()
       |> default_opts()
 
     {opts, allowed_files}
@@ -374,7 +375,7 @@ defmodule Mix.Tasks.Test do
   defp default_opts(opts) do
     # Set autorun to false because Mix
     # automatically runs the test suite for us.
-    [autorun: false, max_fail: :infinity] ++ opts
+    [autorun: false] ++ opts
   end
 
   defp parse_files([], test_paths) do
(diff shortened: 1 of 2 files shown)


lib/mix/lib/mix/tasks/test.ex
```diff
@@ -277,6 +279,9 @@ defmodule Mix.Tasks.Test do
         option_only_present? = Keyword.has_key?(opts, :only)
 
         cond do
+          failures >= opts[:max_fail] ->
```
josevalim: This is only invoked after all tests run, so it is not going to give you the desired effect.

lib/mix/test/mix/tasks/test_test.exs
```diff
@@ -46,21 +52,21 @@ defmodule Mix.Tasks.TestTest do
...
       assert_stale_run_output("""
       No stale tests
       """)
-    end
+    end)
   end
 
   test "--stale: runs tests that depend on modified modules" do
-    in_fixture "test_stale", fn ->
+    in_fixture("test_stale", fn ->
```
josevalim: Whatever you are using to format the code is not considering the .formatter.exs at the root of the Elixir codebase, as it should not add those parens. :)

lib/mix/lib/mix/tasks/test.ex
Kartstig: Right, it will be caught and it won't stop until the suite is done. Should a call to `agent_stop` be made?

lib/mix/test/mix/tasks/test_test.exs
Kartstig: Ah. Yeah it seems like it's coming from my Sublime Text formatter. I'll remove these changes

lib/mix/lib/mix/tasks/test.ex
josevalim: @Kartstig we would need to add support for this in ExUnit instead.

lib/mix/lib/mix/tasks/test.ex
```diff
@@ -58,6 +58,7 @@ defmodule Mix.Tasks.Test do
     * `--listen-on-stdin` - runs tests, and then listens on stdin. Receiving a newline will
       result in the tests being run again. Very useful when combined with `--stale` and
       external commands which produce output on stdout upon file system modification.
+    * `--max-fail` - sets the maximum number of tests to allow to fail before stopping
```
eksperimental: You are missing the trailing period at the end of this sentence

lib/mix/test/mix/tasks/test_test.exs
```diff
@@ -46,11 +52,11 @@ defmodule Mix.Tasks.TestTest do
   defp ex_unit_opts_from_given(passed) do
     passed
     |> ex_unit_opts()
-    |> Keyword.drop([:failures_manifest_file, :autorun])
+    |> Keyword.drop(@default_opts)
   end
 
   test "--stale: runs all tests for first run, then none on second" do
-    in_fixture "test_stale", fn ->
+    in_fixture"test_stale", fn ->
```
eksperimental: Is the removal of the white space intentional?

lib/mix/lib/mix/tasks/test.ex
```diff
@@ -450,6 +457,16 @@ defmodule Mix.Tasks.Test do
     Enum.filter(matched_test_files, &MapSet.member?(allowed_files, Path.expand(&1)))
   end
 
+  defp max_fail_opts(opts) do
+    case Keyword.fetch(opts, :max_fail) do
+      {:ok, _} ->
+        opts
```
eksperimental: What happens when max fail is a negative integer? I think we should raise.

lib/mix/test/mix/tasks/test_test.exs
```diff
@@ -3,10 +3,16 @@ Code.require_file("../../test_helper.exs", __DIR__)
...
+  @default_opts [:failures_manifest_file, :autorun, :max_fail]
+
   test "ex_unit_opts/1 returns ex unit options" do
     assert ex_unit_opts_from_given(unknown: "ok", seed: 13) == [seed: 13]
   end
 
+  test "ex_unit_opts/1 returns max_fail" do
+    assert ex_unit_opts_from_given(max_fail: 13) == [max_fail: 13]
+  end
+
```
eksperimental: It would be good to have a test for negative integers

lib/mix/lib/mix/tasks/test.ex
eksperimental: If max fail is zero, this will fail eventhough no failures happened. So we need to cover this corner case

lib/mix/lib/mix/tasks/test.ex
eksperimental: Actually the logic is wrong. It should be `>`

lib/mix/lib/mix/tasks/test.ex
eksperimental: Actually the logic is wrong. It should be `>`

josevalim:
> I originally branched off v1.6.3, but then when trying to merge the latest master I had a ton of conflicts, so I cherry-picked my commits off the latest master, but now I get the following errors:

Run `make clean` and start again. :)

lib/mix/test/mix/tasks/test_test.exs
Kartstig: Not intentional

eksperimental:
I can take care of finishing this feature but for some reason I thought something had to be decided on the Elixirs core side, or I may be mistaken with another PR

josevalim:
I don't think there is nothing pending on our side. We should call it `--max-failures` and implement it rather on the ExUnit.Runner side. The other thing is what to do with concurrent tests. When a test fails other N tests may be running and we can't just call them, otherwise we would skip cleanup.

Kartstig:
Thanks @eksperimental. I got stuck on this after Jose pointed out that the actual feature needs to be integrated into ExUnit. I couldn't find the right place for implementation and lost track of this. If you could point me in the right direction, I may be able to contribute

PR #7483, opened 2018-03-23
typofix for quote doc

davionxiang:


whatyouhide:
Thanks @davionxiang! 💟

PR #7484, opened 2018-03-24
Improve error message for invalid clauses inside local functions

### LINKED ISSUE #7440: Improve unhandled operator error message
**Code:**
```
defmodule Parallel do
def pmap(collection, fun) do
collection
|> Enum.map(&spam_process(&1, self, fun))
|> Enum.map(&await/1)
end

defp spawn_process(item, parent, fun) do
spawn_link fn ->
send parent, {self, fun.(item)}
end
end

defp await(pid) do
recieve do
{^pid, result} -> result
end
end

end
```

The issue is that I spelled the word `receive` wrong.

**Error Message:**
```
iex processes.ex
Erlang/OTP 20 [erts-9.2] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:10] [hipe] [kernel-poll:false] [dtrace]

warning: variable "self" does not exist and is being expanded to "self()", please use parentheses to remove the ambiguity or change the variable name
processes.ex:4

warning: variable "self" does not exist and is being expanded to "self()", please use parentheses to remove the ambiguity or change the variable name
processes.ex:10

** (CompileError) processes.ex:16: unhandled operator ->
(stdlib) lists.erl:1354: :lists.mapfoldl/3
(stdlib) lists.erl:1355: :lists.mapfoldl/3
```

The above error message talks about line 16 having an unhandled operator however the error isn't clear about the problem. Could this be improved upon?

ggcampinho:
Closes #7440

### DIFF
--- a/lib/elixir/src/elixir_expand.erl
+++ b/lib/elixir/src/elixir_expand.erl
@@ -701,14 +701,15 @@ assert_no_clauses(_Name, _Meta, [], _E) ->
 assert_no_clauses(Name, Meta, Args, E) ->
   assert_arg_with_no_clauses(Name, Meta, lists:last(Args), E).
 
-assert_arg_with_no_clauses(_Name, _Meta, Arg, _E) when not is_list(Arg) ->
-  ok;
-assert_arg_with_no_clauses(_Name, _Meta, [], _E) ->
-  ok;
-assert_arg_with_no_clauses(Name, Meta, [{_, [{'->', _, _} | _]}], E) ->
-  form_error(Meta, ?key(E, file), ?MODULE, {invalid_clauses, Name});
-assert_arg_with_no_clauses(Name, Meta, [_ | Tail], E) ->
-  assert_arg_with_no_clauses(Name, Meta, Tail, E).
+assert_arg_with_no_clauses(Name, Meta, [{Key, Value} | Rest], E) when is_atom(Key) ->
+  case Value of
+    [{'->', _, _} | _] ->
+      form_error(Meta, ?key(E, file), ?MODULE, {invalid_clauses, Name});
+    _ ->
+      assert_arg_with_no_clauses(Name, Meta, Rest, E)
+  end;
+assert_arg_with_no_clauses(_Name, _Meta, _Arg, _E) ->
+  ok.
 
 expand_local(Meta, Name, Args, #{function := nil} = E) ->
   form_error(Meta, ?key(E, file), ?MODULE, {undefined_function, Name, Args});
--- a/lib/elixir/src/elixir_expand.erl
+++ b/lib/elixir/src/elixir_expand.erl
@@ -406,7 +406,6 @@ expand({Name, Meta, Kind} = Var, E) when is_atom(Name), is_atom(Kind) ->
 
 expand({Atom, Meta, Args}, E) when is_atom(Atom), is_list(Meta), is_list(Args) ->
   assert_no_ambiguous_op(Atom, Meta, Args, E),
-  assert_no_clauses(Atom, Meta, Args, E),
 
   elixir_dispatch:dispatch_import(Meta, Atom, Args, E, fun() ->
     expand_local(Meta, Atom, Args, E)
@@ -706,7 +705,7 @@ assert_arg_with_no_clauses(_Name, _Meta, Arg, _E) when not is_list(Arg) ->
   ok;
 assert_arg_with_no_clauses(_Name, _Meta, [], _E) ->
   ok;
-assert_arg_with_no_clauses(Name, Meta, [{Opt, [{'->', _, _} | _]}], E) when Opt =/= rescue, Opt =/= 'catch' ->
+assert_arg_with_no_clauses(Name, Meta, [{_, [{'->', _, _} | _]}], E) ->
   form_error(Meta, ?key(E, file), ?MODULE, {invalid_clauses, Name});
 assert_arg_with_no_clauses(Name, Meta, [_ | Tail], E) ->
   assert_arg_with_no_clauses(Name, Meta, Tail, E).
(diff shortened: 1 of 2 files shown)


lib/elixir/src/elixir_expand.erl
```diff
@@ -1008,6 +1023,20 @@ format_error({op_ambiguity, Name, Arg}) ->
...
+  SupportMessage =
+    case Name of
+      'case' ->
+        "If you are trying to use the \"case\" structure, it should be similar to:\n"
+        "    case value do"
+        "      pattern -> :ok"
+        "    end";
+
+      _ ->
+        "Clauses are only valid for case, cond, receive, catch, rescue and fn."
```
michalmuskala: This is not entirely true. It depends - e.g. in `with`, it's also valid with `else`.

ggcampinho: I've added `else` to the list, I can't figure out a better message without getting too complicated.

lib/elixir/src/elixir_expand.erl
```diff
@@ -406,6 +406,7 @@ expand({Name, Meta, Kind} = Var, E) when is_atom(Name), is_atom(Kind) ->
 
 expand({Atom, Meta, Args}, E) when is_atom(Atom), is_list(Meta), is_list(Args) ->
   assert_no_ambiguous_op(Atom, Meta, Args, E),
+  assert_no_clauses(Atom, Meta, Args, E),
```
josevalim: Aren't we going to have false positives by moving it here? Imagine we have this macro:

```elixir
def escape(quoted) do
  Macro.escape(quoted)
end
```

This should still work:

```elixir
escape do
  :bar
end
```

So the problem here is really when we are sure we have a local (as in expand_local) or a remote (expand_remote) and it contains a `:->` element regardless of the key! By the simple fact it isn't macro, it means `->` won't compile.

lib/elixir/src/elixir_expand.erl
```diff
@@ -696,6 +697,20 @@ assert_no_ambiguous_op(Name, Meta, [Arg], E) ->
...
+assert_no_clauses(_Name, _Meta, [], _E) ->
+  ok;
+assert_no_clauses(Name, Meta, Args, E) ->
+  assert_arg_with_no_clauses(Name, Meta, lists:last(Args), E).
+
+assert_arg_with_no_clauses(_Name, _Meta, Arg, _E) when not is_list(Arg) ->
+  ok;
+assert_arg_with_no_clauses(_Name, _Meta, [], _E) ->
+  ok;
+assert_arg_with_no_clauses(Name, Meta, [{Opt, [{'->', _, _} | _]}], E) when Opt =/= rescue, Opt =/= 'catch' ->
```
josevalim: From what I said above, if we find a `->` during expand_local/expand_remote, then it should always fail, regardless of the context.

lib/elixir/src/elixir_expand.erl
```diff
@@ -1008,6 +1023,20 @@ format_error({op_ambiguity, Name, Arg}) ->
...
+  SupportMessage =
+    case Name of
+      'case' ->
+        "If you are trying to use the \"case\" structure, it should be similar to:\n"
+        "    case value do"
+        "      pattern -> :ok"
+        "    end";
+
+      _ ->
+        "Clauses are only valid for case, cond, fn, receive, else, catch or rescue."
```
josevalim: Clauses they are actually valid in any macro. For now I think we should drop the support message. :)

lib/elixir/src/elixir_expand.erl
```diff
@@ -1008,6 +1023,20 @@ format_error({op_ambiguity, Name, Arg}) ->
...
+    end,
+  io_lib:format("invalid clauses with -> operator passed to local function \"~ts\". ~s",
```
josevalim: What do you think about: "the function "~ts" cannot handle clauses with the -> operator because it is not macro. Please make sure you are invoking the proper name and that it is a macro"?

michalmuskala:
After thinking some more on this, I'm a bit worried about doing this for every single function call - this seems quite expensive to do.

Maybe we should short-circuit `assert_arg_with_no_clauses` if the list is not a keyword list? I think this should make it a bit less expensive - we would effectively only check calls with keyword options at the end.

lib/elixir/src/elixir_expand.erl
```diff
@@ -696,18 +696,36 @@ assert_no_ambiguous_op(Name, Meta, [Arg], E) ->
...
+assert_no_clauses(_Name, _Meta, [], _E) ->
+  ok;
+assert_no_clauses(Name, Meta, Args, E) ->
+  assert_arg_with_no_clauses(Name, Meta, lists:last(Args), E).
+
+assert_arg_with_no_clauses(_Name, _Meta, Arg, _E) when not is_list(Arg) ->
+  ok;
+assert_arg_with_no_clauses(_Name, _Meta, [], _E) ->
+  ok;
+assert_arg_with_no_clauses(Name, Meta, [{_, [{'->', _, _} | _]}], E) ->
```
josevalim: We are not checking this case:

```elixir
wrong_fun do
  foo -> bar
after
  :ok
end
```

Because this clause is matching exclusively on the last entry as it doesn't consider the tail. Given @michalmuskala feedback, I think replacing all clauses above by these two would be fast and consider all cases:

```erlang
assert_arg_with_no_clauses(Name, Meta, [{Key, Value} | Rest], E) when is_atom(Key) ->
  case Value of
    [{'->', _, _} | _] ->
      form_error(Meta, ?key(E, file), ?MODULE, {invalid_clauses, Name});
    _ ->
      assert_arg_with_no_clauses(Name, Meta, Rest, E)
  end;
assert_arg_with_no_clauses(_Name, _Meta, _Arg, _E) ->
  ok.
```

michalmuskala:
Another way to solve this, though, I'm pretty sure you won't like it very much, @josevalim, would be to wrap the `expand_args` call in a try/catch and then throw when we expand and encounter unhandled `->` - at top level we could have a try/catch with some generic error.

josevalim:

PR #7486, opened 2018-03-25
Fix column count on heredoc tokenizer

ggcampinho:


josevalim:

PR #7487, opened 2018-03-25
Improve error messages for lists concatenation operator

### LINKED ISSUE #7485: Erlang error for ++ operator inside match
### Environment

```
Erlang/OTP 20 [erts-9.2] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:10] [hipe] [kernel-poll:false] [dtrace]

Elixir 1.7.0-dev (8c05bb0) (compiled with OTP 20)
```
* Operating system: MacOS High Sierra 10.13.3

### Current behavior

```
iex(1)> "a" ++ "b" = "ab"
** (ErlangError) Erlang error: {:illegal_pattern, {:op, 1, :++, {:bin, 0, [{:bin_element, 0, {:string, 0, 'a'}, :default, :default}]}, {:bin, 0, [{:bin_element, 0, {:string, 0, 'b'}, :default, :default}]}}}
iex(1)> "a" ++ "b"
    :erlang.++("a", "b")
```

### Expected behavior

I believe it should give an error similar to the <> operator:
```
iex(1)> "a" ++ "b" = "ab"
** (ArgumentError) expected list argument in <> operator but got: "a"
iex(1)> "a" ++ "b"
** (ArgumentError) expected list argument in <> operator but got: "a"
```

If the issue makes sense, I would be glad to fix it.

ggcampinho:
Closes #7485

### DIFF
--- a/lib/elixir/src/elixir_expand.erl
+++ b/lib/elixir/src/elixir_expand.erl
@@ -715,6 +715,7 @@ expand_remote(Receiver, DotMeta, Right, Meta, Args, #{context := Context} = E, E
   {EArgs, EA} = expand_args(Args, E),
   Rewritten = elixir_rewrite:rewrite(Receiver, DotMeta, Right, Meta, EArgs),
   maybe_warn_struct_comparison(Rewritten, Args, E),
+  assert_valid_args_for_remote(Rewritten, E),
   case allowed_in_context(Rewritten, Arity, Context) of
     true ->
       {Rewritten, elixir_env:mergev(EL, EA)};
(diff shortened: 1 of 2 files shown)


lib/elixir/src/elixir_expand.erl
```diff
@@ -733,6 +734,16 @@ allowed_in_context(_, _Arity, guard) ->
 allowed_in_context(_, _, _) ->
   true.
 
+assert_valid_args_for_remote({{'.', _, [erlang, '++']}, Meta, [{_, _} = Left, _Right]}, E) ->
```
josevalim: Hrm, I think we are being a bit inconsistent here. For example, this code warns at best:

```elixir
1 + :foo
```

While being obviously invalid. I wouldn't want to add all of those checks to the compiler, since they are already performed by a later Erlang pass anyway.

There are other places, such as guards, where doing so is weird, although semantically valid:

```elixir
def foo when 1 ++ 2 do
  ...
end
```

So my suggestion is for us to raise when the context is a match and whenever the left side is not a list. This will simplify the check, make sure we don't raise an internal erlang error, and make `++` behave like the other operators.

lib/elixir/src/elixir_expand.erl
```diff
@@ -972,6 +983,9 @@ format_error(wrong_number_of_args_for_super) ->
 format_error({invalid_arg_for_pin, Arg}) ->
   io_lib:format("invalid argument for unary operator ^, expected an existing variable, got: ^~ts",
                 ['Elixir.Macro':to_string(Arg)]);
+format_error({invalid_arg_for_union, Arg}) ->
```
josevalim: Instead of "union" let's call it the "lists_concatenation".

lib/elixir/src/elixir_expand.erl
ggcampinho: I did some tests and I believe is not valid semantically inside guards:
```
iex(1)> defmodule A do
...(1)> def foo when 1 ++ 2 do
...(1)> 1
...(1)> end
...(1)> end
** (CompileError) iex:2: illegal guard expression
    (iex) lib/iex/evaluator.ex:245: IEx.Evaluator.handle_eval/5
```

But it's also invalid in Erlang and doesn't impact the solution. I agree that is simpler in the matcher only, and let Erlang handle the other errors, I just don't know if I understood what you mean with changes for `<<>>`, but I think you removed.

lib/elixir/src/elixir_expand.erl
```diff
@@ -972,6 +978,9 @@ format_error(wrong_number_of_args_for_super) ->
 format_error({invalid_arg_for_pin, Arg}) ->
   io_lib:format("invalid argument for unary operator ^, expected an existing variable, got: ^~ts",
                 ['Elixir.Macro':to_string(Arg)]);
+format_error({invalid_arg_for_lists_concatenation, Arg}) ->
+  io_lib:format("invalid argument for ++ operator, expected a list, got: ~ts",
```
josevalim: One last thing. We need to clarify that it is a literal list and that it is only during a match. What do you think about "invalid argument for ++ operator inside a match, expected a literal list, got: ~ts"?

josevalim:

PR #7493, opened 2018-03-26
Optimize tokenizer for commas

ggcampinho:
Small optimization skipping the `handle_terminator` function for commas since they don't require terminators.

I'm not sure if this makes sense, feel free to close it if doesn't.


josevalim:

PR #7499, opened 2018-03-27
Fix small math error in error_logger config option

mitchellhenke:
75% of 500 should be 225 instead of 250 🙂


josevalim:

Issue #7500, opened 2018-03-27
Function capture support using variables


NJichev:
### The issue

Say we have a simple module:

```elixir
defmodule Math do
  def zero?(0), do: true
  def zero?(_), do: false
end
```
And we want to capture the function by using these variables:

```elixir
{m, f, a} = {Math, :zero?, 1}
```
Using `&` - we cannot:
```elixir
&m.f/a
** (CompileError) iex:8: invalid args for &, expected an expression in the format of &Mod.fun/arity, &local/arity or a capture containing at least one argument as &1, got: m.f() / a
```
What we could do is use `make_fun`:
```elixir
:erlang.make_fun(m, f, a).(0) #=> true
```
### Expected behaviour(API could be different):

```elixir
{m, f, a} = {Math, :zero?, 1}
&{m, f, a}
# or
&m.f/a
#=> &Math.zero?/1
```

This will be useful when writing some macros for removing boiler-plate and the code would be only in elixir.

josevalim:
We cannot support `&m.f/a` because we would take `f` to be the literal function name and not a variable.

My suggestion: introduce a `Function` module with a `capture/3` function and a `info/1` and `info/2` functions that delegate to `erlang:fun_info/1` and `erlang:fun_info/2`.

We could add `Function.apply/2` as well but our official policy is to not duplicate functions that exist in `Kernel` in the other modules.

michalmuskala:
I'm not really sure about `Function.capture/3`. Implementation-wise, I think we should expand it into `fun M:F/A` instead of calling `erlang:make_fun/3`, just because it's not documented. But we don't really compile anything outside of `Kernel` into syntax when translating to erlang.

josevalim:
I would say this isn’t much different from inlining, which is something we
do in all data-structure modules.

We also had plans to inline Map.merge and Map.put into instructions, we
just didn’t do it yet due to OTP version constraints.
-- 


*José Valimwww.plataformatec.com.br
<http://www.plataformatec.com.br/>Founder and Director of R&D*

PR #7503, opened 2018-03-27
Add Function module for introspecting functions.

NJichev:
Reason to add it is capturing functions(#7500) when passing in module name,
function name and arity using variables, because the capture operator
can't handle them.

Done by delegating to `:erlang.make_fun(mod, f, arity)`, also exposed
some erlang functions for getting information for a function and removing
:"Elixir.Function" from the reserved modules.

Added docs for capture/3, info1/ and info/2.
Also added some tests.

### DIFF
--- a/lib/elixir/lib/function.ex
+++ b/lib/elixir/lib/function.ex
@@ -26,17 +26,11 @@ defmodule Function do
 
   ## Examples
 
-    iex> Function.capture(String, :length, 1)
-    &String.length/1
-
-  Useful when you want to capture a function when having variables.
-
-    iex> {module, function_name, arity} = {String, :length, 1}
-    iex> Function.capture(module, function_name, arity)
-    &String.length/1
+      iex> Function.capture(String, :length, 1)
+      &String.length/1
 
   """
-  @spec capture(module, atom, integer) :: (... -> any)
+  @spec capture(module, atom, integer) :: fun
   def capture(module, function_name, arity)
       when is_atom(module) and is_atom(function_name) and is_integer(arity) and arity in 0..255 do
     :erlang.make_fun(module, function_name, arity)
@@ -46,37 +40,42 @@ defmodule Function do
   Returns a keyword list with information about a function.
 
   The `{key, value}`s will include the following:
-  - `:type` - `:local` (for anonymous functions) or `:external` (for
+  
+    * `:type` - `:local` (for anonymous functions) or `:external` (for
   named functions)
-    - `:module` - an atom - the module where the function is defined when
+    * `:module` - an atom - the module where the function is defined when
     anonymous or the module which the function refers to when it's a named one.
-    - `:arity` - the number of arguments the function is to be called with
-    - `:name` - the name of the functions
-    - `:env` - a list of the environment or free variables. For named
-    functions, the returned list is always empty.
+    * `:arity` - the number of arguments the function is to be called with
+    * `:name` - the name of the functions
+    * `:env` - a list of the environment or free variables. For named
+      functions, the returned list is always empty.
 
   When it is an anonymous function it will also return info about:
-  - `:pid` - process identifier of the process that originally created
+  
+    * `:pid` - process identifier of the process that originally created
   the funciton
-    - `:index` - an integer - is an index into the module function table.
-    - `:new_index` - an integer - is an index into the module function table.
-    - `:new_uniq` - a binary-  it's a unique value for this function. It is
+    * `:index` - an integer - is an index into the module function table.
+    * `:new_index` - an integer - is an index into the module function table.
+    * `:new_uniq` - a binary-  it's a unique value for this function. It is
     calculated from the compiled code for the entire module.
-    - `:uniq` - an integer, a unique value for this function. As from Erlang/OTP
+    * `:uniq` - an integer, a unique value for this function. As from Erlang/OTP
     R15, this integer is calculated from the compiled code for the entire
     module. Before Erlang/OTP R15, this integer was based on only the body
     of the function.
 
+  **Note**: this function must be used only for debugging purposes.
 
   ## Examples
-    iex> fun = fn x -> x end
-    iex> info = Function.info(fun)
-    iex> info |> Keyword.get(:arity)
-    1
-    iex> info |> Keyword.get(:type)
-    :local
+
+      iex> fun = fn x -> x end
+      iex> info = Function.info(fun)
+      iex> Keyword.get(info, :arity)
+      1
+      iex> Keyword.get(info, :type)
+      :local
+
   """
-  @spec info((... -> any)) :: [{information, value}]
+  @spec info(fun) :: [{information, value}]
   def info(fun) when is_function(fun), do: :erlang.fun_info(fun)
 
   @doc """
@@ -91,19 +90,19 @@ defmodule Function do
   For a named function, the value of any of these items is always the
   atom `:undefined`.
 
-  For more information on each of the value returned - check the docs for
+  For more information on each of the value returned. Check the docs for
   `Function.info/1`.
 
   ## Examples
 
-    iex> f = fn x -> x end
-    iex> Function.info(f, :arity)
-    {:arity, 1}
-    iex> Function.info(f, :type)
-    {:type, :local}
+      iex> f = fn x -> x end
+      iex> Function.info(f, :arity)
+      {:arity, 1}
+      iex> Function.info(f, :type)
+      {:type, :local}
 
   """
-  @spec info((... -> any)) :: {information, value}
+  @spec info(fun) :: {information, value}
   def info(fun, item) when is_function(fun) and is_atom(item) do
     :erlang.fun_info(fun, item)
   end
--- a/lib/elixir/lib/function.ex
+++ b/lib/elixir/lib/function.ex
@@ -33,6 +33,7 @@ defmodule Function do
       &String.length/1
 
   """
+  @since "1.7.0"
   @spec capture(module, atom, integer) :: fun
   def capture(module, function_name, arity) do
     :erlang.make_fun(module, function_name, arity)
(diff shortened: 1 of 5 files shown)


lib/elixir/lib/function.ex
```diff
@@ -0,0 +1,90 @@
...
+  @doc """
+  Captures a function by a given module, function name and arity.
+
+  Useful when you want to capture a function when having variables.
+
+  ## Examples
+
+    iex> Function.capture(String, :length, 1)
+    &String.length/1
+  """
```
eksperimental: This doc should be next to the function it is describing

lib/elixir/lib/function.ex
```diff
@@ -0,0 +1,90 @@
...
+  A set of funtions for working with functions.
+
+  We can have two types of captures functions - external and local -
+  external functions are our normal functions residing in modules while
+  local are the anonymous functions defined with `fn` or the capture
+  operator `&`.
+  """
+
+  @doc """
+  Captures a function by a given module, function name and arity.
```
eksperimental: We can use backticks around module, function and arity

lib/elixir/lib/function.ex
```diff
@@ -0,0 +1,90 @@
...
+          | :module
+          | :arity
+          | :name
+          | :env
+          | :pid
+          | :index
+          | :new_index
+          | :new_uniq
+          | :uniq
+  @type value :: any
```
eksperimental: Tyoe definitions could be moved up bellow moduledoc

lib/elixir/lib/function.ex
```diff
@@ -0,0 +1,90 @@
...
+
+  @spec capture(atom, atom, integer) :: (... -> any)
+  def capture(mod, fun, arity) when is_atom(mod) and is_atom(fun) and is_integer(arity) do
```
eksperimental: I woulf rename args mod to module and fun to function_name.
Fun is used when they are anonimous functions, not atoms

eksperimental: Guard for arity should be
is_integer(arity) and arity >= 0 and arity <= 255

lib/elixir/lib/function.ex
```diff
@@ -0,0 +1,90 @@
...
+          | :name
+          | :env
+          | :pid
+          | :index
+          | :new_index
+          | :new_uniq
+          | :uniq
+  @type value :: any
+
+  @spec capture(atom, atom, integer) :: (... -> any)
```
eksperimental: Args should be: module, atom, arity

lib/elixir/lib/function.ex
```diff
@@ -0,0 +1,90 @@
...
+    - :name - the name of the functions
+    - :env - a list of the environment or free variables. For named functions, the returned list is always empty.
+
+  When it is an anonymous function it will also return info about:
+    - :pid - process identifier of the process that originally created the fun
+    - :index - an integer - is an index into the module fun table.
+    - :new_index - an integer - is an index into the module fun table.
+    - :new_uniq - a binary-  it's a unique value for this fun. It is calculated from the compiled code for the entire module.
+    - :uniq - an integer, a unique value for this fun. As from Erlang/OTP R15, this integer is calculated from the compiled code for the entire module. Before Erlang/OTP R15, this integer was based on only the body of the fun.
+
```
eksperimental: Backticks are needed around every atom

lib/elixir/lib/function.ex
```diff
@@ -0,0 +1,90 @@
+defmodule Function do
+  @moduledoc """
+  A set of funtions for working with functions.
+
+  We can have two types of captures functions - external and local -
+  external functions are our normal functions residing in modules while
+  local are the anonymous functions defined with `fn` or the capture
+  operator `&`.
```
eksperimental: I think that fn and & should have the slash and the arity so it cane be linked in the documentation.. but im not too sure here how we are dealing with it somewhere else

lib/elixir/lib/function.ex
```diff
@@ -0,0 +1,90 @@
...
+          :type
+          | :module
+          | :arity
+          | :name
+          | :env
+          | :pid
+          | :index
+          | :new_index
+          | :new_uniq
+          | :uniq
```
eksperimental: Atoms shoud be sorted alphabetically

lib/elixir/lib/function.ex
```diff
@@ -0,0 +1,90 @@
...
+
+  @spec capture(atom, atom, integer) :: (... -> any)
+  def capture(mod, fun, arity) when is_atom(mod) and is_atom(fun) and is_integer(arity) do
+    :erlang.make_fun(mod, fun, arity)
+  end
+
+  @doc """
+  Returns a keyword list with information about a function.
+
+  The {key,value}s will include the following:
```
eksperimental: Backticks around tuple. Space after coma

lib/elixir/lib/function.ex
```diff
@@ -0,0 +1,90 @@
...
+  We can have two types of captures functions - external and local -
+  external functions are our normal functions residing in modules while
+  local are the anonymous functions defined with `fn` or the capture
+  operator `&`.
+  """
+
+  @doc """
+  Captures a function by a given module, function name and arity.
+
+  Useful when you want to capture a function when having variables.
```
eksperimental: It is not clear what it means. Maybe an example showing it will help

lib/elixir/lib/function.ex
```diff
@@ -0,0 +1,90 @@
...
+  @spec capture(atom, atom, integer) :: (... -> any)
+  def capture(mod, fun, arity) when is_atom(mod) and is_atom(fun) and is_integer(arity) do
+    :erlang.make_fun(mod, fun, arity)
+  end
+
+  @doc """
+  Returns a keyword list with information about a function.
+
+  The {key,value}s will include the following:
+    - :type - :local(for anonymous functions) or :external(for named functions)
```
eksperimental: Add white-Space after :local and :external

lib/elixir/lib/function.ex
```diff
@@ -0,0 +1,90 @@
...
+    - :arity - the number of arguments the function is to be called with
+    - :name - the name of the functions
+    - :env - a list of the environment or free variables. For named functions, the returned list is always empty.
+
+  When it is an anonymous function it will also return info about:
+    - :pid - process identifier of the process that originally created the fun
+    - :index - an integer - is an index into the module fun table.
+    - :new_index - an integer - is an index into the module fun table.
+    - :new_uniq - a binary-  it's a unique value for this fun. It is calculated from the compiled code for the entire module.
+    - :uniq - an integer, a unique value for this fun. As from Erlang/OTP R15, this integer is calculated from the compiled code for the entire module. Before Erlang/OTP R15, this integer was based on only the body of the fun.
```
eksperimental: Line is too long

lib/elixir/lib/function.ex
```diff
@@ -0,0 +1,90 @@
...
+  The {key,value}s will include the following:
+    - :type - :local(for anonymous functions) or :external(for named functions)
+    - :module - an atom - the module where the function is defined when anonymous
+    or the module which the function refers to when it's a named one.
+    - :arity - the number of arguments the function is to be called with
+    - :name - the name of the functions
+    - :env - a list of the environment or free variables. For named functions, the returned list is always empty.
+
+  When it is an anonymous function it will also return info about:
+    - :pid - process identifier of the process that originally created the fun
```
eksperimental: I wouldn't say fun but function. We need the trailing period in this sentence. I think sentence is a bit redundant. I would go with: PID of the process that created the function.

josevalim:
Thanks @NJichev! I have pushed some cosmetic changes:

1. used `*` for bullets
2. indented examples with 4 columns
3. used `fun` in typespecs

lib/elixir/test/elixir/function_test.exs
```diff
@@ -0,0 +1,80 @@
...
+
+  doctest Function
+  import Function
+
+  @information_keys_for_named [:type, :module, :arity, :name, :env]
+  @information_keys_for_anonymous @information_keys_for_named ++
+                                    [:pid, :index, :new_index, :new_uniq, :uniq]
+
+  describe "capture/3" do
+    test "that it can capture module functions with arity 0" do
```
josevalim: We can remove the "that it can" and "it" prefixes from tests. This should read as: `test "captures module functions with arity 0"`. Same for the tests below.

lib/elixir/test/elixir/function_test.exs
```diff
@@ -0,0 +1,80 @@
...
+      end)
+    end
+  end
+
+  describe "info/2" do
+    test "it returns info for every possible information key for named functions" do
+      f = &DummyFunction.zero?/1
+
+      @information_keys_for_named
+      |> Enum.each(fn x ->
```
josevalim: Let's use a for comprehension here. No particular reason rather than that's what we commonly use in Elixir codebase:

    for x <- @information_keys_for_named do

Same for the tests below.

josevalim:
@NJichev we also need to inline those functions. Could you please add entries for all three functions here: https://github.com/elixir-lang/elixir/blob/master/lib/elixir/src/elixir_rewrite.erl#L24

Then update their docs by adding a single paragraph saying "Inlined by the compiler". See the `Process` module for more information.

Also, could you please run the formatter again in lib/elixir/lib/function.ex? I messed it up, apologies.

michalmuskala:
If we do inline the function, we shouldn't be having guards there - we want to have exactly the same errors if called directly (and inlined) or called dynamically (for example through `apply`).

josevalim:
> If we do inline the function, we shouldn't be having guards there - we want to have exactly the same errors if called directly (and inlined) or called dynamically (for example through apply).

Good call. So let's remove the guards too.

josevalim:
@NJichev oh, one last thing! Let's add `@since "1.7.0"` after the `@doc` and before each function in the `Function` module. :)

josevalim:

NJichev:
Thanks for the guidance :bowing_man:

whatyouhide:
@eksperimental I already fixed most of your comments for the `Function` module in #7509.

PR #7509, opened 2018-03-28
Improve documentation and specs in the Function module

whatyouhide:
A bunch of improvements that I was late for making in https://github.com/elixir-lang/elixir/pull/7503 :).

Notable things:

  * I removed the `Function.value/0` type. I think we should be conservative with types since then they're public API and since this type was just defined as `term/0`, I used `term` instead.

  * Added some additional examples.

  * Restructured docs a bit.

  * Fixed the spec for `info/2` (which was defined as `info(fun)`) so that it is defined for `info/2` and so that it encodes the information that the passed item is returned as well.

### DIFF
--- a/lib/elixir/lib/function.ex
+++ b/lib/elixir/lib/function.ex
@@ -2,10 +2,10 @@ defmodule Function do
   @moduledoc """
   A set of functions for working with functions.
 
-  We can have two types of captured functions: external and local.
+  There are two types of captured functions: **external** and **local**.
   External functions are functions residing in modules that are captured
-  with `&/1`, such as `&String.length/1`. Local are the anonymous functions
-  defined with `fn/1` or with the capture operator `&/1` using `&1`, `&2`
+  with `&/1`, such as `&String.length/1`. Local functions are anonymous functions
+  defined with `fn/1` or with the capture operator `&/1` using `&1`, `&2`,
   and so on as replacements.
   """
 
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/function.ex
```diff
@@ -2,10 +2,10 @@ defmodule Function do
   @moduledoc """
   A set of functions for working with functions.
 
-  We can have two types of captured functions: external and local.
+  There are two types of captured functions: **external** and **local**.
```
wojtekmach: s/external/remote?

josevalim: This is the function internal naming. So we need to stick with external.

wojtekmach: TIL, thanks.

michalmuskala: Yes, we should stick with "external" - it's what `:erlang.fun_info(fun, :type)` returns. Although, internally in erts they are called exports.

lib/elixir/lib/function.ex
```diff
@@ -2,10 +2,10 @@ defmodule Function do
...
   External functions are functions residing in modules that are captured
-  with `&/1`, such as `&String.length/1`. Local are the anonymous functions
-  defined with `fn/1` or with the capture operator `&/1` using `&1`, `&2`
+  with `&/1`, such as `&String.length/1`. Local functions are anonymous functions
+  defined with `fn/1` or with the capture operator `&/1` using `&1`, `&2`,
   and so on as replacements.
```
michalmuskala: Would it make sense to add an example using `Function.info(fn -> nil end, :type) => :local` and `Function.info(&String.length/1, :type) => :external`?

whatyouhide: We have that example in `Function.info/1`, I think it's fine to skip it here.

PR #7510, opened 2018-03-28
Add example .formatter.exs for lazy folks

iamvery:
It wasn't immediately obvious to me how the `.formatter.exs` file would look. I probably shouldn't have read the documentation more closely, but I really wanted/expected to see an example. The example I added is from https://hashrocket.com/blog/posts/format-your-elixir-code-now, which seems like a reasonable default.

What do you think?

### DIFF
--- a/lib/mix/lib/mix/tasks/format.ex
+++ b/lib/mix/lib/mix/tasks/format.ex
@@ -17,6 +17,14 @@ defmodule Mix.Tasks.Format do
   formatter configuration. Evaluating this file should return a keyword list
   with any of the options supported by `Code.format_string!/2`.
 
+  ### Example `.formatter.exs`
+
+  ```
+  [
+    inputs: ["{mix,.formatter}.exs", "{config,lib,test}/**/*.{ex,exs}"]
+  ]
+  ```
+
   The `.formatter.exs` also supports other options:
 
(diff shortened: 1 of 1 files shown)


lib/mix/lib/mix/tasks/format.ex
```diff
@@ -17,6 +17,14 @@ defmodule Mix.Tasks.Format do
   formatter configuration. Evaluating this file should return a keyword list
   with any of the options supported by `Code.format_string!/2`.
 
+  ### Example `.formatter.exs`
+
+  ```
+  [
+    inputs: ["mix.exs", "{config,lib,test}/**/*.{ex,exs}"]
```
whatyouhide: I think the best default would be:

```elixir
[
  inputs: ["{mix,.formatter}.exs", "{config,lib,test}/**/*.{ex,exs}"]
]
```

so that the formatter configuration is also formatted :)

iamvery: LGTM! updating

josevalim:
I tried to improve the flow a bit. @iamvery and @whatyouhide, what do you think?

iamvery:
Wonderful! Thank you ✨

josevalim:

Issue #7513, opened 2018-03-29
Also emit warnings for guard checks with `length(Something) > 0`


josevalim:
We already emit for `== 0`, checking for `> 0` would be the negated form of that.

whatyouhide:
Closed by #7514.

PR #7514, opened 2018-03-29
Warn users that uses length to check if list is not empty on guards

ulissesalmeida:
Some Elixir users check if the list is not
empty by using the length function. It's not the optimal solution since it has to transverse the whole
list. Now we're warning the users and suggesting them
to use `!= []`.

Apply https://github.com/elixir-lang/elixir/issues/7513 enhancement.

### DIFF
--- a/lib/elixir/src/elixir_clauses.erl
+++ b/lib/elixir/src/elixir_clauses.erl
@@ -52,7 +52,15 @@ warn_zero_length_guard({{'.', _, [erlang, '==']}, Meta,
   Message = io_lib:format("\"length(~ts) == 0\" is discouraged since it has to "
                           "traverse the whole list to check if it is empty or not. "
                           "Prefer to pattern match on an empty list or use "
-                          "\"~ts == []\" as a guard", [ArgString, ArgString]),
+                          "\"~ts == []\" as a guard.", [ArgString, ArgString]),
+  elixir_errors:warn(?line(Meta), ?key(E, file), Message);
+warn_zero_length_guard({{'.', _, [erlang, '>']}, Meta,
+                        [{{'.', _, [erlang, length]}, _, [Arg]}, 0]}, E) ->
+  ArgString = 'Elixir.Macro':to_string(Arg),
+  Message = io_lib:format("\"length(~ts) > 0\" is discouraged since it has to "
+                          "traverse the whole list to check if it is empty or not. "
+                          "Prefer to pattern match on a non-empty list, such as [_ | _], "
+                          "or use \"~ts != []\" as a guard.", [ArgString, ArgString]),
   elixir_errors:warn(?line(Meta), ?key(E, file), Message);
 warn_zero_length_guard({Op, _, [L, R]}, E) when Op == 'or'; Op == 'and' ->
   warn_zero_length_guard(L, E),
--- a/lib/elixir/test/elixir/kernel/warning_test.exs
+++ b/lib/elixir/test/elixir/kernel/warning_test.exs
@@ -551,20 +551,51 @@ defmodule Kernel.WarningTest do
   end
 
   test "length(list) == 0 in guard" do
-    assert capture_err(fn ->
-             Code.eval_string("""
-             defmodule Sample do
-               def list_case do
-                 v = []
-                 case v do
-                   _ when length(v) == 0 -> :ok
-                   _ -> :fail
-                 end
-               end
-             end
-             """)
-           end) =~
+    error_message =
+      capture_err(fn ->
+        Code.eval_string("""
+        defmodule Sample do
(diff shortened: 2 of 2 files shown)


lib/elixir/src/elixir_clauses.erl
```diff
@@ -54,6 +54,13 @@ warn_zero_length_guard({{'.', _, [erlang, '==']}, Meta,
                           "Prefer to pattern match on an empty list or use "
                           "\"~ts == []\" as a guard", [ArgString, ArgString]),
   elixir_errors:warn(?line(Meta), ?key(E, file), Message);
+warn_zero_length_guard({{'.', _, [erlang, '>']}, Meta,
+                        [{{'.', _, [erlang, length]}, _, [Arg]}, 0]}, E) ->
+  ArgString = 'Elixir.Macro':to_string(Arg),
+  Message = io_lib:format("\"length(~ts) > 0\" is discouraged since it has to "
+                          "traverse the whole list to check if it is not empty. "
+                          "Prefer to use \"~ts != []\" as a guard", [ArgString, ArgString]),
```
josevalim: What do you think about the last sentence being:

> Prefer to pattern match on a cons cell, such as [_ | _], or use \"~ts != []\" as a guard

josevalim: Note: I changed my comment above to say "non-empty list" instead of "cons cell"

josevalim:
@ulissesalmeida please run `bin/elixir bin/mix format` at the root of the project as well to format the warning_test file. :)

brodeuralexis:
Why not modify `Kernel.length/1` to pattern match ?
```ex
def length([]), do: 0
```

ulissesalmeida:
The problem is when the list is not empty. We need to transverse the entire
list to know its size

Em Qui, 29 de mar de 2018 08:05, Alexis Brodeur <notifications@github.com>
escreveu:

lib/elixir/src/elixir_clauses.erl
```diff
@@ -54,6 +54,13 @@ warn_zero_length_guard({{'.', _, [erlang, '==']}, Meta,
                           "Prefer to pattern match on an empty list or use "
                           "\"~ts == []\" as a guard", [ArgString, ArgString]),
   elixir_errors:warn(?line(Meta), ?key(E, file), Message);
+warn_zero_length_guard({{'.', _, [erlang, '>']}, Meta,
+                        [{{'.', _, [erlang, length]}, _, [Arg]}, 0]}, E) ->
+  ArgString = 'Elixir.Macro':to_string(Arg),
+  Message = io_lib:format("\"length(~ts) > 0\" is discouraged since it has to "
```
eksperimental: I would change the meaning of the sentence in an affirmative way:
> Is discouraged since it has to transverse the whole list just to see if it's empty

lib/elixir/src/elixir_clauses.erl
```diff
@@ -52,7 +52,15 @@ warn_zero_length_guard({{'.', _, [erlang, '==']}, Meta,
...
                           "traverse the whole list to check if it is empty or not. "
                           "Prefer to pattern match on an empty list or use "
-                          "\"~ts == []\" as a guard", [ArgString, ArgString]),
+                          "\"~ts == []\" as a guard.", [ArgString, ArgString]),
+  elixir_errors:warn(?line(Meta), ?key(E, file), Message);
+warn_zero_length_guard({{'.', _, [erlang, '>']}, Meta,
+                        [{{'.', _, [erlang, length]}, _, [Arg]}, 0]}, E) ->
+  ArgString = 'Elixir.Macro':to_string(Arg),
+  Message = io_lib:format("\"length(~ts) > 0\" is discouraged since it has to "
+                          "traverse the whole list to check if it is empty or not. "
```
eksperimental: I think just mentioning that the list is empty is enough. I find the "or not" part unnecessary, specially for an error msg that aims to be as concise as possible

eksperimental: Additionally I would rephrase my previous suggestion since it is not clear what is the "it" in the "it has to traverse...". I wouldngo with this:
> is discouraged since the whole list needs to be traversed in order to check if it's empty

josevalim:
Let’s keep the message as is because it is consistent with the existing
one. I will do another pass after merging incorporating suggestions. :)
-- 


josevalim:

lib/elixir/src/elixir_clauses.erl
josevalim: I think we should avoid passive voice in general as we just moved the ambiguity elsewhere. Previously it was unclear what was the "it", now it is unclear who is causing the action altogether.

PR #7518, opened 2018-03-30
Add public types for some Supervisor functions

whitfin:
Hi there! I'm not sure on the guidelines for submitting PRs, so please let me know of any standards I may need to adhere to in future!

The aim of this PR is to simply expose the return types of `Supervisor.which_children/1` and `Supervisor.count_children/1`. In the case you have a function which relays either of these in your own code, you need to copy the specification (and they're kinda noisy) - I find that it's more convenient to expose the types in the Supervisor module.

That's it really, let me know any thoughts :)

### DIFF
--- a/lib/elixir/lib/supervisor.ex
+++ b/lib/elixir/lib/supervisor.ex
@@ -943,7 +943,7 @@ defmodule Supervisor do
     * `modules` - as specified by the child specification
 
   """
-  @spec which_children(supervisor) :: [ child_tuple ]
+  @spec which_children(supervisor) :: [child_tuple]
   def which_children(supervisor) do
     call(supervisor, :which_children)
   end
--- a/lib/elixir/lib/supervisor.ex
+++ b/lib/elixir/lib/supervisor.ex
@@ -539,6 +539,22 @@ defmodule Supervisor do
           optional(:modules) => [module()] | :dynamic
         }
 
+  @typedoc "Return values used by the `which_children/1` function"
+  @type child_tuple :: {
+          term() | :undefined,
+          child | :restarting,
+          :worker | :supervisor,
+          :supervisor.modules()
+        }
+
+  @typedoc "Return values of the `count_children/1` function"
+  @type child_count :: %{
+          specs: non_neg_integer,
+          active: non_neg_integer,
+          supervisors: non_neg_integer,
+          workers: non_neg_integer
+        }
+
   @doc """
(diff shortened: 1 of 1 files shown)


josevalim:
Thanks! The PR looks great but I am not sure if this is the kind of types we want to expose. Thoughts @michalmuskala and @fishcakez?

michalmuskala:
I'm not really sure it makes sense to extract them as separate types - I would prefer for this, if anything, to be defined directly in the typespec for the function.

whatyouhide:
The problem I see is that this could be argued for all functions that have a non-trivial return type, so I agree with @michalmuskala and would prefer this to be defined in the typespec of the function as well.

whitfin:
Hi guys! Thanks for your feedback. When you say that it's defined in the typespec of the function, do you have an example? Would this still allow other modules to point to it as a type?

One of the reasons I opened this is because I've seen a couple of places where libraries actually return the same form that comes back from `Supervisor.which_children/1` (perhaps with some children filtered out, or a specific child), and I found that it would be better to have a type they can point to inside the Supervisor module.

I agree with @whatyouhide in that it does invite that argument, but at the same time `Supervisor` is so widely used that perhaps that's why it makes more sense for this module rather than somewhere else?

fishcakez:
> perhaps with some children filtered out, or a specific child

Supervisors are *not* for discovery. If this is the use case I don't think we should add the types because we would be enabling an anti-pattern. `which_children/1` should be used by release handler and for debugging, not in place of a registry.

whatyouhide:
> When you say that it's defined in the typespec of the function, do you have an example?

It means just not having a separate type and using a type directly in the function spec:

```elixir
@spec foo() :: :ok
def foo() do
  :ok
end
```

> Would this still allow other modules to point to it as a type?

No.

> I agree with @whatyouhide in that it does invite that argument, but at the same time Supervisor is so widely used that perhaps that's why it makes more sense for this module rather than somewhere else?

We mostly can't tell what is widely used and what isn't. What about all the other very widely used modules then? Should we make types for all the return values of functions in those?

whitfin:
@fishcakez that's fair enough in that case. Out of interest, what would you recommend instead?

I think I'll close this then, thanks for feedback!

fishcakez:
> Out of interest, what would you recommend instead?

A registry, such as `Registry` (for non-atom names or duplicates) or `Process.register` (when 1 process per module or similarly static names)

PR #7521, opened 2018-03-31
Modify regex example output

desigrammer:
Noticed while using regex


josevalim:

PR #7523, opened 2018-04-01
Documents the :name option of start_link/2 [ci skip]

fxn:
Title says it all :).

Since I was on it, the patch also deletes the paragraph that repeates the default values of `:max_restarts` and `:max_seconds`, and revises the capitalization of the section title "Name registration" to match the existing convention.


josevalim:

PR #7527, opened 2018-04-03
Adding .elixir_ls to generated gitignore.

seivan:


lib/mix/lib/mix/tasks/new.ex
```diff
@@ -264,6 +264,9 @@ defmodule Mix.Tasks.New do
   # The directory Mix will write compiled artifacts to.
   /_build/
 
+  # The directory generated by Elixir Language Server (ElixirLS).
```
fertapric: This directory depends on the development environment of each user. That's why I wouldn't include it by default here.

I would suggest to include that into a [`global .gitignore`](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore).

josevalim:
As @fertapric said, editor specific ignores belong to global .gitignore.

Thank you @fertapric! <3

seivan:
@josevalim Sorry to bring this up again, but most users won't know about it until it hits them, are there any downsides of keeping it with the generate ignore? 

What are the downsides except duplicated mentions of the same folder in global and local ignore? 

I'm thinking, since you're already generating it, and that folder shouldn't necessarily(?) be committed, why not add it with it? I tend keep per project ignores separate from the global, where things like `.DS_Store` is kept.

lukaszsamson:
ElixirLS generates `.elixir_ls/.gitignore` so this is not a problem since long time. Not sure about other language servers

PR #7533, opened 2018-04-04
Add missing discard_threshold option to Logger configuration

hlhr:
Fix for #7532 

Add missing option `:discard_threshold` and unit test to test the correct implementation of `configure` for valid options

Issue #7532, opened 2018-04-04
Logger cannot set `discard_threshold`


pallix:
### Environment

* Elixir & Erlang versions (elixir --version): 
```
Erlang/OTP 20 [erts-9.3] [source] [64-bit] [smp:2:2] [ds:2:2:10] [async-threads:10] [kernel-poll:false]

Elixir 1.6.0 (compiled with OTP 19)
```
* Operating system: 

Linux

### Current behavior

I can set the `discard_threshold` value for the `Logger` configuration

```
iex(1)> Logger.configure(discard_threshold: 10_000)
:ok
iex(2)> Logger.Config.__data__()   
%{
  async_threshold: 15,
  discard_threshold: 500,
  keep_threshold: 375,
  level: :debug,
  mode: :async,
  sync_threshold: 20,
  translators: [{Logger.Translator, :translate}],
  truncate: 8096,
  utc_log: false
}
```

### Expected behavior

I can set the `discard_threshold` value for the `Logger` configuration, like the documentation indicates.

### Analysis

I think the `discard_threshold` key is missing in the [`valid_options` attribute list](https://github.com/elixir-lang/elixir/blob/v1.6.4/lib/logger/lib/logger.ex#L503).

hlhr:
Had some time to kill so I wrote a unit test for that issue, can confirm it for option `discard_threshold` and `translator_inspect_opts`. I can fix this with the described solution if this is ok?

pallix:
There may be the same problem for `:translator_inspect_opts`, also I don't see this option when printing `Logger.Config.__data__() ` so I am not sure how it is used.

PR #7536, opened 2018-04-07
Add missing translator_inspect_opts option to Logger configuration

hlhr:
Addtional fix for #7532 

After my fix for the option `:discard_threshold` in PR #7533 another missing option `:translator_inspect_opts` was reported. This option is [documented](https://hexdocs.pm/logger/Logger.html#module-runtime-configuration) but not part of the `Logger.Config.__data__()` ouput and not configurable with `Logger.configure/1`.

This PR adds `:translator_inspect_opts` to the output and makes it configurable with `Logger.configure/1`.

### DIFF
--- a/lib/logger/lib/logger.ex
+++ b/lib/logger/lib/logger.ex
@@ -513,7 +513,8 @@ defmodule Logger do
     :truncate,
     :level,
     :utc_log,
-    :discard_threshold
+    :discard_threshold,
+    :translator_inspect_opts
   ]
   @spec configure(keyword) :: :ok
   def configure(options) do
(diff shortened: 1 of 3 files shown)


josevalim:
We only need to add `:translator_inspect_opts` to the list of `@valid_options` and that should be it. There is no need to make it appear in `Logger.Config.__data__` since none of the callers of `Logger.Config.__data__` use it. :)

If you want to test that calling configure works, you can call `Application.get_env` directly in the test.

josevalim:
@hlhr see this commit: ae89ac8525d6bc4ea86dc820eab705f4e818966e where I structure things a bit better, so it should be clearer. :) The idea is that only a handful of those is persisted, only what is needed in the logger front-end.

Could you please rebase accordingly? Thank you!

hlhr:
Ok, thanks for your feedback :)

josevalim:
@hlhr just to clarify, a PR is still welcome, it is just that you need to do less work. :)

hlhr:
ok I thought you did this in your commit too. I had some problems with my git repository and will create a new PR for that.

Issue #7531, opened 2018-04-03
`mix compile` error with invalid or empty `apps_path`


talentdeficit:
### Environment

* Elixir & Erlang versions (elixir --version): 
```
Erlang/OTP 20 [erts-9.3] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:10] [hipe] [kernel-poll:false] [dtrace]

Elixir 1.6.4 (compiled with OTP 20)
```
* Operating system: 
```
17.0.4 Darwin
```

### Current behavior

```
$ cat mix.exs
defmodule Ask.Mixfile do
  use Mix.Project

  def project do
    [
      app: :apps_path_bug,
      version: "1.0.0",
      apps_path: "this_path_does_not_exist_or_is_empty"
    ]
  end
end
```

```
$ mix compile
** (Enum.EmptyError) empty error
    (elixir) lib/enum.ex:1843: Enum.reduce/2
    (mix) lib/mix/tasks/compile.ex:93: Mix.Tasks.Compile.run/1
    (mix) lib/mix/task.ex:314: Mix.Task.run_task/3
    (mix) lib/mix/cli.ex:80: Mix.CLI.run_task/2
    (elixir) lib/code.ex:677: Code.require_file/2
```

### Expected behavior

this should probably have a friendlier failure mode

JulioDeLeon:
I would like to work on this issue.

josevalim:
@JulioDeLeon go ahead. :)

soteras:
Is there a way to debug code with IEx.pry or IO.inspect?

I changed the code and after i run (make compile | make clean_elixir compile) but my code didn't change. I want to check the return of Mix.Task.run("compile.all", args).

josevalim:
How are you testing it? You need to make sure you are using the elixir and mix scripts from the bin directory. Something like: `bin/elixir bin/mix compile`. You can also leverage the test suite and let the tests do the verification for you. :) 

Other than that, there is a debugging chapter on the getting started chapter in the website.

PR #7537, opened 2018-04-08
Fix mix compile error when there is a wrong project path.

soteras:
This PR aims to solve the issue #7531, the problem is Mix.Task.run("compile.all", args) returns a empty list when project path is wrong. 

```
def project do
    [
      app: :apps_path_bug,
      apps_path: "this_path_does_not_exist_or_is_empty"
    ]
 end
```

### DIFF
--- a/lib/mix/lib/mix/tasks/compile.ex
+++ b/lib/mix/lib/mix/tasks/compile.ex
@@ -90,16 +90,18 @@ defmodule Mix.Tasks.Compile do
 
     load_erl_config(opts)
 
-    {res, diagnostics} =
-      case Mix.Task.run("compile.all", args) do
-        [] ->
-          {:error, "path_not_found"}
+    diagnostics =
+      Mix.Task.run("compile.all", args)
+      |> List.wrap()
+      |> Enum.map(&Mix.Task.Compiler.normalize(&1, :all))
 
-        projects ->
-          projects
-          |> List.wrap()
-          |> Enum.map(&Mix.Task.Compiler.normalize(&1, :all))
-          |> Enum.reduce(&merge_diagnostics/2)
+    {res, diagnostics} =
+      case diagnostics do
+        [] ->
+          Enum.reduce(diagnostics, {:ok, []}, &merge_diagnostics/2)
+
+        _ ->
+          Enum.reduce(diagnostics, &merge_diagnostics/2)
       end
 
     if res == :error and "--return-errors" not in args do
--- a/lib/mix/lib/mix/tasks/compile.ex
+++ b/lib/mix/lib/mix/tasks/compile.ex
@@ -91,10 +91,16 @@ defmodule Mix.Tasks.Compile do
     load_erl_config(opts)
 
     {res, diagnostics} =
-      Mix.Task.run("compile.all", args)
-      |> List.wrap()
-      |> Enum.map(&Mix.Task.Compiler.normalize(&1, :all))
-      |> Enum.reduce(&merge_diagnostics/2)
+      case Mix.Task.run("compile.all", args) do
+        [] ->
+          {:error, "path_not_found"}
+
+        projects ->
+          projects
+          |> List.wrap()
+          |> Enum.map(&Mix.Task.Compiler.normalize(&1, :all))
+          |> Enum.reduce(&merge_diagnostics/2)
+      end
 
     if res == :error and "--return-errors" not in args do
       exit({:shutdown, 1})
--- a/lib/mix/lib/mix/tasks/compile.ex
+++ b/lib/mix/lib/mix/tasks/compile.ex
@@ -90,19 +90,11 @@ defmodule Mix.Tasks.Compile do
 
     load_erl_config(opts)
 
-    diagnostics =
+    {res, diagnostics} =
       Mix.Task.run("compile.all", args)
       |> List.wrap()
       |> Enum.map(&Mix.Task.Compiler.normalize(&1, :all))
-
-    {res, diagnostics} =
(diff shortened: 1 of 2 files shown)


lib/mix/lib/mix/tasks/compile.ex
```diff
@@ -91,10 +91,16 @@ defmodule Mix.Tasks.Compile do
     load_erl_config(opts)
 
     {res, diagnostics} =
-      Mix.Task.run("compile.all", args)
-      |> List.wrap()
-      |> Enum.map(&Mix.Task.Compiler.normalize(&1, :all))
-      |> Enum.reduce(&merge_diagnostics/2)
```
josevalim: The simplest fix here would be to replace `Enum.reduce(&merge_diagnostics/2)` by `Enum.reduce({:ok, []}, &merge_diagnostics/2)`. Then we just ensure that calling the umbrella app does not raise.

soteras: Something like that?

```
diagnostics =
      Mix.Task.run("compile.all", args) |> List.wrap()
      |> Enum.map(&Mix.Task.Compiler.normalize(&1, :all))

    {res, diagnostics} =
      case diagnostics do
        [] ->
          Enum.reduce(diagnostics, {:error, []}, &merge_diagnostics/2)

        _ ->
          Enum.reduce(diagnostics, &merge_diagnostics/2)
      end
```

lib/mix/test/mix/tasks/compile_test.exs
```diff
@@ -9,6 +9,12 @@ defmodule Mix.Tasks.CompileTest do
     end
   end
 
+  defmodule WrongPath do
+    def project do
+      [app: :apps_path_bug, apps_path: "this_path_does_not_exist_or_is_empty"]
```
eksperimental: It's wrong the part of the atom that says its_empty. Maybe another module with empty path can be created and tested

lib/mix/lib/mix/tasks/compile.ex
```diff
@@ -90,11 +90,19 @@ defmodule Mix.Tasks.Compile do
...
+    diagnostics =
       Mix.Task.run("compile.all", args)
       |> List.wrap()
       |> Enum.map(&Mix.Task.Compiler.normalize(&1, :all))
-      |> Enum.reduce(&merge_diagnostics/2)
+
+    {res, diagnostics} =
+      case diagnostics do
+        [] ->
+          Enum.reduce(diagnostics, {:ok, []}, &merge_diagnostics/2)
```
josevalim: We can keep the code exactly the same as before, we just need to change one line where we replace Enum.reduce/2 by a Enum.reduce/3. :)

soteras: I need to change old tests. Do you agree?
Assertion with == failed
     code:  assert Mix.Task.run("compile", ["--verbose"]) == {:noop, []}
     left:  {:ok, []}
     right: {:noop, []}
     stacktrace:
       lib/mix/test/mix/tasks/compile_test.exs:54: anonymous fn/0 in Mix.Tasks.CompileTest."test compile a project with mixfile"/1
       (elixir) lib/file.ex:1428: File.cd!/2
       lib/mix/test/test_helper.exs:114: MixTest.Case.in_fixture/3
       lib/mix/test/mix/tasks/compile_test.exs:44: (test)

soteras: Maybe that:

Enum.reduce({:noop, []}, &merge_diagnostics/2)

lib/mix/test/mix/tasks/compile_test.exs
soteras: Yes i'll create

lib/mix/lib/mix/tasks/compile.ex
josevalim: Oh, I see. Yes, `|> Enum.reduce({:noop, []}, &merge_diagnostics/2)` should work!

lib/mix/test/mix/tasks/compile_test.exs
```diff
@@ -157,4 +169,20 @@ defmodule Mix.Tasks.CompileTest do
   after
     Application.delete_env(:erl_config_app, :value)
   end
+
+  test "compile a project with wrong path" do
```
eksperimental: Tests should be in present tense, not in imperative. So it should read:
"compiles a project"

lib/mix/test/mix/tasks/compile_test.exs
```diff
@@ -157,4 +169,20 @@ defmodule Mix.Tasks.CompileTest do
...
+    Mix.Project.push(WrongPath)
+
+    ExUnit.CaptureIO.capture_io(fn ->
+      assert {:noop, []} = Mix.Task.run("compile", ["--force", "--return-errors"])
+    end)
+  end
+
+  test "compile a project with empty path" do
```
eksperimental: it should read:
"compiles a project"

lib/mix/test/mix/tasks/compile_test.exs
soteras: English is not my first language but if you open the file lib/mix/test/mix/tasks/compile_test.exs there're a lot tests with "compile" they are wrong too?

eksperimental: Yes, they should be corrected.
I think it was myself and/or @lexmag who standardized all the tenses but those that you see have been introduced relatively recently.
You can see it has been done before in c6f83b5 for example

josevalim:

PR #7539, opened 2018-04-08
Propsed fix for issue #7537, mix compile errors when apps_path does n…

JulioDeLeon:
This PR addresses the issues found in https://github.com/elixir-lang/elixir/issues/7531. It setups up multiple fixtures where apps_path can be considered invalid or empty. The error message will now display if the value provided fo apps_path is a valid directory and if the directory is empty.


josevalim:
Thanks @JulioDeLeon but a PR has already been sent on #7537, so we will go ahead with it. :heart:

PR #7540, opened 2018-04-08
Add missing translator_inspect_opts option to Logger configuration

hlhr:
Addtional fix for #7532. Discussed in #7536. 

After my fix for the option :discard_threshold in PR #7533 another missing option `:translator_inspect_opts` was reported. This option is documented but not configurable with `Logger.configure/1`.

This PR adds makes it configurable with `Logger.configure/1`.


josevalim:

PR #7542, opened 2018-04-10
Use function module

NJichev:
After having `Function` module might as well make the internal code use it.

* change elixir code to use `Function.fun_info`
* point function type to this module now

Feel free to close this PR if you feel those changes are unnecessary!
Should I change the docs in kernel to reflect that the module is for normal functions too?
More forgotten stuff... :sweat_smile:


josevalim:

PR #7555, opened 2018-04-16
Deprecations table improvements

eksperimental:
- Add link to functions with multiple arities `Enum.chunk_every/2/3/4`
- Add link to special forms in and for
- Remove unncessary use of "Use" word
- Add "Use" word when necessary
- Remove unnecessary "instead" words
- Add descriptive words whether it is a "module attribute", "key" or "value"
- Add missing "AVAILABLE SINCE" version for @doc module attribute
- change to a more logical order:
  - `char_list/0` type
  - `:as_char_lists` value
  - `:char_lists` key
- Be explicit about use of `<%=` (not just `=`)


josevalim:

PR #7557, opened 2018-04-16
Refactors the test suite of compile.app

fxn:
Parses the resource files to make assertions against the generated term.

(This is preliminary work for another patch in which I am whitelisting the keys to avoid generating `extra_applications` in particular.)


lib/mix/test/mix/tasks/compile.app_test.exs
```diff
@@ -60,49 +66,60 @@ defmodule Mix.Tasks.Compile.AppTest do
...
+      assert properties[:applications] == [
+               :kernel,
+               :stdlib,
+               :elixir,
+               :logger,
+               :ok1,
+               :ok3,
+               :ok4,
+               :ok7
+             ]
```
josevalim: Please write this as:

```elixir
assert properties[:applications] ==
        [:kernel, :stdlib, :elixir, :logger, :ok1, :ok3, :ok4, :ok7]
```

Someetimes the formatter needs some hinting on which direction to take. :)

lib/mix/test/mix/tasks/compile.app_test.exs
```diff
@@ -191,34 +208,34 @@ defmodule Mix.Tasks.Compile.AppTest do
...
 
-    in_fixture "no_mixfile", fn ->
+    in_fixture("no_mixfile", fn ->
       message = "Expected :version to be a SemVer version, got: \"0.3\""
 
       assert_raise Mix.Error, message, fn ->
         Mix.Tasks.Compile.App.run([])
       end
-    end
+    end)
```
whatyouhide: Why were parens added to all `in_fixture/2` calls?

lib/mix/test/mix/tasks/compile.app_test.exs
```diff
@@ -3,6 +3,12 @@ Code.require_file("../../test_helper.exs", __DIR__)
 defmodule Mix.Tasks.Compile.AppTest do
   use MixTest.Case
 
+  def parse_resource_file(app) do
+    {:ok, [term]} = :file.consult("_build/dev/lib/#{app}/ebin/#{app}.app")
+    {:application, ^app, properties} = term
+    properties
+  end
```
whatyouhide: Personally I prefer to put helper functions at the bottom of the file, after all tests. But regardless of personal preference, this should be private.

lib/mix/test/mix/tasks/compile.app_test.exs
josevalim: Oh, probably because @fxn ran the formatter from inside lib/elixir and he needs to run from root.

@whatyouhide but honestly, I would just add parens to all in_fixture and all in_tmp calls. WDYT?

whatyouhide: @josevalim I agree, and remove them from `.formatter.exs`. Seems we avoided doing that initially to limit the already huge amount of changes. But I wouldn't do this here, I can do it as a separate PR/commit.

josevalim: Agreed. @fxn, feel free to leave the parens in the `in_fixture` calls. We will bring all of them back anyway in a later commit. Thank you!

fxn:
Hey guys! Revised everything according to your feedback.

Regarding formatting, I would normally have sent a clean patch, but since the code base is being unified I thought to leave them and see if you wanted them in.

@whatyouhide moved the helper down and made it private. I normally put these things up, because people read files from top to bottom. So, if you open this file and see there is a helper right there, that quickly communicates that the thing encapsulated in that helper is important and used later on. So you are set and ready to read the tests. The other workflow is to see the helper used and wonder, what is this doing?

Just explaining the rationale to share it with you, but the personal preferences have to match core preferences, so absolutely fine moving that one down.

PR #7560, opened 2018-04-17
Adds meta generated to for comprehension clauses with bitstrings

### LINKED ISSUE #7508: Dialyzer issue introduced in for expansion of Elixir 1.6.4
With 

```
Erlang/OTP 20 [erts-9.3] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:10] [hipe] [kernel-poll:false]
Elixir 1.6.4 (compiled with OTP 19)
```

the following example code

```
  def f() do
    for a <- [1,2,3], into: <<>>, do: <<a::big-integer-size(8)>>
  end
```
seems to expand to code for which Dialyzer issues the following warnings:
```
lib/ex16.ex:20: The variable _@1 can never match since previous clauses completely covered the type <<_:8>>
lib/ex16.ex:20: Cons will produce an improper list since its 2nd argument is <<_:8>>
```

With earlier versions of Elixir no warnings were issued for the code.

ggcampinho:
Fixes #7508


lexmag:
Thank you @ggcampinho.

PR #7565, opened 2018-04-18
Fix parsing of queries with just "?" in URI.parse/1

### LINKED ISSUE #7563: URI.parse incorrectly handling empty query strings
### Environment

I'm using `Erlang/OTP 20` and `Elixir (1.6.1)`.

### Current behavior

```elixir
URI.parse("/?") 
%URI{
  authority: nil,
  fragment: nil,
  host: nil,
  path: "/",
  port: nil,
  query: nil,
  scheme: nil,
  userinfo: nil
}
```

The query here is `nil` but I think it should be `""`

### Expected behavior

I was expecting the parse function to implement the reverse of the following behaviour

```elixir
iex(1)> "#{%URI{path: "/", query: ""}}"
"/?"
iex(2)> "#{%URI{path: "/", query: nil}}"
"/"
iex(3)> "#{%URI{path: "/", query: "a"}}"
"/?a"
```

For every other usecase (that I looked at) stringifying a parsed url would result in the original binary.
except this one.

```elixir
iex(5)> "/?" |> URI.parse |> URI.to_string
"/"

whatyouhide:
Closes #7563

### DIFF
--- a/lib/elixir/lib/uri.ex
+++ b/lib/elixir/lib/uri.ex
@@ -422,11 +422,13 @@ defmodule URI do
 
     parts = Regex.run(regex, string)
 
-    destructure [_, _, scheme, _, authority, path, _, query, _, fragment], parts
+    destructure [_, _, scheme, _, authority, path, query_with_question_mark, _, _, fragment],
+                parts
+
     scheme = nillify(scheme)
     authority = nillify(authority)
     path = nillify(path)
-    query = nillify(query)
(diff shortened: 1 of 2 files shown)


CrowdHailer:
Do these changes could as a patch change and therefore available in the next patch release? (1.6.5)

btw thanks for the quick work here. :whale:

josevalim:
I want to reserve it for 1.7 only because I can see people relying on this
behavior. :(
-- 

PR #7569, opened 2018-04-19
Fix typespecs for named records

### LINKED ISSUE #7530: Record typespecs are generated in a wrong way
### Environment
```
Erlang/OTP 20 [erts-9.3] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:10] [hipe] [kernel-poll:false] [dtrace]

Elixir 1.6.3 (compiled with OTP 20)
```

* Operating system: 
OS X High Sierra

### Current behavior
If you declare a record using: 
```elixir
defrecord :queue_bind,          :'queue.bind',          Record.extract(:'queue.bind',          from_lib: "rabbit_common/include/rabbit_framing.hrl")
```
And then want to create a typespec:
```elixir
 @type queue_bind :: record(:'queue_bind',
        ticket: term,
        queue: String.t,
        exchange: String.t,
        routing_key: String.t,
        nowait: term,
        arguments: term)
```
This typespec has no chances of working as the type it generates is:
```elixir
type: {:queue_bind,
   {:type, 32, :tuple,
    [
      {:atom, 0, :queue_bind},
      {:type, 33, :term, []},
      {:remote_type, 34, [{:atom, 0, String}, {:atom, 0, :t}, []]},
      {:remote_type, 35, [{:atom, 0, String}, {:atom, 0, :t}, []]},
      {:remote_type, 36, [{:atom, 0, String}, {:atom, 0, :t}, []]},
      {:type, 37, :term, []},
      {:type, 38, :term, []}
    ]}, []}
```

This will never match, as the record is in fact called `:'queue.bind'`. Declaring it with `queue.bind` raises a compilation error: 
```
== Compilation error in file lib/amqp_director/queues.ex ==
** (CompileError) lib/amqp_director/queues.ex:32: unknown record :"queue.bind"
    (elixir) lib/kernel/typespec.ex:1171: Kernel.Typespec.compile_error/2
    (elixir) lib/kernel/typespec.ex:470: Kernel.Typespec.translate_type/3
```

### Expected behavior
I would expect the `record` macro in typespec would take into account the option to rename records inside Elixir code.

Currently the workaround is to define two records, one for actual usage and one without renaming to use in the typespec.

ggcampinho:
Closes #7530

### DIFF
--- a/lib/elixir/lib/kernel/typespec.ex
+++ b/lib/elixir/lib/kernel/typespec.ex
@@ -477,26 +477,26 @@ defmodule Kernel.Typespec do
     typespec({:record, meta, [atom, []]}, vars, caller)
   end
 
-  defp typespec({:record, meta, [atom, fields]}, vars, caller) do
+  defp typespec({:record, meta, [tag, field_specs]}, vars, caller) do
     # We cannot set a function name to avoid tracking
     # as a compile time dependency because for records it actually is one.
-    case Macro.expand({atom, [], [{atom, [], []}]}, caller) do
-      keyword when is_list(keyword) ->
+    case Macro.expand({tag, [], [{}]}, caller) do
+      {_, _, [name, fields | _]} when is_list(fields) ->
         types =
-          Enum.map(keyword, fn {field, _} ->
-            Keyword.get(fields, field, quote(do: term()))
+          Enum.map(fields, fn {field, _} ->
+            Keyword.get(field_specs, field, quote(do: term()))
           end)
 
-        Enum.each(fields, fn {field, _} ->
-          unless Keyword.has_key?(keyword, field) do
-            compile_error(caller, "undefined field #{field} on record #{inspect(atom)}")
+        Enum.each(field_specs, fn {field, _} ->
+          unless Keyword.has_key?(fields, field) do
+            compile_error(caller, "undefined field #{field} on record #{inspect(tag)}")
           end
         end)
 
-        typespec({:{}, meta, [atom | types]}, vars, caller)
+        typespec({:{}, meta, [name | types]}, vars, caller)
 
       _ ->
-        compile_error(caller, "unknown record #{inspect(atom)}")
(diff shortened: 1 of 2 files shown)


lib/elixir/lib/kernel/typespec.ex
```diff
@@ -477,26 +477,26 @@ defmodule Kernel.Typespec do
     typespec({:record, meta, [atom, []]}, vars, caller)
   end
 
-  defp typespec({:record, meta, [atom, fields]}, vars, caller) do
+  defp typespec({:record, meta, [tag, field_specs]}, vars, caller) do
```
ggcampinho: I tried to match the naming with `Record.defrecord`

lib/elixir/lib/kernel/typespec.ex
```diff
@@ -477,26 +477,26 @@ defmodule Kernel.Typespec do
...
     # We cannot set a function name to avoid tracking
     # as a compile time dependency because for records it actually is one.
-    case Macro.expand({atom, [], [{atom, [], []}]}, caller) do
-      keyword when is_list(keyword) ->
+    case Macro.expand({tag, [], [{}]}, caller) do
```
josevalim: An empty tuple is not a valid AST. Maybe you mean a list with an empty list inside?

lib/elixir/lib/kernel/typespec.ex
josevalim: :+1:

lib/elixir/lib/kernel/typespec.ex
ggcampinho: The idea was to match only in the [last condition](https://github.com/elixir-lang/elixir/blob/master/lib/elixir/lib/record.ex#L323) avoiding to call this method twice (one without parameters and another one like it was before) because there is no return to the function that returns the name and the fields together. Since it's quoted I thought it was safe to do this, but I can replace for two calls.

PR #7571, opened 2018-04-20
Elixir wrapper for Erlang :global module

kluevandrew:
The PR provides wrapper for Erlang Global module which provides a global name registration facility.
More about Erlang module: http://erlang.org/doc/man/global.html

It is my first try to contribute elixir. If i break contribution policy when just tell me.


.gitignore
```diff
@@ -11,3 +11,6 @@
 /.eunit
 .elixir.plt
 erl_crash.dump
+
+# Ingore IDE configs
```
fertapric: I think IDE-specific files should be handled by the global `.gitignore` file on the machine and not by the Elixir project.

https://help.github.com/articles/ignoring-files/

kluevandrew: Ok, i can remove revert it. I added it out of habit

NobbZ:
Wouldn't this wrapper mean we had to use `{:via, Global, name}` instead of `{:global, name}` when naming processes?

Don't we have already `Registry` to register process names?

Also its really only wrapping, there is only a single `def`, everything else is `defdelegate`ing to `:global`.

In my opinion using `Global` over `:global` makes things more complicated instead of simpler.

fertapric:
First off, thanks for contributing! 💜 

Unfortunately, I think this module does not provide much value. It's an alias, similar behavior as `alias :global, as: Global`.

josevalim:
Hi @kluevandrew! In Elixir we tend to avoid wrapping Erlang modules only for wrapping them. We recommend developers to use `:global` directly.

PR #7574, opened 2018-04-20
Supports log formatter customization for metadata

adkron:
All log messages except those created by `Logger.bare_log/2` include
module, function, line, and file as attributes. These are very useful in
log messages. The problem with supporting them is that the data may not
be present. This forces the developer to write a custom logger or deal
with the metadata format `module=MyModyle function=my_fun/2`.

This format is hard to read since other places that we use this data it
is usually formatted as `MyModule.my_fun/2`.

In order to keep developers from having to write custom formatters for
basic output of values I thought that any parameter `$foo` could easily
pull from the metadata if it is present. This would keep from breaking
`Logger.bare_log/2` if the format included `$module` and keep it simple
to use.

In the case where the key does not exist in metadata it will be printed
as a string. So `$module` would be printed as `$module` if it isn't
found. That means if a developer wants a `$somthing` to show up in their
log messages that would also now work.

Amos King @adkron <amos@binarynoggin.com>

### DIFF
--- a/lib/logger/lib/logger/formatter.ex
+++ b/lib/logger/lib/logger/formatter.ex
@@ -23,6 +23,8 @@ defmodule Logger.Formatter do
     * `$metadata` - user controlled data presented in `"key=val key2=val2"` format
     * `$levelpad` - sets to a single space if level is 4 characters long,
       otherwise set to the empty space. Used to align the message after level.
+    * `$<any>`    - the value of the given key form the metadata. If there is
+      no matching key then the parameter is treated as a string.
 
   Backends typically allow developers to supply such control
   strings via configuration files. This module provides `compile/1`,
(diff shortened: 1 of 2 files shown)


adkron:
@fertapric, thanks I fixed and squashed that.

fertapric:
I think writing a custom formatter is the way to go here :)

As your last paragraph reflects, missing values require a special behaviour. In your case, you opted by writing the name of the key (`$something`), but others could prefer `-`, `””` or not writing anything. I don’t think Elixir should make that decision for you.

adkron:
The big thing I see is that module, function, line, and file are present in almost all log messages. The only time you don’t have them is in ‘bare_log’.  So they seem like common, simple pieces of data that should be accessible. The rest was just an extension of those same pieces.

michaelkschmidt:
@fertapric $module and $function are basic items to include in a log message.  It is a surprise that they aren't currently supported.

Generalizing this to support any meta data is a great move. It eliminates the need for custom formatters for trivial data like this.

josevalim:
@adkron but now the logging engine needs to handle $module not being an actual module, no? How would we address that?

josevalim:
The other concern is that, if you are using $metadata and $module, then the module information will be duplicated. Since this seems to be specific to a certain use case (i.e. modules are shown as $module and there is duplicated data), I would recommend to tackle it with your own custom formatter.

PR #7575, opened 2018-04-20
Fix AST for records with one field

ggcampinho:
I'm not very sure about the test or it's place

### DIFF
--- a/lib/elixir/lib/record.ex
+++ b/lib/elixir/lib/record.ex
@@ -370,8 +370,11 @@ defmodule Record do
         {new_fields, Keyword.delete(each_keyword, field)}
       end)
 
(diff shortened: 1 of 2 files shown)


josevalim:
@ggcampinho emitting `{:{}, _, [elem1, elem2]}` is not wrong per se. Or is there another reason for this change?

ggcampinho:
I thought it could cause problems if someone is trying to parse some AST and expecting the tuple with two elements only. It's a corner case, but it can happen. Feel free to close the PR if you think it's not necessary :)

josevalim:
I don't think it is necessary right now since both formats are supported but we can always reopen it based on more use cases.

Issue #4119, opened 2015-12-26
Mix.env/0 always returns prod for application dependencies


rbishop:
Elixir version: 1.1.1

I have a dependency ([honeybadger-elixir](https://honeybadger-io/honeybadger-elixir)) and when I'm compiling my application that makes use of that dependency, `Mix.env/0` returns `:prod` when that dependency is being compiled. This happens even if I set `MIX_ENV=dev` before running any `mix` or `iex` commands. When my application is compiling, `Mix.env/0` returns the correct environment.

nuss-justin:
This sounds correct to me. From the [Introduction to Mix](http://elixir-lang.org/getting-started/mix-otp/introduction-to-mix.html):

> Note: If you add dependencies to your project, they will not inherit your project’s environment, but instead run with their :prod environment settings!

rbishop:
Thank you for sharing that note. That seems like surprising behavior to me. Especially with a common pattern of using macros + environment to perform optimizations.

josevalim:
I would argue that the opposite would be much more surprising. Imagine every time you pool Phoenix in, it runs as a dependency without optimizations and bringing its development dependencies (like ex_doc). Dependencies are meant to run in prod. If you want the opposite, pass the `:env` option.

PR #7576, opened 2018-04-21
Allows path-based dependencies to share current env

bryannaegele:
All dependencies compile using `:prod` environment with the exception of umbrella dependencies, which share the current environment. The `in_umbrella` option is simply setting a default path of `../` and setting `env` to the current environment. 

As noted in #4119, a default of anything but `:prod` for dependencies would be unexpected. However, in the case of umbrella applications, the opposite is true, and the current environment is passed to the dependency. It would be helpful to allow the same functionality to path dependencies that `in_umbrella: true` provides without forcing the user to have that dependency as a sibling folder. This PR exposes an option to pass the mix env to the dependency to provide parity with the functionality `in_umbrella` provides.

This could be changed to use an `if` but I wasn't sure about the feelings in core on reassigning like this.

```
cond do
  raw = opts[:path] ->
    Keyword.put(opts, :dest, Path.expand(raw))
        
    if opts[:use_mix_env], do:  Keyword.put_new(opts, :env, Mix.env())

  opts[:in_umbrella] ->
    path = "../#{app}"

    opts
    |> Keyword.put(:dest, Path.expand(path))
    |> Keyword.put_new(:path, path)
    |> Keyword.put_new(:env, Mix.env())

  true ->
    nil
end
```

Or extracting for both `in_umbrella` and `use_mix_env` to a private function and adding it in there. I'm open to whatever works best.

```
def accepts_options(app, opts) do
  cond do
    raw = opts[:path] ->
      Keyword.put(opts, :dest, Path.expand(raw))
      |> add_env()

    opts[:in_umbrella] ->
      path = "../#{app}"

      opts
      |> Keyword.put(:dest, Path.expand(path))
      |> Keyword.put_new(:path, path)
      |> add_env()

    true ->
      nil
  end
end

defp add_env(opts) do
  Keyword.put_new(opts, :env, Mix.env())
end
```

### DIFF
--- a/lib/mix/lib/mix/scm/path.ex
+++ b/lib/mix/lib/mix/scm/path.ex
@@ -17,7 +17,14 @@ defmodule Mix.SCM.Path do
   def accepts_options(app, opts) do
     cond do
       raw = opts[:path] ->
-        Keyword.put(opts, :dest, Path.expand(raw))
+        case opts[:use_mix_env] do
+          true ->
+            Keyword.put(opts, :dest, Path.expand(raw))
+            |> Keyword.put_new(:env, Mix.env())
+
+          _ ->
+            Keyword.put(opts, :dest, Path.expand(raw))
+        end
 
       opts[:in_umbrella] ->
         path = "../#{app}"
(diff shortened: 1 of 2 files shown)


josevalim:
@bryannaegele in those cases, you should set `env: Mix.env` in the dependency in the mix.exs file. I believe there is no need for an extra option. Thanks!

bryannaegele:
Ah, yes! Such a simple solution. Thank you!

PR #7578, opened 2018-04-22
Add __STACKTRACE__/0 to Naming Conventions page and docs

eksperimental:
Additionally sort special forms alphabetically.


lib/elixir/lib/kernel/special_forms.ex
```diff
@@ -8,7 +8,7 @@ defmodule Kernel.SpecialForms do
   forms used to define tuple and binary data structures respectively.
 
   This module also documents macros that return information about Elixir's
-  compilation environment, such as (`__ENV__/0`, `__MODULE__/0`, `__DIR__/0` and `__CALLER__/0`).
+  compilation environment, such as (`__CALLER__/0`, `__DIR__/0`, `__ENV__/0`, `__MODULE__/0` and `__STACKTRACE__/0`).
```
michalmuskala: I'm not sure this is correct - `__STACKTRACE__` does not return information about compilation environment - it's more of a runtime thing. While it looks the same to the others, it's different in the functionality.

eksperimental: You are right @michalmuskala.
How about the new commit?

PR #7580, opened 2018-04-22
Improve String.printable?/2

eksperimental:
- Improve specs to reflect that when limit is 0, it will return true,
  and mention this in the docs.
- Add guards so it will fail when wrong types are give

### DIFF
--- a/lib/elixir/lib/string.ex
+++ b/lib/elixir/lib/string.ex
@@ -246,37 +246,36 @@ defmodule String do
   """
   @spec printable?(t, 0) :: true
   @spec printable?(t, pos_integer | :infinity) :: boolean
-
   def printable?(string, character_limit \\ :infinity)
       when is_binary(string) and
              (character_limit == :infinity or
                 (is_integer(character_limit) and character_limit >= 0)) do
-    do_printable?(string, character_limit)
+    recur_printable?(string, character_limit)
   end
 
-  defp do_printable?(_string, 0), do: true
-  defp do_printable?(<<>>, _character_limit), do: true
+  defp recur_printable?(_string, 0), do: true
+  defp recur_printable?(<<>>, _character_limit), do: true
 
   for char <- 0x20..0x7E do
-    defp do_printable?(<<unquote(char), rest::binary>>, character_limit) do
-      do_printable?(rest, decrement(character_limit))
+    defp recur_printable?(<<unquote(char), rest::binary>>, character_limit) do
+      recur_printable?(rest, decrement(character_limit))
     end
   end
 
   for char <- '\n\r\t\v\b\f\e\d\a' do
-    defp do_printable?(<<unquote(char), rest::binary>>, character_limit) do
-      do_printable?(rest, decrement(character_limit))
+    defp recur_printable?(<<unquote(char), rest::binary>>, character_limit) do
+      recur_printable?(rest, decrement(character_limit))
     end
   end
 
-  defp do_printable?(<<char::utf8, rest::binary>>, character_limit)
+  defp recur_printable?(<<char::utf8, rest::binary>>, character_limit)
        when char in 0xA0..0xD7FF
        when char in 0xE000..0xFFFD
        when char in 0x10000..0x10FFFF do
-    do_printable?(rest, decrement(character_limit))
+    recur_printable?(rest, decrement(character_limit))
   end
 
-  defp do_printable?(string, _character_limit) do
+  defp recur_printable?(string, _character_limit) do
     false
   end
 
--- a/lib/elixir/lib/string.ex
+++ b/lib/elixir/lib/string.ex
@@ -224,10 +224,10 @@ defmodule String do
   @conditional_mappings [:greek]
 
   @doc """
-  Checks if a string contains only printable characters.
+  Checks if a string contains only printable characters up to `character_limit`.
 
-  Takes an optional `limit` as a second argument. `printable?/2` only
-  checks the printability of the string up to the `limit`.
+  Takes an optional `character_limit` as a second argument. If `character_limit` is `0`, this
+  function will return `true`.
 
   ## Examples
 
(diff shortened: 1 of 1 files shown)


eksperimental:
I would like to mention that I think we should correct this function, and not take limit = 0,
since there is no point in doing so.

eksperimental:
@josevalim please review again

lib/elixir/lib/string.ex
```diff
@@ -240,37 +240,48 @@ defmodule String do
...
-  @spec printable?(t, non_neg_integer | :infinity) :: boolean
-  def printable?(string, counter \\ :infinity)
+  @spec printable?(t, 0) :: true
+  @spec printable?(t, pos_integer | :infinity) :: boolean
+
+  def printable?(string, character_limit \\ :infinity)
+      when is_binary(string) and
+             (character_limit == :infinity or
+                (is_integer(character_limit) and character_limit >= 0)) do
+    do_printable?(string, character_limit)
```
lexmag: I believe we're trying to not have `do_` prefixing, I wonder if we can come up with an alternative. 🤔

lib/elixir/lib/string.ex
```diff
@@ -240,37 +240,48 @@ defmodule String do
...
   """
-  @spec printable?(t) :: boolean
-  @spec printable?(t, non_neg_integer | :infinity) :: boolean
-  def printable?(string, counter \\ :infinity)
+  @spec printable?(t, 0) :: true
+  @spec printable?(t, pos_integer | :infinity) :: boolean
+
+  def printable?(string, character_limit \\ :infinity)
+      when is_binary(string) and
+             (character_limit == :infinity or
```
lexmag: Passing integer limit is likely more common case, I guess we should put it first.

eksperimental: I thought that calling `String.printable?/1` would be the most common case.
actually I cannot find anybody using the limit arg
https://github.com/search?utf8=%E2%9C%93&q=%22String.printable%22+language%3Aelixir&type=Code

lib/elixir/lib/string.ex
eksperimental: I could not come up with anything, but this is a common pattern. What can we name functions that take the same arguments as the public one, but we don't do any guard checks.
`printable?_checked`, `printable?_filtered`, `printable?_guarded` ???

lib/elixir/lib/string.ex
lexmag: Indeed, String.printable?/1` should be even more common case.

lib/elixir/lib/string.ex
josevalim: I thought the same but I have no good suggestion either.

lexmag: Maybe `printable_with_limit`?

lib/elixir/lib/string.ex
```diff
@@ -240,37 +240,48 @@ defmodule String do
...
+      iex> String.printable?("abc" <> <<0>>, 0)
+      true
+
   """
-  @spec printable?(t) :: boolean
-  @spec printable?(t, non_neg_integer | :infinity) :: boolean
-  def printable?(string, counter \\ :infinity)
+  @spec printable?(t, 0) :: true
+  @spec printable?(t, pos_integer | :infinity) :: boolean
+
```
whatyouhide: Since there seems to be another round of changes coming up, this line shouldn't be here, we don't leave an empty line between specs and function. It's not added by the formatter I hope :D

lib/elixir/lib/string.ex
eksperimental: Sorry, but that doesn't say anythig to me as the public one uses limit as well.
How about process_printable?, Or pass_, consume_, I'm sure somebody else can find a better verb to describe this action

lexmag: All the prefixed alternatives share the same issue as the `do_`—breaks grouping and navigation. Ideally we should go with suffix. What do you folks think about `printable_impl`? Still not ideal but at least helps to solve aforementioned issue.

josevalim:
I would vote for recur_printable?. The first definition is the guard, this
is the one doing the actual recursion.
-- 


lexmag:
@josevalim `recur_printable` is more clearer than generic `do_printable`, however it still breaks grouping and navigation, how about `printable_recur`?

josevalim:
@lexmag the function is right there after the other. I do not necessarily think they should all start with the same prefix. I would go with whatever reads better.

lexmag:
@josevalim though in the future they might go separately (like it happens in `Enum`), and we get those problems. To me both (`printable_recur` and `recur_printable`) reads equally. :bowtie:

eksperimental:
@lexmag what do you exactly mean by "grouping and navigation"?

eksperimental:
I like "recur" either as prefix or suffix.
What abour "_printable?"? Would we be breaking any convention or falling in an anti-pattern?

josevalim:
`_printable?` has the same issue as `do_printable?`. It is used as a clutch when better names could be used.

eksperimental:
Another issue with using "recur" as suffix is that the function ends with "?", So "printable_recur" should read "printable_recur?" or "printable?_recur"; the former breaks the name and may not be found when searching for the string "printable?", The latter breaks the convention of using the question mark as a suffix. I would choose the latter though, if we are to use recur as a suffix

josevalim:
`recur_printable?` is my vote. I don't have much to add beyond that. Question mark in the middle is not supported.

fertapric:
As the `do_printable?` function always gets the `character_limit`, what do you guys think about `printable_within_character_limit?` or `printable_within_char_limit?` (to shorten it a little bit)?

lexmag:
@eksperimental I mean search in code: find all helper functions for a public punction (`defp printable_` gives this), if some specific helper is needed you basically add one–two more symbols to search for it—`defp printable_re`. Also it is visually easier to comprehend what belong to which function.

We have examples where we went without having question mark in the name: https://github.com/elixir-lang/elixir/blob/master/lib/elixir/lib/enum.ex#L2846.

Ultimately, I think `recur_printable` is indeed better than `do_printable`, nevertheless I still prefer to have it named `printable_reccur`.

josevalim:
Folks, i am taking an executive decision here. :) Please name it `recur_printable?`. Thank you!

PR #7581, opened 2018-04-22
Replace Mixfile with MixProject in docs

fertapric:
`Mixfile` was replaced with `MixProject` in https://github.com/elixir-lang/elixir/commit/c6a0edcd32359eca284ff0166d19a6dc979ee954.

### DIFF
--- a/lib/mix/lib/mix/project.ex
+++ b/lib/mix/lib/mix/project.ex
@@ -330,9 +330,9 @@ defmodule Mix.Project do
   ## Examples
 
       Mix.Project.in_project(:my_app, "/path/to/my_app", fn module ->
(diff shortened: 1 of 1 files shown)


fertapric:
@eksperimental it depends on if we are talking about the module or the project itself. 

I went for the module because it's the argument of `fun` in `Mix.Project.in_project/4` and because that's what we are interpolating `"MixProject is: #{inspect module}"`.

Said that, both sound good to me, so I will change it if you think "Mix project is" is more accurate :)

josevalim:

eksperimental:
@fertapric yes, I was going to suggest "Mix project" or just "Project". Thank you

PR #7585, opened 2018-04-23
[WIP] Add diff to doctest

wasnotrice:
This adds diff output to doctests, as [discussed on elixir-core](https://groups.google.com/forum/#!topic/elixir-lang-core/qGMW6lCYo5w). 

I haven't figured out a good way to add more tests for these changes. I have tested the changes manually, and they produce good output on complex data structures like Maps. Output is diffed just like normal in ExUnit tests. Are the existing (updated) tests sufficient, do you think?

I thought of adding a test to `lib/ex_unit/examples/difference.exs`, which exercises highlighting, but since doctests require creating a BEAM file, I wasn't sure this was the best approach. The current file is simple and understandable, and I didn't want to lose that quality.

I also thought to add a case to the doctests suite, but this would only test that left and right values are included, and the other tests already do that. I don't think we can test the actual highlighting here.

Thoughts?

### DIFF
--- a/lib/ex_unit/lib/ex_unit/doc_test.ex
+++ b/lib/ex_unit/lib/ex_unit/doc_test.ex
@@ -322,7 +322,7 @@ defmodule ExUnit.DocTest do
 
         actual ->
           expr = "#{unquote(String.trim(expr))} === #{unquote(String.trim(expected))}"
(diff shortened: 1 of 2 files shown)


josevalim:

PR #7587, opened 2018-04-23
Add informations about `Code.compile_string/2` insecurities

hauleth:
Elixir can execute any valid code during compilation.  This can
compromise user machine.  Example of such code:

    defmodule Foo do
      @foo File.rm_rf(System.get_env("HOME"))
    end

Such code would remove whole `$HOME` of current user.

This PR introduces warning similar to the one used in
`Code.eval_string/3` that there is possibility for attacker to run code
via compilation (without evaluation).

### DIFF
--- a/lib/elixir/lib/code.ex
+++ b/lib/elixir/lib/code.ex
@@ -801,10 +801,10 @@ defmodule Code do
   given as second argument which will be used for reporting warnings
   and errors.
 
-  **Warning**: `string` can be any Elixir code and some code can be executed
-  with the same privileges as the Elixir compiler: this means that such code
+  **Warning**: `string` can be any Elixir code and code can be executed with
+  the same privileges as the Elixir compiler: this means that such code
   could compromise the machine (for example by executing system commands).
-  Don't use `oompile_string/2` with untrusted input (such as strings coming
+  Don't use `compile_string/2` with untrusted input (such as strings coming
   from the network).
   """
   @spec compile_string(List.Chars.t(), binary) :: [{module, binary}]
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/code.ex
```diff
@@ -800,6 +800,12 @@ defmodule Code do
   and the second one is its bytecode (as a binary). A `file` can be
   given as second argument which will be used for reporting warnings
   and errors.
+
+  **Warning**: `string` can be any Elixir code and some code can be executed
+  with the same privileges as the Elixir compiler: this means that such code
+  could compromise the machine (for example by executing system commands).
+  Don't use `oompile_string/2` with untrusted input (such as strings coming
```
eksperimental: compile_string is misspelt. I would use commas and not parentheses.

eksperimental: Also, it is not that "some code" can be executed with elixirc privileges. "All" the code is executed like that

hauleth: @eksperimental not all code is run during compilation, that was the case for this, not that some will be executed with `elixirc` privileges.

hauleth: Fixed wording.

hauleth:
@eksperimental it is mentioned in `Code.eval_string/3` but with `eval_*` family it is more obvious that "here be dragons" than with `compile_*` as not all languages allows to run any code during compilation only.

josevalim:

lib/elixir/lib/code.ex
```diff
@@ -802,8 +802,8 @@ defmodule Code do
   and errors.
 
   **Warning**: `string` can be any Elixir code and code can be executed with
```
eksperimental: I think here it is correct to say, as it is said the `Code.eval_string/3`, that "**will** be executed with
  the same privileges as the Erlang VM"

PR #7589, opened 2018-04-24
clarify behavior of `:temporary` processes within supervisor strategies

davidsulc:
The documentation for the `:one_for_all` supervisor strategy indicates

> if a child process terminates, all other child processes are terminated

But this is not entirely true. Imagine a supervisor S with a `:one_for_all` strategy. S has 2 children: a `:permanent` process P, and a `:temporary` process T. Killing T will never cause P to get restarted event though that is what seems to be indicated by the documentation (T is a child of S, after all, so P should be terminated given it is an "other child" of the same supervisor).

Hopefully, this commit clarifies that `:temporary` processes are essentially ignored by supervisors with respect to restarting the temporary process or its siblings.


josevalim:

PR #7590, opened 2018-04-24
Fix typespec for DynamicSupervisor.on_start_child

lasseebert:
`DynamicSupervisor.start_child/2` is typespec'ed to return the same type as `Supervisor.start_child/2`, but it can also return `:ignore` and `{:error, :max_children}`

This PR fixes that.


josevalim:

PR #7592, opened 2018-04-24
Optimise String.starts_with?/2

michalmuskala:
The new implementation is 10x to 150x faster on the benchmark data.

The full benchmark: https://gist.github.com/michalmuskala/b8dea400a9902356f5bc53c201161a05

### DIFF
--- a/lib/elixir/lib/string.ex
+++ b/lib/elixir/lib/string.ex
@@ -1897,17 +1897,29 @@ defmodule String do
       true
 
   """
-  @spec starts_with?(t, t | [t]) :: boolean
-  def starts_with?(string, []) when is_binary(string) do
-    false
+  @spec starts_with?(t, pattern) :: boolean
+  def starts_with?(string, prefix) when is_binary(string) and is_binary(prefix) do
+    starts_with_string?(string, byte_size(string), prefix)
   end
 
   def starts_with?(string, prefix) when is_binary(string) and is_list(prefix) do
-    "" in prefix or Kernel.match?({0, _}, :binary.match(string, prefix))
+    string_size = byte_size(string)
+    Enum.any?(prefix, &starts_with_string?(string, string_size, &1))
   end
 
   def starts_with?(string, prefix) when is_binary(string) do
-    "" == prefix or Kernel.match?({0, _}, :binary.match(string, prefix))
+    Kernel.match?({0, _}, :binary.match(string, prefix))
+  end
+
+  @compile {:inline, starts_with_string?: 3}
+  defp starts_with_string?(string, string_size, prefix) when is_binary(prefix) do
+    prefix_size = byte_size(prefix)
+
+    if prefix_size <= string_size do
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/string.ex
```diff
@@ -1898,16 +1898,27 @@ defmodule String do
...
-  def starts_with?(string, []) when is_binary(string) do
-    false
+  def starts_with?(string, prefix) when is_binary(string) and is_binary(prefix) do
+    starts_with_string?(string, byte_size(string), prefix)
   end
 
   def starts_with?(string, prefix) when is_binary(string) and is_list(prefix) do
-    "" in prefix or Kernel.match?({0, _}, :binary.match(string, prefix))
+    string_size = byte_size(string)
+    Enum.any?(prefix, &starts_with_string?(string, string_size, &1))
```
michalmuskala: I decided to always use the recursion with `starts_with_string?` instead of `:binary.match`. I couldn't find an example where `:binary.match` would be faster. See the linked benchmark files for what I've tried.

lib/elixir/lib/string.ex
```diff
@@ -1897,17 +1897,29 @@ defmodule String do
...
   end
 
   def starts_with?(string, prefix) when is_binary(string) do
```
fertapric: Looking at the doc and the specs, the first two `starts_with?/2` should catch all the possible valid combinations (a string and a list of strings). In which cases this function will be executed? 🤔

fertapric: Nevermind. I saw `:binary.cp()` in the `pattern` typespec.

michalmuskala: Yes, it wasn't previously documented with the typespecs, but it's allowed.

lib/elixir/lib/string.ex
```diff
@@ -1897,17 +1897,29 @@ defmodule String do
...
-    "" == prefix or Kernel.match?({0, _}, :binary.match(string, prefix))
+    Kernel.match?({0, _}, :binary.match(string, prefix))
+  end
+
+  @compile {:inline, starts_with_string?: 3}
+  defp starts_with_string?(string, string_size, prefix) when is_binary(prefix) do
+    prefix_size = byte_size(prefix)
+
+    if prefix_size <= string_size do
+      prefix === binary_part(string, 0, prefix_size)
```
whatyouhide: Any reason to use `===` here? As far as I know, it wouldn't make a difference for binaries?

michalmuskala: There was a bug in `==` where it wouldn't compare sizes before starting the comparison, but it doesn't really matter here since the size is always the same.

whatyouhide: @michalmuskala Right, so it looks like we can drop it here? I think for clarity `==` is always better as `===` automatically triggers my smell as I expect it's being used for a particular reason.

PR #6560, opened 2017-09-10
Features/allow multiple registries in supervision tree

yordis:
Related to #6556 

cc: @josevalim

Issue #6556, opened 2017-09-10
Error when starting multiple Registries in supervision tree


jlevycpa:
I have a need to start two different Registry processes in my supervision tree. My list of children includes the following:

```
...
{Registry, [keys: :unique, name: Wif.ConnectionRegistry]},
{Registry, [keys: :unique, name: Wif.TriggerRegistry]},
...
```
I received this error when I compiled:
`** (Mix) Could not start application wif: Wif.Application.start(:normal, []) returned an error: bad child specification, more than one child specification has the id: Registry.`

It looks like [`Registry.child_spec/1`](https://github.com/elixir-lang/elixir/blob/dbcc36ddb25782155cf6c7712019554b599b41d3/lib/elixir/lib/registry.ex#L305) includes the fixed value of `Registry` as the child id, which would prevent multiple Registry processes from being started under the same supervisor without pre-1.5 style child specs or some other workaround. 

I propose that instead, when the `name` option is used, the `name` is used as the child id. Alternatively, a separate `id` parameter could be passed. 

Please let me know if you would be open to a PR or if I'm missing something.

josevalim:
That's a great suggestion. :+1: A PR will definitely be welcome.

yordis:
@josevalim how would you do the testing for this https://github.com/yordis/elixir/commit/9b410159aecca43ef6d2a718ac17f8409705c833 ?

I could just open that function and test the lookup or I could setup the whole environment with the supervision tree so I can see it pass

josevalim:
@yordis for the test, calling `child_spec(opts)` on the Regitry defined in the tests themselves should be enough.

jlevycpa:
Just went to put the PR together, but I see you beat me to it!

yordis:
@jlevycpa sorry I saw this was 5 hours ago when I did it!

jlevycpa:
@yordis no worries. I wrote this on my way out and just got back.

whatyouhide:
Closed by https://github.com/elixir-lang/elixir/pull/6560.

PR #7594, opened 2018-04-24
Allow multiple DynamicSupervisors to be started easily

lmarlow:
Use the provided name for the id of the child_spec if there is one, a la
how Registry does it. Without this you need to provide a more complete
child_spec Map with id, type, and start fields.

### DIFF
--- a/lib/elixir/lib/dynamic_supervisor.ex
+++ b/lib/elixir/lib/dynamic_supervisor.ex
@@ -194,7 +194,7 @@ defmodule DynamicSupervisor do
       case Keyword.get(arg, :name, DynamicSupervisor) do
         name when is_atom(name) -> name
         {:global, name} -> name
-        {:via, _, name} -> name
+        {:via, _module, name} -> name
       end
 
     %{


josevalim:
Thanks @lmarlow! I like this although the question is what happens when the name is not an atom but an entry such as global or, even worse, a via tuple.

We could still use those as IDs... but I thought I would bring this up for discussion.

lmarlow:
Are IDs ever useful to look at or pass around? They seem to be mostly opaque items to me. It wouldn't matter what is in them as long as they are consistent and unique within its supervisor.

The motivation for this was the same as #6556, resolved by #6560.

josevalim:
@lmarlow IDs are used in regular supervisor. That's how you refer to their children. But most of the times you want to refer to them by id giving them an explicit id is likely the way to go.

josevalim:
@fishcakez sounds good to me! @lmarlow what do you think? can you please do the changes above?

lmarlow:
That's fine with me.

What's the benefit of using just `name` instead of `{:global, name}`, though? `Supervisor` uses it to track which of its specs are running, but a developer would use the name they provided to start it when interacting with it.

PR #7597, opened 2018-04-25
Inline Code.ensure_loaded/1

eksperimental:


michalmuskala:
We generally inline only BIFs - `:code.ensure_loaded` is a plain erlang function. I don't think we need to inline it.

PR #7600, opened 2018-04-25
Include endianness in the regex version

josevalim:
This addresses issues with precompiled archives on different
endianness, such as: https://github.com/hexpm/hex/issues/469


josevalim:
@ericmj that is fine. When v1.5 was released, pretty much all archives were doing a recompilation. Given the use case here is archives, it is most likely that the regex recompilation is not a major performance issue. If it is, you should not be using regexes. :)

josevalim:

PR #7601, opened 2018-04-25
Erase elixir_parser_columns from the process dictionary after use

ggcampinho:
Notice this while reading the parser


ggcampinho:
One thing that I noticed is that the `string_to_quoted!` affects the process dictionary. Probably nobody is using those atoms, but we could restore the old state if we stored the old values, I don't know if it makes sense

josevalim:
I think erasing is fine, since the keys are namespace. No need to try to revert it.

josevalim:

PR #7603, opened 2018-04-25
Ignore unused var in String.ex

fertapric:
Removes warning:

```
warning: variable "string" is unused
  lib/string.ex:278
```


whatyouhide:
Thanks @fertapric! 😍

eksperimental:
Thank you

PR #7608, opened 2018-04-26
Warn when a variable that ends in ! or ? is followed by = with no space

### LINKED ISSUE #7583: Operator != behaves as a match operator if there is no space
### Precheck

* For proposing a new feature, please start a discussion on the Elixir Core mailing list

### Environment

```
Erlang/OTP 20 [erts-9.2] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:10] [hipe] [kernel-poll:false]

```

### Current behavior

I couldn't find any issue about this - I hope it's not a duplicate. Here is a sample that show two different behaviors for what I would have expected to be the `!=` operator:

```
iex(1)> x=1
1
iex(2)> x==2
false
iex(3)> x != 2
true
iex(4)> x!= 2
2
iex(5)> x!
2
```
### Expected behavior

I would have expected `!=` to have higher precedence than `=` (`x!= 2` to behave the same as `x != 2`).

whatyouhide:
Closes #7583.

### DIFF
--- a/lib/elixir/src/elixir_tokenizer.erl
+++ b/lib/elixir/src/elixir_tokenizer.erl
@@ -506,6 +506,7 @@ tokenize(String, Line, Column, Scope, Tokens) ->
   case tokenize_identifier(String, Line, Scope) of
     {Kind, Atom, Rest, Length, Ascii, Special} ->
       HasAt = lists:member($@, Special),
+      HasBangOrQuestionMark = lists:member($?, Special) orelse lists:member($!, Special),
 
       case Rest of
         [$: | T] when ?is_space(hd(T)) ->
@@ -520,6 +521,10 @@ tokenize(String, Line, Column, Scope, Tokens) ->
           {error, Reason, String, Tokens};
         _ when Kind == alias ->
           tokenize_alias(Rest, Line, Column, Atom, Length, Ascii, Special, Scope, Tokens);
+        [$= | _] when HasBangOrQuestionMark, Kind == identifier ->
+          Msg = "use a space before = if the previous identifier ends with ! or ?",
+          elixir_errors:warn(Line, Scope#elixir_tokenizer.file, Msg),
+          tokenize_other(Rest, Line, Column, Atom, Length, Scope, Tokens);
         _ when Kind == identifier ->
           tokenize_other(Rest, Line, Column, Atom, Length, Scope, Tokens);
         _ ->
--- a/lib/elixir/src/elixir_tokenizer.erl
+++ b/lib/elixir/src/elixir_tokenizer.erl
@@ -506,27 +506,39 @@ tokenize(String, Line, Column, Scope, Tokens) ->
   case tokenize_identifier(String, Line, Scope) of
     {Kind, Atom, Rest, Length, Ascii, Special} ->
       HasAt = lists:member($@, Special),
-      HasBangOrQuestionMark = lists:member($?, Special) orelse lists:member($!, Special),
 
       case Rest of
         [$: | T] when ?is_space(hd(T)) ->
           Token = {kw_identifier, {Line, Column, nil}, Atom},
           tokenize(T, Line, Column + Length + 1, Scope, [Token | Tokens]);
+
         [$: | T] when hd(T) /= $: ->
           AtomName = atom_to_list(Atom) ++ [$:],
           Reason = {Line, "keyword argument must be followed by space after: ", AtomName},
           {error, Reason, String, Tokens};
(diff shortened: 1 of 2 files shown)


lib/elixir/src/elixir_tokenizer.erl
```diff
@@ -506,6 +506,7 @@ tokenize(String, Line, Column, Scope, Tokens) ->
   case tokenize_identifier(String, Line, Scope) of
     {Kind, Atom, Rest, Length, Ascii, Special} ->
       HasAt = lists:member($@, Special),
+      HasBangOrQuestionMark = lists:member($?, Special) orelse lists:member($!, Special),
```
whatyouhide: I'm not confident this is the best method of finding out if an identifier ends with `?` or `!`.

michalmuskala: Maybe lists:last would be better here?

whatyouhide: @michalmuskala I started with that but `Special` would be `"!@"` for `foo@bar!` IIRC (I tried earlier and I think this was the result).

lib/elixir/src/elixir_tokenizer.erl
```diff
@@ -520,6 +521,10 @@ tokenize(String, Line, Column, Scope, Tokens) ->
           {error, Reason, String, Tokens};
         _ when Kind == alias ->
           tokenize_alias(Rest, Line, Column, Atom, Length, Ascii, Special, Scope, Tokens);
+        [$= | _] when HasBangOrQuestionMark, Kind == identifier ->
+          Msg = "use a space before = if the previous identifier ends with ! or ?",
+          elixir_errors:warn(Line, Scope#elixir_tokenizer.file, Msg),
+          tokenize_other(Rest, Line, Column, Atom, Length, Scope, Tokens);
         _ when Kind == identifier ->
           tokenize_other(Rest, Line, Column, Atom, Length, Scope, Tokens);
```
josevalim: Why not add a case and check for the last entry only inside this clause then?

lib/elixir/src/elixir_tokenizer.erl
```diff
@@ -520,6 +521,10 @@ tokenize(String, Line, Column, Scope, Tokens) ->
           {error, Reason, String, Tokens};
         _ when Kind == alias ->
           tokenize_alias(Rest, Line, Column, Atom, Length, Ascii, Special, Scope, Tokens);
+        [$= | _] when HasBangOrQuestionMark, Kind == identifier ->
+          Msg = "use a space before = if the previous identifier ends with ! or ?",
```
josevalim: I would do something like:

```erlang
    io_lib:format("found an identifier ~ts, ending with ~ts, followed by =. It is unclear if you mean \"~ts ~ts=\" or \"~ts =". Please add a space before or after ~ts to remove the ambiguity", [Identifier, [Last], IdentifierWithoutLast, [Last], Identifier, [Last], [Last]])
```

That's because there are ways to solve the ambiguity, so we should be clear about them.

lib/elixir/src/elixir_tokenizer.erl
josevalim: Basically, let's check for the last only when we are sure we have an identifier.

whatyouhide:
Updated with support for atoms too!

lib/elixir/src/elixir_tokenizer.erl
```diff
@@ -531,6 +538,26 @@ tokenize(String, Line, Column, Scope, Tokens) ->
...
+maybe_warn_for_ambiguous_bang_before_equals(Kind, Atom, [$= | _], Scope, Line) ->
+  {What, Identifier} =
+    case Kind of
+      atom -> {"atom", [$: | atom_to_list(Atom)]};
+      identifier -> {"identifier", atom_to_list(Atom)}
+    end,
+
+  case lists:last(Identifier) of
+    Last when Last == $!; Last == $? ->
+      Msg = io_lib:format("found an ~ts \"~ts\", ending with ~ts, followed by =. "
```
josevalim: We need to remove the article "an", since we are specifying the atom/identifier, and I would wrap `ending with ~ts` as `ending with "~ts"`.

josevalim:
One last comment and we should be good to go!

lib/elixir/src/elixir_tokenizer.erl
whatyouhide: Good catches, pushed!

PR #7609, opened 2018-04-27
Leave blank line before closing comment if idented

eksperimental:
If last line in the docs is idented, leave an emptyline before closing the comment.

Mentioned by @whatyouhide in https://github.com/elixir-lang/elixir/pull/7606#discussion_r184348514

The command/regexp used to find these offending lines is:

    ag "\ {4,}[^\n]+\n\s{2}[\"']{3}"

### DIFF
--- a/lib/elixir/lib/access.ex
+++ b/lib/elixir/lib/access.ex
@@ -762,6 +762,7 @@ defmodule Access do
 
       iex> get_in(%{}, [Access.filter(fn a -> a == 10 end)])
       ** (RuntimeError) Access.filter/1 expected a list, got: %{}
+
   """
   @since "1.6.0"
   @spec filter((term -> boolean)) :: access_fun(data :: list, get_value :: list)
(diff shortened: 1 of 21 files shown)


lib/mix/lib/mix/tasks/new.ex
```diff
@@ -408,6 +408,7 @@ defmodule Mix.Tasks.New do
   # here (which is why it is important to import them last).
   #
   #     import_config "#{Mix.env}.exs"
+  #
```
whatyouhide: We shouldn't have this line in this template.

lib/mix/lib/mix/tasks/new.ex
```diff
@@ -428,6 +429,7 @@ defmodule Mix.Tasks.New do
   #       level: :info,
   #       format: "$date $time [$level] $metadata$message\n",
   #       metadata: [:user_id]
+  #
```
whatyouhide: We shouldn't have this line in this template.

lib/mix/lib/mix/tasks/new.ex
eksperimental: true

lib/mix/lib/mix/tasks/new.ex
eksperimental: oops

eksperimental:
@whatyouhide fixed

josevalim:

PR #7610, opened 2018-04-27
Inline hard-deprecated functions

eksperimental:
These functions were inlined in the past. I think they should still
be inlined until removed in v2.0

I see they were removed in an unrelated commit 9c0d6a66d5fc05868fa0e7cdbe180c0a478b0547

before:

    iex(1)> &Atom.to_charlist/1
    &:erlang.atom_to_list/1
    iex(2)> &Atom.to_char_list/1
    warning: Atom.to_char_list/1 is deprecated, use Atom.to_charlist/1
      iex:2

    &Atom.to_char_list/1

now:

    iex(1)> &Atom.to_charlist/1
    &:erlang.atom_to_list/1
    iex(2)> &Atom.to_char_list/1
    warning: Atom.to_char_list/1 is deprecated, use Atom.to_charlist/1
      iex:2

    &:erlang.atom_to_list/1

### DIFF
--- a/lib/elixir/lib/atom.ex
+++ b/lib/elixir/lib/atom.ex
@@ -39,6 +39,7 @@ defmodule Atom do
 
   # TODO: Remove by 2.0
   # (hard-deprecated in elixir_dispatch)
+  # Inlined by the compiler
   @doc false
   @spec to_char_list(atom) :: charlist
   def to_char_list(atom), do: Atom.to_charlist(atom)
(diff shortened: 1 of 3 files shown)


josevalim:
Inlining is an optimization and it is not really necessary for deprecated code. We document it to avoid confusion and that's it. We should have no breaking change when inlining or not, so I don't see a reason to inline deprecated APIs. Thank you!

eksperimental:
That's true, but my point is that users count on a function being inlined and optimized, and later because Elixir changed its mind on some naming convention or something, the same code won't be optimized if they upgrade Elixir, and I don't think it's up to the to update their code. but that's my main arguing point.
anyway we are talking about literally three lines of code.

PR #7613, opened 2018-04-27
Remove unnecessary module call in Kernel.to_char_list/1

eksperimental:


michalmuskala:
Wouldn't it cause problems if the `Kernel.to_charlist` is not imported in the using module?

eksperimental:
@michalmuskala Is this what you mean?

```elixir
defmodule Foo do
  import Kernel, except: [to_charlist: 1]
  def bar, do: {to_char_list(:good), Kernel.to_char_list(:good)}
end

Foo.bar
#=> {'good', 'good'}
```

josevalim:
@eksperimental yes. It may only work in this case because Kernel is bootstrapped and compiled when Kernel is already defined but generally speaking modules are unable to apply hygiene on functions defined by themselves, so this should stay. :) Thank you!

PR #7616, opened 2018-04-28
Standardize terminology for SECURITY FIXES

eksperimental:


josevalim:
Good point, I have standarized on patches on master, since i believe that is the most common term outside Elixir. thank you!

eksperimental:
For some reason I think "security fixes" sounds more professional than "security patches" :-P

PR #7618, opened 2018-04-28
Standardize use of Erlang, OTP and Erlang/OTP names

eksperimental:
This touches only messages directed to the end user,
So comments are (mostly) ignored.

This is important, specially for new commers, who may be confused
by the interchangeably use of Erlang and OTP.

The expressions used consistenly across the documentation are:
- Erlang/OTP version
- Erlang/OTP 20+
- Erlang/OTP release,
- Erlang/OTP documentation
- Erlang process
- Erlang project
- Erlang's error logger
- Erlang log message
- OTP application
- OTP behaviour

### DIFF
--- a/lib/mix/lib/mix/tasks/compile.app.ex
+++ b/lib/mix/lib/mix/tasks/compile.app.ex
@@ -17,11 +17,11 @@ defmodule Mix.Tasks.Compile.App do
 
   The most commonly used options are:
 
-    * `:extra_applications` - a list of Erlang/Elixir applications
+    * `:extra_applications` - a list of OTP applications
       your application depends on which are not included in `:deps`
       (usually defined in `deps/0` in your `mix.exs`). For example,
       here you can declare a dependency on applications that ship with
-      Erlang or Elixir, like `:crypto` or `:logger`, but anything in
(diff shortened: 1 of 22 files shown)


lib/mix/lib/mix/tasks/compile.app.ex
```diff
@@ -17,11 +17,11 @@ defmodule Mix.Tasks.Compile.App do
 
   The most commonly used options are:
 
-    * `:extra_applications` - a list of Erlang/Elixir applications
+    * `:extra_applications` - a list of OTP/Elixir applications
```
josevalim: I think we should either say Erlang/Elixir or OTP applications but not "OTP/Elixir" applications.

josevalim:
One comment and we should be good to go! :heart:

lib/mix/lib/mix/tasks/compile.app.ex
eksperimental: Let's go with "OTP applications" because in the next sentence, it mentions "For example, here you can declare a dependency on applications that ship with Erlang/OTP or Elixir"

eksperimental:
pushed

whatyouhide:
Thanks @eksperimental! 💟

PR #7619, opened 2018-04-28
Group and sort alphabetically Deprecations table

eksperimental:
This is a big change. 
Please build the docs and tell me what you think.

It happens that as of now the deprecation table is big enough and It will only get bigger with time.
As of now, the table is sorted in reverse order by deprecation version, but If you want to have a look at what has been deprecated, it's really hard.
See how things are organized now. Still if you want to see what go deprecated in the latest version you have the links to each CHANGELOG

feedback is appreciated


josevalim:
I don't think that's the direction we want to go. I would say the goal of this table is not to find if something is deprecated but rather to find out what has been deprecated since a certain release. To find if something is deprecated, we should start leveraging the `@deprecated "reason"` attribute.

whatyouhide:
Agreed with @josevalim, `@deprecated` should be the way to go. Thanks anyways @eksperimental! 💟

PR #7620, opened 2018-04-28
Add .formatter.exs to the list of inputs to format

fertapric:
I saw this practice in [Plug](https://github.com/elixir-plug/plug/blob/master/.formatter.exs#L25) and I always include it in my `.formatter.exs` files.


whatyouhide:
Good catch, thanks @fertapric! 💟

PR #7621, opened 2018-04-28
Add missing paren in config.exs template

fertapric:


josevalim:

Issue #7512, opened 2018-03-28
Improve warning for grouped clauses


josevalim:
We should say that Foo and Bar have different names. When they have the same name, we should say they expect a different number of arguments (i.e. arity).

https://elixirforum.com/t/compiler-warning-clauses-for-the-same-def-should-be-grouped-together/13366/2

zacck:
@josevalim  any hints for a first-time contributor to go hunting for this?

josevalim:
@zacck this is where it is raised: https://github.com/elixir-lang/elixir/blob/master/lib/elixir/src/elixir_def.erl#L399

I would say the current warning is good for functions of different names but probably not enough for only when the arity changes.

aaruel:
Would this be a change in the wording of the existing warning or a separate warning?

josevalim:
A separate warning.

zacck:
@josevalim so I am trying to find out where the `format_error` function is called?

zacck:
Hey @josevalim  I have given it a few bashes and some trials with a bit of effort and help, here is what I came up with 
-  match for the condition
```erlang
check_valid_clause(Meta, File, Name, Arity, Kind, Set, StoredMeta, StoredFile) ->
  case ets:lookup_element(Set, ?last_def, 2) of
    none -> ok;
    {Name, Arity} -> ok; # name and arity here match with arguments 3 lines up
    {Name, OtherArity} -> 
       Relative = elixir_utils:relative_to_cwd(StoredFile),
       elixir_errors:form_warn(Meta, File, ?MODULE,
       {grouped_clause, {Kind, Name, Arity, OtherArity, ?line(StoredMeta), Relative}})
    _ ->
      Relative = elixir_utils:relative_to_cwd(StoredFile),
      elixir_errors:form_warn(Meta, File, ?MODULE,
        {ungrouped_clause, {Kind, Name, Arity, ?line(StoredMeta), Relative}})
  end.
```

- throw this as the issue 
```erlang
format_error({grouped_clause, {Kind, Name, Arity, OtherArity, OrigLine, OrigFile}}) ->
  io_lib:format("clauses with the same name should be grouped together, ~ts ~ts/~B was previously defined (~ts:~B)",
```
Would this suffice? 

Also I am finding it a little difficult to locate the tests for the check valid clause function any pointers to where these would be?

PR #7630, opened 2018-04-30
Provide better warnings when grouping multi clause functions

zacck:
##  Chore 
Relates #7512 
Providing a separate warning for cases where just the function name matches rather than the default warning. 

##  Test Plan 
Added a new unit test to cover the new  case

## Notes
I left the new arity of the function in the arguments in case we need to use it in the warning to the user

### DIFF
--- a/lib/elixir/src/elixir_def.erl
+++ b/lib/elixir/src/elixir_def.erl
@@ -301,8 +301,12 @@ check_valid_kind(Meta, File, Name, Arity, Kind, StoredKind) ->
 
 check_valid_clause(Meta, File, Name, Arity, Kind, Set, StoredMeta, StoredFile) ->
   case ets:lookup_element(Set, ?last_def, 2) of
-    {Name, Arity} -> ok;
     none -> ok;
+    {Name, Arity} -> ok;
+    {Name, OtherArity} ->
+      Relative = elixir_utils:relative_to_cwd(StoredFile),
+      elixir_errors:form_warn(Meta, File, ?MODULE,
+        {grouped_clause, {Kind, Name, OtherArity, ?line(StoredMeta), Relative}});
     _ ->
       Relative = elixir_utils:relative_to_cwd(StoredFile),
       elixir_errors:form_warn(Meta, File, ?MODULE,
--- a/lib/elixir/src/elixir_def.erl
+++ b/lib/elixir/src/elixir_def.erl
@@ -303,14 +303,14 @@ check_valid_clause(Meta, File, Name, Arity, Kind, Set, StoredMeta, StoredFile) -
   case ets:lookup_element(Set, ?last_def, 2) of
     none -> ok;
     {Name, Arity} -> ok;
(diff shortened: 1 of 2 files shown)


lib/elixir/src/elixir_def.erl
```diff
@@ -396,6 +400,10 @@ format_error({clauses_with_defaults, {Kind, Name, Arity}}) ->
     "\n"
    "~ts ~ts/~B has multiple clauses and defines defaults in one or more clauses", [Kind, Name, Arity]);
 
+format_error({grouped_clause, {Kind, Name, _OtherArity, OrigLine, OrigFile}}) ->
+   io_lib:format("clauses with the same name should be grouped together, ~ts ~ts was previously defined (~ts:~B)",
```
whatyouhide: I think we should wrap the `def foo` part in quotes (it reads a bit confusing in the error message in the test below). Maybe we can include the arity too?

> "def foo/2" was previously defined

Thoughts?

zacck: I think the arity will help in identifying which clause needs to be corrected and I  agree with you on the quotes let me put in some time and I'll push those fixes

josevalim: I think we can improve both warnings.

Let's call the first `ungrouped_arity` and the other one `ungrouped_name`. For the arity, we should say:

> clauses with the same name and arity (number of arguments) should be grouped together, \"~ts ~ts/~B\" was previously defined (~ts:~B)

For `ungrouped_name`, we will say:

> clauses with the same name should be grouped together, \"~ts ~ts/~B\" was previously defined (~ts:~B)

lib/elixir/src/elixir_def.erl
```diff
@@ -301,8 +301,12 @@ check_valid_kind(Meta, File, Name, Arity, Kind, StoredKind) ->
 
 check_valid_clause(Meta, File, Name, Arity, Kind, Set, StoredMeta, StoredFile) ->
   case ets:lookup_element(Set, ?last_def, 2) of
-    {Name, Arity} -> ok;
     none -> ok;
+    {Name, Arity} -> ok;
+    {Name, OtherArity} ->
+      Relative = elixir_utils:relative_to_cwd(StoredFile),
+      elixir_errors:form_warn(Meta, File, ?MODULE,
+        {grouped_clause, {Kind, Name, OtherArity, ?line(StoredMeta), Relative}});
```
josevalim: We should pass `Arity` here and not `OtherArity` and make use of `Arity` in the error message.

zacck: Ah

lib/elixir/src/elixir_def.erl
zacck: @josevalim  woop woop! yours looks way cooler I like this one

michalmuskala:
Thank you! :heart:

zacck:
Its been a pleasure, hunting the next issue!

PR #7635, opened 2018-05-02
keys and values of decoded query are always binary

CrowdHailer:
I have warnings that dialyzer could have picked up because when taking values for the map it assumes they can be of any term.


michalmuskala:
Thank you! :heart:

PR #7642, opened 2018-05-03
Normalize trace config docs for Mix.Tasks.Test

mmartinson:
I noticed this while reading though some of the source code, was surprised it wasn't documented, then  realized it was in fact documented, but not in the place I had first looked.


josevalim:

Issue #7627, opened 2018-04-30
Provide "rescue var not in [Exprs]"


michalmuskala:
This should nicely complement the current behavoiour when we want to rescue all exceptions except for some particular kind. Example use-case: 

https://github.com/elixir-plug/plug/blob/c547f2bd7a64aec7c4759dcc1ed88b6eab0c7499/lib/plug/parsers/multipart.ex#L32-L44

With `not in` this could look as:

```elixir
    try do
      parse_multipart(conn, opts)
    rescue
      # Wrap all errors except Plug.UploadError
      e not in Plug.UploadError ->
        reraise Plug.Parsers.ParseError.exception(exception: e), System.stacktrace()
    end
```

Doing this should be more efficient than reraising, since the VM could skip building the symbolic representation of the stacktrace in case it does not match.

whatyouhide:
I'll take a stab at this.

PR #7643, opened 2018-05-04
Add support for "rescue var not in [Exprs]"

whatyouhide:
Addresses #7627.

This is a first stab. The idea is that instead of compiling `in` to a list of guards (that in Erlang AST are equivalent to `Guard1; Guard2; ...`), we build a big chain of `orelse`s as in `Guard1 orelse Guard2`. This way we can implement `not in [Exprs]` by just negating the chain of `orelse`s.

I'm also raising an error for `var not in _` as I don't think it makes sense, but let me know what you think.

Thoughts appreciated!

### DIFF
--- a/lib/elixir/src/elixir_clauses.erl
+++ b/lib/elixir/src/elixir_clauses.erl
@@ -263,6 +263,18 @@ expand_rescue({in, Meta, [Left, Right]}, E) ->
       false
   end;
 
+%% rescue var not in _ (throws an error)
+expand_rescue({'not', Meta, [{in, _, [_Left, {'_', _, UnderscoreContext}]}]}, E)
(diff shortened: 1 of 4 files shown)


whatyouhide:
@michalmuskala yeah I was waiting for that because the number of tests would grow even more than it did. I was waiting for approvals on the implementation and then I'll add a few more tests.

michalmuskala:
@whatyouhide I looked at the assmebly and I think the part for erlang errors is not correct - it seems that one is not negated at all.

whatyouhide:
@michalmuskala you are correct, I had missed that. Adding tests and trying to fix it.

whatyouhide:
We decided to not go ahead with this feature :).

PR #7646, opened 2018-05-04
Fix logger and stacktrace warnings on OTP 21

josevalim:
Those are the minimum fixes which preserve the existing behaviour.
More performant fixes in the logger cases will be applied to master
and the upcoming Elixir v1.7.


josevalim:

PR #7648, opened 2018-05-05
Move to __STACKTRACE__ and lazily compute stacktrace

josevalim:
Erlang/OTP 20 warns if the stacktrace is read outside of a
catch/rescue. This commit mirrors this behaviour by consistently
warning on `System.stacktrace/0` being used outside of a
catch/rescue.

Erlang/OTP 21 warns whenever System.stacktrace/:erlang.get_stacktrace
are used. Therefore we need to promote the usage of `__STACKTRACE__`
and make sure to conditionally compile it according to the OTP version.
This requires changes to the Exception normalization mechanism
so we compute `__STACKTRACE__` only when strictly required.

In future Elixir releases, `System.stacktrace/0` will warn when
used even inside catch/rescue.

### DIFF
--- a/lib/elixir/lib/exception.ex
+++ b/lib/elixir/lib/exception.ex
@@ -2,10 +2,8 @@ defmodule Exception do
   @moduledoc """
   Functions to format throw/catch/exit and exceptions.
 
-  Note that stacktraces in Elixir are updated on throw,
-  errors and exits. For example, at any given moment,
-  `System.stacktrace/0` will return the stacktrace for the
-  last throw/error/exit that occurred in the current process.
+  Note that stacktraces in Elixir are only available inside
+  catch and rescue by using the `__STACKTRACE__/0` variable.
 
   Do not rely on the particular format returned by the `format*`
   functions in this module. They may be changed in future releases
@@ -74,11 +72,6 @@ defmodule Exception do
     end
   end
 
-  @deprecated "Use normalize/3 with an explicit stacktrace instead"
-  def normalize(kind, payload) do
-    normalize(kind, payload, nil)
-  end
-
   @doc """
   Normalizes an exception, converting Erlang exceptions
   to Elixir exceptions.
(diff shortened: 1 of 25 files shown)


lib/elixir/lib/exception.ex
```diff
@@ -1200,7 +1172,7 @@ defmodule ErlangError do
 
   def normalize({:badkey, key}, stacktrace) do
     term =
-      case ensure_stacktrace(stacktrace) do
+      case stacktrace do
         [{Map, :get_and_update!, [map, _, _], _} | _] -> map
         [{Map, :update!, [map, _, _], _} | _] -> map
```
michalmuskala: Do we really need the two clauses? I think we should make the functions in `Map` module raise full errors instead of examining the stacktrace afterwards.

josevalim: @michalmuskala so the idea was to make all of them consistent with Erlang/OTP. But I don't think we need to follow this rule. We can tackle it in a separate PR.

lib/elixir/src/elixir_erl_try.erl
```diff
@@ -19,102 +19,145 @@ reduce_clauses([], Acc, OldStack, SAcc, _S) ->
...
+  {'case', Meta, [
+    Var,
+    [{do, [
+      {'->', Meta, [[{'when', Meta, [Var, Guards]}], {'__STACKTRACE__', Meta, nil}]},
+      {'->', Meta, [[{'_', Meta, nil}], []]}
+    ]}]
+  ]}.
+
+normalize_with_stacktrace() ->
+  ['Elixir.FunctionClauseError', 'Elixir.UndefinedFunctionError', 'Elixir.KeyError', 'Elixir.ErlangError'].
```
michalmuskala: We only need the stacktrace for the two-tuple version of `badkey`

michalmuskala:
I'm not sure about always catching stacktraces for some errors. Building a symbolic stacktrace is explicitly documented as an expensive operation. I'm not sure we should be doing it for the user automatically. We didn't do that before.

I think it's fine to use the stacktrace in normalisation, if the user asks to build the stacktrace anyway, but I don't think we should do it on our own. The question is how should a user signal this, when they don't use the stacktrace otherwise.

One downside of always building a stacktrace for normalization is that we completely invalidate the optimisation that the erlang compiler performs when you do something like:
```erl
try
  ...
catch
  K:R:S -> erlang:raise(K,manipulate_r(R),S)
end
```
i.e. - only use the stacktrace in a call to `erlang:raise/3`. The compiler will skip building a symbolic stacktrace representation in such a case.

michalmuskala:
I should correct my post. We actually did that in case of `function_clause`, `undef` and `{badkey, _}` before by calls to `erlang:get_stacktrace/0`. I'm still not convinced that's the right thing to do, but I guess it's fine for those cases - it's not doing anything we weren't doing before.

josevalim:
@michalmuskala well, the goal for now is to keep everything backwards compatible. We can discuss mechanisms to relax those rules later on.

josevalim:
@michalmuskala i have pushed a fix so we build a `__STACKTRACE__` less frequently and by effectively treating ErlangError as a meta exception that is always dynamic.

josevalim:

lib/elixir/src/elixir_dispatch.erl
```diff
@@ -329,11 +329,24 @@ elixir_imported_macros() ->
...
 %% Inline common cases.
 check_deprecation(Meta, ?kernel, to_char_list, 1, E) ->
-  elixir_errors:warn(?line(Meta), ?key(E, file), "Use Kernel.to_charlist/1 instead");
+  Message = "Kernel.to_char_list/1 is deprecated. Use Kernel.to_charlist/1 instead",
+  elixir_errors:warn(?line(Meta), ?key(E, file), Message);
 check_deprecation(_, ?kernel, _, _, _) ->
   ok;
 check_deprecation(_, erlang, _, _, _) ->
   ok;
+check_deprecation(Meta, 'Elixir.System', stacktrace, 0, #{contextual_vars := Vars} = E) ->
```
michalmuskala: I just realised - should we also warn for `:erlang.get_stacktrace`? Otherwise we suppress the Erlang's warning completely.

josevalim: Erlang get_stacktrace is getting its own warning, we are not touching it. The suppression is only on the v1.6 branch.

PR #7649, opened 2018-05-06
Erlang/OTP 21 logger integration

josevalim:


michalmuskala:
Could we get the optimisations to `Enum` and `Keyword` in separate commit/PR? I think this will help keep the discussion focused.

josevalim:
This is ready for review. I just need to write tests for the new handler and it is ready to go.

lib/logger/lib/logger/erlang_handler.ex
```diff
@@ -0,0 +1,132 @@
...
+      end
+    end)
+  end
+
+  defp erlang_level_to_elixir_level(:emergency), do: :error
+  defp erlang_level_to_elixir_level(:alert), do: :error
+  defp erlang_level_to_elixir_level(:critical), do: :error
+  defp erlang_level_to_elixir_level(:error), do: :error
+  defp erlang_level_to_elixir_level(:warning), do: :warn
+  defp erlang_level_to_elixir_level(:info), do: :info
```
sirihansen: Missing `notice`.

josevalim: @sirihansen Thanks! I am glad you *noticed* that.

josevalim:

josevalim:
I have merged this because it causes some intermittent failures in our suite, making it really hard to work on other Erlang/OTP 21 features. Please drop any other feedback in this thread and I will get to it. :)

sirihansen:
Just one minor doc-thing - logger.ex, line 574 - should probably remove "both":

  ```
* `:flush` - when `true`, guarantees all messages currently sent
  +      to both `Logger` are processed before the backend is removed
```

josevalim:
@sirihansen thanks!

PR #7650, opened 2018-05-06
Add helpful error message for improperly formatted type definitions

### LINKED ISSUE #7647: If you use `@type` when you meant to use `@spec`, error messages are not very helpful and your project won't build
### Environment

jadams ~/e/specbug master λ elixir --version                                                                             4s 190ms
Erlang/OTP 20 [erts-9.3] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:10] [hipe] [kernel-poll:false] [dtrace]

Elixir 1.6.1 (compiled with OTP 19)

* Operating system: 
OS X

### Current behavior

I've created a reproduction [here](https://github.com/dailydrip/specbug) that you can use to see what happens.

```elixir
defmodule Specbug do
  # This should say spec, but your brain betrays you
  @type hello(String.t()) :: String.t()
  def hello(world) do
    world
  end
end
```

Wrong code like the above ^^ yields the following error:

```
Compiling 1 file (.ex)

== Compilation error in file lib/specbug.ex ==
** (CompileError) lib/specbug.ex: internal error in lint_module;
crash reason: badarg

  in function  atom_to_list/1
     called as atom_to_list({'.',[{line,2}],
      [{'__aliases__',[{line,2},{counter,-576460752303422973}],['String']},
       t]})
  in call from erl_lint:'-check_type/2-fun-0-'/3 (erl_lint.erl, line 2808)
  in call from dict:fold_bucket/3 (dict.erl, line 460)
  in call from dict:fold_seg/4 (dict.erl, line 456)
  in call from dict:fold_segs/4 (dict.erl, line 452)
  in call from lists:foldl/3 (lists.erl, line 1263)
  in call from erl_lint:forms/2 (erl_lint.erl, line 689)
  in call from erl_lint:module/3 (erl_lint.erl, line 534)
```

And it's really hard to go from that error to the code that caused it.

### Expected behavior

Something other than badarg? It's possible this is a special case, but I'm not exactly new to Elixir and my stupidity got me stuck for a bit here because my brain refused to read the line's problem.

devonestes:
The original issue (#7647) was posted in a very specific context, but I
figured that this implementation would both address that issue as well
as provide helpful feedback if someone were to attempt to write types
with variables incorrectly in general, since that was what was being
raised here.

Resolves #7647

### DIFF
--- a/lib/elixir/lib/kernel/typespec.ex
+++ b/lib/elixir/lib/kernel/typespec.ex
@@ -127,8 +127,6 @@ defmodule Kernel.Typespec do
 
   def deftypespec(kind, expr, line, file, module, pos)
       when kind in [:type, :typep, :opaque] do
-    validate_type_ast!(expr)
-
     {set, bag} = :elixir_module.data_tables(module)
 
     case type_to_signature(expr) do
@@ -152,17 +150,6 @@ defmodule Kernel.Typespec do
 
     store_typespec(bag, :type, {kind, expr, pos})
   end
-
-  defp validate_type_ast!({:::, _, [{type_name, _, [{{:., _, _}, _, _}]}, _]}) do
-    raise ArgumentError,
-      message: """
-      Type definition for `#{type_name}` is malformed.
-      It appears you are using a concrete type instead of a variable in the type
-      definition.
-      """
-  end
-
-  defp validate_type_ast!(_), do: nil
 
   defp get_typespec(bag, key) do
     :ets.lookup_element(bag, key, 2)
@@ -225,8 +212,9 @@ defmodule Kernel.Typespec do
     :ets.take(bag, key)
   end
 
-  defp translate_type({_, {kind, {:::, _, [{name, _, args}, definition]}, pos}})
+  defp translate_type({_, {kind, expr = {:::, _, [{name, _, args}, definition]}, pos}})
        when is_atom(name) and name != ::: do
+    validate_type_ast!(expr)
     caller = :elixir_locals.get_cached_env(pos)
 
     args =
@@ -261,6 +249,27 @@ defmodule Kernel.Typespec do
     type_spec = Macro.to_string(other)
     compile_error(caller, "invalid type specification: #{type_spec}")
   end
+
+  defp validate_type_ast!({:::, _, [{type_name, _, arguments}, _]}) when is_list(arguments) do
+    unless Enum.all?(arguments, &valid_variable_ast?/1) do
+      raise ArgumentError,
+        message: """
+        Type definition for `#{type_name}` is malformed.
+        It appears you are using a concrete type instead of a variable in the type
+        definition.
+        """
+    end
+  end
+
+  defp validate_type_ast!(_), do: nil
+
+  defp valid_variable_ast?({variable_name, _, nil}) when is_atom(variable_name), do: true
+
+  defp valid_variable_ast?({variable_name, _, module})
+       when is_atom(variable_name) and is_atom(module),
+       do: true
+
+  defp valid_variable_ast?(_), do: false
 
   defp translate_spec({kind, {{:when, _meta, [spec, guard]}, pos}}) do
     caller = :elixir_locals.get_cached_env(pos)
--- a/lib/elixir/lib/kernel/typespec.ex
+++ b/lib/elixir/lib/kernel/typespec.ex
@@ -129,6 +129,11 @@ defmodule Kernel.Typespec do
       when kind in [:type, :typep, :opaque] do
     {set, bag} = :elixir_module.data_tables(module)
 
+    if invalid_type_ast?(expr) do
+      message = message_for_invalid_type_ast(expr, file)
+      raise(ArgumentError, message: message)
+    end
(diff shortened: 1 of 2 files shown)


lib/elixir/lib/kernel/typespec.ex
```diff
@@ -129,6 +129,11 @@ defmodule Kernel.Typespec do
       when kind in [:type, :typep, :opaque] do
     {set, bag} = :elixir_module.data_tables(module)
 
+    if invalid_type_ast?(expr) do
+      message = message_for_invalid_type_ast(expr, file)
+      raise(ArgumentError, message: message)
+    end
```
michalmuskala: What if we merged it all together?

```elixir
defp validate_type_ast!({:::, _, [{type_name, _, [{{:., _, _}, meta, _}]}, _]}) do
  raise ArgumentError, """
  ...
  """
end
```

I don't think we need to fetch the line explicitly - if we raise, it should indicate the line somewhere.

josevalim:
Thanks @devonestes!

There is actually nothing wrong with defining a type before a function of the same name. What is really wrong in the code posted by @knewter was that one of the arguments given to the type definition was not a variable.

So we need to make sure all of the type arguments have a variable AST. Somewhere around the `translate_type` function.

josevalim:
Apologies, it seems my early feedback was partially incorrect.

Your current fix is correct but it is too restrict. It only checks for
function of a single argument and for remote types. Instead we should fail
if any of the arguments is not a variable.

We should also do the check later in the code as mentioned earlier.
-- 


devonestes:
Got it! I've updated to check to be sure that any arguments passed to a type definition are variables, and if any of them are not variables, then it will raise the exception.

josevalim:
Thanks @devonestes! We are still checking for arguments too early though. We need to move the check to inside `translate_type` and make sure to call `compile_error/2` when it comes to erroring. See the others `compile_error/2` calls in this file for examples.

lib/elixir/lib/kernel/typespec.ex
```diff
@@ -249,6 +250,27 @@ defmodule Kernel.Typespec do
...
+        Type definition for `#{type_name}` is malformed.
+        It appears you are using a concrete type instead of a variable in the type
+        definition.
+        """
+    end
+  end
+
+  defp validate_type_ast!(_), do: nil
+
+  defp valid_variable_ast?({variable_name, _, nil}) when is_atom(variable_name), do: true
```
josevalim: We don't need this clause as nil is also atoms.

lib/elixir/test/elixir/typespec_test.exs
```diff
@@ -907,6 +907,26 @@ defmodule TypespecTest do
       end)
     end
 
+    test "fails with a helpful error when non-variables are given as arguments" do
+      assert_raise ArgumentError, ~r/one_bad_variable/, fn ->
```
josevalim: Let's assert a bit more on the error message text.

josevalim:
Regarding the error message, note try just start with lowercase, no
backticks and we cannot use heredocs as they add new lines. I would suggest
something like this:

> @type definitions expect all arguments to be variables. The type
#{type_name}/#{arity} has an invalid argument: #{Macro.to_string(arg)}


-- 


lib/elixir/lib/kernel/typespec.ex
```diff
@@ -212,8 +212,9 @@ defmodule Kernel.Typespec do
     :ets.take(bag, key)
   end
 
-  defp translate_type({_, {kind, {:::, _, [{name, _, args}, definition]}, pos}})
+  defp translate_type({_, {kind, expr = {:::, _, [{name, _, args}, definition]}, pos}})
        when is_atom(name) and name != ::: do
+    validate_type_ast!(expr)
```
lexmag: Since we are only interested in validating arguments, I think we can name it more explicitly and pass less data downstream:
```
validate_type_args!(type_name, arguments)
```

lib/elixir/lib/kernel/typespec.ex
```diff
@@ -249,6 +250,27 @@ defmodule Kernel.Typespec do
...
+      raise ArgumentError,
+        message: """
+        Type definition for `#{type_name}` is malformed.
+        It appears you are using a concrete type instead of a variable in the type
+        definition.
+        """
+    end
+  end
+
+  defp validate_type_ast!(_), do: nil
```
lexmag: Looks like this clause will never match, right?

lib/elixir/test/elixir/typespec_test.exs
```diff
@@ -907,6 +907,26 @@ defmodule TypespecTest do
       end)
     end
 
+    test "fails with a helpful error when non-variables are given as arguments" do
```
lexmag: Maybe `test "non-variables are given as arguments" do` would read equally good. Saying "helpful" is quite subjective.

devonestes:
Thanks for the good recommendation on an error message - much better!

michalmuskala:
Thank you! :heart:

PR #7651, opened 2018-05-06
Add better ExUnit formatting when missing head of list

### LINKED ISSUE #6768: ExUnit diff of two lists is confusing when head is missing
### Environment

```
Erlang/OTP 20 [erts-9.0] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:10] [hipe] [kernel-poll:false]

Elixir 1.5.1
```

* Operating system: 

OS X 10.12.6

### Current behavior

Given this failing test, where the difference between two lists is a missing element at the end:

``` elixir
  test "diffing date lists with a missing date at end" do
    dates_1 = [~D[2017-06-01], ~D[2017-06-02], ~D[2017-06-03]]
    dates_2 = [~D[2017-06-01], ~D[2017-06-02]                ]

    assert dates_1 == dates_2
  end
```

...it produces a really nice, clear diff:

screen shot 2017-10-09 at 3 14 12 pm

...but given this failing test, where the first element is missing:

``` elixir
  test "diffing date lists with a missing date at start" do
    dates_1 = [~D[2017-06-01], ~D[2017-06-02], ~D[2017-06-03]]
    dates_2 = [                ~D[2017-06-02], ~D[2017-06-03]]

    assert dates_1 == dates_2
  end
```

...it produces a confusing diff:

screen shot 2017-10-09 at 3 15 28 pm

### Expected behavior

It would be great if the diff for the latter case was more like the first one, showing that the difference is the missing first element, instead of individually diffing the list elements position by position.  Not sure how feasible that is, but figured I'd ask :).

devonestes:
The current implementation of the Myers algorithm for diffing didn't
give very helpful diff information when comparing lists in the ExUnit
formatter. This is a very naieve implementation, and at the moment it
is only for lists, but I can see that we might conceivably want to have
this "missing head" diff show up for strings as well.

For making just this work, I sort of skipped the whole Myers algorithm
and just hand built the diff script since it follows a pretty simple
pattern. If we want to expand this further, though, we might need to
implement the reverse of the Myers algorithm.

Resolves #6768

### DIFF
--- a/lib/ex_unit/lib/ex_unit/diff.ex
+++ b/lib/ex_unit/lib/ex_unit/diff.ex
@@ -185,8 +185,8 @@ defmodule ExUnit.Diff do
     end
   end
 
-  defp take_common_path([h | tail1], [h | tail2], acc) do
-    take_common_path(tail1, tail2, [h | acc])
+  defp take_common_path([head | tail1], [head | tail2], acc) do
+    take_common_path(tail1, tail2, [head | acc])
   end
 
   defp take_common_path(list1, list2, acc), do: {list1, list2, acc}
--- a/lib/ex_unit/lib/ex_unit/diff.ex
+++ b/lib/ex_unit/lib/ex_unit/diff.ex
@@ -159,11 +159,40 @@ defmodule ExUnit.Diff do
   end
 
   defp script_proper_list(list1, length1, list2, length2, keywords?) do
-    initial_path = {0, 0, list1, list2, []}
+    case diff_subset(list1, list2) do
+      {:ok, diff} ->
+        format_each_fragment(diff, [], keywords?)
 
-    find_script(0, length1 + length2, [initial_path], keywords?)
-    |> format_each_fragment([], keywords?)
+      :error ->
+        initial_path = {0, 0, list1, list2, []}
+
+        find_script(0, length1 + length2, [initial_path], keywords?)
+        |> format_each_fragment([], keywords?)
+    end
   end
+
+  def diff_subset(list1, list2) do
+    {list1, list2, prefix} = take_common_path(list1, list2, [])
+    {list1, list2, suffix} = take_common_path(Enum.reverse(list1), Enum.reverse(list2), [])
+
+    if list1 == [] or list2 == [] do
+      {:ok,
+       wrap_in(:eq, Enum.reverse(prefix)) ++
+         wrap_in(:del, Enum.reverse(list1)) ++
+         wrap_in(:ins, Enum.reverse(list2)) ++ wrap_in(:eq, suffix)}
+    else
+      :error
+    end
+  end
+
+  defp take_common_path([h | tail1], [h | tail2], acc) do
+    take_common_path(tail1, tail2, [h | acc])
+  end
+
+  defp take_common_path(list1, list2, acc), do: {list1, list2, acc}
+
+  defp wrap_in(_, []), do: []
+  defp wrap_in(tag, items), do: [{tag, items}]
 
   defp format_each_fragment([{:diff, script}], [], _keywords?) do
     script
--- a/lib/ex_unit/test/ex_unit/diff_test.exs
+++ b/lib/ex_unit/test/ex_unit/diff_test.exs
@@ -103,17 +103,6 @@ defmodule ExUnit.DiffTest do
 
     assert script(list1, list2) == expected
 
-    list1 = [1, 2]
-    list2 = [1, 1, 2]
-
-    expected = [
-      {:eq, "["},
-      [{:eq, "1"}, {:eq, ", "}, [del: "2", ins: "1"], {:ins, ", "}, {:ins, "2"}],
-      {:eq, "]"}
-    ]
-
-    assert script(list1, list2) == expected
-
     list1 = []
     list2 = [1, 2]
     expected = [{:eq, "["}, [ins: "1, 2"], {:eq, "]"}]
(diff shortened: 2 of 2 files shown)


devonestes:
Got it - I'll go with the full algorithm them.

devonestes:
So I've expanded this a bit, right now just for non-keyword lists. Here are some examples of the current output:

first
second
third
fourth

The first comma in the third and fourth example sort of bothers me, but changing that would require changing existing behavior found [here](https://github.com/elixir-lang/elixir/blob/a2b27f4d0212dd7d92166d332df7760dcd85a51f/lib/ex_unit/lib/ex_unit/diff.ex#L186), so I didn't want to change that for now unless y'all want it changed to better support this. I mean, technically this script is correct, but it would make more sense to me personally if that first comma was a `:del` or `:ins` instead of an `:eq`. I can see how some people might prefer this way, though, so I didn't want to change it just yet.

I realized that when we pull the equal elements from the front and back of the list, the only time that we do anything other than stick with what we're doing now is when the list of common elements at the head of the list was empty and the list of elements at the back of the list wasn't empty, so I substituted that with checking the head and tail of the list to ensure that the tails were equal and the heads weren't equal. In all other cases we would continue with the current behavior, so I left that alone.

If y'all think this is on the right track I'll continue implementing this for the other data types!

lib/ex_unit/lib/ex_unit/diff.ex
```diff
@@ -269,11 +314,9 @@ defmodule ExUnit.Diff do
   defp script_keyword_inner({key, val1}, {key, val2}, true),
     do: [{:eq, format_key(key, true)}, script_inner(val1, val2)]
 
-  defp script_keyword_inner(_pair1, _pair2, true),
-    do: nil
+  defp script_keyword_inner(_pair1, _pair2, true), do: nil
 
-  defp script_keyword_inner(elem1, elem2, false),
-    do: script(elem1, elem2)
```
josevalim: You are not running the latest version of the formatter, that's why this change happened. Make sure you have 1.6.5 in your system or run `bin/elixir bin/mix format`.

devonestes: oh, yeah, sorry! My vim integration isn't set for `bin/elixir bin/mix` 😢

josevalim:
Hi @devonestes! I looked at the current implementation and it feels is more complex than it has to be. Maybe that's the source of the comma issue, I am not sure, but here is what I would try:

```elixir
case diff_subset(list1, list2) do
  {:ok, diff} -> diff
  :error -> ...invokethecodefromtoday...
end
```

where `diff_subset(list1, list2)` would be implemented like this:

```elixir
def diff_subset(list1, list2) do
  {list1, list2, prefix} = take_common_path(list1, list2, [])
  {list1, list2, suffix} = take_common_path(Enum.reverse(list1), Enum.reverse(list2), [])

  if list1 == [] or list2 == [] do
    {:ok, wrap_in(:eq, Enum.reverse(prefix)) ++ wrap_in(:ins, Enum.reverse(list1)) ++ wrap_in(:del, Enum.reverse(list2)) ++ wrap_in(:eq, suffix)}
  else
    :error
  end
end

defp take_common_path([h | t1], [h | t2], acc), do: take_common_path(t1, t2, [h | acc])
defp take_common_path(t1, t2, acc), do: {t1, t2, acc}

defp wrap_in(tag, items), do: Enum.map(items, &{tag, &1})
```

Roughly pseudo code, probably has errors around the corners, but it should point you towards a simpler solution. It may also be that I am missing some context from the code that makes it more complex. In any case, the above is enough for the "take both prefix and suffix out, assert one of the remaining ones is empty and then generate a diff".

josevalim:
Regarding the other types, the only remaining data types where this would apply to are tuples (which probably share the logic with lists) and strings. But let's make sure we got all list cases covered first.

lib/ex_unit/lib/ex_unit/diff.ex
```diff
@@ -143,22 +143,67 @@ defmodule ExUnit.Diff do
...
+
+    initial_path = {0, 0, remainder1, remainder2, []}
+    remainder_length = length(remainder1) + length(remainder2)
+    remainder_script = find_script(0, remainder_length, [initial_path], keywords?)
+    result = format_each_fragment(remainder_script ++ [eq: equal_part], [], keywords?)
+
+    [{:eq, "["}, result, {:eq, "]"}]
   end
 
+  defp format_each_fragment([{:diff, script}], [], _keywords?), do: script
```
josevalim: This diff is also because of the formatter.

lib/ex_unit/test/ex_unit/diff_test.exs
```diff
@@ -125,6 +125,39 @@ defmodule ExUnit.DiffTest do
     assert script(list1, list2) == expected
 
     assert script([], []) == [eq: "[]"]
+
```
josevalim: Let's create a new test case especifically about testing lists that are subsets instead of adding more examples to those already large cases :)

josevalim:
@devonestes can you please rebase once again before moving forward? I have noticed that tuples and lists could share their logic, so I have refactored the code. This means that once you are done with lists, you will be done with tuples too. :D

devonestes:
Ok, rebased! For some reason it didn't click for me until I saw you implementation that we were looking for subsets! I was still doing something totally different.

I think this is along the lines of what y'all want now, and I also added tests for keyword lists and tuples since those are now covered as well. I didn't think we needed additional tests for charlists since those are just lists without any special behavior in the diffing.

devonestes:
Oh, and it looks like strings already are looking for substrings and formatting appropriately, which means that all data types we want to support should be taken care of. I tried writing a failing test and couldn't make it fail, so that's a good thing.

Also, maybe this is definitely for a separate issue, but there seems to be a pretty significant amount of duplication between the implementation of the Myers algorithm in `ExUnit.Diff` and the implementation in `List.myers_difference`. Is there a reason we can't or shouldn't unify those two?

josevalim:
@devonestes I have added the word "non-empty" to the test cases since we have subsets and supersets in other tests, it just happens that they are empty or equal lists. What do you think?

josevalim:
@devonestes can you also please add subset tests for tuples? thank you!

lexmag:
Thank you @devonestes. 💛

Issue #7645, opened 2018-05-04
Provide more topics for IEx.Helpers.runtime_info


michalmuskala:
This function has a lot of potential to be very useful for debugging - as a first tool for discovery where the problem might be in production situations.

I propose providing more topics based on features of the `:c` module and things that might be found in [recon](https://ferd.github.io/recon/) and other places of OTP.

Proposed additional topics:
* `:ets` - list of ets tables with basic info, like `:ets.i()`
* `:modules` - list of loaded modules, like `:c.m()`
* provide information about dirty schedulers in `:system`
* `:memory_details` - based on the new `:instrument` module in OTP 21. Maybe print some ASCII histograms?
* `:sctp`, `:tcp`, `:udp`, `:files` - list of ports of the provided type with some basic information about each of them (I'm not sure how this interacts with OTP 21 where files are no longer ports).

josevalim:
Many of those are welcome but ASCII histograms are probably out of scope (unless somebody has a very short and trivial implementation).

aerosol:
I'd like to tackle this if that's ok

michalmuskala:
Go ahead, @aerosol. This could be also split across multiple commits & PRs if some of the implementations are bigger/more complex.

josevalim:
Btw, let's not add anything based on processes and ports. My concern is that we may want to order, sort, filter this data in multiple ways and the runtime info panel won't provide this flexibility. I will amend the issue description for now.

michalmuskala:
TBH i considered those to be the most useful ones.

I agree a simple list might not be that useful. I'm not sure how to proceed here. You could make the same argument for ets or modules.

josevalim:
What would we show for modules? Alphabetical is properly fine there. I am
not worried about ets because the number of ets tables compared to
processes is way lower.
-- 


*José Valimwww.plataformatec.com.br
<http://www.plataformatec.com.br/>Founder and Director of R&D*

aerosol:
To me it sounds like single `runtime_info` function isn't the perfect interface for a comprehensive runtime inspection. Would it make sense to introduce `Runtime` module that's more specialized and provides the flexibility needed?

josevalim:
@aerosol good call. We could introduce process_info and ports_info in the future but that would be a separate issue. :) For now let's focus on how to improve the ones that fits runtime_info.

aerosol:
[...]

PR #7658, opened 2018-05-08
Add :allocators runtime_info topic

aerosol:
Refs https://github.com/elixir-lang/elixir/issues/7645#issuecomment-386910209

Changes:

  - New `runtime_info` topic: `:allocators`
  - Human-friendly bytes formatting, the unit is now selected based on the input value

image

### DIFF
--- a/lib/iex/lib/iex/helpers.ex
+++ b/lib/iex/lib/iex/helpers.ex
@@ -480,7 +480,7 @@ defmodule IEx.Helpers do
     |> Enum.map_join(", ", &inspect/1)
   end
 
-  @runtime_info_topics [:system, :memory, :memory_details, :limits, :applications]
+  @runtime_info_topics [:system, :memory, :allocators, :limits, :applications]
   @doc """
   Prints vm/runtime information such as versions, memory usage and statistics.
   Additional topics are available via `runtime_info/1`.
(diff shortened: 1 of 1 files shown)


lib/iex/lib/iex/helpers.ex
```diff
@@ -533,14 +533,37 @@ defmodule IEx.Helpers do
...
+    print_memory_detail("Atom Space", :atom_space)
+    print_memory_detail("Atom Table", :atom_table)
+    print_memory_detail("BIF Timer", :bif_timer)
+    print_memory_detail("Bits Bufs Size", :bits_bufs_size)
+    print_memory_detail("Dist Table", :dist_table)
+    print_memory_detail("ETS Misc", :ets_misc)
+    print_memory_detail("Export List", :export_list)
+    print_memory_detail("Export Table", :export_table)
+    print_memory_detail("Function Table", :fun_table)
+    print_memory_detail("Link LH", :link_lh)
```
aerosol: I could not find any info on what that might be. Could use a friendlier name?

lib/iex/lib/iex/helpers.ex
```diff
@@ -533,14 +533,37 @@ defmodule IEx.Helpers do
...
     print_memory("ETS", :ets)
     print_memory("Processes", :processes)
   end
 
+  defp print_runtime_info_topic(:memory_details) do
+    print_pane("Memory Details")
+    print_memory_detail("Atom Space", :atom_space)
+    print_memory_detail("Atom Table", :atom_table)
+    print_memory_detail("BIF Timer", :bif_timer)
+    print_memory_detail("Bits Bufs Size", :bits_bufs_size)
```
aerosol: Does `bits` here refers to bitstrings? :thinking:

lib/iex/lib/iex/helpers.ex
```diff
@@ -596,18 +619,52 @@ defmodule IEx.Helpers do
...
+    case probe do
+      {_, allocated, used} ->
+        IO.puts(
+          "#{pad_key(key)}#{format_bytes(allocated)} (#{format_bytes(used)} used)"
+        )
+
+      {_, allocated} ->
+        IO.puts("#{pad_key(key)}#{format_bytes(allocated)}")
+
+      _ ->
```
aerosol: > erlang:system_info(allocated_areas) is intended for debugging, and the content is highly implementation-dependent. The content of the results therefore changes when needed without prior notice.

lib/iex/lib/iex/helpers.ex
```diff
@@ -596,18 +619,52 @@ defmodule IEx.Helpers do
...
 
-    IO.puts("#{pad_key(key)}#{value} #{unit}")
+    "#{value} #{unit}"
   end
+  defp format_bytes(bytes, :B) when is_integer(bytes), do: "#{bytes} B"
 
+  defp memory_unit(:GB), do: 1024 * 1024 * 1024
   defp memory_unit(:MB), do: 1024 * 1024
-  defp memory_unit(:kB), do: 1024
+  defp memory_unit(:KB), do: 1024
```
aerosol: kB = 1000, KB = 1024

josevalim:
Thanks @aerosol! I would call this ":allocators" instead of ":memory_details". But before doing those changes, let's wait for some feedback from @michalmuskala.

aerosol:
:+1: for the rename FWIW. Was blindly following the base thread :P

aerosol:
A tentative ping, is there anything I can do here to push this forward?

lib/iex/lib/iex/helpers.ex
josevalim: Ping @michalmuskala.

lib/iex/lib/iex/helpers.ex
josevalim: Ping @michalmuskala.

lib/iex/lib/iex/helpers.ex
aerosol: FWIW it's `N/A` on Erlang/OTP 21.0

lib/iex/lib/iex/helpers.ex
michalmuskala: Looking through the OTP code, it seems that it's something allocated when placing integers in binaries, when the binary is not byte-aligned.

lib/iex/lib/iex/helpers.ex
michalmuskala: That's probably related to the links being largely reworked to increase scalability.

josevalim: @aerosol even better, let's remove it then.

michalmuskala:
Thank you! :heart:

PR #7662, opened 2018-05-09
Add helpful error message for Keyword.delete

mmartinson:
Since we're recursively traversing a full list to delete elements, we can validate the keyword list with a pattern match at each step to throw a helpful exception, and avoid the performance penalty of validation in advance with `Keyword.keyword?/1`. This, I think, is preferable to the current behaviour of `(FunctionClauseError) no function clause matching in Keyword.delete_key/3`, since it avoids leaving the developer searching through private functions to track down the offending caller.

I thought it would be helpful enough to show only the offending element in the error messag, rather than changing the function definitions to pass the full list through the recursion for the sake of the error message, but I could certainly change that to make the message similar to `Keyword.merge/3` if anyone thinks that's important.

This is my first PR of code for this project, feel free to give me feedback on anything about the process I may have missed.

### DIFF
--- a/lib/elixir/lib/keyword.ex
+++ b/lib/elixir/lib/keyword.ex
@@ -477,6 +477,14 @@ defmodule Keyword do
     [pair | delete_key_value(tail, key, value)]
   end
 
+  defp delete_key_value([not_pair | _tail], _key, _value) do
+    raise ArgumentError,
+      message:
+        "expected a keyword list as the first argument, but got list with invalid element #{
+          inspect(not_pair)
+        }"
+  end
+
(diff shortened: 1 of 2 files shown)


mmartinson:
I notice that CI failed. Is anyone able to give me some insight into a mistake I made, or is that an unrelated issue?

devonestes:
Here’s the log for the failure: https://travis-ci.org/elixir-lang/elixir/jobs/376627846

At first glance it looks unrelated since it’s testing recompilation, and I don’t see how this new code could affect that, but I could be wrong. The easiest thing you could do first would be to restart that build and see if it passes.

PR #7664, opened 2018-05-09
Display simple coverage report to console

hauleth:
This should reduce need for packages like excoveralls in some use cases
when all what is needed is to have total coverage percentage in any
form available for CI to provide coverage badge.

Reference: https://docs.gitlab.com/ee/user/project/pipelines/settings.html#test-coverage-parsing

### DIFF
--- a/lib/mix/lib/mix/tasks/test.ex
+++ b/lib/mix/lib/mix/tasks/test.ex
@@ -19,10 +19,11 @@ defmodule Mix.Tasks.Test do
       fn ->
         Mix.shell().info("\nGenerating cover results ...\n")
         {:result, ok, _fail} = :cover.analyse(:coverage, :module)
-
         results = Enum.sort_by(ok, &percentage(elem(&1, 1)), &>=/2)
 
-        if summary_opts = Keyword.get(opts, :summary, true), do: console(results, summary_opts)
+        if summary_opts = Keyword.get(opts, :summary, true) do
+          console(results, summary_opts)
+        end
 
         html(results, opts)
       end
@@ -39,16 +40,13 @@ defmodule Mix.Tasks.Test do
         |> Stream.each(&display(&1, opts))
         |> Enum.reduce({0, 0}, fn {_, {covered, not_covered}},
                                   {total_covered, total_not_covered} ->
-          {total_cov + cov, total_not_cov + not_cov}
+          {total_covered + covered, total_not_covered + not_covered}
         end)
 
       Mix.shell().info("-----------|--------------------------")
-
       display({"Total", total}, opts)
       Mix.shell().info("")
     end
-
-    defp console(_, value) when value in [nil, false], do: nil
 
     defp html(results, opts) do
       output = opts[:output]
--- a/lib/mix/lib/mix/tasks/test.ex
+++ b/lib/mix/lib/mix/tasks/test.ex
@@ -37,8 +37,9 @@ defmodule Mix.Tasks.Test do
       total =
         results
         |> Stream.each(&display(&1, opts))
-        |> Enum.reduce({0, 0}, fn {_, {cov, not_cov}}, {tot_cov, tot_not_cov} ->
-          {tot_cov + cov, tot_not_cov + not_cov}
+        |> Enum.reduce({0, 0}, fn {_, {covered, not_covered}},
+                                  {total_covered, total_not_covered} ->
+          {total_cov + cov, total_not_cov + not_cov}
         end)
 
       Mix.shell().info("-----------|--------------------------")
@@ -84,13 +85,13 @@ defmodule Mix.Tasks.Test do
     end
 
     defp percentage({0, 0}), do: 100
-    defp percentage({cov, not_cov}), do: cov / (cov + not_cov) * 100
+    defp percentage({covered, not_covered}), do: covered / (covered + not_covered) * 100
 
-    defp format(num, len) when is_integer(num),
-      do: :io_lib.format("~#{len}b", [num])
+    defp format(number, length) when is_integer(number),
+      do: :io_lib.format("~#{length}b", [number])
 
-    defp format(num, len) when is_float(num),
-      do: :io_lib.format("~#{len}.2f", [num])
+    defp format(number, length) when is_float(number),
+      do: :io_lib.format("~#{length}.2f", [number])
 
     defp format_name(name) when is_binary(name), do: name
     defp format_name(mod) when is_atom(mod), do: inspect(mod)
--- a/lib/mix/lib/mix/tasks/test.ex
+++ b/lib/mix/lib/mix/tasks/test.ex
@@ -1,5 +1,7 @@
 defmodule Mix.Tasks.Test do
   defmodule Cover do
+    @default_threshold 90
+
     @moduledoc false
 
     def start(compile_path, opts) do
@@ -14,17 +16,84 @@ defmodule Mix.Tasks.Test do
           Mix.raise("Failed to cover compile directory: " <> compile_path)
       end
 
-      output = opts[:output]
-
       fn ->
-        Mix.shell().info("\nGenerating cover results ...")
-        File.mkdir_p!(output)
+        Mix.shell().info("\nGenerating cover results ...\n")
+        {:result, ok, _fail} = :cover.analyse(:coverage, :module)
 
-        Enum.each(:cover.modules(), fn mod ->
-          {:ok, _} = :cover.analyse_to_file(mod, '#{output}/#{mod}.html', [:html])
-        end)
+        results = Enum.sort_by(ok, &percentage(elem(&1, 1)), &>=/2)
+
+        if summary_opts = Keyword.get(opts, :summary, true), do: console(results, summary_opts)
+
+        html(results, opts)
       end
     end
+
+    defp console(results, true), do: console(results, [])
+
+    defp console(results, opts) when is_list(opts) do
+      Mix.shell().info("Percentage | Module")
+      Mix.shell().info("-----------|--------------------------")
+
+      total =
+        results
+        |> Stream.each(&display(&1, opts))
+        |> Enum.reduce({0, 0}, fn {_, {cov, not_cov}}, {tot_cov, tot_not_cov} ->
+          {tot_cov + cov, tot_not_cov + not_cov}
+        end)
+
+      Mix.shell().info("-----------|--------------------------")
+
+      display({"Total", total}, opts)
+      Mix.shell().info("")
+    end
+
+    defp console(_, value) when value in [nil, false], do: nil
+
+    defp html(results, opts) do
+      output = opts[:output]
+      File.mkdir_p!(output)
+
+      for {mod, _} <- results do
+        {:ok, _} = :cover.analyse_to_file(mod, '#{output}/#{mod}.html', [:html])
+      end
+
+      Mix.shell().info([
+        "Generated HTML coverage results in '",
+        output,
+        "' directory\n"
+      ])
+    end
+
+    defp color(percentage, true), do: color(percentage, @default_threshold)
+    defp color(_, false), do: ""
+    defp color(percentage, threshold) when percentage > threshold, do: :green
+    defp color(_, _), do: :red
+
+    defp display({name, coverage}, opts) do
+      threshold = Keyword.get(opts, :threshold, @default_threshold)
+      percentage = percentage(coverage)
+
+      Mix.shell().info([
+        color(percentage, threshold),
+        format(percentage, 9),
+        "%",
+        :reset,
+        " | ",
+        format_name(name)
+      ])
+    end
+
+    defp percentage({0, 0}), do: 100
+    defp percentage({cov, not_cov}), do: cov / (cov + not_cov) * 100
+
+    defp format(num, len) when is_integer(num),
+      do: :io_lib.format("~#{len}b", [num])
+
+    defp format(num, len) when is_float(num),
+      do: :io_lib.format("~#{len}.2f", [num])
+
+    defp format_name(name) when is_binary(name), do: name
+    defp format_name(mod) when is_atom(mod), do: inspect(mod)
   end
 
   use Mix.Task
@@ -151,6 +220,12 @@ defmodule Mix.Tasks.Test do
 
     * `:output` - the output for cover results, defaults to `"cover"`
     * `:tool`   - the coverage tool
+    * `:summary` - summary output configuration, can be either boolean
+      or keyword list, when keyword list is passed it can specify `:threshold`
+      which can be boolean or numeric value which would enable colouring
+      of percentages red/green depending either below/over threshold
+      respectively, defaults to `[threshold: 90]`
+
 
   By default, a very simple wrapper around OTP's `cover` is used as a tool,
   but it can be overridden as follows:


josevalim:
Can you please include an example of the report? Does any other client besides gitlab parses coverage reports like that?

lib/mix/lib/mix/tasks/test.ex
```diff
@@ -2,6 +2,8 @@ defmodule Mix.Tasks.Test do
   defmodule Cover do
     @moduledoc false
 
+    @threshold 90
```
josevalim: This would need to be made configurable.

lib/mix/lib/mix/tasks/test.ex
```diff
@@ -17,14 +19,64 @@ defmodule Mix.Tasks.Test do
...
+    defp display({name, coverage}, opts) do
+      threshold = Keyword.get(opts, :threshold, @threshold)
+      percentage = percentage(coverage)
+
+      Mix.shell().info([
+        colour(percentage, threshold),
+        format(percentage, 9),
+        "%",
+        :reset,
+        " | #{name}"
```
josevalim: You need to inspect the module name so it shows only as "MyApp.Module" and not "Elixir.MyApp.Module". But you need to be careful to not inspect the string "Total".

josevalim:
Also please check the build.

lib/mix/lib/mix/tasks/test.ex
hauleth: This is configurable via `threshold` value in `test_coverage` configuration option. I haven't documented it yet thought.

josevalim:
Let's also provide an option that skips the summary if desired.

lib/mix/lib/mix/tasks/test.ex
hauleth: That is the reason why I haven't done it, however I can make it happen.

lib/mix/lib/mix/tasks/test.ex
josevalim: Beautiful. Maybe we should call this `@default_threshold` then?

josevalim:
One last thing: gitlab default's regex expects the word "covered" in there. Do we need to add something similar? Wouldn't just printing percentages be a bit ambiguous? What regex are you using currently?

hauleth:
@josevalim about comments, from top to bottom:

- I am not aware of any that works in such way, however I only skimmed through available options, so maybe there are ones that I am no aware of.
- I need to update tests.
- I was thinking about something other: sort solutions based on their coverage percentage and then display only "bottom N" of options. Then such option could be set to `false` to disable CLI coverage report at all. What do you think? Or rather add additional option that would disable output and separate that would print sorted results?
- Default regex yes, but this can be parsed using `\d+(\.\d+)?%\s*\|\s*Total`.

hauleth:
Example result:

```
Cover results
Percentage | Module
      0.0% | Elixir.Cover
      0.0% | Elixir.ImagerWeb.Instrumenter
     50.0% | Elixir.Imager.Application
      0.0% | Elixir.Imager.Convert
      0.0% | Elixir.ImagerWeb.ChannelCase
      0.0% | Elixir.ImagerWeb.UserSocket
      0.0% | Elixir.ImagerWeb.WebhookController
     6.45% | Elixir.ImagerWeb.Router.Helpers
    15.79% | Elixir.ImagerWeb.ErrorView
    45.45% | Elixir.ImagerWeb.ConnCase
     5.26% | Elixir.ImagerWeb.Endpoint
    23.81% | Elixir.Imager.Stats
     5.71% | Elixir.Imager
      0.0% | Elixir.ImagerWeb.Router
      0.0% | Elixir.ImagerWeb
     7.88% | Total
```

josevalim:
> I was thinking about something other: sort solutions based on their coverage percentage and then display only "bottom N" of options. Then such option could be set to false to disable CLI coverage report at all.

To me it is OK to sort by percentage but is there a reason to allow a cut off?

michalmuskala:
Should we print the module names with elixir syntax?

hauleth:
@michalmuskala @josevalim how about such output:

```
Coverage results:

Percentage | Module
      0.0% | Imager.Convert
      0.0% | Cover
      0.0% | ImagerWeb.Instrumenter
      0.0% | ImagerWeb
      0.0% | ImagerWeb.ChannelCase
      0.0% | ImagerWeb.UserSocket
      0.0% | ImagerWeb.WebhookController
      0.0% | ImagerWeb.Router
     5.26% | ImagerWeb.Endpoint
     5.71% | Imager
     6.45% | ImagerWeb.Router.Helpers
    15.79% | ImagerWeb.ErrorView
    23.81% | Imager.Stats
    45.45% | ImagerWeb.ConnCase
     50.0% | Imager.Application
     7.73% | Total

Generate coverage results in cover
```

hauleth:
I am thinking about changing the last line to `Generated HTML coverage results in 'cover' directory`.

lib/mix/lib/mix/tasks/test.ex
hauleth: No problem

webofbits:
@hauleth could you show us what an example of the HTML page would look like?

hauleth:
@webdeb HTML result is left as is. For more information check Erlang's `cover` module.

webofbits:
You coverage logs are fine. Not sure if would be possible to have some additional info like lines, relevant or missed but I think that would be great to have. Just an opinion though and thanks for this pull request. 
```
COV    FILE                                        LINES RELEVANT   MISSED
100.0% lib/module_name.ex                              9        0        0
```

lib/mix/lib/mix/tasks/test.ex
```diff
@@ -17,14 +19,64 @@ defmodule Mix.Tasks.Test do
       output = opts[:output]
 
       fn ->
-        Mix.shell().info("\nGenerating cover results ...")
         File.mkdir_p!(output)
 
-        Enum.each(:cover.modules(), fn mod ->
-          {:ok, _} = :cover.analyse_to_file(mod, '#{output}/#{mod}.html', [:html])
-        end)
+        Mix.shell().info("\nCover results")
```
josevalim: We can keep this line as it was before because it will also work if folks choose to skip the summary.

hauleth:
Yes this is perfectly possible, however I want to keep output format as simple as possible without any additional clutter that can be provided by external tools.

@josevalim @michalmuskala any opinions about output format?

josevalim:
@hauleth I like your suggested output last. I would just keep the first line of the output and the last line the same as it is today. Let's minimiza changes and make sure the summary is optional.

@webofbits let's start simple for now we can always add more information later.

lib/mix/lib/mix/tasks/test.ex
hauleth: How about keeping this line and adding additional output after HTML result was generated?

hauleth:
@josevalim @webofbits 

```
Generating cover results ...

Coverage results:

Percentage | Module
      0.0% | Imager.Convert
      0.0% | ImagerWeb.ChannelCase
      0.0% | ImagerWeb.UserSocket
      0.0% | ImagerWeb
      0.0% | ImagerWeb.WebhookController
      0.0% | Cover
      0.0% | ImagerWeb.Instrumenter
      0.0% | ImagerWeb.Router
     5.26% | ImagerWeb.Endpoint
     5.71% | Imager
     6.45% | ImagerWeb.Router.Helpers
    15.79% | ImagerWeb.ErrorView
    23.81% | Imager.Stats
    45.45% | ImagerWeb.ConnCase
     50.0% | Imager.Application
     7.65% | Total

Generated HTML coverage results in 'cover' directory
```

webofbits:
Sounds good to me :+1:

josevalim:
@hauleth "Coverage results:" is unnecessary. From the previous line and context we know they are results.

lib/mix/lib/mix/tasks/test.ex
```diff
@@ -1,6 +1,22 @@
 defmodule Mix.Tasks.Test do
   defmodule Cover do
-    @moduledoc false
+    @default_formatters [:console, :html]
+    @default_threshold 90
+
+    @moduledoc """
```
josevalim: Let's please keep this module still private. I don't think we should expose it. Let's just have a section in the `mix test` docs that says the configurations for the default cover handling.

lib/mix/lib/mix/tasks/test.ex
```diff
@@ -14,17 +30,91 @@ defmodule Mix.Tasks.Test do
           Mix.raise("Failed to cover compile directory: " <> compile_path)
       end
 
-      output = opts[:output]
+      formatters = Keyword.get(opts, :formatters, @default_formatters)
```
josevalim: `:console` is not really a formatter because it doesn't print it for every module. It is just a summary. So I would rather have a `summary: true | false` option.

lib/mix/lib/mix/tasks/test.ex
```diff
@@ -14,17 +30,91 @@ defmodule Mix.Tasks.Test do
...
 
       fn ->
+        {:result, ok, _fail} = :cover.analyse(:coverage, :module)
         Mix.shell().info("\nGenerating cover results ...")
```
josevalim: Let's print this before we call analyze, otherwise the user may wait for something and they don't know what it is.

lib/mix/lib/mix/tasks/test.ex
hauleth: According to Erlang `cover:analyse/2` docs it will print data for all cover compiled modules. However I can misunderstanding your comment.

josevalim: Do you mean that we have also added an option to print all of the results to the terminal? If so, that's a different feature than the one we were debating. Let's focus on the summary feature first. Other additions should be considered separately.

hauleth: That is the exact result of what this PR produces https://github.com/elixir-lang/elixir/pull/7664#issuecomment-387730487. It display all modules and their respective coverage percentage. You meant something else? Like we should display only overall coverage like:

```
Generating cover results ...

Coverage: 72.32%

Generated HTML coverage results in 'cover' directory
```

?

josevalim: Ok. Sorry for the misunderstanding. The current output is fine. However, I wouldn't call it a formatter because what we print to the console and what we write to html files are two very distinct things. My proposal is to remove the idea of formatters and always generate HTML as before. Let's call the new reporting that you have added as part of this PR "Summary" and it can be disabled via the `summary: true | false` option.

whatyouhide:
I might be missing something obvious, but why have the percentage first (left) and the module name after (right) instead of the other way around?

OvermindDL1:
@whatyouhide The percentage is pretty bounded in how many characters it can take up, where a module name is not (in comparison), thus I think it is easier to keep both readable instead of the module name occasionally pushing the percentages off the side with long module names.  :-)

hauleth:
Exactly what @OvermindDL1 said. It is easier to display it in that way instead of looping through module names twice and building list based on the longest module name.

whatyouhide:
Yeah I get that it's easier to build but is that an argument for the user of this or for the implementer? I think the difference in implementation complexity is so small that it justifies having what IMO feels more natural to the user, that is, modules first and percentage after. Navigating the list of modules twice shouldn't be a performance issue since these lists are supposed to be on the "very small" side.

josevalim:
I think the most important information here is the percentage and given that we will be ordering by %, I vote to keep it first too. The goal here is not to list modules.

OvermindDL1:
As the user I find the percentages first *significantly* more preferable as well as I want to see what they are before I care about module names.

hauleth:
@whatyouhide I also find it much easier to connect module with percentage in that format as these are next to each other, if there would be bunch of spaces between them and there would be some long and short module names then it would be hard to follow lines without some kind of guide. While it is fairly easy to get guide in your terminal view then it is pretty hard to get one in CI output if the CI UI author didn't thought about it beforehand.

hauleth:
Using example from above:

```
 Module                      | Percentage
-----------------------------|------------
 Imager.Convert              |       0.0%
 ImagerWeb.ChannelCase       |       0.0%
 ImagerWeb.UserSocket        |       0.0%
 ImagerWeb                   |       0.0%
 ImagerWeb.WebhookController |       0.0%
 Cover                       |       0.0%
 ImagerWeb.Instrumenter      |       0.0%
 ImagerWeb.Router            |       0.0%
 ImagerWeb.Endpoint          |      5.26%
 Imager                      |      5.71%
 ImagerWeb.Router.Helpers    |      6.45%
 ImagerWeb.ErrorView         |     15.79%
 Imager.Stats                |     23.81%
 ImagerWeb.ConnCase          |     45.45%
 Imager.Application          |      50.0%
-----------------------------|------------
 Total                       |      7.65%
```

It can get pretty hard to connect `Imager` with `5.71%`, and now imagine if there would be block of short module names and there would be one exceptionally long. It could became very hard to connect values with names on the first sight.

hauleth:
Only open question is whether sort in increasing or decreasing order. I do not think it should be configureable, but there should be defined order.

hauleth:
@whatyouhide another reason why it is better to leave percentage in front is that it is much easier to crop lines at the end in the terminal window rather than crop them in the middle or at the beginning to have visible results in case of very long module names.

OvermindDL1:
> Only open question is whether sort in increasing or decreasing order. I do not think it should be configureable, but there should be defined order.

Honestly I'd prefer decreasing so that the 0%'ers are at the bottom as that is the most immediate thing I will see when running commands on the terminal.

AndrewDryga:
@OvermindDL1 I think it's much better to see them in increasing (ascending) order. I feel like it's better to have issues to be on top of the list (first things should be addressed first), not on the bottom of it.

OvermindDL1:
@AndrewDryga Except with them being at the top then the issues have scrolled off the screen and I can't see them without doing `<Ctrl>+<B>, <[>` and then scroll around with the keys, that's a pain to see the information that I want to see most, where the last printed information is the information that is immediately viewable and accessible.

lib/mix/lib/mix/tasks/test.ex
```diff
@@ -162,6 +232,14 @@ defmodule Mix.Tasks.Test do
   It must return either `nil` or an anonymous function of zero arity that will
   be run after the test suite is done.
 
+  ### Default coverage module options
+
+    * `:threshold` - summary output colours resulting percentage red/green
+      depending on weather given value is below or over given threshold,
```
fertapric: s/weather/whether/

lib/mix/lib/mix/tasks/test.ex
```diff
@@ -162,6 +232,14 @@ defmodule Mix.Tasks.Test do
   It must return either `nil` or an anonymous function of zero arity that will
   be run after the test suite is done.
 
+  ### Default coverage module options
+
+    * `:threshold` - summary output colours resulting percentage red/green
```
fertapric: The `:threshold` option seems to only make sense when the `:summary` option is enabled. In order to reflect that hierarchy, what do you think about?

```elixir
test_coverage: [summary: true] # Enables the summary with the default threshold
test_coverage: [summary: [threshold: 70]] # Enables the summary and sets the threshold to 70
```

If in the future the summary adds more configuration variables, they won't be mixed with the global coverage options.

lib/mix/lib/mix/tasks/test.ex
```diff
@@ -162,6 +232,14 @@ defmodule Mix.Tasks.Test do
   It must return either `nil` or an anonymous function of zero arity that will
   be run after the test suite is done.
 
+  ### Default coverage module options
```
fertapric: Should these options be listed in the "Coverage" section above?

> ## Coverage
>
> The `:test_coverage` configuration accepts the following options:
>
>  * `:output` - the output for cover results, defaults to `"cover"`
>  * `:tool`   - the coverage tool
>  * `:summary` - whether reporter should output coverage data summary to the console, defaults to `true`

josevalim:
Let's go with decreasing order (100%) comes first. Let's also adopt @fertapric's  suggestion.

hauleth:
Everything is updated to what @fertapric requested.

lexmag:
Does it make sense to drop `%` sign given that we explicitly state it is "Percentage"?

lib/mix/lib/mix/tasks/test.ex
```diff
@@ -149,8 +225,14 @@ defmodule Mix.Tasks.Test do
 
   The `:test_coverage` configuration accepts the following options:
 
-    * `:output` - the output for cover results, defaults to `"cover"`
-    * `:tool`   - the coverage tool
+    * `:output`  - the output for cover results, defaults to `"cover"`
```
lexmag: We don't align `-` anymore due to noise it generates in commits.

lib/mix/lib/mix/tasks/test.ex
```diff
@@ -14,17 +16,91 @@ defmodule Mix.Tasks.Test do
...
       end
+
+      Mix.shell().info([
+        "Generated HTML coverage results in '",
+        output,
+        "' directory\n"
+      ])
     end
+
+    defp colour(percentage, true), do: colour(percentage, @default_threshold)
```
lexmag: Except "behaviour" we should use American spelling.

lib/mix/lib/mix/tasks/test.ex
```diff
@@ -14,17 +16,91 @@ defmodule Mix.Tasks.Test do
...
+          {tot_cov + cov, tot_not_cov + not_cov}
         end)
+
+      Mix.shell().info("-----------|--------------------------")
+
+      display({"Total", total}, opts)
+      Mix.shell().info("")
+    end
+
+    defp console(_, _), do: nil
```
lexmag: Maybe match on `[false, nil]` to make `summary: 123` fail?

lexmag:
@hauleth in my opinion this one character does not improve readability (rather brings visual noise).
I think if we want to add other metrics in the future we'll reconsider the current output quite a bit anyway.

lexmag:
@hauleth just checked the common agreement on this with the rest of the folks in the team: let's keep the sign. 👍

lib/mix/lib/mix/tasks/test.ex
```diff
@@ -14,17 +16,91 @@ defmodule Mix.Tasks.Test do
...
+      |> Integer.to_string()
+      |> String.pad_leading(len)
+    end
+
+    defp format(num, len) when is_float(num) do
+      num
+      |> Float.round(2)
+      |> Float.to_string()
+      |> String.pad_leading(len)
+    end
```
michalmuskala: Could we use `:io_lib.format("~#{len}.2f", [num])` for formatting here? This will always give us 2 fields of precision and nicely align the results.

lib/mix/lib/mix/tasks/test.ex
```diff
@@ -14,17 +16,84 @@ defmodule Mix.Tasks.Test do
...
+        :reset,
+        " | ",
+        format_name(name)
+      ])
     end
+
+    defp percentage({0, 0}), do: 100
+    defp percentage({cov, not_cov}), do: cov / (cov + not_cov) * 100
+
+    defp format(num, len) when is_integer(num),
```
eksperimental: Generally we use full name in arguments, so it reads better `number` and `length`

lib/mix/lib/mix/tasks/test.ex
```diff
@@ -14,17 +16,84 @@ defmodule Mix.Tasks.Test do
...
+        format(percentage, 9),
+        "%",
+        :reset,
+        " | ",
+        format_name(name)
+      ])
     end
+
+    defp percentage({0, 0}), do: 100
+    defp percentage({cov, not_cov}), do: cov / (cov + not_cov) * 100
```
eksperimental: It reads better full name arg. `covered` and `not_covered`

lib/mix/lib/mix/tasks/test.ex
hauleth: @eksperimental my problem with `length` is that vim-elixir colours it as a keyword which I find distracting. Also I find `len` and `num` popular enough that I didn't even thought about using full name.

OvermindDL1: > my problem with length is that vim-elixir colours it as a keyword which I find distracting

Wait what why?  It's not a keyword!  o.O

hauleth: elixir-editors/vim-elixir#378

lib/mix/lib/mix/tasks/test.ex
```diff
@@ -14,17 +16,85 @@ defmodule Mix.Tasks.Test do
...
+    defp console(results, opts) when is_list(opts) do
+      Mix.shell().info("Percentage | Module")
+      Mix.shell().info("-----------|--------------------------")
+
+      total =
+        results
+        |> Stream.each(&display(&1, opts))
+        |> Enum.reduce({0, 0}, fn {_, {covered, not_covered}},
+                                  {total_covered, total_not_covered} ->
+          {total_cov + cov, total_not_cov + not_cov}
```
michalmuskala: This line should be:
```
{total_covered + covered, total_not_covered + not_covered}
```

josevalim:
I have pushed a commit that fixes the build, so as soon as it passes we are good to go!

josevalim:

PR #7665, opened 2018-05-09
Document String functions that accept compiled patterns

fertapric:


josevalim:

PR #7668, opened 2018-05-10
Consider if definition should be checked for @impl

### LINKED ISSUE #7617: Do not warn for @impl true missing in generated code
For example, if you define `@impl true` for init in Ecto.Repo, we show plenty of warnings for the internal functions and we should not. The opposite is also true.

ggcampinho:
Fixes #7617

### DIFF
--- a/lib/elixir/src/elixir_def.erl
+++ b/lib/elixir/src/elixir_def.erl
@@ -69,13 +69,13 @@ fetch_definition([Tuple | T], File, Module, Set, Bag, All, Private) ->
 
   try ets:lookup_element(Bag, {clauses, Tuple}, 2) of
     Clauses ->
+      MetaCheck = add_check_to_meta(Check, Meta),
       NewAll =
-        [{Tuple, Kind, add_defaults_to_meta(Defaults, Meta), Clauses} | All],
+        [{Tuple, Kind, add_defaults_to_meta(Defaults, MetaCheck), Clauses} | All],
       NewPrivate =
         case (Kind == defp) orelse (Kind == defmacrop) of
           true ->
-            WarnMeta = case Check of true -> Meta; false -> false end,
-            [{Tuple, Kind, WarnMeta, Defaults} | Private];
+            [{Tuple, Kind, MetaCheck, Defaults} | Private];
           false ->
             Private
         end,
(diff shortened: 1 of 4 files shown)


lib/elixir/src/elixir_def.erl
```diff
@@ -69,13 +69,13 @@ fetch_definition([Tuple | T], File, Module, Set, Bag, All, Private) ->
 
   try ets:lookup_element(Bag, {clauses, Tuple}, 2) of
     Clauses ->
+      MetaCheck = add_check_to_meta(Check, Meta),
```
josevalim: I would prefer to not add a new check flag and rely on the existing ones. In this case, I think we can check for the `:context` flag instead. It is one of the things that we use to set `Check` to true and it should be enough in this case.

lib/elixir/test/elixir/kernel/impl_test.exs
```diff
@@ -28,6 +28,31 @@ defmodule Kernel.ImplTest do
...
+
+  defmodule UseBehaviourWithImpl do
+    @callback foo() :: any
+    @callback bar() :: any
+
+    defmacro __using__(_opts) do
+      quote do
+        @behaviour Kernel.ImplTest.UseBehaviourWithImpl
+        @impl true
+        def foo(), do: :auto_generated
```
josevalim: The question I would have is: If the behaviour has three callbacks and we provide two default implementations but we set `@impl true` to only one of them... would that be checked?

josevalim: Also what would happen if we declare the `@impl` for a non-existing callback? It seems that now we are simply skipping the checks. I think we want to keep the code mostly as is today... we just want to avoid warnings on `warn_missing_impls`.

I think this would be a heuristic. If there is a missing implementation, we get the context for the missing implementation (maybe nil). We only run the missing implementation check for other functions that have exactly the same context.

I think this would actually allow us to simplify the overridable logic we have today but we can do that as another PR.

josevalim:
Hold on, I may have a suggestion incoming!

josevalim:
I have merged your commit manually and then unified the overridable code with the context one! Thank you! :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

PR #7669, opened 2018-05-10
Improve guards in Kernel.*_in (Access-related) functions

whatyouhide:


lib/elixir/lib/kernel.ex
```diff
@@ -2103,7 +2103,7 @@ defmodule Kernel do
   an error will be raised when trying to access it next.
   """
   @spec put_in(Access.t(), nonempty_list(term), term) :: Access.t()
-  def put_in(data, keys, value) do
+  def put_in(data, keys, value) when is_list(keys) and keys != [] do
```
josevalim: Maybe pattern match on `[_ | _] = keys`?

josevalim: I find it simpler, more concise and faster.

lib/elixir/lib/kernel.ex
```diff
@@ -2125,7 +2125,7 @@ defmodule Kernel do
   an error will be raised when trying to access it next.
   """
   @spec update_in(Access.t(), nonempty_list(term), (term -> term)) :: Access.t()
-  def update_in(data, keys, fun) when is_function(fun, 1) do
+  def update_in(data, [_ | _] = keys, fun) when is_function(fun) do
```
whatyouhide: @josevalim on the same philosophy, I removed the arity check from here. If I understand correctly, we would do that in this case too.

michalmuskala:
Thank you! :heart:

PR #7670, opened 2018-05-10
Remove the Access.get/3 callback

whatyouhide:
It was not being used. In fact, `Access.get/3` was relying on the `c:Access.fetch/2` callback for some reason.

### DIFF
--- a/lib/elixir/lib/access.ex
+++ b/lib/elixir/lib/access.ex
@@ -182,28 +182,6 @@ defmodule Access do
   @callback fetch(term :: t, key) :: {:ok, value} | :error
 
   @doc """
-  Invoked in order to access the value stored under `key` in the given term `term`,
-  defaulting to `default` if not present.
-
-  This function should return the value under `key` in `term` if there's
-  such key, otherwise `default`.
-
-  For most data structures, this can be implemented using `fetch/2` internally;
-  for example:
-
-      def get(structure, key, default) do
-        case fetch(structure, key) do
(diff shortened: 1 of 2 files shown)


lib/elixir/lib/access.ex
```diff
@@ -329,43 +307,13 @@ defmodule Access do
...
-
-  def get(%module{} = container, key, default) do
-    try do
-      module.fetch(container, key)
-    rescue
-      exception in UndefinedFunctionError ->
-        raise_undefined_behaviour(exception, module, {^module, :fetch, [^container, ^key], _})
-    else
+  def get(container, key, default \\ nil) do
+    case fetch(container, key) do
```
whatyouhide: I think the gain in code simplicity is pretty big here against the very slight optimization of not calling `fetch/2` internally (we're losing something like 25 lines of code after all), but I can revert if necessary.

michalmuskala: I would keep it for lists and maps. It's called very often in most programs, so even a small optimisation matters.
Custom Access implementations, on the other hand, are probably not as common.

josevalim: Exactly what @michalmuskala said. This is called **a lot**, the more verbose code here is justified.

whatyouhide:
@josevalim @michalmuskala done, this is ready to go.

josevalim:
Ship it! :)
-- 

PR #7672, opened 2018-05-11
Consider possible accumulator in Stream.do_zip_step/2

### LINKED ISSUE #7667: FunctionClauseError in Stream.do_zip_step/2
### Environment

* Elixir & Erlang/OTP version: Elixir 1.7.0-dev (36a6ebd59) (compiled with Erlang/OTP 20)
* Operating system: ArchLinux (Kernel 4.16.6-1)

### Current behavior

```Elixir
enum1 = ["foo"] |> Stream.take(1)
enum2 = ["bar"]
stream = Stream.concat(enum1, enum2)
zipwith = ["!", "!"]
zipped = Stream.zip(stream, zipwith)
Enum.to_list(zipped)
```
results in:
```Elixir
** (FunctionClauseError) no function clause matching in Stream.do_zip_step/2    
    
    The following arguments were given to Stream.do_zip_step/2:
    
        # 1
        "bar"
    
        # 2
        ["foo"]
    
        defp do_zip_step(x, [])
    
    (elixir) lib/stream.ex:1193: Stream.do_zip_step/2
    (elixir) lib/enum.ex:3251: Enumerable.List.reduce/3
    (elixir) lib/stream.ex:897: Stream.do_list_transform/7
    (elixir) lib/stream.ex:1166: Stream.do_zip_next_tuple/5
    (elixir) lib/stream.ex:1142: Stream.do_zip/3
    (elixir) lib/enum.ex:1919: Enum.reverse/1
    (elixir) lib/enum.ex:2588: Enum.to_list/1
```

### Expected behavior
```Elixir
[{"foo", "!"}, {"bar", "!"}]
```

ggcampinho:
Fixes #7667

### DIFF
--- a/lib/elixir/lib/stream.ex
+++ b/lib/elixir/lib/stream.ex
@@ -938,8 +938,7 @@ defmodule Stream do
     else
       # Only take into account outer halts when the op is not halt itself.
       # Otherwise, we were the ones wishing to halt, so we should just stop.
-      {:halted, [:outer | acc]}
-      when op != :halt ->
+      {:halted, [:outer | acc]} when op != :halt ->
         do_transform_user(vals, user_acc, next_op, next, {:cont, acc}, funs)
 
       {:halted, [_ | acc]} ->
@@ -1124,7 +1123,7 @@ defmodule Stream do
 
     enum_funs =
       Enum.map(enumerables, fn enum ->
-        {&Enumerable.reduce(enum, &1, step), :cont}
+        {&Enumerable.reduce(enum, &1, step), [], :cont}
       end)
 
     do_zip(enum_funs, acc, fun)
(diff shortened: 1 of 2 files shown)


lib/elixir/lib/stream.ex
```diff
@@ -1165,18 +1164,20 @@ defmodule Stream do
...
     case fun.({:cont, []}) do
-      {:suspended, [elem], fun} ->
-        do_zip_next_tuple(zips, acc, callback, [elem | yielded_elems], [{fun, :cont} | buffer])
+      {:suspended, [elem | next_acc], fun} ->
+        next_buffer = [{fun, next_acc, :cont} | buffer]
+        do_zip_next_tuple(zips, acc, callback, [elem | yielded_elems], next_buffer)
 
-      {_, [elem]} ->
-        do_zip_next_tuple(zips, acc, callback, [elem | yielded_elems], [{fun, :halt} | buffer])
+      {_, [elem | next_acc]} ->
```
ggcampinho: I couldn't find an easy way to test this case, don't know a good way to procede

josevalim: What do you mean by testing this? The new test case does not hit this?

ggcampinho: I pushed 986f4b9 that clarifies the case

lib/elixir/test/elixir/stream_test.exs
```diff
@@ -1086,6 +1107,9 @@ defmodule StreamTest do
 
     assert Stream.chunk_every([0, 1, 2, 3], 2) |> Stream.zip() |> Enum.to_list() ==
              [{0, 2}, {1, 3}]
+
+    stream = %HaltAcc{acc: 1..3}
```
josevalim: Do we need to use a separate module? Couldn't we use a range + Stream.take for this test?

ggcampinho: If I understood correctly when this happens range + Stream.take would return only the last element halted, so it matches with the previous clause `{_, [elem]}`. In my case, the halt tuple won't match because it has 3 elements

josevalim: Oh, I see. Perfect then!

lib/elixir/test/elixir/stream_test.exs
```diff
@@ -196,6 +212,11 @@ defmodule StreamTest do
     assert Enum.take(stream, 13) == [1, 2, 3, 1, 2, 3, 1, 2, 3, 1, 2, 3, 1]
   end
 
+  test "concat/2 is zippable" do
+    stream = Stream.concat(1..2 |> Stream.take(2), 3..4)
```
whatyouhide: I think this pipe expression is a bit weird. What about:

```elixir
stream = 1..2 |> Stream.take(2) |> Stream.concat(3..4)
```

?

josevalim:

PR #7674, opened 2018-05-11
Remove unnecessary catch from the parser

ggcampinho:
The throws with this pattern are [caught by yecc](https://github.com/erlang/otp/blob/master/lib/parsetools/include/yeccpre.hrl#L67)
and returned as an error. Also renamed throw to error to avoid confusions (matching with function [return_error](https://github.com/erlang/otp/blob/master/lib/parsetools/include/yeccpre.hrl#L51))


josevalim:

PR #7675, opened 2018-05-12
Use first..last in ranges

eksperimental:
As opposed to `left..right` and `x..y`

### DIFF
--- a/lib/elixir/lib/range.ex
+++ b/lib/elixir/lib/range.ex
@@ -59,7 +59,7 @@ defmodule Range do
 
   # TODO: Remove by 2.0
   @doc false
-  @deprecated "Pattern match on left..right instead"
+  @deprecated "Pattern match on first..last instead"
   def range?(term)
   def range?(first..last) when is_integer(first) and is_integer(last), do: true
   def range?(_), do: false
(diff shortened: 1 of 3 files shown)


eksperimental:
This is the terminology used in the Range module
> Ranges can be either increasing (first <= last) or decreasing (first > last). Ranges are also always inclusive.

personally, I don't think it's confusing. if you have a range `100..1`, 100 is still the first integer in the range, regardless one being < than 100.

yordis:
I would prefer to read `first` and `last` because @michalmuskala I see your issue as something that is relative what is `first` and `last` in your terminology (Relativity issue)

lib/elixir/lib/range.ex
```diff
@@ -70,25 +70,20 @@ defimpl Enumerable, for: Range do
     reduce(first, last, acc, fun, _up? = last >= first)
   end
 
-  defp reduce(_x, _y, {:halt, acc}, _fun, _up?) do
-    {:halted, acc}
-  end
```
josevalim: Please avoid changing the syntax. If the issue is about renaming variables, then it should keep the other syntax the same.

josevalim:

PR #7676, opened 2018-05-12
Rename single-char variables from docs

eksperimental:


lib/elixir/lib/enum.ex
```diff
@@ -131,10 +131,10 @@ defprotocol Enumerable do
 
   As an example, here is the implementation of `reduce` for lists:
 
-      def reduce(_,       {:halt, acc}, _fun),   do: {:halted, acc}
-      def reduce(list,    {:suspend, acc}, fun), do: {:suspended, acc, &reduce(list, &1, fun)}
-      def reduce([],      {:cont, acc}, _fun),   do: {:done, acc}
-      def reduce([h | t], {:cont, acc}, fun),    do: reduce(t, fun.(h, acc), fun)
+      def reduce(_list,         {:halt, acc}, _fun),   do: {:halted, acc}
+      def reduce(list,          {:suspend, acc}, fun), do: {:suspended, acc, &reduce(list, &1, fun)}
```
lexmag: It'd be better to have aligning removed, right?

josevalim: Yes, given the renaming is causing things to misalign, let's remove the alignment too.

josevalim:

PR #7677, opened 2018-05-12
Optimization for ranges in Enum.sort/1 and Enum.to_list/1

eksperimental:


lib/elixir/lib/enum.ex
```diff
@@ -2216,6 +2216,19 @@ defmodule Enum do
     :lists.sort(enumerable)
   end
 
+  def sort(first..last = range) do
```
lexmag: Is this realistic use case (range sorting) to warrant special casing?

josevalim:
I agree with @lexmag! Those cases are not common and we already optimize ranges on reduce. This means those operations already perform reasonably well. Thanks @eksperimental!

eksperimental:
I agree as well that the sort for range is not common, but to_list/1 is a different story, and we are unnecessarily traversing the list twice, when just by flipping the range and traversing once works just fine

Issue #7673, opened 2018-05-11
String.downcase/2 with `:ascii ` performs worse than `:default`


tckb:
While routinely checking on performance of `String.downcase`,  I noticed that `String.downcase/2` with `:ascii ` performs much worse the default option. 

As per the document, it _should_ perform better, if not several orders of magnitude better than the default case.  My test input is always an ASCII input without any smell of unicode characters in it. 

I have tried it with a simple benchmark script. 
### Benchmark script
```
Benchee.run(
      %{
        "string downcase" => fn -> String.downcase("ABCDEFGHIJKLMNO") end,
        "string downcase ascii" => fn -> String.downcase("ABCDEFGHIJKLMNO", :ascii) end
      },
      warmup: 5,
      time: 10,
      formatter_options: %{
        console: %{
          extended_statistics: true
        }
      }
    )
```

#### Results
```
Name                            ips        average  deviation         median         99th %
string downcase              1.25 M        0.80 μs  ±1023.30%        0.60 μs        2.40 μs
string downcase ascii        0.61 M        1.65 μs   ±297.88%        1.30 μs        4.40 μs

Comparison:
string downcase              1.25 M
string downcase ascii        0.61 M - 2.06x slower

Extended statistics:

Name                          minimum        maximum    sample size                     mode
string downcase               0.60 μs     4876.20 μs         1.13 M                  0.60 μs
string downcase ascii         1.20 μs     1537.40 μs       572.37 K                  1.30 μs```
```

This seems quite surprising to me, even counter-intuitive. Am I missing something here?

### Environment
Operating System: macOS
CPU Information: Intel(R) Core(TM) i5-5257U CPU @ 2.70GHz
Number of Available Cores: 4
Available memory: 8 GB
Elixir 1.6.4
Erlang 20.3.4

michalmuskala:
I checked myself that this is the case. I don't understand yet, how that's possible, though.

tckb:
Ah @michalmuskala , I was wondering if I have to test it further on another platform. thanks for the confirmation

bbense:
I get similar results on OS X. 

Name                            ips        average  deviation         median         99th %
string downcase            843.38 K        1.19 μs  ±1433.66%        0.70 μs        3.80 μs
string downcase ascii      260.78 K        3.83 μs  ±3878.60%           2 μs           7 μs

But the deviations are so high I wonder what else is going on.

aerosol:
[...]

PR #7680, opened 2018-05-12
Added performant impl for string upcase/downcase `:ascii` mode.

tckb:
the existing implementation with binary comprehensions turned out to be
_far_ slower than the other modes.

The current implementation is >= 2.5X faster than the earlier implementation

---
this PR is a consequence of the discussions in #7673

PS: this is my first PR in elixir

### DIFF
--- a/lib/elixir/lib/string.ex
+++ b/lib/elixir/lib/string.ex
@@ -677,13 +677,10 @@ defmodule String do
     upcase_ascii(string)
   end
 
-  defp upcase_ascii(string), do: upcase_ascii(string, [])
-
-  defp upcase_ascii(<<c, rest::bits>>, acc) when c >= ?a and c <= ?z,
-    do: upcase_ascii(rest, [c - 32 | acc])
-
-  defp upcase_ascii(<<c, rest::bits>>, acc), do: upcase_ascii(rest, [c | acc])
-  defp upcase_ascii(<<>>, acc), do: IO.iodata_to_binary(:lists.reverse(acc))
+  defp upcase_ascii(string), do: IO.iodata_to_binary(upcase_ascii(string))
+  defp upcase_ascii(<<c, rest::bits>>) when c >= ?a and c <= ?z, do: [c - 32 | upcase_ascii(rest)]
+  defp upcase_ascii(<<c, rest::bits>>), do: [c | upcase_ascii(rest)]
+  defp upcase_ascii(<<>>), do: []
 
   def upcase(string, mode) when mode in @conditional_mappings do
     String.Casing.upcase(string, [], mode)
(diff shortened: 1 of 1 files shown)


josevalim:
@tckb Although I said to go with `dn4_1`, @michalmuskala brought good points about `dn4_2` and memory use, so can you please migrate to `dn4_2` instead? Thank you!

tckb:
@josevalim  on further benchmarking with larger & different inputs, I have not seen any ***consistent*** performance difference between `dn4_1(downcase_ascii_patternMatch)` and `dn4_2(downcase_ascii_iodata)`. For memory consumption, dn4_2 has _slightly_ better memory usage than the other. In some case (all characters being same case), I have noticed `dn4_1`  having better memory usage.

https://gist.github.com/tckb/ee13867e8a50e069afb04b25fc9efd85
https://gist.github.com/tckb/4cfd89a3c2e10f307f681679b528b099


In either case, these have methods have outperformed the standard libs method several folds.

https://gist.github.com/tckb/6e0a55e9482e5de785059e3a2de2b4e2

josevalim:
It is your call @michalmuskala.

michalmuskala:
Let's go with the body-recursive version. That's the implementation we use for `Enum.map` (through `:lists.map`), so let's be consistent.

tckb:
thanks @michalmuskala  @josevalim  I've changed the impl to `dn4_2`  I'm assuming the test cases related to this method are already in-place. I haven't created one

michalmuskala:
The `upcase_ascii` function is implemented in-between the `upcase` clauses. This causes a compiler warning. Let's move `upcase_ascii` after all the `upcase` clauses. Similar with `downcase_ascii`.

tckb:
Ah! good catch @michalmuskala , was wondering about that. Is there any specific side-effect of this?

michalmuskala:
Not really, just tidiness.

lib/elixir/lib/string.ex
```diff
@@ -674,15 +674,17 @@ defmodule String do
...
   def upcase(string, :ascii) when is_binary(string) do
-    for <<x <- string>>,
-      do: if(x >= ?a and x <= ?z, do: <<x - 32>>, else: <<x>>),
-      into: ""
+    IO.iodata_to_binary(upcase_ascii(string))
   end
 
   def upcase(string, mode) when mode in @conditional_mappings do
     String.Casing.upcase(string, [], mode)
   end
```
eksperimental: Im not with a computer but would `when char in ?a..?z, do:` work? If so that will read better

lib/elixir/lib/string.ex
```diff
@@ -674,15 +674,17 @@ defmodule String do
...
 
+  defp upcase_ascii(<<c, rest::bits>>) when c >= ?a and c <= ?z, do: [c - 32 | upcase_ascii(rest)]
```
whatyouhide: I think we should do

```elixir
defp upcase_ascii(<<c, rest::bits>>) when c in ?a..?z, do: ...
```

The `in` seems clearer to me and is compiled to the same thing so we should be fine performance-wise.

eksperimental: Within the same minute we both commented the same

josevalim: They do not compile the same (there is an additional is_integer check which hmay be removed by the compiler or not) but we can also run into bootstrap issues, unfortunataely.

tckb: @josevalim  so `in` creates another validation check, regardless of `..` ? I thought only integers are accepted in Range

tckb: right, the `Kernel.".."/2` macros has the integer checks for the range

josevalim: I think the extra check is not supposed to matter in future erlang versions anyway but the biggest issue here is bootstrapping (`in` may not be available when we compile the `String` module).

michalmuskala: I believe the `is_integer` check on extracting data from a binary will only be eliminated on OTP 21 (I contributed that optimisation recently).

tckb:
okay, after further reviewing, here are the todo's

- [ ] change `c` to more "descriptive" name 
- ~use `in` instead of explicit >, <~ not considering as it might cause issue while bootstrapping. 

I'll hold off the push just in-case there are further changes ;) 

@josevalim @michalmuskala @whatyouhide  @eksperimental  :D

lib/elixir/lib/string.ex
tckb: @eksperimental normally, yes -- but as per @josevalim  using `in` at this level will cause bootstrapping issues.

josevalim:

PR #7682, opened 2018-05-13
Extend IEx autocomplete docs

wojtekmach:


wojtekmach:
Btw, given `@doc false` functions are not autocompleted, should `@moduledoc false` modules not be as well?

josevalim:
> Btw, given @doc false functions are not autocompleted, should @moduledoc false modules not be as well?

To me it is fine as is, because you would need to access the module first and if you are trying to reach a private module it is probably intentional. This is different from private functions where they are mixed. Maybe we could exclude it when listing the modules but those lookups become quite expensive (one fs lookup per module shown in the autocomplete).

wojtekmach:
> To me it is fine as is, because you would need to access the module first and if you are trying to reach a private module it is probably intentional.

Oh, I should have mentioned that I hit this when I did: `MyApp.<tab>` and `MyApp.Application` and `MyApp.MixProject` showed up. But yeah, not a big deal 👍

josevalim:

PR #7686, opened 2018-05-15
DynamicSupervisor documentation fix

adkron:
The documentation is refering to an `id` for terminating a child, but
DynamicSupervisor uses a `pid`.

Amos King @adkron <amos@binarynoggin.com>


josevalim:
Thanks @adkron! I have added backticks to pid since we pass it as an argument, so it is part of the signature.

PR #7687, opened 2018-05-15
Use pdict instead of Agent on Mix.Config

josevalim:
This simplifies and speeds up the code as we always modify
the process dictionary directly instead of building multiple
keyword lists and merging them on imports.

This commit also deprecate APIs that were public but not
used anywhere in Mix (nor tested).


josevalim:

Issue #6646, opened 2017-10-09
Code formatter incorrectly formats multiline lists when there is a trailing comma


pragdave:
The docstring says

> As mentioned in the previous section, lists, tuples, bitstrings, maps, and
structs will be broken into multiple lines if they are followed by a newline in
the opening bracket and preceded by a new line in the closing bracket lines.

However, this code:

~~~ elixir
  @deps [
    { :private, "> 0.0.0" },
    { :ex_doc,  "~> 0.14", only: :dev, runtime: false },
  ]
~~~

gets formatted as:

~~~ elixir
  @deps [{:private, "> 0.0.0"}, {:ex_doc, "~> 0.14", only: :dev, runtime: false}]
~~~

Cheers

Dave

fishcakez:
This looks like an issue with multiline lists combined with a trailing comma. The formatter will remove the trailing commas but then not honor the multiline list. Rather it should remove trailing comma and honor the multiline list.

pragdave:
Fwiw, multi line lists are exactly where a trailing comma is good...


On Sun, Oct 08, 2017 at 11:05pm, James Fish < notifications@github.com [notifications@github.com] > wrote:
This looks like an issue with multiline lists combined with a trailing comma. The formatter will remove the trailing commas but then not honor the multiline list. Rather it should remove trailing comma and honor the multiline list.

fishcakez:
> Fwiw, multi line lists are exactly where a trailing comma is good...

Please try to stay objective :). A topic like trailing commas can invite lots of opinions and not many facts. Elixir current style does not use trailing commas and the formatter follows the existing style. If you can show that it would be of benefit please propose it separately.

pragdave:
Objectively:


1. A trailing comma never hurts.


2. A trailing comma makes it easier to add


3. A trailing comma means you can sort a kv literal on key.


On Sun, Oct 08, 2017 at 11:21pm, James Fish < notifications@github.com [notifications@github.com] > wrote:
Fwiw, multi line lists are exactly where a trailing comma is good...

Please try to stay objective :). A topic like trailing commas can invite lots of opinions and not many facts. Elixir current style does not use trailing commas and the formatter follows the existing style. If you can show that it would be of benefit please propose it separately.

fishcakez:
[...]

Issue #6647, opened 2017-10-09
Could code formatter leave spaces in multiline keyword lists?


pragdave:
I like to make multiline keyword lists easier to read by aligning the values into a column:

~~~ elixir
  def project do
    in_production = Mix.env == :prod
     [
      app:     "fred",
      version: "1.2",
      elixir:  ">= 1.4.0",
      deps:    deps()
     ]
end
~~~

The code formatter removes these extra spaces:

~~~ elixir
  def project do
    in_production = Mix.env() == :prod

    [
      app: "fred",
      version: "1.2",
      elixir: ">= 1.4.0",
      deps: deps()
    ]
  end
~~~

I would be really, really nice if it kept them. (And even nicer still if it added them to do the alignment for you... :)

fishcakez:
Hi @pragdave. I believe the Elixir code base used to be aligned more then half the time, and then moved away from aligning about a year ago. The formatter is following the standard style used by Elixir.

Note that aligning like this can cause many lines to be altered when one entry becomes longer. This hurt the review process, history and maintenance. Of course the later point is much less significant now we have a tool to handle it. Unfortunately allowing alignment when all are aligned would hit these issues, even with maintenance as lines need to be done by hand and its possible mistake would unalign. Hopefully developer environments can provide enough features, such as syntax highlighting to keep the unaligned way clear.

With the new formatter there is going to be compromises for everyone (especially amongst those who worked on it so far), and where possible follows Elixir's existing style. I hope people will stay objective about this and see that while not everything is exactly how they personally write code the consistency will benefit everyone.

pragdave:
Well, as José can attest, I've never been a fan of the current Elixir style, and have many times proposed different ways of doing it. The code created by the standard generators, which was then in the standard "Elixir style" was eventually changed into something more reasonable as a result of my suggestions.


Elixir is a lovely language, but the way it is often written makes it look ugly, which is a shame. Everything is packed together, with little whitespace, no alignment of related things, and long line after long line.


The fact that the new formatter preserves most line breaks is great. In fact, I have no problems with 99%of what it does. Two of my three issues were simple bugs I found while trying it on a typical mix.exs file.


The plea to preserve whitespace in multi line lists is different. It's a request to start a discussion. I'd like to have a formatter that reveals the structure and beauty of the code, rather than one that tries to save on whitespace.


I agree it is a personal thing, but ultimately that's not bad if it is based on experience and not just opinion.


[...]

PR #7689, opened 2018-05-16
Add the `:trailing_comma` option to the formatter

gamache:
This commit adds support for the `:trailing_comma` option in `Code.format_string!/2` et al, including in `.formatter.exs`.

When set `true`, multi-line list, map, and struct literals will include a trailing comma after the last item or pair in the data structure.

Does not affect other data structures, or lists/maps/structs rendered on a single line.

--

The topic of trailing commas has been raised previously by @pragdave in #6646.  No resolution was reached; instead, it was considered appropriate to wait for the formatter to have more time in common use before making decisions on which syntax variants to allow.  It's now been a little over six months since the last discussion, and in conversations with Elixir developers, the lack of trailing commas continues to be a common source of friction.

Some possible reasons to prefer trailing commas in multi-line data structures are ease of refactoring, minimal diffs when adding or removing elements, and visual symmetry.  These reasons are not worth discussing directly in this PR.  It should suffice to note that there are advantages and disadvantages to either approach, and that many developers and teams feel strongly about their preference.

This PR's goal is not to dictate the style of the Elixir project or community at large.  In a similar way to the existing `:line_length` option, the `:trailing_comma` option provides an opt-in mechanism by which projects and organizations can benefit from standardized code formatting while retaining a style they find to be more efficient and productive.

### DIFF
--- a/lib/elixir/lib/code.ex
+++ b/lib/elixir/lib/code.ex
@@ -261,6 +261,11 @@ defmodule Code do
       expects a valid `Version` which is usually the minimum Elixir
       version supported by the project.
 
+    * `:trailing_comma` - if set `true`, multi-line list, map, and
+      struct literals will include a trailing comma after the last item
+      or pair in the data structure. Does not affect argument lists,
+      tuples, or lists/maps/structs rendered on a single line.
+
   ## Design principles
 
   The formatter was designed under three principles.
--- a/lib/elixir/lib/code/formatter.ex
+++ b/lib/elixir/lib/code/formatter.ex
@@ -194,7 +194,7 @@ defmodule Code.Formatter do
 
   Returns `{:ok, doc}` or `{:error, parser_error}`.
 
-  See `format!/2` for the list of options.
+  See `Code.format_string!/2` for the list of options.
   """
   def to_algebra(string, opts \\ []) when is_binary(string) and is_list(opts) do
     file = Keyword.get(opts, :file, "nofile")
@@ -260,6 +260,7 @@ defmodule Code.Formatter do
       locals_without_parens: locals_without_parens,
       operand_nesting: 2,
       rename_deprecated_at: rename_deprecated_at,
+      trailing_comma: Keyword.get(opts, :trailing_comma, false),
       comments: comments
     }
   end
@@ -1401,7 +1402,7 @@ defmodule Code.Formatter do
     fun = &quoted_to_algebra(&1, :parens_arg, &2)
 
     {args_doc, _join, state} =
-      args_to_algebra_with_comments(args, meta, false, :none, join, state, fun)
+      args_to_algebra_with_comments(args, meta, false, :state, join, state, fun)
 
     {surround("[", args_doc, "]"), state}
   end
@@ -1412,7 +1413,7 @@ defmodule Code.Formatter do
     {left_doc, state} = fun.(left, state)
 
     {right_doc, _join, state} =
-      args_to_algebra_with_comments(right, meta, false, :none, join, state, fun)
+      args_to_algebra_with_comments(right, meta, false, :state, join, state, fun)
 
     args_doc =
       left_doc
@@ -1428,7 +1429,7 @@ defmodule Code.Formatter do
     fun = &quoted_to_algebra(&1, :parens_arg, &2)
 
     {args_doc, _join, state} =
-      args_to_algebra_with_comments(args, meta, false, :none, join, state, fun)
+      args_to_algebra_with_comments(args, meta, false, :state, join, state, fun)
 
     name_doc = "%" |> concat(name_doc) |> concat("{")
     {surround(name_doc, args_doc, "}"), state}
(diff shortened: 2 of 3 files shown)


pmarreck:
I have been nothing but impressed with the formatter (which I honestly did not originally expect)... _Except for_ this little detail! I love me some trailing commas for all the reasons cited. (Possibly related: I'm also a fan of the Oxford comma!)

In short, I think this would be a useful feature to add.

josevalim:
Hi @gamache! 👋 

As you know, the goal of the formatter is exactly to unify the code style used by the community. While I understand this PR is an argument to relax this particular goal, the formatter has been publicly released for roughy 4 months. We need to get much more mileage and feedback before we can perform such changes.

Note we had long discussions in the past about formatter customization, such as #6647, and the conclusion there was the same: t is too soon in the formatter adoption cycle to warrant configuration options.

I also would like to note that the combination of optional parens and keyword lists make trailing commas slightly inconsistent in Elixir as the following code is ambiguous (and therefore disallowed):

```elixir
some_fun :foo,
                 key: bar,
baz
```

This was one of the main reasons for disallowing trailing commas in Elixir, as in an attempt to standardize a developer could end-up adding comma to warranted places.

For all of those reasons, we won't merge this PR. I understand many will disagree with this decision, but we don't plan to add new formatter options at least until we release Elixir v1.8 (Jan/2019) and the formatter gets more mileage.

gamache:
Thanks, @josevalim.  In that case, I eagerly await the release of 1.7.0 so I can submit this again.  :smiley:

OvermindDL1:
> I also would like to note that the combination of optional parens and keyword lists make trailing commas slightly inconsistent in Elixir as the following code is ambiguous (and therefore disallowed):

Which is exactly why they should not be allowed in parenthesis-less function calls, but should be in multi-line parenthesis function calls, lists, maps, tuples, etc...).

josevalim:
It would be after but I would like to reaffirm my personal position on the formatter in general:

  1. I am personally not in favor of global formatter options, especially options that are divergent (i.e. if you toggle the option on and off, every time you get a different formatting)

  2. The direction that I would most likely accept is one that allows trailing commas with some special annotations, there have been some discussions on this area on #6647

  3. I am personally staying away from formatter discussions, so someone should be responsible to build the proposal, collect feedback and push it through. If you need a reference of what a proposal looks like and how to collect feedback from the community, you can see many of the past proposals done by the Elixir team on the [Elixir Forum](https://elixirforum.com/c/elixir-news). There is also [a great example from @blatyo on the forum](https://elixirforum.com/t/proposal-add-field-puns-map-shorthand-to-elixir/15452). I recommend using the ElixirForum for gathering feedback too but it is up to you. You are welcome to also link to it on the elixir-lang-core mailing list. Once a proposal is written and validated, we will gladly get involved

eagle-te:
So....., `:trailing_comma` option in release v1.10? :p

Personally, I see no reason to not include this. The toggling reason, @josevalim mentioned, could also be applied to line_length option.

There are objectively over weighing good reasons to allow a trailing comma option, such as https://github.com/elixir-lang/elixir/issues/6646#issuecomment-335068731 , which shouldn't be dismissed by non productive personal reasoning.

josevalim:
> The toggling reason, @josevalim mentioned, could also be applied to line_length option.

Not quite the same. If you have line_length set to 100 and then you set to 80, it will surely introduce line breaks in *some* cases. But if you roll it back to 100, everything will remain the same.

Trailing commas change everything every time you change it.

On the good news, Elixir v1.10 makes public many of the private APIs used by formatter, so now it should be simpler and more stable for third party libraries to provide their own formatting, such as the freedom_formatter from @gamache.

eagle-te:
Well, I expected at least an answer of compromise, like with `:allowing_trailing_commas`.
Which on `true` doesn't remove them and add no new ones, but the formatter behaves default, if disabled. This would act non destructive, non toggling, like `:line_length`, I hinted at.

Thx for the info anyway, probably looking for third party formatters on v1.10 release. :)

snowe2010:
Is it time to revisit this? From the arguments I've seen:

Pros:

1. Trailing commas are good in all locations except parenthesis-less function calls
2. Trailing commas should not be _added_ by the formatter, just removed. So toggling the option **off** removes all trailing commas, but turning it **on** does nothing except allow them. 
3. Trailing commas should only be allowed in multiline situations
4. Trailing commas have many upsides and few downsides:
    1. Adding a line only changes that line in your VCS diff
    2. Adding a line is much easier. From copy/paste to duplicating a line, no need to add a comma to the end
    3. Sorting the lines is easier

Cons:

1. Like any formatter option, it allows for something different. 
2. Some people might not understand trailing commas. Personally I think this is a terrible reason, as there are many, many other things beginners wouldn't understand before they got to a trailing comma.

Sija:
Hi! Elixir formatter is a pleasure to work with, except for this one behavior. Could this be revisited, please? Over 4 years has passed since opening this PR and sth like trailing commas are commonly found in virtually every other programming language along with its tooling (linters and/or formatters). Amount of upsides clearly outweigh the few downsides there are.

marcandre:
@Sija in case it helps: I am [maintaining `freedom_formatter`](https://github.com/marcandre/freedom_formatter) which supports trailing commas. The new builtin formatter allows plugins and even the [default elixir formatter can now be overridden](https://github.com/elixir-lang/elixir/pull/11507) this way. Caveats: I haven't taken the time to setup `freedom_formatter` to be directly usable as a formatter plugin, it still isn't as easy as if it was builtin, and I haven't checked what level of support packages like `ElixirLS` / Editor extensions currently offer.

pmarreck:
one could get very clever with custom sigil macros that treated unescaped unquoted whitespace as an item separator in lists, keyword lists, maps, etc., the way lisps do etc. (and the way sigils like ~w already work!) without needing to alter the formatter's interpretation of commas

PR #7690, opened 2018-05-16
Fix reference in Code.Formatter documentation

fertapric:


gamache:
I'm glad at least some of #7689 might make it in!

PR #7691, opened 2018-05-16
More precise started and loaded_applications typespec

arcz:
Update to match the wrapped functions: https://github.com/erlang/otp/blob/1526eaead833b3bdcd3555a12e2af62c359e7868/lib/kernel/src/application.erl#L269


josevalim:

PR #7692, opened 2018-05-17
Fix faulty function call in comment

linduxed:
This PR is nothing but a fix for the documentation of `Task.Supervisor`.

A comment containing an example function call to `Supervisor.start_link/2` is written without any children passed in.


josevalim:

PR #7695, opened 2018-05-19
consistently use "supervision strategy" term

davidsulc:
Using "supervision strategy" consistently makes the documentation easier to follow, and removing the previous uses of "restart strategy" also avoids confusion with the `:restart` value used in child specifications.

### DIFF
--- a/lib/elixir/lib/supervisor.ex
+++ b/lib/elixir/lib/supervisor.ex
@@ -97,7 +97,7 @@ defmodule Supervisor do
   Supervisors support different strategies; in the example above, we
   have chosen `:one_for_one`. Furthermore, each supervisor can have many
   workers and supervisors as children, each of them with their specific
-  configuration, shutdown values, and restart strategies.
+  configuration, shutdown values, and supervision strategies.
 
   The rest of this document will cover how child processes are started,
   how they can be specified, different supervision strategies and more.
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/supervisor.ex
```diff
@@ -97,7 +97,7 @@ defmodule Supervisor do
   Supervisors support different strategies; in the example above, we
   have chosen `:one_for_one`. Furthermore, each supervisor can have many
   workers and supervisors as children, each of them with their specific
-  configuration, shutdown values, and restart strategies.
+  configuration, shutdown values, and supervision strategies.
```
josevalim: In this case, however, I believe we were referring to the restart value, since we are talking about the children. Maybe we can use "shutdown and restart values"?

davidsulc: You're right, of course :). Since this seems to be discussing child specs, maybe this sentence could be changed to

> Furthermore, each supervisor can have many
   workers and/or supervisors as children, with each one having its own configuration (as outlined in the “Child specification” section).

josevalim: Sounds good to me!

davidsulc: Updated, as discussed

josevalim:

PR #7696, opened 2018-05-19
Move cond to Kernel

### LINKED ISSUE #7693: Explore if cond could be a regular macro
Currently `cond` is a special form handled by the compiler. The primary reason for this, in the past, was to correctly handle variable scoping and limit leaking variables between clauses.

Currently, after we changed the scoping rules of `case`, I believe, it should be possible to implement `cond` as a regular macro on top of `case`.

Note: the new implementation should make sure to optimise boolean matches in case a pure-boolean operator is used.

ggcampinho:
This aims to remove cond from special forms and implement it as macro, maintaining the logic as close as possible from the original.

Also, it fixes the following error with invalid keyword lists:
```
iex(1)> cond([{:do, (1 -> 1)}, :foo])
** (FunctionClauseError) no function clause matching in :elixir_clauses.expand_cond/4
```

Now, the error looks like:
```
iex(1)> cond([{:do, (1 -> 1)}, :foo])
** (ArgumentError) invalid arguments for "cond"
```

Closes #7693


lib/elixir/lib/kernel.ex
```diff
@@ -2642,6 +2642,179 @@ defmodule Kernel do
...
+
+  defp build_cond_clauses([], acc, _old_meta, _module) do
+    acc
+  end
+
+  defp build_cond_clauses([{:->, new_meta, [[condition], body]} | tail], acc, old_meta, module) do
+    var = {:cond, [], module}
+
+    case_clause =
+      {:case, new_meta,
```
ggcampinho: I can't find a way to make this look nice

lib/elixir/src/elixir_expand.erl
```diff
@@ -658,21 +685,35 @@ rewrite_case_clauses([{do, [
...
+  ]},
+  {'->', FalseMeta, [
+    [{'_', _, _}],
+    FalseExpr
+  ]}
+]}]) ->
+  {FalseMeta, FalseExpr, TrueMeta, TrueExpr};
 
-rewrite_case_clauses([{do, [
+extract_boolean_clauses([{do, [
```
ggcampinho: I have the impression that this is not necessary anymore

lib/elixir/lib/kernel.ex
```diff
@@ -2642,6 +2642,179 @@ defmodule Kernel do
...
+            end
+
+          build_cond_clauses(reversed_clauses, error, meta, module)
+      end
+
+    replace_case_meta(case_clause, line)
+  end
+
+  defp build_cond([], _caller) do
+    :erlang.error(ArgumentError.exception("missing :do option in \"cond\""))
```
josevalim: Can't we use `raise` here as long as we define it after `defmacro raise`?

lib/elixir/lib/kernel.ex
josevalim: We should use `quote/unquote` here, no?

josevalim:
Thanks @ggcampinho! Since you are translating from Erlang to Elixir, it seems you are doing more work than you should, for example, setting up `module` and `line` metadata. This is all done by the macro system, so should be able to handle those.

Here is the `cond` implementation [before we moved it to a special form](https://github.com/elixir-lang/elixir/commit/64f8541ee61a796d7ecd8782f5ef143e79eb728a), in case it helps with ideas.

lib/elixir/lib/kernel.ex
```diff
@@ -1473,8 +1473,8 @@ defmodule Kernel do
 
   ## Implemented in Elixir
 
-  defp optimize_boolean({:case, meta, args}) do
-    {:case, [{:optimize_boolean, true} | meta], args}
+  defp optimize_boolean({:case, meta, args}, generated \\ true) do
+    {:case, [{:optimize_boolean, generated: generated} | meta], args}
```
josevalim: Sorry but I do not understand this change. Why are we defining such tuple with tuples inside?

lib/elixir/lib/kernel.ex
```diff
@@ -2642,6 +2642,179 @@ defmodule Kernel do
...
+    replace_case_meta(case_clause, line)
+  end
+
+  defp build_cond([], _caller) do
+    :erlang.error(ArgumentError.exception("missing :do option in \"cond\""))
+  end
+
+  defp build_cond(clauses, _caller) when is_list(clauses) do
+    check_at_most_once(:do, clauses, 0, "cond")
+    check_unexpected_options([:do], clauses, "cond")
```
josevalim: It feels we could simplify this code. The code in Erlang was written in a generic way because it is used in multiple parts. I think here we can just bail and say that cond was calling incorrectly and it expects a single do keyword.

lib/elixir/src/elixir_expand.erl
```diff
@@ -407,6 +401,7 @@ expand({Name, Meta, Kind} = Var, E) when is_atom(Name), is_atom(Kind) ->
 
 expand({Atom, Meta, Args}, E) when is_atom(Atom), is_list(Meta), is_list(Args) ->
   assert_no_ambiguous_op(Atom, Meta, Args, E),
+  assert_no_underscore_clause_in_cond(Atom, Args, E),
```
josevalim: We can assert this in Elixir land.

lib/elixir/test/elixir/kernel/expansion_test.exs
```diff
@@ -957,157 +957,6 @@ defmodule Kernel.ExpansionTest do
...
-
-      after_expansion =
-        quote do
-          cond do
-            1 -> :"Elixir.List"
-            2 -> flatten([1, 2, 3])
-          end
-        end
-
-      assert expand(before_expansion) == after_expansion
```
michalmuskala: This test is very important. It would change semantics if it fails.

lib/elixir/src/elixir_expand.erl
```diff
@@ -223,12 +223,6 @@ expand({fn, Meta, Pairs}, E) ->
 
 %% Case/Receive/Try
 
-expand({'cond', Meta, [Opts]}, E) ->
-  assert_no_match_or_guard_scope(Meta, "cond", E),
```
josevalim: We should assert this in the Elixir code as well.

lib/elixir/test/elixir/kernel/expansion_test.exs
```diff
@@ -957,157 +957,6 @@ defmodule Kernel.ExpansionTest do
...
-        quote do
-          cond do
-            1 -> x = 1
-            2 -> y = 2
-          end
-
-          :erlang.+(x(), y())
-        end
-
-      assert expand(before_expansion) == after_expansion
```
michalmuskala: Similar the tests that checks if variable leak between clauses are very important as well.

lib/elixir/lib/kernel.ex
josevalim: To clarify, I think we should be able to not change this, as conds compile as a series of nested case clauses, and our optimization is enough to handle case clauses, no? :)

lib/elixir/lib/kernel.ex
josevalim: Take a look at how `if`, `||` and `&&` are implemented. The principle here should be similar.

lib/elixir/lib/kernel.ex
ggcampinho: Maybe it's not clear enough, the problem with this clause is that in the optimization during the expansion the clauses are marked with `generated: true`, for `cond` this is a problem because of (this)[https://github.com/elixir-lang/elixir/blob/master/lib/elixir/test/elixir/kernel/warning_test.exs#L609] warning.

I can switch this to an atom (but I couldn't find a good one) or bring the generated to Kernel, leaving only the optimization to be done in Erlang.

josevalim: > I can switch this to an atom (but I couldn't find a good one) or bring the generated to Kernel, leaving only the optimization to be done in Erlang.

I like this. Two questions then:

1. What happens if we don't mark them as generated?
2. If something bad happens, then let's bring it to kernel.

lib/elixir/lib/kernel.ex
```diff
@@ -2642,6 +2642,126 @@ defmodule Kernel do
...
+
+  defp get_line(meta) do
+    case :lists.keyfind(:line, 1, meta) do
+      {:line, line} -> line
+      false -> 0
+    end
+  end
+
+  defp replace_case_line({:case, meta, args}, line) do
+    new_meta = :lists.keyreplace(:line, 1, meta, {:line, line})
```
ggcampinho: I didn't know that `quote` can receive `line` option. I've changed in other places, but apparently, it doesn't work with pre-existing meta, with is strange for me:
```
quote line: 2 do
  unquote({:case, [line: 1], [true, [do: [{:->, [line: 1], [[true], true]}]]]})
end
```

lib/elixir/lib/kernel.ex
```diff
@@ -2642,6 +2642,126 @@ defmodule Kernel do
...
+        value when is_atom(value) and value != false and value != nil ->
+          {tail, body}
+
+        _ ->
+          error = quote(do: raise(CondClauseError))
+          {reversed_clauses, error}
+      end
+
+    case_clause = build_cond_clauses(next_clauses, next_acc, meta)
+    replace_case_line(case_clause, line)
```
josevalim: Do we need to do this? It is fine to me if we always use the meta of the `->` and never the meta of the `cond` call itself.

ggcampinho: [There was an explicit test for that](https://github.com/elixir-lang/elixir/blob/master/lib/elixir/test/erlang/control_test.erl#L186), so I was afraid to change. I don't think there are big implications of that, maybe a better error message when there are problems in the first condition.

lib/elixir/lib/kernel.ex
ggcampinho: It creates a bunch of warnings:
```
warning: this clause cannot match because a previous clause always matches
```

I'll move it to kernel

lib/elixir/src/elixir_expand.erl
ggcampinho: Done in a21757d. Although, the error is now an ArgumentError not a CompileError

lib/elixir/lib/kernel.ex
```diff
@@ -1567,23 +1575,29 @@ defmodule Kernel do
...
   end
 
   defmacro !value do
+    build_truthy_check(value, false, true)
+  end
+
+  defp build_truthy_check(condition, true_body, false_body) do
+    truthy_clause =
+      quote generated: true do
+        value when value != nil and value != false -> unquote(true_body)
```
ggcampinho: Changed here to follow the same pattern as `cond`, simplifying the optimization in the expansion. I've checked the history, and doesn't seem that there was a particular reason to use `:"Elixir.Kernel".in`

lib/elixir/test/elixir/kernel/expansion_test.exs
ggcampinho: I believe that this cond is leaking:
```
cond do
  x = 1 -> x
  true -> x
end
```

Because it becomes:
```
case x = 1 do
  value when value != nil and value != false -> x
  _ -> x # this has the same context as the first x
end
```

screen shot 2018-05-23 at 00 01 34

But I have the impression that we can just move the matches to the clauses

lib/elixir/test/elixir/kernel/expansion_test.exs
```diff
@@ -957,157 +957,6 @@ defmodule Kernel.ExpansionTest do
...
-            x = 1 -> x
-            true -> x
-          end
-        end
-
-      after_expansion =
-        quote do
-          cond do
-            x = 1 -> x
-            true -> x()
```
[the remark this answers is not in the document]
ggcampinho: Actually, there is one way, but it's a bit strange. It would be to put the case inside of a try and throw and catch the value and do the nesting inside of the catch.

```
try do
  value = x = 1

  (value == nil or value == false) and throw(:cond)
  x
catch
  :cond ->
    try do
      value = true

      (value == nil or value == false) and throw(:cond)
      x
    catch
      :cond ->
        raise CondClauseError
    end
end
```

lib/elixir/test/elixir/kernel/expansion_test.exs
josevalim: Hrm... so maybe we can't rewrite a `cond` to a `case` after all? I see no way to address this. :(

lib/elixir/lib/kernel.ex
josevalim: So we cannot use `!=` and `and` because otherwise we would need to optimize clauses only after expansion, as we are not quite sure what `and` and `!=` points to. Most likely it points to the kernel implementation but given we can import/unimport them, it may not be true. To avoid the need for expansion, we use the fully qualified module names and calls. So you would have to write: `:erlang.andalso(:erlang.'/='(...), :erlang.'/='(...))`. Given that's too verbose, we use `Elixir.Kernel.in/2` for simplicity.

lib/elixir/test/elixir/kernel/expansion_test.exs
ggcampinho: I don't see either, but the question that keeps hammering in my head is if this is a problem for macros in general?

If I understood correctly, it means that everytime you pass clauses to a macro, the context between clauses will be the same unless you find a way to separate the contexts in the generated code

josevalim:
I don’t think it is a general issue. It just means we really need cond as
an abstraction if we want to provide this kind of binding semantics.
-- 


josevalim:
Thanks @ggcampinho! I will go ahead and close this then, as we can't implement `cond` with `case` and therefore it deserves to be its own special form. Sorry for the false positive and great job as usual!

PR #7697, opened 2018-05-19
Match files starting with dot in "mix format"

fertapric:
Here is the example `.formatter.exs` file given in the documentation of `mix format`:

    [
      inputs: ["{mix,.formatter}.exs", "{config,lib,test}/**/*.{ex,exs}"]
    ]

With the previous implementation, the `.formatter.exs` file would not be formatted because internally `Path.wildcard/2` was used without expanding files starting with dot ".":

    iex> Path.wildcard("{mix,.formatter}.exs")
    ["mix.exs"]

    iex> Path.wildcard("{mix,.formatter}.exs", match_dot: true)
    [".formatter.exs", "mix.exs"]

### DIFF
--- a/lib/mix/lib/mix/tasks/format.ex
+++ b/lib/mix/lib/mix/tasks/format.ex
@@ -358,7 +358,7 @@ defmodule Mix.Tasks.Format do
 
     map =
       for input <- List.wrap(formatter_opts[:inputs]),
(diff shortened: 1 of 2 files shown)


fertapric:
🤔 Or could this be a bug in the `Path.Wildcard` implementation?

whatyouhide:
Good catch, thanks @fertapric! 💟

PR #7698, opened 2018-05-19
Add failing format test for assert with line limit

maxim:
I honestly don't know how to distill this formatting issue to bare minimum, but in my specific case this was generated by context generator:

```elixir
    test "create_account/1 with invalid data returns error changeset" do
      assert {:error, %Ecto.Changeset{}} = Accounts.create_account(@invalid_attrs)
    end
```

And despite that `.formatter.exs` has the setting `line_length: 79`, the assert line above stays at 82 characters no matter what. If I try to manually add a line break after assert or equals, it still resets it to what you see above.

This PR adds a test that also violates `line_length`, but I can't be 100% certain it's the same issue.

The test failure is as follows:

```
  1) test sigils with assert on line limit (Code.Formatter.GeneralTest)
     lib/elixir/test/elixir/code_formatter/general_test.exs:91
     Assertion with == failed
     code:  assert IO.iodata_to_binary(Code.format_string!(bad, opts)) == result
     left:  "assert foobar =\n         bazqux"
     right: "assert(\n  foobar =\n    bazqux\n)"
     stacktrace:
       lib/elixir/test/elixir/code_formatter/general_test.exs:103: (test)
```

As you can see in the left part the `assert foobar =` would violate the 10 char limit imposed by `@short_length`.

Edit: I might've accidentally put this into a wrong describe block. If this kind of test is welcome, happy to fix.

### DIFF
--- a/lib/elixir/test/elixir/code_formatter/general_test.exs
+++ b/lib/elixir/test/elixir/code_formatter/general_test.exs
@@ -88,6 +88,21 @@ defmodule Code.Formatter.GeneralTest do
       assert_format bad, good, @short_length
     end
 
+    test "with assert on line limit" do
+      bad = ~S"""
+      assert foobar = bazqux
+      """
+
+      good = ~S"""
+      assert(
+        foobar =
+          bazqux
+      )
+      """
+
+      assert_format bad, good, @short_length
+    end
+
     test "with heredoc syntax" do
       assert_same ~S"""
       ~s'''


josevalim:
Hi @maxim!

I believe your example is not the same issue. There are instances where the formatter will violate the character limit, such as long variables, long strings, etc. Given `assert` is in the list of calls without parens, I don't think we should add parens in the hope that it will fit the limit. As a slightly longer variable name will make this test just fail again.

Can you please give us the complete file where formatting files and let us know your Elixir version? Thank you.

maxim:
Hey, thanks @josevalim!

Turns out the issue wasn't with the formatter, it was the [ST3 ElixirFormatter plugin](https://github.com/karolsluszniak/elixir-formatter-sublime) not seeing the .formatter.exs file at the root of umbrella app. Running `mix format` by hand fixes those lines. Sorry about that. (This used Elixir 1.6.4).

That said, the test in this PR still begs a question. You're right that line length cannot be accommodated in all cases, but if there is a way, shouldn't we expect formatter to find it? (The test was run against Elixir master.)

josevalim:
In this case the formatter prefers to respect the user choice of not using
parens. You can always change the code to accommodate the line length
though.
-- 


josevalim:
I will close this as the first issue is resolved and the second won't fix. :)

PR #7701, opened 2018-05-22
Avoid calling truncate_n/2 if truncate is :infinity

blambi:
I encountered the following while investigating an issue that occurred in dev environment but not in testing (was apparently parsing some really broken data with Sweet_xml)

```elixir
13:29:32.638 module=Logger.Watcher [error] :gen_event handler Logger.ErrorHandler installed at :error_logger
** (exit) an exception was raised:                    
        (logger) lib/logger/utils.ex:57: Logger.Utils.truncate_n/2    
        (elixir) lib/enum.ex:1397: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
        (elixir) lib/enum.ex:1397: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
        (logger) lib/logger/utils.ex:141: Logger.Utils.scan_inspect/4
        (logger) lib/logger/error_handler.ex:170: Logger.ErrorHandler.translate/6
        (logger) lib/logger/error_handler.ex:92: Logger.ErrorHandler.log_event/6
        (logger) lib/logger/error_handler.ex:31: Logger.ErrorHandler.handle_event/2
        (stdlib) gen_event.erl:574: :gen_event.server_update/4  
```

After some prodding I noticed that `logger, truncate` was set to `:infinity` only in dev and if I added the same to test we got the same crash there.

### DIFF
--- a/lib/logger/lib/logger/utils.ex
+++ b/lib/logger/lib/logger/utils.ex
@@ -145,7 +145,7 @@ defmodule Logger.Utils do
     # arguments according to the truncate limit.
     {args, _} =
       Enum.map_reduce(args, truncate, fn arg, acc ->
-        if is_binary(arg) and is_integer(acc) do
+        if is_binary(arg) and acc != :infinity do
           truncate_n(arg, acc)
         else
           {arg, acc}
--- a/lib/logger/lib/logger/utils.ex
+++ b/lib/logger/lib/logger/utils.ex
@@ -145,7 +145,7 @@ defmodule Logger.Utils do
     # arguments according to the truncate limit.
     {args, _} =
       Enum.map_reduce(args, truncate, fn arg, acc ->
-        if is_binary(arg) do
+        if is_binary(arg) and is_integer(acc) do
           truncate_n(arg, acc)
         else
           {arg, acc}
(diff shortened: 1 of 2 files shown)


lib/logger/test/logger/utils_test.exs
```diff
@@ -90,52 +90,57 @@ defmodule Logger.UtilsTest do
     assert truncate(long_string, :infinity) == long_string
   end
 
-  test "inspect/2 formats" do
+  test "inspect/3 formats" do
```
josevalim: We should rather call it `scan_inspect/3` because that's what we are testing.

lib/logger/lib/logger/utils.ex
```diff
@@ -145,7 +145,7 @@ defmodule Logger.Utils do
     # arguments according to the truncate limit.
     {args, _} =
       Enum.map_reduce(args, truncate, fn arg, acc ->
-        if is_binary(arg) do
+        if is_binary(arg) and is_integer(acc) do
```
josevalim: Maybe we should have `acc != :infinity`? Otherwise someone may pass a completely invalid `acc` and we will silently discard it.

blambi: I do a agree with that, I would feel silly if I introduced another unexpected behaviour

josevalim:

PR #7702, opened 2018-05-23
Adds minor reference for Kernel functions Enum doc

tallysmartins:
- One may refer to Kernel functions to find the max and min values in enumerables
with just two elements.


michalmuskala:
I think it would be better to use slightly different wording, for example:

```
For selecting a minimal value out of two consider `Kernel.min/2`.
```

tallysmartins:
You made it way better. :+1:

eksperimental:
"Consider using" reads better IMO

michalmuskala:
Thank you! :heart:

PR #7717, opened 2018-05-28
Allow excluding input patterns in .formatter.exs

cybrox:
With wildcard support, the formatter configuration makes it very
easy to include whole directories full of files for formatting.
However, excluding a single file is not as trivial. While this is
not a regular use-case, there are definitely situations where it
is very useful.

This adds support for negating include patterns in the config
analog to how `.gitignore` works. The paths are resolved the same
way normal include paths are resolved and dropped from the file
list before passing it to the code formatter.

This allows for configurations like the following:
```
inputs: ["lib/**/*.{ex,exs}", "!lib/static.{ex,exs}"]
```

### DIFF
--- a/lib/mix/lib/mix/tasks/format.ex
+++ b/lib/mix/lib/mix/tasks/format.ex
@@ -356,16 +356,39 @@ defmodule Mix.Tasks.Format do
       Mix.raise("Expected :inputs or :subdirectories key in #{dot_formatter}")
     end
 
-    map =
-      for input <- List.wrap(formatter_opts[:inputs]),
-          file <- Path.wildcard(Path.join(prefix ++ [input]), match_dot: true),
-          do: {file, formatter_opts},
-          into: %{}
+    {included_entries, excluded_entries} = get_entry_types(formatter_opts)
 
-    Enum.reduce(subs, Map.merge(acc, map), fn {sub, formatter_opts_and_subs}, acc ->
+    included_map = expand_set_of_inputs(prefix, formatter_opts, included_entries)
+    excluded_map = expand_set_of_inputs(prefix, formatter_opts, excluded_entries)
+
+    filtered_map =
+      included_map
+      |> Map.drop(Map.keys(excluded_map))
+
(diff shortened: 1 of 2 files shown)


whatyouhide:
I am not sure if we want to allow this yet, but if we do, I would much rather have a separate `:excluded_inputs` key or something like that rather than a less clear `!`.

cybrox:
Out of curiosity, what would be the reason to not allow such functionality?

I did consider moving it into a separate key but this way seemed cleaner to me, especially since most other configs in my project, such as the `.gitignore` use that format for selecting files. Changing it to using a separate key wouldn't be much of an issue, though.

whatyouhide:
I think the biggest benefit would be clarity, plus ease of programmability. But mostly clarity. `.gitignore` is kind of limited to use that format, so I am not sure it's the best example.

josevalim:
I am not a big fan of behaviour that "undoes" a previous behaviour.

Plus nowhere in Elixir we use the `!` convention to refute wildcards, so I agree with @whatyouhide, a better option is clearer and more explicit. `.gitignore` is a textual format, so it makes sense there, here we are talking about a configuration file in Elixir, we should leverage more semantic information since we can.

Finally, note that .formatter.exs is a script, so if you really want this, you can do:

```elixir
inputs: Path.wildcard("lib/**/*.{ex,exs}") -- Path.wildcard("lib/static.{ex,exs}")
```

There is no reason to push this into Elixir itself.

cybrox:
I am not a huge fan of too much scripting in configurations, if an easier solution would be available. I have to agree in this case, though, that it definitely is a nice solution.

Personally, I think this possibility should at least be noted in the docs, though, since a user that is unaware of how path resolving is handled internally (or hasn't worked with `Path` much) would have a hard time coming up with that solution. Would a PR for adding that information to the docs be welcome? 🙂

josevalim:
So I pushed a small commit that mentions `Path.wildcard/1`. Hopefully that will point people to the proper direction.

PR #7720, opened 2018-06-01
Change `:type` to `:test_type` and properly document

### LINKED ISSUE #7719: ExUnit internals leak through the setup block
### Environment
Elixir 1.6.4

### Current behavior

ExUnit's test type leaks through the setup block:

```elixir
defmodule ExUnitThingTest do
  use ExUnit.Case, async: false

  setup do
    {:ok, type: "thing"}
  end

  test "some test" do
    assert true
  end
end
```

```bash
$ mix test
.

Finished in 0.02 seconds
1 thing, 0 failures

Randomized with seed 563574
```

### Expected behavior

```bash
$ mix test
.

Finished in 0.02 seconds
1 test, 0 failures

Randomized with seed 563574
```

### Notes

AFAIK the documentation doesn't mention the `type` parameter, so it should either be mentioned or the `type` should be replaced with something more explicit like `ex_unit_type`. It took my a while to figure out why my test output was this (part of the issue is surely a big `setup` block wrote by someone else):

```
........................

Finished in 0.1 seconds
15 tests, 9 statss, 0 failures

Randomized with seed 27958
```

devonestes:
Resolves #7719

### DIFF
--- a/lib/ex_unit/lib/ex_unit/callbacks.ex
+++ b/lib/ex_unit/lib/ex_unit/callbacks.ex
@@ -399,7 +399,7 @@ defmodule ExUnit.Callbacks do
 
   ## Helpers
 
-  @reserved [:case, :file, :line, :test, :async, :registered, :describe]
+  @reserved [:case, :file, :line, :test, :type, :async, :registered, :describe]
 
   @doc false
   def __callback__(callback, describe) do
--- a/lib/ex_unit/lib/ex_unit/callbacks.ex
+++ b/lib/ex_unit/lib/ex_unit/callbacks.ex
@@ -399,7 +399,7 @@ defmodule ExUnit.Callbacks do
 
   ## Helpers
 
-  @reserved [:case, :file, :line, :test, :type, :async, :registered, :describe]
+  @reserved [:case, :file, :line, :test, :async, :registered, :describe]
 
   @doc false
   def __callback__(callback, describe) do
--- a/lib/ex_unit/lib/ex_unit/case.ex
+++ b/lib/ex_unit/lib/ex_unit/case.ex
@@ -143,7 +143,6 @@ defmodule ExUnit.Case do
     * `:file`       - the file on which the test was defined
     * `:line`       - the line on which the test was defined
     * `:test`       - the test name
-    * `:type`       - the type of the test (`:test`, `:doctest`, `:property`, etc)
     * `:async`      - if the test case is in async mode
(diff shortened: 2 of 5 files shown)


devonestes:
I don't know why, but I'm seeing strange intermittent test failures in CI. I can't replicate them on my local machine, even with the same seed. Here are the examples: https://travis-ci.org/elixir-lang/elixir/builds/386680141
https://travis-ci.org/elixir-lang/elixir/builds/386687681

Think this is something to look into, or just weirdness on Travis we can ignore?

josevalim:
@devonestes it is a file system issue (it can't find a beam file) that we can't replicate either and it has been happening for a good while.

josevalim:
@whatyouhide @devonestes I am wondering if we should reserve this key or if we should namespace it `ex_unit_type` or `test_type` before reserving.

whatyouhide:
@josevalim very good point, type is a common word and we dont really need to reserve it. test_type might be useful to devs too so maybe ex_unit_type, even if I'm not the biggest fan of this being the only kw prefixed with "ex_unit_".

josevalim:
Ok after looking at the code, the type is not really meant to be reserved. We want to allow users to change it. That said, I think then `test_type` is much clearer in intent.

Let's change `:type` to `:test_type` and let's make sure to document that it is meant to be changed by users who want to segregate tests.

What do you think?

devonestes:
Sounds good to me!!

josevalim:
Alright, so let's go in this direction then!

adkron:
I could be way off, but since this is a map could we put all of the reserved keys under another key?

~~~elixir
  %{exunit: %{type: foo, ...}}
~~~

That way we don't have to restrict users as much. Thanks for your consideration.

josevalim:
@adkron the problem is backwards compatibility, I am afraid. `type` has just been released and it is considered experimental (as it is meant to support ptesting), so we can still change it, but I am afraid we just can't namespace everything else. :(

whatyouhide:
I was afraid of that, makes sense. We'll live with this :)

adkron:
Should the new ones be namespaced? Then move the others to the namespace in Elixir 2, if that day ever comes? Or is there a way to add a depreciation?

josevalim:
@adkron we can't deprecate map fields, unfortunately. I would prefer to go ahead with `test_type` for now and then discuss a possible migration altogether.

josevalim:

PR #7724, opened 2018-06-03
[Typespec#spec_to_quoted] ignore_vars should be converted to elixir variables

Sihui:
Refs https://github.com/elixir-lang/elixir/issues/7712

When building the `when` clause `Typespec#spec_to_quoted`, we want to ignore var:

```elixir
    ignore_vars = Keyword.keys(guards)

    vars =
      for type_expr <- args ++ [result],
          var <- collect_vars(type_expr),
          var not in ignore_vars,
          uniq: true,
          do: {var, {:var, meta, nil}}
```

But `ignore_vars` contains a list of codelists where as `collect_vars` returns an elixir variable:

```elixir
  defp collect_vars({:var, _line, var}) do
    [erl_to_ex_var(var)]
  end
```
As a result, this filter (`var not in ignore_vars,`) doesn't work as intended.

Converting ignore_vars to a list of elixir variables will fix it.
`ignore_vars = Keyword.keys(guards) |> Enum.map(&erl_to_ex_var/1)`

Before:
image

After:
image

====
Updates after @michalmuskala 's feedback.

Now the variable names inside the when clause are lower case for the sake of consistency.
image

### DIFF
--- a/lib/iex/test/iex/helpers_test.exs
+++ b/lib/iex/test/iex/helpers_test.exs
@@ -339,7 +339,7 @@ defmodule IEx.HelpersTest do
       assert capture_io(fn -> h(:timer.sleep() / 1) end) == """
              * :timer.sleep/1
 
-               @spec sleep(time) :: :ok when Time: timeout()
+               @spec sleep(time) :: :ok when time: timeout()
 
              Documentation is not available for non-Elixir modules. Showing only specs.
              """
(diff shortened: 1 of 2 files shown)


lib/iex/test/iex/helpers_test.exs
```diff
@@ -339,7 +339,7 @@ defmodule IEx.HelpersTest do
       assert capture_io(fn -> h(:timer.sleep() / 1) end) == """
              * :timer.sleep/1
 
-               @spec sleep(time) :: :ok when Time: timeout(), time: var
+               @spec sleep(time) :: :ok when Time: timeout()
```
michalmuskala: I still don't think this is correct. The function refers to a variable `time` while the guard specifies `Time`. I think they should be the same in both places.

Sihui: so you think it should be `@spec sleep(time) :: :ok when time: timeout()` ?

michalmuskala: Yes, I think so. If we're going for elixir syntax everywhere, let's go all the way.

On the other hand, I think printing everything in erlang syntax could be a reasonable choice too, but I think it would be slightly harder and less useful for newcomers.

Sihui: 👍 good call. I agree. Will push the change in a few mintues.

PR #7725, opened 2018-06-03
raise if piping into bitstring operator <<>>

### LINKED ISSUE #7721: Raise if piping into <<...>>
To avoid confusion when piping into interpolated strings or literal binaries.

Tuxified:
Fixes #7721: 

`1 |> <<1,2,3>>` was valid syntax as it was compiled to
<<1, 1, 2, 3>>. `<<>>` is an operator, the pipe operator (`|>`) prepended an
extra argument. Same goes for situations with interpolation since that leads to
creation of a new string, using `<<>>` and `::`.


josevalim:

PR #7731, opened 2018-06-06
Expands the documentation of Application

fxn:
Yo! This patch expands a bit the documentation of `Application` with things I missed when researching applications in the last months.

I tried to find a balance between what I'd like to know as an Elixir programmer, and the gory details. For example, I believe it is important to have a section about the resource file, and I believe it is good to explain the concept of _loading_ an application. That way you have good context to understand the environment, for example, and to describe what happens when an application that does not have a callback module is started. But we do not talk for example about the application controller, the application master, the process leader, etc.

I took some micro-decisions in writing this so take it as a proposal, as always, please feel free to suggest any changes you may like 😄.


lib/elixir/lib/application.ex
```diff
@@ -4,32 +4,90 @@ defmodule Application do
...
+  otherwise, it continues.
 
-  Starting an application is done via the "application module callback", which
-  is a module that defines the `start/2` function. The `start/2` function should
-  then start a supervisor, which is often called as the top-level supervisor, since
-  it sits at the root of a potentially long supervision tree. When the system is
+  An application callback module is a module that defines at least the callbacks
+  `start/2` and `stop/1`. If there is one configured, the start process invokes
+  its `start/2` function, which must start and link a supervisor, and return its
+  pid. That supervisor is often called _the top-level supervisor_, since it sits
```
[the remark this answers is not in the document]
fxn: I write PID everywhere except in Erlang/Elixir, where I see "pid" is more common in docs and books.

lib/elixir/lib/application.ex
```diff
@@ -4,32 +4,90 @@ defmodule Application do
...
-  ## Start and shutdown
+  The key `mod` of the application resource file specifies whether the
+  application has a callback module. If it doesn't, the start process is done,
+  otherwise, it continues.
 
-  Starting an application is done via the "application module callback", which
-  is a module that defines the `start/2` function. The `start/2` function should
-  then start a supervisor, which is often called as the top-level supervisor, since
-  it sits at the root of a potentially long supervision tree. When the system is
+  An application callback module is a module that defines at least the callbacks
```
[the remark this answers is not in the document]
fxn: Makes sense. There is a section next, let me see if I can merge them or use different headers or something.

lib/elixir/lib/application.ex
```diff
@@ -4,32 +4,90 @@ defmodule Application do
...
+  flags passed to `erl`.
+
+  Loading an application _does not_ load its modules.
+
+  In practice, you rarely load applications by hand because that is part of the
+  start process, explained next.
+
+  ## Starting and stopping applications
+
+  Applications are also _started_, and _stopped_:
```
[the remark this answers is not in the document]
fxn: Done.

lib/elixir/lib/application.ex
```diff
@@ -4,32 +4,90 @@ defmodule Application do
...
+  Applications are _loaded_, which means that the BEAM finds and processes their
+  resource file:
+
+      iex(1)> Application.load(:ex_unit)
+      :ok
+
+  If an application has included applications, they are loaded too. And the
+  procedure recurses if they in turn have included applications. (Included
+  applications are unrelated to applications in Mix umbrella projects, they are
+  an Erlang concept that has to do with coordinated starts.)
```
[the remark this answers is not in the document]
fxn: Removed.

lib/elixir/lib/application.ex
```diff
@@ -4,32 +4,90 @@ defmodule Application do
...
+  OTP application `ex_unit` is called `ex_unit.app`.
+
+  The application resource file must reside in the same `ebin` directory as the
+  compiled modules of the application. In Elixir, the Mix build tool automates
+  its generation. Some of its keys are taken from the keyword lists returned by
+  the `project/0` and `application/0` functions defined in `mix.exs`, and others
+  are generated by Mix itself.
+
+  You can learn more about the generation of application resource files by
+  typing `mix help compile.app`.
```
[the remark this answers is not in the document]
fxn: Good point, let me generate and see which option autolinks.

lib/elixir/lib/application.ex
```diff
@@ -4,32 +4,90 @@ defmodule Application do
...
-  `APP` is the application name, usually in `underscore_case`. The application
-  file must reside in the same `ebin` directory as the compiled modules of the
-  application. In Elixir, the Mix build tool is responsible for compiling your
-  source code and generating your application `.app` file. You can learn more
-  about the generation of `.app` files by typing `mix help compile.app`.
+  ## The application resource file
+
+  Applications are specified in their [_resource
+  file_](http://erlang.org/doc/man/app.html), which is a file called
+  _application_name_`.app`. For example, the application resource file of the
```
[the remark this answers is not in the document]
fxn: Done!

lib/elixir/lib/application.ex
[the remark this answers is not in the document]
fxn: No luck with the autolinks... what about linking with a relative path so it works both in edge and stable docs? For example, this seems to be OK according to my tests:

> You can learn more about the generation of application resource files in the [`mix compile.app`](../mix/Mix.Tasks.Compile.App.html) task.

lib/elixir/lib/application.ex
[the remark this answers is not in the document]
fxn: Oh man I just saw https://github.com/eksperimental/elixir_inconsistencies, that is golden!

lib/elixir/lib/application.ex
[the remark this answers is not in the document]
whatyouhide: Sounds good.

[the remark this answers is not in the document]
fxn: Awesome, it is up.

fxn:
The original docs have this sentence that the initial patch preserved:

> The application resource file must reside in the same `ebin` directory as the compiled modules of the application.

I just pushed an alternative to that sentence to relax a bit the "must":

> You'll find the resource file of an application in its `ebin` directory, it is generated automatically by Mix.

Reason is, I read the source code of `application_controller` back in the day, and [it just searches the code path](https://github.com/erlang/otp/blob/9ae2044073e6433030ce30756658b103ce67c3c1/lib/kernel/src/application_controller.erl#L1446). For example, this works:

```
$ cat foo.app
{application, foo, []}.
$ iex
iex(1)> Application.load(:foo)
:ok
```

and there is no `ebin`. In practice, when you ship an application using the standard structure, `ebin` is the conventional place in the code path where you can throw the resource file, but formally I am not sure it is a "must". This is just an intro, not an OTP reference, so that relaxed sentence may be just fine. Sounds good?

lib/elixir/lib/application.ex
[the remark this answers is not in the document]
fxn: Yep, I think I need to work more on this patch and merge some information. I'll write back!

lib/elixir/lib/application.ex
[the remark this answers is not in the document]
eksperimental: We use PID in documenration since its an acronym, but we use pid for arguments and function names

Refs
https://github.com/elixir-lang/elixir/pull/4883
https://github.com/elixir-lang/elixir/pull/4754

lib/elixir/lib/application.ex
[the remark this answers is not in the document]
eksperimental: IIRC we dont use underscores but asterisks in markdown to make text bold and italics

whatyouhide:
@fxn yes sounds good.

lib/elixir/lib/application.ex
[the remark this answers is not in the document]
whatyouhide: We do.

lib/elixir/lib/application.ex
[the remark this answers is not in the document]
fxn: Awesome, I've revised it.

lib/elixir/lib/application.ex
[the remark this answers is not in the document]
fxn: Updated too!

lib/elixir/lib/application.ex
```diff
@@ -4,32 +4,89 @@ defmodule Application do
...
+  where `APP` is the application name. For example, the application resource
+  file of the OTP application `ex_unit` is called `ex_unit.app`.
+
+  You'll find the resource file of an application in its `ebin` directory, it is
+  generated automatically by Mix. Some of its keys are taken from the keyword
+  lists returned by the `project/0` and `application/0` functions defined in
+  `mix.exs`, and others are generated by Mix itself.
+
+  You can learn more about the generation of application resource files in the
+  [`mix compile.app`](../mix/Mix.Tasks.Compile.App.html) task.
```
eksperimental: Im wondering how is this going to look like when read in IEx.
This is something I would like to propose/implement in exdoc. The ability.to link to other Elixir proyects

lib/elixir/lib/application.ex
```diff
@@ -4,32 +4,89 @@ defmodule Application do
...
+  Note that this environment is unrelated to the operating system environment
+  variables.
+
+  The application environment can be overriden via the `-config` option of
+  `erl`, as well as command-line flags, as we are going to see below.
+
+  ## Loading applications
+
+  Applications are *loaded*, which means that the BEAM finds and processes their
+  resource file:
```
eksperimental: I think it should be "their resource files" (plural), since they are more.than one.

lib/elixir/lib/application.ex
```diff
@@ -4,32 +4,89 @@ defmodule Application do
...
+
+      iex(1)> Application.load(:ex_unit)
+      :ok
+
+  If an application has included applications, they are loaded too. And the
+  procedure recurses if they in turn have included applications. Included
+  applications are unrelated to applications in Mix umbrella projects, they are
+  an Erlang concept that has to do with coordinated starts.
```
eksperimental: I would say "Erlang/OTP concept" or "OTP concept"

lib/elixir/lib/application.ex
```diff
@@ -4,32 +4,89 @@ defmodule Application do
...
+
+  When an application is loaded, the environment specified in its resource file
+  is merged with any overrides from config files passed to `erl` via the
+  `-config` option. It is worth highlighting that releases pass `sys.config`
+  this way. In turn, these are merged again with any specific `-Application`
```
eksperimental: These what? These options, args, ... I think its worth mention it explicitly for clarity

lib/elixir/lib/application.ex
```diff
@@ -4,32 +4,89 @@ defmodule Application do
...
   system. Instead, you start one or more applications, each with their own
   initialization and termination logic.
 
-  Applications also provide an "application environment", which provides one
-  mechanism for configuring long running applications. We will learn more about
-  the tooling, start and shutdown and the application environment in the next
-  sections.
+  When an application is started, the BEAM loads it if it is not already loaded
+  (in the technical sense described above). Then, it checks if the dependencies
+  listed in the `applications` key of the resource file are already started.
```
eksperimental: I think application should be an atom  since its a key

lib/elixir/lib/application.ex
```diff
@@ -4,32 +4,89 @@ defmodule Application do
...
+  Having at least one dependency not started is an error condition.
+
+  When you start an application with `mix run`, Mix takes care of starting all
+  dependencies for you.
```
eksperimental: "starting all the dependencies"

lib/elixir/lib/application.ex
```diff
@@ -4,32 +4,89 @@ defmodule Application do
...
 
-  ## Start and shutdown
+  The key `mod` of the application resource file specifies whether the
```
eksperimental: Mod shouldn't be an atom since its a key?

lib/elixir/lib/application.ex
fertapric: I thought about commenting the same, but if you open the application resource file `APP.app` what you find is `applications`, not `:applications`. So I thought the current documentation was more accurate in the sense of "what you will see if you open the file".

fxn: That was also my rationale, because it is the ... key *of the resource file*.

lib/elixir/lib/application.ex
```diff
@@ -4,32 +4,89 @@ defmodule Application do
...
+  generated automatically by Mix. Some of its keys are taken from the keyword
+  lists returned by the `project/0` and `application/0` functions defined in
+  `mix.exs`, and others are generated by Mix itself.
+
+  You can learn more about the generation of application resource files in the
+  [`mix compile.app`](../mix/Mix.Tasks.Compile.App.html) task.
+
+  ## The application environment
+
+  The key `env` of an application resource file has a list of tuples that map
```
fertapric: If, at the end, the keys of the application resource file are written as Elixir atoms, then this should also be changed.

lib/elixir/lib/application.ex
eksperimental: Yes. It should stay as is then. Sorry i didnt have a computer at hand to check the files when reviewing

lib/elixir/lib/application.ex
josevalim: :+1: for Erlang/OTP. Erlang by itself is the language and the language does not know about applications.

lib/elixir/lib/application.ex
josevalim: I think we should say:

> files in the `Mix.Tasks.Compile.App` task (or also by running `mix help compile.app`).

fxn: Do you prefer one of the two options?

Linking to `Mix.Tasks.Compile.App`, even if manually, is normal: you link to a module's documentation.

On the other hand, I wonder if mentioning the module is natural or not for an end-user, compared to what they normally interact with in the command-line. (I wonder in the sense that I do not know the answer.) In that option, linking `mix help compile.app`... is a bit strange because you are linking a command to the source code of a module. Not a big mismatch, but just someting to take into account.

What do you think?

(BTW, I need to work in the general structure of the module doc and merge some sections, the patch as is needs more work but all this feedback is being incorporated to the second pass.)

josevalim: If we use `Mix.Tasks.Compile.App`, then we will get automatic linking in places like ExDoc, and pointing out to `mix help compile.app` in parens will be helpful for anybody in the command line.

Given that both are public and the relationship between them is public, I think going with both is fine and the most user friendly. Does it make sense?

fxn: Sounds good, @whatyouhide good for you too?

BTW, locally I do not get autolinking from Elixir to Mix, the purpose of the relative path is to link manually in a way that works both in edge and stable docs. Am I missing something?

josevalim: If autolinking to mix modules is not working then it should be treated as an ex_doc bug. We should investigate. :)

eksperimental: Its a missing feature. As far as i know it has never been implemented

fxn: Could work on that too if you will. In that case, could we perhaps leave here just fixed-width font by now?

lib/elixir/lib/application.ex
```diff
@@ -2,61 +2,88 @@ defmodule Application do
...
+  callback module and start argument:
 
       def application do
         [mod: {MyApp, []}]
       end
 
-  Our application now requires the `MyApp` module to provide an application
-  callback. This can be done by invoking `use Application` in that module and
-  defining a `start/2` callback, for example:
+  This key is optional, only needed for applications implemented as a
```
fertapric: I would rephrase this :) 

This sentence seems to suggest that the core of the application is the supervision tree, but that might not be accurate. Maybe use a expression like "applications with a supervision tree" or "applications that have/use a supervision tree"?

fxn: I feel also so-so about that sentence, but there are several things to consider here.

On one hand, a "library" application may launch processes and supervisors.

The other thing is that I copied this sentence [from the Erlang documentation](http://erlang.org/doc/man/app.html):

> Key mod is necessary for an application implemented as a supervision tree, otherwise the application controller does not know how to start it. mod can be omitted for applications without processes, typically code libraries, for example, STDLIB.

While it seems to be a bit forced, I am not an Erlang programmer, so I guessed that way to phrase it would probably be idiomatic, giving the source.

But if we can find something better...

lib/elixir/lib/application.ex
```diff
@@ -93,11 +117,90 @@ defmodule Application do
...
+      #=> :ok
+
+  Once your application is compiled, running your system is a matter of starting
+  your current application and its dependencies. Differently from other languages,
+  Elixir does not have a `main` procedure that is responsible for starting your
+  system. Instead, you start one or more applications, each with their own
+  initialization and termination logic.
+
+  When an application is started, the runtime loads it if it is not already
+  loaded (in the technical sense described above). Then, it checks if the
```
eksperimental: "the runtime loads it if it is hasn't been loaded yet"

lib/elixir/lib/application.ex
```diff
@@ -93,11 +117,90 @@ defmodule Application do
...
+
+  1. If present, invoke the optional callback `c:prep_stop/1`.
+  2. Terminate the top-level supervisor.
+  3. Invoke the required callback `c:stop/1`.
+
+  The arguments passed to the callbacks are related to the state optionally
+  returned by `c:start/2`, and are documented in the section about the callback
+  module above.
+
+  It is important to highlight that step 2 is blocking. Termination of a
```
eksperimental: "that step 2 is a blocking one"

lib/elixir/lib/application.ex
```diff
@@ -93,11 +117,90 @@ defmodule Application do
...
+  supervisor triggers a recursive chain of children terminations, therefore
+  orderly shutting down all descendant processes. The `c:stop/1` callback is
+  invoked only after termination of the whole supervision tree.
+
+  Shutting down a live system cleanly can be done by calling `System.stop/1`.
+  It will shut down all applications in the opposite order they are started.
```
eksperimental: "It will shut down every application in the opposite order they had been started."

lib/elixir/lib/application.ex
```diff
@@ -118,41 +221,14 @@ defmodule Application do
...
-  then those configurations are better placed at runtime. For example,
-  one may configure applications dynamically via `put_env/3`.
+  ## Further information
 
-  Keep in mind that each application is responsible for its environment.
-  Do not use the functions in this module for directly accessing or modifying
-  the environment of other applications (as it may lead to inconsistent
-  data in the application environment).
+  For further details on applications please check the documentation of the
+  [`application`](http://www.erlang.org/doc/man/application.html) Erlang/OTP
```
eksperimental: I would say here "Erlang module"

fxn:
@eksperimental all feedback addressed, thanks!

josevalim:

PR #7732, opened 2018-06-06
Add support for "--organization" in the archive.install task

whatyouhide:
For archives downloaded from Hex, we want to be able to pass in an organization:

    mix archive.install hex my_package --organization my_org

Do we want to support `--repo` as well?

PS: did this pairing up with @idlehands, thanks for the help Jeffrey! 💟

### DIFF
--- a/lib/mix/lib/mix/local/installer.ex
+++ b/lib/mix/lib/mix/local/installer.ex
@@ -216,7 +216,7 @@ defmodule Mix.Local.Installer do
     dep_opts =
       opts
       |> Keyword.take([:organization])
-      |> Keyword.merge(hex: String.to_atom(package_name))
+      |> Keyword.put(:hex, String.to_atom(package_name))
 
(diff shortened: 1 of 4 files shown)


lib/mix/lib/mix/local/installer.ex
```diff
@@ -213,7 +213,12 @@ defmodule Mix.Local.Installer do
         package_name
       end
 
-    {:fetcher, {String.to_atom(app_name), version, hex: String.to_atom(package_name)}}
+    dep_opts =
+      opts
+      |> Keyword.take([:organization])
+      |> Keyword.merge(hex: String.to_atom(package_name))
```
josevalim: Can we use Keyword.put here instead?

whatyouhide: Yes, this was brought over from moving stuff around. Pushed just now.

Issue #7718, opened 2018-05-30
Function with multi-clause style collapses awkwardly with single clause


chrismccord:
### Environment

* Elixir & Erlang/OTP versions (elixir --version): 1.6.0
* Operating system: OSX 10.13.4

### Current behavior

I've found functions with multi-clause style collapses awkwardly when a single clause is used. Given the following code block:

```elixir
SomeModule.long_function_name_that_approaches_max_columns(argument, acc, fn
  %SomeStruct{key: key}, acc -> more_code(key, acc)
end)
```

The formatter produces:

```elixir
SomeModule.long_function_name_that_approaches_max_columns(argument, acc, fn %SomeStruct{
                                                                              key: key
                                                                            },
                                                                            acc ->
  more_code(key, acc)
end)
```

### Expected behavior

The formatter does not change the original code. I realize using the multi-clause style could be "cheating" in this case, but if you add a second clause, the formatter does not change the code, and the formatted code in this case is very awkwardly styled. Thanks for taking a look!

josevalim:
This should be a straight-forward change. Either we are only checking for newlines if you have more than two clauses or the newline checking is only between clauses, which means it won't work here.

Sihui:
@chrismccord 
>  if you add a second clause, the formatter does not change the code

Can you provide an example? 

For my formatted result is a bit different than yours (but still awkard)

```elixir
# Ex 1
# Original
SomeModule.long_function_name_that_approaches_max_columns(argument, acc, fn
   %SomeStruct{key: key}, acc -> more_code(key, acc)
 end)
# Formatted
SomeModule.long_function_name_that_approaches_max_columns(argument, acc, fn %SomeStruct{key: key},
                                                                            acc ->
  more_code(key, acc)
end)

# Ex 2
# Original
SomeModule.long_function_name_that_approaches_max_columns(argument, acc, fn
  %SomeStruct{key: key}, %SomeStruct{key: key}, acc -> more_code(key, acc)
end)
# Formatted
SomeModule.long_function_name_that_approaches_max_columns(argument, acc, fn %SomeStruct{key: key},
                                                                            %SomeStruct{key: key},
                                                                            acc ->
  more_code(key, acc)
end)
```

josevalim:
@Sihui I believe he meant:

```
SomeModule.long_function_name_that_approaches_max_columns(argument, acc, fn
   %SomeStruct{key: key}, acc -> more_code(key, acc)
[...]

PR #7737, opened 2018-06-10
[Formatter : anon_fun] When fn is followed by a newline, use multi-clause formatting style

Sihui:
ref: https://github.com/elixir-lang/elixir/issues/7718

### DIFF
--- a/lib/elixir/src/elixir_parser.yrl
+++ b/lib/elixir/src/elixir_parser.yrl
@@ -649,18 +649,6 @@ meta_from_location({Line, Column, _}) ->
     false -> [{line, Line}]
   end.
 
-meta_from_token_keep_eol(Token) ->
-  meta_from_location_keep_eol(?location(Token)).
-
-meta_from_location_keep_eol({Line, Column, eol}) ->
-  case ?columns() of
-    true -> [{eol, true}, {line, Line}, {column, Column}];
-    false -> [{eol, true}, {line, Line}]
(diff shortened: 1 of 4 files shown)


lib/elixir/src/elixir_parser.yrl
```diff
@@ -659,9 +671,9 @@ meta_from_token_with_end_line(Begin, End) ->
   case ?formatter_metadata() of
     true ->
       [{end_line, line_from_location(?location(End))}
-       | meta_from_token(Begin)];
+       | meta_from_token_keep_eol(Begin)];
     false ->
-      meta_from_token(Begin)
+      meta_from_token_keep_eol(Begin)
```
josevalim: Instead of changing this function to also handle `eol`, what if we change this line:

https://github.com/Sihui/elixir/blob/d9252a8872667e2fd8b77dd78e50667761838c4f/lib/elixir/src/elixir_parser.yrl#L849

To do something like `eol_pair(Fn, End) ++ meta_from_token(Fn)` instead of `meta_from_token_with_end_line(Fn, End)`? I think we will end-up with less code after all.

Btw, great solution to this problem!

Sihui: @josevalim, good call! I agree. Using `meta_from_token_keep_eol` is not very optimal. Updated the PR accordingly :D

josevalim:

PR #7738, opened 2018-06-10
[Formatter : anon_fun] When fn is followed by a newline, preserve the newline.

Sihui:
ref: https://github.com/elixir-lang/elixir/issues/7718

### DIFF
--- a/lib/elixir/src/elixir_parser.yrl
+++ b/lib/elixir/src/elixir_parser.yrl
@@ -305,7 +305,7 @@ eoe -> ';' : '$1'.
 eoe -> eol ';' : '$1'.
 
 fn_eoe -> 'fn' : '$1'.
-fn_eoe -> 'fn' eoe : '$1'.
+fn_eoe -> 'fn' eoe : next_is_eol('$1').
 
 do_eoe -> 'do' : '$1'.
 do_eoe -> 'do' eoe : '$1'.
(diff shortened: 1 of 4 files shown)


lib/elixir/test/elixir/code_formatter/general_test.exs
```diff
@@ -187,6 +187,33 @@ defmodule Code.Formatter.GeneralTest do
...
+      bad = ~S"""
+      SomeModule.long_function_name_that_approaches_max_columns(argument, acc, fn 
+        %SomeStruct{key: key}, acc -> more_code(key, acc)
+      end)
+      """
+
+      assert_format bad, ~S"""
+      SomeModule.long_function_name_that_approaches_max_columns(argument, acc, fn 
+        %SomeStruct{key: key}, acc ->
+        more_code(key, acc)
```
josevalim: This part should be indented with four spaces.

lib/elixir/src/elixir_parser.yrl
```diff
@@ -659,9 +671,9 @@ meta_from_token_with_end_line(Begin, End) ->
   case ?formatter_metadata() of
     true ->
       [{end_line, line_from_location(?location(End))}
-       | meta_from_token(Begin)];
+       | meta_from_token_keep_eol(Begin)];
     false ->
-      meta_from_token(Begin)
+      meta_from_token_keep_eol(Begin)
```
josevalim: Instead of changing this function to also handle `eol`, what if we change this line:

https://github.com/Sihui/elixir/blob/d9252a8872667e2fd8b77dd78e50667761838c4f/lib/elixir/src/elixir_parser.yrl#L849

To do something like `eol_pair(Fn, End) ++ meta_from_token(Fn)` instead of `meta_from_token_with_end_line(Fn, End)`?

Btw, great solution to this problem!

josevalim:
Oh, I see the other pull request, that is definitely the better way to go. Great job! I will close this one and transfer the relevant comments from here there!

PR #7739, opened 2018-06-10
some notes on a distributed logger

pragdave:
I thought that rather than rush to code I'd approach this in steps. So here's an outline of the change I'm suggesting.

Dave

### DIFF
--- a/lib/logger/notes.md
+++ b/lib/logger/notes.md
@@ -1,0 +1,42 @@
+## Distributed Logger Proposal
+
+### Issue Addressed
+
+As Elixir apps are deployed across dozens of nodes, it becomes harder to
+manage and inspect all the independent log files. Some, like Bleacher
+Report, resort to [third party solutions](https://dev.bleacherreport.com/elixir-phoenix-centralized-http-logging-aa50efe3105b).
+
+This seems unnecessary, as this kind of distributed solution is square
+in Elixir's sweet spot.
+
+### Proposal
+
+1. Alter the `console` logger backend to accept a new configuration
+   parameter, `known_as`. This would publish the a name for this logger.
+   The name could be an atom, is which case `global` is used, or it
+   could be a `:via` tuple.
+
+2. Create a new logger backend, `forward`. This will be configured to
+   take the name of the logger created in (1), along with
+
+   * an optional identifying string (defaults to the node name)
+   * a minimum log level to send (defaults to `error`). This is
+     distinct from the logger's own error level, and controls the
+     amount of traffic sent over the network.
+
+### In Use
+
+1. The `forward` backend will typically be configured in parallel with
+   an existing backend (such as console). That way the logger on a
+   client machine will continue to log locally even if the logging
+   server goes down. (maybe: If the server does go down, the forward backend
+   will enter an exponential backoff poll for it to come back.)
+
+2. We'd recommend that folks run the logging server in a separate
+   top-level application, isolating it from failures of application
+   level components.
+
+3. On the logging server, messages received via `forward` from another
+   node will be logged normally, and hence will follow the logging
+   server's log configuration. (Note that this even permits a hierarchy
+   of loggers...)


josevalim:
Thanks @pragdave!

I am thinking we don't need to change the existing backends nor add a new backend.

Instead, we could change the part in `Logger.bare_log` to use a custom name and send it directly to the remote node. This means we would bypass all handlers in the current node. What do you think?

lib/logger/notes.md
```diff
@@ -0,0 +1,42 @@
...
+
+This seems unnecessary, as this kind of distributed solution is square
+in Elixir's sweet spot.
+
+### Proposal
+
+1. Alter the `console` logger backend to accept a new configuration
+   parameter, `known_as`. This would publish the a name for this logger.
+   The name could be an atom, is which case `global` is used, or it
+   could be a `:via` tuple.
```
josevalim: There is a mismatch here. The logger backends are not named, only the Logger `gen_event`. So we would have to name the whole logger backend.

My suggestion is to call this option `:name` or something that ends with `_name` and keep the same naming arguments as a GenServer for consistency. So `name: Logger` would be the default, `name: {:global, ...}` or via would be supported too.

josevalim:
Ok, I see now that you want to have two handlers, so we continue to log locally. If that's the case, I don't think we need to introduce `:known_as`. Since the goal here is to leverage the Erlang distribution, in the Erlang distribution we can access any named process in another by using the `{name, node}`. This means we can forward to any node by doing something like `:gen_event.notify({Logger, other_node_name}, event)`.

pragdave:
Bypassing the handlers in the client node means there's no fallback if the logger server fails. From our previous conversation, you wanted to be able to handle this.

pragdave:
This means we can forward to any node by doing something like :gen_event.notify({Logger, other_node_name}, event)

Then we couple everything to a named node, rather than a named process. Do we do that anywhere else? And is there infrastructure to allow failover of named nodes?

josevalim:
@pragdave good point. So let's consider we need `known_as/name` back.

The issue here is that, since you still want to log locally, we need both the local name and the global name. However, the `:name` option allows only a single name, so the naming change in particular can't belong to Logger gen_event manager per se.

Here is my suggestion for a v1 / proof of concept:

1. Have a handler named global_forwarder (or whatever name) that calls global:register_name/2 (or similar) on its init. It doesn't matter if you win or not.

2. For each event in the global_forwarder, you lookup the server with `global:whereis/1` with three different outcomes:

  a. If it exists and it is `!= self()`, broadcast to it
  b. If it exists and it is `== self()`, ignore it
  c. If it does not exist, discard the message but attempt to register yourself

`2.c` can be an issue. Global registration can take a while and you dpn't want to block the logger. So you can always try to do the registration in a separate process. The other thing is that you said you want only certain nodes to actually be the global logger, so you don't want the handler to always do it. That may further justify the need to move to a separate process.

Thoughts?

michalmuskala:
Thinking of this, my main worry is performance of such a solution. The erlang distribution is just a single TCP socket between the nodes and forwarding all the log messages one-by-one over that might saturate it overloading the distribution channels. Especially in more busy applications, they can log a lot of stuff. So while this is definitely possible, I'm really not sure this is the right approach for global log gathering.

My conclusion is that we should make it possible to do this, but we should not encourage this as a good default solution.

josevalim:
@michalmuskala right, we need to evaluate which changes are necessary as this should be possible at least as a separate project. We know we may need to change the console handler to no longer discard stuff from across nodes. Now we need to investigate if we may need something else.

pragdave:
Option 3:

Simply add a configuration option to `console` to say `remote: name`. If set, then the console does an add_handler to `{ :via, :global, name }`, 

Then also change the console handler to accept events from any node.

josevalim:
@pragdave I don't think it is the job of the console to register the logger under another name, because it won't do anything with this information. Also, as I mentioned above, I don't think only registering it is enough, because you may want to perform certain actions when the node is down (such as re-registering), and we don't want to hide this logic inside the console backend.

I think the backend doing the forwarding also needs to do the registering (or even have another backend with the explicit goal of registering and managing the registration).

pragdave:
No, the console doesn't register. It simply does an add_handler to that name.

The remote logger registers itself.

josevalim:
Sorry @pragdave, I am afraid I don't see it. Do you want the console to add itself to a remote node? If so, what would then happen? Would the console start sending messages to the remote node or receive messages from it?

In case what I am asking doesn't make sense, can you explain precisely what you mean by "If set, then the console does an add_handler to { :via, :global, name }"?

pragdave:
Right now, the fact there;s a [:console] in backends: means that the logger adds a console and does an add_handler for it.

I'm suggesting that if that console has a `remote:` option in its config, that name is what is registered, rather than creating a local process.

josevalim:
> Right now, the fact there;s a [:console] in backends: means that the logger adds a console and does an add_handler for it.

Specifying the `:console` option performs only a single operation, it adds a handler. The `:console` backend is not a separate process. The only name that is registered in Logger is the name of the `Logger` itself, which is used across all handlers.

We may be getting stuck in implementation details here, but to me "registering the console" or "creating a local process for it" does not have any meaning, because we never do this for the console in the first place.

josevalim:
To be more precise, we can't register a name for the console nor for any other backend. We can either replace `Logger` as a whole, which means nothing is logged locally, or we can add multiple backends inside the `Logger` process and those backends may attack extra names to the logger, but we can never name each handler individually.

josevalim:
@pragdave can you give some feedback on why the issue was closed?

FWIW, I am not saying there is nothing to be changed. All I am saying is that for now I can't quite understand the changes being proposed, but that may have been caused to a difference in terminology.

Thanks!

pragdave:
I'll put together an implementation as a separate project.

In a related note: I did a quick github query. Of the 100 repositories with language=elixir, ordered by number of stars, 20% contain `require Logger`.

Perhaps it should be split out of core?

josevalim:
Breaking Logger out of core would be backwards incompatible. So it can't
happen any time before 2.0 and we don't plan to release 2.0 any time soon.
For this reason alone, we have no plans to remove any application (nor any
released feature) from core. On the positive side, we have no plans of
adding new applications either.

The other thing is that Logger aims to provide integration with Erlang's
built-in error_logger. So removing the Logger from Elixir means that all
Erlang logs, such as the ones you get when using Supervisors, GenServer,
etc will be logged in the Erlang format and in Erlang terms. This would be
undesirable and not provide a good development experience. The other issue
is that Erlang's built-in error_logger had performance issues we had to
address, although those are solved in OTP 21.

PR #7744, opened 2018-06-13
Fix type spec for Task.Supervisor.start_child/4

blatyo:
Behind the scenes, this function uses `DynamicSupervisor.start_child/4`, which can return multiple errors. This gives `Task.Supervisor.start_child/4` the return type. This allows dialyzer to not complain when you pass `:max_children` to the `Task.Supervisor` and call `start_child`, which may return `{:error, :max_children}`.


josevalim:

PR #7753, opened 2018-06-14
Refactor __info__ so it generates smaller assembly

michalmuskala:
The phash value changes because clauses for the `__info__` function
are now ordered differently

### DIFF
--- a/lib/elixir/src/elixir_erl.erl
+++ b/lib/elixir/src/elixir_erl.erl
@@ -329,8 +329,8 @@ macros_info(Defmacro) ->
   {clause, 0, [{atom, 0, macros}], [], [elixir_erl:elixir_to_erl(lists:sort(Defmacro))]}.
 
 get_module_info(Module, Key) ->
-  Call = remote(0, erlang, get_module_info, [{atom, 0, Module}, {atom, 0, Key}]),
-  {clause, 0, [{atom, 0, Key}], [], [Call]}.
+  Call = remote(0, erlang, get_module_info, [{atom, 0, Module}, {var, 0, 'Key'}]),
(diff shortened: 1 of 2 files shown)


josevalim:
For curiosity, which difference does it make?

michalmuskala:
The compiler can unify all the clauses calling `erlang:get_module` because they use a variable as argument instead of a literal atom.

josevalim:
Does it change the match semantics though? Does a function clause error become a match error?

michalmuskala:
No. We still match in the clauses. The difference is:

```erl
foo(bar) -> erlang:get_module('Module', bar)
foo(baz) -> erlang:get_module('Module', baz)
```
vs
```erl
foo(bar = Key) -> erlang:get_module('Module', Key)
foo(baz = Key) -> erlang:get_module('Module', Key)
```

The semantics are exactly the same, it's just that the body of all clauses is now the same, so it can appear just once in the final code.

lib/elixir/src/elixir_erl.erl
```diff
@@ -329,8 +329,8 @@ macros_info(Defmacro) ->
   {clause, 0, [{atom, 0, macros}], [], [elixir_erl:elixir_to_erl(lists:sort(Defmacro))]}.
 
 get_module_info(Module, Key) ->
-  Call = remote(0, erlang, get_module_info, [{atom, 0, Module}, {atom, 0, Key}]),
-  {clause, 0, [{atom, 0, Key}], [], [Call]}.
+  Call = remote(0, erlang, get_module_info, [{atom, 0, Module}, {var, 0, 'Key'}]),
```
fertapric: I think it might be a good idea to include some comments like the ones you gave on https://github.com/elixir-lang/elixir/pull/7753#issuecomment-397270877 :)

I think this change might not be so obvious for future Elixir contributors.

PR #7755, opened 2018-06-14
Fix the LFE example in Mix.Compilers.Erlang

meddle0x53:
I wanted to make it easier for Elixir developers to try and play with LFE (as I like LFE), so decided to create a mix task for compiling LFE, which gave birth to [mix_lfe](https://github.com/meddle0x53/mix_lfe).

To my surprise, I found an example for exactly that in `Mix.Compilers.Erlang`, while I was looking at Mix's sources which was almost right, but not quite. I used the idea [here](https://github.com/meddle0x53/mix_lfe/blob/6021715d5af181069eb0fd09dcd1787a020ff0de/lib/mix/compilers/lfe.ex#L26) and modified it to work, so decided to fix this example too:

* The `:outdir` fixes a syntax error in the example and is a real option of the `:lfe_comp.file/2`.
* The `:report` option will make it so that compiler errors are shown when the sources are compiled.

### DIFF
--- a/lib/mix/lib/mix/compilers/erlang.ex
+++ b/lib/mix/lib/mix/compilers/erlang.ex
@@ -22,7 +22,7 @@ defmodule Mix.Compilers.Erlang do
 
       compile manifest, [{"src", dest}], :lfe, :beam, opts, fn input, output ->
         :lfe_comp.file(to_erl_file(input),
-                       [{output_dir, Path.dirname(output)}, :return])
+                       [{:outdir, Path.dirname(output)}, :return, :report])
       end
 
   The command above will:


lib/mix/lib/mix/compilers/erlang.ex
```diff
@@ -22,7 +22,7 @@ defmodule Mix.Compilers.Erlang do
 
       compile manifest, [{"src", dest}], :lfe, :beam, opts, fn input, output ->
         :lfe_comp.file(to_erl_file(input),
-                       [{output_dir, Path.dirname(output)}, :return])
+                       [{:outdir, Path.dirname(output)}, :return, :report])
```
josevalim: If you pass :report, the errors will be reported twice, because once they are returned we printed them too, no?

meddle0x53: Not really, in the beginning I tried without `:report` and I was not seeing them at all, that's why I added the `:report` bit and started seeing them. In the moment, when I compile file with bad syntax the output is OK:

```
> mix compile
Compiling 1 file (.lfe)
src/echo.lfe:11: unknown form
src/echo.lfe:9999: unbound function: #(start 2)
```

josevalim: @meddle0x53 ideally we want to return them because then you get better tooling, such as editor support. Can you double check that we are returning warnings and errors with the `:return` flag and without the `:report` one?

meddle0x53: Also both `:return` and `:report` are passed to the `Erlang` compiler when the `compile.erlang` task is ran (that's where I saw passing the `:report` bit and tried it): https://github.com/elixir-lang/elixir/blob/1c5ed3840d608b4488477c52b549c494409e47d0/lib/mix/lib/mix/tasks/compile.erlang.ex#L79

josevalim:
You are correct, apologies. The diagnostic are returned but never printed.

josevalim:

PR #7761, opened 2018-06-18
Fix small grammar error in Integer.parse docs

FreedomBen:
... a integer -> an integer


whatyouhide:
Thanks @FreedomBen! 💟

PR #7763, opened 2018-06-19
added info on child spec formats

davidsulc:
Referring to the `Supervisor` docs enables the user to determine what the various child spec shapes relate to, and how they can be used to specify specs for dynamic children. In particular, it clarifies that the user doesn't need to provide a full child spec map, but can instead use the module or tuple forms.

### DIFF
--- a/lib/elixir/lib/dynamic_supervisor.ex
+++ b/lib/elixir/lib/dynamic_supervisor.ex
@@ -284,8 +284,9 @@ defmodule DynamicSupervisor do
   @doc """
   Dynamically adds a child specification to `supervisor` and starts that child.
 
-  `child_spec` should be a valid child specification. The child process will
-  be started as defined in the child specification.
+  `child_spec` should be a valid child specification as detailed in the
+  "child_spec/1" section of the documentation for `Supervisor`. The child
+  process will be started as defined in the child specification.
 
   If the child process start function returns `{:ok, child}` or `{:ok, child,
   info}`, then child specification and PID are added to the supervisor and


lib/elixir/lib/dynamic_supervisor.ex
```diff
@@ -284,8 +284,9 @@ defmodule DynamicSupervisor do
   @doc """
   Dynamically adds a child specification to `supervisor` and starts that child.
 
-  `child_spec` should be a valid child specification. The child process will
-  be started as defined in the child specification.
+  `child_spec` should be a valid child specification as detailed in the
+  "child_spec/1" section of the documentation for `Supervisor`. The child
+  process will be started as defined in the child specification.
```
michalmuskala: We can use `Supervisor.child_spec/1` (wrapped in backticks) and it will be auto-linked by ex_doc.

What do you think about something like this:
```
The child process will be started according to the child specification
provided in the `child_spec` argument. To learn more about child specifications
see `Supervisor.child_spec/1`.
```

davidsulc: You're obviously more knowledgeable than I am, but can you confirm the auto-linking will work for section titles? As far as I know, the linking feature works when referencing functions including the module name. However, in this case the relevant documentation is not within a function (see https://hexdocs.pm/elixir/Supervisor.html#module-child_spec-1 for the relevant section I'm talking about).

In addition, I don't see a function where the relevant documentation could be moved as `Supervisor` only defines `child_spec/2` (i.e. different arity).

Other than my concerns about the auto-linking (which wouldn't work according to my understanding, per above), I think your proposed change is great.

josevalim: @michalmuskala it seems @davidsulc wants to link to the "child_spec/1" section in the Supervisor module docs. I agree it is confusing though.

Who got the idea to name a section "child_spec/1"? 🤚haha

davidsulc: @josevalim @michalmuskala  Yes, that's what I meant to refer to. I also believe the information on child specs could be made less confusing for newer users. Would a larger documentation change/refactor be welcome?

I was thinking along the lines of concentrating information on child specs in one place, as it is currently split among the "Child specification" section, the "child_spec/1" section, and within the "start_link/2, init/2, and strategies". In my opinion, it would be helpful to have a "Child specification" section detailing what a child spec is and what the attributes correspond to (which is currently what is in the "Child specification" section), and then have some place to group information about alternative means of providing child specs (module names, tuples), and how that ends up calling the module's `child_spec/1` function (which is often defined via a `use` directive).

Thoughts?

josevalim: @davidsulc this documentation has been rewritten after v1.6 but it has not yet been released, so I would like to get more general feedback before doing more changes.

In any case, you are completely right that it was being spread over three sections, while the original intention was to have two. I have fixed this here: 2de07ba2ab7a2e9c4b68cc8aef2e431e6cd7f544

Thank you!

PR #7764, opened 2018-06-19
Restore and/or optimisation

michalmuskala:
The optimize_boolean option is very sensitive to the order of
clauses and the optimisation was not happening for and/2 and or/2.

Add regression tests to ensure that won't happen again.

---

This change breaks the dialyzer test for optimised booleans:
```elixir
  def and_check_optimized(arg) when is_integer(arg) do
    arg < :infinity and arg
  end
```
this now emits a warning that the pattern `false` can never match. It's weird because we have already all the clauses marked with `generated`, so I'm not sure why this is emitted. Anyway, the warning is very correct in here, since the `false` branch can never be taken, it's just pretty unreadable to the user - I'm not sure what to do about this.

### DIFF
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -1540,14 +1540,9 @@ defmodule Kernel do
     optimize_boolean(
       quote do
         case unquote(check) do
-          true ->
-            unquote(true_clause)
-
-          false ->
-            unquote(false_clause)
-
-          other ->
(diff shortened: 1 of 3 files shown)


josevalim:
Let’s remove the dialyzer test and see what comes out of it. It is only a
warning. It will either help people find bugs or we will be able to write
better tests.
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

michalmuskala:
Updated.

PR #7766, opened 2018-06-20
Link style guide in "mix format" documentation

qcam:
I might be late to the party but thanks for introducing the formatter in 1.6. It truly helps to maintain the code style among the team much more easier.

I would like to propose a change in the documentation for `mix format`. I linked @lexmag's style guide which I assume to be the guide that is being followed by the formatter. Please kindly advise if my assumption was wrong.

I think having a link to the guide in the docs would help the users to know what to expect when running the Mix task.


josevalim:
The Elixir formatter is its own style guide, it doesn't follow any in particular. :)

PR #7767, opened 2018-06-20
Allow force recompilation in IEx

qcam:
I think it would be nice to be able to force recompile right in the IEx during development.

### DIFF
--- a/lib/iex/lib/iex/helpers.ex
+++ b/lib/iex/lib/iex/helpers.ex
@@ -79,7 +79,7 @@ defmodule IEx.Helpers do
   This function is meant to be used for development and
   debugging purposes. Do not depend on it in production code.
   """
-  def recompile do
+  def recompile(force? \\ false) do
     if mix_started?() do
       config = Mix.Project.config()
       consolidation = Mix.Project.consolidation_path(config)
@@ -89,7 +89,8 @@ defmodule IEx.Helpers do
       Code.delete_path(consolidation)
       purge_protocols(consolidation)
 
-      {result, _} = Mix.Task.run("compile")
+      arguments = if force?, do: ["--force"], else: []
+      {result, _} = Mix.Task.run("compile", arguments)
 
       # Reenable consolidation and allow them to be loaded.
       Code.prepend_path(consolidation)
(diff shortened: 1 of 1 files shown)


josevalim:
I would like to avoid the boolean. Maybe `recompile force: true` or `recompile :force`. Thoughts?

qcam:
Yeah I think `force: true` is clearer. I will update the branch.

lib/iex/lib/iex/helpers.ex
```diff
@@ -89,7 +89,10 @@ defmodule IEx.Helpers do
       Code.delete_path(consolidation)
       purge_protocols(consolidation)
 
-      {result, _} = Mix.Task.run("compile")
+      force? = Keyword.fetch!(options, :force)
```
josevalim: You want to do `Keyword.get(options, :force, false)` otherwise `recompile!([])` will raise. This way you can set the default to an empty list.

lib/iex/lib/iex/helpers.ex
```diff
@@ -79,7 +79,7 @@ defmodule IEx.Helpers do
   This function is meant to be used for development and
   debugging purposes. Do not depend on it in production code.
   """
-  def recompile do
+  def recompile(options \\ [force: false]) do
```
josevalim: We need to document those options.

michalmuskala:
Isn't `recompile force: true` a bit lengthy for shell? You also wouldn't be able to easily tab-complete it.

qcam:
@michalmuskala how about just `recompile(:force)`?

JanStevens:
or go for `recompile!()` ? Bang indicating that it will force recompilation,

just my 2 cents flying by :D

josevalim:
@JanStevens we don't use bang with such meaning in Elixir. It means only that it may raise a semantical error.

qcam:
Shall we go `recompile(:force)` or `recompile(force: true)`, or would introducing another helper like `force_recompile` work 🙊?

whatyouhide:
I am in favor of `recompile(force: true)`. `recompile(:force)` is weird IMO, I'd rather type a few more characters.

josevalim:

PR #7768, opened 2018-06-20
Only evaluate Logger macro arguments when the message will be logged

### LINKED ISSUE #7046: Automatically wrap logger arguments in an anonymous function
This removes the need for compile time trickery and gives developers less chance to do expensive calls in production i.e:

    Logger.debug fn -> "foo #{inspect bar}" end

josevalim:
This means that the arguments won't be evaluated if the
message has a lower level than the logger one or if the
logger mode is set to discard.

Closes #7046.

### DIFF
--- a/lib/logger/lib/logger.ex
+++ b/lib/logger/lib/logger.ex
@@ -612,7 +612,7 @@ defmodule Logger do
   @doc """
   Logs a message dynamically.
 
-  Opposite to `log/3`, `debug/2`, `info/2` and friends. The arguments
+  Opposite to `log/3`, `debug/2`, `info/2`, and friends, the arguments
   given to `bare_log/3` are always evaluated. However, you can pass
   anonymous functions to `bare_log/3` and they will only be evaluated
   if there is something to be logged.
--- a/lib/logger/lib/logger.ex
+++ b/lib/logger/lib/logger.ex
@@ -25,23 +25,24 @@ defmodule Logger do
   system. For example, it may be helpful to log whenever a user is deleted.
 
       def delete_user(user) do
-        Logger.info fn ->
-          "Deleting user from the system: #{inspect(user)}"
-        end
+        Logger.info "Deleting user from the system: #{inspect(user)}"
         # ...
       end
 
   The `Logger.info/2` macro emits the provided message at the `:info`
-  level. There are additional macros for other levels. Notice the argument
-  passed to `Logger.info/2` in the above example is a zero argument function.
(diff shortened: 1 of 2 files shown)


adkron:
:heart:

lib/logger/lib/logger.ex
```diff
@@ -25,23 +25,24 @@ defmodule Logger do
...
-          "Deleting user from the system: #{inspect(user)}"
-        end
+        Logger.info "Deleting user from the system: #{inspect(user)}"
         # ...
       end
 
   The `Logger.info/2` macro emits the provided message at the `:info`
-  level. There are additional macros for other levels. Notice the argument
-  passed to `Logger.info/2` in the above example is a zero argument function.
+  level. Note the arguments given to `info/2` will only be evaluated
```
fertapric: I think if-clauses in first conditional can only use _will_ when they express willingness, as in requests. Then:

“Note the arguments given to `info/2` will only be evaluated if a message is logged.”

lib/logger/lib/logger.ex
```diff
@@ -627,41 +628,56 @@ defmodule Logger do
...
+  given to `bare_log/3` are always evaluated. However, you can pass
+  anonymous functions to `bare_log/3` and they will only be evaluated
+  if there is something to be logged.
   """
   @spec bare_log(level, message | (() -> message | {message, keyword}), keyword) ::
           :ok | {:error, :noproc} | {:error, term}
-  def bare_log(level, chardata_or_fun, metadata \\ [])
-      when level in @levels and is_list(metadata) do
+  def bare_log(level, chardata_or_fun, metadata \\ []) do
+    case __maybe_log__(level) do
```
fertapric: I think the name `__maybe_log__` suggests a function that might log or not, but this seems to be fetching the log "context" and checking if it should log or not. So, what do you think about?

```elixir
def bare_log(level, chardata_or_fun, metadata \\ []) do
  case __log_context__(level) do
    :error -> 
      :ok

    log_context ->
     if __should_log__(log_context), do: do_log(log_context, chardata_or_fun, metadata) 
  end
end

@doc false
def __log_context__(level) when level in @levels do
  case __metadata__() do
    {true, pdict} -> {level, Logger.Config.__data__(), pdict}
    {false, _} -> :error
  end
end

@doc false
def __should_log__({level, %{mode: mode, level: min_level}, _pdict}) do
  compare_levels(level, min_level) != :lt and mode != :discard
end
```

josevalim:

PR #7773, opened 2018-06-21
Warn if nested comparisons are found

lukevmorris:
Hi there! Here's a first pass at this. A few things to note:

* What is the rationale behind discouraging nested comparisons? It would be great to include this in the error message
* I continued the case chain in the name of consistency here, but the structure seems not great. I might tackle with a `with` in Elixir, but not sure how best to handle in Erlang
* I captured the "`Node`" in an effort to print out the whole comparison expression. But when I pass it to `to_string` at the end, I'm getting an erlang expression: `:erlang.<(:erlang.<(1, 3), 5)`. Still looking for a way to output the expression in Elixir
* Are there any other test cases you'd like to see?

Looking forward to any feedback y'all might have!


lib/elixir/src/elixir_expand.erl
```diff
@@ -776,23 +776,34 @@ check_erlang_operator_args({{'.', _, [erlang, '++']}, Meta, [ELeft, _]}, [Left,
...
+          true -> {Right, nil};
+          false ->
+            case is_comparison_expression(Left) of
+              true -> {nil, Node};
+              false ->
+                case is_comparison_expression(Right) of
+                  true -> {nil, Node};
+                  false -> false
+                end
+            end
```
josevalim: I think we can avoid the nested case by doing something like this:

```erlang
check_erlang_operator_args({{'.', _, [erlang, Op]}, Meta, [ELeft, ERight]} = Node, [Left, Right], _, E)
     when Op =:= '>'; Op =:= '<'; Op =:= '=<'; Op =:= '>=' ->
  case is_struct_comparison(ELeft, ERight, Left, Right) of
    false ->
      case is_comparison_expression(Op, ELeft, ERight, Left, Right) of
        false -> ...
        Comparison -> warn
    Struct -> warn
  end
```

lib/elixir/src/elixir_expand.erl
```diff
@@ -776,23 +776,34 @@ check_erlang_operator_args({{'.', _, [erlang, '++']}, Meta, [ELeft, _]}, [Left,
...
     case is_struct_expression(ELeft) of
-      true -> Left;
+      true -> {Left, nil};
       false ->
         case is_struct_expression(ERight) of
-          true -> Right;
-          false -> false
+          true -> {Right, nil};
+          false ->
+            case is_comparison_expression(Left) of
```
josevalim: We should not run validations on `Left` and `Right` because they haven't been expanded yet, so we may have false positives in case someone overrides `<` and friends. So you need to run this on `ELeft` and `ERight`.

josevalim:
> I captured the "Node" in an effort to print out the whole comparison expression. But when I pass it to to_string at the end, I'm getting an erlang expression: :erlang.<(:erlang.<(1, 3), 5). Still looking for a way to output the expression in Elixir

You should do something like this:

```erlang
{erlang_comparison_op_to_elixir(Op), Meta, [Left, Right]}
```

Where:

```erlang
erlang_comparison_op_to_elixir('=<') -> '<=';
erlang_comparison_op_to_elixir(Op) -> Op.
```

Note we already have a function named `translate_comp_op` that does this in `elixir_erl_compiler.erl`. So my suggestion is to move it to `elixir_utils.erl`, rename it to `erlang_comparison_op_to_elixir` for clarity and use it in both places.

lib/elixir/src/elixir_expand.erl
```diff
@@ -1165,6 +1181,12 @@ format_error({struct_comparison, StructExpr}) ->
                 "Comparing with a struct literal is unlikely to give a meaningful result. "
                 "Modules typically define a compare/2 function that can be used for "
                 "semantic comparison", [String]);
+format_error({nested_comparison, CompExpr}) ->
+  String = 'Elixir.Macro':to_string(CompExpr),
+  io_lib:format("nested comparisons are discouraged. Instead of ~ts,"
+                "consider joining together each comparison segment "
+                "with an &&.", [String]);
```
josevalim: My take would be this:

```erlang
"Elixir does not supported nested comparisons. Something like\n\n"
"    x < y < z\n\n"
"is equivalent to:\n\n"
"    (x < y) < z\n\n"
"which would ultimately compare the result of (x < y), which is a boolean, with z. "
"Instead consider joining together each comparison segment with an \"and\", such as\n\n"
"    x < y and y < z\n\n"
"You wrote: ~ts"
```

lib/elixir/src/elixir_expand.erl
lukevmorris: Ah great, this idea is cleaner

lib/elixir/src/elixir_expand.erl
lukevmorris: :+1:

lib/elixir/src/elixir_expand.erl
lukevmorris: Makes sense -- I've included this with one minor tweak

lukevmorris:
Thank you for the review! I've addressed your feedback and expanded the two tests slightly. Let me know if there is anything else I should do.

lukevmorris:
you can do it Travis!

lukevmorris:
OTP 20.1 failed with 
```
** (UndefinedFunctionError) function Inspect.inspect/2 is undefined (module Inspect is not available)
    (elixir) Inspect.inspect(List.Chars, %Inspect.Opts{base: :decimal, binaries: :infer, char_lists: :infer, charlists: :infer, limit: 50, pretty: false, printable_limit: 4096, safe: true, structs: true, syntax_colors: [], width: 80})
    (elixir) lib/kernel.ex:1931: Kernel.inspect/2
    (mix) lib/mix/tasks/compile.protocols.ex:151: Mix.Tasks.Compile.Protocols.consolidate/4
    (elixir) lib/task/supervised.ex:88: Task.Supervised.do_apply/2
    (elixir) lib/task/supervised.ex:38: Task.Supervised.reply/5
    (stdlib) proc_lib.erl:247: :proc_lib.init_p_do_apply/3
Function: #Function<9.128747118/0 in Mix.Tasks.Compile.Protocols.consolidate/6>
    Args: []
```

I looked through past builds and have seen this error before -- is this maybe a flaky test, or did I somehow indirectly cause this?

lukevmorris:
For reference, here is another build with this failure pattern: https://travis-ci.org/elixir-lang/elixir/jobs/392425911

josevalim:
@lukevmorris this failure is a race condition in our test suite that we have been chasing for a while since we can't reproduce it locally.

josevalim:

PR #7776, opened 2018-06-21
Take into account structs from stale local deps

### LINKED ISSUE #7765: Mix does not recompile file on local dependency change
### Environment

Erlang/OTP 20 [erts-9.3.3] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:10] [hipe] [kernel-poll:false]

* Operating system: 
Linux de1572f61a8c 4.13.0-43-generic #48~16.04.1-Ubuntu SMP Thu May 17 12:56:46 UTC 2018 x86_64 GNU/Linux

### Current behavior
First I create a application `a` that depends on `b`. `b` application is a local dependency:
```
mix new a
cd a
mix new b
```
Now in `a/mix.exs` I put the dependency `{:b, path: "./b"}`. The `b` app defines a struct 
```
 defstruct [:b1, :b2]
``` 
and `a` uses it like
```
IO.inspect %B{b1: 1, b2: 2}
```
which when executed prints (as expected)
```
%B{b1: 1, b2: 2}
```

Now if I remove the `b2` field from the `B` struct, `a` does not gets recompiled and prints
```
%{__struct__: B, b1: 1, b2: 2}
```
### Expected behavior
I would expect `mix` to recompile `a` (failing), since one of its dependency *defining a struct that `a` uses* (`b`) changed. 

Just saving the file that uses the `B` struct forces a recompile failing with (which should be the expected behaviour)
```

== Compilation error in file lib/a.ex ==
** (KeyError) key :b2 not found in: %B{b1: 1}
    (stdlib) :maps.update(:b2, 2, %B{b1: 1})
    lib/b.ex:6: anonymous fn/2 in B.__struct__/1
    (elixir) lib/enum.ex:1899: Enum."-reduce/3-lists^foldl/2-0-"/3

```

uesteibar:
Fixes #7765

### DIFF
--- a/lib/mix/test/fixtures/umbrella_dep/deps/umbrella/apps/foo/lib/foo_struct.ex
+++ b/lib/mix/test/fixtures/umbrella_dep/deps/umbrella/apps/foo/lib/foo_struct.ex
@@ -1,0 +1,3 @@
+defmodule FooStruct do
+  defstruct [:bar]
+end
--- a/lib/mix/test/fixtures/umbrella_dep/deps/umbrella/apps/foo/lib/foo_struct.ex
+++ b/lib/mix/test/fixtures/umbrella_dep/deps/umbrella/apps/foo/lib/foo_struct.ex
@@ -1,3 +1,0 @@
-defmodule FooStruct do
-  defstruct [:bar]
-end
--- a/lib/mix/lib/mix/compilers/elixir.ex
+++ b/lib/mix/lib/mix/compilers/elixir.ex
@@ -85,13 +85,15 @@ defmodule Mix.Compilers.Elixir do
               into: new_paths,
               do: source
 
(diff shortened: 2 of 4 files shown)


lib/mix/test/fixtures/umbrella_dep/deps/umbrella/apps/foo/lib/foo_struct.ex
```diff
@@ -0,0 +1,3 @@
+defmodule FooStruct do
+  defstruct [:bar]
+end
```
josevalim: Instead of adding a new module, let's just write to `lib/foo.ex` a file with a struct. We should prefer to generate the data in the test itself rather than rely on fixtures. :)

josevalim:
@uesteibar I have added one comment and we should be good to go!

josevalim:

PR #7778, opened 2018-06-21
Include caller information on struct expansion errors, closes #7752

### LINKED ISSUE #7752: No line number on error when using invalid struct key
### Environment

Erlang/OTP 20 [erts-9.3.2] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:10] [hipe] [kernel-poll:false] [dtrace]

### Current behavior

When compiling `lib/my_app/my_module.ex`
```
defmodule MyApp.MyModule do
  def my_function do
    %MyApp.MyStruct{not_existing_key: 1}
  end
end
```
the compiler crashes and outputs no line number 
```
== Compilation error in file lib/my_app/my_module.ex ==
** (KeyError) key : not_existing_key not found in: %MyApp.MyStruct{key: nil}
    (stdlib) :maps.update(:modifier, 1.1, %MyApp.MyStruct{key: nil})
    lib/my_app/my_struct.ex:6: anonymous fn/2 in MyApp.MyStruct.__struct__/1
    (elixir) lib/enum.ex:1899: Enum."-reduce/3-lists^foldl/2-0-"/3
```

It works for similar code

```
defmodule MyApp.MyModule do
  def my_function(%MyApp.MyStruct{not_existing_key: 1}) do
    :ok
  end
end
```

```
== Compilation error in file lib/my_app/my_module.ex ==
** (CompileError) lib/my_app/my_module.ex:2: unknown key : not_existing_key for struct MyApp.MyStruct
    lib/my_app/my_module.ex:2: (module)
```

```
defmodule MyApp.MyModule do
  def my_function(param) do
    %MyApp.MyStruct{not_existing_key: a} = param
    a
  end
end
```

```
== Compilation error in file lib/my_app/my_module.ex ==
** (CompileError) lib/my_app/my_module.ex:3: unknown key : not_existing_key for struct MyApp.MyStruct
    lib/my_app/my_module.ex:3: (module)
```
### Expected behavior
Explicit line number in error message

derekkraan:
Partial credit: @tuxified helped me figure this puzzle out.

### DIFF
--- a/lib/elixir/src/elixir_map.erl
+++ b/lib/elixir/src/elixir_map.erl
@@ -149,10 +149,9 @@ load_struct(Meta, Name, Args, InContext, E) ->
       end;
 
     Kind:Reason ->
-      Stacktrace = erlang:get_stacktrace(),
       Info = [{Name, '__struct__', Arity, [{file, "expanding struct"}]},
               elixir_utils:caller(?line(Meta), ?key(E, file), ?key(E, module), ?key(E, function))],
-      erlang:raise(Kind, Reason, prune_stacktrace(Stacktrace, Name, Arity) ++ Info)
+      erlang:raise(Kind, Reason, Info)
   end.
 
 prune_stacktrace([{Module, '__struct__', Arity, _} | _], Module, Arity) ->
--- a/lib/elixir/test/elixir/kernel/errors_test.exs
+++ b/lib/elixir/test/elixir/kernel/errors_test.exs
@@ -440,7 +440,7 @@ defmodule Kernel.ErrorsTest do
     end
 
     assert_eval_raise KeyError,
-                      "key :age not found in: %Kernel.ErrorsTest.GoodStruct{name: \"john\"}",
+                      "key :age not found",
                       '%#{GoodStruct}{age: 27}'
 
     assert_eval_raise CompileError,


josevalim:
Hi @derekkraan!

Unfortunately the KeyError is just one of many errors that could happen there. [The clause right below the one you added](https://github.com/elixir-lang/elixir/pull/7778/files#diff-fe1fd5d026e67e3b3378536e223ebaacR154) was supposed to add a relevant stacktrace line that says "expanding struct" but it doesn't. Maybe we should investigate why?

Thank you!

josevalim:
@derekkraan maybe we should forego of the [stacktrace altogether here](https://github.com/derekkraan/elixir/blob/14165c918445dac0fd6c9412234ea2a95b17f7bb/lib/elixir/src/elixir_map.erl#L158) and call `erlang:raise(Kind, Reason, Info)` instead?

derekkraan:
I pushed a new commit. Is this what you meant? Then we have the correct handling for a key error and we don't attempt to add stacktrace info for other cases.

lib/elixir/src/elixir_map.erl
```diff
@@ -148,11 +148,13 @@ load_struct(Meta, Name, Args, InContext, E) ->
           form_error(Meta, ?key(E, file), ?MODULE, {undefined_struct, Name, Arity})
       end;
 
+    error:{badkey,Key} ->
+      form_error(Meta, ?key(E, file), ?MODULE, {unknown_key_for_struct, Name, Key});
```
josevalim: Perfect. But now I think we can get rid of this clause altogether. What do you think?

derekkraan: I tested it without this clause, and then you get KeyError, but no line number info. Is it maybe better to change the catch-all clause at the end to also use `form_error`?

derekkraan: Actually I'm not sure how that would work ... I'm not sure whether it's possible to write the general case to give a good error message when you try to use a key that doesn't exist.

derekkraan: `KeyError (key :age not found)` is what we see if we remove this clause.

derekkraan: I stand corrected, the stacktrace is good :joy:

derekkraan: I'm actually of two minds on this. If we remove this clause then we get an error message with a line number, so in that sense it's good. The error message is a little bit different though, so for consistency's sake with the rest of Elixir it might be a good idea to keep it in.

Without this clause:
`key :age not found`
with this clause:
`key :age not found in: %Kernel.ErrorsTest.GoodStruct{name: \"john\"}`

josevalim: @derekkraan Yeah, I noticed that. I am not convinced that printing the struct is good though. It works fine for small structs but in practice we have large ones and they end-up adding too much noise.

derekkraan: Ok. It's also not really necessary to find the error and fix it. I'll remove the clause and squash commits.

michalmuskala:
What do you think about implementing `Exception.blame` for `KeyError` and in case the value is a struct (or a map in general) provide a did-you-mean style help?

josevalim:
@michalmuskala I think it would be great. Btw, the did_you_mean feature is not part of blame, but it probably should.

derekkraan:
I've squashed the commits and pushed again.

Adding `KeyError.blame/2` seems like it's out of the scope of this PR, but if you feel that it isn't then I could give it a shot.

michalmuskala:
I agree. I created a separate issue to track this https://github.com/elixir-lang/elixir/issues/7779

josevalim:

PR #7783, opened 2018-06-22
Adding documentation for an example of variable assignment with the pipe operator

smorin:
Adding documentation for an example of variable assignment with the pipe operator

### DIFF
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -3145,7 +3145,7 @@ defmodule Kernel do
 
   Assignment is also possible with the pipe operator.  In the example below we
   assign flatten list to the variable `flattened_list`:
-  
+
       iex> flattened_list = [1, [2], 3] |> List.flatten()
       [1, 2, 3]
 
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -3143,6 +3143,12 @@ defmodule Kernel do
 
       Enum.map(List.flatten([1, [2], 3]), fn x -> x * 2 end)
 
+  Assignment is also possible with the pipe operator.  In the example below we
+  assign flatten list to the variable `flattened_list`:
+  
+      iex> flattened_list = [1, [2], 3] |> List.flatten()
+      [1, 2, 3]
+
   ## Pitfalls
 
   There are two common pitfalls when using the pipe operator.


lib/elixir/lib/kernel.ex
```diff
@@ -3143,6 +3143,12 @@ defmodule Kernel do
 
       Enum.map(List.flatten([1, [2], 3]), fn x -> x * 2 end)
 
+  Assignment is also possible with the pipe operator.  In the example below we
+  assign flatten list to the variable `flattened_list`:
+
+      iex> flattened_list = [1, [2], 3] |> List.flatten()
+      [1, 2, 3]
+
```
josevalim: I am a little bit skeptical about adding this change here. The result of every operator in the language can be assigned to a variable. This is a general language property and not specific to how this operator is used. Does it mean we need to add a similar sentence to every operator? Thoughts?

smorin:
No that’s not needed to add everywhere.  It’s a special form not usually seen and this helps clear things up.  I believe most other forms are self explanatory.

whatyouhide:
I am 👎 on this for the reason @josevalim suggested. Everything in Elixir is an expression (including most non-weird - say `left . right` - special forms), so you can assign everything to a variable. I don't think the pipe operator is any help here.

josevalim:
Thanks @smorin! I do understand why you want this addition: it is a very common idiom, not existent in many other traditional languages, so such questions may arise. I think we can clarify this in other places, like you did in the getting started guide, but I don't think it belongs to the API reference.

smorin:
I understand not wanting to bloat the documentation and that should this be repeated everywhere concern.  But this is basic part of the language but not a obvious one especially coming from a another languages where your taught and conditioned to learn individual forms.

Insert python if example where assignments don't happen.

```python
>>> x = int(input("Please enter an integer: "))
Please enter an integer: 42
>>> if x < 0:
...     x = 0
...     print('Negative changed to zero')
... elif x == 0:
...     print('Zero')
... elif x == 1:
...     print('Single')
... else:
...     print('More')
...
More
```
Ref: https://docs.python.org/3/tutorial/controlflow.html

Is documentation for the experts that know 'Everything in Elixir is an expression' or the people that don’t know something?

Is documentation to make it easy to learn Elixir or should we be teaching people where to learn and find pieces of information?

I am adding this because I was looking where to find this as I was learning the language and elixir pipe operator assignment doesn't google well.

I would say if you see:

To a newbie which would be correct

```elixir
  [1, [2], 3]
  |> List.flatten()
  |> IO.inspect() 
  |> flattened_list
```
or

```elixir
  [1, [2], 3]
  |> List.flatten()
  |> IO.inspect()
= flattened_list
```
or

```elixir
  [1, [2], 3]
  |> List.flatten()
  |> IO.inspect() =>  flattened_list 
```

Or proper way

```elixir
flattened_list =
  [1, [2], 3]
  |> List.flatten()
  |> IO.inspect()
```
Which should be right to a beginner?

I would argue documentation is for the person that doesn’t know something.

Let’s make it easy as possible for people to learn Elixir.  The reason I want to add this is because it wasn’t obvious to me.  Then I had to go searching on the internet and fiddle around IEX to figure it out something that should be captured here.

Thinking about it maybe assignment should also be added to a few other limited core structures like `case` too.

josevalim:
We don’t disagree with anything you said. Our only concern is that it
doesn’t belong here.

For example, a “patterns and expressions” section in the getting started
that explains how we only have patterns and expressions and no statements
in Elixir could probably do a much better job of explaining those
distinctions at a conceptual level rather than a note that feels specific
to the pipe operator.
-- 


smorin:
If I am searching how to do this I am going to look for pipe operator reference not general assignment reference.

josevalim:
@smorin if we add a bunch of loosely related information to the pipeline operator docs, then finding something that is actually relevant to the pipeline operator gets harder.

smorin:
It’s not loosely, it’s using the pipe operator.  It’s a few lines extra that doesn’t make reference that much harder below with links as part of the navigation. 

We need to make it as practical as possible for newer users to find information in the places search for this information directly.

PR #7784, opened 2018-06-22
Correct grammar in Logger documentation

fertapric:
If-clauses in first conditional shouldn’t double up on the future “will”.

I double check with three native English speakes :)


josevalim:

PR #7785, opened 2018-06-22
Deprecate using non-empty lists with the Collectable protocol

### LINKED ISSUE #7421: Enum.into/2 is broken for keyword lists
See example below:

```iex
iex(6)> Enum.into([a: 2], %{a: 1})[:a]
2
iex(7)> Enum.into([a: 2], [a: 1])[:a]
1
```

The map result is correct, as the left side overrides the right side. To be more precise, `Enum.into/2` for two lists should be equivalent to `left ++ right`. We need to change both the shortcuts in `Enum.into/2` as well as the collectable implementation for lists.

whatyouhide:
Closes #7421.

### DIFF
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -1164,10 +1164,14 @@ defmodule Enum do
   @doc """
   Inserts the given `enumerable` into a `collectable`.
 
+  Note that passing a non-empty list as the `collectable` is deprecated. If you're collecting
+  into a non-empty keyword list, consider using `Keyword.merge/2`. If you're collecting into a
+  non-empty list, consider something like `to_list(enumerable) ++ collectable`.
+
   ## Examples
 
-      iex> Enum.into([1, 2], [0])
-      [0, 1, 2]
+      iex> Enum.into([1, 2], [])
+      [1, 2]
 
       iex> Enum.into([a: 1, b: 2], %{})
       %{a: 1, b: 2}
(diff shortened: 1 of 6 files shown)


josevalim:
I think we need to deprecate something in elixir_erl_for as we also inline
some lists in there.

Or maybe even better: whenever a non empty list is given, let it fallback
to the Collectable implementation, so we get the warning. We could use this
technique in Enum.into/2 too.
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

whatyouhide:
@josevalim this is ready for actual review and I did what you suggested. However I couldn't find any special casing in `elixir_erl_for`, only the call to `Collectable.into/1`.

lib/elixir/lib/enum.ex
```diff
@@ -1180,8 +1184,14 @@ defmodule Enum do
 
   """
   @spec into(Enumerable.t(), Collectable.t()) :: Collectable.t()
+  def into(enumerable, collectable)
+
+  def into(enumerable, []) do
+    to_list(enumerable)
+  end
+
   def into(enumerable, collectable) when is_list(collectable) do
```
josevalim: We can get rid of this clause altogether, no?

whatyouhide: Good catch, removed!

michalmuskala:
I would like to assess how many times this will be triggered in a phoenix or a nerves application. If it's going to print a ton, i don't think we should merge it. It will be extremely annoying especially if the source would be in a library.

whatyouhide:
I didn't have time yet to try it out on bigger stuff (like a new Phoenix project) but it already warns from Hex when fetching dependencies 😄

josevalim:
@whatyouhide one of the things to do is to merge this right after v1.7 is out, to include this on v1.8. This means whoever is on Elixir master will have plenty of time to fix those. Maybe adding it now, one month before the release, is indeed too risky.

josevalim:
Meanwhile @whatyouhide, can you please fix Hex?

whatyouhide:
@josevalim ahah yeah, already on it :)

josevalim:

PR #7787, opened 2018-06-22
Fix description of max_children option

van-mronov:
It returns `{:error, :max_children}` instead of `{:error, :dynamic}`.


josevalim:

PR #7788, opened 2018-06-23
Polish documentation of Kernel module

fertapric:
Minor corrections in the grammar of the documentation of the `Kernel` module:

 * enumerate the arguments of the function in `get_in/2` in the same way as in `get_and_update_in/3`. [This new style was applied recently.](https://github.com/elixir-lang/elixir/commit/19105a3bd2437250113776164ddf1209a65da387)
* use back-ticks to enable auto-linking of the `Access` module.
* use "in case of" instead of "in case". ["in case" shouldn't be used to mean "if".](https://dictionary.cambridge.org/grammar/british-grammar/conditionals-and-wishes/in-case-of)
* add missing "the" articles.

### DIFF
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -2041,8 +2041,11 @@ defmodule Kernel do
   function.
 
   If a key is a function, the function will be invoked
-  passing three arguments, the operation (`:get`), the
-  data to be accessed, and a function to be invoked next.
+  passing three arguments: 
+  
+    * the operation (`:get`)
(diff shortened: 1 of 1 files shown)


fertapric:
Soungs good to me. I’m not a native speaker either. 

To be honest, I also had doubts regarding “in case of”, but then I found the link from the Cambridge Dictionary and I thought “I’ve been using it wrong all the time!” 😅

whatyouhide:
Ah, great to know! Thanks for following up @fertapric! 💟

PR #7790, opened 2018-06-25
Updated with umbrella project example

smorin:
Updated with umbrella project example


whatyouhide:
Great work, thanks @smorin! 💟

PR #7792, opened 2018-06-25
Example: aliases tasks 2 show its capabilities

smorin:
Example of aliases calling tasks to show its capabilities

### DIFF
--- a/lib/mix/lib/mix.ex
+++ b/lib/mix/lib/mix.ex
@@ -182,10 +182,10 @@ defmodule Mix do
 
   In the example above we have created 3 aliases, the first example 
   `taskalias` will run task `hex.info`, then (run)[Mix.Tasks.Run.run/1]
-  to run a elixir script, then (cmd)[Mix.Tasks.Cmd.run/1] to run a 
+  to run an elixir script, then (cmd)[Mix.Tasks.Cmd.run/1] to run a 
   command line bash script. This shows how powerful aliases mixed 
-  with Tasks can be.  `taskalias2` shows a limitation of tasks where 
-  only one of a give task will run, the second execution of `run priv/hello.exs` 
+  with Mix.Task can be. `taskalias2` shows a limitation of tasks where 
+  only one of the given tasks will run, the second execution of `run priv/hello.exs` 
   will not run.
 
   ## Environment variables
--- a/lib/mix/lib/mix.ex
+++ b/lib/mix/lib/mix.ex
@@ -165,13 +165,14 @@ defmodule Mix do
   Aliases can be used very powerfully to also run elixir scripts and
   bash commands, for example:
       
-      #priv/hello.exs
+      # priv/hello.exs
       IO.puts("hello")
 
-      #priv/world.sh
+      # priv/world.sh
       #!/bin/sh
       echo "world!"
 
+      # mix.exs
       defp create_aliases do
         [
           "taskalias": ["hex.info", "run priv/hello.exs", "cmd priv/world.sh"],
--- a/lib/mix/lib/mix.ex
+++ b/lib/mix/lib/mix.ex
@@ -176,17 +176,22 @@ defmodule Mix do
       defp create_aliases do
         [
           "taskalias": ["hex.info", "run priv/hello.exs", "cmd priv/world.sh"],
-          "taskalias2": ["run priv/hello.exs", "run priv/hello.exs"]
+          "taskalias2": ["run priv/hello1.exs", "run priv/hello2.exs"]
         ]
       end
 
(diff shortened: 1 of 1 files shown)


lib/mix/lib/mix.ex
```diff
@@ -161,6 +161,30 @@ defmodule Mix do
...
+          "taskalias": ["hex.info", "run priv/hello.exs", "cmd priv/world.sh"],
+          "taskalias2": ["run priv/hello.exs", "run priv/hello.exs"]
+        ]
+      end
+
+  In the example above we have created 3 aliases, the first example 
+  `taskalias` will run task `hex.info`, then (run)[Mix.Tasks.Run.run/1]
+  to run a elixir script, then (cmd)[Mix.Tasks.Cmd.run/1] to run a 
+  command line bash script. This shows how powerful aliases mixed 
+  with Tasks can be.  `taskalias2` shows a limitation of tasks where
```
fertapric: Double space before `taskalias2`

lib/mix/lib/mix.ex
```diff
@@ -161,6 +161,30 @@ defmodule Mix do
   Note aliases do not show up on `mix help`.
   Aliases defined in the current project do not affect its dependencies and aliases defined in dependencies are not accessible from the current project.
 
+  Aliases can be used very powerfully to also run elixir scripts and bash commands, for example:
```
fertapric: Checking the documentation I think it should be "Elixir"

lib/mix/lib/mix.ex
```diff
@@ -161,6 +161,30 @@ defmodule Mix do
...
+      
+      #priv/hello.exs
```
fertapric: An space should be included after the `#`. [Here is an example in the documentation.](https://github.com/elixir-lang/elixir/blob/master/lib/elixir/lib/kernel/special_forms.ex#L1118)

lib/mix/lib/mix.ex
fertapric: [Here is an example in the documentation](https://github.com/elixir-lang/elixir/blob/master/lib/elixir/lib/float.ex#L21)

lib/mix/lib/mix.ex
```diff
@@ -161,6 +161,30 @@ defmodule Mix do
...
+      IO.puts("hello")
+
+      #priv/world.sh
+      #!/bin/sh
+      echo "world!"
+
+      defp create_aliases do
```
fertapric: I would include here `# mix.exs`

lib/mix/lib/mix.ex
fertapric: `Tasks` could be confused with [`Task`](https://hexdocs.pm/elixir/Task.html#content). Maybe `Mix.Task`, `mix tasks` or simply `tasks`?

lib/mix/lib/mix.ex
```diff
@@ -161,6 +161,30 @@ defmodule Mix do
...
+        [
+          "taskalias": ["hex.info", "run priv/hello.exs", "cmd priv/world.sh"],
+          "taskalias2": ["run priv/hello.exs", "run priv/hello.exs"]
+        ]
+      end
+
+  In the example above we have created 3 aliases, the first example 
+  `taskalias` will run task `hex.info`, then (run)[Mix.Tasks.Run.run/1]
+  to run a elixir script, then (cmd)[Mix.Tasks.Cmd.run/1] to run a
```
fertapric: "an Elixir"

lib/mix/lib/mix.ex
```diff
@@ -161,6 +161,30 @@ defmodule Mix do
...
+          "taskalias2": ["run priv/hello.exs", "run priv/hello.exs"]
+        ]
+      end
+
+  In the example above we have created 3 aliases, the first example 
+  `taskalias` will run task `hex.info`, then (run)[Mix.Tasks.Run.run/1]
+  to run a elixir script, then (cmd)[Mix.Tasks.Cmd.run/1] to run a 
+  command line bash script. This shows how powerful aliases mixed 
+  with Tasks can be.  `taskalias2` shows a limitation of tasks where 
+  only one of a give task will run, the second execution of `run priv/hello.exs`
```
fertapric: Maybe "...shows a limitation of tasks where only one of the **given tasks** will run"?

lib/mix/lib/mix.ex
fertapric: I'm not sure what's the official line length for comments, but I would wrap this line (and the one above) :)

smorin:
Can't tell why this failed anyone have ideas?

josevalim:

Issue #7795, opened 2018-06-25
Introduce assert_no_guard_scope in Kernel


michalmuskala:
Several macros in `Kernel` cannot be used inside guards - `!`, `&&` and `||` being the most notable examples. We should fail gracefully when that happens. Right now the error is "cannot use case in guards", which is not very helpful.

The simplest way to do this would probably to introduce a private `assert_no_guard_scope` function verifying `__CALLER__.context`, similar to `assert_module_scope` and `assert_no_function_scope`.

Sihui:
Hi @michalmuskala, I reuse the error message from [elixir_expand](https://github.com/elixir-lang/elixir/blob/master/lib/elixir/src/elixir_expand.erl#L1102). Not sure if we want something different than that. I also only use the new `assert_no_guard_scope` in  `!`, `&&` and `||`. Let me know if you think I should use it somewhere else as well.

Maybe I miss your point. Why is this better than what we already have?

josevalim:
@Sihui the difference is that we raise the error closer to the user code, so the error message will be more precise and relevant. :)

Sihui:
That makes sense. Thanks, @josevalim! :)

devonestes:
I think this can be closed now. It wasn't closed automatically for some reason when the PR was merged.

PR #7797, opened 2018-06-26
[Kernel] add assert no guard

Sihui:
ref: https://github.com/elixir-lang/elixir/issues/7795

### DIFF
--- a/lib/elixir/test/elixir/kernel/guard_test.exs
+++ b/lib/elixir/test/elixir/kernel/guard_test.exs
@@ -290,19 +290,19 @@ defmodule Kernel.GuardTest do
 
       # Consistent errors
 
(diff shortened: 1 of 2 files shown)


josevalim:

lib/elixir/test/elixir/kernel/guard_test.exs
```diff
@@ -290,19 +290,19 @@ defmodule Kernel.GuardTest do
 
       # Consistent errors
 
-      assert_raise CompileError, ~r"invalid expression in guard", fn ->
+      assert_raise ArgumentError, ~r"invalid expression in guard", fn ->
```
fertapric: I would match on the message too, just to be sure:

```elixir
assert_raise ArgumentError, ~r"invalid expression in guard, ! is not allowed in guards", fn ->
```

PR #7799, opened 2018-06-26
Update mix.ex

fertapric:
Minor changes in documentation:

* use "Elixir" instead of "elixir". [The comment was collapsed in the original PR and it was ignored 😞 ](https://github.com/elixir-lang/elixir/pull/7792#discussion_r197716526)
* correct the number of aliases
* remove double space

cc @smorin


josevalim:

PR #7801, opened 2018-06-26
Make test suite for assert_no_guard more robust

fertapric:
If I got it right, one of the reasons for that change was to get the message:

> invalid expression in guard, **!** is not allowed in guards. To learn more about guards, visit: https://hexdocs.pm/elixir/guards.html  

instead of:

> invalid expression in guard, **case** is not allowed in guards. To learn more about guards, visit: https://hexdocs.pm/elixir/guards.html

This change should make the test suite more robust for future changes or refactors :)

In addition, I've split a couple of strings that were exceeding the maximum line length (98). By the way, should the formatter warn when it cannot keep the code under the desired line length? maybe with a `--strict` flag?

### DIFF
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -4877,7 +4877,8 @@ defmodule Kernel do
 
       if is_list(funs) do
         message =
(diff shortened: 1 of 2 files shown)


michalmuskala:
Thank you! :heart:

PR #7803, opened 2018-06-26
Add `blame` callback for KeyError

### LINKED ISSUE #7779: Provide KeyError.blame
The `KeyError` exception should provide a did-you-mean style suggestion in the optional `blame` callback. It should check the `term` value and if it's a map or a keyword, it should extract available keys and compare against the requested value.
For strings and atoms this is easy, it's probably harder to do for other types - the initial implementation might do this just for strings and atoms.

devonestes:
This starts with an initial implementation of the `blame/2` callback
for `KeyError` to add some helpful `did_you_mean` feedback for
potentially typo'd keys. Right now it's only implemented for maps and
keyword lists, and only for string and atom keys.

Resolves #7779

### DIFF
--- a/lib/elixir/lib/exception.ex
+++ b/lib/elixir/lib/exception.ex
@@ -1091,17 +1091,84 @@ defmodule Protocol.UndefinedError do
 end
 
 defmodule KeyError do
-  defexception [:key, :term]
+  defexception [:key, :term, :message]
 
   @impl true
-  def message(exception) do
-    msg = "key #{inspect(exception.key)} not found"
+  def message(exception = %{message: nil}), do: message(exception.key, exception.term)
+  def message(%{message: message}), do: message
 
-    if exception.term != nil do
-      msg <> " in: #{inspect(exception.term)}"
+  def message(key, term) do
+    msg = "key #{inspect(key)} not found"
+
+    if term != nil do
+      msg <> " in: #{inspect(term)}"
     else
       msg
     end
+  end
+
+  @impl true
+  def blame(exception = %{term: nil}, stacktrace) do
+    message = message(exception.key, exception.term)
+    {%{exception | message: message}, stacktrace}
+  end
+
+  def blame(exception = %{term: term}, stacktrace) when is_map(term) do
+    available_keys = Map.keys(term)
+    hint = did_you_mean(exception.key, available_keys)
+    message = message(exception.key, term) <> IO.iodata_to_binary(hint)
+    {%{exception | message: message}, stacktrace}
+  end
+
+  def blame(exception = %{term: term}, stacktrace) when is_list(term) do
+    available_keys = Keyword.keys(term)
+    hint = did_you_mean(exception.key, available_keys)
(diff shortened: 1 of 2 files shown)


lib/elixir/lib/exception.ex
```diff
@@ -1091,18 +1091,85 @@ defmodule Protocol.UndefinedError do
...
+  defexception [:key, :term, :message]
 
   @impl true
-  def message(exception) do
-    msg = "key #{inspect(exception.key)} not found"
+  def message(exception = %{message: nil}), do: message(exception.key, exception.term)
+  def message(%{message: message}), do: message
+
+  def message(key, term) do
+    msg = "key #{inspect(key)} not found"
```
fertapric: Minor nitpick: what do you think about using `message` instead of `msg` (like the function above) and `distance` instead of `dist` in the functions below?

devonestes: Yeah, works for me!

lib/elixir/lib/exception.ex
```diff
@@ -1091,18 +1091,85 @@ defmodule Protocol.UndefinedError do
...
+
+  def blame(exception = %{term: term}, stacktrace) when is_map(term) do
+    available_keys = Map.keys(term)
+    hint = did_you_mean(exception.key, available_keys)
+    message = message(exception.key, term) <> IO.iodata_to_binary(hint)
+    {%{exception | message: message}, stacktrace}
+  end
+
+  def blame(exception = %{term: term}, stacktrace) when is_list(term) do
+    available_keys = Keyword.keys(term)
```
josevalim: Unfortunately I think there is a bug here. `Keyword.keys` assumes a keyword is given. So if you do `Keyword.fetch!([1, foo: :bar], :bar)`, it will invoke this clause which then fails on Keyword.keys. My suggestion is to unify both clauses above and do something like this:

    if (is_map(term) or Keyword.keyword?(term)) and Enum.all?(term, fn {k, _} -> is_atom(k) end) do
      ...
    else
      {message, stacktrace}
    end

I know we can have mixed keys in maps, but I think we should start with a more conservative approach for now.

lib/elixir/test/elixir/exception_test.exs
```diff
@@ -482,6 +482,42 @@ defmodule ExceptionTest do
...
+    test "annotates key error with suggestions" do
+      message = blame_message(%{first: nil, second: nil}, fn map -> map.firts end)
+
+      assert message == """
+             key :firts not found in: %{first: nil, second: nil}. Did you mean one of:
+
+                   * :first
+             """
+
+      message = blame_message(%{"first" => nil, "second" => nil}, fn map -> map.firts end)
```
josevalim: We should not compare atoms with strings. After all, `%{"first" => nil}.first` also wouldn't work. I think for now we should only provide did you mean on atom access. We can always expand later.

lib/elixir/lib/exception.ex
devonestes: Ah - I didn't think it was possible for a non-keyword list to be passed as the term to a `KeyError`. I'll update.

PR #7805, opened 2018-06-27
Make mix help list aliases

Sihui:
ref: https://trello.com/c/9DzdXbp4/4-make-mix-help-list-aliases

cc @michalmuskala

### DIFF
--- a/lib/mix/lib/mix/tasks/help.ex
+++ b/lib/mix/lib/mix/tasks/help.ex
@@ -195,6 +195,11 @@ defmodule Mix.Tasks.Help do
     end)
   end
 
+  defp task_str(task_name) when is_list(task_name) do
+    task_strs = Enum.map(task_name, &task_str/1) |> Enum.join(", ")
+    "[" <> task_strs <> "]"
+  end
+
   defp task_str(task_name) when is_bitstring(task_name), do: task_name
   defp task_str(task_name), do: inspect(task_name)
 end
--- a/lib/mix/lib/mix/tasks/help.ex
+++ b/lib/mix/lib/mix/tasks/help.ex
@@ -72,7 +72,7 @@ defmodule Mix.Tasks.Help do
     loadpaths!()
 
     modules = Enum.filter(load_tasks(), &String.contains?(Mix.Task.task_name(&1), pattern))
-    aliases = Enum.filter(load_aliases(), &String.contains?(elem(&1, 0), pattern))
+    aliases = Enum.filter(load_aliases(), fn {name, _} -> String.contains?(name, pattern) end)
 
     {docs, max} = build_doc_list(modules, aliases)
 
--- a/lib/mix/lib/mix/tasks/help.ex
+++ b/lib/mix/lib/mix/tasks/help.ex
@@ -4,11 +4,12 @@ defmodule Mix.Tasks.Help do
   @shortdoc "Prints help information for tasks"
 
   @moduledoc """
-  Lists all tasks or prints the documentation for a given task.
+  Lists all aliases and tasks or prints the documentation for a given task.
 
   ## Arguments
 
-      mix help                  - prints all tasks and their shortdoc
+      mix help                  - prints all aliases, tasks and task shortdoc
+      mix help ALIAS            - prints the definition for the given alias
       mix help TASK             - prints full docs for the given task
       mix help --search PATTERN - prints all tasks that contain PATTERN in the name
       mix help --names          - prints all task names and aliases
@@ -89,8 +90,7 @@ defmodule Mix.Tasks.Help do
   def run([task]) do
     loadpaths!()
 
-    module = Mix.Task.get!(task)
-    doc = Mix.Task.moduledoc(module) || "There is no documentation for this task"
+    {doc, location} = verbose_doc(task)
     opts = Application.get_env(:mix, :colors)
 
     if ansi_docs?(opts) do
@@ -102,7 +102,7 @@ defmodule Mix.Tasks.Help do
       IO.puts(doc)
     end
 
-    IO.puts("Location: #{where_is_file(module)}")
+    IO.puts("Location: #{location}")
   end
 
   def run(_) do
@@ -125,6 +125,7 @@ defmodule Mix.Tasks.Help do
   defp load_aliases() do
     Mix.Project.config()[:aliases]
     |> Enum.map(fn {k, v} -> {Atom.to_string(k), v} end)
+    |> Map.new()
   end
 
   defp ansi_docs?(opts) do
@@ -188,18 +189,24 @@ defmodule Mix.Tasks.Help do
   end
 
   defp build_alias_doc_list(aliases) do
-    Enum.reduce(aliases, {[], 0}, fn {alias_name, task_name}, {docs, max} ->
-      doc = "Alias for " <> task_str(task_name)
+    Enum.reduce(aliases, {[], 0}, fn {alias_name, _task_name}, {docs, max} ->
+      doc = "Alias defined in mix.exs"
       task = "mix " <> alias_name
       {[{task, doc} | docs], max(byte_size(task), max)}
     end)
   end
 
-  defp task_str(task_name) when is_list(task_name) do
-    task_strs = Enum.map(task_name, &task_str/1) |> Enum.join(", ")
-    "[" <> task_strs <> "]"
-  end
+  defp verbose_doc(task) do
+    aliases = load_aliases()
 
-  defp task_str(task_name) when is_bitstring(task_name), do: task_name
-  defp task_str(task_name), do: inspect(task_name)
+    if Map.has_key?(aliases, task) do
+      task_name = aliases[task]
+      doc = "Alias for " <> inspect(task_name)
+      {doc, "mix.exs"}
+    else
+      module = Mix.Task.get!(task)
+      doc = Mix.Task.moduledoc(module) || "There is no documentation for this task"
+      {doc, where_is_file(module)}
+    end
+  end
 end
--- a/lib/mix/lib/mix/tasks/help.ex
+++ b/lib/mix/lib/mix/tasks/help.ex
@@ -189,9 +189,12 @@ defmodule Mix.Tasks.Help do
 
   defp build_alias_doc_list(aliases) do
     Enum.reduce(aliases, {[], 0}, fn {alias_name, task_name}, {docs, max} ->
-      doc = "Alias for " <> task_name
+      doc = "Alias for " <> task_str(task_name)
       task = "mix " <> alias_name
       {[{task, doc} | docs], max(byte_size(task), max)}
     end)
   end
+
+  defp task_str(task_name) when is_bitstring(task_name), do: task_name
+  defp task_str(task_name), do: inspect(task_name)
 end
--- a/lib/mix/lib/mix/tasks/help.ex
+++ b/lib/mix/lib/mix/tasks/help.ex
@@ -4,7 +4,7 @@ defmodule Mix.Tasks.Help do
   @shortdoc "Prints help information for tasks"
 
   @moduledoc """
-  Lists all aliases and tasks or prints the documentation for a given task.
+  Lists all tasks and aliases or prints the documentation for a given task or alias.
 
   ## Arguments
 
--- a/lib/mix/lib/mix/tasks/help.ex
+++ b/lib/mix/lib/mix/tasks/help.ex
@@ -59,8 +59,9 @@ defmodule Mix.Tasks.Help do
     tasks = Enum.map(load_tasks(), &Mix.Task.task_name/1)
 
     aliases =
-      Mix.Project.config()[:aliases]
-      |> Enum.map(fn {k, _} -> Atom.to_string(k) end)
+      Enum.map(Mix.Project.config()[:aliases], fn {alias_name, _} ->
+        Atom.to_string(alias_name)
+      end)
 
     for info <- Enum.sort(aliases ++ tasks) do
       Mix.shell().info(info)
@@ -70,13 +71,8 @@ defmodule Mix.Tasks.Help do
   def run(["--search", pattern]) do
     loadpaths!()
 
-    modules =
-      load_tasks()
-      |> Enum.filter(&String.contains?(Mix.Task.task_name(&1), pattern))
-
-    aliases =
-      load_aliases()
-      |> Enum.filter(&String.contains?(elem(&1, 0), pattern))
+    modules = Enum.filter(load_tasks(), &String.contains?(Mix.Task.task_name(&1), pattern))
+    aliases = Enum.filter(load_aliases(), &String.contains?(elem(&1, 0), pattern))
 
     {docs, max} = build_doc_list(modules, aliases)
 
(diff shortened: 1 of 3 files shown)


Sihui:
Not sure why the latest build failed: https://travis-ci.org/elixir-lang/elixir/jobs/397193995.
The second latest build also failed but on a different test: https://travis-ci.org/elixir-lang/elixir/jobs/397182429. 😞 Does the new code introcude something flaky? 😨

smorin:
I was just wanting this exact functionality.  Thanks for doing the pull request.

michalmuskala:
The build failure seems unrelated. I rested the failed suite.

lib/mix/lib/mix/tasks/help.ex
```diff
@@ -169,4 +186,20 @@ defmodule Mix.Tasks.Help do
...
+    Enum.reduce(aliases, {[], 0}, fn {alias_name, task_name}, {docs, max} ->
+      doc = "Alias for " <> task_str(task_name)
+      task = "mix " <> alias_name
+      {[{task, doc} | docs], max(byte_size(task), max)}
+    end)
+  end
+
+  defp task_str(task_name) when is_list(task_name) do
+    task_strs = Enum.map(task_name, &task_str/1) |> Enum.join(", ")
+    "[" <> task_strs <> "]"
```
michalmuskala: I wonder about the format here with the list, but I'm not really sure what would be better.

josevalim: I would say we should just call `inspect`. It is the simplest implementation and it would be similar to what we have today. But I am not sure.

I guess the formatting was the biggest concern for including aliases in the `mix help`, as they can get quite verbose. Another option is to just have a simple text: "# Alias defined in mix.exs" and call it a day.

Thoughts?

fertapric:
I think in order for this feature to be complete, this PR should support`mix help ALIAS`. It might be confusing to be able to run `mix help TASK` on some of the names listed but not on others.

This might also simplify the formatting of the aliases in `mix help`. As @josevalim suggested, `mix help` can just display `my_alias    # Alias defined in mix.exs`; and `mix help my_alias` could display a more verbose description:

```
Alias for ["run my/really/long/path.exs", "cmd another_long_cmd"]
```

Thoughts?

michalmuskala:
I like this solution.

smorin:
I like: `my_alias # Alias defined in mix.exs` as a solution but why not also allow.

```
defp aliases do
    [
      help_make: [alias: "cmd make", description: "my description", new_option: "asdf"]
    ]
  end
```
OR
```
defp aliases do
    [
      help_make: %{alias: "cmd make", description: "my description", new_option: "asdf"}
    ]
  end
```

josevalim:
@smorin If you want to add descriptions and what not, then I would say you should just go ahead and define a proper task.

smorin:
Why force someone to write a task for something that takes 3 words.  It's using a sledge hammer to drive a nail.  Especially in cases, where the command is 2 or 3 separate words and the description is the longest part.  Why clutter a code base, adding steps for a developer, and more lines of code to maintain.

npm does it will good effect: https://docs.npmjs.com/misc/scripts

lib/mix/lib/mix/tasks/help.ex
```diff
@@ -71,9 +73,13 @@ defmodule Mix.Tasks.Help do
       load_tasks()
       |> Enum.filter(&String.contains?(Mix.Task.task_name(&1), pattern))
 
-    {docs, max} = build_task_doc_list(modules)
+    aliases =
+      load_aliases()
+      |> Enum.filter(&String.contains?(elem(&1, 0), pattern))
```
whatyouhide: We try to avoid single pipes. I would suggest to refactor this with a `for` comprehension:

```elixir
aliases = 
  for {alias_name, _} = alias <- load_aliases(),
      String.contains?(alias_name, pattern),
      do: alias
```

josevalim:
> @josevalim Why force someone to write a task for something that takes 3 words. It's using a sledge hammer to drive a nail.

I see your point. We should move this discussion elsewhere then. Maybe a proposal to the elixir-lang-core mailing list as the scope of this particular issue is well defined based on how aliases work today and we should not go off-topic.

lib/mix/lib/mix/tasks/help.ex
Sihui: why do we try to avoid single pipes?

lib/mix/test/mix/tasks/help_test.exs
```diff
@@ -43,6 +57,64 @@ defmodule Mix.Tasks.HelpTest do
...
+
+      output =
+        capture_io(fn ->
+          Mix.Tasks.Help.run(["nested.h"])
+        end)
+
+      assert output =~ "# mix nested.h\n\n"
+
+      assert output =~
+               ~r/Alias for \[#Function<.*\/1 in .*ComplexAliases.project\/0>, \"h foo bar\"\]\n/
```
Sihui: Couldn't find an easy way to format the case for `"nested.h": [&Mix.shell().info(inspect(&1)), "h foo bar"]`.
Thoughts?

Sihui:
^^ Updated to: 
1. use `my_alias # Alias defined in mix.exs` in `mix help` 
2. and add support for `mix help ALIAS`

lib/mix/lib/mix/tasks/help.ex
```diff
@@ -116,6 +122,11 @@ defmodule Mix.Tasks.Help do
     |> Enum.filter(&(Mix.Task.moduledoc(&1) != false))
   end
 
+  defp load_aliases() do
+    Mix.Project.config()[:aliases]
+    |> Enum.map(fn {k, v} -> {Atom.to_string(k), v} end)
```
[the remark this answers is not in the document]
Sihui: updated to 
```elixir
  defp load_aliases() do
    Mix.Project.config()[:aliases]
    |> Enum.map(fn {k, v} -> {Atom.to_string(k), v} end)
    |> Map.new()
  end
```
to make `verbose_doc` easier. (turning it to a map also makes more sense.)

lib/mix/lib/mix/tasks/help.ex
```diff
@@ -4,11 +4,12 @@ defmodule Mix.Tasks.Help do
   @shortdoc "Prints help information for tasks"
 
   @moduledoc """
-  Lists all tasks or prints the documentation for a given task.
+  Lists all aliases and tasks or prints the documentation for a given task.
 
   ## Arguments
 
-      mix help                  - prints all tasks and their shortdoc
+      mix help                  - prints all aliases, tasks and task shortdoc
```
fertapric: I think there is a typo here "prints all aliases, tasks and **their** shortdoc"

lib/mix/lib/mix/tasks/help.ex
```diff
@@ -4,11 +4,12 @@ defmodule Mix.Tasks.Help do
   @shortdoc "Prints help information for tasks"
 
   @moduledoc """
-  Lists all tasks or prints the documentation for a given task.
+  Lists all aliases and tasks or prints the documentation for a given task.
```
fertapric: Minor nitpick: I would change the order of "tasks" and "aliases" here (and below). For some reason, in my head, it makes sense to put "tasks" at the beginning as they are more relevant for this specific task :) So, "Lists all tasks and aliases or prints the documentation for a given task or alias"

lib/mix/test/mix/tasks/help_test.exs
```diff
@@ -29,6 +29,20 @@ defmodule Mix.Tasks.HelpTest do
     end
   end
 
+  test "help list alias task", context do
```
fertapric: Minor nitpick: I think it should be "lists", and maybe "help lists all aliases"?

lib/mix/lib/mix/tasks/help.ex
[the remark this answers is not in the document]
fertapric: What do you think about these names to address the single-letter variables?

```elixir
defp load_aliases() do
  Mix.Project.config()[:aliases]
  |> Enum.map(fn {alias_name, alias_tasks} -> {Atom.to_string(alias_name), alias_tasks} end)
  |> Map.new()
end
```

lib/mix/lib/mix/tasks/help.ex
josevalim: @Sihui mostly because if we don't have multiple expressions, the pipeline ends up being more noise than helping readability. We know it is subjective but that's why we decided to add a "rule", so it is a clear path and we remove subjectivity. :)

lib/mix/test/mix/tasks/help_test.exs
josevalim: @Sihui this is honestly good enough to me. :)

lib/mix/test/mix/tasks/help_test.exs
```diff
@@ -70,6 +142,13 @@ defmodule Mix.Tasks.HelpTest do
       assert_received {:mix_shell, :info, ["mix deps" <> _]}
       assert_received {:mix_shell, :info, ["mix deps.clean" <> _]}
     end)
+
+    Mix.Project.push(Aliases)
+
+    in_tmp(context.test, fn ->
+      Mix.Tasks.Help.run(["--search", "c"])
+      assert_received {:mix_shell, :info, ["mix c" <> _]}
```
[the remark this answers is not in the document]
josevalim: @fertapric good catch!

lib/mix/lib/mix/tasks/help.ex
Sihui: I don't think we have shortdoc for aliases. shortdoc are marked with `@shortdoc`

lib/mix/lib/mix/tasks/help.ex
```diff
@@ -4,11 +4,12 @@ defmodule Mix.Tasks.Help do
...
 
   ## Arguments
 
-      mix help                  - prints all tasks and their shortdoc
+      mix help                  - prints all aliases, tasks and task shortdoc
+      mix help ALIAS            - prints the definition for the given alias
       mix help TASK             - prints full docs for the given task
       mix help --search PATTERN - prints all tasks that contain PATTERN in the name
```
fertapric: "prints all tasks and aliases that contain PATTERN in the name"?

lib/mix/lib/mix/tasks/help.ex
fertapric: "and their summary" or "and their short description"?

lib/mix/lib/mix/tasks/help.ex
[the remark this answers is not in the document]
whatyouhide: You're doing two passes this way. I didn't know you needed a map but if you do, I'd much rather go with `into: %{}` in the `for` comprehension or just `Map.new/2`:

```elixir
aliases = Mix.Project.config()[:aliases]
Map.new(aliases, fn {alias_name, alias_tasks} -> {Atom.to_string(alias_name), alias_tasks} end)
```

(also, definitely let's avoid single-letter variables!)

lib/mix/lib/mix/tasks/help.ex
Sihui: A (weak) counter argument will be using single pipe makes adding additional pipe easier. But that's for another discussion.

I don't think we should avoid `Enum.filter` b/c of avoiding single pipes though. I think `Enum.filter` expresses our intention clearly than the `for` comprehension. How about 

```elixir
Enum.filter(load_aliases, &String.contains?(elem(&1, 0), pattern))
```

A side note, we have quite a few usages of single pipes throughout the file (and the codebase), does it mean we want to replace all of them?

lib/mix/lib/mix/tasks/help.ex
Sihui: good catch!

lib/mix/lib/mix/tasks/help.ex
whatyouhide: Eventually we want to replace all pipes, but we try to avoid small changes like that in the codebase in order to keep the number of changes down. `Enum.filter/2` is also very much fine with me, I just don't love `elem/2` since it hides what the tuple looks like. What about:

```elixir
Enum.filter(load_aliases(), fn {name, _} -> String.contains?(name, pattern) end)
```

lib/mix/test/mix/tasks/help_test.exs
```diff
@@ -70,6 +142,13 @@ defmodule Mix.Tasks.HelpTest do
       assert_received {:mix_shell, :info, ["mix deps" <> _]}
       assert_received {:mix_shell, :info, ["mix deps.clean" <> _]}
     end)
+
+    Mix.Project.push(Aliases)
+
+    in_tmp(context.test, fn ->
+      Mix.Tasks.Help.run(["--search", "h"])
+      assert_received {:mix_shell, :info, ["mix h" <> _]}
```
fertapric: Unfortunately, this might also give false positives with `mix help` itself, `mix hex` or any `mix hex.*`.

fertapric: I would either include a space, or match on the message live above:

```
assert message =~ ~r/# Alias defined in mix.exs/
```

lib/mix/lib/mix/tasks/help.ex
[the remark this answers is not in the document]
Sihui: @whatyouhide I used map b/c I couldn't find an easy way to accomplish `Map.has_key?(aliases, task)`. If there's a way without having to use map, I would love to learn that! :)

lib/mix/lib/mix/tasks/help.ex
Sihui: good call! I not a fan of `elem` either.

smorin:
@josevalim makes sense moving to core-mailing list.
> I would say that ad-hoc aliases would end-up being more clutter than well-structured tasks but I see your point. We should move the discussion about revamping aliases elsewhere then. Maybe a proposal to the elixir-lang-core mailing list as the scope of this particular issue is well defined based on how aliases work today and we should not go off-topic.

Sihui:
@josevalim @michalmuskala @whatyouhide Can this be merged? Anything else I need to do?

josevalim:
We need two reviews to merge, so we are waiting on the second one. :) Once
somebody has some time available, they will review it and likely merge.
-- 


lib/mix/lib/mix/tasks/help.ex
```diff
@@ -56,8 +59,9 @@ defmodule Mix.Tasks.Help do
     tasks = Enum.map(load_tasks(), &Mix.Task.task_name/1)
 
     aliases =
-      Mix.Project.config()[:aliases]
-      |> Enum.map(fn {k, _} -> Atom.to_string(k) end)
+      Enum.map(Mix.Project.config()[:aliases], fn {alias_name, _} ->
+        Atom.to_string(alias_name)
+      end)
```
whatyouhide: Why not use `aliases = load_aliases()` here too?

whatyouhide: I mean that and then `Enum.sort(Map.to_list(aliases) ++ tasks)`.

Sihui: `Map.to_list(aliases)` won't work because aliases is `[{alias1, taks1}, {alias2, task2}, ...]`.
If we want to use `load_aliases()`, we might need to do

```elixir
    aliases = Enum.map(load_aliases(), fn {alias_name, _} -> alias_name end)

    for info <- Enum.sort(aliases ++ tasks) do
      Mix.shell().info(info)
    end
```

At that point, we might as well use `Mix.Project.config()[:aliases]` directly, so we don't have to iterate twice.

whatyouhide: Oh I missed that it was just the alias name, sorry!

whatyouhide:
Thanks @Sihui! 💟

Sihui:
Thank you all for reviewing! ❤️ ❤️ ❤️

Issue #7270, opened 2018-01-27
Code.string_to_quoted raises exception instead of returning error


schnittchen:
### Environment

* Elixir & Erlang versions (elixir --version):
```
Erlang/OTP 20 [erts-9.2] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:10] [hipe] [kernel-poll:false]

Elixir 1.6.0 (compiled with OTP 20)
```
* Operating system: 
Arch Linux

### Current behavior

```
iex(10)> "\"\\x\"" |> Code.string_to_quoted
** (ArgumentError) missing hex sequence after \x, expected \xHH
    (elixir) src/elixir_interpolation.erl:155: :elixir_interpolation.unescape_hex/3
    (elixir) src/elixir_interpolation.erl:81: :elixir_interpolation."-unescape_tokens/2-lc$^0/1-0-"/2
    (elixir) src/elixir_tokenizer.erl:580: :elixir_tokenizer.handle_strings/6
    (elixir) lib/code.ex:568: Code.string_to_quoted/2
```

### Expected behavior

Other syntax errors give a result of `{:error, _}`

schnittchen:
Very similarly:

```
** (ArgumentError) invalid Unicode sequence after \u, expected \uHHHH or \u{H*}
    (elixir) src/elixir_interpolation.erl:182: :elixir_interpolation.unescape_unicode/3
    (elixir) src/elixir_interpolation.erl:81: :elixir_interpolation."-unescape_tokens/2-lc$^0/1-0-"/2
    (elixir) src/elixir_tokenizer.erl:580: :elixir_tokenizer.handle_strings/6
    (elixir) lib/code.ex:568: Code.string_to_quoted/2
```

schnittchen:
I'm fuzzing code with StreamData and rescued those two cases. It looks like there is nothing more coming.

schnittchen:
Here is another one:

```
iex(23)> ".\"\#{}\"" |> Code.string_to_quoted 
** (CaseClauseError) no case clause matching: {1, 7, [{{1, 3, 1}, []}], []}
    (elixir) src/elixir_tokenizer.erl:636: :elixir_tokenizer.handle_dot/6
    (elixir) lib/code.ex:568: Code.string_to_quoted/2
```

schnittchen:
[...]

PR #7809, opened 2018-06-27
Move `Code.string_to_quoted` error logic up

drincruz:
As discussed in #7270, it would help if we moved up the error logic out of Erlang. So, instead of raising an exception, we'll return an error tuple.
Currently, this will only fix the Hex error use case, so does not completely resolve this issue. But, if this is the route we want to take, I can work on the other usecases.

### DIFF
--- a/lib/elixir/lib/code.ex
+++ b/lib/elixir/lib/code.ex
@@ -661,8 +661,12 @@ defmodule Code do
     file = Keyword.get(opts, :file, "nofile")
     line = Keyword.get(opts, :line, 1)
 
-    with {:ok, tokens} <- :elixir.string_to_tokens(to_charlist(string), line, file, opts) do
-      :elixir.tokens_to_quoted(tokens, file, opts)
+    case :elixir.string_to_tokens(to_charlist(string), line, file, opts) do
+      {:ok, tokens} ->
+        :elixir.tokens_to_quoted(tokens, file, opts)
+
+      {:error, _error_msg} = error ->
+        error
     end
   end
 
--- a/lib/elixir/src/elixir_interpolation.erl
+++ b/lib/elixir/src/elixir_interpolation.erl
@@ -78,7 +78,13 @@ unescape_tokens(Tokens) ->
   unescape_tokens(Tokens, fun unescape_map/1).
 
 unescape_tokens(Tokens, Map) ->
-  [unescape_token(Token, Map) || Token <- Tokens].
+  try [unescape_token(Token, Map) || Token <- Tokens] of
+    Unescaped ->
+      {ok, Unescaped}
+  catch
+    {error, _Reason} = Error ->
+      Error
+  end.
 
 unescape_token(Token, Map) when is_binary(Token) -> unescape_chars(Token, Map);
 unescape_token(Other, _Map) -> Other.
@@ -151,8 +157,7 @@ unescape_hex(<<${, A, B, C, D, E, F, $}, Rest/binary>>, Map, Acc) when ?is_hex(A
   append_codepoint(Rest, Map, [A, B, C, D, E, F], Acc, 16);
 
 unescape_hex(<<_/binary>>, _Map, _Acc) ->
-  Msg = <<"missing hex sequence after \\x, expected \\xHH">>,
-  error('Elixir.ArgumentError':exception([{message, Msg}])).
+  throw({error, "missing hex sequence after \\x, expected \\xHH"}).
 
 %% Finish deprecated sequences
 
--- a/lib/elixir/src/elixir_tokenizer.erl
+++ b/lib/elixir/src/elixir_tokenizer.erl
@@ -609,8 +609,13 @@ handle_strings(T, Line, Column, H, Scope, Tokens) ->
       Token = {Key, {Line, Column - 1, nil}, Unescaped},
       tokenize(Rest, NewLine, NewColumn + 1, Scope, [Token | Tokens]);
     {NewLine, NewColumn, Parts, Rest} ->
-      Token = {string_type(H), {Line, Column - 1, nil}, unescape_tokens(Parts, Scope)},
-      tokenize(Rest, NewLine, NewColumn, Scope, [Token | Tokens])
+      case unescape_tokens(Parts, Scope) of
+        {error, Msg} ->
+          {error, {Line, Column, Msg, [H]}, Rest, Tokens};
+        Unescaped ->
+          Token = {string_type(H), {Line, Column - 1, nil}, Unescaped},
+          tokenize(Rest, NewLine, NewColumn, Scope, [Token | Tokens])
+      end
   end.
 
 handle_unary_op([$: | Rest], Line, Column, _Kind, Length, Op, Scope, Tokens) when ?is_space(hd(Rest)) ->
@@ -876,7 +881,12 @@ extract_heredoc_line(Marker, Rest, Buffer, _Counter) ->
   extract_heredoc_line(Marker, Rest, Buffer).
 
 unescape_tokens(Tokens, #elixir_tokenizer{unescape=true}) ->
-  elixir_interpolation:unescape_tokens(Tokens);
+  case elixir_interpolation:unescape_tokens(Tokens) of
+    {ok, UnescapedTokens} ->
+      UnescapedTokens;
+    {error, _Reason} = Error ->
+      Error
+    end;
 unescape_tokens(Tokens, #elixir_tokenizer{unescape=false}) ->
   Tokens.
 
--- a/lib/elixir/test/elixir/code_test.exs
+++ b/lib/elixir/test/elixir/code_test.exs
@@ -118,6 +118,11 @@ defmodule CodeTest do
       assert string_to_quoted.("foo + bar") == {:ok, {:+, [line: 1, column: 5], [foo, bar]}}
     end
 
+    test "returns an error tuple on hex errors" do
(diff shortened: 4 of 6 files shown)


lib/elixir/lib/code.ex
```diff
@@ -661,8 +661,12 @@ defmodule Code do
     file = Keyword.get(opts, :file, "nofile")
     line = Keyword.get(opts, :line, 1)
 
-    with {:ok, tokens} <- :elixir.string_to_tokens(to_charlist(string), line, file, opts) do
-      :elixir.tokens_to_quoted(tokens, file, opts)
+    case :elixir.string_to_tokens(to_charlist(string), line, file, opts) do
```
josevalim: Just a note: the code before and after are equivalent. However, your version is clearer, so let's stick with it. :)

lib/elixir/src/elixir_interpolation.erl
```diff
@@ -78,9 +78,24 @@ unescape_tokens(Tokens) ->
   unescape_tokens(Tokens, fun unescape_map/1).
 
 unescape_tokens(Tokens, Map) ->
-  [unescape_token(Token, Map) || Token <- Tokens].
```
josevalim: This is one of the few cases where I think a `throw` would be clearer. So something like this:

```erlang
unescape_tokens(Tokens, Map) ->
  try [unescape_token(Token, Map) || Token <- Tokens] of
    Unescaped -> {ok, Unescaped}
  catch
    {error, _Msg} = Error -> Error
  end.
```

This way you can keep the implementation the same, all you need to do is to replace the ArgumentError by a throw.

lib/elixir/src/elixir_tokenizer.erl
```diff
@@ -609,8 +609,13 @@ handle_strings(T, Line, Column, H, Scope, Tokens) ->
       Token = {Key, {Line, Column - 1, nil}, Unescaped},
       tokenize(Rest, NewLine, NewColumn + 1, Scope, [Token | Tokens]);
     {NewLine, NewColumn, Parts, Rest} ->
-      Token = {string_type(H), {Line, Column - 1, nil}, unescape_tokens(Parts, Scope)},
-      tokenize(Rest, NewLine, NewColumn, Scope, [Token | Tokens])
+      UnescapedToken = unescape_tokens(Parts, Scope),
+      case UnescapedToken of
+        {error, _Msg} = Error -> Error;
```
josevalim: I don't think we should introduce a new error tuple which is {error, String}. Let's change this line so it returns: `{error, {Line, Column, Message, string_type(H)}, Rest, Tokens}`. This way you won't have to change any of the calls to `:elixir.string_to_token` and ultimately the code will be simpler.

fertapric: Checking the coding style of this file, it seems right-hand clauses are only kept in the same line if all the clauses are kept as single lines. The same applies to the `unescaped_tokens` function in `lib/elixir/src/elixir_interpolation.erl`

lib/elixir/test/elixir/code_test.exs
```diff
@@ -118,6 +118,11 @@ defmodule CodeTest do
       assert string_to_quoted.("foo + bar") == {:ok, {:+, [line: 1, column: 5], [foo, bar]}}
     end
 
+    test "returns an error tuple on hex errors" do
+      assert Code.string_to_quoted("\"\\x\"") ==
+               {:error, "missing hex sequence after \\x, expected \\xHH"}
```
josevalim: This is changing the contract of `Code.string_to_quoted` which returns errors in the format `{error, Tuple}`. So let's proceed as [described here](https://github.com/elixir-lang/elixir/pull/7809/files#r198758547) so we can keep all the error formats the same!

lib/elixir/src/elixir_tokenizer.erl
drincruz: thanks! I'm a bit unfamiliar with the style guide, so I trust your judgement! I'll make updates!

lib/elixir/src/elixir_interpolation.erl
```diff
@@ -151,8 +165,7 @@ unescape_hex(<<${, A, B, C, D, E, F, $}, Rest/binary>>, Map, Acc) when ?is_hex(A
   append_codepoint(Rest, Map, [A, B, C, D, E, F], Acc, 16);
 
 unescape_hex(<<_/binary>>, _Map, _Acc) ->
-  Msg = <<"missing hex sequence after \\x, expected \\xHH">>,
-  error('Elixir.ArgumentError':exception([{message, Msg}])).
+  throw(missing_hex_sequence).
```
drincruz: I was not sure what we would want to throw here and am not sure how we all feel about just throwing `missing_hex_sequence`. ❔

josevalim: This is totally fine but you can also do `throw({error, Msg})` and you catch and return the tuple later on.

drincruz: Brilliant! That sounds like the route we'd want. Let me update! 👍

lib/elixir/src/elixir_interpolation.erl
```diff
@@ -78,9 +78,23 @@ unescape_tokens(Tokens) ->
   unescape_tokens(Tokens, fun unescape_map/1).
 
 unescape_tokens(Tokens, Map) ->
-  [unescape_token(Token, Map) || Token <- Tokens].
+  try [unescape_token(Token, Map) || Token <- Tokens] of
```
josevalim: We should need a try/catch  only on `unescape_tokens` as `unescape_token` itself is private.

lib/elixir/src/elixir.erl
```diff
@@ -348,6 +348,7 @@ parser_line(Meta) ->
   end.
 
 to_binary(List) when is_list(List) -> elixir_utils:characters_to_binary(List);
+to_binary(Binary) when is_binary(Binary) -> Binary;
```
josevalim: Why is this change necessary?

lib/elixir/src/elixir_parser.yrl
```diff
@@ -903,6 +903,8 @@ string_parts(Parts) ->
   [string_part(Part) || Part <- Parts].
 string_part(Binary) when is_binary(Binary) ->
   Binary;
+string_part({error, ErrorMsg}) ->
+  {error, ErrorMsg};
```
josevalim: This should no longer be necessary either.

lib/elixir/src/elixir_tokenizer.erl
```diff
@@ -609,8 +609,13 @@ handle_strings(T, Line, Column, H, Scope, Tokens) ->
       Token = {Key, {Line, Column - 1, nil}, Unescaped},
       tokenize(Rest, NewLine, NewColumn + 1, Scope, [Token | Tokens]);
     {NewLine, NewColumn, Parts, Rest} ->
-      Token = {string_type(H), {Line, Column - 1, nil}, unescape_tokens(Parts, Scope)},
-      tokenize(Rest, NewLine, NewColumn, Scope, [Token | Tokens])
+      Unescaped = unescape_tokens(Parts, Scope),
```
josevalim: Let's return `{ok, Unescaped} | {error, Msg}`

lib/elixir/src/elixir_tokenizer.erl
```diff
@@ -609,8 +609,13 @@ handle_strings(T, Line, Column, H, Scope, Tokens) ->
...
+      case Unescaped of
+        {error, Msg} -> {error, {Line, Column, Msg, string_type(H)}, Rest, Tokens};
```
josevalim: Sorry, I gave a bad advice, the fourth element should be `[H]`.

lib/elixir/src/elixir.erl
drincruz: oh I was getting an error since we are passing a binary from the error message [here](https://github.com/elixir-lang/elixir/pull/7809/files#diff-8e79b6aba4faa9923e1ed2defe97e290R614).

lib/elixir/src/elixir_interpolation.erl
drincruz: @josevalim Oh! thank you for bearing with me on all of these erlang changes! 😬 i'll update.

lib/elixir/src/elixir.erl
josevalim: I see. It is the same "principle" as before then. We should not introduce new "formats" for returning errors. We already have a well defined tuple, we should stick to it. :)

lib/elixir/src/elixir_interpolation.erl
```diff
@@ -78,7 +78,13 @@ unescape_tokens(Tokens) ->
...
+    Unescaped ->
+      {ok, Unescaped}
```
drincruz: so, with this change to return an `ok | error` tuple, i needed to update Kernel and Macro slightly to respect this change as well.

josevalim: Yes, exactly. You should also change the macro ones to raise on error, as otherwise it would be a breaking change. So it would be something like:

```elixir
case :elixir_interpolation.unescape_tokens(...) do
  {:ok, value} -> value
  {:error, reason} -> raise ArgumentError, to_string(reason)
end
```

drincruz: that makes sense. this is a small change, i can get this updated PR in soon. 👍

josevalim:

PR #7810, opened 2018-06-28
disambiguate def/2 docs regarding anonymous functions

davydog187:
I was reading through the docs, and I can see how `single` here can be misconstrued to mean functions of arity 1. I think that sole makes it clearer as referring to a sole arity instead of a single arity function.


josevalim:

PR #7812, opened 2018-06-28
Updated docs für GenServer.terminate

h4cc:
Tried to make the docs more clear that manual cleanup is not needed for opened sockets and files.


michalmuskala:
Thank you! :heart:

PR #7815, opened 2018-06-28
Replace single pipes in lib/mix/mix/tasks/.

Sihui:
Per conversation in https://github.com/elixir-lang/elixir/pull/7805#discussion_r198515317
> We try to avoid single pipes.

Had a small chunck of free time so I took a stab on this to see how it feels.

### DIFF
--- a/lib/mix/lib/mix/tasks/app.tree.ex
+++ b/lib/mix/lib/mix/tasks/app.tree.ex
@@ -49,7 +49,7 @@ defmodule Mix.Tasks.App.Tree do
           {String.to_atom(app), opts}
       end
 
-    excluded = Keyword.get_values(opts, :exclude) |> Enum.map(&String.to_atom/1)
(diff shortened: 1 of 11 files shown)


josevalim:
We are generally against commits that only change the code style. We fix those when we are working on those files but we avoid changing those all at once to avoid breaking existing pull requests and to avoid adding churn to the git history. In any case, thanks @Sihui !

whatyouhide:
I had commented that on that same comment thread btw

> Eventually we want to replace all pipes, but we try to avoid small changes like that in the codebase in order to keep the number of changes down.

Sorry if I wasn't clear!

Sihui:
@whatyouhide oh, that's what you meant. Sorry, for some reason I misread it and thought you meant something else. Why do we want to keep the number of changes down, btw? 
A side note, is [the other PR](https://github.com/elixir-lang/elixir/pull/7805) ready to be merged? Anything you want me to change?

PR #7816, opened 2018-06-29
Make `mix format -` more suitable for use in Vim by preserving surrounding whitespace

puyo:
I like formatting snippets of code in my editor by selecting them and running a command to format just that snippet. There are a few benefits to avoiding formatting an entire file:

- It can be faster
- It can minimise commit diffs / PRs when somebody else has committed unformatted code

But when I do this with `mix format -` (e.g. in Vim, highlight code and type `:!mix format -<CR>`), the selected code becomes unindented (it moves to column 0) and also loses any whitespace such as newlines at the start and the end of the code.

This PR measures and restores the leading/trailing whitespace and the indentation on the Elixir code only when run on stdio via the `mix format -`, making it suitable to run within Vim (or Spacemacs) via the `!` operator.

### DIFF
--- a/lib/mix/lib/mix/tasks/format.ex
+++ b/lib/mix/lib/mix/tasks/format.ex
@@ -394,7 +394,11 @@ defmodule Mix.Tasks.Format do
 
   defp format_file({file, formatter_opts}, task_opts) do
     {input, extra_opts} = read_file(file)
-    output = IO.iodata_to_binary([Code.format_string!(input, extra_opts ++ formatter_opts), ?\n])
+
+    output =
+      [Code.format_string!(input, extra_opts ++ formatter_opts), ?\n]
+      |> IO.iodata_to_binary()
+      |> restore_whitespace(input, file)
 
     check_equivalent? = Keyword.get(task_opts, :check_equivalent, false)
     check_formatted? = Keyword.get(task_opts, :check_formatted, false)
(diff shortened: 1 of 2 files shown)


josevalim:
Thanks @puyo but I don't think this should always be performed on `mix format -`. For example, I may have copied an inner module to a new buffer and in that moment I want the indentation to be removed on formatting. We could instead support a function such as `--keep-indentation` but it feels like this should be implemented by scripts that run inside a text editor rather than on Elixir itself. Thank you.

OvermindDL1:
At least in atom here it runs it through the mix formatter but then re-indents all non-blank lines to the indentation of the first selected line, so it 'just works', perhaps vim should do the same?

sorentwo:
Appending the re-select command `gv` and then the indent `=` command may achieve the same results.

puyo:
> Appending the re-select command gv and then the indent = command may achieve the same results.

Doesn't seem to work for me, since = on a region invokes Vim's own set of rules for indenting Elixir code which is different to `mix format`. Same story for Spacemacs and presumably Emacs.

> At least in atom here it runs it through the mix formatter but then re-indents all non-blank lines to the indentation of the first selected line, so it 'just works', perhaps vim should do the same?

Sure, it can, I just figured since Elixir had taken on the job of formatting code on behalf of code editors, that such functionality that was common across editors should be written in one place rather than into each editor in different languages. I already have a Ruby wrapper around `mix format -` that post-processes the result correctly [1], so it's no trouble for me if this PR isn't merged, I just wanted to contribute code that I thought would be useful to others.

[1] https://github.com/puyo/config/blob/master/bin/exfmt

PR #7818, opened 2018-07-01
Add a --force flag to mix archive.uninstall

AnilRedshift:
Mix archive.install contains a `--force` flag to allow silent installation/upgrades. However, the same is not true for uninstall.

This PR copies the same pattern used in install to add a similar force flag to uninstall.

### DIFF
--- a/lib/mix/lib/mix/local/installer.ex
+++ b/lib/mix/lib/mix/local/installer.ex
@@ -236,9 +236,9 @@ defmodule Mix.Local.Installer do
   @doc """
   A common implementation for uninstalling archives and scripts.
   """
-  @spec uninstall(Path.t(), String.t(), OptionParser.argv()) :: Path.t() | nil
-  def uninstall(root, listing, argv) do
-    {_, argv, _} = OptionParser.parse(argv, switches: [])
+  @spec uninstall(Path.t(), String.t(), OptionParser.argv(), keyword) :: Path.t() | nil
+  def uninstall(root, listing, argv, switches) do
+    {opts, argv} = OptionParser.parse!(argv, switches: switches)
 
     if name = List.first(argv) do
       path = Path.join(root, name)
@@ -249,7 +249,7 @@ defmodule Mix.Local.Installer do
           Mix.Task.rerun(listing)
           nil
 
-        should_uninstall?(path) ->
+        should_uninstall?(path, opts) ->
           File.rm_rf!(path)
           path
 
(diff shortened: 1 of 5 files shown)


AnilRedshift:
I'm not immediately sure why the tests are failing. They work on my local machine and the failing test looks unrelated: `2) test compiles a project with mixfile (Mix.Tasks.CompileTest)`

lib/mix/lib/mix/tasks/escript.uninstall.ex
```diff
@@ -10,8 +10,11 @@ defmodule Mix.Tasks.Escript.Uninstall do
 
   """
 
+  @switches []
```
josevalim: Let's support `force: :boolean` here too.

lib/mix/lib/mix/tasks/archive.uninstall.ex
```diff
@@ -8,8 +8,18 @@ defmodule Mix.Tasks.Archive.Uninstall do
 
       mix archive.uninstall archive.ez
 
+  ## Command line options
+
+    * `--sha512` - checks the archive matches the given SHA-512 checksum. Only
+      applies to installations via URL or local path
```
josevalim: This option does not apply here, does it?

lib/mix/lib/mix/local/installer.ex
```diff
@@ -236,9 +236,9 @@ defmodule Mix.Local.Installer do
   @doc """
   A common implementation for uninstalling archives and scripts.
   """
-  @spec uninstall(Path.t(), String.t(), OptionParser.argv()) :: Path.t() | nil
-  def uninstall(root, listing, argv) do
-    {_, argv, _} = OptionParser.parse(argv, switches: [])
+  @spec uninstall(Path.t(), String.t(), OptionParser.argv(), keyword) :: Path.t() | nil
+  def uninstall(root, listing, argv, switches) do
```
josevalim: We need to declare it as `switches \\ []`, otherwise it is a breaking change.

lib/mix/lib/mix/tasks/archive.uninstall.ex
```diff
@@ -8,8 +8,18 @@ defmodule Mix.Tasks.Archive.Uninstall do
...
+
+    * `--force` - forces uninstallation without a shell prompt; primarily
+      intended for automation
   """
+  @switches [
+    force: :boolean
+  ]
```
whatyouhide: Nitpick, but I would leave a blank line before and after the `@switches` definition.

lib/mix/lib/mix/tasks/archive.uninstall.ex
AnilRedshift: Bad copy/paste. Thanks!

lib/mix/lib/mix/tasks/archive.uninstall.ex
AnilRedshift: Added

lib/mix/lib/mix/tasks/escript.uninstall.ex
AnilRedshift: Added.

PR #7821, opened 2018-07-01
Beginner-friendly revisions to Elixir's documentation

English3000:
Where better to get feedback from than the source? As a relatively new user of Elixir (since v1.6) with a background in teaching methodology, here are a few suggestions to make the Hex documentation more helpful and intuitive for new users.

For context, see: 

* https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/elixir-lang-core/SAejGg2e-io/kpA_3qVVCAAJ
* https://elixirforum.com/t/discussion-incorporating-erlang-otp-21-map-guards-in-elixir/14816/39

PLEASE NOTE: These edits are meant as a starting point to confirm that the design decisions for the documentation--based on their merits--are viewed by the Elixir Core Team as steps in the right direction.

When this is hopefully confirmed (if you are unconvinced that my revisions will be helpful, let's discuss your objections and get some feedback from beginners on my adjustments), I am happy to provide some deeper revisions in regard to the structure/organization of the core pages I have edited (or feel free to go ahead and implement the suggested comments in my respective commits yourself).

Thanks!

### DIFF
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -12,13 +12,14 @@ defmodule Kernel do
   
   It mainly consists of:
   
-    * syntactic sugar - for more readable and intuitive code
-    * [macros](https://elixir-lang.org/getting-started/meta/macros.html) - for defining new
-      functionality (modules, functions, and so on)
-    * [guard](guards.html) checks - for augmenting pattern matching
+    * basic language primitives, such as arithmetic operators, spawning of processes,
+      data type handling, etc
+    * macros for control-flow and defining new functionality (modules, functions, and so on)
+    * [guard](guards.html) checks for augmenting pattern matching
     
-  You can use `Kernel` functions/macros without the `Kernel` prefix anywhere in Elixir code
-  as all its functions and macros are automatically imported. For example, in IEx:
+  You can use `Kernel` functions/macros without the `Kernel` prefix anywhere in
+  Elixir code as all its functions and macros are automatically imported. For
+  example, in IEx:
 
       iex> is_number(13)
       true
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -23,10 +23,10 @@ defmodule Kernel do
       iex> is_number(13)
       true
       
-  If you don't want to import a function or macro from `Kernel`, use `except:` 
-  and then list the functionality by arity:
+  If you don't want to import a function or macro from `Kernel`, use the `:except` 
+  option and then list the function/macro by arity:
   
-        import Kernel, except: [if: 2]
+        import Kernel, except: [if: 2, unless: 2]
 
   Elixir also has special forms that are always imported and
   cannot be skipped. These are described in `Kernel.SpecialForms`.
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -8,32 +8,25 @@ import :elixir_bootstrap
 
 defmodule Kernel do
   @moduledoc """
-  This module is the entry point of the Elixir programming language.
-
-  It provides the default functions and macros Elixir imports into your
-  environment. Those can be directly invoked in your code without the
-  module name:
-
+  `Kernel` is Elixir's set of core functionality, mainly consisting of:
+  
+    * syntactic sugar - for more readable and intuitive code
+    * [macros](https://elixir-lang.org/getting-started/meta/macros.html) - for defining new functionality
+    * [guard](guards.html) checks - for more powerful function definitions
+    
+  `Kernel` is automatically imported into all `.ex` and `.exs` files,
+  meaning that you can use its functions and macros in any file or `IEx` 
+  without using the module name:
+  
       iex> is_number(13)
       true
-
-  These functions and macros can be skipped or cherry-picked via the
-  `Kernel.SpecialForms.import/2` macro. For instance, if you want to tell
-  Elixir not to import the `if/2` macro, you can do:
-
-      import Kernel, except: [if: 2]
+      
+  For optimizations, you can tell Elixir not to import a given function or macro:
+  
+        import Kernel, except: [if: 2]
 
   Elixir also has special forms that are always imported and
   cannot be skipped. These are described in `Kernel.SpecialForms`.
-
-  This module provides a variety of code definition, flow-control
-  and data type functionality. For example, new processes can be
-  created with `spawn/1`, modules can be defined with `defmodule/2`,
-  short-circuit operators are found in `&&/2` and `||/2`, numbers
-  can be added with `+/2`, and more. This module also contains all
-  built-in guards, which are a set of functions that augment pattern
-  matching with complex checks. More information about guards can be
-  found in the [Guards page](guards.html).
 
   ## The standard library
 
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -13,19 +13,22 @@ defmodule Kernel do
   It mainly consists of:
   
     * syntactic sugar - for more readable and intuitive code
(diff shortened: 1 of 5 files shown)


lib/elixir/lib/kernel.ex
```diff
@@ -8,33 +8,26 @@ import :elixir_bootstrap
 
 defmodule Kernel do
   @moduledoc """
-  This module is the entry point of the Elixir programming language.
-
-  It provides the default functions and macros Elixir imports into your
-  environment. Those can be directly invoked in your code without the
-  module name:
-
+  `Kernel` is Elixir's set of core functionality, mainly consisting of:
```
michalmuskala: The first line of moduledoc should be a short self-contained summary. It will be displayed in some module summaries in documentation and similar.

lib/elixir/lib/kernel.ex
```diff
@@ -8,33 +8,26 @@ import :elixir_bootstrap
...
       iex> is_number(13)
       true
-
-  These functions and macros can be skipped or cherry-picked via the
-  `Kernel.SpecialForms.import/2` macro. For instance, if you want to tell
-  Elixir not to import the `if/2` macro, you can do:
-
-      import Kernel, except: [if: 2]
+      
+  For optimizations, you can tell Elixir not to import a given function or macro:
```
michalmuskala: I'm not sure I would call it optimisations - the primary use case I found for this in the wild is defining a `send/2` function in your module that won'r conflict with `Kernel.send/2` - it usually has nothing to do with performance.

lib/elixir/lib/map.ex
```diff
@@ -1,6 +1,21 @@
 defmodule Map do
   @moduledoc """
   A set of functions for working with maps.
+  
+  Please note many functions for maps can be found in `Enum`.
```
michalmuskala: How about something like: "maps implement the `Enumerable` protocol and thus can be manipulated using functions from the `Enum` module.

lib/elixir/lib/tuple.ex
```diff
@@ -1,6 +1,32 @@
 defmodule Tuple do
   @moduledoc """
   Functions for working with tuples.
+  
+  Please note the following functions for tuples are found in `Kernel`:
+  
+    * `elem/2` - for accessing a tuple by index
+    * `put_elem/3`
+    * `tuple_size/3`
```
michalmuskala: Why only have the "shortdoc" for just one of them? I think we should either do that for all or for none.

English3000:
Just revised again.

lib/elixir/lib/kernel.ex
```diff
@@ -8,33 +8,29 @@ import :elixir_bootstrap
...
-
+  `Kernel` is Elixir's set of core functionality.
+  
+  It mainly consists of:
+  
+    * syntactic sugar - for more readable and intuitive code
+    * [macros](https://elixir-lang.org/getting-started/meta/macros.html) - for defining new functionality
+    * [guard](guards.html) checks - for more powerful function definitions
+    
+  `Kernel` is automatically imported into all `.ex` and `.exs` files,
```
whatyouhide: Should we say that `Kernel` is automatically imported in all Elixir code, instead of just `.ex` and `.exs` files?

lib/elixir/lib/list.ex
```diff
@@ -1,6 +1,16 @@
...
+  
+  Please note that most of the functions provided for lists--which implement 
+  the `Enumerable` protocol--are found in `Enum`.
+  
+  Additionally, the following functions for lists are found in `Kernel`:
+  
+    * `hd/1`
+    * `tl/1`
+    * `in/2` - equivalent to `Enum.member?/2`
+    * `length/1` - requires traversing the full list
```
whatyouhide: I don't think this is the place to mention that `length/1` requires traversing the whole list.

lib/elixir/lib/map.ex
```diff
@@ -1,6 +1,22 @@
...
+    * `get_and_update_in/2`
+    * `get_and_update_in/3`
+    * `get_in/2`
+    * `map_size/1`
+    * `pop_in/1`
+    * `pop_in/2`
+    * `put_in/2`
+    * `put_in/3`
+    * `update_in/2`
+    * `update_in/3`
```
whatyouhide: Mh, I am not sure I agree with listing all the `Access` calls that are in `Kernel` here. Those work on more than maps (for example, keyword lists).

lib/elixir/lib/tuple.ex
```diff
@@ -1,60 +1,24 @@
...
-  Tuples being stored contiguously in memory also means that updating a tuple
-  (for example replacing an element with `Kernel.put_elem/3`) will make a
-  shallow copy of the whole tuple. The tuple elements are still shared thanks
-  to immutability.
-
-  Tuples are not meant to be used as a "collection" type but rather as a
-  fixed-size container for multiple elements. That's why it is not possible
+  
+  Please note the following functions for tuples--_all constant-time_
+  --are found in `Kernel`:
```
whatyouhide: I don't think this is the place to mention that these operations are constant-time. Let the documentation for each of those function mention that.

English3000:
Ready for the next round of edits/feedback.

lib/elixir/lib/tuple.ex
```diff
@@ -1,60 +1,35 @@
...
-      3
-
-  Tuples being stored contiguously in memory also means that updating a tuple
-  (for example replacing an element with `Kernel.put_elem/3`) will make a
-  shallow copy of the whole tuple. The tuple elements are still shared thanks
-  to immutability.
-
-  Tuples are not meant to be used as a "collection" type but rather as a
-  fixed-size container for multiple elements. That's why it is not possible
-  to traverse a tuple dynamically using the functions in the `Enum` module.
```
whatyouhide: I think it's important we keep this information.

lib/elixir/lib/list.ex
```diff
@@ -2,14 +2,17 @@ defmodule List do
...
-  the `Enumerable` protocol--are found in `Enum`.
+  Many of the functions provided for lists, which implement 
+  the `Enumerable` protocol, are found in the `Enum` module.
   
-  Additionally, the following functions for lists are found in `Kernel`:
+  Additionally, the following functions and operators for lists are
+  found in `Kernel`:
   
+    * `Kernel.++/2`
+    * `Kernel.--/2`
```
eksperimental: `++/2` and `--/2`  should work with latest exdoc version

josevalim: I prefer the `Kernel` prefix though from outer modules for clarity.

josevalim:
Also please run `bin/elixir bin/mix format` in the project so the build can become green.

lib/elixir/lib/list.ex
eksperimental: So in/2 should use Kernel

English3000:
Cool, sounds like you all have a good sense of the direction you want to take my feedback.

I'll let you all execute it to your liking. When everything is good, I can run the build.

I appreciate your receptiveness to my feedback and am happy to discuss more offline if any of you would be interested.

P.S. One last thought: For the functions that work on a data structure but aren't included in its module, the information could be placed in the navigator's dropdown as its own tab, with links to the functions in their separate modules. That way, it could come after the data structure description but still be very visible to users.

lib/elixir/lib/kernel.ex
mariusbutuc: #TIL ❤️

eksperimental:
Can you please remove "Hex" from the title of the PR, it's kind of misleading

English3000:
All set!

josevalim:

lackac:
@English3000 yes, they will be published to https://hexdocs.pm/elixir/ with the next release. However, you can already see the changes if you go to the documentation of the `master` branch: https://hexdocs.pm/elixir/master/Integer.html

PR #7826, opened 2018-07-02
Adopt EEP 48

### LINKED ISSUE #7198: Adopt EEP 48
http://erlang.org/eep/eeps/eep-0048.html

The goal is to introduce `Code.fetch_docs/1` that returns the format in the EEP (or `{:error, _}` if the chunk is not available). Since the old documentation chunk will no longer exist, `Code.get_docs/2` should be changed to always return `nil` - which is a backwards compatible result - and be scheduled for deprecation.

We will also use the new chunk metadata to store relevant information, such as "defaults", "deprecated" and "since".

Replaces #3589.

lackac:
Related to #7198.

As discussed in the referenced issue the following are in scope for this PR:

- [x] extend `docs_chunk/6` in `elixir_erl.erl` to also generate the new `Docs` chunk according to EEP 48
- [x] read the new chunk in `Code.fetch_docs/1`
- [x] adjust `Protocol` and the `escript.build` mix task to include the new chunk

The following are to be implemented in separate PRs:

* migrate all functionality that uses `Code.get_docs/2` to use `Code.fetch_docs/1`
* remove the old docs chunk and return `nil` from `Code.get_docs/2`
* change how we store docs to be closer to the new docs format
* dependent functionalities (e.g. `@since` and `@deprecated`)

### DIFF
--- a/lib/elixir/lib/code.ex
+++ b/lib/elixir/lib/code.ex
@@ -1045,6 +1045,66 @@ defmodule Code do
   When given a path to a `.beam` file, it will load the docs directly from that
   file.
 
+  It returns the term stored in the documentation chunk in the format defined by
+  [EEP 48](http://erlang.org/eep/eeps/eep-0048.html) or `{:error, reason}` if
+  the chunk is not available.
+
+  ## Examples
+
+      # Module documentation of an existing module
+      iex> {:docs_v1, _, :elixir, _, %{"en" => module_doc}, _, _} = Code.fetch_docs(Atom)
+      iex> module_doc |> String.split("\n") |> Enum.at(0)
+      "Convenience functions for working with atoms."
+
+      # A module that doesn't exist
+      iex> Code.fetch_docs(ModuleNotGood)
+      {:error, :module_not_found}
+
+  """
+  @spec fetch_docs(module) ::
+          {:docs_v1, anno, beam_language, format, module_doc :: doc, metadata,
+           docs :: [{{kind, name, arity}, anno, signature, doc, metadata}]}
+          | {:error, :module_not_found | :docs_chunk_not_found}
+        when anno: :erl_anno.anno(),
+             beam_language: atom,
+             format: binary,
+             doc: %{binary => binary} | :none | :hidden,
+             kind: atom,
+             name: atom,
+             signature: [binary],
+             metadata: map
+  @since "1.7.0"
+  def fetch_docs(module) when is_atom(module) do
+    case :code.get_object_code(module) do
+      {_module, bin, _beam_path} -> do_fetch_docs(bin)
+      :error -> {:error, :module_not_found}
+    end
+  end
+
+  def fetch_docs(binpath) when is_binary(binpath) do
+    do_fetch_docs(String.to_charlist(binpath))
+  end
+
+  @docs_chunk 'Docs'
+
+  defp do_fetch_docs(bin_or_path) do
+    case :beam_lib.chunks(bin_or_path, [@docs_chunk]) do
+      {:ok, {_module, [{@docs_chunk, bin}]}} ->
+        :erlang.binary_to_term(bin)
+
+      {:error, :beam_lib, {:missing_chunk, _, @docs_chunk}} ->
+        {:error, :docs_chunk_not_found}
+    end
+  end
+
+  @doc ~S"""
+  Returns the docs for the given module.
+
+  When given a module name, it finds its BEAM code and reads the docs from it.
+
+  When given a path to a `.beam` file, it will load the docs directly from that
+  file.
+
   The return value depends on the `kind` value:
 
     * `:moduledoc` - tuple `{line, doc}` where `line` is the line on
@@ -1108,14 +1168,14 @@ defmodule Code do
     do_get_docs(String.to_charlist(binpath), kind)
   end
 
-  @docs_chunk 'ExDc'
+  @legacy_docs_chunk 'ExDc'
 
   defp do_get_docs(bin_or_path, kind) do
-    case :beam_lib.chunks(bin_or_path, [@docs_chunk]) do
-      {:ok, {_module, [{@docs_chunk, bin}]}} ->
+    case :beam_lib.chunks(bin_or_path, [@legacy_docs_chunk]) do
+      {:ok, {_module, [{@legacy_docs_chunk, bin}]}} ->
         lookup_docs(:erlang.binary_to_term(bin), kind)
 
-      {:error, :beam_lib, {:missing_chunk, _, @docs_chunk}} ->
+      {:error, :beam_lib, {:missing_chunk, _, @legacy_docs_chunk}} ->
         nil
     end
   end
--- a/lib/elixir/src/elixir_erl.erl
+++ b/lib/elixir/src/elixir_erl.erl
@@ -515,14 +515,29 @@ supports_extra_chunks_option() ->
 docs_chunk(Set, Line, Def, Defmacro, Types, Callbacks) ->
   case elixir_compiler:get_opt(docs) of
     true ->
-      ChunkData = term_to_binary({elixir_docs_v1, [
+      {DocLine, Doc} = get_moduledoc(Line, Set),
+
+      FunctionAndMacroDocs = get_docs(Set, Def, Defmacro),
+      CallbackDocs = get_callback_docs(Set, Callbacks),
+      TypeDocs = get_type_docs(Set, Types),
+
+      DocsChunkData = term_to_binary({docs_v1,
+        erl_anno:new(DocLine),
+        elixir,
+        <<"text/markdown">>,
+        doc_value(Doc),
+        #{},
+        [translate_doc(EntityDoc) || EntityDoc <- FunctionAndMacroDocs ++ CallbackDocs ++ TypeDocs]
+      }, [compressed]),
+
+      ExDcChunkData = term_to_binary({elixir_docs_v1, [
         {moduledoc, get_moduledoc(Line, Set)},
-        {docs, get_docs(Set, Def, Defmacro)},
-        {callback_docs, get_callback_docs(Set, Callbacks)},
-        {type_docs, get_type_docs(Set, Types)}
+        {docs, FunctionAndMacroDocs},
+        {callback_docs, CallbackDocs},
+        {type_docs, TypeDocs}
       ]}, [compressed]),
 
-      [{<<"ExDc">>, ChunkData}];
+      [{<<"Docs">>, DocsChunkData}, {<<"ExDc">>, ExDcChunkData}];
 
     false ->
       []
@@ -531,6 +546,13 @@ docs_chunk(Set, Line, Def, Defmacro, Types, Callbacks) ->
 deprecated_chunk(Deprecated) ->
   ChunkData = term_to_binary({elixir_deprecated_v1, Deprecated}, [compressed]),
   [{<<"ExDp">>, ChunkData}].
+
+doc_value(Doc) ->
+  case Doc of
+    nil -> none;
+    false -> hidden;
+    Doc -> #{<<"en">> => Doc}
+  end.
 
 get_moduledoc(Line, Set) ->
   case ets:lookup_element(Set, moduledoc, 2) of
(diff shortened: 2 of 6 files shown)


lib/elixir/lib/code.ex
```diff
@@ -1045,6 +1045,66 @@ defmodule Code do
...
+           docs :: [{{kind, name, arity}, anno, signature, doc, metadata}]}
+          | {:error, reason :: term}
+        when anno: :erl_anno.anno(),
+             beam_language: atom,
+             format: binary,
+             doc: %{binary => binary} | :none | :hidden,
+             kind: atom,
+             name: atom,
+             signature: [binary],
+             metadata: map
```
lackac: There might be a better way to write this spec. I'm not sure if I can name some parts of the spec for more clarity (e.g. `module_doc :: doc`).

lib/elixir/lib/code.ex
```diff
@@ -1045,6 +1045,66 @@ defmodule Code do
...
+  @since "1.7.0"
+  def fetch_docs(module) when is_atom(module) do
+    case :code.get_object_code(module) do
```
lackac: @josevalim I'm not sure if the current implementation addresses the case when the chunk is not in the beam file, but in a separate "doc/chunks/example.chunk" file. Would this be resolved in by `:beam_lib.chunks/2` or do we need to handle that case ourselves? This is where a good example case from an Erlang module would help.

lib/elixir/src/elixir_erl.erl
```diff
@@ -532,6 +547,13 @@ deprecated_chunk(Deprecated) ->
   ChunkData = term_to_binary({elixir_deprecated_v1, Deprecated}, [compressed]),
   [{<<"ExDp">>, ChunkData}].
 
+doc_value(Doc) ->
+  case Doc of
+    nil -> none;
+    false -> hidden;
+    Doc -> #{<<"en">> => Doc}
```
lackac: @josevalim AFAIK, there is no way to specify the language of the documentation in the code. I chose to default to `"en"` for now, because I think this is something that needs more thought and would raise more questions if we wanted to tackle it now.

lib/elixir/src/elixir_erl.erl
```diff
@@ -559,6 +581,38 @@ get_type_docs(Set, Types) ->
...
+
+signature_to_binary(Name, Signature) ->
+  BinaryName = atom_to_binary(Name, utf8),
+  Args = lists:map(fun
+    FormatArg({'\\\\', _, [Left, Right]}) ->
+      [FormatArg(Left), " \\\\ ", 'Elixir.Macro':to_string(Right)];
+    FormatArg({Var, _, _}) ->
+      atom_to_binary(Var, utf8)
+  end, Signature),
+  iolist_to_binary([BinaryName, "(", lists:join(", ", Args), ")"]).
```
lackac: Here we're also generating the function signatures (similarly to how IEx does it). This might make it unnecessary to include the original signature list (the simplified AST), but I added it in the metadata map for now. I'll see if it's necessary to keep it while tackling the last todo of migrating all functionality. Or we could decide to keep it in the metadata map just in case. I was also thinking whether to generate similar signatures for other kinds (callbacks, etc.) I think that would require touching an earlier compilation step that adds the necessary information to the ETS tables, but I haven't dug into it yet.

lib/elixir/src/elixir_erl.erl
```diff
@@ -559,6 +581,38 @@ get_type_docs(Set, Types) ->
...
+   doc_value(Doc),
+   #{}
+  }.
+
+translate_doc_kind(def)           -> function;
+translate_doc_kind(callback)      -> callback;
+translate_doc_kind(type)          -> type;
+translate_doc_kind(opaque)        -> opaque;
+translate_doc_kind(defmacro)      -> macro;
+translate_doc_kind(macrocallback) -> macrocallback.
```
lackac: @josevalim are you happy with this mapping? I was unsure about whether to use `type` for `opaque` too, but decided to use `opaque` as that's what we're using in `ExDc` at the moment.

lackac: I also thought about adding a generic clause that returns the argument, but in this case I think it's better not to extend the kinds silently if a new one is introduced. Instead a function clause error would provide an opportunity to make a conscious decision about naming.

lib/elixir/src/elixir_erl.erl
```diff
@@ -515,14 +515,29 @@ supports_extra_chunks_option() ->
...
     true ->
-      ChunkData = term_to_binary({elixir_docs_v1, [
+      {DocLine, Doc} = get_moduledoc(Line, Set),
+
+      FunctionAndMacroDocs = get_docs(Set, Def, Defmacro),
+      CallbackDocs = get_callback_docs(Set, Callbacks),
+      TypeDocs = get_type_docs(Set, Types),
+
+      DocsChunkData = term_to_binary({docs_v1,
+        erl_anno:new(DocLine),
```
lackac: @josevalim do we want to include the file in this?

lib/elixir/src/elixir_erl.erl
hauleth: Wouldn't it be clearer to have:

```erlang
translate_doc_kind(def)           -> function;
translate_doc_kind(defmacro)      -> macro;
translate_doc_kind(Type) -> Type.
```

I know that this isn't exactly the same, but it would allow seamless expansion of supported kinds in future without changes there.

lackac: @hauleth I think it's better to let new kinds fail on this first so that there is a conscious decision about how to name them in the docs chunk.

lib/elixir/lib/code.ex
josevalim: We would need to handle it ourselves but let's skip it for now as no project seems to be using the doc/chunks bit.

lib/elixir/src/elixir_erl.erl
```diff
@@ -559,6 +581,38 @@ get_type_docs(Set, Types) ->
...
+  }.
+
+translate_doc_kind(def)           -> function;
+translate_doc_kind(callback)      -> callback;
+translate_doc_kind(type)          -> type;
+translate_doc_kind(opaque)        -> opaque;
+translate_doc_kind(defmacro)      -> macro;
+translate_doc_kind(macrocallback) -> macrocallback.
+
+signature_to_binary(Name, Signature) ->
```
josevalim: I believe we should be able to replace this whole function by `'Elixir.Macro':to_string({Name, [], Signature})`. Thoughts?

lib/elixir/src/elixir_erl.erl
```diff
@@ -559,6 +581,38 @@ get_type_docs(Set, Types) ->
      {_, Line, Doc} <- ets:lookup(Set, {Kind, Tuple})]
   ).
 
+translate_doc({{Name, Arity}, Line, Kind, Signature, Doc}) ->
+  {{translate_doc_kind(Kind), Name, Arity},
+   erl_anno:new(Line),
+   [signature_to_binary(Name, Signature)],
+   doc_value(Doc),
+   #{signature => Signature}
```
josevalim: Do we need to store this in the metadata?

lib/elixir/src/elixir_erl.erl
josevalim: Just the line if fine for now IMO.

lib/elixir/src/elixir_erl.erl
```diff
@@ -559,6 +581,38 @@ get_type_docs(Set, Types) ->
...
+   erl_anno:new(Line),
+   [],
+   doc_value(Doc),
+   #{}
+  }.
+
+translate_doc_kind(def)           -> function;
+translate_doc_kind(callback)      -> callback;
+translate_doc_kind(type)          -> type;
+translate_doc_kind(opaque)        -> opaque;
```
josevalim: I am thinking it may be better to go with `type` and put `opaque` in the metadata. Thoughts?

lib/elixir/src/elixir_erl.erl
josevalim: Yeah, let's go the explicit road for now. :+1:

lib/elixir/src/elixir_erl.erl
josevalim: Perfect.

lib/elixir/src/elixir_erl.erl
josevalim: We should trust the signature in the docs. :) So we can remove the signature from the metadata.

josevalim:
@lackac this looks excellent. Just one or two minor comments. What if we go ahead and merge this and we do the migration to Code.fetch_docs in another PR?

lib/elixir/src/elixir_erl.erl
lackac: Sounds good to me, I'll do the change.

lib/elixir/src/elixir_erl.erl
lackac: Hmm, that works. Why isn't IEx doing the same?

lackac:
@josevalim sounds good, I'll make the two changes (`opaque` and simplifying `signature_to_binary`—is there anything else?) and then you can merge this back.

lib/elixir/src/elixir_erl.erl
josevalim: Oh... no idea. :)

josevalim:
> @josevalim sounds good, I'll make the two changes (opaque and simplifying signature_to_binary—is there anything else?) and then you can merge this back.

I think this is all!

josevalim:
Btw, I think the next steps would be:

  * [ ] migrate all functionality that uses Code.get_docs/2 to use Code.fetch_docs/1
  * [ ] remove the old docs chunk
  * [ ] change how we store docs to be closer to the new docs format. for example, so we don't have to translate the doc name and we can have the metadata builtin. Likely the only step we would do in the erlang side is the generation of signature

But we can do each of those as a separate PR. You also don't need to perform them all if you don't want to, but that should cover everything. :)

lackac:
@josevalim I'm wouldn't mind taking them on. As I was working on this I explored how Elixir was compiled and I find it fascinating.

lib/elixir/src/elixir_erl.erl
lackac: The change for this made CI fail and I could reproduce it locally by running `make clean` first. The error is the following:

```
== Compilation error in file lib/kernel/special_forms.ex ==
** (MatchError) no match of right hand side value: {:map, [], nil}
    /Users/LacKac/Code/elixir-lang/elixir/lib/elixir/lib/macro.ex:623: Macro.to_string/2
** (UndefinedFunctionError) function List.Chars.to_charlist/1 is undefined (module List.Chars is not available)
    List.Chars.to_charlist("lib/kernel/special_forms.ex")
    /Users/LacKac/Code/elixir-lang/elixir/lib/elixir/lib/kernel/parallel_compiler.ex:500: Kernel.ParallelCompiler.get_line/3
    /Users/LacKac/Code/elixir-lang/elixir/lib/elixir/lib/kernel/parallel_compiler.ex:483: Kernel.ParallelCompiler.to_error/4
    /Users/LacKac/Code/elixir-lang/elixir/lib/elixir/lib/kernel/parallel_compiler.ex:383: Kernel.ParallelCompiler.wait_for_messages/6
    /Users/LacKac/Code/elixir-lang/elixir/lib/elixir/lib/kernel/parallel_compiler.ex:135: Kernel.ParallelCompiler.spawn_workers/3
    /Users/LacKac/Code/elixir-lang/elixir/lib/elixir/lib/kernel/cli.ex:469: anonymous fn/2 in Kernel.CLI.process_command/2
```

josevalim: @lackac it is a bootstrap issue. We need some modules to compile Elixir itself and this change makes List.Chars required. Can you add `<<"lib/elixir/lib/list/chars.ex">>` [below here](https://github.com/elixir-lang/elixir/blob/master/lib/elixir/src/elixir_compiler.erl#L151) and it should work. I will investigate after merging to remove back this dependency.

hauleth: @josevalim protip, when you hit `y` in diff then URL will be expanded to canonical form with `master` replaced with commit SHA.

lackac: @josevalim it still fails to compile:

```
== Compilation error in file lib/kernel/special_forms.ex ==
** (MatchError) no match of right hand side value: {:map, [], nil}
    /Users/LacKac/Code/elixir-lang/elixir/lib/elixir/lib/macro.ex:623: Macro.to_string/2
```

I think it's because it's trying to generate the signature for the `%Struct{}` special form. The signature AST is `{:%, [], [{:struct, [], nil}, {:map, [], nil}]}` which `Macro.to_string/2` is trying to render as a `%StructName{}` expression, but the right hand side is not a map representation, so it fails on that.

If I add a special case just for this special form it successfully compiles the rest even without adding `List.Chars` to the bootstrap list. After doing that I wanted to see what signatures were generated for the special form macros. Here's the list:

``` elixir
iex(4)> for {{:macro, n, a}, _, [s], _, _} <- docs, do: {{n, a}, s}
[
  {{:%, 2}, "%struct_name{}"},
  {{:%{}, 1}, "%{args}"},
  {{:&, 1}, "&(expr)"},
  {{:., 2}, "left . right"},
  {{:::, 2}, "left :: right"},
  {{:<<>>, 1}, "<<args>>"},
  {{:=, 2}, "left = right"},
  {{:^, 1}, "^var"},
  {{:__CALLER__, 0}, "__CALLER__()"},
  {{:__DIR__, 0}, "__DIR__()"},
  {{:__ENV__, 0}, "__ENV__()"},
  {{:__MODULE__, 0}, "__MODULE__()"},
  {{:__STACKTRACE__, 0}, "__STACKTRACE__()"},
  {{:__aliases__, 1}, "args"},
  {{:__block__, 1}, "args"},
  {{:alias, 2}, "alias(module, opts)"},
  {{:case, 2}, "case(condition, clauses)"},
  {{:cond, 1}, "cond(clauses)"},
  {{:fn, 1}, "fn\n  [clauses]\nend"},
  {{:for, 1}, "for(args)"},
  {{:import, 2}, "import(module, opts)"},
  {{:quote, 2}, "quote(opts, block)"},
  {{:receive, 1}, "receive(args)"},
  {{:require, 2}, "require(module, opts)"},
  {{:super, 1}, "super(args)"},
  {{:try, 1}, "try(args)"},
  {{:unquote, 1}, "unquote(expr)"},
  {{:unquote_splicing, 1}, "unquote_splicing(expr)"},
  {{:with, 1}, "with(args)"},
  {{:{}, 1}, "{args}"}
]
```

As you can see some of these don't make much sense. It's also not much better than what the implementation borrowed from IEx was doing.

I see three options:

1. keep the original naive implementation based on IEx
2. use `'Elixir.Macro':to_string/2` but with a special case for `%struct_name{}`
3. generalise on the special case and allow macros to define their own signature (e.g. `@signature "%struct_name{key: value, ...}"`)

lib/elixir/lib/code.ex
fertapric: What do you think about being more explicit about the error responses `{:error, :module_not_found | : docs_chunk_not_found}` ?

lackac: I think that's a reasonable thing to do as long as the list of possible errors are limited as it is in this case.

lib/elixir/src/elixir_erl.erl
josevalim: I see. Let's keep the original naive implementation based on IEx then. :)

lackac:
@josevalim commits cleaned up, ready for a final review/merge :)

josevalim:

PR #7828, opened 2018-07-04
Migrate to `Code.fetch_docs/1`

lackac:
Related to #7198.

### DIFF
--- a/lib/elixir/lib/code.ex
+++ b/lib/elixir/lib/code.ex
@@ -1069,7 +1069,11 @@ defmodule Code do
   @spec fetch_docs(module) ::
           {:docs_v1, anno, beam_language, format, module_doc :: doc, metadata,
            docs :: [{{kind, name, arity}, anno, signature, doc, metadata}]}
-          | {:error, :module_not_found | :docs_chunk_not_found}
+          | {:error,
+             :module_not_found
+             | :docs_chunk_not_found
+             | {:unsupported_version, atom}
+             | {:unrecognized_chunk, term}}
         when anno: :erl_anno.anno(),
              beam_language: atom,
              format: binary,
@@ -1095,7 +1099,16 @@ defmodule Code do
   defp do_fetch_docs(bin_or_path) do
     case :beam_lib.chunks(bin_or_path, [@docs_chunk]) do
       {:ok, {_module, [{@docs_chunk, bin}]}} ->
-        :erlang.binary_to_term(bin)
+        case :erlang.binary_to_term(bin) do
+          {:docs_v1, _, _, _, _, _, _} = chunk ->
+            chunk
+
+          tuple when is_tuple(tuple) and is_atom(elem(tuple, 0)) ->
+            {:error, {:unsupported_version, elem(tuple, 0)}}
+
+          unknown ->
+            {:error, {:unrecognized_chunk, unknown}}
+        end
 
       {:error, :beam_lib, {:missing_chunk, _, @docs_chunk}} ->
         {:error, :docs_chunk_not_found}
--- a/lib/elixir/lib/code.ex
+++ b/lib/elixir/lib/code.ex
@@ -1069,11 +1069,8 @@ defmodule Code do
   @spec fetch_docs(module) ::
           {:docs_v1, anno, beam_language, format, module_doc :: doc, metadata,
            docs :: [{{kind, name, arity}, anno, signature, doc, metadata}]}
-          | {:error,
-             :module_not_found
-             | :docs_chunk_not_found
-             | {:unsupported_version, atom}
-             | {:unrecognized_chunk, term}}
+          | {:error, :module_not_found | :chunk_not_found | {:invalid_chunk, binary}}
+          | future_formats
         when anno: :erl_anno.anno(),
              beam_language: atom,
              format: binary,
@@ -1081,7 +1078,8 @@ defmodule Code do
              kind: atom,
              name: atom,
              signature: [binary],
-             metadata: map
+             metadata: map,
+             future_formats: term
   @since "1.7.0"
   def fetch_docs(module) when is_atom(module) do
     case :code.get_object_code(module) do
@@ -1099,19 +1097,14 @@ defmodule Code do
   defp do_fetch_docs(bin_or_path) do
     case :beam_lib.chunks(bin_or_path, [@docs_chunk]) do
       {:ok, {_module, [{@docs_chunk, bin}]}} ->
-        case :erlang.binary_to_term(bin) do
-          {:docs_v1, _, _, _, _, _, _} = chunk ->
-            chunk
-
-          tuple when is_tuple(tuple) and is_atom(elem(tuple, 0)) ->
-            {:error, {:unsupported_version, elem(tuple, 0)}}
-
-          unknown ->
-            {:error, {:unrecognized_chunk, unknown}}
+        try do
+          :erlang.binary_to_term(bin)
+        rescue
+          _ -> {:error, {:invalid_chunk, bin}}
         end
 
       {:error, :beam_lib, {:missing_chunk, _, @docs_chunk}} ->
-        {:error, :docs_chunk_not_found}
+        {:error, :chunk_not_found}
     end
   end
 
--- a/lib/elixir/lib/kernel/typespec.ex
+++ b/lib/elixir/lib/kernel/typespec.ex
@@ -24,10 +24,20 @@ defmodule Kernel.Typespec do
 
   @doc false
   def beam_typedocs(module) when is_atom(module) or is_binary(module) do
-    IO.warn("Kernel.Typespec.beam_typedocs/1 is deprecated, please use Code.get_docs/2 instead")
+    IO.warn("Kernel.Typespec.beam_typedocs/1 is deprecated, please use Code.fetch_docs/1 instead")
 
-    if docs = Code.get_docs(module, :type_docs) do
-      for {tuple, _, _, doc} <- docs, do: {tuple, doc}
+    case Code.fetch_docs(module) do
+      {:docs_v1, _, _, _, _, _, docs} ->
+        for {{:type, name, arity}, _, _, doc, _} <- docs do
+          case doc do
+            :none -> {{name, arity}, nil}
+            :hidden -> {{name, arity}, false}
+            %{"en" => doc_string} -> {{name, arity}, doc_string}
+          end
+        end
+
+      {:error, _} ->
+        nil
     end
   end
 
--- a/lib/elixir/lib/kernel/typespec.ex
+++ b/lib/elixir/lib/kernel/typespec.ex
@@ -23,9 +23,8 @@ defmodule Kernel.Typespec do
   end
 
   @doc false
+  @deprecated "Use Code.fetch_docs/1 instead"
   def beam_typedocs(module) when is_atom(module) or is_binary(module) do
-    IO.warn("Kernel.Typespec.beam_typedocs/1 is deprecated, please use Code.fetch_docs/1 instead")
-
     case Code.fetch_docs(module) do
       {:docs_v1, _, _, _, _, _, docs} ->
         for {{:type, name, arity}, _, _, doc, _} <- docs do
--- a/lib/ex_unit/lib/ex_unit/doc_test.ex
+++ b/lib/ex_unit/lib/ex_unit/doc_test.ex
@@ -428,7 +428,8 @@ defmodule ExUnit.DocTest do
 
   defp extract(module) do
     case Code.fetch_docs(module) do
-      {:docs_v1, line, _, _, moduledoc, _, docs} ->
+      {:docs_v1, anno, _, _, moduledoc, _, docs} ->
+        line = :erl_anno.line(anno)
         extract_from_moduledoc({line, moduledoc}, module) ++ extract_from_docs(docs, module)
 
       {:error, reason} ->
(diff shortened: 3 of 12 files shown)


lackac:
It seems to me that while the approach of parsing the docs chunk always on the caller side does the job, it leads to a lot of repetition. There are currently two implementations that look up specific kind of docs, two or three variations on how to interpret doc values (`:none`, `:hidden`, `%{"en" => doc}`), etc. For another example do a search for `docs_v1` in the PR and consider what would need to change when we need to upgrade to `docs_v2`.

I think that we should either add a `Code.Doc` module to deal with these or provide some minimal helpers in `Code`. I'm leaning towards the former since it would follow the pattern of `Code.Identifier` and `Code.Typespec`. `Code.fetch_docs/1` would also move to `Code.Doc.fetch_docs/1`.

@josevalim thoughts?

lib/iex/lib/iex/introspection.ex
```diff
@@ -745,9 +749,11 @@ defmodule IEx.Introspection do
     end
   end
 
-  defp elixir_module?(module) do
-    function_exported?(module, :__info__, 1)
-  end
+  defp translate_doc(nil), do: nil
+  defp translate_doc(:none), do: nil
+  defp translate_doc(:hidden), do: nil
+  defp translate_doc(binary) when is_binary(binary), do: binary
```
josevalim: When can we hit this clause?

josevalim:
> It seems to me that while the approach of parsing the docs chunk always on the caller side does the job, it leads to a lot of repetition. 

Good point.

I have seen some repetitions:

  * There is some repetition on Behaviour and Kernel.Typespec, but those are deprecated, so I would leave them as is.

  * There is some repetition such as raising exceptions when the chunk is not available, such as in the doctest. For those we could add `Code.fetch_docs!(...)`.

  * Finally there is some repetition in terms of ignoring the `{:error, _}` case, but I don't think there is much to do about it.

Did I miss some cases?

lib/iex/lib/iex/introspection.ex
lackac: From [line 379](https://github.com/lackac/elixir/blob/a59baf22465a279edc88c477b9e19e83e5d9cae9/lib/iex/lib/iex/introspection.ex#L379)

josevalim: Similarly, if we can avoid `nil` and use `:none` instead... but I am not sure how feasible is that.

lackac: This is exactly the symptom. Currently the code is in two minds. Do we keep the atoms used by the chunk and let clients deal with that or translate them early on into idiomatic Elixir? I'm not sure which I would prefer, but certainly not both. :)

The current implementation was mostly a mechanical replacement of `Code.get_docs/1`. I think with spending more time on it we could refactor the IEx related bits more. Moving some things like `count_defaults` to the chunk while sharing some code between `IEx.Autocomplete` and `IEx.Introspection`.

josevalim:
I think we should move to :none and :hidden because they are clearer and
more idiomatic than nil and false.

Regarding the defaults, the plan was to indeed at them to the metadata. So
I would leave the duplication for now and when we make it easier to define
the metadata from the Elixir side we can revisit it.
-- 


josevalim:
Also let’s go ahead with the fetch_docs change. :)
-- 


lackac:
@josevalim ok, let's live with the duplication for now. I completed the outstanding changes. This frees us to move forward and deal with the removal of `ExDc`, related code and prepare the generation of the docs chunk for the inclusion of more metadata.

lib/ex_unit/lib/ex_unit/doc_test.ex
```diff
@@ -427,39 +427,50 @@ defmodule ExUnit.DocTest do
...
+          module: module,
+          message:
+            "could not retrieve the documentation for module #{inspect(module)}. " <>
+              explain_docs_error(reason)
     end
+  end
 
-    moduledocs = extract_from_moduledoc(all_docs[:moduledoc], module)
+  defp explain_docs_error(:module_not_found),
+    do: "The BEAM file of the modoule cannot be accessed"
```
fertapric: Typo “modoule”

lib/iex/lib/iex/introspection.ex
```diff
@@ -401,42 +389,57 @@ defmodule IEx.Introspection do
...
   defp find_doc(docs, fun, arity) do
-    doc = List.keyfind(docs, {fun, arity}, 0) || find_doc_defaults(docs, fun, arity)
+    doc =
+      Enum.find(docs, &match?({_, ^fun, ^arity}, elem(&1, 0))) ||
+        find_doc_defaults(docs, fun, arity)
+
     if doc != nil and has_content?(doc), do: doc
   end
 
   defp find_doc_defaults(docs, function, min) do
```
fertapric: In order to get rid of that `elem(doc, 2)` what do you think about?

```elixir
defp find_doc_defaults(docs, function, min) do
  Enum.find(docs, fn
    {{_, ^function, arity}, _, signature, _, _} when arity > min ->
      defaults = count_defaults(signature)
      arity <= min + defaults

    _ ->
      false
  end)
end
```

lib/elixir/lib/code.ex
```diff
@@ -1095,7 +1099,16 @@ defmodule Code do
...
       {:ok, {_module, [{@docs_chunk, bin}]}} ->
-        :erlang.binary_to_term(bin)
+        case :erlang.binary_to_term(bin) do
+          {:docs_v1, _, _, _, _, _, _} = chunk ->
+            chunk
+
+          tuple when is_tuple(tuple) and is_atom(elem(tuple, 0)) ->
+            {:error, {:unsupported_version, elem(tuple, 0)}}
+
+          unknown ->
```
josevalim: Since the `binary_to_term` could fail altogether, maybe we should do this:

```elixir
try do
  :erlang.binary_to_term(binary)
rescue
  _ -> {:error, {:invalid_chunk, binary}}
end
```

Maybe somebody could put invalid data into the chunk but, if they do that, they deserve the errors they will get (which is true for any other chunk). :)

lib/iex/lib/iex/info.ex
```diff
@@ -45,10 +45,9 @@ defimpl IEx.Info, for: Atom do
 
   defp info_module(mod) do
     extra =
-      if Code.get_docs(mod, :moduledoc) do
-        "Use h(#{inspect(mod)}) to access its documentation.\n"
-      else
-        ""
+      case Code.fetch_docs(mod) do
+        {:docs_v1, _, _, _, %{}, _, _} -> "Use h(#{inspect(mod)}) to access its documentation.\n"
```
josevalim: Maybe here the opposite would be better?

```elixir
{:error, _} -> ""
_ -> "Use h(#{inspect(mod)}) to access its documentation.\n"
```

This way if we add new versions in the future we won't need to change it.

lib/mix/lib/mix/task.ex
```diff
@@ -124,9 +124,10 @@ defmodule Mix.Task do
   """
   @spec moduledoc(task_module) :: String.t() | nil
```
josevalim: According to the typespec we always return `nil`. Maybe we could get away with returning `nil` for `false` as well?

josevalim:
@lackac I have dropped some final comments. :)

lib/iex/lib/iex/info.ex
lackac: The module doc elem can also be `:none` or `:hidden` in which case `h/1` won't be able to return any documentation.

lib/mix/lib/mix/task.ex
lackac: Good catch, I should've looked at the typespec.

lib/elixir/lib/code.ex
```diff
@@ -1095,7 +1099,16 @@ defmodule Code do
   defp do_fetch_docs(bin_or_path) do
     case :beam_lib.chunks(bin_or_path, [@docs_chunk]) do
       {:ok, {_module, [{@docs_chunk, bin}]}} ->
-        :erlang.binary_to_term(bin)
+        case :erlang.binary_to_term(bin) do
+          {:docs_v1, _, _, _, _, _, _} = chunk ->
+            chunk
+
+          tuple when is_tuple(tuple) and is_atom(elem(tuple, 0)) ->
```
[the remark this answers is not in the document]
lackac: @josevalim I see, you want to make `fetch_docs` transparent and put the burden of knowing what's in the chunks entirely on the caller.

With this change and the other about rescuing `binary_to_term` I'm thinking that we should also revisit the typespec for `Code.fetch_docs/1`. We can't really promise to return the current spec unless we match on the result of `binary_to_term`. What should be the approach for the spec in this case? I would like to make the spec more useful than simply `@spec fetch_docs(module) :: term | {:error, term}`.

lib/elixir/lib/kernel/typespec.ex
```diff
@@ -24,10 +24,20 @@ defmodule Kernel.Typespec do
 
   @doc false
   def beam_typedocs(module) when is_atom(module) or is_binary(module) do
-    IO.warn("Kernel.Typespec.beam_typedocs/1 is deprecated, please use Code.get_docs/2 instead")
+    IO.warn("Kernel.Typespec.beam_typedocs/1 is deprecated, please use Code.fetch_docs/1 instead")
```
michalmuskala: Could we migrate to the `@deprecated` attribute here while we're at it?

lib/ex_unit/lib/ex_unit/doc_test.ex
```diff
@@ -427,39 +427,50 @@ defmodule ExUnit.DocTest do
   ## Extraction of the tests
 
   defp extract(module) do
-    all_docs = Code.get_docs(module, :all)
+    case Code.fetch_docs(module) do
+      {:docs_v1, line, _, _, moduledoc, _, docs} ->
```
michalmuskala: Is `line` always going to be a line number? The spec of `Code.get_docs` defines it as `:erl_anno.anno()` which might have more data in it. Should we use `:erl_anno.line(anno)` to get the line here?

lackac: That would certainly be safer. However, in doc tests we know that it's just a line since it was generated by the Elixir compiler where we don't include more information. But since that could change I think it's best to go with your advice.

lib/elixir/lib/code.ex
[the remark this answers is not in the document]
josevalim: We can try:

    @spec fetch_docs(module) :: {:docs_v1, ...} | {:error, term} | future_formats when future_formats: term

But this may have negative implications in dialyzer. We can give it a try though.

lib/ex_unit/lib/ex_unit/doc_test.ex
josevalim: Yeah, lets go with `:erl_anno.line/1`.

lib/iex/lib/iex/info.ex
josevalim: Oh, I see, thank you.

lib/iex/lib/iex/autocomplete.ex
```diff
@@ -437,43 +437,50 @@ defmodule IEx.Autocomplete do
...
         count > 0,
         new_arity <- (arity - count)..arity,
         do: {fun_name, new_arity}
   end
 
-  defp count_defaults(args) do
-    Enum.count(args, &match?({:\\, _, _}, &1))
+  defp count_defaults(signature) do
+    signature
+    |> Stream.flat_map(&Regex.scan(~r/ \\\\ /, &1))
```
michalmuskala: I'm not sure that's reliable - somebody could have a thing like:

michalmuskala: I'm not sure that's reliable - I could imagine code like this that would miscalculate here:
```elixir
iex(3)> quote(do: def foo(bar \\ ~S|\\|)) |> Macro.to_string
"def(foo(bar \\\\ ~S\"\\\\\"))"
```

josevalim: This is just temporary until we move the number of defaults to metadata. :)

lackac: Yes, you're right. We discussed this earlier:

I wrote:

> I think we could replace [`count_defaults`] by adding something like `defaults: 2` or `defaults: [:year, :options]` to the metadata in `elixir_erl`

@josevalim:

> Regarding the defaults, the plan was to indeed at them to the metadata. So
I would leave the duplication for now and when we make it easier to define
the metadata from the Elixir side we can revisit it.

I would look at that after we remove `ExDc` and we have a chance to refactor the chunk generation and how and what we store about docs in ETS during compilation. I think we can live with this edge case in autocompletion until then.

lib/mix/lib/mix/task.ex
```diff
@@ -122,11 +122,12 @@ defmodule Mix.Task do
 
   Returns the moduledoc or `nil`.
   """
-  @spec moduledoc(task_module) :: String.t() | nil
+  @spec moduledoc(task_module) :: String.t() | nil | false
```
josevalim: Apparently the spec was wrong, hence the build failure. I Fixed it and updated the spec. If the build is green tomorrow morning I will go ahead and merge this. :)

lackac: Thanks, I was about to push the same thing. :)

josevalim:

PR #7831, opened 2018-07-04
Use @deprecated instead of IO.warn in Kernel.Typespec

lackac:


josevalim:

PR #7832, opened 2018-07-05
Compile protocols in elixir syntax

michalmuskala:
This siwtches protocol compilation to happen on the epxnaded Elixir AST
instead of the Erlang AST. This is slightly less complex and opens
possibility for easy "partial" compilation of protocols.

Additionally instead of rewriting specs on consolidation, this marks the
functions that end-up being rewritten with @dialyzer :nowarn_function to
disable overly eager dialyzer warnings.

### DIFF
--- a/lib/elixir/lib/protocol.ex
+++ b/lib/elixir/lib/protocol.ex
@@ -168,8 +168,7 @@ defmodule Protocol do
   """
   defmacro derive(protocol, module, options \\ []) do
     quote do
-      module = unquote(module)
-      Protocol.__derive__([{unquote(protocol), unquote(options)}], module, __ENV__)
+      Protocol.__derive__([{unquote(protocol), unquote(options)}], unquote(module), __ENV__)
     end
   end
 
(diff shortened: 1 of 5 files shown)


lib/elixir/src/elixir_erl.erl
```diff
@@ -138,7 +138,13 @@ elixir_to_erl_cons2([], Acc) ->
 scope(_Meta) ->
   #elixir_erl{}.
 
-%% Compilation hook.
+%% Static compilation hook, used in protocol consolidation
+
+compile(Map, TypeSpecs, Chunks) ->
```
josevalim: Let's call this `consolidate` for now, it is very specific to the protocol mechanism and not a general compilation mechanism.

lib/elixir/lib/protocol.ex
```diff
@@ -549,6 +479,10 @@ defmodule Protocol do
           nil
         end
 
+      # Disable dialyzer checks - before and after consolidation
+      # the types could be more strict
+      @dialyzer {:nowarn_function, __protocol__: 1, impl_for: 1, impl_for!: 1}
```
josevalim: I would like to keep the previous behaviour because `impl_for/1` and `impl_for/2` is used in applications and changing the specs on consolidation meant we could catch bugs that now we won't.

lib/elixir/src/elixir_erl.erl
michalmuskala: I was thinking about moving the ETS table access up to the callers of `compile/1` and have a pure function here in general.

:+1: for `consolidate/3`.

lib/elixir/lib/protocol.ex
josevalim: Actually, ignore this. We were not rewriting `impl_for` specs and the `__protocol__` ones are not relevant to what I said above.

michalmuskala:
Updated

josevalim:
Since we are using `:debug_info`, this requires Erlang/OTP 20+, so we can only merge it after v1.7 is out.

michalmuskala:
Aaaaand, this has to wait for 1.8 after we drop support for OTP 19 - the `:debug_info` chunk is not available there.

josevalim:
@michalmuskala this can now be applied to master. Please rebase. :)

michalmuskala:
Updated.

PR #7836, opened 2018-07-07
[WIP] Return an error tuple for `handle_dot`

drincruz:
Also mentioned in #7270, we want to return an error tuple as well for errors in `handle_dot`.
We're getting a CaseClauseError in `handle_dot` whenver `Part` is a non-binary.


drincruz:
So, we're getting a CaseClauseError within `handle_dot` whenever `Part` is not a binary. Would we want to return an error tuple similar to how we handled Hex issues?

lib/elixir/src/elixir_tokenizer.erl
```diff
@@ -700,6 +700,8 @@ handle_dot([$., H | T] = Original, Line, Column, DotInfo, Scope, Tokens) when ?i
         {error, Reason} ->
           {error, Reason, Original, Tokens}
       end;
+    {_NewLine, _NewColumn, [_Part], Rest} ->
+        {error, {Line, Column, "some sort of error message", [H]}, Rest, Tokens};
```
drincruz: If we were okay with this implementation, yes, we would want to change the error message to something useful.

josevalim: The error message should be: "interpolation is not allowed when invoking functions".

lib/elixir/src/elixir_tokenizer.erl
```diff
@@ -700,6 +700,8 @@ handle_dot([$., H | T] = Original, Line, Column, DotInfo, Scope, Tokens) when ?i
         {error, Reason} ->
           {error, Reason, Original, Tokens}
       end;
+    {_NewLine, _NewColumn, [_Part], Rest} ->
```
josevalim: We should rather match on `{_NewLine, _NewColumn, _Parts, Rest}` since this code also fails with a CaseClauseError: `Foo."f#{:o}o"`. Basically any interpolation is disallowed here.

drincruz: This makes more sense! Thank you for the direction. 👍

PR #7838, opened 2018-07-07
Warn on unnecessary quotes

### LINKED ISSUE #7634: Warning for using quotes around keywords when they are not necessary, such as `"key": value`
### Environment

```
Erlang/OTP 19 [erts-8.3.5.3] [source] [64-bit] [smp:8:8] [async-threads:10] [hipe] [kernel-poll:false] [dtrace]

Elixir 1.7.0-dev (e07a41d9d) (compiled with OTP 19)
```
### Current behavior

The difference in meaning of `"key":` and `"key" =>` in maps can lead to some quite interesting and really hard to find bugs.

The two syntaxes for writing map keys are similar enough to be easily confused and not spotted instantly. 

While the convenience of writing `key: value`  in a map is great mixing it with the syntax of complex atoms makes it icky

### Expected behavior

I would like to propose a warning for this, it is a valid syntax so while confusing it shouldn't be an error in my opinion (even so I'd personally kind of like it :P).

I'm not sure how far to take it.

The first and kind of 'safest' option in my eyes would be to warn that `"key": value` should be written as `:"key" => value` to prevent confusion and make it obvious that `key` is an atom here.

I can understand however that enforcing a convention like that might not be appreciated by everyone, so as a middle ground it might be okay to warn on this re-write of `"key": value` and `key:` are the same (the atom would not need quoting). While not catching all the cases it'd probably warn on a majority of mistaken uses of `"key": value`

josevalim:
Closes #7634


josevalim:
Also, should it be `:warn_on_unnecessary_quote` or `:warn_on_unnecessary_quotes`? I went with the second but I thought I would double check.

lib/elixir/lib/code.ex
```diff
@@ -650,6 +650,10 @@ defmodule Code do
       when non-existing atoms are found by the tokenizer.
       Defaults to `false`.
 
+    * `:warn_on_unnecessary_quotes` - when `false`, does not warn
+      when atoms, keywords and calls have unnecessary quotes on
+      them
```
fertapric: I would add a dot at the end and the default value to keep consistency with previous descriptions.

> * `:warn_on_unnecessary_quotes` - when `false`, does not warn when atoms, keywords and calls have unnecessary quotes on them. Defaults to `true`.

fertapric: And maybe use and "or" instead of "and"? 

> does not warn when atoms, keywords **or** calls have unnecessary quotes on them

lib/elixir/lib/code.ex
```diff
@@ -650,6 +650,10 @@ defmodule Code do
       when non-existing atoms are found by the tokenizer.
       Defaults to `false`.
 
+    * `:warn_on_unnecessary_quotes` - when `false`, does not warn
+      when atoms, keywords or calls have unnecessary quotes on
+      them.
```
fertapric: I think the documentation about the default value is handy :)

I just want to be sure that part was not overlooked from my previous comments. If you disagree, no worries :)

josevalim: I missed it. :) Tks for double checking.

josevalim:
Yes, the rebar3 plugin should do the same. The options given to eval are not forwarded to string_to_quoted but you can just call it directly and put it all together. This approach should also be backwards compatibility. `string_to_quoted` and `eval_quoted` have been there forever!

Issue #7494, opened 2018-03-26
Augment logger translations with metadata


josevalim:
The following metadata should just work:

  * [x] `:name` ([source](https://github.com/appsignal/appsignal-elixir/blob/e3d304e2424a8e648f66ee1f23cc7da427bacea2/lib/appsignal/error_handler.ex#L81-L84)). This option appears optionally in `:registered_name` in proc_lib crash reports and it may be present in behaviours as an atom or PID

The following metadata requires new formatter rules:

  * [x] `:initial_call` ([source](https://github.com/getsentry/sentry-elixir/blob/747964b81dab4a3f79a154d3f918ba3c9fd95c2a/lib/sentry/logger.ex#L121)) - available only in proc_lib crash reports.

The following metadata is always ignored unless explicitly fetched by the logger backend:

  * [x] `:crash_reason` as `{reason, stack}` - ([source](https://github.com/getsentry/sentry-elixir/blob/747964b81dab4a3f79a154d3f918ba3c9fd95c2a/lib/sentry/logger.ex#L108-L115)) ([source](https://github.com/elixir-addicts/rollbax/blob/7ff1181b0f8e2ed1878b5bb4c8496f9eda144510/lib/rollbax/reporter/standard.ex#L121-L140)) ([source](https://github.com/jarednorman/bugsnag-elixir/blob/master/lib/bugsnag/logger.ex#L21-L23)). It is available as `{kind, reason, stack}` in `proc_lib` and it is in the return of `format_reason` in behaviours

josevalim:
We can also use the first stacktrace entry to fill in the module and function.

mitchellhenke:
happy to take this if nobody is actively working on it. would be able to get started on it this week 🙂

josevalim:
@mitchellhenke I am already working on it :) but reviews will be really appreciated when I send a PR. :)

mitchellhenke:
would be glad to!

mitchellhenke:
comment updated

josevalim:
We want to include some of it (pid, registered_name, error_info, etc) but not all of it. It should be easy to add new information at any time though. Is there anything in particular you would like to add?

mitchellhenke:
dictionary is particularly useful I think. It would include `logger_metadata` which could even be pulled out and passed to the eventual Logger call as well?

josevalim:
[...]

PR #7841, opened 2018-07-08
include crash_reason in Logger translator

mitchellhenke:
Related to #7494

@josevalim we hadn't discussed it, but for the other metadata mentioned in #7494 (`:name`, `:initial_call`), should it be pulled from places like https://github.com/elixir-lang/elixir/blob/8a971fcb44391bd8b16456666f3033b633c6ff77/lib/logger/lib/logger/translator.ex#L348 and https://github.com/elixir-lang/elixir/blob/8a971fcb44391bd8b16456666f3033b633c6ff77/lib/logger/lib/logger/translator.ex#L346?

I also added the `:inspect` option to the test logger backend to print events to the console as it was helpful to be able to turn it on and off per test.  I wasn't sure if it should be merged in, so let me know if it should be removed/refactored.

### DIFF
--- a/lib/logger/lib/logger/backends/console.ex
+++ b/lib/logger/lib/logger/backends/console.ex
@@ -191,7 +191,9 @@ defmodule Logger.Backends.Console do
     |> color_event(level, colors, md)
   end
 
-  defp take_metadata(metadata, :all), do: metadata
+  defp take_metadata(metadata, :all) do
+    Keyword.drop(metadata, [:crash_reason])
+  end
 
   defp take_metadata(metadata, keys) do
     Enum.reduce(keys, [], fn key, acc ->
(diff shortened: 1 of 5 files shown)


josevalim:
@mitchellhenke regarding initial call and registered name, yes. You can also check the original issue and compare with how tools are extracting this information.

josevalim:
Merged manually, thank you! :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

josevalim:
@mitchellhenke I went down a rabbit whole to make the tests pass on Erlang/OTP 19 and even had to do some changes based on how tasks work. But it should all be good now!

I have also changed the test backend to be added on setup_all, so we don't need to do it per test.

I think all is ready now to receive a patch that also returns initial_call and registered_name in the metadata. :)

josevalim:
@mitchellhenke Btw, don't forget to rebase before working on initial_call, as there has been changes around crash reports (which is where initial_call and registered_name are defined).

PR #7844, opened 2018-07-09
Include initial_call and registered_name in metadata

### LINKED ISSUE #7494: Augment logger translations with metadata

josevalim:
Closes #7494.

### DIFF
--- a/lib/logger/lib/logger.ex
+++ b/lib/logger/lib/logger.ex
@@ -301,6 +301,17 @@ defmodule Logger do
       but it can be useful to other backends, such as the ones that report
       errors to third-party services
 
+    * `:initial_call` - the initial call that started the process
+
+    * `:registered_name` - the process registered name as an atom
+
+  Note that all metadata is optional and may not always be available.
+  The `:module`, `:function`, `:line` and similar metadata are automatically
+  included when using the `Logger` macros. `Logger.bare_log/3` does not
+  include such metadata by default. Other metadata, such as `:crash_reason`,
+  `:initial_call` and `:registered_name` are extracted from Erlang/OTP
+  crash reports and available only under those circumstances.
+
   ### Custom formatting
 
   The console backend allows you to customize the format of your log messages
--- a/lib/logger/lib/logger/translator.ex
+++ b/lib/logger/lib/logger/translator.ex
@@ -88,6 +88,7 @@ defmodule Logger.Translator do
     case message do
       {'** Generic server ' ++ _, [name, last, state, reason | client]} ->
         {formatted, reason} = format_reason(reason)
+        metadata = [crash_reason: reason] ++ registered_name(name)
 
         msg =
           ["GenServer #{inspect(name)} terminating", formatted] ++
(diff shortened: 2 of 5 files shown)


lib/logger/lib/logger/translator.ex
```diff
@@ -534,6 +546,9 @@ defmodule Logger.Translator do
     end
   end
 
+  defp registered_name(name) when is_atom(name), do: [registered_name: name]
+  defp registered_name(_name), do: []
```
michalmuskala: What about via tuples here? Are we ignoring them?

josevalim: @michalmuskala yes, as the name can be any term and it is not guaranteed we can log those. So I am sticking with the otp reports conventions here (either an atom or nothing).

lib/logger/lib/logger.ex
```diff
@@ -301,6 +301,17 @@ defmodule Logger do
       but it can be useful to other backends, such as the ones that report
       errors to third-party services
 
+    * `:initial_call` - the initial call that started the process
+
+    * `:registered_name` - the process registered name as an atom
+
+  Note that all metadata is optional and may not always be available.
+  The `:module`, `:function`, `:line` and similar metadata are automatically
```
fertapric: I think the docs use Oxford commas, what do you think about using it here?

> `:module`, `:function`, `:line`, and similar metadata are automatically included when using the `Logger` macros.

**Note:** I would also remove the "The" at the start.

> Other metadata, such as  such as `:crash_reason`, `:initial_call`, and `:registered_name` are extracted from Erlang/OTP crash reports and available only under those circumstances.

PR #7852, opened 2018-07-10
Support adding arbitrary docs metadata with `@doc keyword()`

### LINKED ISSUE #7830: Rename @since ... to @doc since: ...
The new docs chunk has a metadata field but currently there is no way for us to attach metadata besides using specific attributes like `@since`. However, if we need to add a new module attribute for each possible documentation, it won't scale. Things are made worse by the fact that we would need `@since`, `@modulesince` and `@typesince`.

To solve this problem, we should allow a keyword list to be given to `@doc`, `@moduledoc` and `@typedoc` which will be stored and attached as metadata.

So this code:

```elixir
@since "1.3.0"
def foo
```

will become:

```elixir
@doc since: "1.3.0"
def foo
```

This means this feature is now automatically available to types and modules too:

```elixir
@moduledoc since: "1.3.0"
```

and we can attach any metadata we want, for example:

```elixir
@moduledoc authors: ["Alice", "Bob"],
           cross_references: [AnotherModule]
```

The format is a bit more verbose but the change is worth in the sense it supports multiple metadata across different places.

This feature depends on us finishing adopting EEP 48 which is still work in progress.

**Note:** we won't change `@deprecated` because it has functional implications and it is not only used for documentation purposes.

lackac:
Closes #7830.

This PR introduces a common way of defining arbitrary metadata for documentation
using the existing `@doc`, `@moduledoc` and `@typedoc` module attributes. This
is set to replace the `@since` attribute and potentially also the `@deprecated`
attribute.

### Example

``` elixir
  @moduledoc "Module for working with integers"
  @moduledoc author: "José Valim"

  @typedoc since: "1.5.0"
  @type t :: integer

  @doc """
  Returns the greatest common divisor of the two given integers.
  """
  @doc since: "1.5.0"
  def gcd(integer1, integer2)

  @doc "Multiple calls with keywords are merged"
  @doc since: "1.7.0", color: :red
  @doc color: :blue, stable: true # results in %{since: "1.7.0", color: :blue, stable: true} in metadata
  @doc "Calling @doc twice with a binary still raises a warning"
```

### TODO

- [x] allow specifying docs metadata with `@doc` and `@typedoc` attributes
- [x] allow specifying module doc metadata with `@moduledoc`
- [x] warn and ignore reserved keys (currently `defaults` and `opaque`)
- [x] add and amend documentation
- [x] search and replace `@since ...` with `@doc since: ...`

The last two might be better to do after 1.7 so that tooling (primarily ExDoc) has time to catch up.

### DIFF
--- a/lib/elixir/lib/module.ex
+++ b/lib/elixir/lib/module.ex
@@ -1714,28 +1714,6 @@ defmodule Module do
     :ok
   end
 
-  # This is the same list of attributes as in :elixir_module.
-  # We do not insert into the :attributes key in the bag table
-  # because those attributes are deleted on every definition.
-  defp put_attribute(module, key, value, line, set, _bag)
-       when (key in [:doc, :typedoc, :moduledoc] and not is_list(elem(value, 1))) or
-              key in [:impl, :since, :deprecated] do
-    try do
-      :ets.lookup_element(set, key, 3)
-    catch
-      :error, :badarg -> :ok
-    else
-      unread_line when is_integer(line) and is_integer(unread_line) ->
-        message = "redefining @#{key} attribute previously set at line #{unread_line}"
-        IO.warn(message, attribute_stack(module, line))
-
-      _ ->
-        :ok
-    end
-
-    :ets.insert(set, {key, value, line})
-  end
-
   # If any of the doc attributes are called with a keyword list that
   # will become documentation metadata. Multiple calls will be merged
   # into the same map overriding duplicate keys.
@@ -1751,6 +1729,27 @@ defmodule Module do
         current_metadata = :ets.lookup_element(set, {key, :meta}, 2)
         :ets.update_element(set, {key, :meta}, {2, Map.merge(current_metadata, metadata_map)})
     end
+  end
+
+  # This is the same list of attributes as in :elixir_module.
+  # We do not insert into the :attributes key in the bag table
+  # because those attributes are deleted on every definition.
+  defp put_attribute(module, key, value, line, set, _bag)
+       when key in [:doc, :typedoc, :moduledoc, :impl, :since, :deprecated] do
+    try do
+      :ets.lookup_element(set, key, 3)
+    catch
+      :error, :badarg -> :ok
+    else
+      unread_line when is_integer(line) and is_integer(unread_line) ->
+        message = "redefining @#{key} attribute previously set at line #{unread_line}"
+        IO.warn(message, attribute_stack(module, line))
+
+      _ ->
+        :ok
+    end
+
+    :ets.insert(set, {key, value, line})
   end
 
   defp put_attribute(_module, key, value, line, set, bag) do
--- a/lib/elixir/lib/module.ex
+++ b/lib/elixir/lib/module.ex
@@ -1717,9 +1717,9 @@ defmodule Module do
   # If any of the doc attributes are called with a keyword list that
   # will become documentation metadata. Multiple calls will be merged
   # into the same map overriding duplicate keys.
-  defp put_attribute(_module, key, {_, metadata}, line, set, _bag)
+  defp put_attribute(module, key, {_, metadata}, line, set, _bag)
        when key in [:doc, :typedoc, :moduledoc] and is_list(metadata) do
-    metadata_map = Map.new(metadata)
+    metadata_map = preprocess_doc_meta(metadata, module, line, %{})
 
     case :ets.insert_new(set, {{key, :meta}, metadata_map, line}) do
       true ->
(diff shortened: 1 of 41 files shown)


lib/elixir/lib/module.ex
```diff
@@ -1727,6 +1736,20 @@ defmodule Module do
...
+  # into the same map overriding duplicate keys.
+  defp put_attribute(_module, key, {_, metadata}, line, set, _bag)
+       when key in [:doc, :typedoc, :moduledoc] and is_list(metadata) do
+    case :ets.lookup(set, {key, :meta}) do
+      [] ->
+        :ets.insert(set, {{key, :meta}, Enum.into(metadata, %{}), line})
+
+      [{{^key, :meta}, current_metadata, _}] ->
+        :ets.update_element(set, {key, :meta}, {2, Enum.into(metadata, current_metadata)})
+    end
```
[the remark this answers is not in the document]
lackac: Yes, this would be better, I'll do the change.

lib/elixir/lib/module.ex
```diff
@@ -1727,6 +1736,23 @@ defmodule Module do
     :ets.insert(set, {key, value, line})
   end
 
+  # If any of the doc attributes are called with a keyword list that
+  # will become documentation metadata. Multiple calls will be merged
+  # into the same map overriding duplicate keys.
+  defp put_attribute(_module, key, {_, metadata}, line, set, _bag)
```
josevalim: Let's move this clause up so we don't need to change the guards of the clause before.

josevalim:
> deprecate @since

You don't need to deprecate it. Just remove all usage as well as the code that handles it. I will add the deprecation warning in a separate commit because I need to look into the most efficient way of doing it.

> same treatment for @deprecated?

We won't change `@deprecated` because it impacts the language in other ways. It may be confusing though. We will see how it goes.

lib/elixir/lib/module.ex
lackac: Good catch, will do.

lackac:
> We won't change @deprecated because it impacts the language in other ways. It may be confusing though. We will see how it goes.

I don't think it would be complicated to introduce `@doc deprecated: ...`. It looks to me that we could keep most of the internal implementation for now. We would only have to make sure that the same information is put into the ETS bag about deprecations even if it comes from a `@doc deprecated: ...` invocation. That would leave `ExDp` and xref untouched for now. Those could be changed later with less impact to users.

josevalim:
> I don't think it would be complicated to introduce @doc deprecated: .... It looks to me that we could keep most of the internal implementation for now. 

It is not an implementation concern. It is just that I don't think a documentation annotation should change how the code behaves. For all purposes, we can even support `@doc deprecated: ...`, if you want to mark something as deprecated but you don't want it to warn when invoked (soft deprecation). Although we should still keep `@deprecated` around and also add the `@deprecated` value to the metadata.

lackac:
I see, thank you for explaining, I haven't thought of that. :)

josevalim:
And I should have explained this since the beginning. :) We should document it too.

josevalim:
Btw, don't worry about ex_doc. ex_doc master already supports the new chunks.

lackac:
@josevalim I started to look into the changes that were necessary there last week. Noticed that you rewrote `signature_to_binary` based on what ExDoc was doing. :)

lib/elixir/lib/module.ex
```diff
@@ -1752,15 +1806,18 @@ defmodule Module do
...
         value
 
+      {line, [{key, _} | _]} when is_integer(line) and is_atom(key) ->
+        value
+
       {line, doc} when is_integer(line) ->
         raise ArgumentError,
               "@#{key} is a built-in module attribute for documentation. It should be " <>
-                "a binary, a boolean, or nil, got: #{inspect(doc)}"
+                "a binary, boolean, keyword list, or nil, got: #{inspect(doc)}"
```
fertapric: Not a native English speaker here, but “It should be a binary, a bookean, a keyword list or nil” sounds better to me :)

(same applies to the text below)

lib/elixir/lib/module.ex
```diff
@@ -1854,6 +1911,29 @@ defmodule Module do
...
+
+  defp preprocess_doc_meta([{key, value} | tail], module, line, map) when is_atom(key) do
+    validate_doc_meta(key, value)
+    preprocess_doc_meta(tail, module, line, Map.put(map, key, value))
+  end
+
+  defp validate_doc_meta(:since, value) when not is_binary(value) do
+    raise ArgumentError,
+          ":since is a built-in documentation metadata key. " <>
+            "It should be a string representing the version a function, macro, type or " <>
```
fertapric: I think there is something missing in this sentence:

> It should be a string representing the version **in which** a function, a macro, a type or a callback was added

(Btw, I'm not sure if **to which** is the proper preposition, maybe better to ask a native English speaker)

josevalim:
Thanks you @lackac! Since this is blocking some work in the ExDoc front, I will go ahead and merge it and do some changes as well as the ones requested by @fertapric!

josevalim:

lib/elixir/lib/module.ex
lackac: I don't know, also not a native speaker, and for me including an indefinite article before each item sounds superfluous. I tried now to look up, but couldn't find a generic rule for this. This is the closest I've come: https://english.stackexchange.com/questions/253041/should-an-article-be-used-when-enumerating-items

My favorite bit from that page:

> shows that the rules governing article usage are highly complex

PR #7855, opened 2018-07-10
Reenable tasks without side-effects, that only run code

### LINKED ISSUE #7848: Mix run can't execute multiple scripts
### Environment

```
Erlang/OTP 20 [erts-9.3.1] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:10] [hipe] [kernel-poll:false] [dtrace]

```

* Operating system: MacOS

### Current behavior

If I have a mix alias like this:

```
"my.alias": ["run script1.exs", "run script2.exs"]
```

only the first script will be run.

### Expected behavior

I would expect both scripts to run.

As mentioned in several places [[1]] [[2]] [[3]], mix tasks can only be run once. I'm not sure why the limitation exists, but in the context of `run script.exs` it doesn't make sense to me.

[1]: https://stackoverflow.com/questions/48291606/how-to-alias-run-twice-in-mix-exs
[2]: https://stackoverflow.com/questions/36846041/elixir-mix-aliases-two-run-tasks-in-one-alias
[3]: https://github.com/elixir-lang/elixir/issues/6488

While it makes sense that we wouldn't want the exact same mix task to run twice, it seems like it should be conditional on the task __and arguments__ rather than just the task.

josevalim:
Closes #7848.

### DIFF
--- a/lib/mix/lib/mix/tasks/run.ex
+++ b/lib/mix/lib/mix/tasks/run.ex
@@ -40,11 +40,14 @@ defmodule Mix.Tasks.Run do
 
       elixir --sname hello -S mix run --no-halt
 
+  This task is automatically reenabled, so it can be called multiple times
+  with different arguments.
+
   ## Command-line options
 
     * `--config`, `-c`  - loads the given configuration file
     * `--eval`, `-e` - evaluates the given code
-    * `--require`, `-r` - requires pattern before running the command
+    * `--require`, `-r` - executes the given pattern/file
     * `--parallel`, `-p` - makes all requires parallel
     * `--preload-modules` - preloads all modules defined in applications
     * `--no-compile` - does not compile even if files require compilation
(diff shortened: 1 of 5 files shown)


lib/mix/lib/mix/tasks/run.ex
```diff
@@ -40,11 +40,14 @@ defmodule Mix.Tasks.Run do
...
 
+  This task is automatically reenabled, so it can be called multiple times
+  with different arguments.
+
   ## Command-line options
 
     * `--config`, `-c`  - loads the given configuration file
     * `--eval`, `-e` - evaluates the given code
-    * `--require`, `-r` - requires pattern before running the command
+    * `--require`, `-r` - executes the given pattern/file
```
fertapric: I wonder if this change is consistent with the description of the next flag:

> `parallel`, `-p` - makes all requires parallel

Maybe “requires (and executes) the given pattern/file”?

josevalim: I think so, since the word "requires" is talking about the `--require` flag. Thoughts?

michalmuskala:
I think this is a sensible change, I don't see any downsides to this.

josevalim:

PR #7856, opened 2018-07-11
Be precise in documentation about after values in receive/1

eksperimental:
* 4_294_967_295 is still a valid number
* it is not correct to say that :infinity is the same as not using a timeout value,
  as this is needed if used the after clause. To what's equivalent is as to not to
  use the after clause itself.
* use the formal name "hexadecimal notation" in favor of "hex notation"

[ci skip]


josevalim:

PR #7858, opened 2018-07-11
Add comment in typespecs page next to port type

eksperimental:
Mention that `port()` is a "port identifier", in the same vein as
`pid()` is commented as being a "proccess identifier"

[ci skip]


josevalim:

PR #7860, opened 2018-07-11
Clarify documentation in GenServer

eksperimental:
Since the documented function uses a "module" arg., it was not
clear whether "module" referrer to the argument or to GenServer

[ci skip]


josevalim:

PR #7863, opened 2018-07-11
Change arg names in example in receive/1

eksperimental:
1. avoid use of single-char variables with not descriptive names ("i")
2. "value" can be thought of a number, so to make it clearer I replace it with "name"


josevalim:

PR #7864, opened 2018-07-11
Improve specs and documentation for List.wrap/1

eksperimental:


josevalim:

lib/elixir/lib/list.ex
```diff
@@ -390,10 +390,10 @@ defmodule List do
   end
 
   @doc """
-  Wraps the argument in a list.
+  Wraps `term` in a list if this is not list.
```
lackac: I'm not a native speaker, but this sentence feels strange. "... if it is not already a list." might read better.

josevalim: Yup, good catch. Looks like a typo.

PR #7865, opened 2018-07-11
Use backtick in "returns true" and "returns false" in docs

eksperimental:
[ci skip]


josevalim:

PR #7871, opened 2018-07-11
Improve Supervisor docs and specs

eksperimental:


lib/elixir/lib/supervisor.ex
```diff
@@ -529,12 +528,12 @@ defmodule Supervisor do
           | {:max_seconds, pos_integer}
 
   @typedoc "Supported strategies"
-  @type strategy :: :one_for_one | :one_for_all | :rest_for_one
+  @type strategy :: :one_for_one | :one_for_all | :rest_for_one | :simple_one_for_one
```
eksperimental: Eventhough it has been deprecated, it should still be reflected in the typespecs

lib/elixir/lib/supervisor.ex
```diff
@@ -529,12 +528,12 @@ defmodule Supervisor do
...
 
   # Note we have inlined all types for readability
   @typedoc "The supervisor specification"
   @type child_spec :: %{
-          required(:id) => term(),
+          required(:id) => module() | term(),
```
eksperimental: Even though it's redundant it shows what's the default :id

lib/elixir/lib/supervisor.ex
josevalim: We have recently agreed in another issue that we won'tkeep in typespecs as we do want the warning there still.

lib/elixir/lib/supervisor.ex
josevalim: I think `atom()` is better as the `module()` is just a consequence of the atom.

josevalim:

PR #7872, opened 2018-07-11
Specify that "mix new" APP name only takes lowercase ASCII letters

eksperimental:


josevalim:

jdmarshall:
Why so many limitations on app names?

PR #7873, opened 2018-07-11
Rename apply/3 second arg.

eksperimental:
Second argument should be named "function_name" and not "fun" as "fun"
is generally used for anonymous functions.


josevalim:

PR #7875, opened 2018-07-11
Change wording in error giving options

eksperimental:
Message read weird (didn't read natural):

    supervisors expect each child to be one of:
      * a module
      * a {module, arg} tuple
      * a child specification as a map with at least the :id and :start fields
      * or a tuple with 6 elements generated by Supervisor.Spec (deprecated)

changed to
    supervisors expect each child to be one of the following:
      * a module
      ....


josevalim:

PR #7883, opened 2018-07-12
Fix time unit depracation messages

lukaszsamson:
Fixes
```
warning: deprecated time unit: :nanosecond. A time unit should be :second, :millisecond, :microsecond, :nanosecond, or a positive integer
```


josevalim:
Hrm, good catch, we need both. We need one to use when warning and another to return.

lukaszsamson:
Replaced with https://github.com/elixir-lang/elixir/pull/7885

PR #7885, opened 2018-07-12
Fix time unit depracation messages

lukaszsamson:
Fixes
```
warning: deprecated time unit: :nanosecond. A time unit should be :second, :millisecond, :microsecond, :nanosecond, or a positive integer
```


josevalim:

PR #7886, opened 2018-07-12
Show documentation metadata in IEx (only since and deprecated for now)

lackac:


lackac:
@josevalim so far this only deals with moduledoc, functions, and macros. I'm adding types and callbacks now. :)

lib/elixir/lib/io/ansi/docs.ex
```diff
@@ -53,6 +55,37 @@ defmodule IO.ANSI.Docs do
...
+
+  defp print_single_metadata({{:since, value}, options}, _printed) do
+    IO.puts([metadata_label(:since, options), " ", value])
+  end
+
+  # only printing deprecated and since for now
+  defp print_single_metadata(_, printed), do: printed
+
+  defp metadata_label(key, options) do
+    "#{color(:doc_metadata, options)}#{key}:#{IO.ANSI.reset()}"
```
fertapric: What do you think about? `color(:doc_metadata, options) <> "#{key}:" <> IO.ANSI.reset()`

lackac: @fertapric are you recommending this because you think it looks clearer, or for performance benefits?

fertapric: Because I think it looks clearer :)

lackac: I agree, changing it. :)

lib/elixir/lib/io/ansi/docs.ex
```diff
@@ -53,6 +55,37 @@ defmodule IO.ANSI.Docs do
...
+
+  See `default_options/0` for docs on the supported options.
+  """
+  @spec print_metadata(map, keyword) :: :ok
+  def print_metadata(metadata, options \\ []) when is_map(metadata) do
+    options = Keyword.merge(default_options(), options)
+    print_each_metadata(metadata, options) && newline_after_block()
+  end
+
+  defp print_each_metadata(metadata, options) do
```
fertapric: I was wondering if we might implement a more generic version. This would be my take on this code:

```elixir
@printed_metadata [:since, :deprecated]

defp print_each_metadata(metadata, options) do
  Enum.reduce(metadata, false, fn
    {key, value}, _printed when key in @printed_metadata ->
      label = color(:doc_metadata, options) <> "#{key}:" <> IO.ANSI.reset()
      indent = String.duplicate(" ", length_without_escape(label) + 1)
      write_with_wrap([label | String.split(value, @spaces)], options[:width], indent, true)

    _metadata, printed ->
      printed
  end)
end
```

What I like about this proposal is that if we decide to include new keys, developers would only need to add the key to the module attribute (instead of generating a new clause for `print_single_metadata` which I think it would be the tendency looking at the code); if we decide to print all metadata it would be just removing the second clause and the guard of the first clause; and if we decide to blacklist the keys (for example `defaults`) then we can't simply turn `when key in @printed_metadata` to `when key in @blacklisted_metadata_keys`.

Thoughts?

Btw, I don't like the naming of the module attribute 😞, we should probably find a better name.

lib/elixir/lib/io/ansi/docs.ex
lackac: On second thought, we are constructing a string out of 4 elements, so I think the use of interpolation is ok. Changing this would require squash or rebase to make the history cleaner, and a new CI run, etc. While those are not significant overheads, in this case the benefit is not significant either to warrant them. I'll change this as well if other things come up too. Otherwise leaving this for another time. :)

Talking about overheads while writing an essay as a reply to a trivial change request. :)

lib/elixir/lib/io/ansi/docs.ex
lackac: @fertapric I started out with that approach, but abandoned it because it's not necessary for now to be generic and we don't know what values we're going to need to deal with in the future.

For example, your snippet assumes that `value` is a string. But in the general case that will not be true. In addition we won't necessarily need or want to wrap values for other keys and we don't need to do it for `since` either.

lib/elixir/lib/io/ansi/docs.ex
fertapric: Ok, no problem.

> Changing this would require squash or rebase to make the history cleaner

I think José squashes everything into a single commit when he merges the PR (unless you are talking about the history for reviewing the PR)

lib/elixir/lib/io/ansi/docs.ex
fertapric: `when is_binary(value) and key in @metadata_keys`? 😅 

I see what you mean. I guess we can make it more generic later when we add more keys.

lib/elixir/lib/io/ansi/docs.ex
lackac: Not always. If the history in the PR is already clean he might not do it. :)

lib/elixir/lib/io/ansi/docs.ex
```diff
@@ -53,6 +55,37 @@ defmodule IO.ANSI.Docs do
     newline_after_block()
   end
 
+  @doc """
+  Prints documentation metadata (only `since` and `deprecated` for now).
+
+  See `default_options/0` for docs on the supported options.
+  """
+  @spec print_metadata(map, keyword) :: :ok
+  def print_metadata(metadata, options \\ []) when is_map(metadata) do
```
josevalim: Can we please move this function to IEx.Introspection? We only keep in this module functionality used by both IEx and Mix and Mix is not currently using any of the metadata formatting. :)

lib/iex/lib/iex/introspection.ex
```diff
@@ -637,8 +637,9 @@ defmodule IEx.Introspection do
       {:ok, types} ->
         printed =
           for {_, {^type, _, args}} = typespec <- types do
-            doc = {format_type(typespec), type_doc(module, type, length(args))}
-            print_typespec(doc)
+            type_doc(module, type, length(args))
+            |> Tuple.insert_at(0, format_type(typespec))
+            |> print_typespec()
```
josevalim: It seems we end-up prepending `format_type/1` both times we call `type_doc/3` (here and below)... so what if we call `format_type/1` inside `type_doc/3` and skip this dance?

lib/elixir/lib/io/ansi/docs.ex
josevalim: Another advantage of moving this to introspection is what we can unify the code with https://github.com/elixir-lang/elixir/pull/7886/files#diff-9500c68886c2a36c242cdad92fe0f5e6R743. Otherwise, some of the logic you added, such as deprecated wrapping, only applies if ANSI is enabled.

josevalim: To be precise: I would just get rid of the `write_with_wrap` logic when moving this to introspection and have both behave the same (without wrapping).

lackac: We already print the docs without wrapping if ANSI is not enabled since the code simply calls `IO.puts(doc)` in that case. Moving this to `IEx.Introspection` would also require duplicating the wrapping logic there or exporting it from `IO.ANSI.Docs`.

I wouldn't want to duplicate code. Would you be okay with exporting `write_with_wrap/4`?

lackac: The code also relies on a new `:doc_metadata` option for the color, and other helpers in `IO.ANSI.Docs`. It wouldn't be clear how to configure the color if moved.

josevalim: @lackac I see. So let's keep this here but what if we also change this code to be able to handle when the color is disabled too? Basically, if color is not enabled, we don't add the color+reset. Would that be a fair compromise?

lib/elixir/lib/io/ansi/docs.ex
```diff
@@ -53,6 +55,37 @@ defmodule IO.ANSI.Docs do
...
+    print_each_metadata(metadata, options) && newline_after_block()
+  end
+
+  defp print_each_metadata(metadata, options) do
+    Enum.reduce(metadata, false, &print_single_metadata({&1, options}, &2))
+  end
+
+  defp print_single_metadata({{:deprecated, value}, options}, _printed) do
+    [metadata_label(:deprecated, options) | String.split(value, @spaces)]
+    |> write_with_wrap(options[:width], "            ", true)
```
josevalim: Instead of `"            "`, can we compute this value? Is it `String.duplicate(" ", String.length("deprecated") + 2)`? I would prefer to have those things dynamically computed on the key, similar to what @fertapric suggested, than hardcode all individual entries just because adding new binaries metadata in the future (likely to be the most common ones) will be easier and have less code duplication.

lackac: Fair enough, in this case I'm inclined to refactor and go with something similar to what @fertapric suggested.

lib/elixir/lib/io/ansi/docs.ex
lackac: It would :)

lib/elixir/lib/io/ansi/docs.ex
lackac: @fertapric implementing your suggestion after the discussion below. Wdyt about `@metadata_filter` for the attribute name?

fertapric: `@visible_metadata`, `@printable_metadata`, `@doc_metadata_keys`? I think `@doc_metadata_keys` is the one I like the most :)

lackac: @fertapric sorry, I didn't receive notification of this until after the commit. In any case I would either stay with `@metadata_filter` or use `@visible_metadata`.

fertapric: 🤔 

`@metadata_filter` can be interpreted as a whitelist or as a blacklist. For example, `@metadata_filter [:defaults]` (blacklist) and `@metadata_filter [:since, :deprecated]` (whitelist).

So, I would go with `@visible_metadata` :)

josevalim: In Elixir `filter` typically means holding the data (i.e. Enum.filter) so I think the current semantics are fine.

josevalim:

Issue #7882, opened 2018-07-12
Support floor/1 and ceil/1 guard BIFs


josevalim:
They have been added in Erlang/OTP 20+ which is a requirement in master.

PR #7888, opened 2018-07-12
Add Kernel.ceil/1 and Kernel.floor/1 buildins

lukaszsamson:
Fixes https://github.com/elixir-lang/elixir/issues/7882


josevalim:
Looks great! Two things remaining: simple tests (see kernel_test.exs) and
the spec for ceil is wrong.

-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

fertapric:
There is also `Float.ceil/2` and `Float.floor/2`, should those functions be mentioned here? Or viceversa?

josevalim:
Yeah, we should probably say that if they want floats back with a certain
amount of decimal places they should refer to the float ones.
-- 


lukaszsamson:
I added tests and fixed the spec. I also added doc references to Float functions.

michalmuskala:
Thank you! :heart:

PR #7896, opened 2018-07-13
Clarify how to configure :test_coverage

wojtekmach:
Given `ExUnit.configure` was previously mentioned in these docs, I thought that's where `:test_coverage` is to be put and was surprised that it didn't work.


josevalim:

PR #7897, opened 2018-07-13
Update --cover docs

wojtekmach:
We can't actually specify directory via command line options, need to configure it via mix configuration instead.


josevalim:

PR #7900, opened 2018-07-13
Add Function module docs to "Basic Types" group

amatalai:
At the moment `Function` module is in one group (without group?) with Kernel and Kernel.SpecialForms.

I'm not sure if "Basic Types" is correct group, but functions in this module are taking/returning other functions so function looks like kind of type to me.


josevalim:

PR #7904, opened 2018-07-14
Clarify how to assert on exits from linked processes.

favetelinguis:
Added some additional information as requested in the following [post](https://elixirforum.com/t/asserting-exit-logg-and-state-of-running-process/15148/2)

### DIFF
--- a/lib/ex_unit/lib/ex_unit/assertions.ex
+++ b/lib/ex_unit/lib/ex_unit/assertions.ex
@@ -762,11 +762,12 @@ defmodule ExUnit.Assertions do
   Asserts `expression` will exit.
   Returns the exit status/message of the current process or fails otherwise.
 
-  To assert exits from linked processes started from the test use `Process.flag(:trap_exit, true)` with `assert_received/2`.
-
   ## Examples
 
       assert catch_exit(exit 1) == 1
+      
+  To assert exits from linked processes started from the test, trap exits
+  with `Process.flag/2` and assert the exit message with `assert_received/2`.
 
       Process.flag(:trap_exit, true)
       pid = spawn_link(fn -> Process.exit(self(), :normal) end)
--- a/lib/ex_unit/lib/ex_unit/assertions.ex
+++ b/lib/ex_unit/lib/ex_unit/assertions.ex
@@ -760,14 +760,17 @@ defmodule ExUnit.Assertions do
 
   @doc """
   Asserts `expression` will exit.
-  Returns the exit status/message or fails otherwise.
+  Returns the exit status/message of the current process or fails otherwise
 
-  catch_exit is about catching an exit raised by the current process itself.
-  To assert exits from linked processes stared from the test use Process.flag(:trap_exit, true) with assert_received.
+  To assert exits from linked processes started from the test use `Process.flag(:trap_exit, true)` with `assert_received/2`.
 
   ## Examples
 
       assert catch_exit(exit 1) == 1
+
+      Process.flag(:trap_exit, true)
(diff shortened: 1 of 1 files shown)


lib/ex_unit/lib/ex_unit/assertions.ex
```diff
@@ -762,6 +762,9 @@ defmodule ExUnit.Assertions do
   Asserts `expression` will exit.
   Returns the exit status/message or fails otherwise.
 
+  catch_exit is about catching an exit raised by the current process itself.
```
fertapric: I would include the arity when mentioning functions (that’s how usually functions are identified in Elixir) and backticks (to display them as code and autolinking): `catch_exit/1` and `assert_received/2`.

In addition, I would also use backticks with the code `Process.flag(:trap_exit, true)`; and there is a typo `s/stared/started/` :)

lib/ex_unit/lib/ex_unit/assertions.ex
```diff
@@ -762,6 +762,9 @@ defmodule ExUnit.Assertions do
   Asserts `expression` will exit.
   Returns the exit status/message or fails otherwise.
```
fertapric: What do tou think about updating this line with “Returns the exit status/message of the current process or fails otherwise” and remove the next line?

lib/ex_unit/lib/ex_unit/assertions.ex
fertapric: Also, what do you think about giving a more complete example:

```elixir
Process.flag(:trap_exit, true)
pid = spawn_link(fn -> Process.exit(self(), :normal) end)
assert_receive {:EXIT, ^pid, :normal}
```

favetelinguis:
Your comments sound very good, I have made changes. Just tell me if there is anything else needed change.

lib/ex_unit/lib/ex_unit/assertions.ex
```diff
@@ -760,12 +760,18 @@ defmodule ExUnit.Assertions do
...
-  Returns the exit status/message or fails otherwise.
+  Returns the exit status/message of the current process or fails otherwise.
+
+  To assert exits from linked processes started from the test use `Process.flag(:trap_exit, true)` with `assert_received/2`.
 
   ## Examples
 
       assert catch_exit(exit 1) == 1
 
+      Process.flag(:trap_exit, true)
```
fertapric: I would put this example closer to the paragraph:

> To assert exits from linked processes started from the test use `Process.flag(:trap_exit, true)` with `assert_received/2`"

lib/ex_unit/lib/ex_unit/assertions.ex
```diff
@@ -760,12 +760,18 @@ defmodule ExUnit.Assertions do
 
   @doc """
   Asserts `expression` will exit.
-  Returns the exit status/message or fails otherwise.
+  Returns the exit status/message of the current process or fails otherwise.
+
+  To assert exits from linked processes started from the test use `Process.flag(:trap_exit, true)` with `assert_received/2`.
```
fertapric: I think this line exceeds the (hypothetical) maximum line length for comments.

josevalim:
@fertapric we usually don’t worry about style comments when reviewing
because they are not written anywhere, so it is not fair to expect from
committers to know or habdle those. In most cases, we just go ahead and fix
them before merging. :)
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

josevalim:

PR #7906, opened 2018-07-15
Minor update in Kernel's doc

fertapric:
Looking at the documentation, final dots are not included when listing items.


whatyouhide:
Thanks @fertapric 💟

PR #7909, opened 2018-07-15
Clarify documentation of @compile :inline flag

fertapric:
This is one of the things I learnt today in https://github.com/elixir-lang/elixir/pull/7905 :)


michalmuskala:
Thank you! :heart:

PR #7911, opened 2018-07-15
Changelog: explain `mix test --failed` more accurately.

myronmarston:
`mix test --failed` runs all tests that failed the last time they ran,
even if they were not included in the last run of the test suite. This
is important, because it means that you can run individual failed tests
to focus on fixing them without worrying about mix losing track of the
other failed tests.


josevalim:

PR #7914, opened 2018-07-16
Time zone behaviour and functionality (RFC/WIP)

lau:
This provides a behaviour for providing time zone data to Elixir (`lib/elixir/lib/calendar/time_zone_database.ex`). It allows anyone to implement a library that provides time zone data and use that with Elixir.

In addition to the behaviour there is code changes that enables Elixir to get a `DateTime` for the time right now in any timezone, to change between timezones and to create a new `DateTime` in any timezone. All using the behaviour defined by `TimeZoneDatabase`.

The behaviour includes functions for getting timezone information for a certain period for a certain time zone. Either by the time in UTC or the "wall time". Wall time being the time you would see on a clock on the wall, that is taking into account UTC offset as well as DST.

There are also functions for getting leap second information. This information is provided by IANA in the time zone data tar file. It is useful for validating UTC datetimes and for calculating differences where leap seconds are taken into account. It could also be used for calculating TAI and GPS time. In this PR there is currently not any functions that use the leap second data, but I might update the PR with that later.

Additionally there is functionality that uses the time zone data in `date_time.ex`.

The functions take an optional module name with a module that implement the behaviour:

```elixir
iex> DateTime.from_naive(~N[2018-10-28 02:30:00], "Europe/Copenhagen", Tzdata.TimeZoneDatabase)
{:ambiguous,
 [#DateTime<2018-10-28 02:30:00+02:00 CEST Europe/Copenhagen>,
  #DateTime<2018-10-28 02:30:00+01:00 CET Europe/Copenhagen>]}
```

Instead of providing the module as a argument, Elixir can be configured to use a specific module.

`:elixir_config.put(:time_zone_module, Tzdata.TimeZoneDatabase)`

Then the argument is not needed for the functions and the configured module will be used instead:

```elixir
iex> DateTime.from_naive(~N[2018-10-28 02:30:00], "Europe/Copenhagen")
{:ambiguous,
 [#DateTime<2018-10-28 02:30:00+02:00 CEST Europe/Copenhagen>,
  #DateTime<2018-10-28 02:30:00+01:00 CET Europe/Copenhagen>]}
```

gaps and overlaps (ambiguity) for instance during "spring forward" and autumn are handled.

```
DateTime.from_naive(~N[2018-10-28 05:30:00], "Europe/Copenhagen", Tzdata.TimeZoneDatabase)
{:ok, #DateTime<2018-10-28 05:30:00+01:00 CET Europe/Copenhagen>}
```

You can get the time right now in any timezone:

```elixir
{:ok, datetime} = DateTime.now("Europe/London")
{:ok, #DateTime<2018-07-16 18:03:15.070876+01:00 BST Europe/London>}
```

Get the same time for a datetime in another timezone:

```elixir
iex(6)> DateTime.shift_zone(datetime, "America/New_York")
{:ok, #DateTime<2018-07-16 13:03:15.070876-04:00 EDT America/New_York>}
```

This is not considered finished. Some of the details could be optimised. The main thing I focused on is the TimeZoneData behaviour. The `iex>` examples above were made with a working implementation of the behaviour not included here.

### DIFF
--- a/lib/elixir/lib/calendar/datetime.ex
+++ b/lib/elixir/lib/calendar/datetime.ex
@@ -172,7 +172,9 @@ defmodule DateTime do
   Converts the given `NaiveDateTime` to `DateTime`.
 
   It expects a time zone to put the NaiveDateTime in.
-  Currently it only supports "Etc/UTC" as time zone.
+
+  It only supports "Etc/UTC" as time zone if a TimeZoneDatabase
+  is not provided as a third argument.
 
   ## Examples
 
@@ -180,12 +182,124 @@ defmodule DateTime do
       iex> datetime
       #DateTime<2016-05-24 13:26:08.003Z>
 
+  When the datetime is ambiguous - for instance during changing from summer
+  to winter time - the two possible valid datetimes are returned. First the one
+  that happens first, then the one that happens after.
+
+      iex> {:ambiguous, [first_dt, second_dt]} = DateTime.from_naive(~N[2018-10-28 02:30:00], "Europe/Copenhagen", FakeTimeZoneDatabase)
+      iex> first_dt
+      #DateTime<2018-10-28 02:30:00+02:00 CEST Europe/Copenhagen>
+      iex> second_dt
+      #DateTime<2018-10-28 02:30:00+01:00 CET Europe/Copenhagen>
+
+  When there is a gap in wall time - for instance in spring when the clocks are
+  turned forward - the latest valid datetime just before the gap and the first
+  valid datetime just after the gap.
+
+      iex> {:gap, [just_before, just_after]} = DateTime.from_naive(~N[2019-03-31 02:30:00], "Europe/Copenhagen", FakeTimeZoneDatabase)
+      iex> just_before
+      #DateTime<2019-03-31 01:59:59.999999+01:00 CET Europe/Copenhagen>
+      iex> just_after
+      #DateTime<2019-03-31 03:00:00+02:00 CEST Europe/Copenhagen>
+
+  Most of the time there is one, and just one, valid datetime for a certain
+  date and time in a certain time zone.
+
+      iex> {:ok, datetime} = DateTime.from_naive(~N[2018-07-28 12:30:00], "Europe/Copenhagen", FakeTimeZoneDatabase)
+      iex> datetime
+      #DateTime<2018-07-28 12:30:00+02:00 CEST Europe/Copenhagen>
   """
   @doc since: "1.4.0"
-  @spec from_naive(NaiveDateTime.t(), Calendar.time_zone()) :: {:ok, t}
-  def from_naive(naive_datetime, time_zone)
+  @spec from_naive(NaiveDateTime.t(), Calendar.time_zone(), TimeZoneDatabase.t() | :from_config) ::
+          {:ok, t}
+          | {:ambiguous, [TimeZoneDatabase.time_zone_period()]}
+          | {:gap, [TimeZoneDatabase.time_zone_period()]}
+          | {:error, :time_zone_not_found}
 
-  def from_naive(%NaiveDateTime{} = naive_datetime, "Etc/UTC") do
+  def from_naive(naive_datetime, time_zone, time_zone_data_module \\ :from_config)
+
+  def from_naive(%NaiveDateTime{} = naive_datetime, "Etc/UTC", _) do
+    do_from_naive(naive_datetime, "Etc/UTC", 0, 0, "UTC")
+  end
+
+  def from_naive(%NaiveDateTime{} = naive_datetime, time_zone, time_zone_data_module) do
+    gregorian_seconds =
+      naive_datetime
+      |> NaiveDateTime.to_erl()
+      |> :calendar.datetime_to_gregorian_seconds()
+
+    case by_wall(time_zone_data_module, time_zone, gregorian_seconds) do
+      {:single, period} ->
+        do_from_naive(
+          naive_datetime,
+          time_zone,
+          period.std_offset,
+          period.utc_offset,
+          period.zone_abbr
+        )
+
+      {:ambiguous, [first_period, second_period]} ->
+        {:ok, first_datetime} =
+          do_from_naive(
+            naive_datetime,
+            time_zone,
+            first_period.std_offset,
+            first_period.utc_offset,
+            first_period.zone_abbr
+          )
+
+        {:ok, second_datetime} =
+          do_from_naive(
+            naive_datetime,
+            time_zone,
+            second_period.std_offset,
+            second_period.utc_offset,
+            second_period.zone_abbr
+          )
+
+        {:ambiguous, [first_datetime, second_datetime]}
+
+      {:gap, [first_period, second_period]} ->
+        # `until_wall` is not valid, but any time just before is.
+        # So by subtracting a second and adding .999999 seconds
+        # we get the last microsecond just before.
+        before_naive =
+          first_period.until_wall
+          |> :calendar.gregorian_seconds_to_datetime()
+          |> NaiveDateTime.from_erl!({999_999, 6})
+          |> NaiveDateTime.add(-1)
+
+        after_naive =
+          second_period.from_wall
+          |> :calendar.gregorian_seconds_to_datetime()
+          |> NaiveDateTime.from_erl!()
+
+        {:ok, latest_datetime_before} =
+          do_from_naive(
+            before_naive,
+            time_zone,
+            first_period.std_offset,
+            first_period.utc_offset,
+            first_period.zone_abbr
+          )
+
+        {:ok, first_datetime_after} =
+          do_from_naive(
+            after_naive,
+            time_zone,
+            second_period.std_offset,
+            second_period.utc_offset,
+            second_period.zone_abbr
+          )
+
+        {:gap, [latest_datetime_before, first_datetime_after]}
+
+      {:error, _} = error ->
+        error
+    end
+  end
+
+  defp do_from_naive(naive_datetime, time_zone, std_offset, utc_offset, zone_abbr) do
     %{
       calendar: calendar,
       hour: hour,
@@ -206,10 +320,10 @@ defmodule DateTime do
       minute: minute,
       second: second,
       microsecond: microsecond,
-      std_offset: 0,
-      utc_offset: 0,
-      zone_abbr: "UTC",
-      time_zone: "Etc/UTC"
+      std_offset: std_offset,
+      utc_offset: utc_offset,
+      zone_abbr: zone_abbr,
+      time_zone: time_zone
     }
 
     {:ok, datetime}
@@ -219,18 +333,21 @@ defmodule DateTime do
   Converts the given `NaiveDateTime` to `DateTime`.
 
   It expects a time zone to put the NaiveDateTime in.
-  Currently it only supports "Etc/UTC" as time zone.
 
   ## Examples
 
       iex> DateTime.from_naive!(~N[2016-05-24 13:26:08.003], "Etc/UTC")
       #DateTime<2016-05-24 13:26:08.003Z>
 
+      iex> DateTime.from_naive!(~N[2018-05-24 13:26:08.003], "Europe/Copenhagen", FakeTimeZoneDatabase)
+      #DateTime<2018-05-24 13:26:08.003+02:00 CEST Europe/Copenhagen>
+
   """
   @doc since: "1.4.0"
-  @spec from_naive!(NaiveDateTime.t(), Calendar.time_zone()) :: t
-  def from_naive!(naive_datetime, time_zone) do
-    case from_naive(naive_datetime, time_zone) do
+  @spec from_naive!(NaiveDateTime.t(), Calendar.time_zone(), TimeZoneDatabase.t() | :from_config) ::
+          t
+  def from_naive!(naive_datetime, time_zone, time_zone_data_module \\ :from_config) do
+    case from_naive(naive_datetime, time_zone, time_zone_data_module) do
       {:ok, datetime} ->
         datetime
 
@@ -238,6 +355,75 @@ defmodule DateTime do
         raise ArgumentError,
               "cannot parse #{inspect(naive_datetime)} to datetime, reason: #{inspect(reason)}"
     end
+  end
+
+  @doc """
+  Takes a DateTime and a time zone.
+
+  Returns a DateTime for the same point in time, but instead at the
+  time zone provided.
+
+      iex> cph_datetime = DateTime.from_naive!(~N[2018-07-16 12:00:00], "Europe/Copenhagen", FakeTimeZoneDatabase)
+      iex> {:ok, pacific_datetime} = DateTime.shift_zone(cph_datetime, "America/Los_Angeles", FakeTimeZoneDatabase)
+      iex> pacific_datetime
+      #DateTime<2018-07-16 03:00:00-07:00 PDT America/Los_Angeles>
+  """
+  @spec shift_zone(t, Calendar.time_zone(), TimeZoneDatabase.t() | :from_config) ::
+          {:ok, t} | {:error, :time_zone_not_found}
+  def shift_zone(
+        %{calendar: Calendar.ISO} = datetime,
+        time_zone,
+        time_zone_data_module \\ :from_config
+      ) do
+    in_utc = datetime |> to_unix |> from_unix! |> Map.put(:microsecond, datetime.microsecond)
+
+    gregorian_seconds_utc =
+      :calendar.datetime_to_gregorian_seconds(in_utc |> NaiveDateTime.to_erl())
+
+    case by_utc(time_zone_data_module, time_zone, gregorian_seconds_utc) do
+      {:ok, period} ->
+        gregorian_seconds_wall = gregorian_seconds_utc + period.utc_offset + period.std_offset
+
+        naive_datetime =
+          gregorian_seconds_wall
+          |> :calendar.gregorian_seconds_to_datetime()
+          |> NaiveDateTime.from_erl!(datetime.microsecond)
+
+        do_from_naive(
+          naive_datetime,
+          time_zone,
+          period.std_offset,
+          period.utc_offset,
+          period.zone_abbr
+        )
+
+      {:error, :time_zone_not_found} ->
+        {:error, :time_zone_not_found}
+    end
+  end
+
+  @doc """
+  Returns the current datetime in the provided time zone.
+
+  Requires a module implementing the `TimeZoneDatabase` behaviour.
+
+  ## Examples
+
+      iex> {:ok, datetime} = DateTime.now("Europe/Copenhagen", FakeTimeZoneDatabase)
+      iex> datetime.time_zone
+      "Europe/Copenhagen"
+
+  """
+  @spec now(Calendar.time_zone(), TimeZoneDatabase.t() | :from_config) ::
+          {:ok, t} | {:error, :time_zone_not_found}
+  def now(time_zone, time_zone_data_module \\ :from_config)
+
+  def now("Etc/UTC", _) do
+    {:ok, utc_now()}
+  end
+
+  def now(time_zone, time_zone_data_module) do
+    utc_now() |> shift_zone(time_zone, time_zone_data_module)
   end
 
   @doc """
@@ -898,6 +1084,46 @@ defmodule DateTime do
 
   defp apply_tz_offset(iso_days, offset) do
     Calendar.ISO.add_day_fraction_to_iso_days(iso_days, -offset, 86400)
+  end
+
+  @no_valid_time_zone_database_error "No valid TimeZoneDatabase provided or configured. Configure with :elixir_config.put(:time_zone_module, module_name)"
+  @spec by_wall(TimeZoneDatabase.t(), Calendar.time_zone(), TimeZoneDatabase.gregorian_seconds()) ::
+          {:single, TimeZoneDatabase.time_zone_period()}
+          | {:ambiguous, [TimeZoneDatabase.time_zone_period()]}
+          | {:gap, [TimeZoneDatabase.time_zone_period()]}
+          | {:error, :time_zone_not_found}
+  defp by_wall(time_zone_data_module, time_zone, gregorian_seconds) do
+    time_zone_data_module = time_zone_data_module_from_parameter(time_zone_data_module)
+
+    try do
+      time_zone_data_module.by_wall(time_zone, gregorian_seconds)
+    rescue
+      UndefinedFunctionError ->
+        raise @no_valid_time_zone_database_error
+    end
+  end
+
+  @spec by_utc(TimeZoneDatabase.t(), Calendar.time_zone(), TimeZoneDatabase.gregorian_seconds()) ::
+          {:ok, TimeZoneDatabase.time_zone_period()} | {:error, :time_zone_not_found}
+  defp by_utc(time_zone_data_module, time_zone, gregorian_seconds) do
+    time_zone_data_module = time_zone_data_module_from_parameter(time_zone_data_module)
+
+    try do
+      time_zone_data_module.by_utc(time_zone, gregorian_seconds)
+    rescue
+      UndefinedFunctionError ->
+        raise @no_valid_time_zone_database_error
+    end
+  end
+
+  @spec time_zone_data_module_from_parameter(:from_config | TimeZoneDatabase.t()) ::
+          TimeZoneDatabase.t()
+  defp time_zone_data_module_from_parameter(:from_config) do
+    :elixir_config.safe_get(:time_zone_module, :from_config)
+  end
+
+  defp time_zone_data_module_from_parameter(time_zone_data_module) do
+    time_zone_data_module
   end
 
   defimpl String.Chars do
--- a/lib/elixir/lib/calendar/datetime.ex
+++ b/lib/elixir/lib/calendar/datetime.ex
@@ -544,8 +544,8 @@ defmodule DateTime do
           period.zone_abbr
         )
 
-      {:error, :time_zone_not_found} ->
-        {:error, :time_zone_not_found}
+      {:error, _} = error ->
+        error
     end
   end
 
@@ -587,6 +587,8 @@ defmodule DateTime do
       iex> {:ok, datetime} = DateTime.now("Europe/Copenhagen", FakeTimeZoneDatabase)
       iex> datetime.time_zone
       "Europe/Copenhagen"
+      iex> DateTime.now("not a real time zone name", FakeTimeZoneDatabase)
+      {:error, :time_zone_not_found}
 
   """
   @spec now(Calendar.time_zone(), TimeZoneDatabaseClient.tz_db_or_config()) ::
--- a/lib/elixir/lib/calendar/datetime.ex
+++ b/lib/elixir/lib/calendar/datetime.ex
@@ -260,13 +260,11 @@ defmodule DateTime do
         # we get the last microsecond just before.
         before_naive =
           first_period.until_wall
-          |> Calendar.ISO.iso_seconds_to_datetime()
           |> NaiveDateTime.from_erl!({999_999, 6})
           |> NaiveDateTime.add(-1)
 
         after_naive =
           second_period.from_wall
-          |> Calendar.ISO.iso_seconds_to_datetime()
           |> NaiveDateTime.from_erl!()
 
         {:ok, latest_datetime_before} =
@@ -365,23 +363,26 @@ defmodule DateTime do
   """
   @spec shift_zone(t, Calendar.time_zone(), TimeZoneDatabase.t() | :from_config) ::
           {:ok, t} | {:error, :time_zone_not_found}
+  def shift_zone(datetime, time_zone, time_zone_database \\ :from_config)
+
+  def shift_zone(%{time_zone: time_zone} = datetime, time_zone, _) do
+    # When the desired time_zone is the same as the existing time_zone just
+    # return the passed datetime unchanged.
+    {:ok, datetime}
+  end
+
   def shift_zone(
         %{calendar: Calendar.ISO} = datetime,
         time_zone,
-        time_zone_data_module \\ :from_config
+        time_zone_database
       ) do
-    in_utc = datetime |> to_unix |> from_unix! |> Map.put(:microsecond, datetime.microsecond)
+    in_utc = to_zero_total_offset(datetime)
 
-    iso_seconds_utc = Calendar.ISO.datetime_to_iso_seconds(in_utc |> NaiveDateTime.to_erl())
-
-    case by_utc(time_zone_data_module, time_zone, in_utc) do
+    case by_utc(time_zone_database, time_zone, in_utc) do
       {:ok, period} ->
-        iso_seconds_wall = iso_seconds_utc + period.utc_offset + period.std_offset
-
         naive_datetime =
-          iso_seconds_wall
-          |> Calendar.ISO.iso_seconds_to_datetime()
-          |> NaiveDateTime.from_erl!(datetime.microsecond)
+          in_utc
+          |> NaiveDateTime.add(period.utc_offset + period.std_offset)
 
         do_from_naive(
           naive_datetime,
@@ -394,6 +395,17 @@ defmodule DateTime do
       {:error, :time_zone_not_found} ->
         {:error, :time_zone_not_found}
     end
+  end
+
+  # Takes Calendar.naive_datetime and makes sure it has a zero total offset
+  @spec to_zero_total_offset(Calendar.naive_datetime()) :: Calendar.naive_datetime()
+  defp to_zero_total_offset(%{utc_offset: utc_offset, std_offset: std_offset} = datetime)
+       when utc_offset + std_offset == 0 do
+    datetime
+  end
+
+  defp to_zero_total_offset(%{calendar: Calendar.ISO} = datetime) do
+    datetime |> NaiveDateTime.add(-1 * (datetime.utc_offset + datetime.std_offset))
   end
 
   @doc """
@@ -410,14 +422,14 @@ defmodule DateTime do
   """
   @spec now(Calendar.time_zone(), TimeZoneDatabase.t() | :from_config) ::
           {:ok, t} | {:error, :time_zone_not_found}
-  def now(time_zone, time_zone_data_module \\ :from_config)
+  def now(time_zone, time_zone_database \\ :from_config)
 
   def now("Etc/UTC", _) do
     {:ok, utc_now()}
   end
 
-  def now(time_zone, time_zone_data_module) do
-    utc_now() |> shift_zone(time_zone, time_zone_data_module)
+  def now(time_zone, time_zone_database) do
+    utc_now() |> shift_zone(time_zone, time_zone_database)
   end
 
   @doc """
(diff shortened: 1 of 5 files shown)


michalmuskala:
I really like where this is going 👏. Thank you for working on this! Here are couple design considerations when it comes to the time zone database interface - I haven't looked at the functions in `DateTime` too closely yet.

* We should probably use the `Calendar.iso_days` abstraction instead of gregorian seconds in the interfaces - this will keep consistency with other elements of the calendar system and should limit required conversions.

* The `:elixir_config` module is private - we should probably have some wrapper for it or just use the application env.

* In the data structure returned from the database callbacks there are fields that are not used in the actual implementation. The implementation only uses the values of `utc_offset`, `std_offset`, and `zone_abbr`. I think we should keep the interface minimal and not include the other data, if we don't need it. An option could be also to use tuples instead of maps, but I'm not sure about that one - it might limit us on extensibility in the future.

* Some functions, when information is lacking, return a list of possible values. Is it actually a list or can there always be just two values? If it's just two values, I think it should return something like `{:gap, lower, higher}` instead.

* I think, by default, we might provide a database module that has only information for UTC and raises nice errors for other timezones. This should simplify some code paths.

* I'm not sure about the `leap_seconds` callback. It seems to me that the information we want to obtain there is - "What is the difference between TAI and UTC at the specified point in time". If so, I think this function should receive a datetime value (probably in the "iso days" format) and return the integer offset. If we go with that approach, it might return: `{:ok, non_neg_integer()} | {:incomplete_data, offset}` or something similar, if requested for data too far into the future.

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -177,20 +177,134 @@ defmodule DateTime do
...
+      iex> {:ok, datetime} = DateTime.from_naive(~N[2018-07-28 12:30:00], "Europe/Copenhagen", FakeTimeZoneDatabase)
+      iex> datetime
+      #DateTime<2018-07-28 12:30:00+02:00 CEST Europe/Copenhagen>
   """
   @doc since: "1.4.0"
-  @spec from_naive(NaiveDateTime.t(), Calendar.time_zone()) :: {:ok, t}
-  def from_naive(naive_datetime, time_zone)
+  @spec from_naive(NaiveDateTime.t(), Calendar.time_zone(), TimeZoneDatabase.t() | :from_config) ::
+          {:ok, t}
+          | {:ambiguous, [TimeZoneDatabase.time_zone_period()]}
```
fertapric: I think this should be `{:ambiguous, [t]}` and `{:gap, [t]}`

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -177,20 +177,134 @@ defmodule DateTime do
   Converts the given `NaiveDateTime` to `DateTime`.
 
   It expects a time zone to put the NaiveDateTime in.
-  Currently it only supports "Etc/UTC" as time zone.
+
+  It only supports "Etc/UTC" as time zone if a TimeZoneDatabase
```
fertapric: I would include backticks for autolinking in \`TimezoneDatabase\`

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -177,20 +177,134 @@ defmodule DateTime do
...
+      #DateTime<2019-03-31 01:59:59.999999+01:00 CET Europe/Copenhagen>
+      iex> just_after
+      #DateTime<2019-03-31 03:00:00+02:00 CEST Europe/Copenhagen>
+
+  Most of the time there is one, and just one, valid datetime for a certain
+  date and time in a certain time zone.
+
+      iex> {:ok, datetime} = DateTime.from_naive(~N[2018-07-28 12:30:00], "Europe/Copenhagen", FakeTimeZoneDatabase)
+      iex> datetime
+      #DateTime<2018-07-28 12:30:00+02:00 CEST Europe/Copenhagen>
```
fertapric: I would include a new line after the examples :)

lib/elixir/lib/calendar/time_zone_database.ex
```diff
@@ -0,0 +1,92 @@
...
+          zone_abbr: Calendar.zone_abbr(),
+          from_utc: time_zone_period_limit,
+          from_wall: time_zone_period_limit,
+          until_utc: time_zone_period_limit,
+          until_wall: time_zone_period_limit
+        }
+
+  @doc """
+  Takes a time zone name and a point in time for UTC and returns a
+  `time_zone_period` for that point in time.
```
fertapric: I think you have to use `t:time_zone_period/0` for linking types

lib/elixir/lib/calendar/time_zone_database.ex
```diff
@@ -0,0 +1,92 @@
...
+  """
+  @callback by_utc(Calendar.time_zone(), gregorian_seconds) ::
+              {:ok, time_zone_period} | {:error, :time_zone_not_found}
+
+  @doc """
```
fertapric: I think the first line should be a summary of the function.

lib/elixir/lib/calendar/time_zone_database.ex
```diff
@@ -0,0 +1,92 @@
...
+  and the `time_zone_period` is returned.
+  """
+  @callback by_wall(Calendar.time_zone(), gregorian_seconds) ::
+              {:single, time_zone_period}
+              | {:ambiguous, [time_zone_period]}
+              | {:gap, [time_zone_period]}
+              | {:error, :time_zone_not_found}
+
+  @doc """
+  Returns a list of all known leap seconds. Each element in the list is a tuple
```
fertapric: I would add a new line after the first sentence, so `Returns a list of all known leap seconds.` becomes the summary.

lib/elixir/lib/calendar/time_zone_database.ex
```diff
@@ -0,0 +1,92 @@
...
+  with the first element being the UTC datetime for the leap second and the
+  second element being the difference between TAI and UTC in seconds. (TAI-UTC)
+  """
+  @callback leap_seconds() :: [{:calendar.datetime(), integer()}]
+
+  @doc """
+  Returns a datetime tuple with the UTC datetime for when the leap second
```
fertapric: datetime map/struct?

lib/elixir/lib/calendar/time_zone_database.ex
```diff
@@ -0,0 +1,92 @@
...
+
+  @doc """
+  When the provided `gregorian_seconds` is ambiguous for the datetime a tuple with `:ambiguous` and a list of two possible
+  periods. The periods in the list are sorted with the first element being the one that begins first.
+
+  When the provided `gregorian_seconds` datetime is in a gap - for instance during the "spring forward" when going
+  from winter time to summer time, a tuple with `:gap` and a list of two time zone periods are returned. The first
+  period in the list is the period before the gap and the second period is the period just after the gap.
+
+  If there is only a single possible period for the provided `gregorian_seconds`, the a tuple with `:single`
```
fertapric: Typo `s/the a tuple/then a tuple`

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -245,6 +362,75 @@ defmodule DateTime do
     end
   end
 
+  @doc """
+  Takes a DateTime and a time zone.
+
+  Returns a DateTime for the same point in time, but instead at the
```
fertapric: I would make this line the first one so it becomes the summary

fertapric: I would also include backticks in DateTime and reference the variable (and some rewording):

```
Returns a `DateTime` for the same point in time in the given `time_zone`.
```

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -245,6 +362,75 @@ defmodule DateTime do
...
+  time zone provided.
+
```
fertapric: I would include `## Examples` header and new line after the examples

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -245,6 +362,75 @@ defmodule DateTime do
...
+      #DateTime<2018-07-16 03:00:00-07:00 PDT America/Los_Angeles>
+  """
+  @spec shift_zone(t, Calendar.time_zone(), TimeZoneDatabase.t() | :from_config) ::
+          {:ok, t} | {:error, :time_zone_not_found}
+  def shift_zone(
+        %{calendar: Calendar.ISO} = datetime,
+        time_zone,
+        time_zone_data_module \\ :from_config
+      ) do
+    in_utc = datetime |> to_unix |> from_unix! |> Map.put(:microsecond, datetime.microsecond)
```
fertapric: Looking at this dance `to_unix |> from_unix!`, what do you think about including a `to_utc/1`?

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -245,6 +362,75 @@ defmodule DateTime do
...
+
+    gregorian_seconds_utc =
+      :calendar.datetime_to_gregorian_seconds(in_utc |> NaiveDateTime.to_erl())
```
fertapric: I would write this either:

```elixir
gregorian_seconds_utc = :calendar.datetime_to_gregorian_seconds(NaiveDateTime.to_erl(in_utc))
```

or 

```elixir
gregorian_seconds_utc =
  in_utc
  |> NaiveDateTime.to_erl()
  |> :calendar.datetime_to_gregorian_seconds()
```

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -177,20 +177,134 @@ defmodule DateTime do
...
+            after_naive,
+            time_zone,
+            second_period.std_offset,
+            second_period.utc_offset,
+            second_period.zone_abbr
+          )
+
+        {:gap, [latest_datetime_before, first_datetime_after]}
+
+      {:error, _} = error ->
```
fertapric: I would match here on `:time_zone_not_found` like you did on `shift_zone/3`

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -177,20 +177,134 @@ defmodule DateTime do
...
   @doc since: "1.4.0"
-  @spec from_naive(NaiveDateTime.t(), Calendar.time_zone()) :: {:ok, t}
-  def from_naive(naive_datetime, time_zone)
+  @spec from_naive(NaiveDateTime.t(), Calendar.time_zone(), TimeZoneDatabase.t() | :from_config) ::
+          {:ok, t}
+          | {:ambiguous, [TimeZoneDatabase.time_zone_period()]}
+          | {:gap, [TimeZoneDatabase.time_zone_period()]}
+          | {:error, :time_zone_not_found}
+
+  def from_naive(naive_datetime, time_zone, time_zone_data_module \\ :from_config)
```
fertapric: Maybe drop the `_module` suffix and use `time_zone_database`?

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -245,6 +362,75 @@ defmodule DateTime do
...
+  @spec now(Calendar.time_zone(), TimeZoneDatabase.t() | :from_config) ::
+          {:ok, t} | {:error, :time_zone_not_found}
+  def now(time_zone, time_zone_data_module \\ :from_config)
+
+  def now("Etc/UTC", _) do
+    {:ok, utc_now()}
+  end
+
+  def now(time_zone, time_zone_data_module) do
+    utc_now() |> shift_zone(time_zone, time_zone_data_module)
```
fertapric: I think single pipes are discouraged, maybe `shift_zone(utc_now(), time_zone, time_zone_data_module)`?

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -245,6 +362,75 @@ defmodule DateTime do
...
+          period.zone_abbr
+        )
+
+      {:error, :time_zone_not_found} ->
+        {:error, :time_zone_not_found}
+    end
+  end
+
+  @doc """
+  Returns the current datetime in the provided time zone.
```
fertapric: I would reference the variable:

```
Returns the current datetime in the provided `time_zone`.
```

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -889,6 +1075,46 @@ defmodule DateTime do
...
+          | {:ambiguous, [TimeZoneDatabase.time_zone_period()]}
+          | {:gap, [TimeZoneDatabase.time_zone_period()]}
+          | {:error, :time_zone_not_found}
+  defp by_wall(time_zone_data_module, time_zone, gregorian_seconds) do
+    time_zone_data_module = time_zone_data_module_from_parameter(time_zone_data_module)
+
+    try do
+      time_zone_data_module.by_wall(time_zone, gregorian_seconds)
+    rescue
+      UndefinedFunctionError ->
```
fertapric: I might move these checks to the function clause `time_zone_data_module_from_parameter(:from_config)`:

```elixir
defp time_zone_data_module_from_parameter(:from_config) do
  case :elixir_config.safe_get(:time_zone_module, :from_config) do
    time_zone_data_module -> time_zone_module
    nil -> raise "<A good error message>"
  end
end
```

If a `time_zone_data_module` is provided, it should implement the `TimeZoneDatabase` behaviour. In those cases, I think the error is quite easy to understand: `** (UndefinedFunctionError) function MyTimeZoneDatabase. by_utc/2 is undefined or private`.

Beside, I would use the same name for the variable (now `time_zone_data_module`) and for the key in the configuration (now `time_zone_module`).

lib/elixir/lib/calendar/time_zone_database.ex
fertapric: Maybe clarify that this datetime is not an Elixir's `DateTime` map/struct saying something like `Returns an Erlang datetime tuple`?

On the other hand, being this defined in the Elixir world, what are the benefits of returning `:calendar.datetime()` instead of `Calendar.datetime()`?

josevalim:
Since this is an earlier proposal, we should focus on reviewing the code and its features. Once we agree on the API and everything else, we can do a more thorough review on the docs.  :)

lau:
Thank you for the feedback @michalmuskala 

## We should probably use the Calendar.iso_days abstraction instead of gregorian seconds in the interfaces - this will keep consistency with other elements of the calendar system and should limit required conversions.

A nice thing about gregorian seconds is that it is a simple integer and it fits well with the time zone data. The tzdata from IANA is not calendar agnostic -- it is in the proleptic Gregorian calendar (which Calendar.ISO implements). Furthermore the level of detail is at the seconds level. So implementing the database and querying can be done for instance by comparing integers in a guard clause in a macro or in an ETS query. Imagine that you have a UNIX timestamp (1531785600) and want to find out the time for that in a certain timezone. All you have to do is: `TimeZoneDatabase.by_utc("Europe/Copenhagen", 1531785600 + 62167219200)`. And the database can use the same format internally.

However, in the examples I have called the functions in the Erlang calendar module directly for converting to and from gregorian seconds. To avoid that we can instead use functions in `Calendar.ISO` module. There is already a private function for converting iso seconds (ie. gregorian seconds) to a datetime. That could be made public and a reverse could be made: datetime to iso seconds.

This way we avoid adding complexity to the database API and the database implementations.

## The :elixir_config module is private - we should probably have some wrapper for it or just use the application env.

Yeah the calling :elixir_config directly part is one of the details that aren't ready yet and was just to have something simple working.

## In the data structure returned from the database callbacks there are fields that are not used in the actual implementation. The implementation only uses the values of utc_offset, std_offset, and zone_abbr. I think we should keep the interface minimal and not include the other data, if we don't need it. An option could be also to use tuples instead of maps, but I'm not sure about that one - it might limit us on extensibility in the future.

The other ones are used for when there is a gap. Then it is nice to know what the limits of the gap are. However currently only `from_wall` and `until_wall` are used in the current functionality, so `from_utc` and `until_utc` could be dropped. I considered make two kinds of periods: one with the limits and one without. (For simplicity in the first draft I stuck with a single period type.) For the `by_utc` functions the "lighter" periods can be returned. For `by_wall` - and then maybe only when there is a gap - the "fuller" version that includes the limits can be returned.

## Some functions, when information is lacking, return a list of possible values. Is it actually a list or can there always be just two values? If it's just two values, I think it should return something like {:gap, lower, higher} instead.

Yes, there are always two periods if it is ambiguous or there is a gap. Having them directly in the tuple sounds good.

## I think, by default, we might provide a database module that has only information for UTC and raises nice errors for other timezones. This should simplify some code paths.

Regarding a mock database that only has UTC, I could see that having some advantages, but we might want to have special shortcuts for UTC anyway even when there is a proper database. And then for the user how would that be explained? "You don't have a database (but internally there really is one anyway)" or "You haven't configured a database, so there is this mock database that doesn't really have any data". I think there are advantages to saying: "error, you have not configured a database". And the concept of a lack of a database being more simple than a mock one. If someone tries to get the time for a timezone when there is no timezone database the error should probably be something like "database missing" instead of "that timezone doesn't exist - well maybe it does, but this is a mock database".

## I'm not sure about the leap_seconds callback. It seems to me that the information we want to obtain there is - "What is the difference between TAI and UTC at the specified point in time". If so, I think this function should receive a datetime value (probably in the "iso days" format) and return the integer offset. If we go with that approach, it might return: {:ok, non_neg_integer()} | {:incomplete_data, offset} or something similar, if requested for data too far into the future.

We also want to know for instance "is this a leap second"? "What is the net leap second difference between these two datetimes"? There will be more functions in the API. If we want more or a different functionality in the future then the API will have to change. If we just have all of the leap seconds, then the API won't have to change, but of course it adds more complication to the Elixir (TimeZoneDatabase client) side. It might make sense to implement the features we want now and then see what features are needed and see what would be needed for those in the API. I will do some work on that.

josevalim:
> A nice thing about gregorian seconds is that it is a simple integer and it fits well with the time zone data. The tzdata from IANA is not calendar agnostic 

We can likely still pass iso_days to the database and the database can then convert it to gregorian seconds and perform the lookups. We should not mix the contract with what is the most convenient implementation. :)

In particular, we have removed all the dependencies we had in the `:calendar` module by our own contract, which is iso_days, so we should not add calls to `:calendar` back unless it is happening in the time zone database implementation.

lib/elixir/lib/calendar/datetime.ex
lau: Yes, this was a quick way to have something working. It can be optimised in various ways. For instance if the DateTime total offset (UTC + standard) is 0 then it does not have to be changed.

lau:
I switched from ISO seconds to Erlang style datetime tuples (`:calendar.datetime`). This means less conversion on the Elixir standard library side because the `DateTime` and `NaiveDateTime` structs already have all of the elements of such a datetime tuple.

Non-ISO calendars are now supported in the time zone functions. They have to be compatible (convertible) with the ISO calendar because the time zones are defined in ISO terms.

josevalim:
Thanks @lau!

We worked really hard to no longer depend on Erlang's `:calendar` library, as that caused a confusion of formats. Sometimes we were passing tuples, sometimes we were passing structs and sometimes we were passing iso days. So I wouldn't like to add dependency on the Erlang formats back.

You have already said that the timezone database would compare integers, as that is the most efficient format, and the iso days already get you half way there without wasting CPU work, so why not rely on iso_days?

lau:
@josevalim 

To be more clear about what I wrote the other day:

> However, in the examples I have called the functions in the Erlang calendar module directly for converting to and from Gregorian seconds. To avoid that we can instead use functions in Calendar.ISO module. 

Ie. we don't have to depend on the Gregorian seconds functions in Erlang's :calendar module even if using ISO seconds. One of the functions already has a pure Elixir implementation in Calendar.ISO (https://github.com/elixir-lang/elixir/blob/c114300727659dfb20109ff97dce43b87ed7fd10/lib/elixir/lib/calendar/iso.ex#L802). I tried using that and added the reverse function: datetime to iso seconds. That worked fine. But then I tried the Erlang style datetime instead of ISO seconds and that was simpler. Just like the ISO seconds solution, the current implementation with datetime tuples does not use any functions from the Erlang :calendar module.

ISO days (with the fraction needed) is not an integer, but rather a nested tuple with different integers. So the TimeZoneDatabase implementation would probably still want to convert it to an integer like ISO seconds. If you provide a `DateTime` or `NaiveDateTime` to a function that needs to convert it, you already have the year, month, hour etc. and just need to take those values and put in a tuple.

ISO days seems like a useful solution for converting between calendars, but this is all Gregorian. The time zones are defined in Gregorian terms in the database. All of them would look like this: `{_, {_, 86400000000}}` and `86400000000` would always have to be the same and that would be sent back and forth billions of times. As I see it, it is an extra step: going from the values such as `year, month, ... second`, calculating the equivalent ISO days tuple, converting that to whatever the TimeZoneDatabase uses internally (maybe something like ISO seconds).

The Erlang style tuples (e.g. `DateTime.utc_now |> NaiveDateTime.to_erl`) are simple, easy to read and reason about and the values are already available in the structs passed to the functions. ISO seconds are compact and useful for comparisons.

lau:
@josevalim Thank you for your thoughts on it. I enjoy discussing the rationale behind it to exchange ideas on it :) It am not dead set on any particular type.

When it comes to supporting leap seconds, we would want a type that supports leap seconds. For instance you can do `~N[2016-12-31 23:59:60] |> NaiveDateTime.to_erl |> NaiveDateTime.from_erl!` and get back to `~N[2016-12-31 23:59:60]`.

With ISO days I don't get the same thing back:

```elixir
iex> Calendar.ISO.naive_datetime_to_iso_days(2016, 12, 31, 23, 59, 60, {0, 0}) |> Calendar.ISO.naive_datetime_from_iso_days
{2016, 12, 31, 24, 0, 0, {0, 6}}
```

Should we use ISO days for the time zones and some kind of tuple or map that supports leap seconds for leap seconds? I also remember talk from a long time ago about a NaiveDateTime struct, but just with the struct part removed or ignored. Basically the `Calendar.naive_datetime` type.

BTW I saw that `:calendar.datetime` is used in `File.touch/2`. Should that also accept a UTC `DateTime` struct in addition to the tuple? And have the use of a tuple deprecated there?

P.S. @michalmuskala I forgot to mention that I moved the results out of the list for gaps and ambiguous as you suggested. E.g: `{:gap, lower, higher}`.

josevalim:
> BTW I saw that :calendar.datetime is used in File.touch/2. Should that also accept a UTC DateTime struct in addition to the tuple? And have the use of a tuple deprecated there?

We will need to do something eventually but that's a separate discussion since `:calendar.datetime` is part of the contract there and we need to remain backwards compatible (i.e. it was added before we had calendar types).

lau:
@josevalim Yes, `Calendar.naive_datetime`s are used instead of Erlang style tuples. For now Erlang style tuples are used internally in the FakeTimeZoneDatabase used in tests. 

Some more details about leap second validation:

The leap second data is good until a certain time. For the example for June 2018 we get `{:error, :invalid_leap_second}` because the tzdata knows that there was no leap second.

If you try to create a DateTime with the second being 60 and it is beyond the limit for the known leap seconds, this will be the result:

```elixir
iex> DateTime.from_naive(~N[2090-06-30 23:59:60], "Etc/UTC")
{:error, :outside_leap_second_data_validity_range}
```

There might be a leap second in June in the year 2090, but we don't know, because it is very far into the future and has not been decided yet by [IERS](https://www.iers.org/).

josevalim:
@lau what about `DateTime.diff`? What does it happen when we give it leap seconds? For example:

    DateTime.diff(beginning_of_the_day_with_leapseconds, end_of_the_day_with_leapseconds)

Would it return 86400? Because, if it does, then this won't return true:

    diff = DateTime.diff(beginning_of_the_day_with_leapseconds, end_of_the_day_with_leapseconds)
    DateTime.add(beginning_of_the_day_with_leapseconds, diff) == end_of_the_day_with_leapseconds

I think it is best if we do not take leap seconds into account when converting from and to iso. This means that `end_of_the_day_with_leapseconds` is a valid instance but when we compute the diff, the leap seconds is not taken into account. Otherwise we need to make sure all addition/diff operations consider them.

Thoughts?

lau:
@josevalim I think restricting leap second consideration to only be allowed for Calendar.ISO datetimes makes sense. At least initially.

josevalim:
@lau I am not sure if that answer applies to my latest comment or an earlier one. Can you please re-read my last comment and let me know if I need to clarify something? I have asked a different question, which is about considering leap seconds in the DateTime.diff and DateTime.add operations. Thanks.

lau:
@josevalim Sure. I've updated it to clarify. Yes, I meant for diff and add functions. I'm working on functions for those now.

lau:
A callback `leap_second_diff(Calendar.naive_datetime(), Calendar.naive_datetime())` has been added to the `TimeZoneDatabase` behaviour.

lib/elixir/lib/calendar/naive_datetime.ex
```diff
@@ -270,7 +270,7 @@ defmodule NaiveDateTime do
   """
   @doc since: "1.4.0"
   @spec add(t, integer, System.time_unit()) :: t
-  def add(%NaiveDateTime{} = naive_datetime, integer, unit \\ :second)
+  def add(naive_datetime, integer, unit \\ :second)
```
josevalim: Why this change? I believe the checking for NaiveDateTime is intentional due to the inherent loss of precision.

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -177,20 +177,209 @@ defmodule DateTime do
...
+          | {:outside_leap_second_data_range, t}
+          | {:ambiguous, t, t}
+          | {:gap, t, t}
+          | {:error, :time_zone_not_found}
+          | {:error, :incompatible_calendars}
+          | {:error, :no_time_zone_database}
+
+  def from_naive(naive_datetime, time_zone, tz_db_or_config \\ :from_config)
+
+  def from_naive(%{second: 60} = naive_datetime, "Etc/UTC", tz_db_or_config) do
```
josevalim: This is specific to the ISO calendar. I think we will have to move the leap seconds handling to the ISO calendar but I am not quite sure about how to do that.

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -245,6 +440,149 @@ defmodule DateTime do
...
+        time_zone,
+        tz_db_or_config
+      ) do
+    in_utc = to_zero_total_offset(datetime)
+
+    case TimeZoneDatabaseClient.by_utc(in_utc, time_zone, tz_db_or_config) do
+      {:ok, period} ->
+        naive_datetime =
+          in_utc
+          |> naive_add_preserve_second_60(period.utc_offset + period.std_offset)
```
josevalim: I think this is very inefficient. to_zero_total_offset converts to iso days, then back to naive, then we get the period, convert back to iso days, adds the offsets, and convert back to a datetime again.

josevalim:
Hi @l4u! I have added some comments but I think it would be best if we could Skype and discuss some of those things together. I am thinking it is best if the `by_utc`API is actually `by_iso_days`. This would allow us to implement operations like `shift_zone` and `add` much more efficient then what we have right now. I want to understand the `by_wall` usage though.

Also, would you be ok if give me commit access to your fork? Then we can work on this together. Thanks!

josevalim:
@l4u thanks. :)

lib/elixir/lib/calendar/naive_datetime.ex
lau: The same reason you can pass a DateTime or NaiveDateTime to `Date.leap_year?/1` or `Date.to_iso8601` etc. We started to change the functions to accept e.g. `Calendar.date` instead of only `Date`. But unfortunately some functions were not checked/changed. For consistency we can change the rest.

In this case I see that in the specs `t` should be changed to `Calendar.naive_datetime`.

josevalim: We agreed though, and I think it was your suggestion, to not allow them in `add/2` because of the inherent loss of precision that happens when you pass a `DateTime` to `NaiveDateTime.add`. For example offsets are not considered. So mostly we allow `Calendar.naive_datetime` on read-only operations and the struct on write ones.

Still, I think if we want to remove this restriction, we should probably do it consistently and in another pull request, as this one is already getting quite large. :)

lau:
@josevalim Sure, let's talk on Skype. I've added you to my fork.

josevalim:
I had another talk with @lau and we planned the following changes:

1. Rename by_wall to something clearer, such as `time_zone_period_from_wall_time`
2. Rename by_utc to something clearer and make it receive iso_days, something like `time_zone_period_from_iso_days`
3. Unify the time zone period types and convert the `:gap` tuple to a five element tuple

@lau you will have to rebase again, apologies. :)

lau:
@josevalim 
The functions/callbacks have been renamed.

The time_zone_period types have also been renamed.  While the idea of using a 5 tuple for the wall time limits initially sounded good, I did not think about that those limits are different for each period. So a 5 tuple element would not be enough. The wall time limits really belong to each period. Now they are called `time_zone_period` and `time_zone_period_with_wall_limits`.

josevalim:
Hi @lau! I will review this for merge. I will drop comments here, for guidance, but you don't have to worry about them, as I will tackle them post merge. If I have any questions I will ping you.

josevalim:
I also want to add that, after a IRC convo in July, @lau and I agreed to not build leap seconds or consider leap seconds in the add/diff APIs. That's because if we start building leap seconds and handing off to other APIs, they can flat-out deny leap seconds. So we will follow on the footsteps of JodaTime and parse leap seconds but convert them to the previous second.

lau:
> I also want to add that, after a IRC convo in July, @lau and I agreed to not build leap seconds or consider leap seconds in the add/diff APIs. That's because if we start building leap seconds and handing off to other APIs, they can flat-out deny leap seconds. So we will follow on the footsteps of JodaTime and parse leap seconds but convert them to the previous second.

Yes, and I would just like to repeat that this ignoring leap seconds was just about the default behaviour standard add/diff functions. So that leap seconds will be ignored for those functions by default. E.g. if you pass a positive leap second to an add function or diff function it will be treated as the second before. And any leap second between the two datetimes will be ignored in the calculation.

Parsing of strings to DateTime/NaiveDateTime/Time structs should still correctly parse the perfectly valid datetime. E.g. `DateTime.from_iso8601("2016-12-31T23:59:60Z")` should still return
`{:ok, #DateTime<2016-12-31 23:59:60Z>, 0}` as it does today.

Regarding talking to APIs that are not able to handle valid ISO datetimes if they are leap seconds - this should not be a problem as long as the defaults are used. Most computers today do not create timestamps with leap seconds anyway. Even `DateTime.utc_now` does not.

Now for something we also discussed adding to the API at some point:
We also talked about a separate function to calculate leap second diffs `DateTime.leap_seconds_between` which could then be added to the result in order to calculate diffs/adds taking leap seconds into account. In most cases with a small interval this function would return 0. But the absolute result of this if the diff is between two datetimes many years apart e.g. between the years 2014 and 2004 the result would be 3 seconds. For a `diff` these 3 seconds could simply be added to a result of a `diff` and the result in seconds would be 3 more. So far so good.

However now that I think about an add operation where we want to take leap seconds into account: Again this is a situation where a user is not using the defaults of an `add` function, but actually wants to do calculations where leap seconds are taken into account. First we use `DateTime.leap_seconds_between` and get a number of seconds, then add that to another amount of seconds before passing it to `DateTime.add`: The add function would have to take the leap seconds into account in order to get the right result in case the result should be a leap second. E.g. `2016-12-31 23:59:60` is a leap second so adding 1 to `2016-12-31 23:59:59` we want a function to return `2016-12-31 23:59:60`. So for adding I am not sure that a `DateTime.leap_seconds_between` function is enough. Some ideas are either a boolean in the add function `take_leap_second_into_account \\ false` or adding a separate function e.g. `DateTime.add_with_leap_seconds` function that works just like `DateTime.add` except it takes leap seconds into account.

lib/elixir/lib/calendar/datetime.ex
josevalim: Stream of thoughts:

The fact we may have negative leap seconds means that we also need to check for `%{second: 59}` and see if it actually exists. I understand that we haven't had a negative leap second yet but not handling those cases (or not thinking explicitly about them) may mean two things:

1. whenever they happen, old elixir versions will be unable to cope with them
2. our API may be unable to be augmented whenever they are introduced

My suggestion is for leap seconds to be layered into DateTime somehow. This can either be `DateTimeWithLeapSecond` (or a new field in DateTime) or a `Calendar.ISOWithLeapSecond`. The latter would mean that `Calendar.ISO` today should have been called `Calendar.ISOWithoutLeapSecond` but unfortunately this ship has sailed. I will have to think more about this but this is inlined to other libraries where leap seconds are opted in.

lau: > The fact we may have negative leap seconds

From what I have read in various places it is unlikely that a negative leap second will occur anytime soon.

From https://web.archive.org/web/20130602095155/http://iraf.noao.edu/~seaman/leap/leapsec.txt :

> One thing to note is the steep slope resulting from initializing UT to the
 old Ephemeris Time definition of so many seconds in 1900.  This slope all
 but guarantees that a negative leap second will never occur - not only would
 the Earth's current slowing trend have to be combatted, but the accumulated
 bias over a coherent 1.5 year timescale would have to be reversed.  We don't
 have leap seconds because of current slowing - we have leap seconds because
 the Earth has already spun down.

josevalim: Awesome, thanks for the info!

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -172,20 +172,213 @@ defmodule DateTime do
...
+        {:ambiguous, first_datetime, second_datetime}
+
+      {:gap, {first_period, first_period_until_wall}, {second_period, second_period_from_wall}} ->
+        # `until_wall` is not valid, but any time just before is.
+        # So by subtracting a second and adding .999999 seconds
+        # we get the last microsecond just before.
+        before_naive =
+          first_period_until_wall
+          |> Map.put(:microsecond, {999_999, 6})
+          |> NaiveDateTime.add(-1)
```
josevalim: @lau why can't we use first_period_until_wall as is and then document that the `:gap` function returns the first invalid value and the first valid value from then?

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -172,20 +172,213 @@ defmodule DateTime do
...
+    end
+  end
+
+  def from_naive(%{calendar: calendar} = naive_datetime, time_zone, tz_db_or_config)
+      when calendar != Calendar.ISO do
+    # For non-ISO calendars, convert to ISO, create ISO DateTime, and then
+    # convert to original calendar
+    iso_result =
+      with {:ok, in_iso} <- NaiveDateTime.convert(naive_datetime, Calendar.ISO) do
+        from_naive(in_iso, time_zone, tz_db_or_config)
```
josevalim: I think it would be more efficient if in this case we use the TimeZoneClient function that works with iso days, as it would require less conversions. I am playing with it and I will let you know.

josevalim: Actually, the current approach is correct and my suggestion is wrong. We cannot assume the given data is correct, so we do need to convert and validate.

lib/elixir/lib/calendar/datetime.ex
josevalim: We can't do what I propose because then it would return an invalid datetime. I wonder then if we should ask the timezone database to return the last valid instant so we can skip this conversion. Thoughts @lau?

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -172,20 +172,213 @@ defmodule DateTime do
...
+
+      {:error, _} = error ->
+        error
+    end
+  end
+
+  def from_naive(%{calendar: calendar} = naive_datetime, time_zone, tz_db_or_config)
+      when calendar != Calendar.ISO do
+    # For non-ISO calendars, convert to ISO, create ISO DateTime, and then
+    # convert to original calendar
```
josevalim: I want to mention this is not the most efficient nor the most correct way to go about this because it doesn't give the opportunity for databases which cannot be converted to ISO to handle timezones. The correct way would probably to add a new callback to `Calendar`. However, I am not even sure if there is a calendar that fits this criteria, so I would rather wait until an use case proves this necessary. So I am perfectly happy to ship this as is! :+1: :heart:

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -240,6 +439,170 @@ defmodule DateTime do
...
+          period.zone_abbr
+        )
+
+      {:error, _} = error ->
+        error
+    end
+  end
+
+  def shift_zone(%{calendar: calendar} = datetime, time_zone, tz_db_or_config)
+      when calendar != Calendar.ISO do
```
josevalim: Since shift_zone relies on iso days conversion, we can completely skip those conversions and generalize the clause above. I have a patch ready.

josevalim:
Closing in favor of #8383.

lib/elixir/lib/calendar/datetime.ex
josevalim: Moved this to the new issue.

PR #7917, opened 2018-07-17
Use "time zone" instead of "timezone" everywhere

fertapric:
There is a mix of "timezone" and "time zone" in the codebase. For what I read, "time zone" is usually more correct.

In addition, there are some variables and types named `time_zone`, so I think the best option is to go with "time zone".

### DIFF
--- a/lib/elixir/lib/calendar/datetime.ex
+++ b/lib/elixir/lib/calendar/datetime.ex
@@ -17,7 +17,7 @@ defmodule DateTime do
   Such functions expect `t:Calendar.datetime/0` in their typespecs
   (instead of `t:t/0`).
 
-  Developers should avoid creating the DateTime struct directly
+  Developers should avoid creating the `DateTime` struct directly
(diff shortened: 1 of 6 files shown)


lib/elixir/lib/calendar/datetime.ex
```diff
@@ -26,7 +26,7 @@ defmodule DateTime do
   You will notice this module only contains conversion
   functions as well as functions that work on UTC. This
   is because a proper DateTime implementation requires a
-  TimeZone database which currently is not provided as part
+  Time Zone database which currently is not provided as part
```
michalmuskala: I think this should probably just say "time zone". I'm not sure why it was capitalised like that.

fertapric: Agree. Changing it.

fertapric:
I've also added some backticks

michalmuskala:
Thank you! :heart:

PR #7920, opened 2018-07-17
Fixed indentation in StringIO.open/3 examples and added missing @doc :since metadata

amatalai:
At the moment examples are interpreted as normal text because indentation is too small
https://hexdocs.pm/elixir/1.7.0-rc.0/StringIO.html#open/3

Also, `@doc :since` metadata was missing and I cannot find anything about `StringIO` in v1.7 CHANGELOG. Function was added in #7477

### DIFF
--- a/lib/elixir/lib/string_io.ex
+++ b/lib/elixir/lib/string_io.ex
@@ -15,7 +15,6 @@ defmodule StringIO do
 
   use GenServer
 
-  @doc since: "1.7.0"
   @doc ~S"""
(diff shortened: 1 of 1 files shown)


josevalim:

josevalim:
thanks @amatalai ! I removed the since because only a certain clause expects the anonymous function but part of the other clause can handle anonymous functions too, so i thought keeping the since around could be slightly confusing.

amatalai:
Sure

PR #7921, opened 2018-07-17
Check duplicated @on_load attributes

fertapric:
This is an attempt to implement the check from #5800: `multiple_on_loads`: multiple `on_load` attributes.

I'm not sure if this is the right spot to perform the check :)

### DIFF
--- a/lib/elixir/lib/module.ex
+++ b/lib/elixir/lib/module.ex
@@ -1797,7 +1797,7 @@ defmodule Module do
         :ets.insert(set, {:on_load, value, line})
         :ets.insert(bag, {:attributes, :on_load})
     else
-      _ -> raise ArgumentError, "the @on_load attribute can only be called once per module"
+      _ -> raise ArgumentError, "the @on_load attribute can only be set once per module"
     end
   end
 
--- a/lib/elixir/lib/module.ex
+++ b/lib/elixir/lib/module.ex
@@ -466,7 +466,7 @@ defmodule Module do
       of the corresponding setting in `Code.compiler_options/1`
 
     * `@compile {:inline, some_fun: 2, other_fun: 3}` - inlines the given
-      name/arity pairs. Inlining is applied locally, calls from another 
+      name/arity pairs. Inlining is applied locally, calls from another
       module are not affected by this option
 
     * `@compile {:autoload, false}` - disables automatic loading of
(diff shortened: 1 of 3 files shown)


lib/elixir/lib/module.ex
```diff
@@ -1789,6 +1789,22 @@ defmodule Module do
...
+  defp put_attribute(module, :on_load, value, line, _set, bag) do
+    try do
+      :ets.lookup_element(bag, {:accumulate, :on_load}, 2)
+    catch
+      :error, :badarg -> :ets.insert(bag, {{:accumulate, :on_load}, value})
+    else
+      _ ->
+        [{_, _, _, file: file, line: line}] = attribute_stack(module, line)
+
+        raise CompileError,
```
josevalim: Nowhere in this module we raise a compile error, so I am thinking a `raise ArgumentError` is enough as it will contain relevant stacktrace info.

fertapric:
I wasn't sure if this part should be changed https://github.com/elixir-lang/elixir/blob/master/lib/elixir/src/elixir_module.erl#L241. At the end, the `on_load` attribute cannot be "accumulated".

josevalim:
@fertapric yes, I am thinking the same. However, we need to guarantee that `@on_load` will warn if set and not used. This code warns:

```elixir
defmodule Foo do
  @example :foo
  @example :bar
end
```

but for some reasons this code does not:

```elixir
defmodule Foo do
  @on_load :foo
  @on_load :bar
  def foo, do: :ok
  def bar, do: :ok
end
```

fertapric:
Agree. Besides, it will be a different behaviour than the one we are replacing.

fertapric:
I've updated the PR with the changes. Builds are failing for something that seems not to be related to these changes.

lib/elixir/lib/module.ex
```diff
@@ -1789,6 +1789,18 @@ defmodule Module do
...
 
+  defp put_attribute(_module, :on_load, value, line, set, bag) do
+    try do
+      :ets.lookup_element(set, :on_load, 3)
+    catch
+      :error, :badarg ->
+        :ets.insert(set, {:on_load, value, line})
+        :ets.insert(bag, {:attributes, :on_load})
+    else
+      _ -> raise ArgumentError, "the @on_load attribute can only be called once per module"
```
whatyouhide: I think we should say "can only be **set** once per module", wdyt?

fertapric: 👌

josevalim:

PR #7923, opened 2018-07-17
Do not use user defined templateDir and hooksDir

hauleth:
This can cause problems with some user configurations as such flags can
introduce unwanted output/behaviour in some test cases.

### DIFF
--- a/lib/mix/lib/mix/scm/git.ex
+++ b/lib/mix/lib/mix/scm/git.ex
@@ -94,13 +94,13 @@ defmodule Mix.SCM.Git do
     File.mkdir_p!(path)
 
     File.cd!(path, fn ->
-      git!(~w[-c core.hooksPath='' -c core.templateDir='' init --quiet])
+      git!(~w[-c core.hooksPath='' -c init.templateDir='' init --quiet])
 
       git!([
         "-c",
         "core.hooksPath=''",
         "-c",
-        "core.templateDir=''",
+        "init.templateDir=''",
         "--git-dir=.git",
         "remote",
         "add",
--- a/lib/mix/lib/mix/scm/git.ex
+++ b/lib/mix/lib/mix/scm/git.ex
@@ -95,19 +95,7 @@ defmodule Mix.SCM.Git do
 
     File.cd!(path, fn ->
       git!(~w[-c core.hooksPath='' -c init.templateDir='' init --quiet])
-
-      git!([
-        "-c",
-        "core.hooksPath=''",
-        "-c",
-        "init.templateDir=''",
-        "--git-dir=.git",
-        "remote",
-        "add",
-        "origin",
-        opts[:git]
-      ])
-
+      git!(["remote", "add", "origin", opts[:git]])
       checkout(path, opts)
     end)
   end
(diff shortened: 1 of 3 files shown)


josevalim:
Maybe we should set [`GIT_CONFIG_NOSYSTEM`](https://git-scm.com/docs/git#git-codeGITCONFIGNOSYSTEM) instead?

hauleth:
@josevalim this will only ignore system-wide git configuration, not user-wide one.

lib/mix/lib/mix/scm/git.ex
```diff
@@ -94,8 +94,20 @@ defmodule Mix.SCM.Git do
     File.mkdir_p!(path)
 
     File.cd!(path, fn ->
-      git!(["init", "--quiet"])
-      git!(["--git-dir=.git", "remote", "add", "origin", opts[:git]])
+      git!(~w[-c core.hooksPath='' -c init.templateDir='' init --quiet])
```
josevalim: I don't think we need to pass `core.hooksPath` here.

josevalim:
> @josevalim this will only ignore system-wide git configuration, not user-wide one.

As the documentation for the variable suggested, we would need to set that and the HOME variable to avoid all configs. But I actually don't think that is a good idea since SSH configuration goes in the config files and we need that. :)

lib/mix/lib/mix/scm/git.ex
```diff
@@ -94,8 +94,20 @@ defmodule Mix.SCM.Git do
...
+
+      git!([
+        "-c",
+        "core.hooksPath=''",
+        "-c",
+        "init.templateDir=''",
```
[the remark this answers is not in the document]
hauleth: Unfortunately, no. This is also used by `git clone` and I believe that `submodules` use this as well.

lib/mix/lib/mix/scm/git.ex
josevalim: If it isn't needed, then we shouldn't keep it. I don't think consistency here matters. Different commands, different options.

lib/mix/lib/mix/scm/git.ex
```diff
@@ -94,8 +94,20 @@ defmodule Mix.SCM.Git do
     File.mkdir_p!(path)
 
     File.cd!(path, fn ->
-      git!(["init", "--quiet"])
-      git!(["--git-dir=.git", "remote", "add", "origin", opts[:git]])
+      git!(~w[-c core.hooksPath='' -c init.templateDir='' init --quiet])
+
+      git!([
+        "-c",
+        "core.hooksPath=''",
```
josevalim: Which hooks are invoked on `remote add origin`?

lib/mix/lib/mix/scm/git.ex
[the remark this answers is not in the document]
josevalim: Hrm, I see, thanks. But it certainly isn't used by git-remote, is it?

lib/mix/lib/mix/scm/git.ex
```diff
@@ -94,8 +94,8 @@ defmodule Mix.SCM.Git do
     File.mkdir_p!(path)
 
     File.cd!(path, fn ->
-      git!(["init", "--quiet"])
-      git!(["--git-dir=.git", "remote", "add", "origin", opts[:git]])
+      git!(~w[-c core.hooksPath='' -c init.templateDir='' init --quiet])
+      git!(["remote", "add", "origin", opts[:git]])
```
josevalim: You removed the --git-dir configuration from this previous command.

lib/mix/lib/mix/scm/git.ex
josevalim: @hauleth and did you look into setting `--templates=""` instead here and below?

josevalim:
Ping!

hauleth:
@josevalim I will finish this later as I had other tasks today (in and out of work). Sorry

josevalim:
Ping :D

josevalim:

PR #7925, opened 2018-07-18
Fix nested use of Mix.Config.eval!/2

Gazler:
Previously, if a config file uses Mix.Config.eval!/2 then it would fail with the
following error:

 > ** (RuntimeError) could not set configuration via Mix.Config. This usually
 >  means you are trying to execute a configuration file directly instead of
 > using the proper command, such as mix loadconfig
 > code: assert Mix.Config.eval!(fixture_path("configs/nested_import.exs")) ==
 > stacktrace:
 >   (mix) lib/mix/config.ex:71: Mix.Config.raise_improper_use!/0
 >   (mix) lib/mix/config.ex:222: Mix.Config.eval!/2
 >   lib/mix/test/mix/config_test.exs:110: (test)

### DIFF
--- a/lib/mix/test/fixtures/configs/nested_import.exs
+++ b/lib/mix/test/fixtures/configs/nested_import.exs
@@ -1,0 +1,3 @@
+use Mix.Config
+Mix.Config.eval!(Path.join([__DIR__, "good_config.exs"]))
+:done
--- a/lib/mix/lib/mix/config.ex
+++ b/lib/mix/lib/mix/config.ex
@@ -47,7 +47,6 @@ defmodule Mix.Config do
 
   defp put_config(value) do
     Process.put(@config_key, value)
-    :ok
   end
 
   defp delete_config() do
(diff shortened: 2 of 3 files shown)


lib/mix/test/fixtures/configs/nested_import.exs
```diff
@@ -0,0 +1,3 @@
+use Mix.Config
+Mix.Config.eval!(Path.join([__DIR__, "good_config.exs"]))
+:done
```
whatyouhide: Do we use `:done` for anything here?

Gazler: I copied it from https://github.com/Gazler/elixir/blob/a7787e6077e2d5df7625509fd3c776b65d6bc9ec/lib/mix/test/fixtures/configs/good_import.exs

whatyouhide: Yeah I was wondering if we use that one for anything also 😄 Let's merge this and we'll check later. 👍

PR #7927, opened 2018-07-18
Optimize metadata merging and ensure no nil leakage

### LINKED ISSUE #7926: Logger metadata not removed
### Environment

* Elixir & Erlang/OTP versions: Erlang/OTP 20.1, Elixir 1.6.6

### Current behavior

When setting a logger metadata to `:nil` using the `Logger.metadata/1` function the given metadata is removed from the keyword list and not printed. If instead this is done via a single log call (for instance `Logger.debug("bar", foo: nil)` the metadata is not removed and for example the console backend prints it as `foo=`

```
iex> require Logger
Logger
iex> Logger.configure_backend(:console, metadata: [:foo])
:ok
iex> Logger.metadata(foo: 123)
:ok
iex> Logger.debug("bar")
foo=123 [debug] bar
iex> Logger.debug("bar", foo: 456)
foo=456 [debug] bar
iex> Logger.debug("bar", foo: nil)
foo= [debug] bar
iex> Logger.metadata(foo: nil)
:ok
iex> Logger.debug("bar")
[debug] bar
iex> Logger.debug("bar", foo: nil)
foo= [debug] bar
```

### Expected behavior

I would expect that the behavior is the same, specifically that setting a metadata to `nil` removes it always (in the case of single call only for that call).

I think it is sufficient to implement the logic of the function `Logger.metadata/1` in the function `Logger.bare_log/3` but without saving the changed metadata in the process dictionary.

I can try to push a PR if you think is worth.

On the other hand if this is intended I think we should specify in the documentation that setting a metadata to nil in single log calls does not remove it.

josevalim:
Closes #7926

### DIFF
--- a/lib/logger/lib/logger.ex
+++ b/lib/logger/lib/logger.ex
@@ -416,7 +416,7 @@ defmodule Logger do
   @type backend :: :gen_event.handler()
   @type message :: IO.chardata() | String.Chars.t()
   @type level :: :error | :info | :warn | :debug
-  @type metadata :: keyword(String.Chars.t())
+  @type metadata :: keyword()
   @levels [:error, :info, :warn, :debug]
 
   @metadata :logger_metadata
@@ -436,15 +436,23 @@ defmodule Logger do
   @spec metadata(metadata) :: :ok
   def metadata(keyword) do
     {enabled?, metadata} = __metadata__()
-
-    metadata =
-      Enum.reduce(keyword, metadata, fn
-        {key, nil}, acc -> Keyword.delete(acc, key)
-        {key, val}, acc -> Keyword.put(acc, key, val)
-      end)
-
-    Process.put(@metadata, {enabled?, metadata})
+    Process.put(@metadata, {enabled?, into_metadata(keyword, metadata)})
     :ok
+  end
+
+  defp into_metadata([], metadata), do: metadata
+  defp into_metadata(keyword, metadata), do: into_metadata(keyword, [], metadata)
+
+  defp into_metadata([{key, nil} | keyword], prepend, metadata) do
+    into_metadata(keyword, prepend, :lists.keydelete(key, 1, metadata))
+  end
+
+  defp into_metadata([{key, _} = pair | keyword], prepend, metadata) do
+    into_metadata(keyword, [pair | prepend], :lists.keydelete(key, 1, metadata))
+  end
+
+  defp into_metadata([], prepend, metadata) do
+    prepend ++ metadata
   end
 
   @doc """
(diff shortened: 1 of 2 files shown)


michalmuskala:
I wonder if at that point it wouldn't be more efficient to maintain the matadata as a map internally and convert to keyword on access.

lib/logger/lib/logger.ex
```diff
@@ -671,7 +679,7 @@ defmodule Logger do
   @doc false
   def __do_log__({level, config, pdict}, chardata_or_fun, metadata) when is_list(metadata) do
     %{utc_log: utc_log?, truncate: truncate, mode: mode} = config
-    metadata = [pid: self()] ++ Keyword.merge(pdict, metadata)
+    metadata = [pid: self()] ++ into_metadata(metadata, pdict)
```
fertapric: Micro-micro optimization: `[{:pid, self()} | into_metadata(metadata, pdict)]`?

josevalim:
@michalmuskala Yeah, I tried it but it was a bit messy. I think the internal metadata should always be a map, as it would also speed up the backends, so I created #7928 so we can track those changes.

lib/logger/lib/logger.ex
josevalim: They are equivalent when the left side of `++` is a literal.

eksperimental: Interesting. Is this documented? If not we should

michalmuskala: It is a compiler optimisation. There are many of them and they also change between versions - I don't think it would be useful to document them all.

In general, one should write the most straightforward code and expect the compiler to translate it to the most efficient code.

josevalim:

lib/logger/lib/logger.ex
fertapric: Sorry, my Erlarng-literals-fu is weak 😅 

I was wondering how `[pid: self()]` can be treated as a literal if `self()` is obtained at runtime. Is there any compiler optimization magic ✨ I can look into?

(Sorry for using this for learning purposes instead of the forum)

josevalim:
As long as the end of the list on the left side is known, we can
concatenate at compile time. Like we can transform [1, 2, 3] ++ [4, 5] into
[1, 2, 3, 4, 5].
-- 


lib/logger/lib/logger.ex
michalmuskala: It's a "partial" literal - since some structure is known (the length of the list in particular), the compiler can emit some better code than first building the entire list and then calling the append operation.

PR #7933, opened 2018-07-19
Enable URI.encode_query/1 to encode list values

spieker:
This enables the URI.encode_query/1 function, to encode lists into url parameters. Since it's currently throwing argument errors, it's not possible to create URLs like this:

```
filter[id_in][]=1&filter[id_in][]=2&filter[id_in][]=3
```

This PR adds the following two possible cases:

```
iex> URI.encode_query(%{key: [:a, :list]})
"key%5B%5D=a&key%5B%5D=list"

iex> URI.encode_query(%{key: [{:foo, :a}, {:bar, :list}]})
"key%5Bfoo%5D=a&key%5Bbar%5D=list"
```

### DIFF
--- a/lib/elixir/lib/uri.ex
+++ b/lib/elixir/lib/uri.ex
@@ -76,7 +76,7 @@ defmodule URI do
   values are URL encoded as per `encode_www_form/1`.
 
   Keys and values can be any term that implements the `String.Chars`
-  protocol, except lists which are explicitly forbidden.
+  protocol. However, it's explicitly forbidden for keys to be lists.
 
   ## Examples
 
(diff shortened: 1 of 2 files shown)


whatyouhide:
As [the documentation](https://github.com/elixir-lang/elixir/blob/6d8ebd4327f2312354f812abf15a4e666fa1937b/lib/elixir/lib/uri.ex#L79) states, lists are **explicitly** forbidden. I can't quite remember why but I remember discussing this with the others at some point and agreeing again that we shouldn't encode lists.

spieker:
@josevalim  Thanks a lot for your explanation. I've kind of expected the PR to be rejected for exactly this reason. Also thanks, for pointing me in the direction of Plug, I'll have a look there.

fertapric:
The documentation states that list are explicitly forbidden, but it does not explain why. Should the documentation reflect that?

josevalim:
@fertapric the documentation is likely referring to charlists.

I don't think it is the job of the documentation though to say why we don't implement such features.

PR #7934, opened 2018-07-19
Fix a typo in the unnecessary quotes warning

ejpcmac:
I’m not a native speaker, but it seems to me the warning on unnecessary quotes sounds strange. Here is a quick fix for that :)


michalmuskala:
Thank you! :heart:

PR #7942, opened 2018-07-20
Fixed misplaced parenthesis in System.cmd/3 examples

amatalai:


josevalim:
Unfortunately echo doesn't work the same on all OSes. :(

PR #7949, opened 2018-07-20
Allow to prune metadata while escaping

### LINKED ISSUE #7947: Macros may leak hygiene info into compiled modules
### Environment

```
Erlang/OTP 21 [erts-10.0] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [hipe]

Elixir 1.6.6 (compiled with OTP 19)
```

* Operating system: macOS 10.12.6

### Current behavior

To give you some background on how I discovered this, I'm building an extension library for ExUnit which provides helpers for running tests against a clustered application. I got most of the way done and was writing tests, when I encountered something strange: a subset of my tests worked fine, but when using my custom `Case` implementation, equivalent tests would fail with a `:badfun` error.

- [Example of working test module](https://github.com/bitwalker/ex_unit_clustered_case/blob/master/test/cluster_test.exs)
- [Example of roughly equivalent test using ClusteredCase which fails](https://github.com/bitwalker/ex_unit_clustered_case/blob/master/test/clustered_case_test.exs)

The way the `scenario/3` macro works in the latter of those two examples, is that it will generate more or less something like the following:

```elixir
@attribute_with_scenario_config scenario_config
describe "scenario description" do
  setup context do
    # initializes a cluster or gets the pid of the one already running for this scenario
    cluster = ...
    Map.put(context, :cluster, cluster)
  end

  unquote(block)
end
```

Initially I thought it was something in how I wrote this, but I did a sanity check by simply wrapping my working test module (the first of the two links above) in a `describe`, and running them again - this time the tests failed with `:badfun`!

The cause of the `:badfun` is that the module md5 is changing on every compilation. This is a problem for my use case in particular because `:erl_boot_server` and `:erl_prim_loader` require a file on disk in order to load code remotely, and there is no API for fetching the BEAM binary for modules compiled in memory so that I can use `:code.load_binary/3` via RPC. Since test modules are compiled in memory, I don't have a way to send those modules to nodes in a cluster. Instead, my workaround for this, is that I load test module code onto the nodes in a cluster using more or less the same method that the test compiler uses to compile them for a normal test run, i.e., finding the set of files which match the test file pattern, and compiling them via `Kernel.ParallelCompiler.require/2`.

What I found, by inspecting the output of `:beam_lib.chunks(module_bin, [:compile_info])`, is that the md5 of the module is different on every compile. By then looking at the `:abstract_code` chunk, I found that by adding `describe` to my `cluster_test.exs` module (first link above), a counter was added in three places:

```erlang
⟩ diff -u abstract_code{1,2}.erl
--- abstract_code1.erl  2018-07-20 14:25:39.000000000 -0400
+++ abstract_code2.erl  2018-07-20 14:25:39.000000000 -0400
@@ -204,7 +204,7 @@
                           {cons,0,
                            {tuple,0,
                             [{atom,0,counter},
-                             {integer,0,-576460752303372252}]},
+                             {integer,0,-576460752303398107}]},
                            {nil,0}}},
                          {cons,0,
                           {atom,0,'ExUnit'},
@@ -355,7 +355,7 @@
                           {cons,0,
                            {tuple,0,
                             [{atom,0,counter},
-                             {integer,0,-576460752303372252}]},
+                             {integer,0,-576460752303398107}]},
                            {nil,0}}},
                          {cons,0,
                           {atom,0,'ExUnit'},
@@ -400,7 +400,7 @@
                                      {cons,0,
                                       {tuple,0,
                                        [{atom,0,counter},
-                                        {integer,0,-576460752303372252}]},
+                                        {integer,0,-576460752303398107}]},
                                       {nil,0}}},
                                     {cons,0,{atom,0,'Node'},{nil,0}}]},
                                   {cons,0,{atom,0,self},{nil,0}}}]},

```

Since the value of this counter is different on every compilation, the md5 changes as well.

I brought this up with José, and his opinion was that likely some macro env info (namely the counter for hygiene) is being leaked into the compiled module. From what I can recall when looking at the abstract code, there were other bits of macro env in there as well (like line numbers and aliases), but the counter is the one which causes the issue.

### Expected behavior

Compiling a module with the same contents, in the same compilation environment (e.g. there isn't code in the module deliberately generating dynamic content each time) more than once should produce the same BEAM bytecode, most importantly, the module md5 should not change.

josevalim:
We use this option in ExUnit as we are escaping
the code to eventually convert it to a string
representation and therefore the metadata is not
relevant.

Closes #7947.

### DIFF
--- a/lib/elixir/lib/macro.ex
+++ b/lib/elixir/lib/macro.ex
@@ -384,10 +384,10 @@ defmodule Macro do
       As an example, `ExUnit` stores the AST of every assertion, so when
       an assertion fails we can show code snippets to users. Without this
       option, each time the test module is compiled, we get a different
-      MD5 of the module byte code, because the AST contains metadata
-      specific, such as counters, specific the compilation environment.
-      By pruning the metadata, we ensure that the module is deterministic
-      and reduce the amount of data ExUnit needs to keep around.
+      MD5 of the module byte code, because the AST contains metadata,
+      such as counters, specific the compilation environment. By pruning
+      the metadata, we ensure that the module is deterministic and reduce
+      the amount of data `ExUnit` needs to keep around.
 
   ## Comparison to `Kernel.quote/2`
 
--- a/lib/elixir/lib/macro.ex
+++ b/lib/elixir/lib/macro.ex
@@ -385,7 +385,7 @@ defmodule Macro do
       an assertion fails we can show code snippets to users. Without this
       option, each time the test module is compiled, we get a different
       MD5 of the module byte code, because the AST contains metadata,
-      such as counters, specific the compilation environment. By pruning
+      such as counters, specific to the compilation environment. By pruning
       the metadata, we ensure that the module is deterministic and reduce
       the amount of data `ExUnit` needs to keep around.
 
--- a/lib/elixir/lib/macro.ex
+++ b/lib/elixir/lib/macro.ex
@@ -380,6 +380,14 @@ defmodule Macro do
       nodes. Note this option changes the semantics of escaped code and
       it should only be used when escaping ASTs, never values. Defaults
       to false.
+      
+      As an example, `ExUnit` stores the AST of every assertion, so when
+      an assertion fails we can show code snippets to users. Without this
+      option, each time the test module is compiled, we get a different
+      MD5 of the module byte code, because the AST contains metadata
+      specific, such as counters, specific the compilation environment.
+      By pruning the metadata, we ensure that the module is deterministic
+      and reduce the amount of data ExUnit needs to keep around.
 
   ## Comparison to `Kernel.quote/2`
 
--- a/lib/elixir/lib/macro.ex
+++ b/lib/elixir/lib/macro.ex
@@ -356,12 +356,7 @@ defmodule Macro do
   def decompose_call(_), do: :error
 
   @doc """
-  Recursively escapes a value so it can be inserted
-  into a syntax tree.
-
-  One may pass `unquote: true` to `escape/2`
-  which leaves `unquote/1` statements unescaped, effectively
-  unquoting the contents on escape.
+  Recursively escapes a value so it can be inserted into a syntax tree.
 
   ## Examples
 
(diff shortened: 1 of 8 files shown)


bitwalker:
Just wanted to chime in here and let you know that this PR solves the problem I was seeing - the tests that were previously failing due to `:badfun` now succeed!

michalmuskala:
This looks good, but I'm still not sure when I would use something like this writing code in Elixir - could you give some examples and explanation why it's necessary, @josevalim?

josevalim:
@michalmuskala done! please let me know if it is clearar now!

lib/elixir/lib/macro.ex
```diff
@@ -374,6 +369,26 @@ defmodule Macro do
...
+
+    * `:prune_metadata` - when true, removes metadata from escaped AST
+      nodes. Note this option changes the semantics of escaped code and
+      it should only be used when escaping ASTs, never values. Defaults
+      to false.
+      
+      As an example, `ExUnit` stores the AST of every assertion, so when
+      an assertion fails we can show code snippets to users. Without this
+      option, each time the test module is compiled, we get a different
+      MD5 of the module byte code, because the AST contains metadata
```
fertapric: I think there is something missing in this part:
> because the AST contains metadata specific, such as counters, specific the compilation environment.

josevalim: One word too many! Tks, fixed!

michalmuskala:
Looks good to me :)

lib/elixir/lib/macro.ex
fertapric: 😬

But, might a **”to”** be missed too? Right now it reads “because the AST contains metadata specific the compilation environment”, should it be “because the AST contains metadata specific **to** the compilation environment”?

josevalim: Good catch again! :D

josevalim:

arjan:
@josevalim any chance this gets backported into 1.6?

josevalim:
@arjan the feature or the fix?

arjan:
Good question, b4c3ea9549fee3529ca778f7ddb76b2ea7b3c060 I guess, it's for [this library](https://github.com/bitwalker/ex_unit_clustered_case#exunit-clustered-cases). cc @bitwalker

bitwalker:
@josevalim I'd be curious to know which release this would fall in, and if backporting to 1.6 is an option (assuming it's in 1.7), that'd open up the ability to use libraries like the one linked with current applications which may not upgrade for some time, though I'm not too concerned about that if it'll be in 1.7. If it's not going to be in 1.7 either, that'd be a problem for my library, but if there are problems with dropping it that soon, I just need to provide people with instructions on how to use master in the meantime.

josevalim:
We have backported to v1.7. We won’t backport it to v1.6 though as from
today v1.7 is the main branch for bug fixes.
-- 


bitwalker:
Thanks José! Sounds good to me :)

PR #7956, opened 2018-07-21
Fixed too wide example for Calendar.ISO.datetime_to_string

amatalai:
How it looks like now:
z1

How longest line should looks like after change:
z2


josevalim:

PR #7966, opened 2018-07-22
Make OptionParser docs more formatter-compatible

amatalai:
This also fixes one `OptionParser.parse/2` example that was too wide for hexdocs.


josevalim:

PR #7968, opened 2018-07-22
Improve Module.defines?/3 examples

fertapric:
The current examples always returning false, and I thought that showing when it would return `true` would be helpful (and it matches `Module.defines?/2`) :)

I've also formatted a few examples in this module.


josevalim:

PR #7973, opened 2018-07-22
Validate @on_load function in Elixir

fertapric:
This is an attempt to implement the check from #5800: `{undefined_on_load,{F,A}}`: function for `on_load` undefined.

These are my first steps with Erlang 🎉 , any feedback is more than appreciated :) 

During the development of this PR, I had some doubts regarding the visibility of the `@on_load` function. According to the [Erlang's documentation](http://erlang.org/doc/reference_manual/code_loading.html#on_load), "It is not necessary to export the function". (I'm currently learning the Elixir internals and what I'm about to say might not make sense, so please be bear with me 😅) I would think that public functions map to exported functions, and private functions to non-exported functions. If that's the case, this should work (as the function does not need to be public/exported):

```elixir
defmodule Example
  @on_load :foo

  defp foo do
    IO.puts("HELLO")
    :ok
  end
end
```

However, when executing the code, this error is returned:

```
warning: function foo/0 is unused
  lib/test.ex:4

** (CompileError) lib/test.ex:1: function foo/0 undefined
```

I wasn't sure if this was an expected behavior. I've assumed the current implementation was right, and this PR will make the compiler raise a `CompileError` is the function used in `@on_load` is private. I've also updated the documentation to reflect the visibility of the `on_load` function.

### DIFF
--- a/lib/elixir/src/elixir_module.erl
+++ b/lib/elixir/src/elixir_module.erl
@@ -169,7 +169,6 @@ validate_on_load_attribute({on_load, Def}, Defs, File, Line) ->
   end;
 validate_on_load_attribute(false, _Defs, _File, _Line) -> ok.
 
-
 %% An undef error for a function in the module being compiled might result in an
 %% exception message suggesting the current module is not loaded. This is
 %% misleading so use a custom reason.
--- a/lib/elixir/src/elixir_module.erl
+++ b/lib/elixir/src/elixir_module.erl
@@ -92,6 +92,10 @@ compile(Line, Module, Block, Vars, E) ->
     RawCompileOpts = bag_lookup_element(DataBag, {accumulate, compile}, 2),
     CompileOpts = validate_compile_opts(RawCompileOpts, AllDefinitions, Unreachable, File, Line),
 
+    OnLoadAttribute = lists:keyfind(on_load, 1, Attributes),
+    OnLoadAttribute /= false andalso
+      validate_on_load_attribute(OnLoadAttribute, AllDefinitions, Unreachable, File, Line),
+
     ModuleMap = #{
       module => Module,
       line => Line,
@@ -154,6 +158,16 @@ validate_inlines([Inline | Inlines], Defs, Unreachable, Acc) ->
       end
   end;
 validate_inlines([], _Defs, _Unreachable, Acc) -> {ok, Acc}.
+
+validate_on_load_attribute({on_load, Def}, Defs, Unreachable, File, Line) ->
+  case lists:keyfind(Def, 1, Defs) of
+    false ->
+      elixir_errors:form_error([{line, Line}], File, ?MODULE, {undefined_on_load, Def});
+    _ ->
+      lists:member(Def, Unreachable) andalso
+        elixir_errors:form_error([{line, Line}], File, ?MODULE, {private_on_load, Def})
+  end.
+
 
 %% An undef error for a function in the module being compiled might result in an
 %% exception message suggesting the current module is not loaded. This is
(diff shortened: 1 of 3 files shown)


josevalim:
>  I would think that public functions map to exported functions, and private functions to non-exported functions. If that's the case, this should work (as the function does not need to be public/exported).

Correct!. The reason why it doesn't work is because when you  define `:on_load` as a private function, we think it is never invoked, and then we completely removed it from the module, making compilation fail. We could fix this by letting the compiler know that the `@on_load` function has been used. But given that this never worked anyway, I would rather keep things as is for now.

lib/elixir/src/elixir_module.erl
```diff
@@ -92,6 +92,10 @@ compile(Line, Module, Block, Vars, E) ->
     RawCompileOpts = bag_lookup_element(DataBag, {accumulate, compile}, 2),
     CompileOpts = validate_compile_opts(RawCompileOpts, AllDefinitions, Unreachable, File, Line),
 
+    OnLoadAttribute = lists:keyfind(on_load, 1, Attributes),
+    OnLoadAttribute /= false andalso
+      validate_on_load_attribute(OnLoadAttribute, AllDefinitions, Unreachable, File, Line),
```
josevalim: I would remove `OnLoadAttribute /= false` and check for false inside `validate_on_load_attribute`.

lib/elixir/src/elixir_module.erl
```diff
@@ -155,6 +159,16 @@ validate_inlines([Inline | Inlines], Defs, Unreachable, Acc) ->
   end;
 validate_inlines([], _Defs, _Unreachable, Acc) -> {ok, Acc}.
 
+validate_on_load_attribute({on_load, Def}, Defs, Unreachable, File, Line) ->
+  case lists:keyfind(Def, 1, Defs) of
+    false ->
+      elixir_errors:form_error([{line, Line}], File, ?MODULE, {undefined_on_load, Def});
+    _ ->
+      lists:member(Def, Unreachable) andalso
```
josevalim: Because you are looking up on unreachable, it means this code will work:

```elixir
defmodule Foo do
  @on_load :baz
  def bar, do: baz
  defp baz, do: :ok
end
```

So I think we should rather check the result returned by `lists:keyfind/3`. One of the elements in the tuple should say if it is `def`, `defmacro`, `defp` or `defmacrop`.

josevalim:
Great job, I have added some comments!

fertapric:
I've updated the PR with a few changes:

* As suggested [here](https://github.com/elixir-lang/elixir/pull/7973#discussion_r204254404), the check for `false` is now made in the `validate_on_load_attribute` function.
* As pointed out [here](https://github.com/elixir-lang/elixir/pull/7973#discussion_r204254474), it now uses the definition to check that the `@on_load` function was implemented as a public function.
* I've changed the error when the `@on_load` attribute is defined with the wrong kind. That should cover the cases of `defp`, `defmacro` and `defmacrop`.
* I've updated the documentation to reflect the difference between the requirements of the `@on_load` function and its behavior.

lib/elixir/test/elixir/kernel/errors_test.exs
```diff
@@ -719,6 +719,30 @@ defmodule Kernel.ErrorsTest do
...
 
+  test "@on_load attribute with undefined function" do
+    assert_eval_raise CompileError,
+                      "nofile:1: @on_load function foo/0 is undefined",
+                      'defmodule UndefinedOnLoadAttribute do @on_load :foo end'
+  end
+
+  test "wrong kind for @on_load attribute" do
+    # Capture warning: function foo/0 is unused
+    ExUnit.CaptureIO.capture_io(:stderr, fn ->
```
josevalim: If we do this, the testcase cannot be async. So let's define a public function that calls `defp foo` so we don't need to capture the warning.

fertapric: Good catch! I completely forgot about that!

fertapric: Done in https://github.com/elixir-lang/elixir/pull/7973/commits/86c46d744f560c8f3f785c6c0788174778d5474b

josevalim:

PR #7977, opened 2018-07-23
Update Process.info spec to reflect the possible nil return value

boydm:
Process.info returns nil if the process is no longer alive, however the @spec for it did not reflect that possible return value .

This causes dialyzer to throw errors if you check for the nil return value.

This pull request updates the @spec for Process.info to allow for the nil return value.

Tests passed. No doc changes necessary as it already said it could return nil.


fishcakez:
Thank you :heart:

PR #7978, opened 2018-07-23
Make Inspect.Algebra docs examples more formatter-compatible

amatalai:


lib/elixir/lib/inspect/algebra.ex
```diff
@@ -324,18 +324,37 @@ defmodule Inspect.Algebra do
...
-      iex> doc = Inspect.Algebra.container_doc("[", Enum.to_list(1..5), "]",
-      ...>         %Inspect.Opts{limit: :infinity}, fn i, _opts -> to_string(i) end)
+      iex> doc =
+      ...>   Inspect.Algebra.container_doc(
+      ...>     "[",
+      ...>     Enum.to_list(1..5),
+      ...>     "]",
+      ...>     %Inspect.Opts{limit: :infinity},
+      ...>     fn i, _opts -> to_string(i) end
+      ...>   )
```
josevalim: Can we break this one into variables instead?

lib/elixir/lib/inspect/algebra.ex
```diff
@@ -324,18 +324,37 @@ defmodule Inspect.Algebra do
...
-      iex> doc = Inspect.Algebra.container_doc("[", Enum.to_list(1..5), "]",
-      ...>         %Inspect.Opts{limit: 3}, fn i, _opts -> to_string(i) end)
+      iex> doc =
+      ...>   Inspect.Algebra.container_doc(
+      ...>     "[",
+      ...>     Enum.to_list(1..5),
+      ...>     "]",
+      ...>     %Inspect.Opts{limit: 3},
+      ...>     fn i, _opts -> to_string(i) end
+      ...>   )
```
josevalim: Can we break this one into variables instead?

lib/elixir/lib/inspect/algebra.ex
```diff
@@ -324,18 +324,37 @@ defmodule Inspect.Algebra do
...
-      ...>         %Inspect.Opts{limit: 3}, fn i, _opts -> to_string(i) end, separator: "!")
+      iex> doc =
+      ...>   Inspect.Algebra.container_doc(
+      ...>     "[",
+      ...>     Enum.to_list(1..5),
+      ...>     "]",
+      ...>     %Inspect.Opts{limit: 3},
+      ...>     fn i, _opts -> to_string(i) end,
+      ...>     separator: "!"
+      ...>   )
```
josevalim: Can we break this one into variables instead?

lib/elixir/lib/inspect/algebra.ex
amatalai: Sure, I'll update it in next few minutes

josevalim:

PR #7979, opened 2018-07-23
Make Code.eval_string/2 doc example more formatter-compatible

amatalai:


josevalim:
I know the formatter will emit the second one but for this particular example, I think the current version is clearer. In an actual codebase base I would make binding a variable, but that's unnecessary here, so I think we should keep the current version. Thanks! :)

PR #7982, opened 2018-07-24
Improvements to the compiler

michalmuskala:
This is several loosely related changes to the compiler that either clean things up or improve performance. With all the changes, on some benchmarks the compiler performance improved about 5%.

### DIFF
--- a/lib/elixir/src/elixir_quote.erl
+++ b/lib/elixir/src/elixir_quote.erl
@@ -174,12 +174,12 @@ quote(Expr, Binding, Q, E) ->
   } || {K, V} <- Binding],
 
   TExprs = do_quote(Expr, Q, E),
-  {'{}', [], ['__block__', [], Vars ++ [TExprs] ]}.
+  {'{}', [], ['__block__', [], Vars ++ [TExprs]]}.
 
 %% Actual quoting and helpers
 
-do_quote({quote, _, Args} = Tuple, #elixir_quote{unquote=true} = Q, E) when length(Args) == 1; length(Args) == 2 ->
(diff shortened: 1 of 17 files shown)


lib/elixir/src/elixir_quote.erl
```diff
@@ -418,7 +430,8 @@ keystore(_Key, Meta, nil) ->
 keystore(Key, Meta, Value) ->
   lists:keystore(Key, 1, Meta, {Key, Value}).
 keynew(Key, Meta, Value) ->
-  case keyfind(Key, Meta) of
-    {Key, _} -> Meta;
-    _ -> keystore(Key, Meta, Value)
+  case lists:keymember(Key, 1, Meta) of
+    true -> Meta;
+    false when Value =:= nil -> Meta;
```
josevalim: It seems we always return `Meta` when the `Value` is `nil`. When is the value `nil` in the first place?

josevalim: I can't identify any case where the value would be `nil`.

michalmuskala: I just inlined the code from keystore that explicitly checks for the value being `nil`. Maybe it has some other uses there, but here it is useless?

josevalim: @michalmuskala so maybe we should move the `nil` check up, like we do for `keystore`. Or remove it altogether. Your call.

michalmuskala: Yeah, it seems `keynew` is only ever called with either a `line` or `counter` and a value of `nil` makes no sense in there. I'll remove the check.

PR #7986, opened 2018-07-25
Optimize ISO-8601 Parsing

dcuddeback:
This PR builds on the optimizations recently committed to the v1.7 branch: https://github.com/elixir-lang/elixir/commit/04c45b031a8e766969b865015b7ba5c1b4ba04d7. I found it to be cheaper to apply the TZ offset directly to the calendar date than to go through the conversions to and from ISO days. Since TZ offsets are no more than 24 hours, we avoid a lot of the complication of counting leap days over a range of years that's handled by doing calendar math in ISO days. At most, we have to carry over +/- 1 day. I added tests for carrying over month and year boundaries and to and from leap days to verify that the date carrying logic is correct.

Note that `calendar.valid_date?` and `calendar.valid_time?` are handled inside the `Date.new/4` and `Time.new/5` constructors and produce the same error messages.

The results look like this (measured on an AMD TR 1950X):

iso8601-speedup

* Dark green = Elixir 1.6.6
* Medium green = Elixir v1.7 HEAD
* Light green = This PR

The scale on the left is microseconds.

### DIFF
--- a/lib/elixir/lib/calendar/datetime.ex
+++ b/lib/elixir/lib/calendar/datetime.ex
@@ -525,6 +525,24 @@ defmodule DateTime do
         not calendar.valid_time?(hour, minute, second, microsecond) ->
           {:error, :invalid_time}
 
+        offset == 0 ->
+          datetime = %DateTime{
+            calendar: calendar,
+            year: year,
+            month: month,
+            day: day,
+            hour: hour,
+            minute: minute,
+            second: second,
+            microsecond: microsecond,
+            std_offset: 0,
+            utc_offset: 0,
+            zone_abbr: "UTC",
+            time_zone: "Etc/UTC"
+          }
+
+          {:ok, datetime, 0}
+
         is_nil(offset) ->
           {:error, :missing_offset}
 
--- a/lib/elixir/lib/calendar/naive_datetime.ex
+++ b/lib/elixir/lib/calendar/naive_datetime.ex
@@ -171,10 +171,35 @@ defmodule NaiveDateTime do
           Calendar.microsecond(),
           Calendar.calendar()
         ) :: {:ok, t} | {:error, atom}
-  def new(year, month, day, hour, minute, second, microsecond \\ {0, 0}, calendar \\ Calendar.ISO) do
-    with {:ok, date} <- Date.new(year, month, day, calendar),
-         {:ok, time} <- Time.new(hour, minute, second, microsecond, calendar),
-         do: new(date, time)
+  def new(year, month, day, hour, minute, second, microsecond \\ {0, 0}, calendar \\ Calendar.ISO)
+
+  def new(year, month, day, hour, minute, second, microsecond, calendar)
+      when is_integer(microsecond) do
+    new(year, month, day, hour, minute, second, {microsecond, 6}, calendar)
+  end
+
+  def new(year, month, day, hour, minute, second, microsecond, calendar) do
+    cond do
+      not calendar.valid_date?(year, month, day) ->
+        {:error, :invalid_date}
+
+      not calendar.valid_time?(hour, minute, second, microsecond) ->
+        {:error, :invalid_time}
+
+      true ->
+        naive_datetime = %NaiveDateTime{
+          calendar: calendar,
+          year: year,
+          month: month,
+          day: day,
+          hour: hour,
+          minute: minute,
+          second: second,
+          microsecond: microsecond
+        }
+
+        {:ok, naive_datetime}
+    end
   end
 
   @doc """


lib/elixir/lib/calendar/datetime.ex
```diff
@@ -889,6 +877,80 @@ defmodule DateTime do
     Calendar.ISO.add_day_fraction_to_iso_days(iso_days, -offset, 86400)
   end
 
+  defp apply_tz_offset_to_naive(%NaiveDateTime{} = naive_datetime, 0, _), do: naive_datetime
```
dcuddeback: This could conceivably be implemented as an optimization for `NaiveDateTime.add/3` instead, and that could be used inside `from_iso8601/2`. That would give `NaiveDateTime.add/3` very different performance characteristics depending on the size of the interval being added. I'm not sure how you would feel about that. I thought it might be an unwelcome surprise to some users, which is why I went with a private function and only optimized `from_iso8601/2`.

lib/elixir/test/elixir/calendar_test.exs
```diff
@@ -624,6 +624,195 @@ defmodule DateTimeTest do
              }
   end
 
+  test "from_iso8601/1 with tz offsets that carry over date boundry" do
+    # backwards past midnight
+    assert DateTime.from_iso8601("2017-06-02T02:30:15+04:00")
```
fertapric: Minor nitpick: what do you think about using this pattern 
```elixir
assert DateTime.from_iso8601("...") == {:ok, %DateTime{...}}
``` 
to remove the pipe `|>` and the `elem(1)`?

dcuddeback: I'd be okay with that. I was aiming to be consistent with the style of the other tests by copying from here:

https://github.com/elixir-lang/elixir/blob/2d4be9850b04ded001d73464c66c04ffcd00bcce/lib/elixir/test/elixir/calendar_test.exs#L529-L532

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -518,33 +518,21 @@ defmodule DateTime do
...
-              second,
-              microsecond
-            )
-            |> apply_tz_offset(offset)
-            |> from_iso_days("Etc/UTC", "UTC", 0, 0, calendar, precision)
-
-          {:ok, %{datetime | microsecond: microsecond}, offset}
+      with {:ok, date} <- Date.new(year, month, day, calendar),
+           {:ok, time} <- Time.new(hour, minute, second, microsecond, calendar),
+           {:ok, naive} = NaiveDateTime.new(date, time) do
```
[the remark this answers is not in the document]
dcuddeback: That sounds like a good idea. I can try that later today. (I'm going to be away on errands for a few hours.)

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -889,6 +877,80 @@ defmodule DateTime do
...
+        end
+
+      true ->
+        %{naive_datetime | day: day}
+    end
+  end
+
+  defp div_mod(int1, int2) do
+    div = div(int1, int2)
+    rem = int1 - div * int2
```
fertapric: Another minor nitpick: `rem = rem(int1, int2)`

fertapric: Nevermind. I see why you didn’t use it.

josevalim:
Thanks @dcuddeback!

Unfortunately I am not convinced about this PR. We are replacing 27 LOC by 278 LOC. Maybe a compromise would be to convert only the time to a day fraction and:

1. if the tzoffset does not cause an overlap, we can create the naive date time directly
2. if the tzoffset does overlap, we do the computation of today

The expensive part is precisely on the conversion to iso_days, so if we can avoid that, we should already reap a good part of the benefits. Thoughts?

dcuddeback:
@josevalim No worries. I understand your hesitation. I added the tests because I thought the extra logic for carrying the dates is tricky, which implies a maintenance burden, so I think your hesitation is fair.

To be honest, the most valuable part to me is optimizing the UTC strings. I don't commonly use timestamps that aren't UTC. My optimization was originally only for UTC, but I thought to generalize it for other TZ offsets before upstreaming it to make it more broadly applicable, which I thought would help make it acceptable in the standard library.

For UTC, skipping the ISO days conversion should be trivial, so optimizing only UTC timestamps would be another possible compromise. If that would be acceptable to you, that should be a quick change that I could have ready soon.

I like your suggestion, too. I don't know how many people use non-UTC timestamps, and I'd have to find time to look into the day fraction code. Is there any sense of how often people use non-UTC timestamps?

josevalim:
Yup. An optimization for UTC looks like a great starting point to me.
-- 


dcuddeback:
Okay. Updated to optimize UTC only. I also added @michalmuskala's suggestion for inlining the date and time constructors inside `NaiveDateTime.new/7`. Let me know if I've misunderstood that.

@michalmuskala Is https://github.com/elixir-lang/elixir/pull/7986/commits/46f057f5546f45f254176701cfa5619bed46e963 what you had in mind? That sped up the UTC benchmark by about 20% (from 1.8 to 1.4 microseconds on my computer).

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -519,6 +519,12 @@ defmodule DateTime do
       {hour, minute, second} = unquote(read_time)
 
       cond do
+        offset == 0 ->
```
josevalim: Actually, I think this will be even faster if you move this clause [here](https://github.com/dcuddeback/elixir/blob/46f057f5546f45f254176701cfa5619bed46e963/lib/elixir/lib/calendar/datetime.ex#L536). And then you never call NaiveDateTime, you just build the DateTime directly, [like done here](https://github.com/dcuddeback/elixir/blob/46f057f5546f45f254176701cfa5619bed46e963/lib/elixir/lib/calendar/datetime.ex#L205-L218).

fertapric: Micro-optimization: what do you think about?

```elixir
cond do
  not calendar.valid_date?(year, month, day) ->
    {:error, :invalid_date}
 
  not calendar.valid_time?(hour, minute, second, microsecond) ->
    {:error, :invalid_time}

  offset == 0 ->
    datetime = %DateTime{
      calendar: calendar,
      year: year,
      month: month,
      day: day,
      hour: hour,
      minute: minute,
      second: second,
      microsecond: microsecond,
      std_offset: 0,
      utc_offset: 0,
      zone_abbr: "UTC",
      time_zone: "Etc/UTC"
    }

    {:ok, datetime, 0}

    # Rest...
```

fertapric: You were a few seconds faster than me 😂 

I suggested to put it before `is_nil(offset)` to speed up the happy case, what do you think?

josevalim: Sounds good to me too!

dcuddeback: Makes sense. We won't be using the change that I made to `NaiveDateTime.new/7` anymore: https://github.com/elixir-lang/elixir/pull/7986/commits/46f057f5546f45f254176701cfa5619bed46e963. Still keep that? Or should I pull that out when I make this change?

dcuddeback: Sounds good.

dcuddeback: Wow. That made another ~18% improvement. Nice suggestion. Will push it up in a minute.

dcuddeback:
It's now 1.15 microseconds on my computer, down from 1.8 microseconds when I opened the PR. Thanks for the suggestions!

michalmuskala:
Thank you! :heart:

PR #7988, opened 2018-07-25
Improve compile_doc warning

goofansu:
`definition without a body` in the warning confused me when I upgraded 1.7.0 for my pheonix project.

I found in other documents, it is called `function head` or `header`.

For example:

1. https://github.com/elixir-lang/elixir/blob/691460a6e661f174476c69153776cba2f8a47341/lib/elixir/src/elixir_def.erl#L412-L426

2. https://github.com/elixir-lang/elixir/blob/caa90b563ade4818a4e49b3e20149a44ff0c00c8/lib/elixir/pages/Writing%20Documentation.md#function-arguments

Or else, is it better to add an example in the warning? Like `def foo(a, b)`

### DIFF
--- a/lib/elixir/lib/module.ex
+++ b/lib/elixir/lib/module.ex
@@ -1347,7 +1347,8 @@ defmodule Module do
           message =
             "redefining @doc attribute previously set at line #{current_line}. " <>
               "If you want to redefine a previously specified doc, " <>
-              "use a definition without a body after the original documentation"
+              "use a definition without a body after the original documentation\n" <>
+              "Example for definition without a body: def foo(a, b)\n"
 
           :elixir_errors.warn(line, env.file, message)
         end


josevalim:
Thanks! We were used to call it function head, but people found it confusing, and we generally agreed to call it definition without a body. Either way, I think including an example and going with a consistent name is the way to go.

goofansu:
screenshot of iterm2 2018-7-26 3-52-02 pm

@josevalim I updated the warning like above screenshot.

Giving it a short name is a must. I think `function head` is already suitable, it is just lack of explanation. For example, we can add this concept in the [getting-started guide](https://elixir-lang.org/getting-started/modules-and-functions.html#named-functions)

josevalim:

josevalim:
I was wrong, the term we settled on was indeed function head. I will finish wrapping this up locally. Thanks!

goofansu:
@josevalim Thank you. So it is called `function head` again, it is clear.

PR #7991, opened 2018-07-26
build: allow downstream packagers to tweak manpage generation

dch:
downstream packagers such as FreeBSD support building and shipping binary-style packages with or without manpages enabled, or in different locations, in addition to setting `PREFIX` or `SHARE_PREFIX`
directly.

This simple patch allows them to do so by setting `MAN_PREFIX` and Make will Do The Right Thing.

It should have no impact on users not using this feature, as the default config will fall through to the current value.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 PREFIX ?= /usr/local
 SHARE_PREFIX ?= $(PREFIX)/share
+MAN_PREFIX ?= $(SHARE_PREFIX)/man
 CANONICAL := master/ # master/ or vMAJOR.MINOR/
 ELIXIRC := bin/elixirc --verbose --ignore-module-conflict
 ERLC := erlc -I lib/elixir/include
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,5 +1,6 @@
 PREFIX ?= /usr/local
 SHARE_PREFIX ?= $(PREFIX)/share
+MAN_PREFIX ?= $(PREFIX)/man
```
michalmuskala: The old default destination was `$(SHARE_PREFIX)/man`, but now it's `$(PREFIX)/man`.

dch: @michalmuskala  fixed & pushed, sorry.

I got this 100% exactly the wrong way around, this is what I'm aiming for downstream BTW - https://www.freebsd.org/cgi/man.cgi?query=hier&apropos=0&sektion=0&manpath=FreeBSD+11.2-RELEASE+and+Ports&arch=default&format=html

> local/	  local executables, libraries, etc.  Also used as the default destination for the ports(7) framework.  Within local/, the general layout sketched out by hier for /usr should be used. Exceptions are the man directory (directly under local/ rather than under local/share/),

josevalim:

PR #8000, opened 2018-07-28
Consistently expand patterns and guards in assertions

### LINKED ISSUE #7999: Can't use `and` or `or` with guards in assert_receive
The following test fails to compile:
```
test "assert receive" do
    send(self(), [1, 2, 3])
    assert_receive [a, b, 3] when is_binary(a) and is_integer(b)
end
```
with the error `(ArgumentError) invalid expression in match, and is not allowed on patterns such as function clauses, case clauses or on the left side of the = operator`

This appears to be because of a combination of
https://github.com/elixir-lang/elixir/blob/master/lib/ex_unit/lib/ex_unit/assertions.ex#L439
puts the environment in a match context and
https://github.com/elixir-lang/elixir/blob/ffb931a5ac805ade8dca2e9dae78c0032655e3ce/lib/elixir/lib/kernel.ex#L1593
disallows the use of `or` in a match context

josevalim:
Closes #7999

### DIFF
--- a/lib/ex_unit/lib/ex_unit/assertions.ex
+++ b/lib/ex_unit/lib/ex_unit/assertions.ex
@@ -436,10 +436,9 @@ defmodule ExUnit.Assertions do
     binary = Macro.to_string(pattern)
 
     # Expand before extracting metadata
-    caller = Macro.Env.to_match(caller)
-    expanded = expand_pattern(pattern, caller)
-    vars = collect_vars_from_pattern(expanded)
-    pins = collect_pins_from_pattern(expanded, Macro.Env.vars(caller))
+    pattern = expand_pattern(pattern, caller)
+    vars = collect_vars_from_pattern(pattern)
+    pins = collect_pins_from_pattern(pattern, Macro.Env.vars(caller))
 
     pattern =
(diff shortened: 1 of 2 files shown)


lib/ex_unit/lib/ex_unit/assertions.ex
```diff
@@ -612,13 +611,20 @@ defmodule ExUnit.Assertions do
     |> elem(1)
   end
 
+  defp expand_pattern({:when, meta, [left, right]}, caller) do
+    left = expand_pattern_except_vars(left, Macro.Env.to_match(caller))
+    right = expand_pattern_except_vars(right, %{caller | context: :guard})
```
michalmuskala: Should we have a corresponding `Macro.Env.to_guard/1`?

josevalim: We created to_match because we were doing in that in 3 or 4 different places. I think to_guard is a good idea... but right now this will be its second use. And one of them is inside Kernel.Utils, where we can't invoke Macro.Env for bootstrap reasons. :(

josevalim:

Issue #6248, opened 2017-06-22
Improve ExUnit output for assertions on process mailbox


devonestes:
Moving this over from the mailing list per instructions from José. Here's the link to the mailing list discussion: https://groups.google.com/forum/#!topic/elixir-lang-core/qLO2IQ9PWUg

In short, the current output for ExUnit when you're asserting that a process receives a certain message is not very helpful, especially on messages with a large number of arguments. Some improvements in the output would make for a much better user experience.

Here is an example of the current behavior:
![mailbox](https://user-images.githubusercontent.com/8422484/27427366-02852c86-573f-11e7-85d1-d2f34ea31520.jpg)

That output makes finding the differences between the expected messages and the received messages hard. It would be really helpful to perform some more illustrative diffing to highlight the differences.

Props to @bradurani for finding this and recommending that it be improved upon!

kelvinst:
This is the next one I'll try. Not sure if I have the skills needed, but let's do it!

kelvinst:
Another thing: I'll start by maps and structs, and when I have them I'll open a WIP PR for you to review the work so far.

kelvinst:
This is actually a lot harder than I thought it would be, the differ will need to receive the AST of the left side and also the pinned variables. I will invest my time now on some easier issues and leave this one for later when I get some time to deeply study this.

Also, if someone is interested to help me on it, be my guest. 😉

jeremyowensboggs:
I'd like to have a look, kelvinst, do you already have some tests and or scenarios written up?

kelvinst:
@jeremyowensboggs nope! I was just studying ways to do it first. Feel free to do what you want, I'm not able to work on it right now. Thanks for asking!

jeremyowensboggs:
I've looked things over, and just want to talk over the approach out-loud.  Feel free to ignore or offer opinions/comments

Inside the assert macro, when we have a pattern, we will put the escaped AST of the pattern, a keyword list of the pinned vars and their values, and the unbound vars into a struct that we will assign to the `:left` option of `assert`.  I will call this struct ExUnit.Pattern.

In the ExUnit.Diff, create a `def script(%ExUnit.Pattern{} = left, right)` function to do the diff.
In ExUnit.Pattern, create a functions to step through the compare, and accumulate the tagged output

for the pattern
  is lhs a map, and is rhs a map? 
    pop the first key from lhs, and attempt to pop the same key from rhs
    if no rhs, {:del, key}
[...]

PR #8002, opened 2018-07-28
Pattern diff

jeremyowensboggs:
In regards to https://github.com/elixir-lang/elixir/issues/6248

Highlight where a message doesn't match the given pattern for assert_receive, match? and =


lib/ex_unit/lib/ex_unit/assertions.ex
```diff
@@ -108,6 +110,14 @@ defmodule ExUnit.Assertions do
     vars = collect_vars_from_pattern(left)
     pins = collect_pins_from_pattern(left, Macro.Env.vars(__CALLER__))
 
+    left_pattern = escape_quoted(:pattern, left)
```
josevalim: We should have a `escape_pattern` function instead of wrapping this in `pattern` just to immediately remove it.

lib/ex_unit/lib/ex_unit/assertions.ex
```diff
@@ -108,6 +110,14 @@ defmodule ExUnit.Assertions do
...
+
+    pattern_vars =
+      vars
+      |> Enum.map(fn {var, _, _} -> var end)
+      |> Enum.uniq()
+      |> Enum.map(&{&1, :ex_unit_unbound_var})
```
josevalim: You cannot trust variable names. Take this code:

```elixir
var = quote(do: right)
quote do: {right, unquote(var)}
```

They have the same name but they are not the same.

lib/ex_unit/lib/ex_unit/assertions.ex
```diff
@@ -485,15 +524,29 @@ defmodule ExUnit.Assertions do
...
         after
-          timeout -> flunk(unquote(failure_message))
+          timeout ->
+            case unquote(failure_message) do
+              {:flunk, msg} ->
+                flunk(msg)
+
+              {:pattern, messages, msg} ->
+                assert false,
+                  right: {:ex_unit_mailbox_contents, messages},
```
josevalim: We shuold just pass `messages` instead of a special tuple and have a key in `ExUnit.Pattern` that says if it is meant to match a single element or multiple.

lib/ex_unit/lib/ex_unit/assertions.ex
```diff
@@ -485,15 +524,29 @@ defmodule ExUnit.Assertions do
             timeout
           )
         end
+      end
 
     quote do
       timeout = unquote(timeout)
 
+      {:pattern, [], [left]} = unquote(left_pattern)
+      pattern = Pattern.new(left, unquote(pins), unquote(pattern_vars))
```
josevalim: We should build a pattern only if there is an error.

lib/ex_unit/lib/ex_unit/assertions.ex
jeremyowensboggs: Is the meta data ever significant for variables, or is it just the first element and the last element that are significant?

josevalim: A variable is uniquely identified by this: `fn {var, meta, context} -> {var, meta[:counter] || context} end`.

lib/ex_unit/lib/ex_unit/assertions.ex
```diff
@@ -62,6 +62,8 @@ defmodule ExUnit.Assertions do
   or handling exceptions.
   """
 
+  alias ExUnit.Pattern
+
   @doc """
```
fertapric: There are a few examples in the documentation that should include the new `left:` key.

josevalim:
Also please rebase and take especially this commit in consideration: https://github.com/elixir-lang/elixir/pull/8039 :)

lib/ex_unit/lib/ex_unit/assertions.ex
jeremyowensboggs: Hi, would you elaborate please, I don't understand.

fertapric: Sure.

For instance, [this example](
https://github.com/elixir-lang/elixir/pull/8002/files#diff-699b0020c6a196231a7d473322f1744dR91) should be updated to:

```elixir
match (=) failed
code:  [1] = [2]
left:  [1]
right: [2]
```

like you did in [this test](
https://github.com/elixir-lang/elixir/pull/8002/files#diff-dcedcba19e3df61c281c3e36e50e53d9R797).

lib/ex_unit/lib/ex_unit/assertions.ex
```diff
@@ -156,11 +170,24 @@ defmodule ExUnit.Assertions do
     match? = {:match?, meta, [left, Macro.var(:right, __MODULE__)]}
     pins = collect_pins_from_pattern(left, Macro.Env.vars(__CALLER__))
```
jeremyowensboggs: Should the collection fo the pins & vars, as well as the expand_pattern be moved from the assertions module into the pattern module?

lib/ex_unit/lib/ex_unit/assertions.ex
jeremyowensboggs: Thanks.  I updated the vars.  I had a go at the pins, but was not able to get a pinned var from another context - is that true, or just a lack of imagination on my part?

jeremyowensboggs:
Updated with changes and rebased, taking special consideration of #8039

jeremyowensboggs:
Hi, is there something else I should be doing with this PR?

josevalim:
@jeremyowensboggs no, sorry. It is blocking on us, but I am personally focused on getting Ecto 3.0 out.

lib/ex_unit/lib/ex_unit/pattern_diff.ex
```diff
@@ -0,0 +1,391 @@
...
+  # {:%{}, _, members}
+  # {:{}, _, members}
+  # {:_, _, _}
+  # {var, _, context}
+  # {key, val}
+  # list
+  # integer
+  # float
+  # string
+  # atom
```
josevalim: What abut binary patterns `{:<<>>, _, _}`?

lib/ex_unit/test/ex_unit/pattern_format_test.exs
```diff
@@ -0,0 +1,581 @@
...
+      "1",
+      ", ",
+      "2",
+      ", ",
+      "3",
+      ", ",
+      {:diff_insert, "4"},
+      "]",
+      " ",
+      {:diff_delete, "when is_binary(a)"}
```
josevalim: This is interesting. We are showing the guard but we don't have a variable. I am wondering if this can be confusing or not.

Another approach would be to show the diff on the pattern. But this complicates things for `assert_receive`. I.e. we would need to show the pattern alongside each attempted message.

What are your thoughts between the approaches?

lib/ex_unit/lib/ex_unit/pattern_diff.ex
```diff
@@ -0,0 +1,391 @@
...
+
+  def cmp(pattern, r) do
```
eksperimental: Could we rename this function to "compare", I don't think it's good practice to use abbrevation for function names in Elixir core

fertapric:
What do you think about namespacing all the "pattern" logic into `ExUnit.Pattern` (`ExUnit.Pattern.Diff`, `ExUnit.Pattern.Format`...)?

josevalim:
Folks, this is a large PR that will likely go through a couple changes still. I believe we should first focus on reviewing the functionality and, once all is said and done, we can start discussing naming, code organization, etc.

lib/ex_unit/lib/ex_unit/pattern_diff.ex
jeremyowensboggs: So far I have punted on those, focusing on the lists & maps, which I think covers most of what is desired.  Since the approach for those will most likely be the same as we do here, I would prefer to work out any issues with this PR before getting too deep in them?  Thoughts?

lib/ex_unit/test/ex_unit/pattern_format_test.exs
jeremyowensboggs: I went back and forth between the variable, and value, and I can be easily talked into either.   I think showing `when is_binary(2)` in red actually makes it easier to see what the problem is, but `when is_binary(a)` points directly to the pattern - and highlights that this is being brought in from the pattern, and not a part of the item being matched.  What are your preferences?

lib/ex_unit/lib/ex_unit/pattern_diff.ex
jeremyowensboggs: Yes.

lib/ex_unit/lib/ex_unit/pattern_diff.ex
josevalim: Makes sense to punt. Doing a full diff on binaries would also be hard so it is likely that we will just compile them to something like `match?(binary_pattern, value)` and treat the whole thing as true or false.

lib/ex_unit/test/ex_unit/pattern_format_test.exs
josevalim: I think was not clear enough.I was thinking about a different approach where the diff actually applies to the pattern. So in such cases, we would show the pattern using green/red and not the value. I can see some advantages of this approach:

For example, on a match like:

    %{foo: _, bar: _} = really_large_map

where the bar key is missing, it would show foo in green, bar in red, so it would be clearer that the bar key is missing. If we had a large map on the right side, we can't show misses.

This approach would also work better with guards.

This approach can also be useful in other situations, for example, in the `UndefinedFunctionClauseError`, we could perform a better diff there too, as we show the patterns there.

The only problem I can think is regarding `assert_receive`, where we would need to show the pattern alongside each value... but maybe it is not that bad?

Thoughts?

jeremyowensboggs: Maybe more of an `and` than an `or`?  Colored pattern on the left, colored map on the right?  I think it would just be screen spam for assert_receive if there were more than couple of messages in the mailbox.  However, we now have a mechanism to determine how closely a messages matches, so we could zero in on a one  'best choices'?

josevalim: I am worried about the complexity of supporting both approaches. But I am not sure how to handle `assert_receive`. Maybe we should try some "mock ups" of how the CLI would look like? I can send something soon. :D

lib/ex_unit/lib/ex_unit/assertions.ex
fertapric: @jeremyowensboggs it seems to be intended https://github.com/elixir-lang/elixir/commit/6931b9d7be8b7f88439bd2434fb1256b58d1f2a9

lib/ex_unit/lib/ex_unit/formatter.ex
```diff
@@ -303,6 +334,16 @@ defmodule ExUnit.Formatter do
     padding <> Enum.join(reasons, "\n" <> padding) <> "\n"
   end
 
+  defp format_sides(%{left: %ExUnit.Pattern{} = left, right: right}, formatter, inspect) do
+    case format_pattern(left, right, formatter) do
+      {l, r} ->
+        {IO.iodata_to_binary(l), IO.iodata_to_binary(r)}
+
+      nil ->
+        {if_value(left, inspect), if_value(right, inspect)}
```
fertapric: I think `left` is always going to have a value here, so I think `left.binary` can be used directly here.

In addition, I would remove the `else` clause from `format_pattern/3`, and handle`nil`s here because with the current implementation if `pattern_diff/2` returns `nil`, I think `if_value(left, inspect)` will print the `Pattern` struct.

lib/ex_unit/test/ex_unit/pattern_format_test.exs
```diff
@@ -0,0 +1,581 @@
...
+
+    expected = ["[", "1", ", ", "2", ", ", "3", "]", " ", {:diff_delete, "when is_binary(a)"}]
+    assert actual == expected
+  end
+
+  test "using guards, but guard matches" do
+    # send(self(), [1, 2, 3])
+    # assert_receive [a, b, 3, 4] when is_integer(a)
+
+    left = quote do: [a, b, 3, 4] when is_binary(a)
```
fertapric: I think this should be `is_integer(a)` so the guard matches

lib/ex_unit/test/ex_unit/pattern_format_test.exs
fertapric: Here are some mockups:

screen shot 2018-10-10 at 01 40 22

I'll work tomorrow on the ones for `assert_receive` 🙂

josevalim: Thanks @fertapric! Were those mockups generated by hand or are you running this PR?

fertapric: By hand. 

I think we can use them to discuss the user experience of the CLI, and then check the implementation.

This is what I used: https://gist.github.com/fertapric/c7ee41e890214b1123e76f17dfd288c4. The code couldn't be uglier 🙈, but it works!

lib/ex_unit/test/ex_unit/pattern_format_test.exs
```diff
@@ -0,0 +1,581 @@
...
+    # send(self(), [1, 2, 3])
+    # send(self(), [1, 2])
+    # send(self(), [1, 2, 3, 4, 5])
+    # assert_receive [1, 2, 3, 4]
+
+    left = quote do: [1, 2, 3, 4]
+    pattern = Pattern.new(left, [], %{})
+
+    actual = PatternFormat.script(pattern, [1, 2, 3])
+    expected = ["[", "1", ", ", "2", ", ", "3", ", ", {:diff_insert, "4"}, "]"]
```
josevalim: I just want to note that we were used to this when we first added diffing and it was very confusing and everyone was puzzled. Basically, it is hard for us to see something like `[1, 2, 3, 4]` where `, 4` is green and understand it as "the list is actually only `[1, 2, 3]` and `, 4` is what is missing". That's when we decided for the original diff implementation that we should never change what is printed, we can only annotate it with colors. This means we can show only reds or greens.

lib/ex_unit/lib/ex_unit/formatter.ex
```diff
@@ -130,6 +130,37 @@ defmodule ExUnit.Formatter do
     format_assertion_error(%{}, struct, [], :infinity, fn _, msg -> msg end, "")
   end
 
+  def format_assertion_error(
+        test,
+        %{left: %ExUnit.Pattern{match_single?: false} = pattern, right: mailbox} = struct,
```
fertapric: What do you think about adding the concept of "mailbox" instead of `:match_single?`? I think `ExUnit.Pattern` should represent just the pattern, and not how it should be applied to the right side. The right side would be then either a value or a "mailbox".

Beside, this structure would make more sense if we finally end up reusing part of this logic in `FunctionClauseError`s:

> This approach can also be useful in other situations, for example, in the `UndefinedFunctionClauseError`, we could perform a better diff there too, as we show the patterns there.

In short words, I'm suggesting to rewrite this line to: 
```elixir
%{left: %ExUnit.Pattern{} = pattern, right: %ExUnit.Mailbox{} = mailbox} = struct
```
Thoughts?

jeremyowensboggs: I like the suggestion.

josevalim:
I believe that diffing the pattern is the way to go, with one caveat. In this comment I want to explain why.

## Assumptions

We will assume that will only generate a diff of the pattern or of the value. At least as an initial goal, we won't diff both due to the complexity. This introduces one important limitation.

## Limitation

If we will only diff the pattern or the value, then we are only capable of showing absences. That's because showing green/red on the same is extremely confusing.

Let's understand why. Consider the two examples below:

    assert [1, 2, 3] = [1, 2, 3, 4]
    assert [1, 2, 3, 4] = [1, 2, 3]

This is the failure output today:

    pattern: [1, 2, 3]
      value: [1, 2, 3, 4]

    pattern: [1, 2, 3, 4]
      value: [1, 2, 3]

Assuming that you can only diff the values, here is the output (where + means green, and - means red) considering that we will show both additions (in +, representing green) and removals (in -, representing red):

    pattern: [1, 2, 3]
      value: [1, 2, 3-, 4-]

    pattern: [1, 2, 3, 4]
      value: [1, 2, 3+, 4+]

Our experience with the original diff implementation is that this approach is **very confusing**, as in the second case, we will literally see the list `[1, 2, 3, 4]` (with `, 4` in green) but the actual list is `[1, 2, 3]`.

This issue only happens with diffs on patterns, but in the opposite order:

    pattern: [1, 2, 3+, 4+]
      value: [1, 2, 3, 4]

    pattern: [1, 2, 3-, 4-]
      value: [1, 2, 3]

Obviously, the solution here would be to perform diffs on both patterns and values:

    pattern: [1, 2, 3]
      value: [1, 2, 3-, 4-]

    pattern: [1, 2, 3-, 4-]
      value: [1, 2, 3]

But our assumption is that this approach is too complex, so we can't do it. So what happens when the entity we are diffing (a pattern or a value) has too few entries? I think the simplest suggestion is to simply mark everything as failed. For patterns:

    pattern: -[1, 2, 3]-
      value: [1, 2, 3, 4]

    pattern: [1, 2, 3-, 4-]
      value: [1, 2, 3]

For values:

    pattern: [1, 2, 3]
      value: [1, 2, 3-, 4-]

    pattern: [1, 2, 3, 4]
      value: -[1, 2, 3]-

Alternatively, we could show only the opening and/or closing brackets in red.

With this limitation out of the way, we still need to answer, between diffing on values and diffing on patterns, which one is the best approach? When we will need to resort to techniques like the above less frequently?

## Diffs on patterns vs diffs on values

We need to consider the following scenarios:

  1. For list (as well as tuples and strings)
    a. When the list on the pattern has too many elements
    b. When the list on the value has too many elements

  2. For maps
    a. When the map on the pattern has too many elements
    b. When the map on the value has too many elements

  3. Guards - when a guard fails

  4. `assert_receive`

  5. Function clause errors

### For list (tuples and strings)

When a list on the pattern has too many elements and we diff on patterns, we will have:

    pattern: [1, 2, 3-, 4-]
      value: [1, 2, 3]

which is correct.

However, when we diff on values, we need to resort to the "technique" from the previous section and mark the whole value in red:

    pattern: [1, 2, 3, 4]
      value: -[1, 2, 3]-

The opposite occurs when the value is the one with too many elements. When we diff on patterns, we will have to show the whole pattern in red:

    pattern: -[1, 2, 3]-
      value: [1, 2, 3, 4]

This means that for lists, tuples and strings, there is no advantage for diffing on patterns or values. Whenever they don't have the same amount of elements, patterns will be more precise in half of the scenarios and the values in the other half.

### For maps

For maps though, we have a different story, since maps in patterns only match a subset.

When the map on the pattern has too many elements and we diff on the pattern, we show exactly what is missing:

    pattern: %{a: 1, -b: 2-}
      value: %{a: 1}

However, when we diff on the value, we need to show the whole value in red:

    pattern: %{a: 1, b: 2}
      value: -%{a: 1}-

On the other hand, when the map on the value has too amny elements, it will never be a report, because the map will always match. In other words, this will never be an error report:

    pattern: %{a: 1}
      value: %{a: 1, b: 2}

This means that, for working with maps, diffing on the pattern will always be superior.

### For guards

Because guards only exist in the pattern, it is natural to show each guard that fails in red when diffing on the pattern.

If we can only diff values, the only way to make this work is to copy the guards to the values too. On the positive side, this can be an opportunity for showing the guards with values filled in, which can be seen as a bonus:

    pattern: %{a: a} when is_integer(a)
      value: %{a: "foo"} when -is_integer("foo")-

However, there is nothing stopping us from adding more information when diffing on patterns. For example, we could say which variables have been bound, since we already say which variables were pinned:

    The following variables were bound:
      a = "foo"
    pattern: %{a: a} when -is_integer(a)-
      value: %{a: "foo"}

While I personally believe that diffing on the pattern here is more natural, showing the guards alongside the values can be useful too.

### For assert_receive

`assert_receive` is different to everything we have discussed so far because here we want to match multiple values against the same pattern. Today we have this report:

    No message matching {:ok, 3} after 100ms.
    Process mailbox:
      {:ok, 1}
      :message_in_my_inbox
      {:ok, 2}
      :another_message

Since today's report emits multiple values, we can easily adapt it to diff on each of those values:

    No message matching {:ok, 3} after 100ms.
    Process mailbox:
      {:ok, -1-}
      -:message_in_my_inbox-
      {:ok, -2-}
      -:another_message-

However, if we are diffing on pattern, the report gets more verbose, as we would need to show each pattern alongside the value:

    No message matching {:ok, 3} after 100ms.
    Process mailbox (4 messages):

      pattern: {:ok, -3-}
      message: {:ok, 1}

      pattern: -{:ok, 3}-
      message: :message_in_my_inbox

      pattern: {:ok, -3-}
      message: {:ok, 2}

      pattern: -{:ok, 3}-
      message: :another_message

Can anybody think of a better approach to format those messages when diffing on patterns? Without other solutions, we can safely say that the diffing on the values is simpler and less verbose.

### Function clause errors

It is worth reminding though that, outside of the test framework, there is at last one other situation we were match values against patterns. Which is in undefined function clause errors:

    ** (FunctionClauseError) no function clause matching in Access.fetch/2

    The following arguments were given to Access.fetch/2:

        # 1
        :foo

        # 2
        :bar

    Attempted function clauses (showing 5 out of 5):

        def fetch(%module{} = container, key)
        def fetch(map, key) when is_map(map)
        def fetch(list, key) when is_list(list) and is_atom(key)
        def fetch(list, key) when is_list(list)
        def fetch(nil, _key)

The implementation today has a very simply diff algorithm that only matches arguments against arguments, without any depth. If we perform diffing on patterns, we could also use the same algorithm to improve function clause errors themselves, which is a clear advantage to perform diffs on patterns.

## Summary

This comment proposes to perform diffing only on patterns or values and argues that, between the two, performing diffs on patterns yields better results, except on `assert_receive`.

lib/ex_unit/test/ex_unit/pattern_format_test.exs
jeremyowensboggs: I assume that the 'code:', 'left:', 'right:' and 'stacktrace:' are supposed to be in the existing blue, and not green as shown here?   I have a slight concern that this depicts symmetrically sized left & right, and in practice, we generally resort to asserting a pattern match primarily for large, nested structs, so there is almost always asymmetry in size between the pattern on the left (a short value), and the target on the right (often a very long value).   But I think that what is detailed here will work well.

fertapric:
`[head | tail]` expressions are all or nothing?

```
   code: assert [2 | _] = [1, 2, 3]
pattern: -[2 | _]-
  value: [1, 2, 3]

   code: assert [%{a: 1, b: 2} | _] = [%{a: 1}, 2, 3]
pattern: -[%{a: 1, b: 2} | _]-
  value: [%{a: 1}, 2, 3]
```

or we diff in that case?

```
   code: assert [2 | _] = [1, 2, 3]
pattern: [-2- | _]
  value: [1, 2, 3]

   code: assert [%{a: 1, b: 2} | _] = [%{a: 1}, 2, 3]
pattern: [%{a: 1, -b: 2-} | _]
  value: [%{a: 1}, 2, 3]
```

jeremyowensboggs:
In practice, I have found that most of the time of the time, the pattern is the standard by which we measure the value.  In terms of which one is going to need to be changed to make the test pass, it is almost always going the be the value.  I'm concerned if we focus on where the pattern conflicts, instead of where the value violates the pattern, then we are not helping the user zero in on where the actual problem is (perhaps deep in a large nested map?)

josevalim:
@jeremyowensboggs that's a good point. Maybe diffing on the pattern is better to tell us what went wrong but diffing on the value is better to tell us where it went wrong? I am certain though that any of them is better than none of them, which is why I think focusing in one of them is enough for now. The pattern one seems to be generally more useful.

Also, regardless of the conclusion of this discussion, I would suggest for you to keep this branch and start the work on a new branch built on top of this one, just in case we decide to explore a new venue and we run into a dead-end.

josevalim:
@fertapric we would diff for cons.

jeremyowensboggs:
Like to request a quick sanity check -
1) Formatting only the pattern for now
screen shot 2018-11-16 at 11 38 12 am
Is this what we have been talking about?

josevalim:
@jeremyowensboggs that looks perfect to me!

ggcampinho:
I know that @jeremyowensboggs is working a long time in this issue, but some months ago I decided to also give a try just in case because I really wanted this feature. I struggled in some points and had to go back and forward a few times until I found a way to represent all the cases without problems. In my experience, the pattern and the value diffs need to be apart from each other to cover all the cases, otherwise, it's confusing when you trying to print the diff.

Anyway, I got matcher working for the cases that could come up with and I believe it covers most of the issues, but maybe there are a few corner cases missing. The matcher also does Myers difference in the list and partially in the map, so I believe it gives similar results to the assertions with `==`.

I don't want to step on anyone's foot, but I want to at least share [the test cases](https://github.com/ggcampinho/elixir/blob/match-diff/lib/elixir/test/elixir/matcher/match_test.exs) and [the matcher code](https://github.com/ggcampinho/elixir/blob/match-diff/lib/elixir/lib/code/matcher.ex) to give some guidance on how this may work.

jeremyowensboggs:
I redid the formatting to match what we talked about.  I simplified the formatting of the value, but left formatting of the value disabled for now, and just highlight where the pattern doesn't match.  For assert_receive, we print the pattern & the value for each of the last 10 items in the mailbox.  It is on the branch https://github.com/jeremyowensboggs/elixir/tree/format-pattern-diff - should I merge it into this branch?  How should we proceed?
@ggcampinho thanks for the examples, it looks like you covered some things that I did not.  What are you proposing?

josevalim:
@jeremyowensboggs awesome! You can send a separate PR and maybe @ggcampinho can drop some comments in there and contribute some of the missing pieces in another step. :)

josevalim:
I will close this in favor of the new PR (#8440) but, if we decide to roll back, we can always reopen it. :) Thanks everyone for the feedback so far!

PR #8003, opened 2018-07-29
Fix formatting in docs for access.ex

FreedomBen:
The code block was rendering as regular text because it needs one more
space to be considered code.


josevalim:

PR #8004, opened 2018-07-29
Speed up from_iso8601 when offset does not change days

josevalim:
The speed up comes from the fact we don't convert to and
from iso days for the date part in certain cases.

Discussed on #7986. /cc @dcuddeback.

### DIFF
--- a/lib/elixir/lib/calendar/datetime.ex
+++ b/lib/elixir/lib/calendar/datetime.ex
@@ -547,22 +547,36 @@ defmodule DateTime do
           {:error, :missing_offset}
 
         true ->
-          {_, precision} = microsecond
+          day_fraction = Calendar.ISO.time_to_day_fraction(hour, minute, second, {0, 0})
 
-          datetime =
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/calendar/iso.ex
```diff
@@ -207,7 +207,6 @@ defmodule Calendar.ISO do
 
   # Converts count of days since 0000-01-01 to {year, month, day} tuple.
   @doc false
-  @doc since: "1.5.0"
```
fertapric: I'm not sure if this comment is out of the scope of this PR 😅  

But, there are other cases within the same module with `@doc false` and `@doc since: "1.5.0` :)

josevalim: Good catch. I have done in a separate commit here but also pushed to master. So we should be see an empty diff for that file here now. :)

dcuddeback:
Looks good to me. This goes from about 4.5 to 2.9 microseconds on my computer, so ~35% faster!

josevalim:

PR #8005, opened 2018-07-29
Fix warning for unused var in Module

fertapric:
Fixes:

```
==> elixir (compile)
warning: variable "body" is unused
  lib/module.ex:1334
```


josevalim:

PR #8006, opened 2018-07-29
Treat warnings as errors

michalmuskala:


josevalim:
@michalmuskala I think we should always enable them then.

josevalim:
> Edit2: By doing so, the CI will not be allowed to test soft-deprecated code as they emit warnings ? 🤔 that is bad

This covers only compilation, we should be fine for the tests themselves!

josevalim:

zkessin:
there should be a way to say only in my code and not in mix dependencies

fertapric:
@zkessin I think this is only applied to the Elixir codebase. Personal Elixir projects are not affected by this PR.

zkessin:
Thanks

PR #8007, opened 2018-07-29
Registry: Associate value to process using via tuple

electricshaman:
When using Registry with `:via` tuples, in some cases it may be desired to associate a value to the process at the same time.  Currently this is not possible, resulting in a value of `nil` unless you call `Registry.update_value` as an additional step after the process has started.

This PR introduces an additional `:via` tuple element for the value to be associated with the process (while retaining the existing behavior, of course):

      {:ok, _} = Registry.start_link(keys: :unique, name: Registry.ViaTest)
      name = {:via, Registry, {Registry.ViaTest, "agent", :hello}}
      {:ok, _} = Agent.start_link(fn -> 0 end, name: name)
      Registry.lookup(Registry.ViaTest, "agent")
      #=> [{self(), :hello}]

Thoughts?

### DIFF
--- a/lib/elixir/lib/registry.ex
+++ b/lib/elixir/lib/registry.ex
@@ -36,6 +36,22 @@ defmodule Registry do
       Agent.get(name, & &1)
       #=> 1
 
+  In the previous example, we were not interested in associating a value to the
+  process:
+
+      Registry.lookup(Registry.ViaTest, "agent")
+      #=> [{self(), nil}]
+
+  However, in some cases it may be desired to associate a value to the process
+  using the alternate `{:via, Registry, {registry, key, value}}` tuple:
+
+      {:ok, _} = Registry.start_link(keys: :unique, name: Registry.ViaTest)
+      name = {:via, Registry, {Registry.ViaTest, "agent", :hello}}
+      {:ok, _} = Agent.start_link(fn -> 0 end, name: name)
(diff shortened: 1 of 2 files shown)


lib/elixir/lib/registry.ex
```diff
@@ -209,10 +225,22 @@ defmodule Registry do
     end
   end
 
+  @doc false
+  @doc since: "1.x.x"
+  def whereis_name({registry, key, _value}) do
+    whereis_name({registry, key})
```
michalmuskala: I'd prefer if we had helper functions called from both callbacks, for example:

```
def whereis_name({registry, key, _value}), do: whereis_name(registry, key)
def whereis_name({registry, key}), do: whereis_name(registry, key)

defp whereis_name(registry, key) do
```

electricshaman: Great point.  Congrats on joining the core team, btw!

electricshaman:
I'm assuming I don't need to worry about adding a catch-all function for `whereis_name/2` or `register_name/1`?  The other callbacks don't have one defined, so I'm assuming we want to defer handling FunctionClauseError to the client code.

josevalim:
@electricshaman yup, functionclauseerror is fine!

michalmuskala:
Thank you! :heart:

PR #8011, opened 2018-07-30
Improve test suite for unnecessary quotes

fertapric:
Tests for uppercase atoms/keywords would pass even if the code was emitting the "unnecessary quotes" warning.

### DIFF
--- a/lib/elixir/test/elixir/kernel/warning_test.exs
+++ b/lib/elixir/test/elixir/kernel/warning_test.exs
@@ -39,18 +39,22 @@ defmodule Kernel.WarningTest do
     assert output =~ "found atom \":foo!\", ending with \"!\""
   end
 
-  test "unnecessary quotes" do
-    _ = Code.eval_string(~s/:"Foo"/)
-    _ = Code.eval_string(~s/["Foo": :bar]/)
+  describe "unnecessary quotes" do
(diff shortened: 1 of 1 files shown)


lib/elixir/test/elixir/kernel/warning_test.exs
```diff
@@ -39,18 +39,22 @@ defmodule Kernel.WarningTest do
     assert output =~ "found atom \":foo!\", ending with \"!\""
   end
 
-  test "unnecessary quotes" do
-    _ = Code.eval_string(~s/:"Foo"/)
-    _ = Code.eval_string(~s/["Foo": :bar]/)
+  describe "unnecessary quotes" do
+    test "does not warn for unnecessary quotes in uppercase atoms/keywords" do
+      refute capture_err(fn -> Code.eval_string(~s/:"foo"/) end) =~ "found quote atom"
```
josevalim: This should be in uppercase.

Also it is better to write `assert capture(fn -> ... end) == ""` as we should check there is no warning. If we check for "found quote atom", we may change the error message and then we would be asserting from the wrong thing.

fertapric: Right 🙈

It seems I pushed some changes I did to check the tests were failing 😅

> Also it is better to write `assert capture(fn -> ... end) == ""` as we should check there is no warning

Great idea!

PR #8017, opened 2018-07-31
Remove :since metadata when @doc false

fertapric:
Similar to [this commit](https://github.com/elixir-lang/elixir/commit/233341699d303f853cd808567f73df5731e511fa).

I've checked all the codebase, these were the only cases I found.


josevalim:

PR #8018, opened 2018-07-31
Change name of first argument of StringIO.open/2 to string

lackac:
Caught this while perusing the docs. `string` makes more sense here.


josevalim:

Issue #8015, opened 2018-07-30
mix help XXX when XXX is both a task and an alias prints the alias help


davydog187:
### Environment

* Elixir & Erlang/OTP versions (elixir --version): 1.7.1 / OTP 21
* Operating system: macOS 10.13.6

### Current behavior

Running `mix help XXX` when `XXX` is the name of both the name of a mix alias and a mix task prints the help data for the alias. 

This was surprising to me when I ran `mix help test` and got 

```elixir
$ mix help test

mix test

Alias for ["ecto.create --quiet", "ecto.migrate", "test"]

Location: mix.exs
```

### Expected behavior

I'm not sure exactly what should happen.

1. mix could let you know that there are multiple mix tasks with that name defined, and provide a way to display help for all of them individually.
2. mix could default to displaying the mix task over the alias. This makes sense in my case but maybe not for other cases. Its unclear to me.

Happy to tackle this one myself, but would appreciate feedback on how to proceed.

fertapric:
> 1. mix could let you know that there are multiple mix tasks with that name defined, and provide a way to display help for all of them individually.

I think we could include a flag in `mix help` to show the documentation of tasks, ignoring aliases. Something like `mix help --task TASK`.

And as mentioned by @davydog187, in case of conflict display, a hint under the documentation of the alias:

```
$ mix help test

mix test

Alias for ["ecto.create --quiet", "ecto.migrate", "test"]

Location: mix.exs

This alias augments an existent task, run `mix help --task test` to get the 
documentation of the underlying task.
```

As a side note, I think showing the documentation of the augmented task is important due to:

[Mix docs](https://hexdocs.pm/mix/Mix.html)
> Arguments given to the alias will be appended to the arguments of the last task in the list, if the last task is a function they will be given as a list of strings to the function.

If the last task is the one being augmented, users will need access the documentation of the task to know the options.

josevalim:
I think we should show both. It will be the better use experience. The documentation for alias is always short. So we can start with the alias and add a note that says "There is also a tasked named \"test\". The documentation is shown next.".

fertapric:
I forgot to mention this, but alias can override an existent task without augmenting it. In that case, the task cannot be executed so I'm not sure if we should show documentation for it.

fertapric:
Showing both sound good to me.

davydog187:
Great! Starting on a PR now

PR #8020, opened 2018-07-31
mix help when theres a task and alias

davydog187:
Addresses #8015

### DIFF
--- a/lib/mix/lib/mix/tasks/help.ex
+++ b/lib/mix/lib/mix/tasks/help.ex
@@ -86,11 +86,40 @@ defmodule Mix.Tasks.Help do
   def run([task]) do
     loadpaths!()
 
-    {doc, location} = verbose_doc(task)
     opts = Application.get_env(:mix, :colors)
 
+    opts =
+      if ansi_docs?(opts) do
+        [width: width()] ++ opts
+      else
+        opts
+      end
+
+    case verbose_doc(task) do
+      [single] ->
+        print_doc(task, single, opts)
+
+      [alias_doc, task_doc] ->
+        print_doc(task, alias_doc, opts)
+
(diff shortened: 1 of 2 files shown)


lib/mix/lib/mix/tasks/help.ex
```diff
@@ -86,11 +86,40 @@ defmodule Mix.Tasks.Help do
...
+    case verbose_doc(task) do
+      [single] ->
+        print_doc(task, single, opts)
+
+      [alias_doc, task_doc] ->
+        print_doc(task, alias_doc, opts)
+
+        note = "\nThere is also a task named \"#{task}\". The documentation is shown next.\n"
+
+        if ansi_docs?(opts) do
```
fertapric: I think this can be a simple `IO.puts` since there are not any special formatting.

davydog187: What about width? I think it needs to be considered

fertapric: You are right 😓 Sorry for the noise

lib/mix/lib/mix/tasks/help.ex
```diff
@@ -86,11 +86,40 @@ defmodule Mix.Tasks.Help do
...
+      end
+
+    case verbose_doc(task) do
+      [single] ->
+        print_doc(task, single, opts)
+
+      [alias_doc, task_doc] ->
+        print_doc(task, alias_doc, opts)
+
+        note = "\nThere is also a task named \"#{task}\". The documentation is shown next.\n"
```
josevalim: What if you add this note to `alias_doc` in the line above?

    alias_doc <> "\nThere is also a task named \"#{task}\". The documentation is shown next.\n"

This way you don't need to check for `ansi_docs?(opts)` and the like.

josevalim:

PR #8021, opened 2018-07-31
Remove duplicate myers implementation

josevalim:
This is done by allowing custom diff scripts when traversing
the list.

### DIFF
--- a/lib/elixir/lib/list.ex
+++ b/lib/elixir/lib/list.ex
@@ -863,27 +863,43 @@ defmodule List do
   corresponding key is `:del`), or left alone (if the corresponding key is
   `:eq`) in `list1` in order to be closer to `list2`.
 
+  A `diff_script` function can be given in case it is desired to compute
+  nested differences. The function may return a list with the inner edit
+  script or `nil` in case there is no such script. The returned inner edit
+  will be under the `:diff` key.
+
   ## Examples
 
       iex> List.myers_difference([1, 4, 2, 3], [1, 2, 3, 4])
       [eq: [1], del: [4], eq: [2, 3], ins: [4]]
 
+  To handle a list of words, we can pass `String.myers_difference/2` for
+  the inner diffs:
+
+      iex> List.myers_difference(["a", "db", "c"], ["a", "bc"], &String.myers_difference/2)
+      [eq: ["a"], diff: [del: "d", eq: "b", ins: "c"], del: ["c"]]
+
   """
   @doc since: "1.4.0"
-  @spec myers_difference(list, list) :: [{:eq | :ins | :del, list}]
-  def myers_difference(list1, list2) when is_list(list1) and is_list(list2) do
+  @spec myers_difference(list, list) :: [{:eq | :ins | :del | :diff, list}]
+  def myers_difference(list1, list2, diff_script \\ nil)
+      when is_list(list1) and is_list(list2) do
     path = {0, list1, list2, []}
-    find_script(0, length(list1) + length(list2), [path])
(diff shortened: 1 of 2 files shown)


lib/elixir/lib/list.ex
```diff
@@ -872,19 +874,49 @@ defmodule List do
...
+  such script. The returned inner edit will be under the `:diff` key.
+
+  ## Examples
+
+      iex> List.myers_difference(["a", "db", "c"], ["a", "bc"], &String.myers_difference/2)
+      [eq: ["a"], diff: [del: "d", eq: "b", ins: "c"], del: ["c"]]
+
+  """
+  @doc since: "1.8.0"
+  @spec myers_difference(list, list, diff_script) :: [{:eq | :ins | :del | :diff, list}]
```
fertapric: What do you think about being more specific with the results of the `diff_script` in the `@spec`?

```elixir
when diff_script: (term, term -> [{:eq | :ins | :del | :diff, term}] | nil)
```

josevalim: Good catch, thanks!

lib/elixir/lib/list.ex
```diff
@@ -872,19 +874,49 @@ defmodule List do
...
+  such script. The returned inner edit will be under the `:diff` key.
+
+  ## Examples
+
+      iex> List.myers_difference(["a", "db", "c"], ["a", "bc"], &String.myers_difference/2)
+      [eq: ["a"], diff: [del: "d", eq: "b", ins: "c"], del: ["c"]]
+
+  """
+  @doc since: "1.8.0"
+  @spec myers_difference(list, list, (term, term -> script)) :: script
```
fertapric: You removed the `nil` from the `diff_script` with the last commit. `(term, term -> script | nil)`

lib/elixir/lib/list.ex
```diff
@@ -872,19 +874,49 @@ defmodule List do
...
+    myers_difference_with_diff_script(list1, list2, nil)
+  end
+
+  @doc """
+  Returns a keyword list that represents an *edit script* with nested diffs.
+
+  This is an extension of `myers_difference/2` where a `diff_script` function
+  can be given in case it is desired to compute nested differences. The function
+  may return a list with the inner edit script or `nil` in case there is no
+  such script. The returned inner edit will be under the `:diff` key.
```
lexmag: inner edit **script** will

lib/elixir/lib/list.ex
```diff
@@ -897,50 +929,68 @@ defmodule List do
...
     {y, list1, rest, [{:ins, elem} | edits]}
   end
 
-  defp move_right({y, list1, [], edits}) do
+  defp move_right({y, list1, [], edits}, _diff_script) do
     {y, list1, [], edits}
   end
 
-  defp move_down({y, [elem | rest], list2, edits}) do
+  defp move_right({y, [elem1 | rest1], [elem2 | rest2] = list2, edits}, diff_script)
```
lexmag: This should be `move_down`.

lexmag: Also `move_right` (L#963) does `y + 1` in both branches, while `move_down` only with inner diff.

josevalim: > This should be move_down.

Thanks. Surprised the test suite didn't catch it. I guess it makes sense though since we always use the move_right branch until there are lefovers and the leftovers go to other branches.

> Also move_right (L#963) does y + 1 in both branches, while move_down only with inner diff.

`move_right` only moves y+1 when moving both lists.

PR #8022, opened 2018-07-31
Add String.bag_distance/2

josevalim:
This extracts a functionality for a function that was previously buried inside `ExUnit.Diff`.

### DIFF
--- a/lib/elixir/lib/string.ex
+++ b/lib/elixir/lib/string.ex
@@ -2252,11 +2252,82 @@ defmodule String do
   end
 
   @doc """
-  Returns a float value between 0 (equates to no similarity) and 1 (is an exact match)
-  representing [Jaro](https://en.wikipedia.org/wiki/Jaro–Winkler_distance)
+  Computes the bag distance of two strings.
+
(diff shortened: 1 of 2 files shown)


lib/elixir/lib/string.ex
```diff
@@ -2252,11 +2252,82 @@ defmodule String do
...
+      0.0
+      iex> String.bag_distance("abcd", "a")
+      0.25
+      iex> String.bag_distance("abcd", "ab")
+      0.5
+      iex> String.bag_distance("abcd", "abc")
+      0.75
+      iex> String.bag_distance("abcd", "abcd")
+      1.0
+
```
fertapric: What do you think about including a note mentioning this function uses graphemes to calculate the distance?

```elixir
iex> String.bag_distance("éṩab", "éṩa")
0.75
```

fertapric: Nevermind. I see almost all the functions in this module use graphemes (as mentioned in the documentation of the module), and their documentation do not include any references to graphemes.

josevalim:

PR #8025, opened 2018-08-01
insert_tail: shortcut common inputs to List.insert_at/3 for performance

fenollp:
The main idea is to skip length(list) computation when adding to the tail.
Two possible incompatibilities however:
* Added `is_list/1` guard so some inputs will now `FunctionClauseError`
* Not computing `length(list)` on index `-1` implies that this call
  no longer ensures input list is a proper list

Micro benchmark code & results can be seen at:
* code: https://github.com/fenollp/benchees/blob/a84b7db544f21a2359c3892e56d8cc36800e56b2/lib/insert_tail.exs
* results: https://circleci.com/gh/fenollp/benchees/3
Comparison:
append            65.75
insert_at         16.72 - 3.93x slower

### DIFF
--- a/lib/elixir/lib/list.ex
+++ b/lib/elixir/lib/list.ex
@@ -543,11 +543,20 @@ defmodule List do
 
   """
   @spec insert_at(list, integer, any) :: list
-  def insert_at(list, index, value) when is_integer(index) do
-    if index < 0 do
-      do_insert_at(list, length(list) + index + 1, value)
-    else
-      do_insert_at(list, index, value)
+  def insert_at(list, index, value)
+      when is_integer(index) and is_list(list) do
+    cond do
+      index == -1 ->
+        list ++ [value]
+
+      index == 0 ->
+        [value | list]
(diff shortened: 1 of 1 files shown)


michalmuskala:
Looking at the code, it seems to me we should be able to optimise the do_* functions in general by only pattern matching on 0 and removing the < 0 check. It should be safe provided we're careful on the calling code. This should give a nice perf boost in general.

fenollp:
Are you saying that, as an additional optimization, we could replace the `<= 0` of https://github.com/fenollp/elixir/blob/d5637102a818cc2bdba920bbef0276f24133b630/lib/elixir/lib/list.ex#L1044 with `== 0`?

michalmuskala:
Yes. I mean it's enough to have one check at the beginning and not on each iteration - if the initial calculation gives a positive number and then we just subtract by 1, we can never arrive below zero bypassing zero itself. This should save one check on every loop iteration.

PR #8026, opened 2018-08-01
Change error message for calling missing function in a guard

### LINKED ISSUE #8023: Bad error message if defguard is defined after function
### Environment

* Elixir & Erlang/OTP versions (elixir --version): 1.7.1 / OTP 21
* Operating system:  Arch Linux

### Current behavior

Using `defguard` to define a new a guard clause after its usage:

```elixir
defmodule Test do
  def test(x) when is_a_test(x) do
    {:ok}
  end

  defguard is_a_test(x) when x > 0
end
```
Returns the following error:
```
** (CompileError) test.exs:2: cannot invoke local is_a_test/1 inside guard, called as: is_a_test(x)
```
This is because `defguard` should be defined before its usage, i.e.:

```elixir
defmodule Test do
  defguard is_a_test(x) when x > 0

  def test(x) when is_a_test(x) do
    {:ok}
  end
end
```

The problem is, the error message is not clear. It gives the impression that Elixir found a `is_a_test` function (`defguard` is a macro) and is trying to call it.

### Expected behavior

A better error description instead of `cannot invoke local is_a_test/1` would make debugging this kind of issue easier, especially for beginners. Something like `cannot find is_a_test/1`.

xtian:
Fixes #8023 

Let me know if you have any suggestions for improving the wording here.

### DIFF
--- a/lib/elixir/src/elixir_expand.erl
+++ b/lib/elixir/src/elixir_expand.erl
@@ -1134,6 +1134,8 @@ format_error({invalid_call, Call}) ->
   io_lib:format("invalid call ~ts", ['Elixir.Macro':to_string(Call)]);
 format_error({invalid_quoted_expr, Expr}) ->
   io_lib:format("invalid quoted expression: ~ts", ['Elixir.Kernel':inspect(Expr, [])]);
+format_error({invalid_local_invocation, guard, {Name, _, Args} = Call}) ->
+  io_lib:format("cannot find local ~ts/~B", [Name, length(Args)]);
 format_error({invalid_local_invocation, Context, {Name, _, Args} = Call}) ->
   io_lib:format("cannot invoke local ~ts/~B inside ~ts, called as: ~ts",
                 [Name, length(Args), Context, 'Elixir.Macro':to_string(Call)]);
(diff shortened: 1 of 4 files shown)


lib/elixir/src/elixir_expand.erl
```diff
@@ -1134,6 +1134,8 @@ format_error({invalid_call, Call}) ->
   io_lib:format("invalid call ~ts", ['Elixir.Macro':to_string(Call)]);
 format_error({invalid_quoted_expr, Expr}) ->
   io_lib:format("invalid quoted expression: ~ts", ['Elixir.Kernel':inspect(Expr, [])]);
+format_error({invalid_local_invocation, guard, {Name, _, Args} = Call}) ->
+  io_lib:format("cannot find local ~ts/~B", [Name, length(Args)]);
 format_error({invalid_local_invocation, Context, {Name, _, Args} = Call}) ->
   io_lib:format("cannot invoke local ~ts/~B inside ~ts, called as: ~ts",
```
josevalim: I think it would be better to change this message, instead of special casing the guard one, because can have custom macros in both match and guard. Also note that the old message is still valid, so we need to make sure to keep it to. I propose something like:

    cannot find or cannot invoke local  ~ts/~B inside ~ts.  Only macros can be invoked in a ~ts and they must be defined before their invocation. Called as: ~ts

What do you think?

josevalim:

PR #8031, opened 2018-08-02
Add Keyword.fetch/3

Olshansk:
In more involved pipelined operations, matches or with
statements, it is useful to have Keyword.fetch have an
option to return a specific value as part of the error
tuple like `{:error, values}` instead of just `:error`.

### DIFF
--- a/lib/elixir/lib/keyword.ex
+++ b/lib/elixir/lib/keyword.ex
@@ -367,6 +367,30 @@ defmodule Keyword do
   end
 
   @doc """
+  Fetches the value for a specific `key` and returns it in a tuple.
+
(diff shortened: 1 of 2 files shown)


fertapric:
In general, I don't think the responses of a function should be configurable by the caller. That's why I wouldn't include such functionality in Elixir.

I would have a separate function that converts the error into something more semantic. For example:

```elixir
defp fetch_my_key(kw) do
  case Keyword.fetch(kw, :my_key) do
    {:ok, value} -> {:ok, value}
    :error -> {:error, :missing_my_key}
  end
end
```

josevalim:
Thanks @Olshansk! But this is something that can be trivially implemented as part of your application and that's what we would promote here.

PR #8037, opened 2018-08-03
Reconsolidate protocols if stale local deps change

### LINKED ISSUE #5987: Protocol implementations lost in umbrellas
### Environment

```
Erlang/OTP 19 [erts-8.2] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:10] [hipe] [kernel-poll:false]

Elixir 1.5.0-dev (1822cdbc8)
```
* Operating system: macOS

### Current behaviour

When working inside the umbrella and alternating between running mix tasks from the umbrella root and one of the applications, it seems that the compiler tends to "lose" some protocol implementations. I suspect this might be due to the fact that one application in the umbrella depends on a package that implements the protocol, while others don't. This most commonly happens with the `String.Chars` protocol for the `Postgrex.Query` type.

The issue is "fixed" by purging the `_build` folder and recompiling everything.

I tried creating a reproducible case multiple times but never managed to successfully create one - that's why I waited with reporting the issue, but maybe someone else will have better luck. 
I face the issue regularly - sometimes even couple times a day. I've also seen other people repeatedly report similar problems. The issue is definitely present in 1.4 and current master, I'm not sure about earlier versions.

### Expected behaviour

Protocols are consolidated correctly.

josevalim:
Prior to this commit, a protocol change in a path
dependency would only be reconcolidated if there at
least a runtime change associated to it. This commit
makes it so changes in local deps always triggers
a potential reconsolidation.

Closes #5987

### DIFF
--- a/lib/mix/lib/mix/compilers/elixir.ex
+++ b/lib/mix/lib/mix/compilers/elixir.ex
@@ -38,6 +38,7 @@ defmodule Mix.Compilers.Elixir do
 
     {all_modules, all_sources} = parse_manifest(manifest, dest)
     modified = Mix.Utils.last_modified(manifest)
+    stale_local_deps = stale_local_deps(manifest, modified)
     prev_paths = for source(source: source) <- all_sources, into: MapSet.new(), do: source
 
     removed =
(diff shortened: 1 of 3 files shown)


josevalim:
@michalmuskala this is not a regression though and I think it deserves a bit more testing before backporting to avoid yet a new release.

josevalim:

PR #8039, opened 2018-08-03
Show last n messages from mailbox on assert receive fail

Apelsinka223:
Show last message 10 from message mailbox instead of first 10 on tests fails on assert_receive, because usually errors are about the last messages


michalmuskala:
Thank you! :heart:

PR #8041, opened 2018-08-04
Fix binding for bind_quoted variables with same name

### LINKED ISSUE #7709: Nested quote with bind_quoted does not work as expected
### Environment

```
Erlang/OTP 20 [erts-9.3] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:10] [hipe] [kernel-poll:false]
Interactive Elixir (1.7.0-dev) - press Ctrl+C to exit (type h() ENTER for help)* 
```

Operating system: Linux/Ubuntu18

### Current behavior

Nested `quote(bind_quoted: ..., do: ...)` does not bind the quoted arguments.

```elixir
defmodule Foo do
  # works as expected
  defmacrop macro0(arg) do
    quote bind_quoted: [arg: arg] do
      {{:., [], [{:__aliases__, [alias: false], [:IO]}, :puts]}, [], [arg]}
    end 
  end
  # works as expected
  defmacrop macro1(arg) do
    quote bind_quoted: [arg: arg] do
      quote do: IO.puts(unquote arg)
    end
  end
  # raises ** (CompileError) foo.ex:37: undefined function arg/0
  defmacrop macro2(arg) do
    quote bind_quoted: [arg: arg] do
      # quote bind_quoted: [arg: arg], do: IO.puts(arg)
    end
  end
  
  defmacro __using__(_) do
    quote do
      def test0 do
        str = "world 0"
        unquote(macro0(quote do: str))
      end
      def test1 do
        str = "world 1"
        unquote(macro1(quote do: str))
      end
      def test2 do
        str = "world 2"
        unquote(macro2(quote do: str))
      end
    end
  end
end

defmodule Bar do
  use Foo 
end

Bar.test0()
Bar.test1()
Bar.test2()
```

### Expected behavior

The latter macro `macro2` should behave the same way as `macro1` since unless it’s nested, they behave the same way.

ggcampinho:
If the bind_quoted variable has the same name as a previous variable, its references need to be updated to the new variable.

Fixes #7709

### DIFF
--- a/lib/elixir/src/elixir_quote.erl
+++ b/lib/elixir/src/elixir_quote.erl
@@ -161,10 +161,18 @@ quote({unquote_splicing, _, [_]}, _Binding, #elixir_quote{unquote=true}, _) ->
     "wrap it in parens if you want it to work with one-liners">>);
 
 quote(Expr, nil, Q, E) ->
-  do_quote(Expr, Q, E);
+  do_quote(Expr, Q, E, #{});
 
 quote(Expr, Binding, Q, E) ->
   Context = Q#elixir_quote.context,
+
+  ReplaceBinding =
+    [
+      {{Var, meta(Meta, Q), VarContext}, {K, [], Context}}
+      || {K, {Var, Meta, VarContext}} <- Binding
+    ],
+
+  TExprs = do_quote(Expr, Q, E, maps:from_list(ReplaceBinding)),
 
   Vars = [ {'{}', [],
     [ '=', [], [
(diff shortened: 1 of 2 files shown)


ggcampinho:
Not sure if it's the best approach to fix this, but I think it I found the issue

josevalim:
Hi @ggcampinho! I have not looked into this issue so far but I think this is not the direction we want to go, because I believe this is an issue with variable hygiene.

I have simplified the case to this:

```elixir
defmodule Foo do
  defmacrop macro2(arg) do
    quote bind_quoted: [arg: arg] do
      quote bind_quoted: [arg: arg], do: IO.puts(arg)
    end
  end
  
  defmacro __using__(_) do
    quote do
      def test2 do
        str = "world 2"
        unquote(macro2("foo"))
      end
    end
  end
end

defmodule Bar do
  use Foo 
end

Bar.test2()
```

I have added a puts statement to see the AST of test2, and we get this:

```elixir
{:test2, 0,
 [
   {[line: 40, context: Foo], [], [],
    {:__block__, [line: 40],
     [
       {:=, [line: 40], [{:str, [line: 40, counter: 2], Foo}, "world 2"]},
       {:__block__, [line: 40],
        [
          {:=, [line: 40], [{:arg, [line: 40, counter: 2], Foo}, "foo"]},
          {{:., [line: 40], [IO, :puts]}, [line: 40],
           [{:arg, [line: 40, counter: 4], []}]}
        ]}
     ]}}
 ]}
```

You can see that `str` and the first `arg` has counter of 2. This counter is identifying the quoted scope from `__using__`. However, you can see that the inner `arg` is having a counter of 4, which is coming from `macro2`. However, because `arg` is coming from a quote inside a quote, we want to discard its counter. This is inline with the `elixir_quote:annotate/2`, which does discards the counter for aliases and imports, but we forgot it for variables.

I have a fix that I will push soon. I know the best would have been for me to give directions so you can fix this but this is one of the hard issues where the solution only becomes obvious after you see it. In any case, I am glad for got started on this and pushed us towards the proper path. Thank you! :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

josevalim:
Closed in favor of #8042.

PR #6205, opened 2017-06-08
Clean up the Map module by checking for maps and reusing code

whatyouhide:

PR #8042, opened 2018-08-04
Delete counter for vars in nested quotes

### LINKED ISSUE #7709: Nested quote with bind_quoted does not work as expected

josevalim:
Consider the following example:

    defmodule NestedQuote do
      defmacrop macro(arg) do
        quote bind_quoted: [arg: arg] do
          quote bind_quoted: [arg: arg], do: String.duplicate(arg, 2)
        end
      end

      defmacro __using__(_) do
        quote do
          def test do
            unquote(macro("foo"))
          end
        end
      end
    end

    defmodule UseNestedQuote do
      use NestedQuote
    end

It has two macro expansions, one inside __using__/1 when it invokes
macro/1, and another when used. The issue, however, is that the
`arg` of `String.duplicate(arg, 2)` gets tagged with the context
of `macro/1` but it should be tagged with the context of `__using__`
since it is a quote inside a quote. This patch addresses this.

Closes #7709

### DIFF
--- a/lib/elixir/src/elixir_rewrite.erl
+++ b/lib/elixir/src/elixir_rewrite.erl
@@ -207,6 +207,10 @@ rewrite(?map, fetch, [Map, Key]) ->
   {maps, find, [Key, Map]};
 rewrite(?map, 'fetch!', [Map, Key]) ->
   {maps, get, [Key, Map]};
+rewrite(?map, 'get', [Map, Key]) ->
+  {maps, get, [Key, Map, nil]};
+rewrite(?map, 'get', [Map, Key, Default]) ->
+  {maps, get, [Key, Map, Default]};
 rewrite(?map, 'has_key?', [Map, Key]) ->
   {maps, is_key, [Key, Map]};
 rewrite(?map, put, [Map, Key, Value]) ->
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -3518,7 +3518,7 @@ defmodule Kernel do
   end
 
   defp ensure_evaled_var(elem, {index, ast}) do
-    var = {String.to_atom("var" <> Integer.to_string(index)), [], __MODULE__}
+    var = {String.to_atom("arg" <> Integer.to_string(index)), [], __MODULE__}
     {var, {index + 1, [{var, elem} | ast]}}
   end
 
--- a/lib/elixir/lib/kernel/utils.ex
+++ b/lib/elixir/lib/kernel/utils.ex
@@ -196,8 +196,8 @@ defmodule Kernel.Utils do
 
     quote do
       case Macro.Env.in_guard?(__CALLER__) do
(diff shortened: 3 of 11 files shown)


lib/elixir/src/elixir_rewrite.erl
```diff
@@ -207,6 +207,10 @@ rewrite(?map, fetch, [Map, Key]) ->
   {maps, find, [Key, Map]};
 rewrite(?map, 'fetch!', [Map, Key]) ->
   {maps, get, [Key, Map]};
+rewrite(?map, 'get', [Map, Key]) ->
```
fertapric: Should the implementation of `Maps.get/3` use `:maps.get(key, map, default)` then? 

I was looking at past PRs, and [this was not inlined](https://github.com/elixir-lang/elixir/pull/6205#discussion_r120950701) in the past because `get/3` was not a BIF. I checked Erlang's source code, and it seems [it is not included in the `%% BIFs` section](https://github.com/erlang/otp/blob/master/lib/stdlib/src/maps.erl#L23).

I've also made some benchmarks, and apparently Elixir's implementation is faster than Erlang's.

josevalim: Good catch. I will revert this. The only reason I added it here is because of bootstrap issues but performance should be the main criteria.

OvermindDL1:
Ah, this explains some oddities I've seen when I work with macro's!

Looking forward to the release then!

PR #8046, opened 2018-08-04
Move parentheses in System.cmd() docs

scott-ad-riley:
This is fixed upstream - but reading through the docs for 1.7.1 it caught me out.


josevalim:

PR #8047, opened 2018-08-04
Inspect.Algebra typespecs improvement

amatalai:


sasa1977:
@josevalim could we port this into 1.7 branch? I think that this would fix a dialyzer error on `Code.format_string`.

josevalim:
@sasa1977 done!

PR #8048, opened 2018-08-05
Minor improvements in match operators code

fertapric:
Minor things I spotted while reading https://github.com/elixir-lang/elixir/pull/8043 :)

cc @michalmuskala

### DIFF
--- a/lib/elixir/test/elixir/kernel/expansion_test.exs
+++ b/lib/elixir/test/elixir/kernel/expansion_test.exs
@@ -552,16 +552,16 @@ defmodule Kernel.ExpansionTest do
 
     test "in matches" do
       message =
-        ~r"cannot invoke remote function Hello.something_that_does_not_exist/0 inside pattern"
+        ~r"cannot invoke remote function Hello.fun_that_does_not_exist/0 inside pattern matching"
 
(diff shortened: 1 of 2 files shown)


lib/elixir/test/elixir/kernel/expansion_test.exs
```diff
@@ -552,16 +552,16 @@ defmodule Kernel.ExpansionTest do
 
     test "in matches" do
       message =
-        ~r"cannot invoke remote function Hello.something_that_does_not_exist/0 inside pattern"
+        ~r"cannot invoke remote function Hello.fun_that_does_not_exist/0 inside pattern matching"
```
fertapric: I haven't modified this message, but I think this could be "inside a match" instead of "inside a pattern matching" to use the same terms when talking about match/pattern matching. What do you think?

fertapric: Done in https://github.com/elixir-lang/elixir/pull/8048/commits/e1e4b2dde5e3a4aa1093aaf84af344c45968c208

michalmuskala:
Thank you! :heart:

PR #8055, opened 2018-08-05
Remove error message from IEx test suite

fertapric:
Due to race conditions, the capturing logic introduced by `ExUnit.CaptureLog.capture_log/2` might not be there when the error messages logged by the Erlang's code server arrived. If that happens, this error message is printed:

screen shot 2018-08-05 at 11 18 47 pm


josevalim:

eksperimental:
thank you @fertapric

PR #8060, opened 2018-08-07
Add documentation example for a common use of Task.Supervised.async_n…

jeremyowensboggs:
Quick example of how to use Task.Supervised.async_nolink with a GenServer.

### DIFF
--- a/lib/elixir/lib/task/supervisor.ex
+++ b/lib/elixir/lib/task/supervisor.ex
@@ -159,27 +159,39 @@ defmodule Task.Supervisor do
 
   ## Examples
 
-  With a GenServer
+  Typically, you use async_nolink when there is a reasonable expectation that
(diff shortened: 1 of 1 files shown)


josevalim:
Thanks @jeremyowensboggs . Can you please provide a bit more context? When would you want to do this? What are the benefits?

Also, regarding the code, how do you know the reference you are getting is really from a task you started?

jeremyowensboggs:
Is that any better?  I'm trying to keep the example code simple.   Something like this was presented by one of the members of our KC meetup, and I thought it was 1) really neat, 2) perfect for what I was working on that day, and 3) Really hard to find another example of it online.  This is mainly an attempt to remedy the lack of an example.  I had read through the task documentation several times, but never clicked that tasks could be used this way.

josevalim:

PR #8062, opened 2018-08-07
Make sure records use the dsetelement optimisation

### LINKED ISSUE #8059: Make sure our records implementation triggers dsetel optimization
This means that we need to allocate arguments to vars during expansion and start by the higher position first. We should also apply the same logic as found in `erl_expand_records` and convert some updates to matches if most of the data is being updated.

michalmuskala:
Closes #8059

Additionally some more refactorings of the Record module

---

This needs some more tests before merging

### DIFF
--- a/lib/elixir/lib/record.ex
+++ b/lib/elixir/lib/record.ex
@@ -379,38 +379,30 @@ defmodule Record do
 
   # Gets the index of field.
   defp index(tag, fields, field) do
-    if index = find_index(fields, field, 0) do
-      # Convert to Elixir index
-      index - 1
-    else
+    find_index(fields, field, 1) ||
       raise ArgumentError, "record #{inspect(tag)} does not have the key: #{inspect(field)}"
-    end
   end
 
   # Creates a new record with the given default fields and keyword values.
   defp create(tag, fields, keyword, caller) do
-    in_match = Macro.Env.in_match?(caller)
-    keyword = apply_underscore(fields, keyword)
+    # Using {} here is safe, since it's not valid AST
+    default = if Macro.Env.in_match?(caller), do: {:_, [], nil}, else: {}
+    {default, keyword} = Keyword.pop(keyword, :_, default)
 
-    {match, remaining} =
-      Enum.map_reduce(fields, keyword, fn {field, default}, each_keyword ->
-        new_fields =
-          case Keyword.fetch(each_keyword, field) do
-            {:ok, value} -> value
-            :error when in_match -> {:_, [], nil}
-            :error -> Macro.escape(default)
-          end
-
-        {new_fields, Keyword.delete(each_keyword, field)}
+    {elements, remaining} =
+      Enum.map_reduce(fields, keyword, fn {key, field_default}, remaining ->
+        case Keyword.pop(remaining, key, default) do
+          {{}, remaining} -> {Macro.escape(field_default), remaining}
+          {default, remaining} -> {default, remaining}
+        end
       end)
 
     case remaining do
       [] ->
-        {:{}, [], [tag | match]}
+        quote(do: {unquote(tag), unquote_splicing(elements)})
 
-      _ ->
-        keys = for {key, _} <- remaining, do: key
-        raise ArgumentError, "record #{inspect(tag)} does not have the key: #{inspect(hd(keys))}"
+      [{key, _} | _] ->
+        raise ArgumentError, "record #{inspect(tag)} does not have the key: #{inspect(key)}"
     end
   end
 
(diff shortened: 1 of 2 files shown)


lib/elixir/lib/record.ex
```diff
@@ -397,35 +389,85 @@ defmodule Record do
       raise ArgumentError, "cannot invoke update style macro inside match"
     end
 
-    keyword = apply_underscore(fields, keyword)
+    case build_updates(keyword, fields, [], [], []) do
```
josevalim: @michalmuskala erl_expand_record has a small heuristic in that, if there are more fields updating than fields being kept, then it does not use setelem at all, and instead allocates a new tuple: https://github.com/erlang/otp/blob/83bf5bea4b872be40253a8f25c9aec3b79d6d66f/lib/stdlib/src/erl_expand_records.erl#L673-L702

Should we have the same?

lib/elixir/lib/record.ex
```diff
@@ -397,35 +389,85 @@ defmodule Record do
...
-  defp get(tag, fields, var, key) do
-    index = find_index(fields, key, 0)
+  defp build_updates([{:_, value} | rest], fields, updates, vars, exprs) do
+    {updates, vars, exprs} = build_updates(rest, fields, updates, vars, exprs)
 
-    if index do
-      quote do
-        :erlang.element(unquote(index), unquote(var))
+    if simple_argument?(value) do
+      updates = add_defaults(fields, updates, value, 2)
```
ggcampinho: If you add this test to the suit:
```
record = user(user(), _: :_, name: "meg", age: 12)
assert user(record, :name) == "meg"
assert user(record, :age) == 12
```

You will have the following error:
```
** (FunctionClauseError) no function clause matching in Record.add_defaults/4    
    
    The following arguments were given to Record.add_defaults/4:
    
        # 1
        []
    
        # 2
        [{2, "meg"}]
    
        # 3
        :_
    
        # 4
        4
    
    Attempted function clauses (showing 3 out of 3):
    
        defp add_defaults([_ | fields], [{i, _} = update | updates], value, i)
        defp add_defaults([_ | fields], updates, value, i)
        defp add_defaults([], [], _, _)
    
    (elixir) lib/record.ex:444: Record.add_defaults/4
    (elixir) lib/record.ex:445: Record.add_defaults/4
    (elixir) lib/record.ex:449: Record.add_defaults/4
    (elixir) lib/record.ex:420: Record.build_updates/5
    (elixir) lib/record.ex:392: Record.update/5
    expanding macro: RecordTest.user/2
    lib/elixir/test/elixir/record_test.exs:134: RecordTest."test records with default values"/1
```

Also, when the `:_` comes at the end of the keyword list or when you have repeated keys, it will execute some useless code. I believe it doesn't belong to this PR, but it would be nice if repeated keys we had some kind of warning.

michalmuskala:
@josevalim I think we should actually benchmark this. The erlang implementation was done some time ago and the VM was refactored couple times since.

josevalim:
> @josevalim I think we should actually benchmark this. The erlang implementation was done some time ago and the VM was refactored couple times since.

Excellent points.Agreed!

michalmuskala:
@josevalim I took one more look at this and I'm not sure if we can do the same optimisation the erlang code is doing without changing the semantics.

The erlang code always verifies the record type on update, where we don't. Our current update is just a sequence of `setelement` calls where they might fail if a wrong tuple is passed, but they might also succeed by accident.
If we switch to the erlang-style code that explicitly unpacks the whole tuple and repacks it back, we need to verify the whole record and not just the fields we're updating. This might make the code fail where it didn't fail before.

OvermindDL1:
> This might make the code fail where it didn't fail before.

Honestly though, if it was failing after such a verification then that was extremely likely a bug to begin with.

josevalim:
@michalmuskala good points, let's keep the unpacking/packing optimization in particular then. We can still apply the other ones.

michalmuskala:
I just realised - updating the record combined with `_: :_` doesn't make much sense - it's basically always creating a new record since all the fields are going to be changed

michalmuskala:
I think this is ready for review

lib/elixir/lib/record.ex
```diff
@@ -420,35 +412,64 @@ defmodule Record do
       raise ArgumentError, "cannot invoke update style macro inside match"
     end
 
-    keyword = apply_underscore(fields, keyword)
+    if Keyword.has_key?(keyword, :_) do
+      IO.warn("updating a record with a default (:_) is equivalent to creating a new record")
```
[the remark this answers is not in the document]
michalmuskala: I knew I saw somewhere a way to build a stacktrace for the compiler, but I just couldn't find it. Makes sense :+1:

lib/elixir/test/elixir/record_test.exs
```diff
@@ -127,9 +129,16 @@ defmodule RecordTest do
     assert match?(user(_: _), user())
     refute match?(user(_: "other"), user())
 
-    record = user(user(), _: :_, name: "meg")
-    assert user(record, :name) == "meg"
-    assert user(record, :age) == :_
+    captured =
```
josevalim: Let's move this to a separate test maybe?

michalmuskala:
Updated

PR #8069, opened 2018-08-10
Small optimization on expand_args/2

ggcampinho:
`elixir_env:mergea(EV, EC)` is always equivalent to `EV` so we can avoid the operation


josevalim:

josevalim:
Great catch!

PR #8074, opened 2018-08-11
Small optimization on elixir_bitstring:expand/4

ggcampinho:
`elixir_env:mergea(EV, EC)` is always equivalent to `EV` so we can avoid the operation.
Similar to #8069


josevalim:

PR #8077, opened 2018-08-13
Extend System.get_env/1 to support a default value

fenollp:
It is common to read an environment variable then to choose
whether to return a default value. For instance in config files
one can call `System.get_env("MY_VAR")` to fill some field, but
if one wants to have a way to handle defaults from within the config
one has to repeat a case block, or define a module that exports the
desired wrapped around `System.get_env/1`.


fenollp:
Sorry I missed that! I quite like the `a() || b()` notation as it's lazy on `b` :)

josevalim:
Closing for the reasons mentioned. :)

PR #8079, opened 2018-08-14
Fix arity of try/1 in :elixir_import.special_form

### LINKED ISSUE #8078: Strange try/catch behaviour
### Environment
Erlang/OTP 21 [erts-10.0] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1]
Elixir 1.7.0 (compiled with Erlang/OTP 20)

Linux version 4.15.0-30-generic (buildd@lgw01-amd64-060) (gcc version 7.3.0 (Ubuntu 7.3.0-16ubuntu3))
### Current behavior
After switching to elixir 1.7 some some test in our ExUnit cases no longer compiled.
An easy reproduction is:
```elixir
ExUnit.start()

defmodule AssertionTest do
  use ExUnit.Case, async: true

  test "the truth" do
    try do
      raise "error"
      false
    rescue
      _ -> true
    end
    |> assert
  end
end
```
i yields an `** (CompileError) ... : unhandled operator ->` error.

Strange thing is inserting an `IO.inspect` in between fixes it.
```elixir
ExUnit.start()

defmodule AssertionTest do
  use ExUnit.Case, async: true

  test "the truth" do
    try do
      raise "error"
      false
    rescue
      _ -> true
    end
    |> IO.inspect()
    |> assert
  end
end
```

### Expected behavior
After consulting the ExUnit docs I`m aware that there are assertions especially for cases in which an Exception is expected.

Still the code example resulting in an error should work since its syntactically correct, especially considering that the `IO.inspect` should not make any difference.

fertapric:
Fixes #8078

### DIFF
--- a/lib/elixir/src/elixir_import.erl
+++ b/lib/elixir/src/elixir_import.erl
@@ -239,6 +239,6 @@ special_form('for', _) -> true;
 special_form('with', _) -> true;
 special_form('cond', 1) -> true;
 special_form('case', 2) -> true;
-special_form('try', 2) -> true;
+special_form('try', 1) -> true;
 special_form('receive', 1) -> true;
 special_form(_, _) -> false.


josevalim:

PR #8080, opened 2018-08-14
Remove impossible clause on DateTime.from_naive!/2

fertapric:
Currently `DateTime.from_naive/2` always returns `{:ok, datetime}`.

Other option is to update `DateTime.from_naive/2` to include a fallback clause that returns errors. For example, `{:error, :unsupported_time_zone}` when the `time_zone` attribute is not "Etc/UTC".


josevalim:
Good catch! I believe this will no longer be true once #7914 will be merged, so I think we should just wait until it is merged. Thanks!

PR #8087, opened 2018-08-15
First stab at implementation of `after_suite/1` callback

### LINKED ISSUE #8082: Add ExUnit.after_suite
See https://github.com/elixir-ecto/ecto/pull/2620 for discussion.

devonestes:
This is a first take at the `after_suite/1` callback in ExUnit. I've
done basic testing by making this work with an app on my local machine,
but I'm going to think about automated tests tomorrow. I also still
need to document this better, but I wanted to get this up just to make
sure this implementation is along the lines of what y'all are looking
for.

Eventually resolves #8082

### DIFF
--- a/lib/ex_unit/lib/ex_unit.ex
+++ b/lib/ex_unit/lib/ex_unit.ex
@@ -319,7 +319,8 @@ defmodule ExUnit do
   map containing the results of the test suite's execution.
 
   If `after_suite/1` is called multiple times, the callbacks will be called in
-  reverse order - i.e. the last callback set will be the first to be called.
+  reverse order. In other words, the last callback set will be the first to be
+  called.
   """
   @spec after_suite(
           (%{
--- a/lib/ex_unit/lib/ex_unit.ex
+++ b/lib/ex_unit/lib/ex_unit.ex
@@ -313,16 +313,26 @@ defmodule ExUnit do
   end
 
   @doc """
-  Allows users to configure functions to run after the test suite has finished.
-  Functions passed in will be executed after all tests have finished running. If
-  `after_suite/1` is called multiple times, the functions will be executed in
-  reverse order (i.e. the last function given will be executed first).
+  Sets a callback to be executed after the completion of a test suite.
+
+  Callbacks set with `after_suite/1` must accept a single argument, which is a
+  map containing the results of the test suite's execution.
+
+  If `after_suite/1` is called multiple times, the callbacks will be called in
+  reverse order - i.e. the last callback set will be the first to be called.
   """
-  @spec after_suite(fun) :: :ok
-  def after_suite(function) do
+  @spec after_suite(
+          (%{
+             excluded: non_neg_integer,
+             failures: non_neg_integer,
+             skipped: non_neg_integer,
+             total: non_neg_integer
+           } ->
+             any)
+        ) :: :ok
+  def after_suite(function) when is_function(function) do
     current_callbacks = Application.fetch_env!(:ex_unit, :after_suite)
     configure(after_suite: [function | current_callbacks])
-    :ok
   end
 
   # Persists default values in application
--- a/lib/ex_unit/lib/ex_unit.ex
+++ b/lib/ex_unit/lib/ex_unit.ex
@@ -312,6 +312,19 @@ defmodule ExUnit do
     ExUnit.Runner.run(config, nil)
   end
 
+  @doc """
+  Allows users to configure functions to run after the test suite has finished.
+  Functions passed in will be executed after all tests have finished running. If
+  `after_suite/1` is called multiple times, the functions will be executed in
+  reverse order (i.e. the last function given will be executed first).
+  """
+  @spec after_suite(fun) :: :ok
+  def after_suite(function) do
+    current_callbacks = Application.fetch_env!(:ex_unit, :after_suite)
+    configure(after_suite: [function | current_callbacks])
+    :ok
+  end
+
   # Persists default values in application
   # environment before the test suite starts.
   defp persist_defaults(config) do
--- a/lib/ex_unit/lib/ex_unit/runner.ex
+++ b/lib/ex_unit/lib/ex_unit/runner.ex
@@ -21,6 +21,8 @@ defmodule ExUnit.Runner do
     EM.suite_finished(config.manager, run_us, load_us)
     result = ExUnit.RunnerStats.stats(stats)
     EM.stop(config.manager)
+    after_suite_callbacks = Application.fetch_env!(:ex_unit, :after_suite)
+    Enum.each(after_suite_callbacks, fn callback -> callback.(result) end)
     result
   end
 
--- a/lib/ex_unit/test/ex_unit_test.exs
+++ b/lib/ex_unit/test/ex_unit_test.exs
@@ -498,6 +498,40 @@ defmodule ExUnitTest do
     ExUnit.configure(seed: global_seed)
   end
 
+  describe "after_suite/1" do
+    test "executes all callbacks set in reverse order" do
+      Process.register(self(), :after_suite_test_process)
+
+      defmodule MultipleAfterSuiteTest do
+        use ExUnit.Case
+
+        test "true" do
+          send(:after_suite_test_process, :in_first_test)
+        end
+      end
+
+      ExUnit.Server.modules_loaded()
+
+      ExUnit.after_suite(fn _ -> send(:after_suite_test_process, :first_after_suite) end)
+      ExUnit.after_suite(fn result -> send(:after_suite_test_process, result) end)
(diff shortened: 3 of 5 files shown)


lib/ex_unit/lib/ex_unit.ex
```diff
@@ -182,6 +182,8 @@ defmodule ExUnit do
           time = ExUnit.Server.modules_loaded()
           config = persist_defaults(configuration())
           %{failures: failures} = ExUnit.Runner.run(config, time)
+          Application.fetch_env!(:ex_unit, :after_suite)
+          |> Enum.each(fn fun -> fun.() end)
```
josevalim: Now that I think about it, maybe we should pass the result of `run` as argument to the function.

lib/ex_unit/lib/ex_unit.ex
```diff
@@ -182,6 +182,8 @@ defmodule ExUnit do
           time = ExUnit.Server.modules_loaded()
           config = persist_defaults(configuration())
           %{failures: failures} = ExUnit.Runner.run(config, time)
```
josevalim: There is also a `run` function in this module and I think we should call the after_run in there too. So maybe we should unify it in a private function?

lib/mix/lib/mix/tasks/test.ex
```diff
@@ -384,6 +384,9 @@ defmodule Mix.Tasks.Test do
       {:ok, %{excluded: excluded, failures: failures, total: total}} ->
         cover && cover.()
 
+        Application.fetch_env!(:ex_unit, :after_suite)
+        |> Enum.each(fn fun -> fun.() end)
```
josevalim: If we do the changes I mentioned above, we won't need this change, as we will run this always after run.

lib/ex_unit/lib/ex_unit.ex
```diff
@@ -312,6 +314,19 @@ defmodule ExUnit do
...
   end
 
+  @doc """
+  Allows users to configure functions to run after the test suite has finished.
+  Functions passed in will be executed after all tests have finished running. If
+  `after_suite/1` is called multiple times, the functions will be executed in
+  reverse order (i.e. the last function given will be executed first).
+  """
+  @spec after_suite(fun) :: :ok
+  def after_suite(function) do
```
fertapric: Maybe use a guard here `is_function(function, 0)`?

josevalim: We generally prefer `is_function(function)` as it leads to better error messages when invoking with the wrong arity later on.

lib/ex_unit/lib/ex_unit.ex
```diff
@@ -312,6 +314,19 @@ defmodule ExUnit do
     ExUnit.Runner.run(config, nil)
   end
 
+  @doc """
+  Allows users to configure functions to run after the test suite has finished.
```
josevalim: The first paragraph of the docs should always be a short summary. :) So I recommend adding a newline after this.

I also think that the "Allows users" doesn't add anything, as we could prefix most function documentations with "Allows users". :)

lib/ex_unit/lib/ex_unit.ex
devonestes: Are you thinking that there could be conditional code in these callbacks that would, for example, only be executed if there are failures? I can see people that would want to do such things.

lib/ex_unit/lib/ex_unit.ex
devonestes: Yeah, I have a lot of work on the documentation of this feature for sure.

lib/ex_unit/lib/ex_unit.ex
devonestes: Funny enough I was thinking the same thing last night as I was lying in bed! I'll definitely add that.

lib/ex_unit/lib/ex_unit.ex
josevalim: Yeah, I can't think of any use cases right now, but it doesn't seem to be a weird thing to want.

lib/ex_unit/lib/ex_unit.ex
```diff
@@ -312,6 +312,19 @@ defmodule ExUnit do
...
+  Functions passed in will be executed after all tests have finished running. If
+  `after_suite/1` is called multiple times, the functions will be executed in
+  reverse order (i.e. the last function given will be executed first).
+  """
+  @spec after_suite(fun) :: :ok
+  def after_suite(function) do
+    current_callbacks = Application.fetch_env!(:ex_unit, :after_suite)
+    configure(after_suite: [function | current_callbacks])
```
fertapric: I think `configure` returns `:ok` (it uses `Enum.each/2` which always returns `:ok`). 

Probably out of the scope of this PR, but I would include the spec `@spec configure(Keyword.t()) :: :ok` in the `configure/1` function :)

devonestes: Yeah, I saw a couple of functions here missing typespecs, so I was going to go and add them in a separate PR.

lib/ex_unit/lib/ex_unit.ex
```diff
@@ -312,6 +312,30 @@ defmodule ExUnit do
...
+             excluded: non_neg_integer,
+             failures: non_neg_integer,
+             skipped: non_neg_integer,
+             total: non_neg_integer
+           } ->
+             any)
+        ) :: :ok
+  def after_suite(function) when is_function(function) do
+    current_callbacks = Application.fetch_env!(:ex_unit, :after_suite)
+    configure(after_suite: [function | current_callbacks])
```
josevalim: I think in this case we can set `Application.put_env` directly, can't we? Is there a reason to go through configure?

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -21,6 +21,8 @@ defmodule ExUnit.Runner do
     EM.suite_finished(config.manager, run_us, load_us)
     result = ExUnit.RunnerStats.stats(stats)
     EM.stop(config.manager)
+    after_suite_callbacks = Application.fetch_env!(:ex_unit, :after_suite)
+    Enum.each(after_suite_callbacks, fn callback -> callback.(result) end)
```
josevalim: NIce, i like moving this here! :+1:

lib/ex_unit/test/ex_unit_test.exs
```diff
@@ -510,4 +544,12 @@ defmodule ExUnitTest do
     output = capture_io(fn -> Process.put(:capture_result, ExUnit.Runner.run(opts, nil)) end)
     {Process.get(:capture_result), output}
   end
+
+  defp next_message_in_mailbox() do
+    receive do
+      msg -> msg
+    after
+      1 -> nil
```
josevalim: This should be zero. :)

josevalim:
@devonestes two last comments and we are good to go!

lib/ex_unit/lib/ex_unit.ex
devonestes: I just went through configure because it seemed like that was the way this sort of configurations were being done elsewhere in this module (like [here](https://github.com/elixir-lang/elixir/blob/f0392717ade146680140022fc833c586de6bee9c/lib/ex_unit/lib/ex_unit.ex#L300)). We can totally use `Application.put_env` directly here, though.

josevalim: I see. Nevermind. Let's keep this as is then. :)

lib/ex_unit/test/ex_unit_test.exs
devonestes: I must have missed that line in the documentation 😝

josevalim:
Btw, you don't need to squash and rebase your commits in general. We can do that when merging. :)

lib/ex_unit/test/ex_unit_test.exs
```diff
@@ -498,6 +498,40 @@ defmodule ExUnitTest do
...
+      # test callbacks here, otherwise it will attempt to execute them after
+      # every subsequent call to `ExUnit.run()` in any tests run after these.
+      Application.put_env(:ex_unit, :after_suite, [])
+
+      assert next_message_in_mailbox() == :in_first_test
+      assert next_message_in_mailbox() == :third_after_suite
+      assert next_message_in_mailbox() == %{excluded: 0, failures: 0, skipped: 0, total: 1}
+      assert next_message_in_mailbox() == :first_after_suite
+      # Check to make sure the mailbox is empty after these four messages
+      refute_receive _
```
fertapric: Should this be `refute_received` to avoid the timeout?

josevalim:

lib/ex_unit/test/ex_unit_test.exs
josevalim: @fertapric yes, it should. Can you please send a super tiny PR? :)

PR #8089, opened 2018-08-16
Add missing typespecs to public functions in `ExUnit`

devonestes:


lib/ex_unit/lib/ex_unit.ex
```diff
@@ -305,8 +308,14 @@ defmodule ExUnit do
   if ExUnit is started via `start/1`.
 
   Returns a map containing the total number of tests, the number
-  of failures and the number of skipped tests.
+  of failures, the number of excluded tests and the number of skipped tests.
   """
+  @spec run() :: %{
```
fertapric: Looking at the `ExUnit.after_suite/1` PR, I think we could move the "result" map into a new type :) That way, it can be reused in `ExUnit.run/0`, `ExUnit.after_suite/1` and in the `ExUnit.Runner`.

devonestes: My original thinking was that `ExUnit.Runner` is private, and so I'm hesitant about exposing any types defined in that module. We could, however, define a `ExUnit.suite_result()` type, and use that in both `ExUnit.run/0` and `ExUnit.after_suite/1`. What do you think about that?

josevalim: Good catch. Let's hold this one until after_suite is merged, then we can act accordingly.

josevalim: :+1: for `ExUnit.suite_result()` too.

PR #8090, opened 2018-08-16
Add missing typespecs to public functions in `System`

devonestes:


lib/elixir/lib/system.ex
```diff
@@ -234,6 +237,7 @@ defmodule System do
 
   Returns the current working directory or raises `RuntimeError`.
   """
+  @spec cwd!() :: String.t() | no_return
```
josevalim: We don't need `no_return`. `no_return` is always implicit. We only use `no_return` when the only option is `no_return`. One way to look at it is to consider `no_return` as an equivalent to `none`. So `something | none` will always be `something`. :)

devonestes: For reference for these `!` functions I looked at how the types were defined in `File` for the `!` functions there, and they all use `:ok | no_return` (like [here](https://github.com/elixir-lang/elixir/blob/1cce731a363a92c494871af1823f935e48a790b5/lib/elixir/lib/file.ex#L236)). To be consistent, should we remove `no_return` in types with that signature and only use `no_return` in places where it's the only possible outcome?

fertapric: I would use `no_return` only in the places where it's the only possible outcome.

One could argue that `something | no_return` could specify that an error/exception is expected. However, I think the documentation and the naming convention of using a `!` should be enough :)

josevalim: @devonestes yes, those uses of `| no_return` is wrong and we should/could remove them.

lib/elixir/lib/system.ex
```diff
@@ -331,6 +339,7 @@ defmodule System do
 
   The function must receive the exit status code as an argument.
   """
+  @spec at_exit((non_neg_integer -> any)) :: list(fun)
```
josevalim: I think we should return `:ok` here. The fact it returns `list(fun)` is an accident.

lib/ex_unit/lib/ex_unit.ex
```diff
@@ -177,7 +177,7 @@ defmodule ExUnit do
   If you want to run tests manually, you can set the `:autorun` option
   to `false` and use `run/0` to run tests.
   """
-  @spec start(Keyword.t()) :: nil | list(fun)
+  @spec start(Keyword.t()) :: nil | :ok
```
whatyouhide: Do we need to know if this returns `nil` or `:ok`, or can we just force the return of `:ok`? If we need the two return values I think we should document what they mean.

devonestes: We can definitely force the return of `:ok` here.

lib/elixir/lib/system.ex
```diff
@@ -331,8 +339,10 @@ defmodule System do
 
   The function must receive the exit status code as an argument.
   """
+  @spec at_exit((non_neg_integer -> any)) :: :ok
   def at_exit(fun) when is_function(fun, 1) do
```
fertapric: Is this a breaking change? I'm not sure if this is considered a bug.

The same applies to `ExUnit.start/1`.

devonestes: Well, technically it is a breaking change, but given that previously both had undocumented return types, it might not _really_ be considered a breaking change. If they‘re undocumented, then what gets returned is just sort of a happy accident in the eyes of many folks.

josevalim:

PR #8092, opened 2018-08-16
Remove unnecessary `no_return` types in specs

devonestes:


josevalim:
Any function can fail any time in face of invalid input. So I wouldn't add those docs with the exception of Task.await, to make it clear it exits with timeout once timeout is reached (which is mostly assumed behaviour based on experience with other parts of the language so it doesn't hurt to make it clearer).

whatyouhide:
@fertapric let's do the changes to `Task.await/2` in another PR so that this one is clean and focused :).

Thanks @devonestes 💟

PR #8093, opened 2018-08-16
Remove check for legacy @export_type attribute

fertapric:
Erlang uses this attribute to export types, but Elixir abstracts it via `@type`, `@typep` and `@opaque`.

Besides, adding `@export_type` to a module does not have any effect; and it's not documented.


josevalim:

PR #8097, opened 2018-08-16
Allow deprecation of structures

hauleth:
This adds support for deprecation of structures by keeping `@deprecated`
attribute for both defined function.  This will make `%Foo{foo: 1}` to
raise deprecation warnings when used.


josevalim:
@hauleth good catch!

i believe though the lack of warning is because we were not tracking struct expansions correctly. I will push a fix shortly.

hauleth:
@josevalim I believe that the same problem is with `defrecord`/`defrecordp`. I couldn't find other macros that generate more than one function in language core though.

josevalim:
@hauleth you are correct, `defrecord` does have this issue.

PR #8100, opened 2018-08-17
Use `@t :: %__MODULE__{}` convention in `Range`.

basdirks:
Use `@t :: %__MODULE__{}` convention.


josevalim:
Both ways are fine. Since Range is a short module name, it makes sense to keep it as is. Thanks!

PR #8101, opened 2018-08-17
Use `@t :: %__MODULE{}` convention in `Time`.

basdirks:


josevalim:
Both ways are fine. Since Time is a short module name, it makes sense to keep it as is. Thanks!

PR #8102, opened 2018-08-17
Use `@t :: %__MODULE__{}` convention in `Date`.

basdirks:


josevalim:
Both ways are fine. Since Date is a short module name, it makes sense to keep it as is. Thanks!

PR #8103, opened 2018-08-17
Use `@t :: %__MODULE__{}` convention in `NaiveDateTime`.

basdirks:


josevalim:
Now you got me. :)

According to my logic, we would need to change to `__MODULE__`, but then the calendar modules would be inconsistent between them. So after all it is probably better to be consistent.

If we are going to do so, however, I think we should discuss which approach we prefer and do one change altogether for all modules at once, instead of separate pull requests.

So I propose the following:

1. Let's create a new issue to discuss which one we want to follow `__MODULE__` or the literal module name.

2. Then we do a single PR that does the change at once.

What do you think?

basdirks:
That's fine with me. Will you or shall I?

josevalim:
@basdirks please go ahead :)

josevalim:
Closing this for now then, thank you!

PR #8108, opened 2018-08-17
Do not run doctests with empty :only option

fertapric:
It's a weird case, but I guess that's how it should behave according to the semantics of `:only`.

Raising an `ArgumentError` could be an alternative :)

### DIFF
--- a/lib/ex_unit/lib/ex_unit/doc_test.ex
+++ b/lib/ex_unit/lib/ex_unit/doc_test.ex
@@ -232,12 +232,20 @@ defmodule ExUnit.DocTest do
   end
 
   defp filter_by_opts(tests, opts) do
-    only = opts[:only] || []
-    except = opts[:except] || []
+    except = Keyword.get(opts, :except, [])
 
-    tests
-    |> Stream.reject(&(&1.fun_arity in except))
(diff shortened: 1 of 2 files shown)


lib/ex_unit/lib/ex_unit/doc_test.ex
```diff
@@ -232,12 +232,20 @@ defmodule ExUnit.DocTest do
...
 
   defp filter_by_opts(tests, opts) do
-    only = opts[:only] || []
-    except = opts[:except] || []
+    except = Keyword.get(opts, :except, [])
 
-    tests
-    |> Stream.reject(&(&1.fun_arity in except))
-    |> Stream.filter(&(Enum.empty?(only) or &1.fun_arity in only))
+    case Keyword.get(opts, :only) do
```
whatyouhide: Could we replace this with `Keyword.fetch/2`?

fertapric: Much better! Thanks!

fertapric: Done in https://github.com/elixir-lang/elixir/pull/8108/commits/a900bbde0f52561bd470ae001335d246ef8ec1dd

josevalim:

PR #8109, opened 2018-08-17
Consider :runtime and :app from mix.exs

josevalim:
Since those values are not stored on Hex, we should not
expect them to be available on Mix.Dep.cached. Instead
we traverse the deps in `mix.exs` to lift this information.

### DIFF
--- a/lib/mix/lib/mix/dep.ex
+++ b/lib/mix/lib/mix/dep.ex
@@ -117,11 +117,6 @@ defmodule Mix.Dep do
     end
   end
 
-  # optional and runtime only matter at the top level.
-  # Any non-top level dependency that is optional and
-  # is still available means it has been fulfilled.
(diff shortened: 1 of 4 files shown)


josevalim:

bitwalker:
Hey @josevalim, does this mean that Distillery should not expect to get that metadata from `Mix.Dep.loaded` or `loaded_by_name` anymore and should consult the project `mix.exs` instead? To refresh your memory, Distillery traverses deps to warn about apps missing from the release (either directly or transitively)

josevalim:
`Mix.Dep` is private API, so it would be best if distillery did not use it. I can help with that. But to answer your question more directly, runtime will work for the top level but not for dependencies (including umbrella children). So it may work depending on how you use it but it is not guaranteed. However, if you move to the approach listed here, it should be backwards compatible and not have such pitfalls.

PR #8110, opened 2018-08-17
Minor tweaks to Agent docs

fxn:
Minor tweaks I did while reading the source code of `Agent`:

* The sample code quoting `Mix.TasksServer` is out of sync. After some discussion, it has been replaced by a simple counter.
* The `Agent` API receives both anonymous functions and MFAs, the patch removes "anonymous" in a sentence of the introduction and refers more vaguely to _functions_.
* I believe functions are _called_ or _invoked_, but I wouldn't say a function is _executed_. You execute a _function call_, or a _callback_, though.
* Specifies that the function passed to `start_link/2` is invoked in the server process.
* Underlines that the value returned by the function passed to `start_link/2` is the _initial_ state of the agent.

### DIFF
--- a/lib/elixir/lib/agent.ex
+++ b/lib/elixir/lib/agent.ex
@@ -11,45 +11,42 @@ defmodule Agent do
 
   ## Examples
 
-  For example, in the Mix tool that ships with Elixir, we need
-  to keep a set of all tasks executed by a given project. Since
-  this set is shared, we can implement it with an agent:
+  For example, the following agent implements a counter:
 
-      defmodule Mix.TasksServer do
+      defmodule Counter do
         use Agent
 
-        def start_link(_) do
-          Agent.start_link(fn -> MapSet.new() end, name: __MODULE__)
+        def start_link do
+          Agent.start_link(fn -> 0 end, name: __MODULE__)
         end
 
-        @doc "Checks if the task has already executed"
-        def executed?(task, project) do
-          item = {task, project}
-
-          Agent.get(__MODULE__, fn set ->
-            item in set
-          end)
+        def value do
+          Agent.get(__MODULE__, & &1)
         end
 
-        @doc "Marks a task as executed"
-        def put_task(task, project) do
-          item = {task, project}
-          Agent.update(__MODULE__, &MapSet.put(&1, item))
-        end
-
-        @doc "Resets the executed tasks and returns the previous list of tasks"
-        def take_all() do
-          Agent.get_and_update(__MODULE__, fn set ->
-            {Enum.into(set, []), MapSet.new()}
-          end)
+        def increment do
+          Agent.update(__MODULE__, &(&1 + 1))
         end
       end
 
-  Agents provide a segregation between the client and server APIs (similar
-  to GenServers). In particular, the anonymous functions given to the `Agent`
-  are executed inside the agent (the server). This distinction is important
-  because you may want to avoid expensive operations inside the agent,
-  as they will effectively block the agent until the request is fulfilled.
+  Usage would be:
+
+      Counter.start_link
+
+      Counter.value     #=> 0
+      Counter.increment #=> :ok
+      Counter.increment #=> :ok
+      Counter.value     #=> 2
+
+  Thanks to the agent server process, the counter can be safely incremented
+  concurrently.
+
+  Agents provide a segregation between the client and server APIs (similar to
+  GenServers). In particular, the functions passed as arguments to the calls to
+  `Agent` functions are invoked inside the agent (the server). This distinction
+  is important because you may want to avoid expensive operations inside the
+  agent, as they will effectively block the agent until the request is
+  fulfilled.
 
   Consider these two examples:
 
@@ -179,9 +176,9 @@ defmodule Agent do
 
   This is often used to start the agent as part of a supervision tree.
 
-  Once the agent is spawned, the given function `fun` is invoked and its return
-  value is used as the agent state. Note that `start_link/2` does not return
-  until the given function has returned.
+  Once the agent is spawned, the given function `fun` is invoked in the server
+  process, and should return the initial agent state. Note that `start_link/2`
+  does not return until the given function has returned.
 
   ## Options
 


lib/elixir/lib/agent.ex
```diff
@@ -18,37 +18,27 @@ defmodule Agent do
...
+          Agent.update(__MODULE__, fun, 30000)
         end
       end
 
-  Agents provide a segregation between the client and server APIs (similar
-  to GenServers). In particular, the anonymous functions given to the `Agent`
-  are executed inside the agent (the server). This distinction is important
-  because you may want to avoid expensive operations inside the agent,
+  Agents provide a segregation between the client and server APIs (similar to
+  GenServers). In particular, the functions passed as arguments to the `Agent`
```
fertapric: I find “to the `Agent` function calls” a bit confusing. What do you think about?

> In particular, the functions passed as arguments to the function calls of `Agent` are invoked inside the agent (the server).

lib/elixir/lib/agent.ex
```diff
@@ -18,37 +18,27 @@ defmodule Agent do
...
+          update(&Map.put(&1, tuple, true))
         end
 
-        @doc "Resets the executed tasks and returns the previous list of tasks"
-        def take_all() do
-          Agent.get_and_update(__MODULE__, fn set ->
-            {Enum.into(set, []), MapSet.new()}
-          end)
+        defp update(fun) do
+          Agent.update(__MODULE__, fun, 30000)
```
michalmuskala: I'm actually not sure I agree with having an example like that in the documentation.

I'd say allowing passing arbitrary functions to the agent from outside leaks some implementation details - I think only having a well defined set of operations exposed as functions is a better approach. This is similar to having direct `GenServer.call/cast` calls in modules outside of the server itself.

josevalim: @michalmuskala this is a private function though? I often like to do exactly this, to encapsulate the agent name+timeout in a single place.

michalmuskala: Oh, I'm sorry. For some reason I missed the `p`. This looks fine now.

I guess it's just a little weird because we have write functions on the agent, but not a single read function.

fxn: Good point. What if we change the example with a very simple `get`/`set` agent?

lib/elixir/lib/agent.ex
fxn: Sounds good!

lib/elixir/lib/agent.ex
fxn: The "reading" function in `Mix.TasksServer` is

```elixir
  def run(tuple) do
    get_and_update(fn set ->
      {not Map.has_key?(set, tuple), Map.put(set, tuple, true)}
    end)
  end
```

But without more context, I guess a function called `run` that doesn't execute anything may be confusing.

The KV example in https://elixir-lang.org/getting-started/mix-otp/agent.html, for example, seems more simple to me. It could be even an agent that stores a single scalar like a number or something like that (but still with a realistic use case).

fxn: One possibility could be an agent that provides a synchronized sequence within a node:

```elixir
defmodule Sequence do
  use Agent

  def start_link do
    Agent.start_link(fn -> 0 end, name: __MODULE__)
  end

  def next do
    Agent.get_and_update(__MODULE__, &({&1, &1 + 1}))
  end
end
```

As first useful agent example, it can't get more simple than that! 😄

michalmuskala: Yes, I think either a counter or a get-set register would be good examples, that don't require the reader to also learn the problem, but just the solution.

lib/elixir/lib/agent.ex
fxn: On a second read, "the function calls of Agent" does not read well to me, because `Agent` has functions, not function calls.

In the original sentence we refer to "`Agent` function calls", meaning calls to `Agent` functions. 

WDYT?

fxn:
@michalmuskala I've pushed a counter, to have a simple read/write interface in the example. WDYT?

lib/elixir/lib/agent.ex
fertapric: “the calls to `Agent` functions”?

josevalim:
Sorry for the delay folks.

The reason I don't like a counter is because it is very inefficient and we have this tool built-in with `System.unique_integer`. So I am afraid that, although it is simple, it is not practical.

Here are other examples to consider:

  1. A getter/setter (althought it is very simple too)

  2. Something that reads the data on start_link and then you can query it. This is a common use case. Sometimes people want to read the number of countries in a file separated by new lines... or a JSON configuration and have it accessible through the application life cycle. This should be simple to implement and generally applicable

Thoughts?

fxn:
@josevalim good point re `System.unique_integer`, an example cannot be artificial in that sense.

What do you think about quoting [`Mix.State`](https://github.com/elixir-lang/elixir/blob/master/lib/mix/lib/mix/state.ex)?

fxn:
@josevalim let me clarify a detail nonetheless.

At some point in the discussion, I posted a `Sequence` agent with a `next/0` function that reimplemented `Syste.unique_integer([:positive, :monotonic])` (was not aware of this API). But the current patch implements a counter, which is different. Processes in a node can read this counter as many times as they want, and it is atomically incremented.

Would you implement this particular agent in a different way?

I ask because the example is very simple and has a read/write API. If idiomatic, I would lean this one to `Mix.State`.

fxn:
@fertapric 👍 about the rewording.

I have it ready locally, will force push when we finish the discussion about the example.

josevalim:
> I ask because the example is very simple and has a read/write API. If idiomatic, I would lean this one to Mix.State.

Oh, I see. The current approach is also fine by me. :+1: Your call!

lib/elixir/lib/agent.ex
```diff
@@ -179,9 +176,9 @@ defmodule Agent do
 
   This is often used to start the agent as part of a supervision tree.
 
-  Once the agent is spawned, the given function `fun` is invoked and its return
-  value is used as the agent state. Note that `start_link/2` does not return
-  until the given function has returned.
+  Once the agent is spawned, the given function `fun` is invoked in the server
+  process, and should return the initial agent state. Note that `start_link/2`
```
umamaistempo: I like being explicit about such details :)

josevalim:

PR #8111, opened 2018-08-17
Update docs for `ExUnit.configure/1` and `ExUnit.Formatter`

devonestes:
This includes documentation explicitly allowing the passing of arbitrary
configuration options in `ExUnit.configure/1`, and also tells how to do
runtime configuration of custom formatters by using this arbitrary
configuration.


lib/ex_unit/lib/ex_unit.ex
```diff
@@ -270,6 +270,9 @@ defmodule ExUnit do
       prints each test case and test while running. Note that in trace mode test timeouts
       will be ignored.
 
+  Any arbitrary configuration can also be passed to `configure/1`, and these options
```
fertapric: I think you can also pass the configuration options to `start/1`, should that be mentioned here?

josevalim:

PR #8116, opened 2018-08-20
Update test module sent to formatters

### LINKED ISSUE #8112: Include test times when sending `:module_finished` event to ExUnit formatters
### Environment

Erlang/OTP 21 [erts-10.0] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1] [hipe]

Elixir 1.7.1 (compiled with Erlang/OTP 19)
* Operating system: MacOS 10.13.6

### Current behavior

When writing a custom formatter, the `%ExUnit.TestModule{}` struct that is sent as part of the `:module_finished` cast includes all of the tests in that module, but it does not include the correct run times of those tests. For example, in that message I get:

```elixir
%ExUnit.TestModule{
  name: TestmetricsElixirClientTest.BasicTest,
  state: nil,
  tests: [
    %ExUnit.Test{
      case: TestmetricsElixirClientTest.BasicTest,
      logs: "",
      module: TestmetricsElixirClientTest.BasicTest,
      name: :"test false",
      state: nil,
      tags: %{
        async: true,
        describe: nil,
        describe_line: nil,
        file: "/Users/devoncestes/sandbox/testmetrics/testmetrics_elixir_client/test/testmetrics_elixir_client_test.exs",
        line: 13,
        registered: %{},
        test_type: :test
      },
      time: 0
    },
    %ExUnit.Test{
      case: TestmetricsElixirClientTest.BasicTest,
      logs: "",
      module: TestmetricsElixirClientTest.BasicTest,
      name: :"test true",
      state: nil,
      tags: %{
        async: true,
        describe: nil,
        describe_line: nil,
        file: "/Users/devoncestes/sandbox/testmetrics/testmetrics_elixir_client/test/testmetrics_elixir_client_test.exs",
        line: 9,
        registered: %{},
        test_type: :test
      },
      time: 0
    }
  ]
}
```

### Expected behavior

I would expect that the `:time` for each test in the test module would be updated to reflect how long the test actually took to run. Essentially, I would expect that the `:tests` key would be updated to include the new structs that are sent as part of the `:test_finished` message, so they include the run times of each test and an updated `:state` of each test, including failure information for failed tests.

devonestes:
Now the `:module_finished` message will include the results of the
finished tests.

Resolves #8112

### DIFF
--- a/lib/ex_unit/lib/ex_unit/runner.ex
+++ b/lib/ex_unit/lib/ex_unit/runner.ex
@@ -117,9 +117,11 @@ defmodule ExUnit.Runner do
     # run and which ones were skipped.
     tests = prepare_tests(config, test_module.tests)
 
-    {test_module, pending} =
+    {test_module, pending, finished_tests} =
       if Enum.all?(tests, & &1.state) do
-        {test_module, tests}
+        # The pending tests here aren't actually run, so they're already
+        # "finished"
+        {test_module, tests, tests}
       else
         spawn_module(config, test_module, tests)
       end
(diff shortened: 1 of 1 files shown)


josevalim:

PR #8120, opened 2018-08-21
Missing comma in README.md

bugaiov:
Hi All,

Just wanted to make my first contribution to elixir and immediately found a place in `README.md` where I could do it.

A very-very minor contribution. I know :-)

Nick

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,7 +138,7 @@ We outline our process below to clarify the roles of everyone involved.
 
 All pull requests must be approved by two committers before being merged into
 the repository. If any changes are necessary, the team will leave appropriate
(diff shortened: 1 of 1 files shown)


josevalim:
Thanks @bugaiov! I believe the last time we had a similar discussion the consensus is that the comma is not actually required, so I would rather leave things as is. Thoughts?

bugaiov:
@josevalim From my english lessons I remember the following pattern: if the word `Unfortunately` goes first in the sentence it should be accompanied by comma. But if you already discussed this topic, I'm totally fine with leaving it as is.

PR #8122, opened 2018-08-22
Docs: change wording about WithClauseError

schnittchen:
This way I hope people will not fall into the trap that no else has different error semantics, as mentioned above in the docs.


josevalim:
Maybe we should be clearer then: "If an `else` block is used and there is no matching clause, ...". Thoughts?

schnittchen:
I'm all for clarity.

Unless somebody comes up with something better, I will change and push to that wording in a couple of hours.

whatyouhide:
Agreed with @josevalim. Also please use the word "clause" over "condition" 💟

josevalim:

PR #8127, opened 2018-08-24
Clarify that multiple guards are useful for exceptions

nathanl:
The previous explanation here talked about failing guards, but it wasn't
clear to me that exceptions were the specific kind of failure where
multiple guards are more correct than multiple expressions separated by
`or`.


josevalim:

PR #8129, opened 2018-08-24
max-failures implementation for ExUnit

hylkealons:
Max failures implementation for Exunit

It's still a work in progress, but I would like to ask some feedback about a proper way to stop testing. 

What is advised to get back to the `Runner` and hit the `EventManager.stop`?

### DIFF
--- a/lib/ex_unit/lib/ex_unit.ex
+++ b/lib/ex_unit/lib/ex_unit.ex
@@ -181,7 +181,6 @@ defmodule ExUnit do
   @spec start(Keyword.t()) :: :ok
   def start(options \\ []) do
     {:ok, _} = Application.ensure_all_started(:ex_unit)
-
     configure(options)
 
     if Application.fetch_env!(:ex_unit, :autorun) do
@@ -243,6 +242,9 @@ defmodule ExUnit do
       different modules run in parallel. It defaults to `System.schedulers_online * 2`
       to optimize both CPU-bound and IO-bound tests;
 
+    * `:max_failures` - maximum number of tests that may fail. When the maximum is reached, the
+       execution of tests is stopped and the results are returned.
+
     * `:module_load_timeout` - the timeout to be used when loading a test module,
       defaults to `60_000` milliseconds;
 
@@ -290,6 +292,15 @@ defmodule ExUnit do
     |> put_seed()
     |> put_slowest()
     |> put_max_cases()
+    |> put_max_failures()
+  end
+
+  defp put_max_failures(opts) do
+    if opts[:max_failures] > 0 do
+      Keyword.put(opts, :max_failures, opts[:max_failures])
+    else
+      opts
+    end
   end
 
   @doc """
@@ -349,7 +360,7 @@ defmodule ExUnit do
   # Persists default values in application
   # environment before the test suite starts.
   defp persist_defaults(config) do
-    config |> Keyword.take([:max_cases, :seed, :trace]) |> configure()
+    config |> Keyword.take([:max_cases, :seed, :trace, :max_failures]) |> configure()
     config
   end
 
--- a/lib/ex_unit/lib/ex_unit/runner.ex
+++ b/lib/ex_unit/lib/ex_unit/runner.ex
@@ -1,6 +1,21 @@
 defmodule ExUnit.Runner do
   @moduledoc false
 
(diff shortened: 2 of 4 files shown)


josevalim:
Thanks @hylkealons. I think the max_failures should be handled in the runner process directly and once we reach it, we just proceed to stop the runner as a whole.

hylkealons:
not sure how that would work @josevalim, as the runner_stats are in charge of keeping track of the failed test. Could you elaborate a bit? I am new to this codebase ;)

josevalim:
The runner stats is just for statistical purposes, it doesn't have control over the execution of the tests. It should not be changed at all. `ExUnit.Runner` is the one responsible for spawning tests and it should be the one that tracks the number of failures compared to max failures. Does it make sense? You are right that it doesn't track this information at all. But now it has to.

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -1,14 +1,30 @@
 defmodule ExUnit.Runner do
   @moduledoc false
 
+  defmodule MaxFailureCounter do
```
fertapric: Should this be `FailuresCounter`? This module counts the number of failures, but it has no knowledge of "maximum number of failures allowed".

I would also rename all `fail_counter` variables to `failures_counter` for consistency.

lib/ex_unit/lib/ex_unit.ex
```diff
@@ -243,6 +242,9 @@ defmodule ExUnit do
       different modules run in parallel. It defaults to `System.schedulers_online * 2`
       to optimize both CPU-bound and IO-bound tests;
 
+    * `:max_failures` - maximum number of tests that may fail. When the maximum is reached, the
+       execution of tests is stopped and the results are returned.
```
fertapric: Minor nitpick: looking at other items, I think this line should end with a semicolon `;` for consistency (and not with a dot `.`)

lib/mix/lib/mix/tasks/test.ex
```diff
@@ -156,6 +156,8 @@ defmodule Mix.Tasks.Test do
       external commands which produce output on stdout upon file system modification
     * `--max-cases` - sets the maximum number of tests running async. Only tests from
       different modules run in parallel. Defaults to twice the number of cores
+    * `--max-failures` - maximum number of tests that may fail. When the maximum is reached, the
+       execution of tests is stopped and the results are returned.
```
fertapric: Minor nitpick: looking at other items, I think this line should not end with a dot `.` for consistency.

lib/ex_unit/lib/ex_unit.ex
```diff
@@ -290,6 +292,15 @@ defmodule ExUnit do
     |> put_seed()
     |> put_slowest()
     |> put_max_cases()
+    |> put_max_failures()
+  end
+
+  defp put_max_failures(opts) do
```
fertapric: I think this function is always returning the same `opts`, no matter what. I think it could be removed.

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -241,9 +259,19 @@ defmodule ExUnit.Runner do
       end
 
     EM.test_finished(config.manager, test)
+    handle_result(config, test)
     test
   end
 
+  defp handle_result(config, %{state: {:failed, _}}) do
```
fertapric: Micro-optimization: ignore this logic if the `--max-failures` flag is not given.

```elixir
defp handle_result(%{max_failures: max_failures}, %{state: {:failed, _}}) when is_integer(max_failures) do
```

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -241,9 +259,19 @@ defmodule ExUnit.Runner do
...
+    failures = MaxFailureCounter.increase(config.fail_counter)
```
josevalim: Is there a way to not use an agent? We should avoid mutable state as much as possible. I would prefer to track this in the runner state itself.

hylkealons:
I'll close this PR for now as it's more complex than I thought initially.

eksperimental:
I will give this a try and submit a PR

eksperimental:
related #4586 and #7480

PR #8131, opened 2018-08-27
"Fall back" the verb is 2 separate words

eksperimental:


lib/elixir/lib/kernel.ex
```diff
@@ -4477,12 +4477,12 @@ defmodule Kernel do
...
   If a protocol implementation is not found for a given type,
   invoking the protocol will raise unless it is configured to
-  fallback to `Any`. Conveniences for building implementations
+  fall back to `Any`. Conveniences for building implementations
   on top of existing ones are also available, look at `defstruct/1`
   for more information about deriving
   protocols.
 
-  ## Fallback to any
+  ## Fall back to `Any`
```
whatyouhide: I think here "fallback" is used as a noun, as in "The fallback to `Any` scenario" rather than "How to fall back to `Any`". But this might be my interpretation, so let's wait for others to pitch in :)

josevalim:

PR #8132, opened 2018-08-27
Use "file system" instead of "filesystem"

eksperimental:
Standardize use of this expression.
While both accepted, Google Books Ngram Viewer shows that "file system"
is the prefered one.


josevalim:

PR #8133, opened 2018-08-27
[WIP] Include arg name in FunctionClauseError hint

eksperimental:
Adds the argument name next the argument number, as shown bellow:

iex> Enum.reverse_slice([:a, :b, :c], 1, :two)
** (FunctionClauseError) no function clause matching in Enum.reverse_slice/3

    The following arguments were given to Enum.reverse_slice/3:

        # 1 (enumerable)
        [:a, :b, :c]

        # 2 (start)
        1

        # 3 (count)
        :two

### DIFF
--- a/lib/elixir/lib/exception.ex
+++ b/lib/elixir/lib/exception.ex
@@ -1048,22 +1048,38 @@ defmodule FunctionClauseError do
   end
 
   def blame(exception, inspect_fun, ast_fun) do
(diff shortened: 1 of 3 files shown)


josevalim:
Given this requires to access the docs chunk on top of evaling, I don't want to add even more work and complexity to the evaluation of blame. It is fine if we lift those from the AST but I would prefer to not require handling of the docs chunk. Thanks!

PR #8134, opened 2018-08-27
Correct blame message in FunctionClauseError for funs of arity 1

eksperimental:
before
iex> Time.to_string(:foo)
** (FunctionClauseError) no function clause matching in Time.to_string/1

    The following arguments were given to Time.to_string/1:

        # 1
        :foo

    (elixir) lib/calendar/time.ex:166: Time.to_string/1

Now the message reads:
The following argument was given to Time.to_string/1:

### DIFF
--- a/lib/elixir/lib/exception.ex
+++ b/lib/elixir/lib/exception.ex
@@ -1051,7 +1051,14 @@ defmodule FunctionClauseError do
     %{module: module, function: function, arity: arity, kind: kind, args: args, clauses: clauses} =
       exception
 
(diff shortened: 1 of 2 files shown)


whatyouhide:
I'm on the fence on this one because in this kind of "scientific" contexts I see lists as lists of one or more element with no distinction for lists of one element vs more than one. I'll leave this to others :)

josevalim:
I am flag to keep it as is as well for the reasons @whatyouhide mentioned. I believe we even discussed a why to keep the sentence "counting neutral" when we first added it. Thanks @eksperimental!

PR #8136, opened 2018-08-27
Add section "Reserved aliases" in defmodule/2

eksperimental:


lib/elixir/lib/kernel.ex
```diff
@@ -3748,6 +3748,19 @@ defmodule Kernel do
   Note that, when a dynamic name is used, Elixir won't nest the name under the
   current module nor automatically set up an alias.
 
+  ## Reserved aliases
```
fertapric: I find the term "Reserved aliases" confusing 😕 Reading it, I would expect the error when using `alias XX.PID`, for example:

```elixir
iex(1)> defmodule Foo.PID do
...(1)> def foo do
...(1)> 1
...(1)> end
...(1)> end
{:module, Foo.PID,
 <<70, 79, 82, 49, 0, 0, 3, 192, 66, 69, 65, 77, 65, 116, 85, 56, 0, 0, 0, 125,
   0, 0, 0, 13, 14, 69, 108, 105, 120, 105, 114, 46, 70, 111, 111, 46, 80, 73,
   68, 8, 95, 95, 105, 110, 102, 111, 95, ...>>, {:foo, 0}}
iex(2)> alias Foo.PID # Reading "reserved aliases", I would expect the error here
Foo.PID
iex(3)> PID.foo
1 
```

Maybe "Reserver module names"?

whatyouhide: The name of something like `X` or `Foo.Bar` is indeed "alias" and not module name. However `PID` is not reserved as an alias, only as a module name, so 👍 from me on "Reserved module names".

PR #8139, opened 2018-08-27
Improvements in Task module 

eksperimental:


lib/elixir/lib/task.ex
```diff
@@ -183,24 +183,40 @@ defmodule Task do
...
+  @spec child_spec(term) :: Supervisor.child_spec()
   def child_spec(arg) do
-    %{
+    spec = %{
       id: Task,
       start: {Task, :start_link, [arg]},
       restart: :temporary
     }
+
+    Supervisor.child_spec(spec, [])
```
whatyouhide: Why this change?

lib/elixir/lib/task.ex
```diff
@@ -701,11 +743,12 @@ defmodule Task do
...
   end
 
-  defp yield_many([%Task{ref: ref, owner: owner} = task | rest], timeout_ref, timeout) do
-    if owner != self() do
-      raise ArgumentError, invalid_owner_error(task)
-    end
+  defp yield_many([%Task{owner: owner} = task | _rest], _timeout_ref, _timeout)
+       when owner != self() do
+    raise ArgumentError, invalid_owner_error(task)
+  end
```
whatyouhide: I think this is better with just one clause so we do the whole pattern matching on the head just once. If we really want to go this way, I would invert the clauses and go with `when owner == self()` first and then fall back to the error case otherwise.

lib/elixir/lib/task.ex
```diff
@@ -515,13 +547,16 @@ defmodule Task do
 
   """
   @spec await(t, timeout) :: term
-  def await(task, timeout \\ 5000)
+  def await(%Task{} = task, timeout \\ 5000)
```
fertapric: Two comments here:

1. I think we could use `defguardp` here to define a `is_timeout`
2. I'm not sure if this kind of defensive programming is needed here (I like it for my personal projects, but I'm not sure if that's also the preference in the Elixir codebase). If you pass an invalid value in `timeout`, it will raise `** (ErlangError) Erlang error: :timeout_value` which I think is self-explanatory (or maybe we don't want to raise `ErlangError`?).

lib/elixir/lib/task.ex
josevalim: I think it is trying to be consistent with other modules but other modules do not accept options here, so this change is not necessary.

lib/elixir/lib/task.ex
josevalim: Yes, let's keep the conditional here. :+1:

lib/elixir/lib/task.ex
josevalim: :+1: for the `is_timeout` guard. I think having a guard is good, so let's keep it.

lib/elixir/lib/task.ex
```diff
@@ -229,30 +250,35 @@ defmodule Task do
...
 
+  `fun` must be a zero-arity anonymous function.
+
   This is only used when the task is used for side-effects
   (i.e. no interest in the returned result) and it should not
   be linked to the current process.
   """
   @spec start((() -> any)) :: {:ok, pid}
-  def start(fun) do
+  def start(fun) when is_function(fun, 0) do
```
fertapric: @josevalim, does [this comment](https://github.com/elixir-lang/elixir/pull/8087#discussion_r210400252) also apply in this case?

lib/elixir/lib/task.ex
eksperimental: sorry, I thought a map with default values was created by Supervisor.child_spec/2, that's why I passed it through

lib/elixir/lib/task.ex
```diff
@@ -735,8 +778,8 @@ defmodule Task do
   Returns `{:ok, reply}` if the reply is received while shutting down the task,
   `{:exit, reason}` if the task died, otherwise `nil`.
 
-  The second argument is either a timeout or `:brutal_kill`. In case
-  of a `timeout`, a `:shutdown` exit signal is sent to the task process
+  The second argument is either a timeout value in milliseconds, :infinity or `:brutal_kill`.
```
eksperimental: I think it makes no sense to accept `:infinity` as a timeout, please correct me if I'm overseeing something.
This is already checked in the guards but doc and spec needs to be updated

lib/elixir/lib/task.ex
eksperimental: done

lib/elixir/lib/task.ex
eksperimental: Great, I have taken it a step further and apply this in other places

lib/elixir/lib/task.ex
```diff
@@ -735,8 +771,9 @@ defmodule Task do
   Returns `{:ok, reply}` if the reply is received while shutting down the task,
   `{:exit, reason}` if the task died, otherwise `nil`.
 
-  The second argument is either a timeout or `:brutal_kill`. In case
-  of a `timeout`, a `:shutdown` exit signal is sent to the task process
+  The second argument is either a timeout value in milliseconds or`:brutal_kill`.
```
fertapric: Minor nitpick: missing space between "or" and "`:brutal_kill`"

lib/elixir/lib/task.ex
```diff
@@ -183,18 +183,35 @@ defmodule Task do
     * `:ref` - the task monitor reference
 
     * `:owner` - the PID of the process that started the task
-
```
fertapric: Minor thingy: I would keep this empty line :)

lib/elixir/lib/task.ex
```diff
@@ -752,18 +789,29 @@ defmodule Task do
...
+      owner != self() ->
+        raise ArgumentError, invalid_owner_error(task)
+
+      true ->
+        shutdown(task, pid, timeout)
+    end
   end
 
-  def shutdown(%Task{pid: pid} = task, :brutal_kill) do
+  def shutdown(task, pid, :brutal_kill) do
```
fertapric: I think this should be a private function.

fertapric: In addition, the `pid` variable is contained in the `task` variable (`%Task{pid: pid} = task`), does it make sense to pass it as an extra argument?

lib/elixir/lib/task.ex
```diff
@@ -752,18 +789,29 @@ defmodule Task do
...
+          {:ok, term} | {:exit, term} | nil
+  def shutdown(task, timeout_or_brutal_kill \\ 5000)
 
-  def shutdown(%Task{pid: nil} = task, _) do
-    raise ArgumentError, "task #{inspect(task)} does not have an associated task process"
-  end
+  def shutdown(%Task{pid: pid, owner: owner} = task, timeout)
+      when is_timeout(timeout) or timeout == :brutal_kill do
+    cond do
+      timeout == :infinity ->
```
josevalim: `:infinity` is definitely a valid value here. `:infinity` is a valid shutdown value for supervisors and therefore it should be a valid value here too. Usually when we say timeout, `:infinity` is included, that's why we don't say it explicitly in the docs.

josevalim: Let's please revert the changes to `shutdown` function. Let's keep only the improve `timeout_or_brutal_kill` variable name for the signature and make sure to use `is_timeout(timeout)` in the last clause of the previous implementation.

lib/elixir/lib/task.ex
```diff
@@ -752,18 +789,29 @@ defmodule Task do
   a response waiting in the message queue this function will return
   `{:exit, :noproc}` as the result or exit reason can not be determined.
   """
-  @spec shutdown(t, timeout | :brutal_kill) :: {:ok, term} | {:exit, term} | nil
```
josevalim: The spec here was clear that `:infinity` was allowed.

lib/elixir/lib/task.ex
```diff
@@ -752,18 +789,29 @@ defmodule Task do
...
+        raise ArgumentError,
+              ":infinity is not a valid timeout, please use a value in milliseconds or :brutal_kill"
 
-  def shutdown(%Task{owner: owner} = task, _) when owner != self() do
-    raise ArgumentError, invalid_owner_error(task)
+      is_nil(pid) ->
+        raise ArgumentError, "task #{inspect(task)} does not have an associated task process"
+
+      owner != self() ->
+        raise ArgumentError, invalid_owner_error(task)
```
josevalim: Shutdown does not require this check because we always monitor ourselves (to avoid data races).

lib/elixir/lib/task.ex
josevalim: Actually, `shutdown` was a perfectly valid name too. Let's keep it.

lib/elixir/lib/task.ex
eksperimental: @josevalim took care of it

lib/elixir/lib/task.ex
eksperimental: Besides `:infinity` being a valid timeout, what is the point of allowing that in the `shutdown` function.\
It will never shutdown, or am I missing something,
Otherwise I think we should raise and inform the user that is going to do something that doesn't make sense.

lib/elixir/lib/task.ex
eksperimental: see the discussion above https://github.com/elixir-lang/elixir/pull/8139#discussion_r216138474

josevalim: On shutdown we send an exit signal. Passing `:infinity` means you are going to give it all the time it needs to shutdown. It doesn't mean we just wait, as a signal is sent anyway.

Some processes are guaranteed to shutdown, such as supervisors, so giving them time to terminate is fair.

eksperimental:
I think this is ready for merging

lib/elixir/lib/task.ex
```diff
@@ -736,7 +773,7 @@ defmodule Task do
   `{:exit, reason}` if the task died, otherwise `nil`.
 
   The second argument is either a timeout or `:brutal_kill`. In case
-  of a `timeout`, a `:shutdown` exit signal is sent to the task process
+  of a timeout, a `:shutdown` exit signal is sent to the task process
   and if it does not exit within the timeout, it is killed. With `:brutal_kill`
   the task is killed straight away. In case the task terminates abnormally
   (possibly killed by another process), this function will exit with the same reason.
```
fertapric: Minor thingy: there is a double space a few paragraphs below: 
```
If a task's monitor has already been demonitored or received  and 
```

between "received" and "and". Do you mind fixing that too? 🙂

eksperimental: not at all! thank you and fixed

josevalim:

PR #8140, opened 2018-08-27
Minor tweaks to Enum documentation

fertapric:
* Replace `zip/*` with `zip/1` and `zip/2` to enable linking
* Use relative linking for functions within the same module

cc @eksperimental


whatyouhide:
Thanks @fertapric! 💟

PR #8144, opened 2018-08-28
Fix typo in Enum.filter/2 @doc

gusaiani:
Hi all, 

added a missing word:

`…function a truthy value.` → `…function returns a truthy value.`

Cheers!


whatyouhide:
Thanks @gusaiani! 💟

Issue #8128, opened 2018-08-24
Date.add/2 seems broken for when the year rolls into a negative leap year


paulanthonywilson:
### Precheck

* For bugs, do a quick search and make sure the bug has not yet been reported

Can't find anything.

### Environment

```
Hex:    0.18.1
Elixir: 1.7.3
OTP:    21.0.5
```
Mac OS High Sierra 10.13.6

### Current behaviour

```
iex(1)> Date.add(~D[-0005-12-31], 1)
~D[-0005-12-32]
```

Note that this is only an issue with dates with negative years.

### Expected behaviour

```
iex(1)> Date.add(~D[-0005-12-31], 1)
~D[-0005-12-32]
```

**Updated 2018-08-26** : Originally found when preparing a Property Testing demonstration; just now narrowed the issue down to being about negative leap years.

paulanthonywilson:
Ok, so I've narrowed it down to a problem with the undocumented `Calendar.ISO.date_from_iso_days/1`:

(excuse the pipe)
```
iex(1)> Calendar.ISO.date_to_iso_days(-4, 1, 1) |> Calendar.ISO.date_from_iso_days()  
{-5, 12, 32}
```
Specifically the issue seems to be in the private `days_to_year/x` functions, but I am really struggling to understand the maths.

josevalim:
Closing in favor of the PR, thanks!

PR #8145, opened 2018-08-28
Fix for converting from negative iso days on New Year in a leap year

paulanthonywilson:
👋 

This is a fix for https://github.com/elixir-lang/elixir/issues/8128

tldr; it was a bug with the 1st of January, on negative leap years - converting back from `iso_days` gave us the 32nd of December on the previous year.

### DIFF
--- a/lib/elixir/lib/calendar/iso.ex
+++ b/lib/elixir/lib/calendar/iso.ex
@@ -206,20 +206,11 @@ defmodule Calendar.ISO do
 
   # Converts count of days since 0000-01-01 to {year, month, day} tuple.
   @doc false
-  def date_from_iso_days(days) when days in 0..3_652_424 do
+  def date_from_iso_days(days) when days in -3_652_059..3_652_424 do
     {year, day_of_year} = days_to_year(days)
     extra_day = if leap_year?(year), do: 1, else: 0
     {month, day_in_month} = year_day_to_year_date(extra_day, day_of_year)
     {year, month, day_in_month + 1}
-  end
-
-  def date_from_iso_days(days) when days in -3_652_059..-1 do
-    {year, day_of_year} = days_to_year(-days)
-    previous_extra_day = if leap_year?(year), do: 1, else: 0
-    extra_day = if leap_year?(year + 1), do: 1, else: 0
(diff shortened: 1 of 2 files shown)


josevalim:
This looks great @paulanthonywilson! There are some comments remaining on the test file, shall we remove them and then go ahead with merging? Thank you!

paulanthonywilson:
@josevalim Thanks - removed comments. Do you want me to move and/or prune the tests?

paulanthonywilson:
(Pretty sure the CI failure isn't me).

josevalim:
@paulanthonywilson i pushed a commit that remove the symptons test, as they theoretically belong to the other module and are already covered by the remaining ones.

The CI failure is indeed a race condition on another test, I will fix it on master, thanks!

lib/elixir/lib/calendar/iso.ex
```diff
@@ -757,20 +748,41 @@ defmodule Calendar.ISO do
...
   end
 
+  defp days_to_year(days) when days < 0 do
+    year_estimate = -div(-days, @days_per_nonleap_year) - 1
+
+    {year, days_before_year} =
+      days_to_year(year_estimate, days, days_to_end_of_epoch(year_estimate))
+
+    leap_year_pad = if leap_year?(year), do: 1, else: 0
+    {year, leap_year_pad + 365 + days - days_before_year}
```
fertapric: Minor nitpick: I would replace `365` with `@days_per_nonleap_year`

lib/elixir/lib/calendar/iso.ex
```diff
@@ -757,20 +748,41 @@ defmodule Calendar.ISO do
...
+  end
+
   defp days_to_year(days) do
-    year = Integer.floor_div(days, @days_per_nonleap_year)
+    year = div(days, @days_per_nonleap_year)
```
fertapric: Another minor nitpick: I like the `year_estimate` naming given in the function above, I would use the same naming here (also for consistency) 🙂

lib/elixir/lib/calendar/iso.ex
paulanthonywilson: Good catch

josevalim:

Issue #7625, opened 2018-04-30
Formatter doesn't work under windows


fklement:
### Environment

* Erlang/OTP 20 [erts-9.3] [64-bit] [smp:6:6] [ds:6:6:10] [async-threads:10]
   Elixir 1.6.4 (compiled with OTP 19)
* Operating system: Windows 10 Pro

### Current behavior

In our project the `.formatter.exs` looks like this:  
```
[
  inputs: [
    "lib/**/*.{ex,exs}",
    "mix.exs",
    "test/support/**/*.{ex,exs}"
  ],
  locals_without_parens: [
    # Kernel
    inspect: 1,
    inspect: 2,

    # Phoenix
    plug: 1,
    plug: 2,
    action_fallback: 1,
    render: 2,
    render: 3,
    render: 4,
    redirect: 2,
    resources: 2,
    resources: 3,
    resources: 4,
    get: 4,
    post: 4,
    delete: 4,
    forward: 4,
    get: 3,
    post: 3,
    delete: 3,
    forward: 3,
    get: 2,
    post: 2,
    delete: 2,
    forward: 2,

    # Ecto Schema
    field: 2,
    field: 3,
    belongs_to: 2,
    belongs_to: 3,
    has_one: 2,
    has_one: 3,
    has_many: 2,
    has_many: 3,
    embeds_one: 2,
    embeds_one: 3,
    embeds_many: 2,
    embeds_many: 3,
    many_to_many: 2,
    many_to_many: 3,
    add: 3,

    # Ecto Query
    from: 2
  ]
]
```

The problem is, that the list get's correctly returned but it does not affect the formatted code when i hit save.  

### Expected behavior
For example:
`embeds_one :rates, Type.Rates` becomes `embeds_one(:rates, Types.Rates)` which should not be.  

My colleague is using MacOS and here works everyting fine.

josevalim:
Does it work if you run “mix format” from your project root? I.e. from the
same directory where .formatter.exs is placed?
-- 


*José Valimwww.plataformatec.com.br
<http://www.plataformatec.com.br/>Founder and Director of R&D*

fklement:
I tried. But nothing happens at all.

josevalim:
It doesn’t happen because the code is already formatted? If you add “foo 1,
2” to a file that is listed, does it add parens?
-- 


*José Valimwww.plataformatec.com.br
<http://www.plataformatec.com.br/>Founder and Director of R&D*

[...]

PR #8147, opened 2018-08-29
Expand path so it works on Windows

princemaple:
https://github.com/elixir-lang/elixir/issues/7625#issuecomment-416017189

- `Path.wildcard` is meant to only work with linux style path
- currently `mix format ANY_WINDOWS_PATH` fails because `Path.wildcard` return empty list
- Assuming `Path` module is correct, I think the path unification should happen in `mix format`
- `Path.expand` is the easiest way I could find that
  - changes windows style path to linux style path
  - keeps wildcard expression working after path transformation


princemaple:
Having to change that test, I'm thinking maybe this change is not good now. Feel free to close.

princemaple:
I've run format on both changed files, they seem to be fine. Why is travis failing?

josevalim:
@princemaple instead of changing the test, change this line:

https://github.com/elixir-lang/elixir/blob/b1dde91d76378afd6ac638e50fc1853a67773f54/lib/mix/lib/mix/tasks/format.ex#L450

To call `#{Path.relative_to_cwd(file)}`. Then the test should still pass.

princemaple:
Thanks for the pointer @josevalim . The failed test seems to be unrelated.

josevalim:

Issue #8146, opened 2018-08-28
Should the compiler warn about an unmatched `else`?


nathanl:
### Environment

* Elixir & Erlang/OTP versions (elixir --version): Elixir 1.7.1 (compiled with Erlang/OTP 21)
* Operating system: MacOS High Sierra (10.13.6)

## Problem

I was just in a debugging session with a colleague where the bug came down to an incorrectly-nested `else`. A minimal reproduction is:

```elixir
defmodule Else do
  def weirdness do
    with true <- true do
      IO.puts("yay")
    end
  else
    match -> IO.inspect(["wat", match])
  end
end

Else.weirdness() # => prints both `"yay"` and `["wat", :ok]`
```

We were very confused to have both the `do` clause of the `with` executed and also the `else`. Eventually we realized that the problem is that the `else` is not nested in the `with`. (Side note: the value of `_match` here is `:ok`, but I don't know why.)

Perhaps an even more minimal reproduction is:

```elixir
defmodule LonelyElse do
  def here do
  else
    match -> IO.inspect(["wat", match])
  end
end
LonelyElse.here() # => prints `["wat", nil]`
```

I don't think either of these `else` clauses makes sense. Could the compiler issue a warning in cases where an `else` does not have a matching `if` or `with` or `catch`?

josevalim:
Looks good to me. If there is no expression on try or the try is a literal, the else is irrelevant. :+1:

josevalim:
This also could be added to the Erlang compiler. /cc @michalmuskala

whatyouhide:
Mh, do we document anywhere that `else` can be used for functions? The docs for `def/2` only mention `rescue`, `catch`, and `after`. I think that we should add it there anyways, or stop supporting it (which I don't think is what we want since everything else works).

OvermindDL1:
Huh, I just tried it, `else` in a def acts like a new case body that the previous main `do` body gets piped in to.  Interesting functionality, I kind of like it, but it's still easy enough to pipe into a new case anyway to do the same thing so probably worth removing anyway.

whatyouhide:
@josevalim we don't.

```
iex(1)> try do
...(1)> :ok
...(1)> end
** (CompileError) iex:1: missing :catch/:rescue/:after/:else option in "try"
```

josevalim:
@whatyouhide so i think we should raise if the try only has `else` too. I can't see any practical usage for it.

nathanl:
[...]

PR #8148, opened 2018-08-29
Document that "else" can be used directly in function definitions

whatyouhide:
Related to #8146.


whatyouhide:
@umamaistempo the `try/1` will be autolinked to the documentation for `try/1` so maybe that's enough as a call to action?

umamaistempo:
It's fine to me. My idea was more to tell the user explicitly to read the `try` documentation but it would be a bit weird :P

So LGTM :+1:

josevalim:
I misunderstood, apologies.

PR #8149, opened 2018-08-29
Document Path.wildcard/2 expects / dir separator

fertapric:
Related to https://github.com/elixir-lang/elixir/pull/8147

As documented in [`filelib:wildcard/1`](http://erlang.org/doc/man/filelib.html#wildcard-1).


josevalim:
I have added a small note, what do yo think?

fertapric:
@josevalim I like it

josevalim:

PR #8152, opened 2018-08-30
Added @impl to the Mix.Task example.

linjunpop:


josevalim:

linjunpop:
Should we also make all the `@impl` in the documentation explicitly? Currently, most of them are using  `@impl true`.

If so, I can submit another PR to change those.

josevalim:
I actually prefer @impl true because 99% of the cases you have only one
behavior and @impl true has no ambiguity. So we should likely stick with
@impl true here too.
-- 


linjunpop:
👌

PR #8154, opened 2018-08-30
Fix invalid expansion for remote

ggcampinho:
When left side is expanded but become invalid, the expansion result in the AST leaking for the user, for example:

```
iex(1)> __ENV__.line.foo
{{:., [line: 1], [1, :foo]}, [line: 1], []}
```

This checks the expanded left to ensure that it's safe


josevalim:

PR #8155, opened 2018-08-31
Remove Version.matchable/0 type

wojtekmach:
It used to be used by public function but hasn't been for a while.

### DIFF
--- a/lib/elixir/lib/version.ex
+++ b/lib/elixir/lib/version.ex
@@ -94,7 +94,6 @@ defmodule Version do
   @type patch :: non_neg_integer | nil
   @type pre :: [String.t() | non_neg_integer]
   @type build :: String.t() | nil
-  @type matchable :: {major :: major, minor :: minor, patch :: patch, pre :: pre}
   @type t :: %__MODULE__{major: major, minor: minor, patch: patch, pre: pre, build: build}
 
   defmodule Requirement do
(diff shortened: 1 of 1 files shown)


devonestes:
Oh, yeah, sorry, you're totally right. GitHub was hiding the fact that we have nested modules in there in the diff. I thought it was just referenced internally.

whatyouhide:
Only concern is that this is a public type. Do we care about breaking the API?

josevalim:
I don’t think it has any use though. We never returned a value matching
this type anywhere, did we?
-- 


whatyouhide:
We didn't I think. So we're probably good to go.

whatyouhide:
Thanks @wojtekmach! 💟

sasa1977:
I found this PR while I wanted to fix #8205. 

One slight comment I'd like to add is that alternatively you could have moved the type into the `Parser` module, or inlined it into the spec for `parse_version`. 

Whether or not it makes sense to typespec internal functions is another matter. It's worth mentioning that [Parser.parse_requirement is still specced](https://github.com/elixir-lang/elixir/blob/v1.7.3/lib/elixir/lib/version.ex#L391), while other exported functions of this module are not. Perhaps it's then worth removing the remaining internal spec too?

sasa1977:
I mostly don't spec private funs, nor funs from internal modules. Occasionally I break this rule to improve the reading experience, but in this particular case, I wouldn't spec the functions from the `Parser` module. So I think that removing the spec is fine, and that the other typespec should probably be removed too, because it looks a bit odd to have just one function in that internal module specced.

PR #8161, opened 2018-09-03
Raise on invalid type annotation in typespecs

### LINKED ISSUE #8143: Invalid type annotations allowed in Elixir
### Environment

```

Elixir 1.8.0-dev (9ab98af68) (compiled with Erlang/OTP 21)
```
* Operating system: Mac OS X 10.13.6

### Current behavior

While working on `{undefined_type, {TypeName, Arity}}` and `{unused_type, {TypeName, Arity}}` of https://github.com/elixir-lang/elixir/issues/5800, I've found that some type annotations are parsed as `:user_type`. This is an example:

https://github.com/elixir-lang/elixir/blob/de42c4cf20a3ac907b5bff81468b509b07d7a907/lib/elixir/lib/process.ex#L28

The resulting AST is:

```elixir
{:type,
 {:type,
  {:::, [line: 7],
   [
     {:dest, [line: 7], nil},
     {:::, [line: 7],
      [
        {:|, [line: 7],
         [
           {:pid, [line: 7], nil},
           {:|, [line: 7],
            [{:port, [line: 7], nil}, {:registered_name, [line: 7], nil}]}
         ]},
        {:|, [line: 7],
         [
           {:atom, [line: 7], nil},
           {{:::, [line: 7],
             [{:registered_name, [line: 7], nil}, {:atom, [line: 7], nil}]},
            {:node, [line: 7], nil}}
         ]}
      ]}
   ]}, {Foo, {7, 1}}}}
```

and the translation passed to build the forms:

```elixir
 {type,
     {dest,0},
     7,
     {dest,
         {ann_type,7,
             [{type,7,union,
                  [{type,7,pid,[]},
                   {type,7,port,[]},
                   {user_type,7,registered_name,[]}]},
              {type,7,union,
                  [{type,7,atom,[]},
                   {type,0,tuple,
                       [{ann_type,7,
                            [{var,7,registered_name},{type,7,atom,[]}]},
                        {type,7,node,[]}]}]}]},
         []},
     true}
```

As seen above, the type annotation `registered_name` is translated as `{user_type,7,registered_name,[]}`.

It seems to be related to operator precedence, because this typespec is parsed properly:

```elixir
@type dest :: pid | port | (registered_name :: atom) | {registered_name :: atom, node}
```

```elixir
{type,{dest,0},
       7,
       {dest,{type,7,union,
                   [{type,7,pid,[]},
                    {type,7,port,[]},
                    {ann_type,7,[{var,7,registered_name},{type,7,atom,[]}]},
                    {type,0,tuple,
                          [{ann_type,7,
                                     [{var,7,registered_name},
                                      {type,7,atom,[]}]},
                           {type,7,node,[]}]}]},
             []},
       true}
```

I tried doing the same with Erlang, but it raises `test.erl:3: syntax error before: '::'`:

```erlang
-type dest() :: pid() | port() | RegisteredName::atom() | {RegisteredName::atom(), node()}.
```

but works if parens are added:

```erlang
-type dest() :: pid() | port() | (RegisteredName::atom()) | {RegisteredName::atom(), node()}.
```

### Expected behavior

I would mirror Erlang and raise a `SyntaxError` in such cases.

If I may, I would like to work on this issue if the expected behavior is desired 🙂

fertapric:
Fixes #8143

### DIFF
--- a/lib/elixir/lib/kernel/typespec.ex
+++ b/lib/elixir/lib/kernel/typespec.ex
@@ -535,29 +535,28 @@ defmodule Kernel.Typespec do
   end
 
   # Handle type operator
-  defp typespec({:::, meta, [var, expr]} = orig, vars, caller) do
-    case var do
-      {var_name, _meta, context} when is_atom(var_name) and is_atom(context) ->
-        case typespec(expr, vars, caller) do
-          {:ann_type, _, _} ->
-            message =
-              "invalid type annotation. Type annotations cannot be nested: " <>
-                "#{Macro.to_string(orig)}"
-
-            compile_error(caller, message)
-
-          right ->
-            left = typespec(var, [elem(var, 0) | vars], caller)
-            {:ann_type, line(meta), [left, right]}
-        end
-
-      _ ->
+  defp typespec({:::, meta, [{var_name, _meta, context} = var, expr]} = ann_type, vars, caller)
+       when is_atom(var_name) and is_atom(context) do
+    case typespec(expr, vars, caller) do
+      {:ann_type, _, _} ->
         message =
-          "invalid type annotation. When using the | operator to represent the union of " <>
-            "types, make sure to wrap type annotations in parentheses: #{Macro.to_string(orig)}"
+          "invalid type annotation. Type annotations cannot be nested: " <>
+            "#{Macro.to_string(ann_type)}"
 
         compile_error(caller, message)
+
+      right ->
+        left = typespec(var, [var_name | vars], caller)
+        {:ann_type, line(meta), [left, right]}
     end
+  end
+
+  defp typespec({:::, _meta, _args} = ann_type, _vars, caller) do
+    message =
+      "invalid type annotation. When using the | operator to represent the union of types, " <>
+        "make sure to wrap type annotations in parentheses: #{Macro.to_string(ann_type)}"
+
+    compile_error(caller, message)
   end
 
   # Handle unary ops
--- a/lib/elixir/lib/kernel/typespec.ex
+++ b/lib/elixir/lib/kernel/typespec.ex
@@ -535,10 +535,29 @@ defmodule Kernel.Typespec do
   end
 
   # Handle type operator
-  defp typespec({:::, meta, [var, expr]}, vars, caller) do
-    left = typespec(var, [elem(var, 0) | vars], caller)
-    right = typespec(expr, vars, caller)
-    {:ann_type, line(meta), [left, right]}
+  defp typespec({:::, meta, [var, expr]} = orig, vars, caller) do
+    case var do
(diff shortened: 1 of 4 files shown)


lib/elixir/lib/kernel/typespec.ex
```diff
@@ -535,10 +535,29 @@ defmodule Kernel.Typespec do
...
   # Handle type operator
-  defp typespec({:::, meta, [var, expr]}, vars, caller) do
-    left = typespec(var, [elem(var, 0) | vars], caller)
-    right = typespec(expr, vars, caller)
-    {:ann_type, line(meta), [left, right]}
+  defp typespec({:::, meta, [var, expr]} = orig, vars, caller) do
+    case var do
+      {var_name, _meta, context} when is_atom(var_name) and is_atom(context) ->
+        case typespec(expr, vars, caller) do
+          {:ann_type, _, _} ->
```
fertapric: I'm not 100% sure if this error should be raised, but I couldn't think if any valid use case for nested type annotations. 

Besides, I will raise a `CompileError` with typing errors like this one `@type my_type, {pid, name :: atom :: atom}`🙂

Thoughts?

lib/elixir/lib/kernel/typespec.ex
```diff
@@ -535,10 +535,29 @@ defmodule Kernel.Typespec do
...
+            message =
+              "invalid type annotation. Type annotations cannot be nested: " <>
+                "#{Macro.to_string(orig)}"
+
+            compile_error(caller, message)
+
+          right ->
+            left = typespec(var, [elem(var, 0) | vars], caller)
```
josevalim: Since we have already matched on the `var_name`, we can use it here directly, instead of `elem`.

lib/elixir/lib/kernel/typespec.ex
```diff
@@ -535,10 +535,29 @@ defmodule Kernel.Typespec do
   end
 
   # Handle type operator
-  defp typespec({:::, meta, [var, expr]}, vars, caller) do
-    left = typespec(var, [elem(var, 0) | vars], caller)
-    right = typespec(expr, vars, caller)
-    {:ann_type, line(meta), [left, right]}
+  defp typespec({:::, meta, [var, expr]} = orig, vars, caller) do
+    case var do
+      {var_name, _meta, context} when is_atom(var_name) and is_atom(context) ->
```
josevalim: Maybe we can move this matching to the function head and split the current code in two clauses? Other than that, everything looks good!

fertapric:
Feedback done in https://github.com/elixir-lang/elixir/pull/8161/commits/5535c22d948c1f118c3b7515bbad89e921e413a2

josevalim:

PR #8163, opened 2018-09-04
Check for unused, undefined, and redefined types

fertapric:
This PR implements the checks on #5800: 

* `{undefined_type, {TypeName, Arity}}`
* `{unused_type, {TypeName, Arity}}`
* `{redefine_type, {TypeName, Arity}}`

Unused types are removed from the final typespecs form, which, if I recall correctly, will be similar to what's currently being done with unused functions.

I'm still a Erlang newbie 🙇 , any feedback is more than appreciated 🙂

### DIFF
--- a/lib/elixir/src/elixir_erl.erl
+++ b/lib/elixir/src/elixir_erl.erl
@@ -297,16 +297,20 @@ typespecs_form(Map, Set, Bag, MacroNames) ->
       {Types, Specs, Callbacks, MacroCallbacks, OptionalCallbacks} =
         'Elixir.Kernel.Typespec':translate_typespecs_for_module(Set, Bag),
 
+      DefinedTypePairs = collect_defined_type_pairs(Map, Types),
+      UsedTypePairs = collect_used_type_pairs(Map, DefinedTypePairs, Types, Specs, Callbacks, MacroCallbacks),
+      UsedTypes = filter_used_types(Map, UsedTypePairs, Types),
+
       AllCallbacks = Callbacks ++ MacroCallbacks,
       MacroCallbackNames = [NameArity || {_, NameArity, _, _} <- MacroCallbacks],
       validate_behaviour_info_and_attributes(Map, AllCallbacks),
       validate_optional_callbacks(Map, AllCallbacks, OptionalCallbacks),
 
       Forms0 = [],
-      Forms1 = types_form(Types, Forms0),
+      Forms1 = types_form(UsedTypes, Forms0),
       Forms2 = callspecs_form(spec, Specs, [], MacroNames, Forms1, Map),
       Forms3 = callspecs_form(callback, AllCallbacks, OptionalCallbacks, MacroCallbackNames, Forms2, Map),
-      {Types, AllCallbacks, Forms3}
+      {UsedTypes, AllCallbacks, Forms3}
   end.
 
 %% Types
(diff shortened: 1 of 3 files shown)


lib/elixir/src/elixir_erl.erl
```diff
@@ -297,16 +297,20 @@ typespecs_form(Map, Set, Bag, MacroNames) ->
       {Types, Specs, Callbacks, MacroCallbacks, OptionalCallbacks} =
         'Elixir.Kernel.Typespec':translate_typespecs_for_module(Set, Bag),
 
+      DefinedTypePairs = collect_defined_type_pairs(Map, Types),
+      UsedTypePairs = collect_used_type_pairs(Map, DefinedTypePairs, Types, Specs, Callbacks, MacroCallbacks),
+      UsedTypes = filter_used_types(Map, UsedTypePairs, Types),
```
josevalim: Thanks @fertapric! Instead of doing another pass on the types, can't we collect which types are being used as we translate them? So `translate_typespecs_for_module` would return UsedTypes too? This would also reduce the amount of Erlang code we have to write. :)

fertapric: I have other two branches trying to solve this code 😅 

One of them collects the used/unused types during the translation. This was actually my first option, due to the performance benefits it provides, but I found these "issues":

* After the translation, the `caller` env is not available. Errors and warnings require the file and the line, but only the `line` information was there (needed for the `unused_type` warning). I found that I could get the file using `:ets.lookup_element(set, {:cache_env, 0}, 2).file` but I wasn't sure that was the best way to get it. One alternative was to pass the `Map` variable as argument, but I didn't want to change the interface because this is a public function (although it's for internal use only). Other option was to keep the caller when collecting the defined type pairs, something like `%{type_pair => env}`, but I thought the Erlang code was simpler to understand and more intention revealing.
* Collecting the used types during the translation required to pass around some state on any call to `typespec`, and the resulting code was a bit more convoluted. For example, this:

  ```elixir
  defp typespec({:%{}, meta, fields} = map, vars, caller) do
    fields =
      Enum.map(fields, fn
        {k, v} when is_atom(k) ->
          args = [typespec(k, vars, caller), typespec(v, vars, caller)]
          {:type, line(meta), :map_field_exact, args}

        {{:required, meta2, [k]}, v} ->
          args = [typespec(k, vars, caller), typespec(v, vars, caller)]
          {:type, line(meta2), :map_field_exact, args}

        {{:optional, meta2, [k]}, v} ->
          args = [typespec(k, vars, caller), typespec(v, vars, caller)]
          {:type, line(meta2), :map_field_assoc, args}
  ```

  was be changed to:

  ```elixir
  defp typespec({:%{}, meta, fields} = map, vars, caller, types_state) do
    fields =
      Enum.map(fields, fn
        {k, v} when is_atom(k) ->
          {arg1, types_state} = typespec(k, vars, caller, types_state)
          {arg2, types_state} = typespec(v, vars, caller, types_state)
          args = [arg1, arg2]
          {{:type, line(meta), :map_field_exact, args}, types_state}

        {{:required, meta2, [k]}, v} ->
          {arg1, types_state} = typespec(k, vars, caller, types_state)
          {arg2, types_state} = typespec(v, vars, caller, types_state)
          args = [arg1, arg2]
          {{:type, line(meta2), :map_field_exact, args}, types_state}
        
        {{:optional, meta2, [k]}, v} ->
          {arg1, types_state} = typespec(k, vars, caller, types_state)
          {arg2, types_state} = typespec(v, vars, caller, types_state)
          args = [arg1, arg2]
          {{:type, line(meta2), :map_field_assoc, args}, types_state}
  ```
  `typespec/*` functions were doing more than one thing in that case, returning the typespec (its main responsibility) and returning the list of used types.

The other branch does the same thing than the current PR in Elixir, but getting the file using `:ets.lookup_element(set, {:cache_env, 0}, 2).file`.

I'm glad to open a new PR with any of those branches if you think that's the way to go 🙂

lib/elixir/lib/kernel/typespec.ex
```diff
@@ -192,20 +192,65 @@ defmodule Kernel.Typespec do
...
+          if builtin_type?(name, arity) do
+            message = "type #{name}/#{arity} is a builtin type and it cannot be redefined"
+            compile_error(env, message)
+          end
+
+          if Map.has_key?(type_pairs, type_pair) do
+            compile_error(env, "type #{name}/#{arity} is already defined")
+          end
+
+          Map.put(type_pairs, type_pair, env)
```
fertapric: This is only used in `filter_used_types/2`, so I might store `{file, line}` as value instead of the whole `env`, wdyt?

lib/elixir/lib/kernel/typespec.ex
```diff
@@ -192,20 +192,65 @@ defmodule Kernel.Typespec do
 
   @doc false
   def translate_typespecs_for_module(_set, bag) do
-    types = Enum.map(take_typespec(bag, :type), &translate_type/1)
-    specs = Enum.map(take_typespec(bag, :spec), &translate_spec/1)
-    callbacks = Enum.map(take_typespec(bag, :callback), &translate_spec/1)
-    macrocallbacks = Enum.map(take_typespec(bag, :macrocallback), &translate_spec/1)
+    type_typespecs = take_typespec(bag, :type)
+    defined_type_pairs = collect_defined_type_pairs(type_typespecs)
+    state = %{defined_type_pairs: defined_type_pairs, used_type_pairs: []}
```
fertapric: I quite not like the name `state`, but it the best name I came up with 😅 

I also disliked that `:defined_type_pairs` and `:used_type_pairs` have different structures but the same suffixes, but that's the minimum info that is required. I guess not a big deal 🤷‍♂️

lib/elixir/lib/kernel/typespec.ex
josevalim: :+1:

fertapric:
@josevalim I will be off for vacation (currently writing this from the plane ✈️), so feel free to do the changes if you want this to be merged 🙂

PR #8164, opened 2018-09-04
Add note to `Task.yield_many` about output ordering

jordan-brough:
This isn't stated explicitly, but it is true as far as I can tell and there is already [a test that requires ordered output](https://github.com/elixir-lang/elixir/blob/v1.7.3/lib/elixir/test/elixir/task_test.exs#L315-L325).


josevalim:

PR #8167, opened 2018-09-05
Clarify the usage of 'var' keyword in spec guards

fertapric:
Related to https://github.com/elixir-lang/elixir/issues/5800#issuecomment-418693648

### DIFF
--- a/lib/elixir/pages/Typespecs.md
+++ b/lib/elixir/pages/Typespecs.md
@@ -191,7 +191,7 @@ If you want to specify more than one variable, you separate them by a comma.
 
     @spec function(arg1, arg2) :: {arg1, arg2} when arg1: atom, arg2: integer
 
-Type variables with no restriction can also be defined using the `var` keyword.
+Type variables with no restriction can also be defined using `var`.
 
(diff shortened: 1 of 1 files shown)


lib/elixir/pages/Typespecs.md
```diff
@@ -191,7 +191,7 @@ If you want to specify more than one variable, you separate them by a comma.
 
     @spec function(arg1, arg2) :: {arg1, arg2} when arg1: atom, arg2: integer
 
-Type variables with no restriction can also be defined.
+Type variables with no restriction can also be defined using the `var` keyword.
```
josevalim: I don't like the use of `keyword` here. People may either think it is a special language keyword or the key of a keyword list, which is not the case. Maybe we could say "by giving it the pseudo-type `var`" or similar?

fertapric: Yeah. Maybe simplify it to “can also be defined using `var`.”? That way we avoid words like “keyword” and “pseudo-type”?

josevalim: Sounds good to me!

josevalim:

PR #8174, opened 2018-09-08
[WIP] Implement ExUnit max-failures feature

eksperimental:
Implements: #4586
Related: #7480, #8129

Initially it was proposed to call this option "max-fail", but the terminology used in ExUnit is failures,
so I went with "max-failures"

This is a work in progress, I still need to add some more tests related to concurrent tests, but I think it's in a good stage to share it and get feedback.

### DIFF
--- a/lib/ex_unit/lib/ex_unit.ex
+++ b/lib/ex_unit/lib/ex_unit.ex
@@ -64,10 +64,10 @@ defmodule ExUnit do
 
     1. Passed (also represented by `nil`)
     2. Failed
-    3. Skipped (via @tag :skip)
+    3. Skipped (via `@tag :skip`)
     4. Not executed, due to `:max_failures` passed before test completed.
-    5. Excluded (via :exclude filters)
-    6. Invalid (when setup_all fails)
+    5. Excluded (via `:exclude` filters)
+    6. Invalid (when `setup_all/1/2` fail)
 
   """
   @type state ::
--- a/lib/ex_unit/lib/ex_unit.ex
+++ b/lib/ex_unit/lib/ex_unit.ex
@@ -65,12 +65,18 @@ defmodule ExUnit do
     1. Passed (also represented by `nil`)
     2. Failed
     3. Skipped (via @tag :skip)
-    4. Excluded (via :exclude filters)
-    5. Invalid (when setup_all fails)
+    4. Not executed, due to `:max_failures` passed before test completed.
+    5. Excluded (via :exclude filters)
+    6. Invalid (when setup_all fails)
 
   """
   @type state ::
-          nil | {:failed, failed} | {:skipped, binary} | {:excluded, binary} | {:invalid, module}
+          nil
+          | {:failed, failed}
+          | {:skipped, binary}
+          | {:not_executed, binary}
+          | {:excluded, binary}
+          | {:invalid, module}
 
   @typedoc "The error state returned by `ExUnit.Test` and `ExUnit.TestModule`"
   @type failed :: [{Exception.kind(), reason :: term, Exception.stacktrace()}]
@@ -79,6 +85,7 @@ defmodule ExUnit do
   @type suite_result :: %{
           excluded: non_neg_integer,
           failures: non_neg_integer,
+          not_executed: non_neg_integer,
           skipped: non_neg_integer,
           total: non_neg_integer
         }
@@ -181,7 +188,6 @@ defmodule ExUnit do
   @spec start(Keyword.t()) :: :ok
   def start(options \\ []) do
     {:ok, _} = Application.ensure_all_started(:ex_unit)
-
     configure(options)
 
     if Application.fetch_env!(:ex_unit, :autorun) do
@@ -242,6 +248,8 @@ defmodule ExUnit do
     * `:max_cases` - maximum number of tests to run in parallel. Only tests from
       different modules run in parallel. It defaults to `System.schedulers_online * 2`
       to optimize both CPU-bound and IO-bound tests;
+
+    * `:max_failures` - maximum number of allowed test failures, defaults to `:infinity`;
 
     * `:module_load_timeout` - the timeout to be used when loading a test module,
       defaults to `60_000` milliseconds;
--- a/lib/ex_unit/lib/ex_unit.ex
+++ b/lib/ex_unit/lib/ex_unit.ex
@@ -249,7 +249,8 @@ defmodule ExUnit do
       different modules run in parallel. It defaults to `System.schedulers_online * 2`
       to optimize both CPU-bound and IO-bound tests;
 
-    * `:max_failures` - maximum number of allowed test failures, defaults to `:infinity`;
+    * `:max_failures` - the suite stops evaluating tests when this number of test failures
+      is reached. Defaults to `:infinity`;
 
     * `:module_load_timeout` - the timeout to be used when loading a test module,
       defaults to `60_000` milliseconds;
--- a/lib/ex_unit/lib/ex_unit/event_manager.ex
+++ b/lib/ex_unit/lib/ex_unit/event_manager.ex
@@ -2,6 +2,8 @@ defmodule ExUnit.EventManager do
   @moduledoc false
   @timeout 30000
 
+  @type manager :: {supervisor_manager :: pid, event_manager :: pid}
+
   # TODO: Remove support for GenEvent formatters on 2.0
 
   @doc """
@@ -10,6 +12,7 @@ defmodule ExUnit.EventManager do
   This is what power formatters as well as the
   internal statistics server for ExUnit.
   """
+  @spec start_link() :: {:ok, manager}
   def start_link() do
     {:ok, sup} = DynamicSupervisor.start_link(strategy: :one_for_one)
     {:ok, event} = :gen_event.start_link()
--- a/lib/ex_unit/lib/ex_unit/event_manager.ex
+++ b/lib/ex_unit/lib/ex_unit/event_manager.ex
@@ -54,8 +54,8 @@ defmodule ExUnit.EventManager do
 
   def module_started(ref, test_module) do
     # TODO: Remove case_started in Elixir v2.0
-    notify(ref, {:module_started, test_module})
     notify(ref, {:case_started, Map.put(test_module, :__struct__, ExUnit.TestCase)})
+    notify(ref, {:module_started, test_module})
   end
 
   def module_finished(ref, test_module) do
@@ -70,6 +70,10 @@ defmodule ExUnit.EventManager do
 
   def test_finished(ref, test) do
     notify(ref, {:test_finished, test})
+  end
+
+  def max_failures_exceeded(ref) do
+    notify(ref, :max_failures_exceeded)
   end
 
   defp notify({sup, event}, msg) do
--- a/lib/ex_unit/lib/ex_unit/event_manager.ex
+++ b/lib/ex_unit/lib/ex_unit/event_manager.ex
@@ -72,8 +72,8 @@ defmodule ExUnit.EventManager do
     notify(ref, {:test_finished, test})
   end
 
-  def max_failures_exceeded(ref) do
-    notify(ref, :max_failures_exceeded)
+  def max_failures_reached(ref) do
+    notify(ref, :max_failures_reached)
   end
 
   defp notify({sup, event}, msg) do
--- a/lib/ex_unit/lib/ex_unit/on_exit_handler.ex
+++ b/lib/ex_unit/lib/ex_unit/on_exit_handler.ex
@@ -34,7 +34,7 @@ defmodule ExUnit.OnExitHandler do
 
   @spec get_registered_pids(term) :: list()
   def get_registered_pids(manager) do
-    :ets.match(@name, {:"$#{@pid}", manager, :_, :_, :_})
+    :ets.match(@name, {:"$#{@pid}", manager, :_, :_})
     |> List.flatten()
   end
 
@@ -90,6 +90,7 @@ defmodule ExUnit.OnExitHandler do
   @spec get_failure_counter({pid, pid}) :: non_neg_integer()
   def get_failure_counter(manager) do
     [{{:failure_counter, _manager}, counter}] = :ets.lookup(@name, {:failure_counter, manager})
+
     counter
   end
 
--- a/lib/ex_unit/lib/ex_unit/on_exit_handler.ex
+++ b/lib/ex_unit/lib/ex_unit/on_exit_handler.ex
@@ -11,7 +11,7 @@ defmodule ExUnit.OnExitHandler do
 
   use Agent
 
-  @type manager :: ExUnit.EventManager.use(Agent)
+  @type manager :: ExUnit.EventManager.manager()
 
   @spec start_link(keyword()) :: {:ok, pid}
   def start_link(_opts) do
--- a/lib/ex_unit/lib/ex_unit/on_exit_handler.ex
+++ b/lib/ex_unit/lib/ex_unit/on_exit_handler.ex
@@ -2,8 +2,9 @@ defmodule ExUnit.OnExitHandler do
   @moduledoc false
 
   @name __MODULE__
-  @supervisor 2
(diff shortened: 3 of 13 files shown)


eksperimental:
Please if you have a real life scenario where to use it,
it's fully functional (or at least supposed to), and give me some feedback and the logic of this
was more complex than I thought and there may be yet some corner-cases that haven't been covered

Kartstig:
Code looks great. Thank you for working on this!

lib/mix/lib/mix/tasks/test.ex
```diff
@@ -288,6 +290,7 @@ defmodule Mix.Tasks.Test do
     cover: :boolean,
     trace: :boolean,
     max_cases: :integer,
+    max_failures: :string,
```
fertapric: The `--slowest` flag uses `0` instead of `:infinity` to disable the feature, what do you think about taking the same approach here?

eksperimental: @fertapric you have to remember that you will always have one more failure than max-failure.
so `--max-failure 0`  will stop on the first failure.
If we want something like that we should rename the feature to something like "stop-on-nth-failure" (I'm not really good at coming up with new names :-)

hylkealons: agree with @fertapric, it makes more sense to have `--max-failures 1` -> stopping after first failure.

hylkealons:
Great work @eksperimental, I will try to test it and give some feedback. Can't wait to see this merged 😎

lib/mix/lib/mix/tasks/test.ex
eksperimental: I have changed the behaviour to stop when max-failures is reached. Though I haven't changed the default value being `:infinity` until we reach consensus.

eksperimental:
I have added some new commits, the main change is that before we were executing setup and setup_all callbacks every time, but now we follow the rule that if the test is discarded or excluded, the callbacks are never run. So if we have reached the max-failures limit, we don't run any callback

lib/mix/lib/mix/tasks/test.ex
fertapric: I thought about that too, and I also think `:infinity` is better for this feature. Using `0` with `--slowest` makes sense, but using it for disabling `--max-failures` would not be semantically correct. 

However, the behavior of `--max-failures 0` is a bit confusing. I guess I would expect to not run any tests 🤔

eksperimental: I think we could raise if max-failures is 0, since it's probably an undesired behaviour

lib/ex_unit/lib/ex_unit/on_exit_handler.ex
```diff
@@ -2,20 +2,40 @@ defmodule ExUnit.OnExitHandler do
...
 
   @spec start_link(keyword()) :: {:ok, pid}
   def start_link(_opts) do
     Agent.start_link(fn -> :ets.new(@name, @ets_opts) end, name: @name)
   end
 
-  @spec register(pid) :: :ok
-  def register(pid) when is_pid(pid) do
-    :ets.insert(@name, {pid, nil, []})
+  @spec register(pid, term) :: :ok
```
fertapric: Is the `manager` also here `{pid, pid}`? (same for `get_registered_pids /1`)

eksperimental: yes. I guess we could enforce that in the spec

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -323,18 +440,31 @@ defmodule ExUnit.Runner do
...
+          %{test_or_case | state: state}
+      end
+    else
+      test_or_case
     end
   end
 
   ## Helpers
 
+  defp max_failures_reached?(failure_counter, max_failures)
```
fertapric: What do you think about pattern-match on `:infinity` to speed up the happy case?

```elixir
defp max_failures_reached?(_, :infinity), do: false

defp max_failures_reached?(manager, max_failures) when is_tuple(manager) do
  get_failure_counter(manager) >= max_failures
end

defp max_failures_reached?(failures, max_failures), do: failures >= max_failures
```

lib/ex_unit/lib/ex_unit/on_exit_handler.ex
eksperimental: I have added it as a type

lib/ex_unit/lib/ex_unit/runner.ex
eksperimental: I'm not too sure about removing the type checking guards since we are not validating max_failures when the functions is called directly.

lib/mix/lib/mix/tasks/test.ex
```diff
@@ -569,6 +575,34 @@ defmodule Mix.Tasks.Test do
     end
   end
 
+  defp validate_max_failures_opts(opts) do
```
fertapric: What do you think about using `Integer.parse/2` to avoid control-flow via exceptions?

```elixir
defp validate_max_failures_opts(opts) do
  case Keyword.fetch(opts, :max_failures) do
    {:ok, value} when value not in ["infinity", ":infinity"] ->
      case Integer.parse(value) do
        {max_failures, _} when max_failures >= 1 -> Keyword.put(opts, :max_failures, :infinity)
        :error -> raise "--max-failuers : Expected type positive integer or infinity, got '#{value}'"
      end
      
    _ -> 
      Keyword.put(opts, :max_failures, :infinity)
  end
end
```

Additionally, it covers:

* the edge case of `--max-failures 0`
* invalid values like `--max-failures invalid`. Similar to what other flags behave:

    ```
    $ mix test --slowest a
    ** (Mix) Could not invoke task "test": 1 error found!
    --slowest : Expected type integer, got "a"
    ```

lib/ex_unit/lib/ex_unit/runner.ex
fertapric: That's also part of another comment [I was currently writing](https://github.com/elixir-lang/elixir/pull/8174/files#r217171086). If applied, it would ensure that `max_failures` is either `:infinity` or a positive integer 🙂

lib/ex_unit/lib/ex_unit.ex
```diff
@@ -65,12 +65,18 @@ defmodule ExUnit do
     1. Passed (also represented by `nil`)
     2. Failed
     3. Skipped (via @tag :skip)
```
fertapric: Minor nitpick: do you mind adding backticks here? `@tag :skip`, `:exclude`, `setup_all/1,2` 🙂

lib/mix/lib/mix/tasks/test.ex
eksperimental: looks greate. i renamed it `normalize_max_failures`, and I changed a bit the logic of your code to deal with invalid decimal numbers
$ mix test --max-failures 0
** (RuntimeError) --max-failures : Expected type positive integer or infinity, got '0'

lib/mix/lib/mix/tasks/test.ex
```diff
@@ -569,6 +575,22 @@ defmodule Mix.Tasks.Test do
     end
   end
 
+  defp normalize_max_failures(opts) do
+    case Keyword.fetch(opts, :max_failures) do
+      value when value in ["infinity", ":infinity"] ->
```
fertapric: I think this should be `{:ok, value}`

fertapric: 🤔And I think the `:error` value cannot be matched

eksperimental: oops

lib/ex_unit/lib/ex_unit/on_exit_handler.ex
```diff
@@ -2,20 +2,42 @@ defmodule ExUnit.OnExitHandler do
...
   @ets_opts [:public, :named_table, read_concurrency: true, write_concurrency: true]
 
+  # ETS column numbers
+  @pid 1
+  @supervisor 3
+  @on_exit 4
+
   use Agent
 
+  @type manager :: ExUnit.EventManager.use(Agent)
```
fertapric: Should this be `ExUnit.EventManager.manager()`?

eksperimental: fixed. thank you for all your input

lib/mix/lib/mix/tasks/test.ex
```diff
@@ -569,6 +575,22 @@ defmodule Mix.Tasks.Test do
     end
   end
 
+  defp normalize_max_failures(opts) do
+    case Keyword.fetch(opts, :max_failures) do
```
fertapric: 😅 I think you missed the comment related to the `:error`. I think this code will raise `** (CaseClauseError) no case clause matching: :error`

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -299,14 +351,79 @@ defmodule ExUnit.Runner do
...
+    end
+  end
+
+  defp process_failure(%ExUnit.Test{} = test, _options) do
+    {:ok, test}
+  end
+
+  # Stops all running tests in the suite
+  defp terminate_all_running_processes(manager, parent_pid, current_pid) do
+    for registered_pid when registered_pid not in [parent_pid, current_pid] <-
```
fertapric: Minor change for readability:

```elixir
for registered_pid <- ExUnit.OnExitHandler.get_registered_pids(manager),
    registered_pid not in [parent_pid, current_pid],
    Process.alive?(registered_pid) do
  Process.exit(registered_pid, :max_failures_reached)
end
```

lib/mix/lib/mix/tasks/test.ex
fertapric: I think it should work now 👍

lib/ex_unit/lib/ex_unit/runner.ex
eksperimental: proper use of filters. thank you

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -244,48 +263,81 @@ defmodule ExUnit.Runner do
...
+        :timer.tc(fn ->
+          if max_failures_reached?(config.manager, config.max_failures) do
+            %{test | state: {:not_executed, @not_executed_msg}, time: 0}
+          else
             case exec_test_setup(test, context) do
               {:ok, test} ->
                 exec_test(test)
 
               {:error, test} ->
+                # %{test | state: {:invalid, test}}
```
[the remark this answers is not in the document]
eksperimental: no.  I need to investigate why setting it as :invalid it will not work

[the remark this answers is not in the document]
eksperimental: i mean, it will not pass the tests

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -299,14 +351,78 @@ defmodule ExUnit.Runner do
...
+        registered_pid not in [parent_pid, current_pid],
+        Process.alive?(registered_pid) do
+      Process.exit(registered_pid, :max_failures_reached)
+    end
+  end
+
+  defp get_failure_counter(manager),
+    do: ExUnit.OnExitHandler.get_failure_counter(manager)
+
+  defp increment_failure_counter(manager, test, increment \\ 1)
```
fertapric: [`ExUnit.OnExitHandler.increment_failure_counter/2`](https://github.com/elixir-lang/elixir/pull/8174/files#diff-258d820db9fe3b2672410c8255211706R88) also validates the `increment` argument, so we might not need to also check for it here.

This internal function is only called from [here](https://github.com/elixir-lang/elixir/pull/8174/files#diff-1e5179d66aaabef6fbc8efa93ae33493R368) and [here](https://github.com/elixir-lang/elixir/pull/8174/files#diff-1e5179d66aaabef6fbc8efa93ae33493R387). Keeping that in mind,  I think the last clause  `defp increment_failure_counter(manager, _, _increment)` won't be executed; and the `tag` on `ExUnit.TestModule` won't never be `:not_executed` 🙂

I think everything could be simplified to (the `tag` is checked on the parent calls):

```elixir
defp increment_failure_counter(manager, increment \\ 1) do
  ExUnit.OnExitHandler.increment_failure_counter(manager, increment)
end
```

lib/ex_unit/lib/ex_unit/on_exit_handler.ex
```diff
@@ -51,11 +73,29 @@ defmodule ExUnit.OnExitHandler do
...
   @spec run(pid, timeout) :: :ok | {Exception.kind(), term, Exception.stacktrace()}
   def run(pid, timeout) when is_pid(pid) do
-    [{^pid, sup, callbacks}] = :ets.take(@name, pid)
+    [{^pid, _manager, sup, callbacks}] = :ets.take(@name, pid)
     error = terminate_supervisor(sup, timeout)
     exec_on_exit_callbacks(Enum.reverse(callbacks), timeout, error)
   end
 
+  @spec start_failure_counter(manager) :: :ok
+  def start_failure_counter(manager) do
```
fertapric: Does the "failure counter" belong to the `ExUnit.OnExitHandler`, conceptually? (just to double check 🙂)

eksperimental: It doesn't . I thought of that too. But thought creating another ETS for just a counter would be overkill.
let's ask what @josevalim  thinks about it

eksperimental: probably we could implement it using an Agent

lib/ex_unit/lib/ex_unit/event_manager.ex
```diff
@@ -54,8 +57,8 @@ defmodule ExUnit.EventManager do
 
   def module_started(ref, test_module) do
     # TODO: Remove case_started in Elixir v2.0
-    notify(ref, {:module_started, test_module})
     notify(ref, {:case_started, Map.put(test_module, :__struct__, ExUnit.TestCase)})
+    notify(ref, {:module_started, test_module})
```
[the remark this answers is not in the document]
fertapric: I wondered the same. I think it’s for the code to be closer to the comment. And for what I saw, I think the order should not matter :)

josevalim:
Thanks @eksperimental.

I do have some concerns about the implementation. I would like to not have to introduce a new status, not_executed. I believe we can abort the runner immediately and just not care about any subsequent test.

The other thing is that I don't like the runner is using the `ExUnit.OnExitHandler` process to retrieve information about current processes. The runner should have all of this information. The only reason the ExUnit.OnExitHandler process exists is because the test needs to send information to the runner.

Doing those changes should simplify the implementation a lot.

josevalim:
@eksperimental if the user is interested in quick feedback, then I would propose for us to abort as soon as possible. There is no reason to wait for the rest of the test suite to finish loading only to not run any of it. It will be less LOC in the implementation too. :)

eksperimental:
Well i guess we could figure out that number based on the total tests, the failed ones, but we would need to keep track of the passed tests which we are not doing now

eksperimental:
So my approach is to create a exunit.runner.tracker module that will keep track of the test pids and thenl counter in a ets table, similar to how was done in onexit.handler. 
Are you ok with this, or you prefer something different?

lib/ex_unit/lib/ex_unit/event_manager.ex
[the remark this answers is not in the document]
eksperimental: Yes, as fertapric said. The comment is right above the line that needs to be removed

eksperimental:
Or just an Agent

josevalim:
> So my approach is to create a exunit.runner.tracker module that will keep track of the test pids and the failure counter in a ets table, similar to how was done in onexit.handler.

IMO the existing runner should know all of the pids and should be able to track all information about the processes it runs. We should not need a separate state storage. Everything could be kept in the runner state. Maybe by adding one or two new fields.

Issue #8183, opened 2018-09-10
 [error] Can't load module ':lists' that resides in sticky dir


eksperimental:
I'm getting this error message on screen, but no test fails.

    [error] Can't load module ':lists' that resides in sticky dir

It doesn't happen all the time, but you can reproduce it with seed `513799`
the offending line is in file: iex/helpers_test.exs
```elixir
assert nl([node()], :lists) == {:ok, [{:nonode@nohost, :error, :sticky_directory}]}
```
I have tried running inside `capture_io` to no avail.
any idea how to stop the error message from spilling into the output?

```
==> iex (exunit)
...............................................
21:38:56.786 [error] Can't load module ':lists' that resides in sticky dir

...................................................................................................................................................................

Finished in 12.1 seconds (1.4s on load, 10.7s on tests)
210 tests, 0 failures

Randomized with seed 513799
```

eksperimental:
Running
Erlang/OTP 21 [erts-10.0.5]

fertapric:
Related https://github.com/elixir-lang/elixir/pull/8055

fertapric:
The test case `assert nl([node()], :lists) == {:ok, [{:nonode@nohost, :error, :sticky_directory}]}` always produces an error message. That error message should be captured by `ExUnit.CaptureLog.capture_log/2`, but my hypothesis if that the error message arrives too late and the error is printed.

In #8055, I tried moving the error message from the bottom to give some time to the message to arrive. We might move the assertion to the top, but that does not guarantee that the error message won't appear.

eksperimental:
It has happened again @fertapric 

Elixir 1.8.0-dev (de9207895) (compiled with Erlang/OTP 21)

```
==> iex (ex_unit)
................................................
12:08:41.253 [error] Can't load module ':lists' that resides in sticky dir

....................................................................................................................................................................

Finished in 6.4 seconds (1.3s on load, 5.1s on tests)
212 tests, 0 failures

Randomized with seed 967841
```

fertapric:
@eksperimental this message is printed due to a race condition. The PRs reduced the odds, but they did not remove it. 

I have plans of [adding a `logger:flush/0`](https://github.com/elixir-lang/elixir/pull/8186#issuecomment-420315058) function that we can use in our tests, but I've not worked on it yet (but I will!) 🙂

fertapric:
@eksperimental this message is printed due to a race condition. The PRs reduced the odds, but they did not remove it. 

[...]

PR #8186, opened 2018-09-10
Rearrange assertions to (try to) capture error messages

fertapric:
Reduces the chances of #8183. Related to https://github.com/elixir-lang/elixir/pull/8055.

I've investigated a bit more about this issue. If I understood well the code, the Erlang's code server sends the error message (`[error] Can't load module ':lists' that resides in sticky dir`) to the Erlang's `:logger`; the Elixir's `Logger` is added as a handler of the Erlang's `:logger`; and the `ExUnit.CaptureLog` proxy is added as a handler of Elixir's `Logger` when `capture_log/2` is used (and the `:console` backend is removed). 

Thanks to `Logger.flush/0`, we can be sure that all the messages sent to Elixir's `Logger` during the execution of the function are consumed by all the Elixir's `Logger` handlers. However, that function does not have any effect on Erlang's `:logger`, nor `:logger` provides a similar functionality (or at least I haven't found one).

Here is my reaffirmed hypothesis: due to race conditions, the error is printed instead of being captured.

Moving the assertion one line above will give more time to process the messages sent to Erlang's `:logger`.

cc @josevalim @eksperimental

### DIFF
--- a/lib/iex/test/iex/helpers_test.exs
+++ b/lib/iex/test/iex/helpers_test.exs
@@ -1002,8 +1002,8 @@ defmodule IEx.HelpersTest do
   describe "nl" do
     @tag :capture_log
     test "loads a given module on the given nodes" do
-      assert nl(:nonexistent_module) == {:error, :nofile}
       assert nl([node()], :lists) == {:ok, [{:nonode@nohost, :error, :sticky_directory}]}
+      assert nl(:nonexistent_module) == {:error, :nofile}
       assert nl([:nosuchnode@badhost], Enum) == {:ok, [{:nosuchnode@badhost, :badrpc, :nodedown}]}
       assert nl([node()], Enum) == {:ok, [{:nonode@nohost, :loaded, Enum}]}
     end


josevalim:

fertapric:
@josevalim do you think it makes sense to send a proposal to Erlang/OTP so their `logger` supports the "flush" behavior?

`Logger.flush/0` would be then:

```elixir
  @spec flush :: :ok
  def flush do
    _ = Process.whereis(:error_logger) && :gen_event.which_handlers(:error_logger)
    :ok = :logger.flush()
    :gen_event.sync_notify(Logger, :flush)
  end
```

josevalim:
@fertapric I don't think we can. The new logger is completely descentralized. But I don't think it is a logger issue. I would start by checking if the current code loader code guarantees that it logs before it returns.

fertapric:
> I would start by checking if the current code loader code guarantees that it logs before it returns.

It seems to [send the error message](https://github.com/erlang/otp/blob/d6285b0a347b9489ce939511ee9a979acd868f71/lib/kernel/src/code_server.erl#L1436-L1442) to the logger, [before it replies](https://github.com/erlang/otp/blob/d6285b0a347b9489ce939511ee9a979acd868f71/lib/kernel/src/code_server.erl#L1119-L1120).

josevalim:
Oh, so in this case they do have a process where the VM and other processes
may send messages to, although that’s not the way logger does dispatching.

In this case, flush makes sense. How is the logger process implemented? Is
it a GenServer?
-- 


fertapric:
> In this case, flush makes sense. How is the logger process implemented? Is
it a GenServer?

Yeah, [it's a `gen_server`](https://github.com/erlang/otp/blob/master/lib/kernel/src/logger_server.erl#L22).

According to [this code comment](https://github.com/erlang/otp/blob/master/lib/kernel/src/logger_server.erl#L269-L280):

```erlang
%% Interface for those who can't call the API - e.g. the emulator, or
%% places related to code loading.
%%
%% This can also be log events from remote nodes which are sent from
%% logger.erl when the group leader of the client process is on a
%% same node as the client process itself.
handle_info({log,Level,Format,Args,Meta}, State) ->
    logger:log(Level,Format,Args,Meta),
    {noreply, State};
handle_info({log,Level,Report,Meta}, State) ->
    logger:log(Level,Report,Meta),
    {noreply, State};
handle_info(...) ->
    ....
```

josevalim:
@fertapric I think adding a `logger:flush/0` there would be great. Would you send a PR there? I believe we can also simulate it by calling `Process.whereis(:logger) && :sys.resume(:logger)`.

fertapric:
@josevalim Sure! 👍

PR #8188, opened 2018-09-11
Raise errors with duplicated entries on import

### LINKED ISSUE #8187: Confusing error on duplicate import
### Environment

* Elixir & Erlang/OTP versions (elixir --version): Elixir 1.7.1 (compiled with Erlang/OTP 21)
* Operating system: MacOS High Sierra (10.13.6)

### Current behavior

A duplicate function in the `import only:` list fails with a confusing error.

```elixir
defmodule Foo do
  def hi, do: "hi there"
end

defmodule Bar do
  import Foo, only: [hi: 0, hi: 0]
  def greet, do: hi()
end

IO.puts Bar.greet()
```

The problem is that `hi: 0` is repeated in the `import`. But the `CompileError` says `cannot import Foo.hi/0 because it is undefined or private`. This is confusing because `hi: 0` clearly is defined and public.

### Expected behavior

The compiler should give a warning or error that identifies the problem as a duplicate import. Something like `cannot import Foo.hi/0 because it has already been imported`, or `hi/0 is specified multiple times in the same import statement`.

fertapric:
Fixes #8187

### DIFF
--- a/lib/elixir/src/elixir_import.erl
+++ b/lib/elixir/src/elixir_import.erl
@@ -65,7 +65,8 @@ record_warn(Meta, Ref, Opts, Added, E) ->
 calculate(Meta, Key, Opts, Old, File, Existing) ->
   New = case keyfind(only, Opts) of
     {only, Only} when is_list(Only) ->
-      ok = ensure_keyword_list(Meta, File, Only, only),
+      ensure_keyword_list(Meta, File, Only, only),
+      ensure_no_duplicates(Meta, File, Only, only),
       case keyfind(except, Opts) of
         false ->
           ok;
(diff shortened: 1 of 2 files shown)


lib/elixir/src/elixir_import.erl
```diff
@@ -65,7 +65,8 @@ record_warn(Meta, Ref, Opts, Added, E) ->
 calculate(Meta, Key, Opts, Old, File, Existing) ->
   New = case keyfind(only, Opts) of
     {only, Only} when is_list(Only) ->
-      ok = ensure_keyword_list(Meta, File, Only, only),
```
fertapric: I removed this match for symmetry with `ensure_no_duplicates` and `ensure_no_special_form_conflict`; and because it will raise a `CompileError` if the given value is not a keyword list 🙂

josevalim:

nathanl:
Yay, thanks! 😄 👏 👏

PR #8196, opened 2018-09-13
Remove `Guard test tuple_size(...) can never succeed` dialyzer warning on try

### LINKED ISSUE #8157: Guard test tuple_size(...) can never succeed warning
### Environment

```
Erlang/OTP 21 [erts-10.0.5] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [hipe] [dtrace]

Elixir 1.7.2 (compiled with Erlang/OTP 21)
```

* Operating system: Mac OS/X High Sierra 10.13.6

### Current behavior

Dialyzer produces warnings on generated exception-handling code in the `try` macro when a `rescue` clause matches on `ErlangError`. Minimal example:

```elixir
defmodule TryTupleSize do
  def demo do
    try do
      :erlang.error(:badarg)
    rescue
      e in ErlangError ->
        {:ok, e}
    end
  end
end
```

Dialyzer output:

```
try_tuple_size.ex:6: Guard test tuple_size(__@5::#{'__exception__':='true', '__struct__':=_, _=>_}) can never succeed
```

The decompiled-to-Erlang version of the `demo/0` function:

```erlang
demo() ->
    try
        error(badarg)
    catch
        error:__@3:___STACKTRACE__@1 when not is_map(__@3) ->
            _e@1 =
                'Elixir.Exception':normalize(error,
                                             __@3,
                                             case __@3 of
                                                 __@5
                                                     when
                                                         tuple_size(__@5)
                                                         ==
                                                         2
                                                         andalso
                                                         element(1,
                                                                 __@5)
                                                         ==
                                                         badkey;
                                                         __@5 == undef;
                                                         __@5
                                                         ==
                                                         function_clause ->
                                                     ___STACKTRACE__@1;
                                                 _ ->
                                                     []
                                             end),
            {ok,_e@1};
        error:#{'__struct__' := __@4,'__exception__' := true} = __@3:___STACKTRACE__@1
            when __@4 == 'Elixir.ErlangError' ->
            _e@1 =
                'Elixir.Exception':normalize(error,
                                             __@3,
                                             case __@3 of
                                                 __@5
                                                     when
                                                         tuple_size(__@5)
                                                         ==
                                                         2
                                                         andalso
                                                         element(1,
                                                                 __@5)
                                                         ==
                                                         badkey;
                                                         __@5 == undef;
                                                         __@5
                                                         ==
                                                         function_clause ->
                                                     ___STACKTRACE__@1;
                                                 _ ->
                                                     []
                                             end),
            {ok,_e@1}
    end.
```

### Expected behavior

In the case where the (second) generated `catch` branch has already matched on an exception map/struct, it should not use `tuple_size/1` in that clause. Furthermore, since it has already established that the error has been normalized, it shouldn't be trying to normalize the error again. This could remove the type warning from the generated code.

ggcampinho:
Fixes #8157

### DIFF
--- a/lib/elixir/src/elixir_erl_try.erl
+++ b/lib/elixir/src/elixir_erl_try.erl
@@ -73,7 +73,7 @@ dynamic_normalize(Meta, Var, [H | T]) ->
   {'case', Meta, [
     Var,
     [{do, [
-      {'->', Meta, [[{'when', Meta, [Var, Guards]}], {'__STACKTRACE__', Meta, nil}]},
+      {'->', ?generated(Meta), [[{'when', Meta, [Var, Guards]}], {'__STACKTRACE__', Meta, nil}]},
       {'->', Meta, [[{'_', Meta, nil}], []]}
     ]}]
   ]}.
@@ -86,7 +86,7 @@ erl_rescue_stacktrace_for(_Meta, _Var, 'Elixir.ErlangError') ->
   error(badarg);
 erl_rescue_stacktrace_for(Meta, Var, 'Elixir.KeyError') ->
   %% Only the two element tuple requires stacktrace.
-  erl_and(Meta, erl_tuple_size(Meta, Var, 2), erl_record_compare(Meta, Var, badkey));
+  erl_and(Meta, erl_tuple_size(?generated(Meta), Var, 2), erl_record_compare(Meta, Var, badkey));
 erl_rescue_stacktrace_for(Meta, Var, Module) ->
   erl_rescue_guard_for(Meta, Var, Module).
 
--- a/lib/elixir/src/elixir_erl_try.erl
+++ b/lib/elixir/src/elixir_erl_try.erl
@@ -69,14 +69,14 @@ dynamic_normalize(Meta, Var, [H | T]) ->
 
   Guards =
     lists:foldl(fun(Alias, Acc) ->
-      {'when', Meta, [erl_rescue_stacktrace_for(Generated, Var, Alias), Acc]}
+      {'when', Generated, [erl_rescue_stacktrace_for(Generated, Var, Alias), Acc]}
     end, erl_rescue_stacktrace_for(Generated, Var, H), T),
 
-  {'case', Meta, [
+  {'case', Generated, [
     Var,
     [{do, [
-      {'->', Generated, [[{'when', Meta, [Var, Guards]}], {'__STACKTRACE__', Meta, nil}]},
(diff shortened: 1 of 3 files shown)


lib/elixir/src/elixir_erl_try.erl
```diff
@@ -86,7 +86,7 @@ erl_rescue_stacktrace_for(_Meta, _Var, 'Elixir.ErlangError') ->
   error(badarg);
 erl_rescue_stacktrace_for(Meta, Var, 'Elixir.KeyError') ->
   %% Only the two element tuple requires stacktrace.
-  erl_and(Meta, erl_tuple_size(Meta, Var, 2), erl_record_compare(Meta, Var, badkey));
+  erl_and(Meta, erl_tuple_size(?generated(Meta), Var, 2), erl_record_compare(Meta, Var, badkey));
```
ggcampinho: I need to test with the other versions of Erlang, but it seems that only tuple_size that is causing warnings is this one, even with other types of exception

ggcampinho: I tried to add the minimum quantity of generated, but I think we would also make all the metas in `erl_rescue_stacktrace_for` were generated

josevalim: I think that when we call `erl_rescue_stacktrace_for` in the first place, we can already pass a ?generated(Meta), so everything is marked as generated from the beginning, across all clauses. What do you think?

ggcampinho: Agreed

lib/elixir/src/elixir_erl_try.erl
```diff
@@ -65,15 +65,17 @@ normalize_rescue(Meta, Var, Pattern, Expr, ErlangAliases) ->
   prepend_to_block(Meta, {'=', Meta, [Pattern, Normalized]}, Expr).
 
 dynamic_normalize(Meta, Var, [H | T]) ->
+  Generated = ?generated(Meta),
```
josevalim: I would go ahead and apply this to all of the `Meta`s in this function, no?

ggcampinho: I thought we usually apply the minimum, but it's done

josevalim: Since this whole thing is generated, I think it makes sense to mark it all. :)

josevalim:

seancribbs:
This is so awesome, thank you! 💖

PR #8200, opened 2018-09-15
Fix registry doc test

ggcampinho:
This doc test uses the same name as the doc test for match/4 causing a race condition

Spotted at https://travis-ci.org/elixir-lang/elixir/jobs/429080162


josevalim:

PR #8201, opened 2018-09-17
Improve documentation regarding Child Specifications

eksperimental:


lib/elixir/lib/agent.ex
```diff
@@ -140,7 +140,7 @@ defmodule Agent do
   @doc """
   Returns a specification to start an agent under a supervisor.
 
-  See `Supervisor`.
+  See `Supervisor.Spec` for more information.
```
josevalim: `Supervisor.Spec` is deprecated. We shouldn't be linking to it.

lib/elixir/lib/agent.ex
```diff
@@ -78,13 +78,13 @@ defmodule Agent do
     * `:id` - the child specification identifier, defaults to the current module
     * `:start` - how to start the child process (defaults to calling `__MODULE__.start_link/1`)
     * `:restart` - when the child should be restarted, defaults to `:permanent`
-    * `:shutdown` - how to shut down the child
+    * `:shutdown` - the shut down strategy
```
josevalim: Using the word strategy is confusing, because we also have the supervision strategy. I think we can try to be more descriptive here.

lib/elixir/lib/agent.ex
eksperimental: I always get confused by these. I will fix it

lib/elixir/lib/agent.ex
```diff
@@ -78,13 +78,13 @@ defmodule Agent do
...
 
   For example:
 
       use Agent, restart: :transient, shutdown: 10_000
 
-  See the `Supervisor` docs for more information.
+  See the "Child specification" section in the `Supervisor` module for more detailed information..
```
davydog187: double period is not necessary here

eksperimental: Thank you, I had fixed it

lib/elixir/lib/agent.ex
eksperimental: my change is because "how to" sound weird when we are talking about a `timeout`. We need to come up with a way that covers `milliseconds`, `:infinity` and `:brutal_kill`

lib/elixir/lib/gen_server.ex
```diff
@@ -567,13 +567,15 @@ defmodule GenServer do
...
 
-  If part of a supervision tree, a `GenServer`'s `Supervisor` will send an exit
-  signal when shutting it down. The exit signal is based on the shutdown
-  strategy in the child's specification. If it is `:brutal_kill` the `GenServer`
-  is killed and so `c:terminate/2` is not called. However if it is a timeout the
-  `Supervisor` will send the exit signal `:shutdown` and the `GenServer` will
-  have the duration of the timeout to call `c:terminate/2` - if the process is
-  still alive after the timeout it is killed.
+  If part of a supervision tree, a `GenServer`'s [supervisor](`Supervisor`) will send an exit
+  signal when shutting it down. The exit signal is based on the `shutdown`
```
whatyouhide: Why "shutdown" in backticks?

lib/elixir/lib/supervisor/spec.ex
```diff
@@ -97,11 +97,11 @@ defmodule Supervisor.Spec do
       to give the subtree enough time to shutdown; it can also be used with
       workers with care
 
-    * any integer - the value of `:shutdown` can also be any integer meaning
+    * any integer >= 0 - the amount of time in milliseconds
```
whatyouhide: This will not render nicely if without backticks. I'd much rather go with "any non-negative integer".

josevalim:
Thanks @whatyouhide! I still would like to remove the word "shutdown strategy". We can either rollback to what it was before or come with something more descriptive but the term "strategy" here is already overloaded.

whatyouhide:
@josevalim on it.

whatyouhide:
@josevalim done.

josevalim:

eksperimental:
Thank you for taking care of it, as I wasn't able to

PR #8206, opened 2018-09-20
Small optimization in `Stream.Reducers.chunk_every/5`

ggcampinho:
Due to the check `acc_count >= limit` inside `chunk_fun`, `acc_count` can only be greater than `count` if `step` is bigger than `count`, there is no need for this check.


josevalim:

Issue #8205, opened 2018-09-20
Dialyzer errors in Version module


sasa1977:
### Environment

Elixir 1.7.3 and Erlang 21.0.9

### Current behavior

Some functions from `Version` lead to dialyzer errors. For example:

```elixir
def error1(), do: Version.parse!("1.2.3")
def error2(), do: Version.compare("1.2.3", "4.5.6")
```

Dialyzer will report "function has no local return" for both of these functions.

### More info

I did a hacky experiment on a local project by copying over the `Version` source code and renaming it to `Version2`. After some experiments, my impression is that the cause of the error is in the wrong typespec for matchable. In 1.7.3 the spec is `{major :: major(), minor :: minor(), patch :: patch(), pre :: pre()}`, while it should be `{major :: major(), minor :: minor(), patch :: patch(), pre :: pre(), build_parts: [String.t]}`. At least in my hacky version making that fix took care of the dialyzer errors.

I wanted to make a PR, but I've noticed that `matchable` has been removed in #8155. Testing that change on my local hack also took care of dialyzer warnings. Thus, I think that the fix for this boils down to porting #8155 into 1.7 branch.

josevalim:
I don't want to backport it on master because, in the very rare case someone depends on it, i don't want to emit new warnings on a patch release. On v1.8 we will at least have time to test it out and collect feedback during RCs.

sasa1977:
Alternatively, we could fix the spec on 1.7. and then resolve the conflict by merging 1.7 into the master.

In any case, I'm not sure that the issue should be closed without a fix. On 1.7.x using `Version.parse` and `Version.compare` leads to dialyzer errors in the client code, which is definitely a bug. At work, I had to hack around this by using `apply(Version, :parse, [...])` to make our CI tests pass, which is IMO not very pretty.

josevalim:
We will be happy with a typespec fix for v1.7 if somebody sends a PR. :)

sasa1977:
Somebody will do it :-) In the meantime, would you mind reopening this issue, so it can be properly fixed?

josevalim:
@sasa1977 the issue is already fixed in master. There is no guarantee we backport all fixes as those branches have diverged and we don't merge them. If people ask, we usually backport, but since the backport is not possible in this case, somebody has to do the work.

Note also there is no guarantee we will have another v1.7 release before v1.8.

sasa1977:
[...]

PR #8207, opened 2018-09-20
Fix typespec in Version

sasa1977:
Adds a fix for #8205 to 1.7 branch. 

Note that in the master branch the fix is different (typespec has been removed), which means that this commit might cause a conflict between 1.7 branch and the master.


josevalim:

PR #8211, opened 2018-09-21
Clear up `Macro.Env.context` description in the documentation

zoldar:
The way it's phrased now doesn't tell what concrete values can `context` contain.


whatyouhide:
Thanks @zoldar! 💟

PR #8214, opened 2018-09-21
Raise on empty accumulators only on the first time inside `Stream.cycle/1`

### LINKED ISSUE #8209: Stream.cycle constructed with range appears not to repeat through Stream.zip
### Environment

* Elixir & Erlang/OTP versions (elixir --version): 1.7.3
* Operating system: macOS High Sierra 10.13.4

### Current behavior
This line works: 
`Stream.zip(1..255, Stream.cycle(1..5)) |> Enum.take(5)`
and produces this output: 
`[{1, 1}, {2, 2}, {3, 3}, {4, 4}, {5, 5}]`

However, this line: 
`Stream.zip(1..255, Stream.cycle(1..5)) |> Enum.take(6)`
fails, and throws this error:
`(ArgumentError) cannot cycle over empty enumerable`

This line, on the other hand
`Stream.zip(1..255, Stream.cycle([1, 2, 3, 4, 5]) |> Enum.take(6)`
produces the expected output: 
`[{1, 1}, {2, 2}, {3, 3}, {4, 4}, {5, 5}, {6, 1}`

### Expected behavior
We expect that this line of code:
`Stream.zip(1..255, Stream.cycle(1..5)) |> Enum.take(6)`
should have returned: 
`[{1, 1}, {2, 2}, {3, 3}, {4, 4}, {5, 5}, {6, 1}`

It appears that the problem has something to do with the use of a range in `Stream.cycle` / `Stream.zip`, since using a list as the argument to `Stream.cycle` works as expected (as in the third example in the Current Behavior section).

ggcampinho:
`Stream.zip/1` uses empty accumulators to get the next element and this causes an issue

Fixes #8209

### DIFF
--- a/lib/elixir/lib/stream.ex
+++ b/lib/elixir/lib/stream.ex
@@ -1244,7 +1244,8 @@ defmodule Stream do
     fn acc, fun ->
       inner = &do_cycle_each(&1, &2, fun)
       outer = &Enumerable.reduce(enumerable, &1, inner)
-      do_cycle(outer, outer, acc)
+      reduce = check_cycle_first_element(outer)
+      do_cycle(reduce, outer, acc)
     end
   end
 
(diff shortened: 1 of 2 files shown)


whatyouhide:
Thanks @ggcampinho! 💟

Issue #8210, opened 2018-09-20
Better error message when defining the same record twice


wojtekmach:
### Environment

* Elixir & Erlang/OTP versions: Elixir 1.8.0-dev (13ced80fc) (compiled with Erlang/OTP 21)

### Current behavior

I've accidentally defined the same record twice (bad git merge) and got pretty cryptic error. It was   easy enough to see what's wrong as it says "defines defaults multiple times" but perhaps there's a way for something more explicit. Here's a repro:

```elixir
defmodule M do
  import Record
  defrecord :r, [:a]
  defrecord :r, [:a]
end
```

```
** (CompileError) a.ex:4: defmacro r/1 defines defaults multiple times. Elixir allows defaults to be declared once per definition. Instead of:

    def foo(:first_clause, b \\ :default) do ... end
    def foo(:second_clause, b \\ :default) do ... end

one should write:

    def foo(a, b \\ :default)
    def foo(:first_clause, b) do ... end
    def foo(:second_clause, b) do ... end

    a.ex:4: (module)
```

### Expected behavior

Better error message, e.g.: `"record :a is already defined"`

josevalim:
We can use Module.defines? to check that something with the same name does not already exist.

cryic:
Thanks for the starter issue! I'd like to take a first pass at this. I'll respond here if I have any questions.

PR #8215, opened 2018-09-21
Add a check for previously defined records

cryic:
An implementation for #8210 , about a better error message for previously defined Records.

The previous error message can be seen in the linked issue.

The new error message is:
```
== Compilation error in file m.ex ==
** (CompileError) m.ex:5: record :r was previously defined
    m.ex:5: (module)
```

Any feedback is appreciated, especially around potentially pulling the `defines?` check and error message into its own function. The check and CompileError logic feels a little long in the `defrecord`/`defrecordp` functions, however, I couldn't figure out how to extract them since its inside macro definitions.

### DIFF
--- a/lib/elixir/lib/record.ex
+++ b/lib/elixir/lib/record.ex
@@ -255,7 +255,15 @@ defmodule Record do
   """
   defmacro defrecord(name, tag \\ nil, kv) do
     quote bind_quoted: [name: name, tag: tag, kv: kv] do
+      if Module.defines?(__MODULE__, {name, 0}, :defmacro) do
+        raise CompileError,
+          file: __ENV__.file,
+          line: __ENV__.line,
+          description: "record #{inspect(name)} was previously defined"
+      end
+
       tag = tag || name
+
       fields = Record.__fields__(:defrecord, kv)
 
       defmacro unquote(name)(args \\ []) do
@@ -273,7 +281,15 @@ defmodule Record do
   """
   defmacro defrecordp(name, tag \\ nil, kv) do
     quote bind_quoted: [name: name, tag: tag, kv: kv] do
+      if Module.defines?(__MODULE__, {name, 0}, :defmacrop) do
+        raise CompileError,
+          file: __ENV__.file,
+          line: __ENV__.line,
+          description: "record #{inspect(name)} was previously defined"
+      end
+
       tag = tag || name
+
       fields = Record.__fields__(:defrecordp, kv)
 
       defmacrop unquote(name)(args \\ []) do
--- a/lib/elixir/lib/record.ex
+++ b/lib/elixir/lib/record.ex
@@ -255,9 +255,14 @@ defmodule Record do
   """
   defmacro defrecord(name, tag \\ nil, kv) do
     quote bind_quoted: [name: name, tag: tag, kv: kv] do
-      if Module.defines?(__MODULE__, {name, 0}) do
+      defined_arity =
+        Enum.find(0..2, fn arity ->
+          Module.defines?(__MODULE__, {name, arity})
+        end)
+
+      if defined_arity do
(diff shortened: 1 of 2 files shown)


whatyouhide:
Do we want to point to the line of the previous definition?

cryic:
@whatyouhide I could see that being helpful. The more specific we can be, the better. I'll look into this a little bit.

lib/elixir/lib/record.ex
```diff
@@ -255,7 +255,15 @@ defmodule Record do
   """
   defmacro defrecord(name, tag \\ nil, kv) do
     quote bind_quoted: [name: name, tag: tag, kv: kv] do
+      if Module.defines?(__MODULE__, {name, 0}, :defmacro) do
+        raise CompileError,
```
josevalim: We can raise a regular ArgumentError here. :) The stacktrace will be correct without having to resort to CompileError.

cryic: The original error being thrown was a CompileError, so I kept it. Having an ArgumentError makes sense, since we're talking about `name` as an argument to `defrecord`. I'll switch it.

lib/elixir/lib/record.ex
```diff
@@ -255,7 +255,15 @@ defmodule Record do
   """
   defmacro defrecord(name, tag \\ nil, kv) do
     quote bind_quoted: [name: name, tag: tag, kv: kv] do
+      if Module.defines?(__MODULE__, {name, 0}, :defmacro) do
```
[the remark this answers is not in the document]
wojtekmach: I think what José meant was this situation:

```
defmodule M do
  defmacro a() do
  end

  require Record
  Record.defrecord :a, [:a]
end
```

`record :a was previously defined` isn't correct here.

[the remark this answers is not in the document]
josevalim: Yes. Furthermore, @wojtekmach's example will also fail for `def a()`.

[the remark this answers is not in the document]
cryic: Ahh, I see. That is really helpful. Thanks so much for the clarification. I had attempted to define multiple records with the same name, as well as, a function definition, but hadn't done the the defmacro. I'll remove the def_kind option and double-check these scenarios. Thanks again!

cryic:
I've changed the error type and message according to the previous feedback (thanks @josevalim and @wojtekmach).

Should @whatyouhide 's idea be a part of this or should that be a separate PR (if I can get it working)?

lib/elixir/lib/record.ex
```diff
@@ -255,7 +255,13 @@ defmodule Record do
   """
   defmacro defrecord(name, tag \\ nil, kv) do
     quote bind_quoted: [name: name, tag: tag, kv: kv] do
+      if Module.defines?(__MODULE__, {name, 0}) do
```
wojtekmach: this is somewhat pedantic, but defrecord defines 0, 1 and 2 arity defs so perhaps we should handle all of them? For example, if there's a 1-arity def already, we'll get a bad error too:

```
defmodule M do
  defmacro a(_) do
  end

  require Record
  Record.defrecord :a, [:a]
end
```

```
warning: defmacro a/1 has multiple clauses and also declares default values. In such cases, the default values should be defined in a header. Instead of:

    def foo(:first_clause, b \\ :default) do ... end
    def foo(:second_clause, b) do ... end

one should write:

    def foo(a, b \\ :default)
    def foo(:first_clause, b) do ... end
    def foo(:second_clause, b) do ... end

  a.ex:6

warning: this clause cannot match because a previous clause at line 2 always matches
  a.ex:6
```

cryic: Right, this would leave out `name/1` and `/2`. I'll push a try at that.

josevalim:

PR #8216, opened 2018-09-22
Fixed broken link

davidsulc:


josevalim:
@davidsulc great find. I have changed it to use a relative link, so it works across stable/master too. Thanks!

PR #8217, opened 2018-09-22
Move rewrite match to kernel

ggcampinho:
I saw a possibility to move the rewrites made in `:elixir_rewrite.rewrite_match/5` close to their source in `Kernel` simplifying the logic, so I gave it a try. If it does not make sense, feel free to close it.

After this, it's not possible to call directly `:erlang.++/2`, `:erlang.+/1` and `:erlang.-/1`, but this behavior is similar to Erlang.

### DIFF
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -1187,9 +1187,34 @@ defmodule Kernel do
 
   """
   @doc guard: true
-  @spec +value :: value when value: number
-  def +value do
-    :erlang.+(value)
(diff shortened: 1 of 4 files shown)


josevalim:
The issue is that when we rewrite them to macros, then we can no longer
apply them. For example, we can’t call apply(Kernel, :+, [1, 2]). So we
prefer to tackle this in the compiler rather than push optimizations to
user land. :)
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

PR #8219, opened 2018-09-22
Forbid using only the "else" clause in "try"

### LINKED ISSUE #8146: Should the compiler warn about an unmatched `else`?

whatyouhide:
Closes #8146.

Now using `else` without a `rescue`, `catch`, or `after` in `try` expressions is not allowed anymore.

I started with not allowing `else` by itself but I wonder if it would make sense to also disallow `else` if the only other clause is `after`. `else` is there to handle what happens if nothing is caught/rescued, so it might make sense to disallow `else`/`after` as well. Let me know your thoughts!

### DIFF
--- a/lib/elixir/src/elixir_clauses.erl
+++ b/lib/elixir/src/elixir_clauses.erl
@@ -187,7 +187,12 @@ expand_with_else(Meta, Opts, E, HasMatch) ->
 'try'(Meta, [{do, _}], E) ->
   form_error(Meta, ?key(E, file), elixir_expand, {missing_option, 'try', ['catch', 'rescue', 'after', 'else']});
 'try'(Meta, [{do, _}, {else, _}], E) ->
-  form_error(Meta, ?key(E, file), ?MODULE, try_with_only_else_clause);
+  Kind =
+    case lists:keyfind(origin, 1, Meta) of
+      {origin, Origin} -> Origin;
+      false -> 'try'
+    end,
+  form_error(Meta, ?key(E, file), ?MODULE, {try_with_only_else_clause, Kind});
 'try'(Meta, Opts, E) when not is_list(Opts) ->
   form_error(Meta, ?key(E, file), elixir_expand, {invalid_args, 'try'});
 'try'(Meta, Opts, E) ->
@@ -346,9 +351,9 @@ format_error(invalid_rescue_clause) ->
 format_error(catch_before_rescue) ->
   "\"catch\" should always come after \"rescue\" in try";
 
-format_error(try_with_only_else_clause) ->
-  "\"else\" can't be used as the only clause in \"try\" since it's equivalent to not "
-    "having the \"try\" in the first place";
+format_error({try_with_only_else_clause, Kind}) ->
+  io_lib:format("\"else\" can't be used as the only clause in \"~ts\" since it's equivalent to not "
+                "having the \"try\" in the first place", [Kind]);
 
 format_error(unmatchable_else_in_with) ->
   "\"else\" clauses will never match because all patterns in \"with\" will always match";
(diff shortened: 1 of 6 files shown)


fishcakez:
else/after has valid use case, e.g.
```elixir
file = File.open!("file")
try do
   handle_file(file)
else
  result ->
    handle_result(result)
after
  File.close!(file)
end
```

whatyouhide:
@fishcakez isn't that the same as:

```elixir
file = File.open!("file")
try do
   result = handle_file(file)
    handle_result(result)
after
  File.close!(file)
end
```

?

lib/elixir/test/elixir/kernel/expansion_test.exs
```diff
@@ -1653,6 +1657,21 @@ defmodule Kernel.ExpansionTest do
...
+            try do
+              :ok
+            else
+              other -> other
+            end
+          end
+
+        expand(code)
+      end
+    end
```
josevalim: We also need a test for this:

```elixir
def foo do
  :ok
else
  other -> other
end
```

and we need to make sure that we don't say "try" in the error message but rather definition or something similar. If you look at `elixir_def.erl`, you can see that we set the `{origin, _}` attribute in the `try` node, exactly so we know the origin of the try.

whatyouhide: I pushed an update, please have a look cause I'm not sure it's the best way of doing this. FWIW, there are other places where we say something bad happened in `try` but it really happened in `def`:

```
iex(2)> defmodule Foo do
...(2)> def foo() do
...(2)> :ok
...(2)> catch
...(2)> _ -> :catch
...(2)> rescue
...(2)> _ -> :rescue
...(2)> end
...(2)> end
warning: "catch" should always come after "rescue" in try
  iex:3
```

Should we fix them (separate PR)>

josevalim: > Should we fix them (separate PR)?

Up to you!

whatyouhide: Yep, separate PR! :)

lib/elixir/src/elixir_clauses.erl
```diff
@@ -186,6 +186,13 @@ expand_with_else(Meta, Opts, E, HasMatch) ->
   form_error(Meta, ?key(E, file), elixir_expand, {missing_option, 'try', [do]});
 'try'(Meta, [{do, _}], E) ->
   form_error(Meta, ?key(E, file), elixir_expand, {missing_option, 'try', ['catch', 'rescue', 'after', 'else']});
+'try'(Meta, [{do, _}, {else, _}], E) ->
+  Kind =
+    case lists:keyfind(origin, 1, Meta) of
+      {origin, Origin} -> Origin;
+      false -> 'try'
+    end,
```
josevalim: I believe we already do this elsewhere in this file too, maybe we should put this behind a private function? Something like `origin(Meta, 'try')`?

whatyouhide: @josevalim I want to do this as the next PR, where I fix the other similar cases I mentioned. Would that work for you?

PR #8223, opened 2018-09-25
 Optimizes multiple :maps put and merge operations

### LINKED ISSUE #7353: Inline multiple :maps calls into VM operations
We can do it for `:maps.put/3` (always) and for `:maps.merge/2` when the left side is a map.

ggcampinho:
Optimizes `:maps.put/3` to a VM operation instead of a remote call, including nested calls to it.
It does the same for `:maps.merge/2` with a simple map on the right side.

Keys that will be overridden are removed from the operation to avoid warnings.

Closes #7353

### DIFF
--- a/lib/elixir/src/elixir_erl_pass.erl
+++ b/lib/elixir/src/elixir_erl_pass.erl
@@ -578,6 +578,39 @@ translate_remote('Elixir.String.Chars', to_string, Meta, [Arg], S) ->
         {clause, Generated, [Var], [], [Slow]}
       ]}, VS}
   end;
+translate_remote(maps, put, Meta, [Key, Value, Map], S) ->
+  Ann = ?ann(Meta),
+  {TExpr, ES} =
+    case translate_args([Key, Value, Map], S) of
+      {[TKey, TValue, {map, _, InnerMap, Pairs}], TS} ->
+        NewPairs = merge_pairs(Pairs, [{map_field_assoc, Ann, TKey, TValue}]),
+        {{map, Ann, InnerMap, NewPairs}, TS};
+
+      {[TKey, TValue, {map, _, Pairs}], TS} ->
+        NewPairs = merge_pairs(Pairs, [{map_field_assoc, Ann, TKey, TValue}]),
+        {{map, Ann, NewPairs}, TS};
+
+      {[TKey, TValue, TMap], TS} ->
+        {{map, Ann, TMap, [{map_field_assoc, Ann, TKey, TValue}]}, TS}
+    end,
+  {TExpr, mergev(S, ES)};
+translate_remote(maps, merge, Meta, [Map1, Map2], S) ->
+  Ann = ?ann(Meta),
+  {TExpr, ES} =
+    case translate_args([Map1, Map2], S) of
+      {[{map, _, Pairs1}, {map, _, Pairs2}], TS} ->
+        {{map, Ann, merge_pairs(Pairs1, Pairs2)}, TS};
+
+      {[{map, _, InnerMap1, Pairs1}, {map, _, Pairs2}], TS} ->
+        {{map, Ann, InnerMap1, merge_pairs(Pairs1, Pairs2)}, TS};
+
+      {[TMap1, {map, _, Pairs2}], TS} ->
+        {{map, Ann, TMap1, Pairs2}, TS};
+
+      {[TMap1, TMap2], TS} ->
+        {{call, Ann, {remote, Ann, {atom, Ann, maps}, {atom, Ann, merge}}, [TMap1, TMap2]}, TS}
+    end,
+  {TExpr, mergev(S, ES)};
 translate_remote(Left, Right, Meta, Args, S) ->
   {TLeft, SL} = translate(Left, S),
   {TArgs, SA} = translate_args(Args, mergec(S, SL)),
@@ -598,6 +631,26 @@ translate_remote(Left, Right, Meta, Args, S) ->
     false ->
       {{call, Ann, {remote, Ann, TLeft, TRight}, TArgs}, SC}
   end.
+
+merge_pairs(List1, List2) ->
+  IsDuplicatedKey = fun ({map_field_assoc, _, Key1, _}) ->
+    lists:all(fun ({_, _, Key2, _}) ->
+      not will_override(Key1, Key2, true)
+    end, List2)
+  end,
+  lists:filter(IsDuplicatedKey, List1) ++ List2.
+
+will_override(_, _, false) ->
+  false;
+will_override({Atom, _, Arg}, {Atom, _, Arg}, Result) when is_atom(Atom) ->
+  Result;
+will_override({cons, _, Head1, Tail1}, {cons, _, Head2, Tail2}, Result) ->
+  HeadOverride = will_override(Head1, Head2, Result),
+  will_override(Tail1, Tail2, HeadOverride);
+will_override({nil, _}, {nil, _}, Result) ->
+  Result;
+will_override(_, _, _) ->
+  false.
 
 is_always_string({{'.', _, [Module, Function]}, _, Args}) ->
   is_always_string(Module, Function, length(Args));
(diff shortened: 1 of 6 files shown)


lib/elixir/test/erlang/control_test.erl
```diff
@@ -235,3 +235,50 @@ optimized_inspect_interpolation_test() ->
...
+
+optimized_nested_map_put_variable_test() ->
+  {block, _,
+    [_,
+     {map, _, {var, _, _},
+       [{map_field_assoc, _, {atom, _, a}, {integer, _, 1}},
+        {map_field_assoc, _, {atom, _, b}, {integer, _, 2}}]
+     }]
+  } = to_erl("x = %{}; Map.put(Map.put(x, :a, 1), :b, 2)").
+
```
fertapric: What do you think about including this test? It covers nesting + overrides 🙂

```erlang
  {block, _,
    [_,
     {map, _, {var, _, _},
       [{map_field_assoc, _, {atom, _, a}, {integer, _, 3}},
        {map_field_assoc, _, {atom, _, b}, {integer, _, 2}}]
     }]
  } = to_erl("x = %{}; Map.put(Map.put(Map.put(x, :a, 1), :b, 2), :a, 3)").
```

A similar test can also be included for merge

lib/elixir/src/elixir_erl_pass.erl
```diff
@@ -599,6 +632,36 @@ translate_remote(Left, Right, Meta, Args, S) ->
       {{call, Ann, {remote, Ann, TLeft, TRight}, TArgs}, SC}
   end.
 
+merge_pairs(List1, List2) ->
```
josevalim: Is this to avoid warnings? If so, one option is to move the warning to Elixir-land and ignore the warnings that come from the Erlang side of things.

lib/elixir/src/elixir_erl_pass.erl
```diff
@@ -599,6 +632,36 @@ translate_remote(Left, Right, Meta, Args, S) ->
...
+  IsDuplicatedKey = fun ({map_field_assoc, _, Key1, _}) ->
```
fertapric: Pairs could potentially contain other expressions, like `map_field_exact`. Here is a sample code that fails with the current implementation:

```elixir
defmodule Foo do
  defstruct [:a, :b]
end

foo = %Foo{a: 1, b: 2}

Map.merge(%{foo | a: 3}, %{b: 4})
```

Using the code in the current branch, it raises:

```elixir
** (FunctionClauseError) no function clause matching in anonymous fn/1 in :elixir_erl_pass.merge_pairs/2

    The following arguments were given to anonymous fn/1 in :elixir_erl_pass.merge_pairs/2:

        # 1
        {:map_field_exact, 5, {:atom, 0, :a}, {:integer, 0, 3}}

    (stdlib) lists.erl:1286: :lists."-filter/2-lc$^0/1-0-"/2
```

josevalim: Good catch. But ideally we would remove the merging altogether and let the Erlang compiler do it, as they already do it anyway, and we just take care of the obvious ones for warning purposes. It should be easier in Elixir because literals do not have annotations. :)

fertapric: I posted the comment and then I saw yours 😅 Moving the warning to Elixir-lang sounds good to me 👍

lib/elixir/src/elixir_erl_pass.erl
ggcampinho: I moved them, but I'm not sure if I should add tests covering the fact that the maps can have repeated keys

ggcampinho:
One thing that I didn't understand very well here is the OTP versions on the table https://github.com/elixir-lang/elixir/blob/master/lib/elixir/pages/Compatibility%20and%20Deprecations.md#L52

It says that Elixir 1.7 supports Erlang/OTP 19, I guess it was for previous versions because I also noticed that master refuses to compile with Erlang/OTP 19. Should the table be updated?

lib/elixir/src/elixir_map.erl
```diff
@@ -86,16 +86,46 @@ validate_match_key(Meta, List, E) when is_list(List) ->
 validate_match_key(_, _, _) ->
   ok.
 
+validate_not_repeated(Meta, Key, Used, E) ->
```
josevalim: In my opinion, we don't need to worry about this level of complexity. We can just check for exact matches, which will catch literals and keys in the same line, and that's good enough.

ggcampinho:
@fertapric You're right, I forgot that

lib/elixir/src/elixir_map.erl
ggcampinho: It makes sense, I fixed it adding the whole key to the `Used`

josevalim:
@ggcampinho I think we also need tests for `Map.merge(%{a | b: 1}, %{b: 2})` and with the argument swapped. I would add a handful of runtime tests too to `map_test.exs`, making it clear we are testing `Map.merge/2` with literals.

lib/elixir/src/elixir_map.erl
josevalim: I just realized this approach may make this match though: `%{next() => 1, next() => 2}` where `next()` is based on a side-effect. So I do think we need a quick literal check though. Something like:

```erlang
is_literal({_, _, _)) -> false;
is_literal({Left, Right}) -> is_literal(Left) andalso is_literal(Right);
is_literal([_ | _] = List) -> lists:all(fun is_literal/1, List);
is_literal(_) -> true.
```

ggcampinho:
@josevalim the tests with the update syntax in the `control_test.erl`, you mean?

josevalim:
@ggcampinho probably in both. :)

josevalim:
Most common cases of binaries will be literals, so we are fine with it as
literal.
-- 


lib/elixir/test/erlang/control_test.erl
```diff
@@ -235,3 +235,68 @@ optimized_inspect_interpolation_test() ->
...
+  {block, _,
+    [_,
+     {map, _, {var, _, _},
+       [{map_field_exact, _, {atom, _, a}, {integer, _, 2}},
+        {map_field_assoc, _, {atom, _, b}, {integer, _, 3}}]
+     }]
+  } = to_erl("x = %{a: 1}; Map.merge(%{x | a: 2}, %{b: 3})"),
+  {block, _,
+    [_,
+     {call, _, {remote, _, {atom, _, maps}, {atom, _, merge}},
```
ggcampinho: Notice here that this is not an actual optimization because we can't unify them in one unique instruction

lib/elixir/test/elixir/map_test.exs
```diff
@@ -149,6 +149,39 @@ defmodule MapTest do
     end
   end
 
+  test "put/3 optimized by the compiler" do
```
ggcampinho: I also added some extra tests to put/3 because of the optimizations

lib/elixir/test/elixir/kernel/warning_test.exs
```diff
@@ -589,6 +589,21 @@ defmodule Kernel.WarningTest do
...
+    pid = start_supervised!({Agent, fn -> 1 end})
+
+    next = fn ->
+      Agent.get_and_update(pid, fn acc -> {acc, acc + 1} end)
+    end
+
+    output =
+      capture_err(fn ->
+        defmodule NonLiteralDuplicateMapKeys do
+          assert %{next.() => 1, next.() => 2} == %{1 => 1, 2 => 2}
```
josevalim: Nitpick: You can use `System.unique_integer()`, so no need for the agent. Also, you don't have to wrap this in a module, just write the code, without captyre_log and all. If it warns, we will know when we run the suite. :)

So I would do something like this:

    assert map_size(%{System.unique_integer() => 1, System.unique_integer() => 2}) == 2

lib/elixir/test/erlang/control_test.erl
josevalim: Beautiful.

josevalim:
One last comment and we are good to go. :D

lib/elixir/test/elixir/kernel/warning_test.exs
ggcampinho: I didn't like the `Agent` approach, but it was the only idea that I had. I like your idea!

josevalim:

michalmuskala:
I wonder if we could simplify this and avoid doing the key analysis.

The simple way would be to emit just one put operation for each key separately and then implement a merging (if it's not already there) in the Erlang compiler. This should give us a much simpler implementation on the Elixir side and a compiler optimisation that is more powerful and potentially useful in other places.

PR #8226, opened 2018-09-26
Change `args` to `init_arg` in GenServer start_link/3 and init/1

derekkraan:
I tend to get confused when I read the GenServer docs on how the second argument of `start_link/3` gets transmitted to `init/1`. It's especially confusing in the context of the normal flow of child spec -> start_link -> init.

Here is an example:
`Supervisor.start_child(%{id: 1, start: {Foo, :start_link, [1,2,3]}})`

results in the following being called:
`Foo.start_link(1, 2, 3)`

So far so good. The `args` in `{mod, fun, args}` gets unpacked to multiple arguments in the `start_link` function.

The function header for `GenServer.start_link/3` looks like this: `start_link(module, args, options \\ [])`. Here we also see an `args`, but `args` actually represents a singular argument. This is confusing because it makes it unclear how `args` gets sent to `init/1`. With an argument like `args` I am expecting the following behaviour:

```elixir
GenServer.start_link(Foo, [1,2,3], [])
# Foo.init(1,2,3) gets called
```

What really happens: `Foo.init([1,2,3])` gets called.

Calling it `args` creates the expectation that it should be multiple arguments, like in `{mod, fun, args}`. I think we can clarify this by changing the name to be singular (`arg` for example). IMO it can be further improved by indicating its use: `init_arg`.

This is a deviaton from the Erlang docs, but I think it might be a worthwhile one.

### DIFF
--- a/lib/elixir/lib/gen_server.ex
+++ b/lib/elixir/lib/gen_server.ex
@@ -805,7 +805,7 @@ defmodule GenServer do
   This is often used to start the `GenServer` as part of a supervision tree.
 
   Once the server is started, the `c:init/1` function of the given `module` is
-  called with `init_arg` as its arguments to initialize the server. To ensure a
+  called with `init_arg` as its argument to initialize the server. To ensure a
   synchronized start-up procedure, this function does not return until `c:init/1`
   has returned.
 
--- a/lib/elixir/lib/gen_server.ex
+++ b/lib/elixir/lib/gen_server.ex
@@ -368,7 +368,7 @@ defmodule GenServer do
   Invoked when the server is started. `start_link/3` or `start/3` will
   block until it returns.
 
-  `args` is the argument term (second argument) passed to `start_link/3`.
+  `init_arg` is the argument term (second argument) passed to `start_link/3`.
 
   Returning `{:ok, state}` will cause `start_link/3` to return
   `{:ok, pid}` and the process to enter its loop.
@@ -405,7 +405,7 @@ defmodule GenServer do
   `{:error, reason}` and the process to exit with reason `reason` without
   entering the loop or calling `c:terminate/2`.
   """
-  @callback init(args :: term) ::
+  @callback init(init_arg :: term) ::
               {:ok, state}
               | {:ok, state, timeout | :hibernate | {:continue, term}}
               | :ignore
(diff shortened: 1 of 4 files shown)


josevalim:
Thanks @derekkraan! I like this. Can we please do a similar change for Supervisor and DynamicSupervisor too, where relevant?

lib/elixir/lib/gen_server.ex
```diff
@@ -805,7 +805,7 @@ defmodule GenServer do
   This is often used to start the `GenServer` as part of a supervision tree.
 
   Once the server is started, the `c:init/1` function of the given `module` is
-  called with `args` as its arguments to initialize the server. To ensure a
+  called with `init_arg` as its arguments to initialize the server. To ensure a
```
fertapric: Should this be then "as its argument to initialize the server"?

josevalim:
@derekkraan let's avoid changing the code, yes. Especially inner state and things like that. User facing API is cool to change though. :+1:

derekkraan:
@josevalim I've reverted the state change. I'm not 100% sure where the border is between user-facing API and code (there is inevitably some overlap) so I'm open to a further judgment call on that. I've tried to make it super clear that `args` in the state is actually `init_arg` to keep things as consistent as possible.

josevalim:
@derekkraan if the funciton has a `@doc`, then it is public, otherwise we don't have to worry about it. 

The DynamicSupervisor one looks good! I think we just need to change `Supervisor` now and we should be good to go!

derekkraan:
@josevalim Just did a double check, `Supervisor` was looking pretty good already (`arg` in almost all places). I only had to change it at the very end for the `init/1` callback. Unless you meant something else?

josevalim:
@derekkraan oh, the change was so small that I missed it. I see that `Supervisor.start_link/3` is already using `arg` but maybe it should also use `init_arg` for consistency?

derekkraan:
@josevalim Good catch, I've updated that one and any other examples in the `Supervisor` documentation that eventually passed `arg` down to the input of `init/1` to try to keep things consistent. I think (???) this is how it should be. Lots of judgment calls.

whatyouhide:
Thanks a lot @derekkraan! 💟

PR #8228, opened 2018-09-26
Link docs for defdelegate functions

gj:
Associated issue: https://github.com/elixir-lang/ex_doc/issues/902

### DIFF
--- a/lib/elixir/lib/io/ansi/docs.ex
+++ b/lib/elixir/lib/io/ansi/docs.ex
@@ -78,9 +78,8 @@ defmodule IO.ANSI.Docs do
       {key, value}, _printed when is_boolean(value) and key in @metadata_filter ->
         IO.puts([metadata_label(key, options), ' ', to_string(value)])
 
-      {key, value}, _printed when is_tuple(value) and key == :delegate_to ->
-        label = metadata_label(key, options)
-        {m, f, a} = value
+      {:delegate_to, {m, f, a}}, _printed ->
+        label = metadata_label(:delegate_to, options)
         IO.puts([label, ' ', Exception.format_mfa(m, f, a)])
 
       _metadata, printed ->
--- a/lib/elixir/lib/io/ansi/docs.ex
+++ b/lib/elixir/lib/io/ansi/docs.ex
@@ -66,7 +66,7 @@ defmodule IO.ANSI.Docs do
     print_each_metadata(metadata, options) && IO.write("\n")
   end
 
-  @metadata_filter [:delegate_to, :deprecated, :guard, :since]
+  @metadata_filter [:deprecated, :guard, :since]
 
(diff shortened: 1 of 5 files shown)


gj:
I'll push a commit to use `:erlang.length/1` @josevalim

fertapric:
[`length/1`](https://hexdocs.pm/elixir/Kernel.html#length/1) is inlined by the compiler, so `length(as_args)` and `:erlang.length(as_args)` is the same :)

fertapric:
Oh! Right 🙈 Sorry for the noise!

fertapric:
I was wondering if it would make sense to make `delegate_to` a reserved key. However, after looking into it, I discovered we would have the same issues than we did with the `guard` key ([explained here](https://github.com/elixir-lang/elixir/pull/7893#issuecomment-404797818)):

> We have access to the module, but it's going to be a bit more complex than that. Currently we're doing the check in `Module.preprocess_doc_meta/4` which is called from `Module.put_attribute/6`. With this approach we can't distinguish between a `@doc guard: true` call that comes from regular module source and one that is generated by `defguard`.
> 
> I'm thinking of creating an internal (but not private) function `put_doc_metadata` that would effectively replace the `put_attribute` clause that we're currently using now. This way we'll be able to call this from `do_at` for regular `@doc keyword()` calls with a flag that enables validation of reserved keys while we can also call it from other internal places to skip validation (e.g. `defguard`).

As explained above, I think we cannot distinguish between `@doc delegate_to` that comes from `defdelegate` and one that is explicitly included by the user without modifying the current code.

Although, we might [validate the content of `delegate_to`](https://github.com/lackac/elixir/blob/4bf7cbbbf167fea89c62886007867a7a83a1b9a0/lib/elixir/lib/module.ex#L1928):

```elixir
  defp validate_doc_meta(:delegate_to, value) do
    case value do
      {m, f, a} when is_atom(m) and is_atom(f) and is_integer(a) and a >= 0 -> 
        :ok

      _ ->
        raise ArgumentError,
              ":delegate_to is a built-in documentation metadata key. " <>
                "It should be ..., got: #{inspect(value)}"
    end
  end 
```

Thoughts?

gj:
Threw that in just about verbatim, @fertapric 😄 

Do we care that users can set the `:delegate_to` metadatum on non-`defdelegate`-declared functions? Maybe we should make it a reserved key?

I can't think of a use-case for when I would want to override the value produced automatically by `defdelegate`.

josevalim:

josevalim:
We cannot make it a reserved key. If we make it a reserved key, then it means no even ourselves can set it.

The goal behind Elixir is that the language is extensible. The tools available to the language is also available to its users, so it is important that this tool is available to everyone, and not restricted to the language. :)

PR #8231, opened 2018-09-27
Let the undefined function error be performed at the Elixir-land

ggcampinho:
Refers to #5800

### DIFF
--- a/lib/elixir/src/elixir_erl_compiler.erl
+++ b/lib/elixir/src/elixir_erl_compiler.erl
@@ -74,10 +74,6 @@ handle_file_error(File, {Line, Module, Desc}) ->
 
 %% Custom formatting
 
-%% Normalize formatting of functions
-format_error(erl_lint, {undefined_function, {F, A}}) ->
-  io_lib:format("undefined function ~ts/~B", [F, A]);
-
 %% Normalize formatting of specs
 format_error(erl_lint, {spec_fun_undefined, {M, F, A}}) ->
   io_lib:format("spec for undefined function ~ts.~ts/~B", [elixir_aliases:inspect(M), F, A]);
--- a/lib/elixir/test/elixir/kernel/errors_test.exs
+++ b/lib/elixir/test/elixir/kernel/errors_test.exs
@@ -357,7 +357,7 @@ defmodule Kernel.ErrorsTest do
                       end
                       '''
 
-    assert_eval_raise CompileError, ~r"nofile:2: undefined function foo/0", ~C'''
+    assert_eval_raise CompileError, ~r"nofile:2: function foo/0 undefined", ~C'''
     defmodule Kernel.ErrorsTest.ClauseWithDefaults5 do
       def hello(foo, bar \\ foo())
       def hello(foo, bar), do: foo + bar
@@ -384,7 +384,7 @@ defmodule Kernel.ErrorsTest do
   end
 
   test "bad form" do
-    assert_eval_raise CompileError, "nofile:2: undefined function bar/0", '''
+    assert_eval_raise CompileError, "nofile:2: function bar/0 undefined", '''
     defmodule Kernel.ErrorsTest.BadForm do
       def foo, do: bar()
     end
--- a/lib/elixir/test/elixir/kernel/guard_test.exs
+++ b/lib/elixir/test/elixir/kernel/guard_test.exs
@@ -161,7 +161,7 @@ defmodule Kernel.GuardTest do
         end
       end
 
-      assert_raise CompileError, ~r"undefined function is_even/1", fn ->
+      assert_raise CompileError, ~r"function is_even/1 undefined", fn ->
         defmodule IntegerPrivateFunctionUtils do
           import IntegerPrivateGuards
 
--- a/lib/mix/test/mix/tasks/test_test.exs
+++ b/lib/mix/test/mix/tasks/test_test.exs
@@ -238,7 +238,7 @@ defmodule Mix.Tasks.TestTest do
 
       Port.command(port, "\n")
 
-      message = "undefined function error_not_a_var"
+      message = "function error_not_a_var/0 undefined"
       assert receive_until_match(port, message, "") =~ "test/b_test_stale.exs"
 
       File.write!("test/b_test_stale.exs", """


lib/elixir/src/elixir_erl_compiler.erl
```diff
@@ -74,10 +74,6 @@ handle_file_error(File, {Line, Module, Desc}) ->
 
 %% Custom formatting
 
-%% Normalize formatting of functions
-format_error(erl_lint, {undefined_function, {F, A}}) ->
-  io_lib:format("undefined function ~ts/~B", [F, A]);
```
josevalim: I think the error is still coming from erl_lint, no? It just happens that now we are using the default formatting instead of our custom one.

ggcampinho: Sorry, I got confused hahaha

I'll work on it

fertapric: I was also investigating this issue a few weeks ago before going on vacation, and I was planning on working on it today 😅 But don't worry, I didn't go so far. Last thing I remember was reading and understanding the code of `Kernel.LexicalTracker` (which might not even be related to the issue! 😬).

If you don't mind, let's "call dibs" on the issues of #5800 so we don't duplicate work 🙂

josevalim: I don't think it is the Kernel.LexicalTracker. Most likely the Module.LocalsTracker or something similar. :)

fertapric: Sorry, it was `Module.LocalsTracker`. [There is a "lightweight" digraph using ETS](https://github.com/elixir-lang/elixir/blob/master/lib/elixir/lib/module/locals_tracker.ex#L181-L205), and I thought the solution might be there...or not 😅

fertapric: @josevalim you're always faster than me 🙂 (I remembered the name was something with "tracker" 😬 )

ggcampinho: Yeah, I was playing with the `LocalsTracker` but when I saw the tests were failing only on the string I thought the error where dealt somewhere else.  I've pushed a working version now

lib/elixir/lib/module/locals_tracker.ex
```diff
@@ -178,6 +189,17 @@ defmodule Module.LocalsTracker do
     end)
   end
 
+  defp undefined_from(bag, pair, meta, undefined) do
```
ggcampinho: This code is similar to the `reachable_from` and maybe they can be merged to run faster. Although, I'm not sure if it's better to keep them isolated to be easier to reason about it

lib/elixir/lib/module/locals_tracker.ex
```diff
@@ -98,6 +98,17 @@ defmodule Module.LocalsTracker do
...
   end
 
+  def collect_undefined_locals({_set, bag}, all_defined) do
+    undefined =
+      Enum.reduce(all_defined, %{}, fn {pair, _, meta, _}, acc ->
+        undefined_from(bag, pair, meta, acc)
+      end)
+
+    for {pair, {:undefined_function, meta}} <- undefined do
+      {meta, {:undefined_function, pair}}
```
josevalim: Unfortunately I think this meta is not enough. We need to use the meta of the call site and not of the definition. We will probably have to change how things are stored in ETS.

josevalim:
I think you will have to change it so `{:local, {name, arity}}` points to `{{name, arity}, line}` instead of `{name, arity}`.

ggcampinho:
Just the line? I added the full meta

lib/elixir/lib/module/locals_tracker.ex
```diff
@@ -98,6 +98,17 @@ defmodule Module.LocalsTracker do
     {unreachable(reachable, reattached, private), collect_warnings(reachable, private)}
   end
 
+  def collect_undefined_locals({_set, bag}, all_defined) do
+    undefined =
+      Enum.reduce(all_defined, %{}, fn {pair, _, _, _}, acc ->
+        undefined_from(bag, pair, acc)
+      end)
+
+    for {pair, {:undefined_function, meta}} <- undefined do
```
ggcampinho: Should this be sorted by `{fun, arity}` like the others?

ggcampinho: I guess it's not a problem, since only the first error really matter

lib/elixir/lib/module/locals_tracker.ex
```diff
@@ -98,6 +98,20 @@ defmodule Module.LocalsTracker do
     {unreachable(reachable, reattached, private), collect_warnings(reachable, private)}
   end
 
+  @doc """
+  Collect undefined functions based on local calls and existing definitions\][=]
+  """
+  def collect_undefined_locals({_set, bag}, all_defined) do
+    undefined =
+      Enum.reduce(all_defined, %{}, fn {pair, _, _, _}, acc ->
+        undefined_from(bag, pair, acc)
```
josevalim: I think you can simplify this by passing the `set` table to undefined_from and using `:ets.pmember(set, {:def, {name, arity}})` to check if the function exists. This way you dont need a map mixed with ok and undefined function, but you can just build the list directly, something like:

```elixir
for {pair, _, _, _} <- all_defined,
    {local, line} <- out_neighbours(pair),
    not :ets.member(set, local),
    do: {get_meta(line), {:undefined_function, local}}
```

lib/elixir/lib/module/locals_tracker.ex
```diff
@@ -98,6 +98,20 @@ defmodule Module.LocalsTracker do
     {unreachable(reachable, reattached, private), collect_warnings(reachable, private)}
   end
 
+  @doc """
+  Collect undefined functions based on local calls and existing definitions\][=]
```
fertapric: Minor thingy: the `\][=]` at the end

lib/elixir/src/elixir_def.erl
```diff
@@ -61,6 +61,7 @@ fetch_definitions(File, Module) ->
 
   {All, Private} = fetch_definition(Entries, File, Module, Set, Bag, [], []),
   Unreachable = elixir_locals:warn_unused_local(File, Module, All, Private),
```
fertapric: Minor nitpick: what do you think about reordering this to?

```erlang
{All, Private} = fetch_definition(Entries, File, Module, Set, Bag, [], []),
elixir_locals:ensure_no_undefined_local(File, Module, All),
elixir_locals:ensure_no_import_conflict(File, Module, All),
Unreachable = elixir_locals:warn_unused_local(File, Module, All, Private),
{All, Unreachable}.
```

So it fails faster in case of undefined functions and import conflicts, and the variable assignment is closer to its use?

josevalim: I would prefer to leave the errors for last, especially because the unused warning alongside the undefined local error may reveal things such as misplaced arity.

fertapric: Oh! I see. It makes sense.

PR #8232, opened 2018-09-27
Warn on underspecified opaque types

fertapric:
Refers to #5800 

Performs the `erl_lint` check `{underspecified_opaque, {TypeName, Arity}}` in Elixir.

After reading the Erlang/OTP codebase, it seems an "underspecified opaque" is an opaque type of arity 0 with a spec of `term` or `any`. [Here is the implementation in the Erlang/OTP codebase](https://github.com/erlang/otp/blob/master/lib/stdlib/src/erl_lint.erl#L2796-L2798) and [the test suite](https://github.com/erlang/otp/blob/master/lib/stdlib/test/erl_lint_SUITE.erl#L2606-L2613).

### DIFF
--- a/lib/elixir/lib/kernel/typespec.ex
+++ b/lib/elixir/lib/kernel/typespec.ex
@@ -283,6 +283,11 @@ defmodule Kernel.Typespec do
       compile_error(caller, message)
     end
 
+    if kind == :opaque and arity == 0 and underspecified?(spec) do
+      message = "opaque type #{name}/0 is underspecified and therefore meaningless"
+      :elixir_errors.warn(caller.line, caller.file, message)
+    end
+
     {{kind, {name, arity}, caller.line, type, export}, state}
   end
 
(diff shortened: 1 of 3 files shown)


lib/elixir/lib/kernel/typespec.ex
```diff
@@ -283,6 +283,11 @@ defmodule Kernel.Typespec do
       compile_error(caller, message)
     end
 
+    if kind == :opaque and arity == 0 and underspecified?(spec) do
```
josevalim: Maybe we should move the opaque and arity inside the `underspecified?` function?

fertapric: I started with `underspecified?(kind, arity, spec)` but it felt weird that the message have the "opaque" hardcoded. 

But rethinking this part, maybe we can reword the message to include the kind? For example:

```elixir
if underspecified?(kind, arity, spec) do
  message = "@#{kind} type #{name}/#{arity} is underspecified and therefore meaningless"
  :elixir_errors.warn(caller.line, caller.file, message)
end
```

josevalim: LGTM!

whatyouhide:
Thanks @fertapric! 💟

PR #8236, opened 2018-09-29
Improve the docs around async_stream

whatyouhide:


lib/elixir/lib/task.ex
```diff
@@ -481,13 +484,6 @@ defmodule Task do
       iex> Enum.reduce(stream, 0, fn {:ok, num}, acc -> num + acc end)
       47
 
-  If you do not care about the results of the computation, you can run
-  the stream with `Stream.run/1`. Also set `ordered: false`, as you don't
-  care about the order of the results either:
-
-      stream = Task.async_stream(collection, fn item -> ... end, ordered: false)
-      Stream.run(stream)
-
```
josevalim: Why was this section removed? 🤔

whatyouhide: It's the exact same as in the `async_stream/4` docs. I'd rather have most of the docs in `async_stream/4`, and basically point `async_stream/2` to those, like we do for options. If you disagree, I can revert it.

josevalim: I don't think having more examples hurt but I am fine either way, your call!

whatyouhide: Okay, I'll go with removing this then. I want to promote the pattern of having big docs with lots of info in one "central" place, and have other places point to them. In this case I think it doesn't hurt the documentation, that's why I'm fine with removing it :)

PR #8237, opened 2018-09-29
Improve docs for @impl

### LINKED ISSUE #7866: `@impl` documentation is not clear
reading the documentatation for `@impl`

```
  ### `@impl`

  To aid in the correct implementation of behaviours, you may optionally declare
  `@impl` for implemented callbacks of a behaviour. This makes callbacks
  explicit and can help you to catch errors in your code. The compiler will warn
  you if you mark a function as `@impl` when in fact it is not a callback, and
  vice-versa. It also helps with maintainability by making it clear to other
  developers that the function's purpose is to implement a callback.
```

What is the situation that it is exactly trying to describe by **vice-versa**?
does it mean that it will warn you when you don't add @impl to a call back? either case... it is not clear what vice-versa means and I suggest to explicitly mention it instead of using "vice-versa"

Sorry If it's just me, if not please somebody explain and I will submit a PR., thank you.

whatyouhide:
Closes #7866.

### DIFF
--- a/lib/elixir/lib/module.ex
+++ b/lib/elixir/lib/module.ex
@@ -60,9 +60,17 @@ defmodule Module do
   To aid in the correct implementation of behaviours, you may optionally declare
   `@impl` for implemented callbacks of a behaviour. This makes callbacks
   explicit and can help you to catch errors in your code. The compiler will warn
-  you if you mark a function as `@impl` when in fact it is not a callback. It will
-  also warn when you mark a function with `@impl` when it isn't a callback, but only
-  if you marked other callbacks with `@impl`.
+  in these cases:
+
+    * if you mark a function with `@impl` when that function is not a callback.
+
+    * if you don't mark a function with `@impl` when other functions are marked
+      with `@impl`. If you mark one function with `@impl`, you must mark all
+      other callbacks for that behaviour as `@impl`.
+
+  `@impl` works on a per-context basis. If you generate a function through a macro
+  and mark it with `@impl`, that won't affect the module where that function is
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/module.ex
```diff
@@ -60,11 +60,14 @@ defmodule Module do
   To aid in the correct implementation of behaviours, you may optionally declare
   `@impl` for implemented callbacks of a behaviour. This makes callbacks
   explicit and can help you to catch errors in your code. The compiler will warn
-  you if you mark a function as `@impl` when in fact it is not a callback, and
-  vice-versa. It also helps with maintainability by making it clear to other
-  developers that the function's purpose is to implement a callback.
+  you if you mark a function as `@impl` when in fact it is not a callback. It will
+  also warn when you mark a function with `@impl` when it isn't a callback, but only
+  if you marked other callbacks with `@impl`.
```
fertapric: What do you think about clarifying that:

* if you mark a function with `@impl`, you must mark all other functions of that behavior with `@impl`. Otherwise, you get: 
  ```
  warning: module attribute @impl was not set for function bar/0 callback (specified in Foo). 
  This either means you forgot to add the "@impl true" annotation before the definition or that you are accidentally overriding this callback
  ```
* `@impl` is applied per context. If you mark functions generated by a macro with `@impl`, that won't affect the functions in the module you `use/2` that macro.

whatyouhide: Good call. I rephrased your suggestion but added it in the latest commit.

Issue #7786, opened 2018-06-22
MatchError in Task.Supervisor.async/1 in case of custom max_children option 


van-mronov:
### Environment

* Elixir & Erlang/OTP versions (elixir --version):  
Erlang/OTP 21 [erts-10.0] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [hipe] [dtrace]

Elixir 1.7.0-dev (c3b0caa8c) (compiled with Erlang/OTP 21)

* Operating system: 
macOS High Sierra

### Current behavior

```elixir
iex(1)> fun = fn -> receive do: (true -> true) end
#Function<20.127694169/0 in :erl_eval.expr/5>
iex(2)> {:ok, pid} = Task.Supervisor.start_link(max_children: 2)
{:ok, #PID<0.107.0>}
iex(3)> Task.Supervisor.async(pid, fun)
%Task{
  owner: #PID<0.104.0>,
  pid: #PID<0.109.0>,
  ref: #Reference<0.1607025971.2254700546.50905>
}
iex(4)> Task.Supervisor.async(pid, fun)
%Task{
  owner: #PID<0.104.0>,
  pid: #PID<0.111.0>,
  ref: #Reference<0.1607025971.2254700546.50934>
}
iex(5)> Task.Supervisor.async(pid, fun)
** (MatchError) no match of right hand side value: {:error, :max_children}
    (elixir) lib/task/supervisor.ex:376: Task.Supervisor.async/6
```

### Expected behavior

`Task.Supervisor.async/1` and other calls to start async task should return `{:error, :max_children}` if `DynamicSupervisor` cannot start child.

josevalim:
Good catch. Not sure about the behaviour here. We can't return `{:error, :max_children}` because we just return the task, so that would be confusing. So we may have to raise but something more meaningful.

van-mronov:
Yeah, I see. Also we have the same issue with `async_stream` calls. And, perhaps, raising error in spawn function makes sense.

josevalim:
`async_stream` allows us to return errors, so it is not a big problem there.

van-mronov:
So, what if add something like `MaxChildrenError`?

josevalim:
@van-mronov we generally do not add new exceptions for cases like this. But it may be worth if it can be raised on async and used as a signal on async_stream.

michalmuskala:
[...]

PR #8238, opened 2018-09-29
Raise an error in Task.Supervisor.async when reaching max_children

whatyouhide:
Related to #7786.

This PR raises when the max number of children for a `Task.Supervisor` is reached when starting a task with `Task.Supervisor.async/2,4` or `Task.Supervisor.async_nolink/2,4`. I am not handling anything related to `async_stream` but this should be a start.

The big diff in tests is related to moving things around in order to be able to use `describe` blocks. I just added one test for `async/2` and one for `async_nolink/2`.

### DIFF
--- a/lib/elixir/lib/task/supervisor.ex
+++ b/lib/elixir/lib/task/supervisor.ex
@@ -101,6 +101,9 @@ defmodule Task.Supervisor do
   The task will still be linked to the caller, see `Task.async/3` for
   more information and `async_nolink/2` for a non-linked variant.
 
+  Raises an error if `supervisor` has reached the maximum number of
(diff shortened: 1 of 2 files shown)


whatyouhide:
@josevalim should we leave `async_stream` alone, and just document that `:max_concurrency` should be <= than `:max_children`? What if I use a supervisor with `max_children: 10`, `async_stream` with `max_concurrency: 8`, but then spawn other things under the supervisor?

josevalim:
I think we should shutdown the async stream and have it raise.

whatyouhide:
@josevalim awesome, I'll work on that.

PR #8239, opened 2018-09-30
Raise when reaching max chidren in Task.Supervisor.async_stream

### LINKED ISSUE #7786: MatchError in Task.Supervisor.async/1 in case of custom max_children option

whatyouhide:
Closes #7786.

### DIFF
--- a/lib/elixir/lib/task/supervised.ex
+++ b/lib/elixir/lib/task/supervised.ex
@@ -412,16 +412,15 @@ defmodule Task.Supervised do
         Map.put(waiting, spawned, {pid, :running})
 
       {:max_children, ^monitor_ref} ->
-        stream_cleanup_inbox(monitor_pid, monitor_ref)
-        Process.demonitor(monitor_ref, [:flush])
+        stream_close(monitor_pid, monitor_ref, timeout)
 
         raise """
         reached the maximum number of tasks for this task supervisor. The maximum number \
         of tasks that are allowed to run at the same time under this supervisor can be \
         configured with the :max_children option passed to Task.Supervisor.start_link/1. When \
-        using async_stream, make sure to configure :max_concurrency to be lower or equal to \
-        :max_children and pay attention to whether other tasks are also spawned under the same \
-        task supervisor.\
+        using async_stream or async_stream_nolink, make sure to configure :max_concurrency to \
+        be lower or equal to :max_children and pay attention to whether other tasks are also \
+        spawned under the same task supervisor.\
         """
 
       {:DOWN, ^monitor_ref, _, ^monitor_pid, reason} ->
@@ -495,10 +494,8 @@ defmodule Task.Supervised do
             stream_monitor_loop(running_tasks, config)
 
           {:error, :max_children} ->
-            Process.flag(:trap_exit, true)
-            kill_all_running_tasks(running_tasks)
             send(parent_pid, {:max_children, monitor_ref})
-            exit(:normal)
+            stream_waiting_for_stop_loop(running_tasks, config)
         end
 
       # The parent process is telling us to stop because the stream is being
@@ -552,6 +549,36 @@ defmodule Task.Supervised do
 
       {:EXIT, _, _} ->
         stream_monitor_loop(running_tasks, config)
+    end
+  end
+
+  defp stream_waiting_for_stop_loop(running_tasks, config) do
+    %{
+      parent_ref: parent_ref,
+      monitor_ref: monitor_ref
+    } = config
+
+    receive do
+      # The parent process is telling us to stop because the stream is being
+      # closed. In this case, we forcibly kill all spawned processes and then
+      # exit gracefully ourselves.
+      {:stop, ^monitor_ref} ->
+        Process.flag(:trap_exit, true)
+        kill_all_running_tasks(running_tasks)
+        exit(:normal)
+
+      # The parent process went down with a given reason. We kill all the
+      # spawned processes (that are also linked) with the same reason, and then
+      # exit ourselves with the same reason.
+      {:DOWN, ^parent_ref, _, _, reason} ->
+        for {_ref, %{type: :link, pid: pid}} <- running_tasks do
+          Process.exit(pid, reason)
+        end
+
+        exit(reason)
+
+      _other ->
+        stream_waiting_for_stop_loop(running_tasks, config)
     end
   end
 
--- a/lib/elixir/lib/task/supervised.ex
+++ b/lib/elixir/lib/task/supervised.ex
@@ -498,24 +498,6 @@ defmodule Task.Supervised do
             stream_waiting_for_stop_loop(running_tasks, config)
         end
 
-      # The parent process is telling us to stop because the stream is being
-      # closed. In this case, we forcibly kill all spawned processes and then
-      # exit gracefully ourselves.
-      {:stop, ^monitor_ref} ->
-        Process.flag(:trap_exit, true)
-        kill_all_running_tasks(running_tasks)
-        exit(:normal)
-
-      # The parent process went down with a given reason. We kill all the
-      # spawned processes (that are also linked) with the same reason, and then
-      # exit ourselves with the same reason.
-      {:DOWN, ^parent_ref, _, _, reason} ->
-        for {_ref, %{type: :link, pid: pid}} <- running_tasks do
-          Process.exit(pid, reason)
-        end
-
-        exit(reason)
-
       # One of the spawned processes went down. We inform the parent process of
       # this and keep going.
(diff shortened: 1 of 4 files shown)


lib/elixir/lib/task/supervised.ex
```diff
@@ -411,6 +411,19 @@ defmodule Task.Supervised do
         send(pid, {self(), {monitor_ref, spawned}})
         Map.put(waiting, spawned, {pid, :running})
 
+      {:max_children, ^monitor_ref} ->
+        stream_cleanup_inbox(monitor_pid, monitor_ref)
+        Process.demonitor(monitor_ref, [:flush])
```
whatyouhide: This is the thing I'm the least sure about. Basically, if the monitor process can't spawn a task because `:max_children`, it will kill all other running tasks and send `{:max_children, monitor_ref}` to the parent, then it will `exit(:normal)`. The only other cases it `exit(:normal)`s is when the parent process tells it to stop, but here we can't use `stream_close` (and tell the monitor process to stop) because we need to stop all processes and the monitor process right away. Demonitoring here works but I'm not sure it's what we want since we don't do it anywhere else. Definitely open to better solutions.

lib/elixir/lib/task/supervised.ex
```diff
@@ -411,6 +411,19 @@ defmodule Task.Supervised do
...
+
+        raise """
+        reached the maximum number of tasks for this task supervisor. The maximum number \
+        of tasks that are allowed to run at the same time under this supervisor can be \
+        configured with the :max_children option passed to Task.Supervisor.start_link/1. When \
+        using async_stream, make sure to configure :max_concurrency to be lower or equal to \
```
fertapric: Minor nitpick: should we mention `async_stream_nolink` here?

lib/elixir/lib/task/supervised.ex
whatyouhide: Sounds like a better solution yes.

whatyouhide: @josevalim pushed a commit that fixes this. I think I did what you suggested. There's a bit of code repetition but it should work. Thoughts?

lib/elixir/lib/task/supervised.ex
```diff
@@ -457,43 +469,41 @@ defmodule Task.Supervised do
...
+                timed_out?: false
+              })
 
-        running_tasks = Map.put(running_tasks, ref, task_info)
-        stream_monitor_loop(running_tasks, config)
+            stream_monitor_loop(running_tasks, config)
+
+          {:error, :max_children} ->
+            send(parent_pid, {:max_children, monitor_ref})
+            stream_waiting_for_stop_loop(running_tasks, config)
```
josevalim: You can remove the duplication by adding a `catch all` to this receive, something like:

```
  ...
  other ->
    handle_stop_or_exit(other, running_tasks, config)
    stream_monitor_loop(running_tasks, config)
```

And then you just need to create a `stream_stop_loop` that reuses the same logic.

whatyouhide: Okay. I am not a huge fan of this solution because 1. it complicates the loop by adding indirection since now you don't see all possible messages and 2. it implies that `handle_stop_or_exit` exits the monitor process. But it works. I pushed a commit, if you like it we can get that in in the meantime and I can try and think of a cleaner solution.

lib/elixir/lib/task/supervised.ex
```diff
@@ -539,9 +531,55 @@ defmodule Task.Supervised do
...
+
+      other ->
+        handle_stop_or_parent_down(other, running_tasks, config)
+        stream_monitor_loop(running_tasks, config)
     end
   end
 
+  defp stream_waiting_for_stop_loop(running_tasks, config) do
+    %{
+      parent_ref: parent_ref,
```
fertapric: I think these are no longer used

lib/elixir/lib/task/supervised.ex
```diff
@@ -539,9 +531,55 @@ defmodule Task.Supervised do
...
+      monitor_ref: monitor_ref
+    } = config
+
+    receive do
+      message ->
```
fertapric: Just to double check: this was matching before https://github.com/elixir-lang/elixir/pull/8239/commits/bbef09885a0ea8ba0e12fe82786373afffb68fce on:

* `{:stop, ^monitor_ref}`
* `{:DOWN, ^parent_ref, _, _, reason}`
* `_other` (match all)

However, I think it now only matches on:

* `{:stop, ^monitor_ref}` (the first clause of `handle_stop_or_parent_down`) 
* `{:DOWN, ^parent_ref, _, _, reason}` (the second clause of `handle_stop_or_parent_down`)

The previous match-all is gone. Could that produce undesired `FunctionClauseError`s?

josevalim: It will lead to undefined FunctionClauseError. We need to ignore the message in case it is not the one we are looking for.

lib/elixir/lib/task/supervised.ex
```diff
@@ -539,9 +531,55 @@ defmodule Task.Supervised do
...
+        {:DOWN, ^ref, _, _, _} -> :ok
+      end
+    end
+
+    exit(:normal)
+  end
+
+  defp handle_stop_or_parent_down({:DOWN, parent_ref, _, _, reason}, running_tasks, %{
+         parent_ref: parent_ref
+       }) do
```
josevalim: We rarely use this pattern in Elixir, can we please keep all arguments in a single line or one per line? Thanks. Same for the clauses above and then tests.

lib/elixir/lib/task/supervised.ex
whatyouhide: The problem with this is that before the `stream_monitor_loop` didn't accept random messages as the catch-all clause was `{:EXIT, ...}`. Not sure how to approach this.

josevalim: You can make `handle_stop_or_parent_down` return true (if handled) or false (if unknown) and raise accordingly. `handle_stop... || raise "unexpected message #{inspect other}"`.

whatyouhide: Yeah that makes sense. I'm now raising in `stream_monitor_loop` but not in `stream_waiting_for_stop_loop`. Should work :)

lib/elixir/lib/task/supervised.ex
```diff
@@ -539,9 +530,62 @@ defmodule Task.Supervised do
 
       {:EXIT, _, _} ->
         stream_monitor_loop(running_tasks, config)
+
+      other ->
+        handle_stop_or_parent_down(other, running_tasks, config) ||
+          raise "unexpected message: #{inspect(other)}"
```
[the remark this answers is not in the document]
whatyouhide: Well we weren't handling those before either, so might be something for another PR?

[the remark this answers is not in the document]
josevalim: Before they would sit in the process inbox, now they will cause a crash.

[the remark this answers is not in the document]
josevalim: I think it is best if we revert the last commit and simply discard messages.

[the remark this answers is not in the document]
whatyouhide: Argh, that's correct. Reverted.

PR #8240, opened 2018-09-30
Exclude deprecated functions in hints

qcam:
I'd like to propose excluding deprecated functions in "did you mean" message.

If a function is deprecated, it's useless to suggest it in the hints, since the compiler would eventually show some warnings if the users go with the deprecated suggestion.


whatyouhide:
Great idea @qcam! Thanks 💟

PR #8241, opened 2018-10-01
Bypass erl_lint and spawn compiler processes manually

josevalim:
This speeds up the Elixir compiler from 5-10% as we no
longer do duplicate work on erl_lint and expand_records.
We also avoid the copying of messages by controlling
exactly when the compiler spawns new processes.


josevalim:

josevalim:
The dialyzer option is not passed to v3_core, as we call it by hand, so we are safe.

PR #8247, opened 2018-10-02
Version 3: Implement ExUnit max-failures feature

eksperimental:
I decided to submit a new PR and keep the previous PR #8174

Implements: #4586
Related: #8174, #7480, #8129

/cc @ibizaman, @hylkealons, @Kartstig, @stevegraham, @DenisGorbachev, @fertapric

### DIFF
--- a/lib/ex_unit/lib/ex_unit.ex
+++ b/lib/ex_unit/lib/ex_unit.ex
@@ -62,11 +62,11 @@ defmodule ExUnit do
 
   A finished test can be in one of five states:
 
-    1. Passed (also represented by `nil`)
+    1. Passed (represented by `nil`)
     2. Failed
-    3. Skipped (via @tag :skip)
-    4. Excluded (via :exclude filters)
-    5. Invalid (when setup_all fails)
+    3. Skipped (via `@tag :skip`)
+    4. Excluded (via `:exclude` filters)
+    5. Invalid (when `setup_all/1,2` fail)
 
   """
   @type state ::
@@ -242,6 +242,10 @@ defmodule ExUnit do
     * `:max_cases` - maximum number of tests to run in parallel. Only tests from
       different modules run in parallel. It defaults to `System.schedulers_online * 2`
       to optimize both CPU-bound and IO-bound tests;
+
+    * `:max_failures` - the suite stops evaluating tests when this number of test failures
+      is reached. All tests within a module that fails when using the `setup_all/1,2` callbacks
+      are counted as failures. Defaults to `:infinity`;
 
     * `:module_load_timeout` - the timeout to be used when loading a test module,
       defaults to `60_000` milliseconds;
--- a/lib/ex_unit/lib/ex_unit.ex
+++ b/lib/ex_unit/lib/ex_unit.ex
@@ -244,7 +244,7 @@ defmodule ExUnit do
       to optimize both CPU-bound and IO-bound tests;
 
     * `:max_failures` - the suite stops evaluating tests when this number of test failures
-      is reached. All tests within a module that fails when using the `setup_all/1,2` callbacks
+      is reached. All tests within a module that fail when using the `setup_all/1,2` callbacks
       are counted as failures. Defaults to `:infinity`;
 
     * `:module_load_timeout` - the timeout to be used when loading a test module,
--- a/lib/ex_unit/lib/ex_unit.ex
+++ b/lib/ex_unit/lib/ex_unit.ex
@@ -62,11 +62,11 @@ defmodule ExUnit do
 
   A finished test can be in one of five states:
 
-    1. Passed (represented by `nil`)
+    1. Passed (also represented by `nil`)
     2. Failed
-    3. Skipped (via `@tag :skip`)
-    4. Excluded (via `:exclude` filters)
-    5. Invalid (when `setup_all/1,2` fail)
+    3. Skipped (via @tag :skip)
+    4. Excluded (via :exclude filters)
+    5. Invalid (when setup_all fails)
 
   """
   @type state ::
--- a/lib/ex_unit/lib/ex_unit/cli_formatter.ex
+++ b/lib/ex_unit/lib/ex_unit/cli_formatter.ex
@@ -153,8 +153,29 @@ defmodule ExUnit.CLIFormatter do
     test_counter =
       Enum.reduce(test_module.tests, config.test_counter, &update_test_counter(&2, &1))
 
-    failure_counter = config.failure_counter + tests_length
-    config = %{config | test_counter: test_counter, failure_counter: failure_counter}
+    config = %{
+      config
+      | test_counter: test_counter,
+        failure_counter: config.failure_counter + tests_length
+    }
+
+    {:noreply, config}
+  end
+
+  def handle_cast(:max_failures_reached, config) do
+    if config.trace() do
+      "\n  --max-failures reached"
+      |> failure(config)
+      |> IO.puts()
+    end
+
+    {:noreply, config}
+  end
+
+  def handle_cast(:aborting_max_failures_reached, config) do
+    "\n--max-failures reached, aborting test suite"
+    |> failure(config)
+    |> IO.puts()
 
     {:noreply, config}
   end
--- a/lib/ex_unit/lib/ex_unit/cli_formatter.ex
+++ b/lib/ex_unit/lib/ex_unit/cli_formatter.ex
@@ -153,11 +153,8 @@ defmodule ExUnit.CLIFormatter do
     test_counter =
       Enum.reduce(test_module.tests, config.test_counter, &update_test_counter(&2, &1))
 
-    config = %{
-      config
-      | test_counter: test_counter,
-        failure_counter: config.failure_counter + tests_length
-    }
+    failure_counter = config.failure_counter + tests_length
+    config = %{config | test_counter: test_counter, failure_counter: failure_counter}
 
     {:noreply, config}
   end
--- a/lib/ex_unit/lib/ex_unit/runner.ex
+++ b/lib/ex_unit/lib/ex_unit/runner.ex
@@ -2,13 +2,14 @@ defmodule ExUnit.Runner do
   @moduledoc false
 
   alias ExUnit.EventManager, as: EM
+  alias ExUnit.RunnerStats
 
   @rand_algorithm :exs1024
 
   def run(opts, load_us) when (is_integer(load_us) or is_nil(load_us)) and is_list(opts) do
     opts = normalize_opts(opts)
     {:ok, manager} = EM.start_link()
-    {:ok, stats_pid} = EM.add_handler(manager, ExUnit.RunnerStats, opts)
+    {:ok, stats_pid} = EM.add_handler(manager, RunnerStats, opts)
     config = configure(opts, manager, self(), stats_pid)
 
     :erlang.system_flag(:backtrace_depth, Keyword.fetch!(opts, :stacktrace_depth))
@@ -19,9 +20,15 @@ defmodule ExUnit.Runner do
         loop(config, :async, 0)
       end)
 
+    if max_failures_reached?(config.stats_pid, config.max_failures) do
+      EM.notify(config.manager, :aborting_max_failures_reached)
+    end
+
     EM.suite_finished(config.manager, run_us, load_us)
-    stats = ExUnit.RunnerStats.stats(stats_pid)
+    stats = RunnerStats.stats(stats_pid)
+
     EM.stop(config.manager)
+
     after_suite_callbacks = Application.fetch_env!(:ex_unit, :after_suite)
     Enum.each(after_suite_callbacks, fn callback -> callback.(stats) end)
     stats
@@ -36,6 +43,7 @@ defmodule ExUnit.Runner do
       include: opts[:include],
       manager: manager,
       max_cases: opts[:max_cases],
+      max_failures: opts[:max_failures],
       only_test_ids: opts[:only_test_ids],
       runner_pid: runner_pid,
       seed: opts[:seed],
@@ -82,7 +90,7 @@ defmodule ExUnit.Runner do
 
       # No more modules, we are done!
       [] ->
-        config
+        :ok
     end
   end
 
@@ -91,8 +99,11 @@ defmodule ExUnit.Runner do
   # counter and attempt to spawn new ones.
   defp wait_until_available(config, modules, taken) do
     receive do
-      {_pid, :module_finished, _test_case} ->
+      {_pid, :module_finished} ->
         loop(config, modules, taken - 1)
+
+      {_pid, :max_failures_reached} ->
+        {:error, :max_failures_reached}
     end
   end
 
@@ -107,28 +118,39 @@ defmodule ExUnit.Runner do
   end
 
   defp run_module(config, module) do
-    test_module = module.__ex_unit__()
-    EM.module_started(config.manager, test_module)
+    if max_failures_reached?(config.stats_pid, config.max_failures) do
+      # This is reached when spawn_modules launches several test_modules,
+      # and one is invalid, and max_failures is reached
+      send(config.runner_pid, {self(), :max_failures_reached})
+    else
+      test_module = module.__ex_unit__()
+      EM.module_started(config.manager, test_module)
 
-    # Prepare tests, selecting which ones should
-    # run and which ones were skipped.
-    tests = prepare_tests(config, test_module.tests)
+      # Prepare tests, selecting which ones should be run or skipped
+      tests = prepare_tests(config, test_module.tests)
 
-    {test_module, pending, finished_tests} =
-      if Enum.all?(tests, & &1.state) do
-        # The pending tests here aren't actually run, so they're already
-        # "finished"
-        {test_module, tests, tests}
-      else
-        spawn_module(config, test_module, tests)
-      end
+      {test_module, test_results} =
+        if Enum.all?(tests, & &1.state) do
+          # The pending tests here aren't actually run,
+          # so they're marked as "finished".
+          {test_module, %{pending: tests, finished: tests}}
+        else
+          spawn_module(config, test_module, tests)
+        end
 
-    # Run the pending tests. We don't actually spawn those
-    # tests but we do send the notifications to formatter.
-    Enum.each(pending, &run_test(config, &1, []))
-    test_module = %{test_module | tests: finished_tests}
-    EM.module_finished(config.manager, test_module)
-    send(config.runner_pid, {self(), :module_finished, test_module})
+      # Do not run pending tests.
+      # Just send the notifications to the formatter.
+      Enum.each(test_results.pending, fn test ->
+        EM.test_started(config.manager, test)
+        EM.test_finished(config.manager, test)
+      end)
+
+      test_module = %{test_module | tests: test_results.finished}
+      EM.module_finished(config.manager, test_module)
+
+      # message loop/2
+      send(config.runner_pid, {self(), :module_finished})
+    end
   end
 
   defp prepare_tests(config, tests) do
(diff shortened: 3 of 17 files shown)


eksperimental:
@josevalim I initially decided to go with the approach of loading all test cases, because I think it is important to inform the user how many tests are in total. But now that the implementation is fully working, if it needs to be removed it's easier to do if you considered so.

eksperimental:
I'm running OTP 21.0.6, and the `make clean test` is passing, I don't know why CI is failing in every version

josevalim:
Thanks @eksperimental! We are getting closed to have this merged but the PR is still mixed with unrelated changes. For example, renaming of variables, change of aliases. Can you please revert those changes? The runner code is complex so we should do the minimum amount of changes so we can focus on reviewing the important parts. The renaming and refactoring of related parts can be done in a separate PR.

josevalim:
Also, to keep this simpler, let's also revert the statuses changes. I don't want to add a new status "not executed" or at least not in the initial implementation. So we shouldn't worry about tracking or notifying those. We can make this clear in the failure message "--max-failures reached, aborting test suite".

eksperimental:
It will be less time consuming for me to submit a PR that renames variables and change the order of arguments in functions, and build thIS PR on top of that, and it will be quick to be reviewed and approve than this one. Ley me know if that suits you

josevalim:
Your call. :)

eksperimental:
Just to be clear, so a test with 10 tests, where 2nd, 3rd and 4th fail, and --max-failures 2. What should the stats return? Total: 3, failures: 2
Or should we return nil because the suites been aborted?

josevalim:
As soon as we collect --max-failures, we will stop reporting the other tests and failures. We will just do enough work to make sure they terminate correctly.

eksperimental:
This is finally ready for a review

lib/ex_unit/lib/ex_unit.ex
```diff
@@ -62,11 +62,11 @@ defmodule ExUnit do
...
 
-    1. Passed (also represented by `nil`)
+    1. Passed (represented by `nil`)
     2. Failed
-    3. Skipped (via @tag :skip)
-    4. Excluded (via :exclude filters)
-    5. Invalid (when setup_all fails)
+    3. Skipped (via `@tag :skip`)
+    4. Excluded (via `:exclude` filters)
+    5. Invalid (when `setup_all/1,2` fail)
```
josevalim: Unrelated to this change, please do it in a separate commit.

lib/ex_unit/lib/ex_unit/cli_formatter.ex
```diff
@@ -153,8 +153,29 @@ defmodule ExUnit.CLIFormatter do
     test_counter =
       Enum.reduce(test_module.tests, config.test_counter, &update_test_counter(&2, &1))
 
-    failure_counter = config.failure_counter + tests_length
-    config = %{config | test_counter: test_counter, failure_counter: failure_counter}
+    config = %{
+      config
+      | test_counter: test_counter,
+        failure_counter: config.failure_counter + tests_length
+    }
```
josevalim: Unrelated to this change, please revert.

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -332,6 +416,87 @@ defmodule ExUnit.Runner do
...
+  end
+
+  defp register_test_module(pid) do
+    ExUnit.OnExitHandler.register(pid)
+  end
+
+  defp get_failure_counter(stats_pid) when is_pid(stats_pid),
+    do: RunnerStats.get_failure_counter(stats_pid)
+
+  defp increment_failure_counter(stats_pid, %struct{state: {tag, _}}, increment \\ 1)
```
josevalim: Please avoid default arguments on private functions.

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -332,6 +416,87 @@ defmodule ExUnit.Runner do
...
+  defp process_failure(
+         process_failure_config,
+         %ExUnit.TestModule{state: {tag, _}, tests: tests} = test_module
+       )
+       when tag in [:failed, :invalid] do
+    %{config: config} = process_failure_config
+    failure_counter = increment_failure_counter(config.stats_pid, test_module, length(tests))
+
+    if max_failures_reached?(failure_counter, config.max_failures) do
+      max_failures_has_been_reached(config.manager, config.stats_pid, nil)
```
josevalim: Now that I see the code, I think killing all running tests will be harmful and potentially generate tons of logs. That's because if the test process has spawned other processes, killing the test process will be considered as a cascading failure. I think it is best just to not kill them and let them run their "course".

Also, don't we need to make a distinction between "max_failures" has just been reached and "max failures" has been surpassed? When it has just been reached, you still want to log that test but you want to stop the run. For subsequent cases, you don't want to log, nor continue the run.

lib/ex_unit/lib/ex_unit.ex
eksperimental: done

lib/ex_unit/lib/ex_unit/runner.ex

lib/ex_unit/lib/ex_unit/runner.ex
eksperimental: Just wondering if exiting like this would avoid the logging and also the cascading effect
```elixir
Process.exit(registered_pid, {:shutdown, :exit_max_failures_reached})
```

eksperimental: So I have implemented your suggestion. Let me know if that works.
Regarding max failures reached and surpassed, we I don't think we need that, because I log the tests first and then I do `process_failures` which is based on the passed test, and just then I update the failure counter, if it's reached, I launch the `max_failures_have_been_reached` function, and any subsequent test will just be ignored and marked as :not_executed if they were still running

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -107,28 +118,39 @@ defmodule ExUnit.Runner do
...
+      Enum.each(test_results.pending, fn test ->
+        EM.test_started(config.manager, test)
+        EM.test_finished(config.manager, test)
+      end)
+
+      test_module = %{test_module | tests: test_results.finished}
+      EM.module_finished(config.manager, test_module)
+
+      # message loop/2
+      send(config.runner_pid, {self(), :module_finished})
```
fertapric: I see you simplified the message by not including the module, could this be further simplified by just sending to `loop/3` the messages `:module_finished` and `:max_failures_reached`?

Btw, the comment should say `loop/3` :) But, is it necessary?

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -157,35 +179,82 @@ defmodule ExUnit.Runner do
...
 
-    {exec_on_exit(test_module, module_pid, timeout), pending, finished_tests}
+    test_module = exec_on_exit(test_module, module_pid, timeout)
+    {test_module, test_results}
+  end
+
+  defp spawn_module_monitor(config, test_module, parent_pid, tests) do
+    spawn_monitor(fn ->
+      module_pid = self()
+      register(module_pid)
```
fertapric: We are missing information here: "on exit".

fertapric: In addition, I prefer `self()` because it always means "the current process"

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -157,35 +179,82 @@ defmodule ExUnit.Runner do
...
+      end
+
+      exit(:shutdown)
+    end)
+  end
+
+  # returns a map in the shape of
+  # %{:not_executed => list(), :finished => list()}
+  defp run_tests(config, tests, context) do
+    Enum.map(tests, fn test ->
```
fertapric: Minor nitpick:

```elixir
tests
|> Enum.map(tests, &run_test(config, &1, context)) # or  fn test -> run_test(config, test, context)
|> Enum.group_by(&elem(&1, 0), &elem(&1, 1))
|> Map.put_new(:not_executed, [])
|> Map.put_new(:finished, [])
```

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -107,28 +118,39 @@ defmodule ExUnit.Runner do
...
+      EM.module_started(config.manager, test_module)
+
+      # Prepare tests, selecting which ones should be run or skipped
+      tests = prepare_tests(config, test_module.tests)
+
+      {test_module, test_results} =
+        if Enum.all?(tests, & &1.state) do
+          # The pending tests here aren't actually run,
+          # so they're marked as "finished".
+          {test_module, %{pending: tests, finished: tests}}
```
josevalim: Can you please revert those changes? Previously it was returning a tuple with three elements and it should continue to return a tuple with three elements, as this change is not related to this PR.

I have already said this a couple times but these unrelated changes make the PR really hard to review. After a couple rounds, we are still finding layers of code that were just moved around. Next time, please, please consider making the PR as small as possible. Any refactoring should happen elsewhere.

lib/ex_unit/lib/ex_unit/runner.ex
josevalim: Please let's continue sending the three element tuple. If the previous element was not necessary, it should be removed in another PR.

lib/ex_unit/lib/ex_unit/runner.ex
fertapric: @eksperimental I like this change, but it generates some noise in this not-so-small PR. 

I understand that while working on big PRs like this one you'll see a lot of points for improvement, but please, try to keep you focused on the main goal of the PR 🙂

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -316,13 +420,17 @@ defmodule ExUnit.Runner do
...
 
   defp exec_on_exit(test_or_case, pid, timeout) do
-    case ExUnit.OnExitHandler.run(pid, timeout) do
-      :ok ->
-        test_or_case
-
-      {kind, reason, stack} ->
-        state = test_or_case.state || failed(kind, reason, prune_stacktrace(stack))
-        %{test_or_case | state: state}
+    if ExUnit.OnExitHandler.registered?(pid) do
```
josevalim: Why do we need to wrap this on `registered?`? `run` should return `:ok` if there is nothing registered. We probably don't need a pre-check.

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -157,35 +179,82 @@ defmodule ExUnit.Runner do
...
-    {test_module, pending, finished_tests} =
+    {test_module, test_results} =
       receive do
-        {^module_pid, :module_finished, test_module, failed_tests, finished_tests} ->
-          Process.demonitor(module_ref, [:flush])
-          {test_module, failed_tests, finished_tests}
+        {^module_pid, :module_finished, test_module, tests} ->
+          if max_failures_reached?(config.stats_pid, config.max_failures) do
+            Process.demonitor(module_ref, [:flush])
+            {test_module, %{pending: [], finished: []}}
```
josevalim: Let's continue with the three element tuple here too. No need to put it in a map only to extract it out.

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -157,35 +179,82 @@ defmodule ExUnit.Runner do
...
+          else
+            process_failure(config, test_module)
+            Process.demonitor(module_ref, [:flush])
```
josevalim: This call exists in both `if` and `else` branches, so we should probably move it up.

lib/ex_unit/lib/ex_unit/runner.ex
josevalim: We should not return a map here though. In fact, we should stop running the other tests as soon as we get the first `not_executed`.

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -217,68 +286,101 @@ defmodule ExUnit.Runner do
...
-        run_test_with_capture_log(capture_log, config, test, Map.merge(tags, context))
-      else
-        test
+    if max_failures_reached?(config.stats_pid, config.max_failures) do
+      {:not_executed, test}
+    else
+      EM.test_started(config.manager, test)
+
+      # We need to retrieve the status, because the test can be aborted
+      # during its execution due to max_failures being reached
```
josevalim: When could this happen?

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -157,35 +179,82 @@ defmodule ExUnit.Runner do
...
-        end
-
-        exit(:shutdown)
-      end)
+    {module_pid, module_ref} = spawn_module_monitor(config, test_module, parent_pid, tests)
 
-    {test_module, pending, finished_tests} =
+    {test_module, test_results} =
       receive do
-        {^module_pid, :module_finished, test_module, failed_tests, finished_tests} ->
```
josevalim: Please let's keep this same message format as before.

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -157,35 +179,82 @@ defmodule ExUnit.Runner do
...
-          Process.demonitor(module_ref, [:flush])
-          {test_module, failed_tests, finished_tests}
+        {^module_pid, :module_finished, test_module, tests} ->
+          if max_failures_reached?(config.stats_pid, config.max_failures) do
+            Process.demonitor(module_ref, [:flush])
+            {test_module, %{pending: [], finished: []}}
+          else
+            process_failure(config, test_module)
+            Process.demonitor(module_ref, [:flush])
+            {test_module, tests}
```
josevalim: Let's keep it the same three element tuple as before too.

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -157,35 +179,82 @@ defmodule ExUnit.Runner do
...
+          if max_failures_reached?(config.stats_pid, config.max_failures) do
+            {test_module, %{pending: [], finished: []}}
+          else
+            test_module = %{test_module | state: failed({:EXIT, module_pid}, error, [])}
+            process_failure(config, test_module)
+            {test_module, %{pending: [], finished: []}}
+          end
       end
 
-    {exec_on_exit(test_module, module_pid, timeout), pending, finished_tests}
```
josevalim: Let's keep it the same three element tuple as before too.

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -157,35 +179,82 @@ defmodule ExUnit.Runner do
...
+             %{pending: [], finished: tests_run.finished}}
+          )
+
+        {:error, test_module} ->
+          invalid_tests = Enum.map(tests, &%{&1 | state: {:invalid, test_module}})
+
+          send(
+            parent_pid,
+            {module_pid, :module_finished, test_module, %{pending: invalid_tests, finished: []}}
+          )
```
josevalim: We should not change the message format.

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -157,35 +179,82 @@ defmodule ExUnit.Runner do
...
+      case exec_module_setup(test_module) do
+        {:ok, _test_module, context} ->
+          # max_failures can be reached during the execution of test_module,
+          # so we keep track of which tests are not executed and which ones are finished.
+          tests_run = run_tests(config, tests, context)
+
+          send(
+            parent_pid,
+            {module_pid, :module_finished, test_module,
+             %{pending: [], finished: tests_run.finished}}
```
josevalim: We should not change the message format.

lib/mix/lib/mix/tasks/test.ex
```diff
@@ -569,6 +574,25 @@ defmodule Mix.Tasks.Test do
     end
   end
 
+  defp normalize_max_failures(opts) do
+    case Keyword.fetch(opts, :max_failures) do
+      {:ok, value} when value in ["infinity", ":infinity"] ->
+        Keyword.put(opts, :max_failures, :infinity)
```
josevalim: I don't think we need to support this. The equivalent of setting it to infinity is not setting it at all.

lib/mix/test/mix/tasks/test_test.exs
```diff
@@ -257,6 +258,75 @@ defmodule Mix.Tasks.TestTest do
     end)
   end
 
+  describe "--max-failures" do
```
josevalim: I don't believe we need those integration tests as they are very expensive. We should be able to check all of those cases through the ExUnit suite itself.

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -217,68 +286,101 @@ defmodule ExUnit.Runner do
...
 
-  defp receive_test_reply(test, test_pid, test_ref, timeout) do
+  defp receive_test_reply(config, test, test_pid, test_ref, timeout) do
     receive do
       {^test_pid, :test_finished, test} ->
+        if max_failures_reached?(config.stats_pid, config.max_failures) do
+          {:not_executed, test}
+        else
+          process_failure(config, test)
+          Process.demonitor(test_ref, [:flush])
```
josevalim: The demonitor should be moved up.

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -157,35 +179,82 @@ defmodule ExUnit.Runner do
...
+          if max_failures_reached?(config.stats_pid, config.max_failures) do
+            Process.demonitor(module_ref, [:flush])
+            {test_module, %{pending: [], finished: []}}
+          else
+            process_failure(config, test_module)
+            Process.demonitor(module_ref, [:flush])
+            {test_module, tests}
+          end
+
+        {^module_pid, :max_failures_reached, test_module, tests} ->
```
josevalim: I don't see when this message would be ever delivered. Is it still necessary?

lib/ex_unit/lib/ex_unit/runner.ex
eksperimental: @fertapric what do you exactly mean by missing on exit information?

fertapric: The previous code was `ExUnit.OnExitHandler.register(pid)` and now is `register(pid)`. Before it was clear that you were registering this process for something related to “on exit” callbacks.

lib/mix/test/mix/tasks/test_test.exs
```diff
@@ -257,6 +258,75 @@ defmodule Mix.Tasks.TestTest do
...
+
+        assert mix(:concurrent, ~w[test --max-failures 2 --max-cases 3]) =~
+                 "\n5 tests, 2 failures\n"
+
+        assert mix(:concurrent, ~w[test --max-failures 8 --max-cases 10]) =~
+                 "\n12 tests, 3 failures, 5 invalid\n"
+      end)
+    end
+
+    test "aborts concurrent running tests" do
```
josevalim: This test and the other two below are excellent. Can we please move them to the ExUnit suite? They are really testing ExUnit and not the mix integration. Thanks! :heart:

lib/ex_unit/lib/ex_unit/runner.ex
eksperimental: True. It was a left over from when we were also registering the PIDs in the runnerstats

josevalim:
Thank you! Merged manually in 8c373044dcde1654670d8da10448e4b8fecf0b93! 

Note I did some changes before merging. Namely, I reduced the amount of times we check for `max_reached_finished?`, as it felt a lot of times we should have known the answer as long as we passed it around.

I have also pushed some fixes that made it work properly for umbrellas.

stevegraham:
Thank you @eksperimental! 🙌🏻

PR #8254, opened 2018-10-05
Add option for disabling 'Randomized with seed <int>' message after running tests

thecodeboss:
I've always personally found this message adds noise to our tests, especially in umbrella projects where we run through all the tests for a bunch of applications.

This change should be fully backwards compatible, in the unlikely event something expects these messages to be printed to stdio.

### DIFF
--- a/lib/ex_unit/lib/ex_unit.ex
+++ b/lib/ex_unit/lib/ex_unit.ex
@@ -252,6 +252,9 @@ defmodule ExUnit do
     * `:refute_receive_timeout` - the timeout to be used on `refute_receive`
       calls, defaults to `100` milliseconds;
 
+    * `:print_seed` - whether to print out the randomized seed when the test suite
+      finishes. This is enabled by default;
+
     * `:seed` - an integer seed value to randomize the test suite. This seed
       is also mixed with the test module and name to create a new unique seed
       on every test, which is automatically fed into the `:rand` module. This
@@ -288,6 +291,7 @@ defmodule ExUnit do
   def configuration do
     Application.get_all_env(:ex_unit)
     |> put_seed()
+    |> put_print_seed()
     |> put_slowest()
     |> put_max_cases()
   end
@@ -364,6 +368,10 @@ defmodule ExUnit do
       |> System.convert_time_unit(:native, :microsecond)
       |> rem(1_000_000)
     end)
+  end
+
+  defp put_print_seed(opts) do
+    Keyword.put_new(opts, :print_seed, true)
   end
 
   defp put_max_cases(opts) do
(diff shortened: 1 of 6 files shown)


whatyouhide:
I lean on 👎 for this change because if people start using `print_seed: false`, then they'll lose the benefit of being able to rerun the tests with the same seed and they won't be able to recover the seed if their tests fail. But it's not a strong 👎 for me, so I'll be happy to wait for others to chime in.

wojtekmach:
How about writing a custom formatter instead?

FWIW, if the only thing we need is a very slightly change in the output, writing a custom formatter is a bit annoying because I believe there aren't any extension points to re-use existing one, I think we'd have to pretty much copy-paste the whole formatter and then make the minor change. Thus, I'd personally prefer to focus on making existing formatter more extensible.

thecodeboss:
Writing a custom formatter does not seem like the right solution here. Copying and pasting an entire module just so I can remove one line is a bad pattern. I think it's better to just have an option for disabling it in the formatter.

wojtekmach:
Agreed, the point I was trying to make was that making it easier to extend existing formatter would solve this use case and other similar use cases without having to change ex_unit.

fertapric:
I agree with @whatyouhide.

However, I think we might not print the "Randomized with seed <int>" message if randomization is disabled `--seed 0`. 

@thecodeboss you could use then `mix test --seed 0` directly, or define an alias for `mix test` like `test: ["test --seed 0"]` in your projects.

Thoughts?

thecodeboss:
Updated

josevalim:
Setting seed 0 will remove the randomization, which is an adverse side-effect if you just want to remove the warning. The best way to go is a custom formatter, as @wojtekmach said, as it doesn't make sense for Elixir to support many different options people may want to format their output. Thanks.

fertapric:
Would the changes of this PR make sense if it would have had a different tittle (for example, “Do not print ‘Randomized with seed 0’ message if it’s disabled”)? Another option would be to print a different message.

josevalim:
We could change the message to mean something else but I don't think it would solve the originally reported issue.

thecodeboss:
> it doesn't make sense for Elixir to support many different options people may want to format their output.
I think the core language should have as many options as possible to accommodate peoples' needs, instead of forcing them to copy and paste a [378 line module](https://github.com/elixir-lang/elixir/blob/master/lib/ex_unit/lib/ex_unit/cli_formatter.ex) just so they can change one line of it. I thought adding the `print_seed: false` solution was a pretty good one since it's fully backwards-compatible, but apparently no one liked that idea 😢

I'll just deal with the extra output and let myself out.

PR #8255, opened 2018-10-05
Use Git, GitHub, Hex instead of git, github, hex

fertapric:
Minor thing spotted while reviewing https://github.com/elixir-lang/elixir/pull/8253

### DIFF
--- a/lib/mix/lib/mix/local/installer.ex
+++ b/lib/mix/lib/mix/local/installer.ex
@@ -65,7 +65,7 @@ defmodule Mix.Local.Installer do
       {:fetcher, dep_spec} ->
         if opts[:sha512] do
           Mix.raise(
-            "--sha512 is not supported when installing from git/github/hex\n\n" <> usage(module)
+            "--sha512 is not supported when installing from Git, GitHub, or Hex\n\n" <> usage(module)
           )
         end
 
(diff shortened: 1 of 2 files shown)


lib/mix/lib/mix/tasks/archive.install.ex
```diff
@@ -34,8 +34,9 @@ defmodule Mix.Tasks.Archive.Install do
 
       mix some_task
 
-  Note that installing via Git/GitHub/Hex fetches the source of the archive
-  and builds it, while using URL/local path fetches a pre-built archive.
+  Note that installing via Git, GitHub, or Hex fetches the source
+  of the archive and builds it, while using a URL, or a local path
```
whatyouhide: No comma after URL.

fertapric: Done. Thanks!

lib/mix/lib/mix/local/installer.ex
```diff
@@ -65,7 +65,8 @@ defmodule Mix.Local.Installer do
       {:fetcher, dep_spec} ->
         if opts[:sha512] do
           Mix.raise(
-            "--sha512 is not supported when installing from git/github/hex\n\n" <> usage(module)
+            "--sha512 is not supported when installing from Git, GitHub, or Hex\n\n" <>
```
josevalim: I think in here we are referring to the second argument given to archive.install, which is `git`, `github` and `hex`.

PR #8257, opened 2018-10-08
ExUnit: Rename variables and refactor config and loop 

eksperimental:
This is a preliminary PR in order to implement the --max-failures feature in ExUnit

It only renames arguments, variables, and reorganizes code.
As requested in https://github.com/elixir-lang/elixir/pull/8247#issuecomment-426530597

### DIFF
--- a/lib/ex_unit/lib/ex_unit/event_manager.ex
+++ b/lib/ex_unit/lib/ex_unit/event_manager.ex
@@ -2,6 +2,8 @@ defmodule ExUnit.EventManager do
   @moduledoc false
   @timeout 30000
 
+  @typep manager :: {supervisor_manager :: pid, event_manager :: pid}
+
   # TODO: Remove support for GenEvent formatters on 2.0
 
   @doc """
@@ -10,6 +12,7 @@ defmodule ExUnit.EventManager do
   This is what power formatters as well as the
   internal statistics server for ExUnit.
   """
+  @spec start_link() :: {:ok, manager}
   def start_link() do
     {:ok, sup} = DynamicSupervisor.start_link(strategy: :one_for_one)
     {:ok, event} = :gen_event.start_link()
@@ -44,35 +47,35 @@ defmodule ExUnit.EventManager do
     end
   end
 
-  def suite_started(ref, opts) do
-    notify(ref, {:suite_started, opts})
+  def suite_started(manager, opts) do
+    notify(manager, {:suite_started, opts})
   end
 
-  def suite_finished(ref, run_us, load_us) do
-    notify(ref, {:suite_finished, run_us, load_us})
+  def suite_finished(manager, run_us, load_us) do
+    notify(manager, {:suite_finished, run_us, load_us})
   end
 
-  def module_started(ref, test_module) do
+  def module_started(manager, test_module) do
     # TODO: Remove case_started in Elixir v2.0
-    notify(ref, {:module_started, test_module})
-    notify(ref, {:case_started, Map.put(test_module, :__struct__, ExUnit.TestCase)})
+    notify(manager, {:case_started, Map.put(test_module, :__struct__, ExUnit.TestCase)})
+    notify(manager, {:module_started, test_module})
   end
 
-  def module_finished(ref, test_module) do
+  def module_finished(manager, test_module) do
     # TODO: Remove case_finished in Elixir v2.0
-    notify(ref, {:case_finished, Map.put(test_module, :__struct__, ExUnit.TestCase)})
-    notify(ref, {:module_finished, test_module})
+    notify(manager, {:case_finished, Map.put(test_module, :__struct__, ExUnit.TestCase)})
+    notify(manager, {:module_finished, test_module})
   end
 
-  def test_started(ref, test) do
-    notify(ref, {:test_started, test})
+  def test_started(manager, test) do
+    notify(manager, {:test_started, test})
   end
 
-  def test_finished(ref, test) do
-    notify(ref, {:test_finished, test})
+  def test_finished(manager, test) do
+    notify(manager, {:test_finished, test})
   end
 
-  defp notify({sup, event}, msg) do
+  defp notify({sup, event} = _manager, msg) do
     :gen_event.notify(event, msg)
 
     for {_, pid, _, _} <- Supervisor.which_children(sup) do
--- a/lib/ex_unit/lib/ex_unit/runner.ex
+++ b/lib/ex_unit/lib/ex_unit/runner.ex
@@ -5,23 +5,21 @@ defmodule ExUnit.Runner do
 
   @rand_algorithm :exs1024
 
-  @type manager :: ExUnit.EventManager.manager()
-
-  def run(load_time, opts) when (is_integer(load_time) or is_nil(load_time)) and is_list(opts) do
+  def run(opts, load_us) when (is_integer(load_us) or is_nil(load_us)) and is_list(opts) do
     opts = normalize_opts(opts)
-
     {:ok, manager} = EM.start_link()
     {:ok, stats_pid} = EM.add_handler(manager, ExUnit.RunnerStats, opts)
     config = configure(opts, manager, self(), stats_pid)
+
     :erlang.system_flag(:backtrace_depth, Keyword.fetch!(opts, :stacktrace_depth))
 
-    {run_time, _} =
+    {run_us, _} =
       :timer.tc(fn ->
         EM.suite_started(config.manager, opts)
-        loop(config)
+        loop(config, :async, 0)
       end)
 
-    EM.suite_finished(config.manager, run_time, load_time)
+    EM.suite_finished(config.manager, run_us, load_us)
     stats = ExUnit.RunnerStats.stats(stats_pid)
     EM.stop(config.manager)
     after_suite_callbacks = Application.fetch_env!(:ex_unit, :after_suite)
@@ -55,34 +53,32 @@ defmodule ExUnit.Runner do
     |> Keyword.put(:include, include)
   end
 
-  defp loop(modules \\ :async, taken \\ 0, config)
-
-  defp loop(:async, taken, config) do
+  defp loop(config, :async, taken) do
     available = config.max_cases - taken
 
     cond do
       # No modules available, wait for one
       available <= 0 ->
-        wait_until_available(:async, taken, config)
+        wait_until_available(config, :async, taken)
 
       # Slots are available, start with async modules
       modules = ExUnit.Server.take_async_modules(available) ->
-        spawn_modules(modules, [], taken, config)
+        spawn_modules(config, modules, [], taken)
 
       true ->
         modules = ExUnit.Server.take_sync_modules()
-        loop(modules, taken, config)
+        loop(config, modules, taken)
     end
   end
 
-  defp loop(modules, taken, config) do
+  defp loop(config, modules, taken) do
     case modules do
       _ when taken > 0 ->
-        wait_until_available(modules, taken, config)
+        wait_until_available(config, modules, taken)
 
       # So we can start all sync modules
       [head | tail] ->
-        spawn_modules([head], tail, taken, config)
+        spawn_modules(config, [head], tail, taken)
 
       # No more modules, we are done!
       [] ->
@@ -93,30 +89,30 @@ defmodule ExUnit.Runner do
   # Loop expecting messages from the spawned modules. Whenever
   # a module has finished executing, decrease the taken modules
   # counter and attempt to spawn new ones.
-  defp wait_until_available(modules, taken, config) do
+  defp wait_until_available(config, modules, taken) do
     receive do
       {_pid, :module_finished, _test_case} ->
-        loop(modules, taken - 1, config)
+        loop(config, modules, taken - 1)
     end
   end
 
-  defp spawn_modules(modules, modules_left, taken, config) do
+  defp spawn_modules(config, modules, modules_remaining, taken) do
     Enum.each(modules, fn module ->
       spawn_link(fn ->
-        run_module(module, config)
+        run_module(config, module)
       end)
     end)
 
-    loop(modules_left, taken + length(modules), config)
+    loop(config, modules_remaining, taken + length(modules))
   end
 
-  defp run_module(module, config) do
+  defp run_module(config, module) do
     test_module = module.__ex_unit__()
     EM.module_started(config.manager, test_module)
 
     # Prepare tests, selecting which ones should
     # run and which ones were skipped.
-    tests = prepare_tests(test_module.tests, config)
+    tests = prepare_tests(config, test_module.tests)
 
     {test_module, pending, finished_tests} =
       if Enum.all?(tests, & &1.state) do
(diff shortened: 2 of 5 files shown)


lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -2,14 +2,18 @@ defmodule ExUnit.Runner do
   @moduledoc false
 
   alias ExUnit.EventManager, as: EM
+  alias ExUnit.RunnerStats
```
eksperimental: This alias is used in several instances in the --max-failures implementation

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -229,13 +231,13 @@ defmodule ExUnit.Runner do
...
 
-  defp run_test(config, %{tags: tags} = test, context) do
+  defp run_test(%{tags: tags} = test, context, config) do
     EM.test_started(config.manager, test)
 
     test =
       if is_nil(test.state) do
-        capture_log? = Map.get(tags, :capture_log, config.capture_log)
-        run_test_with_capture_log(capture_log?, config, test, Map.merge(tags, context))
+        capture_log = Map.get(tags, :capture_log, config.capture_log)
```
eksperimental: I have removed the `?` from this variables, as it is not only boolean values that will hold

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -335,6 +344,10 @@ defmodule ExUnit.Runner do
 
   ## Helpers
 
+  defp generate_test_seed(%ExUnit.Test{module: module, name: name}, %{seed: seed}) do
+    :rand.seed(@rand_algorithm, {:erlang.phash2(module), :erlang.phash2(name), seed})
+  end
+
```
eksperimental: this was moved here under the Helper section

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -187,7 +190,6 @@ defmodule ExUnit.Runner do
           {test_module, [], []}
       end
 
-    timeout = get_timeout(%{}, config)
```
eksperimental: moved to the top of the function, where we define variables

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -2,14 +2,18 @@ defmodule ExUnit.Runner do
...
+  @type manager :: ExUnit.EventManager.manager()
+
   def run(opts, load_us) do
     {:ok, manager} = EM.start_link()
-    {:ok, stats} = EM.add_handler(manager, ExUnit.RunnerStats, opts)
+    {:ok, stat_sup} = EM.add_handler(manager, RunnerStats, opts)
     {opts, config} = configure(manager, opts)
-
+    config = Map.put(config, :stat_sup, stat_sup)
+    config = Map.put(config, :top_pid, self())
```
eksperimental: this is the only addition to adds "a feature", we store the top_pid and stat_sup pids, and we avoid to add an extra argument to functions. Maybe there's a better name for top_pid?

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -264,21 +267,27 @@ defmodule ExUnit.Runner do
...
 
-  defp generate_test_seed(%{seed: seed}, %ExUnit.Test{module: module, name: name}) do
-    :rand.seed(@rand_algorithm, {:erlang.phash2(module), :erlang.phash2(name), seed})
+    exec_on_exit(test, test_pid, timeout)
   end
 
-  defp receive_test_reply(test, test_pid, test_ref, timeout) do
+  defp receive_test_reply(
+         test,
+         %{pid: test_pid, ref: test_ref, timeout: timeout} = options
```
josevalim: Please move the pattern to inside the function.

lib/ex_unit/lib/ex_unit/event_manager.ex
```diff
@@ -2,6 +2,8 @@ defmodule ExUnit.EventManager do
   @moduledoc false
   @timeout 30000
 
+  @type manager :: {supervisor_manager :: pid, event_manager :: pid}
```
josevalim: Please make it `@typep`.

lib/ex_unit/lib/ex_unit/runner.ex
josevalim: So lets make sure that we add those when we create the map and not afterwards. Otherwise we get the impression the map is dynamic but it keys are all static (i.e. known upfront). Let's also call those `stats_pid` and `runner_pid`. After all, we don't know that the pid for stats is a sup, that would be leaking implementation detail.

josevalim:
Tks @eksperimental. What is the reason to change the order of config?

josevalim:
Before this PR, the only function that received `config` as last argument was `get_timeout`. When we change the argument order, we are generating a bunch of churn for not clear benefit. Unless we have a strong reason for doing so, I would recommend to revert the change. There is nothing that says the `state` like argument has to be the last one and it is not a convention we particularly follow for internal functions, especially when a behaviour is *not* involved.

lib/ex_unit/lib/ex_unit/runner_stats.ex
```diff
@@ -4,8 +4,8 @@ defmodule ExUnit.RunnerStats do
   use GenServer
   alias ExUnit.{FailuresManifest, Test, TestModule}
 
-  def stats(pid) do
-    GenServer.call(pid, :stats, :infinity)
+  def stats(sup) do
```
fertapric: I would use `pid` here

eksperimental:
> Before this PR, the only function that received `config` as last argument was `get_timeout`. So if the goal is consistency, I would change `get_timeout`.
> 
> When we change the argument order, we are generating a bunch of churn for not clear benefit. Unless we have a strong reason for doing so, I would recommend to revert the change. There is nothing that says the "state-like" argument has to be the last one and it is not a convention we particularly follow for internal functions, especially when a behaviour is _not_ involved.

I have never considered `config` a state-like argument. It has always looked to me as an `options` argument, and that's the reason why it was moved to the end of the arg list.
Looking into it, I realized that the only state value was `modules`, and the pids and the manager, but those while not provided by the user, they never change so they could be considered part of config.
I have refactored the code. Now it does more than just renaming variables, but I think it defines the line in what's state and what's a configuration option

lib/ex_unit/lib/ex_unit/runner_stats.ex
eksperimental: fixed

lib/ex_unit/lib/ex_unit/event_manager.ex
eksperimental: done

lib/ex_unit/lib/ex_unit/runner.ex

lib/ex_unit/lib/ex_unit/runner.ex

eksperimental:
please do a second review if you like @fertapric, @josevalim; as I have updated the PR

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -5,45 +5,46 @@ defmodule ExUnit.Runner do
 
   @rand_algorithm :exs1024
 
-  def run(opts, load_us) do
-    {:ok, manager} = EM.start_link()
-    {:ok, stats} = EM.add_handler(manager, ExUnit.RunnerStats, opts)
-    {opts, config} = configure(manager, opts)
+  @type manager :: ExUnit.EventManager.manager()
+
+  def run(load_time, opts) when (is_integer(load_time) or is_nil(load_time)) and is_list(opts) do
```
eksperimental: Not sure if its a wise choice to change the argument order here for the sake of consistency.
Could it affect third party libraries? 
WDYT @josevalim?

josevalim: Let's please just avoid those changes generally. The benefits are very little over the history change and potential breakage.

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -5,45 +5,46 @@ defmodule ExUnit.Runner do
...
+    %{
       capture_log: opts[:capture_log],
       exclude: opts[:exclude],
       include: opts[:include],
       manager: manager,
       max_cases: opts[:max_cases],
       only_test_ids: opts[:only_test_ids],
+      runner_pid: runner_pid,
       seed: opts[:seed],
-      modules: :async,
```
eksperimental: so now, `modules` is passed around by the loop function, as there is no need to keep it in config.

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -5,45 +5,46 @@ defmodule ExUnit.Runner do
 
   @rand_algorithm :exs1024
 
-  def run(opts, load_us) do
-    {:ok, manager} = EM.start_link()
-    {:ok, stats} = EM.add_handler(manager, ExUnit.RunnerStats, opts)
-    {opts, config} = configure(manager, opts)
+  @type manager :: ExUnit.EventManager.manager()
```
josevalim: We can't do this because the type is privat. Are we even using it?

eksperimental: I have the feeling I do in the main PR.
I will remove it for now and add it back if needed

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -5,45 +5,46 @@ defmodule ExUnit.Runner do
...
+
+  def run(load_time, opts) when (is_integer(load_time) or is_nil(load_time)) and is_list(opts) do
+    opts = normalize_opts(opts)
 
+    {:ok, manager} = EM.start_link()
+    {:ok, stats_pid} = EM.add_handler(manager, ExUnit.RunnerStats, opts)
+    config = configure(opts, manager, self(), stats_pid)
     :erlang.system_flag(:backtrace_depth, Keyword.fetch!(opts, :stacktrace_depth))
 
-    {run_us, _} =
```
josevalim: I think we are losing important information here, which is that the runtime is in `us`.

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -54,32 +55,34 @@ defmodule ExUnit.Runner do
     |> Keyword.put(:include, include)
   end
 
-  defp loop(%{modules: :async} = config, taken) do
+  defp loop(modules \\ :async, taken \\ 0, config)
```
josevalim: I personally find an antipattern to use default arguments for private functions. Can we please pass all arguments the time we call it? Thanks.

josevalim:
@eksperimental I have added some comments. Can we please keep this PR focused? In your latest commits there are new changes compared to the earlier commit, so we have to review the whole PR again, and probably another pass later on once you do your changes again. It is generally preferred for us to keep PRs smalls and focused.

lib/ex_unit/lib/ex_unit/runner.ex
josevalim: It shouldn't be necessary in the main PR either because we don't use spec for private functions or private functionality and everything here is private.

lib/ex_unit/lib/ex_unit/runner.ex
fertapric: Agree with @josevalim. I would use `usec` though.

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -90,52 +93,50 @@ defmodule ExUnit.Runner do
...
 
-  defp run_module(config, pid, module) do
+  defp run_module(module, config) do
     test_module = module.__ex_unit__()
     EM.module_started(config.manager, test_module)
 
     # Prepare tests, selecting which ones should
     # run and which ones were skipped.
-    tests = prepare_tests(config, test_module.tests)
+    tests = prepare_tests(test_module.tests, config)
```
josevalim: Please revert everywhere we change the order of arguments. It is an unnecessary change that complicates history with no benefit. We will keep `config` as first argument. Thanks.

lib/ex_unit/lib/ex_unit/runner.ex
josevalim: So I would go with `runtime_us`.

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -264,21 +266,25 @@ defmodule ExUnit.Runner do
...
+
+    test = receive_test_reply(test, reply_opts)
 
     exec_on_exit(test, test_pid, timeout)
   end
 
-  defp generate_test_seed(%{seed: seed}, %ExUnit.Test{module: module, name: name}) do
-    :rand.seed(@rand_algorithm, {:erlang.phash2(module), :erlang.phash2(name), seed})
-  end
+  defp receive_test_reply(test, opts) do
```
josevalim: It should not be called `opts`, it should be called `config`, as it is a structured map.

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -90,52 +93,50 @@ defmodule ExUnit.Runner do
...
       {_pid, :module_finished, _test_case} ->
-        loop(config, taken - 1)
+        loop(modules, taken - 1, config)
     end
   end
 
-  defp spawn_modules(config, modules, taken) do
-    pid = self()
-
+  defp spawn_modules(modules, modules_left, taken, config) do
```
josevalim: `remaining_modules` or `modules_remaining` may be a better name then `modules_left`.

lib/ex_unit/lib/ex_unit/runner.ex
eksperimental: good to know, I didn't know that was the difference

lib/ex_unit/lib/ex_unit/runner.ex
eksperimental: The reason I changed it its because `run_us` and `load_us` since it has a meaning in English, it was confusing, I have never seen that term before and neither it is used in Elixir core, only one time here.
Took me a while to figure it out.

If I had read `run_usec` I would have figured it out instantly

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -264,21 +262,24 @@ defmodule ExUnit.Runner do
...
         exit(:shutdown)
       end)
 
-    timeout = get_timeout(test.tags, config)
-    test = receive_test_reply(test, test_pid, test_ref, timeout)
+    reply_config = %{
+      pid: test_pid,
+      ref: test_ref,
+      timeout: timeout
+    }
```
josevalim: Can we please keep the previous call? It is fine for us to call a function with four arguments. I don't think it makes much sense to build a map just to unpack it right after.

josevalim:
Thanks @eksperimental, one last comment and we should be good to go.

whatyouhide:
Nice job @eksperimental, thanks! 💟

whatyouhide:
Nice job @eksperimental, thanks! 💟

eksperimental:
thank you @whatyouhide,

BTW, there is an error in my implementation when passing the modules which most of them are being lost.
I'm working on it, so please don't merge any more PRs as the tests are passing but very few tests are being run.
sorry for the inconvenience

PR #8258, opened 2018-10-08
Fix flaky test in "mix compile.elixir"

fertapric:
This is an attempt to fix the flaky test: https://travis-ci.org/elixir-lang/elixir/jobs/437950338

### DIFF
--- a/lib/mix/test/mix/tasks/compile.elixir_test.exs
+++ b/lib/mix/test/mix/tasks/compile.elixir_test.exs
@@ -502,15 +502,15 @@ defmodule Mix.Tasks.Compile.ElixirTest do
     end)
   end
 
-  test "returns error diagnostics" do
-    in_fixture("no_mixfile", fn ->
+  test "returns error diagnostics", context do
+    in_tmp(context.test, fn ->
+      File.mkdir_p!("lib")
+
       File.write!("lib/a.ex", """
       defmodule A do
         def my_fn(), do: $$$
       end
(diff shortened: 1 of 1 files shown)


whatyouhide:
I find this confusing: if I'm reading this test and I see that the `b.ex` file is removed, I'm gonna really ask myself why 🤔 Is there a way to do this more generically or maybe at least write a comment for why you're doing this?

josevalim:
@fertapric maybe better than removing a file, should we just make sure that this particular test starts with a clean slate (i.e. no files in lib) and then just creates the files it needs? i think one of our fixtures do not have anything in lib (or at least we should have a feature like that, otherwise we are doing a lot of work we don't need to in many tests).

josevalim:
I think we can use something like `in_tmp` and just create a temp path with custom files. I am guessing now that most of the tests that do `in_fixture "no_mixfile"` could be rewritten to just start from scratch and it should be more efficient, especially if most of them are using only one of the two files in there.

fertapric:
@whatyouhide @josevalim good idea. I'll change it.

josevalim:
@fertapric beautiful, we should probably do another pass later on and change tests to rely on "no_mixfile" only when necessary.

josevalim:
@fertapric will you send a PR for the parallel compiler issue too? :)

fertapric:
I'm currently working on it

josevalim:

PR #8259, opened 2018-10-08
Wait for worker termination in parallel compiler

fertapric:
This is also related to the flaky test: https://travis-ci.org/elixir-lang/elixir/jobs/437950338

### DIFF
--- a/lib/elixir/lib/kernel/parallel_compiler.ex
+++ b/lib/elixir/lib/kernel/parallel_compiler.ex
@@ -391,7 +391,11 @@ defmodule Kernel.ParallelCompiler do
         discard_down(child_pid)
         print_error(file, kind, reason, stack)
         cancel_waiting_timer(queued, child_pid)
-        terminate(queued)
+
+        queued
+        |> List.keydelete(child_pid, 0)
+        |> terminate()
+
         {:error, [to_error(file, kind, reason, stack)], warnings}
 
       {:DOWN, ref, :process, _pid, reason} ->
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/kernel/parallel_compiler.ex
```diff
@@ -464,6 +472,7 @@ defmodule Kernel.ParallelCompiler do
   defp terminate(queued) do
     for {pid, _, _, _} <- queued do
       Process.exit(pid, :kill)
+      discard_down(pid)
```
josevalim: Let's do this in a separate process so we send and handle all kills concurrently. So it would be:

```elixir
for {pid, _, _, _} <- queued, do: Process.exit(pid, :kill)
for {pid, _, _, _} <- queued, do: discard_down(pid)
```

fertapric: 👍

Actually, that was original implementation, but I changed after seeing [this PR](https://github.com/elixir-lang/elixir/pull/8239/files#diff-e1812f5a5c51b7f59c6bcdf0040eb7cbR561). 

Should I open another PR and change also that one?

josevalim: Ys, I would change the other one too. :)

fertapric: Done https://github.com/elixir-lang/elixir/pull/8260

josevalim:

PR #8263, opened 2018-10-09
Include optional dependencies in extra_applications

josevalim:
Previously when using optional dependencies, they would
not be automatically included by compile.app in extra_applications.

This meant that an optional dependency would not be guaranteed
to be started before the application. Forcing users to handle
this logic in an ad-hoc fashion that is incompatible with
releases.


josevalim:

PR #8264, opened 2018-10-09
Fix bug introduced to ExUnit in #8257

eksperimental:
Bug was introduced in ca8ce9201aa59b59ea209086c2ead2c8cec47615

Causes almost all tests being log when `mix test` was run, and still tests passing


eksperimental:
I checked the number of tests in #8257 with this patch, and to the previous commit, and they are exactly the same, so I can assure the patch works

PR #8265, opened 2018-10-09
Use ex_unit and not exunit in Makefile log

eksperimental:
`ex_doc` task was using the proper name

    ==> ex_doc (ex_unit)


josevalim:

PR #8266, opened 2018-10-09
Use setup_all instead of "setup all" to refer to the callback

eksperimental:
[ci skip]


whatyouhide:
Thanks @eksperimental 💟

PR #8267, opened 2018-10-09
ExUnit: Replace truthy and falsy functions in assertion tests

eksperimental:
It is misleading, as they are meant to be used as true or false.


eksperimental:
please review again as I accidentally updated a branch that was missing a second commit.

josevalim:
@eksperimental can you please keep this PR minimum?

1. Please don't rename modules
2. Please don't import modules. Since we were previously aliasing, we can assume that an alias was the original intention

I also don't think we need to add new assertions. true and false is a subset of truthy/falsy, so as long as we assert for actually `:truthy/nil`, we should be good to go.

eksperimental:
please have a look

josevalim:

PR #8268, opened 2018-10-10
Correct use of truthy and not true

eksperimental:
Additionally, it corrects the CondClauseError message to:
"no cond clause evaluated to a truthy value"

### DIFF
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -2966,7 +2966,7 @@ defmodule Kernel do
       if(foo, do: bar)
 
   In the example above, `bar` will be returned if `foo` evaluates to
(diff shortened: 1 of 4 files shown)


lib/elixir/lib/kernel.ex
```diff
@@ -2966,7 +2966,7 @@ defmodule Kernel do
       if(foo, do: bar)
 
   In the example above, `bar` will be returned if `foo` evaluates to
-  `true` (i.e., it is neither `false` nor `nil`). Otherwise, `nil` will be
+  a thruthy value (neither `false` nor `nil`). Otherwise, `nil` will be
```
fertapric: s/thruthy/truthy

whatyouhide:
@eksperimental went ahead and fixed @fertapric's comment. Thanks for the contribution! 💟

eksperimental:
Awesome

PR #8270, opened 2018-10-10
Use "note that" instead of just "note"

eksperimental:


josevalim:
Thanks @eksperimental! Keep in mind that in practice [we can remove](https://en.wikipedia.org/wiki/Wikipedia:%22Note_that%22_is_unnecessary) most uses of "note that".

PR #8274, opened 2018-10-10
Improve grammar in defstruct docs

eksperimental:


eksperimental:
I think this paragraph can be improved a bit.
It is not clear whether any should be `Any` in the text bellow

```
  For each protocol in the `@derive` list, Elixir will assert whether there is an
  implementation of that protocol for any (regardless if `@fallback_to_any`
  is `true`) and check if the any implementation defines a `__deriving__/3`
  callback (via `Protocol.derive/3`). If so, the callback is invoked,
  otherwise an implementation that simply points to the `Any` implementation
  is automatically derived.
```

josevalim:
@eksperimental I don't think "whether" is correct because there needs to exist an implementation for Any. I have rewritten this paragraph, what do you think?

lib/elixir/lib/kernel.ex
```diff
@@ -4191,12 +4191,12 @@ defmodule Kernel do
...
       MyProtocol.call(john) #=> works
 
-  For each protocol in the `@derive` list, Elixir will assert there is an
-  implementation of that protocol for any (regardless if `@fallback_to_any`
-  is `true`) and check if the any implementation defines a `__deriving__/3`
-  callback (via `Protocol.derive/3`). If so, the callback is invoked,
-  otherwise an implementation that simply points to the `Any` implementation
-  is automatically derived.
+  For each protocol in the `@derive` list, Elixir will assert the protocol has
+  been implemented for the `Any` type. If the `Any` implementation defines a
```
fertapric: Is `Any` a type? Maybe just:

> For each protocol in the `@derive` list, Elixir will assert the protocol has been implemented for `Any`.

whatyouhide:
Thanks @eksperimental! 💟

PR #8278, opened 2018-10-11
Default System.cmd to File.cwd to fix running on slave nodes

mobileoverlord:
Running `System.cmd` in a slave node without passing `cd: File.cwd!` will use the current working directory of the master node and not the slave node.

For example:

If you start a slave node and use `File.cd "/somewhere"` and then `System.cmd("pwd", [])` the returned result will be that of the master node and not `/somewhere`. This is problematic especially because of the use of `File.cd!/2` where calls in the function expect this to work.

### DIFF
--- a/lib/elixir/lib/system.ex
+++ b/lib/elixir/lib/system.ex
@@ -634,6 +634,12 @@ defmodule System do
         :os.find_executable(cmd) || :erlang.error(:enoent, [command, args, opts])
       end
 
+    opts =
+      case File.cwd() do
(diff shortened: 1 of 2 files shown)


josevalim:
I am not convinced we should change this generally speaking. After all, this is based on ports, and ports still won't work as you describe. Nor anything else built on top ports. So the guarantee simply is not there.

I think you either need to change the relevant calls or start a discussion with Erlang/OTP about the behaviour between nodes and ports.

mobileoverlord:
I'll fix the issue in the places where its a problem.

PR #8279, opened 2018-10-11
Set the cwd for system calls in the git scm

mobileoverlord:
The Git SCM doesn't function properly when running under a slave node because the current working directory is pulled from the master instead of the current slave node. This PR updates calls to `System.cmd/3` to set the current working directory.


lib/mix/lib/mix/scm/git.ex
```diff
@@ -307,4 +310,11 @@ defmodule Mix.SCM.Git do
     {int, _} = Integer.parse(string)
     int
   end
+
+  defp cmd_opts(opts \\ []) do
```
josevalim: I think we should document why this function exists and why we are calling File.cwd!.

mobileoverlord:
I've done further testing with running `mix` inside a slave node and this appears to be the only stumbling point that I have found. I have updated this to address the comments.

josevalim:

PR #8286, opened 2018-10-16
Declare var/0 as built-in type

fertapric:
As discussed in https://github.com/elixir-lang/elixir/issues/5800#issuecomment-418888952

Redefining the pseudo-type "var" might lead to ambiguous typespecs.

Here is an example:

```elixir
    defmodule DefineVarType do
      @type my_var :: integer
      @type var :: integer

      @spec foo(x, x, var) :: boolean when x: var
      def foo(x, y, z), do: x + y == z

      @spec bar(x, x, my_var) :: boolean when x: my_var
      def bar(x, y, z), do: x + y == z
    end
```

### DIFF
--- a/lib/elixir/lib/kernel/typespec.ex
+++ b/lib/elixir/lib/kernel/typespec.ex
@@ -373,6 +373,7 @@ defmodule Kernel.Typespec do
   defp builtin_type?(:nonempty_charlist, 0), do: true
   defp builtin_type?(:keyword, 0), do: true
   defp builtin_type?(:keyword, 1), do: true
+  defp builtin_type?(:var, 0), do: true
   defp builtin_type?(name, arity), do: :erl_internal.is_type(name, arity)
 
   defp ensure_no_defaults!(args) do
--- a/lib/elixir/lib/macro/env.ex
+++ b/lib/elixir/lib/macro/env.ex
@@ -64,13 +64,13 @@ defmodule Macro.Env do
   @type macros :: [{module, [name_arity]}]
   @type context_modules :: [module]
   @type lexical_tracker :: pid | nil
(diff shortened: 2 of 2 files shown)


lib/elixir/lib/kernel/typespec.ex
```diff
@@ -373,6 +373,7 @@ defmodule Kernel.Typespec do
   defp builtin_type?(:nonempty_charlist, 0), do: true
   defp builtin_type?(:keyword, 0), do: true
   defp builtin_type?(:keyword, 1), do: true
+  defp builtin_type?(:var, 0), do: true
```
fertapric: This is a breaking change.

It seems small, but I understand if you prefer to emit a warning in v1.8.0 that can be turned into an error in future Elixir versions.

lib/elixir/lib/macro/env.ex
```diff
@@ -64,13 +64,13 @@ defmodule Macro.Env do
   @type macros :: [{module, [name_arity]}]
   @type context_modules :: [module]
   @type lexical_tracker :: pid | nil
-  @type var :: {atom, atom | term}
+  @type variable :: {atom, atom | term}
```
fertapric: This is also a breaking change, but the test suite won't pass even if we turn the error into a warning 😕

lib/elixir/lib/kernel/typespec.ex
josevalim: On the positive side, we explicitly document that we may introduce new functions in kernel, new modules or types and those may cause conflict which could be breaking. So I think it would be worth giving this a try anyway.

josevalim:

Issue #8076, opened 2018-08-13
Imperative Assignements are breaking the application in 1.7 update


jfayad:
### Environment

* Elixir & Erlang/OTP versions (elixir --version): Erlang/OTP 21 [erts-10.0] [ source] [64-bit]
Elixir 1.7.0 (Compiled with Erlang/OTP 20)
* Operating system: Ubuntu 16.04

### Current behavior

After updated Elixir from 1.6 to 1.7 the `params` variable will always be equal to %{} outside the if statements

```
params = %{}

        if Map.has_key?(entities, "email") do
          params = Map.put_new(params, :email, get_values(entities["email"]))
        end
```

### Expected behavior

The `params` variable should be correctly assigned for now and updating to 1.7 shouldn't break such a code since nothing has been mentioned in the release not of 1.7

jfayad:
Actually no @fertapric since the variable is defined outside the if statement.

wojtekmach:
I can confirm the behaviour changed between 1.6 and 1.7:

```
~/foo% asdf local elixir 1.6.5 && elixir -e "x = 1; if(true, do: x = 2); IO.inspect x" 2> /dev/null
2
~/foo% asdf local elixir 1.7.0 && elixir -e "x = 1; if(true, do: x = 2); IO.inspect x" 2> /dev/null
1
```

fertapric:
🤔 That's odd! 

Before posting, I checked the warning was there using this small example:

```elixir
params = %{}

if true do
  params = Map.put_new(params, :email, 1)
end

IO.inspect params
```

The warning was emitted for the assignment in the line 4 :)

wojtekmach:
@fertapric I got the warning in both cases as expected, I removed it with `2> /dev/null` for brevity.

josevalim:
Duplicate of #7403.

Rationale is here:
https://github.com/elixir-lang/elixir/issues/7403#issuecomment-368902698

We should have a CHANGELOG entry too. But I will be glad to clarify it.
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

jfayad:
Thanks for the clarification @josevalim indeed updating the changelog would have saved us (and probably others) some precious time searching for the reason behind these errors.

josevalim:
Julien, sorry for not being clear. I meant that this is already in the
CHANGELOG but apparently not clear enough. Could you please give it a
[...]

PR #8288, opened 2018-10-16
Remove warning related to imperative assignments

### LINKED ISSUE #8285: Variable definition inside case block warning is printed from with block
### Environment

```

Elixir 1.7.2 (compiled with Erlang/OTP 21)
```
### Current behavior

This code prints a warning that I think should not be there
```
  with :ok <- Foo.ensure_permissions(subject, Authorizer.read_permission()),
         {:ok, messages} <- Message.Mutator.mark_messages_as_read_by_user(query, subject.user),
         {:ok, conversation} <- Conversation.Mutator.decrement_unread_count(conversation, length(messages)) do
      :ok
    end
```

The printed warning can be confusing because it tells us about the variable definition in case block which is generated with macros:
```
warning: variable "conversation" is unused

Note variables defined inside case, cond, fn, if and similar do not leak. If you want to conditionally override an existing
 variable "conversation", you will have to explicitly return the variable. For example:

    if some_condition? do
      atom = :one
    else
      atom = :two
    end

should be written as


    atom =
      if some_condition? do
        :one
      else
        :two
      end

Unused variable found at:
  lib/my_app_domain/messaging.ex:194
```

Replacing `conversation` with `_conversation` would silence the whole output.

### Expected behavior

Only `warning: variable "conversation" is unused` warning should be printed without information about variables in case, cond and other blocks.

fertapric:
Closes #8285

This warning has to be removed because all of the forms were unified in v1.7, making case/with/if to behave the same. That leads to undesired warnings in cases such as:

```elixir
a = 1

with {:ok, a} <- my_fun(a) do
  :ok
end
```

### DIFF
--- a/lib/elixir/src/elixir_env.erl
+++ b/lib/elixir/src/elixir_env.erl
@@ -105,7 +105,7 @@ merge_vars(V1, V2) ->
 %% UNUSED VARS
 
 check_unused_vars(#{unused_vars := Unused} = E) ->
-  [elixir_errors:form_warn([{line, Line}], ?key(E, file), ?MODULE, {unused_var, Name, false}) ||
+  [elixir_errors:form_warn([{line, Line}], ?key(E, file), ?MODULE, {unused_var, Name}) ||
     {{{Name, _}, _}, Line} <- maps:to_list(Unused), Line /= false, not_underscored(Name)],
   E.
 
@@ -136,12 +136,11 @@ merge_and_check_unused_vars(Unused, ClauseUnused, E) ->
 
           %% Otherwise we must warn
           _ ->
-            {{Name, _} = Pair, _} = Key,
+            {{Name, _}, _} = Key,
 
             case not_underscored(Name) of
               true ->
-                IsShadowing = maps:is_key(Pair, ?key(E, current_vars)),
-                Warn = {unused_var, Name, IsShadowing},
+                Warn = {unused_var, Name},
                 elixir_errors:form_warn([{line, ClauseValue}], ?key(E, file), ?MODULE, Warn);
 
               false ->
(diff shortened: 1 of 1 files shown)


whatyouhide:
Thanks @fertapric! 💟

adrianomitre:
The way I see it, this was an unfortunate implementation-driven decision which disregarded usability concerns, especially given the warning was related to a [breaking change made in Elixir 1.7+](https://github.com/elixir-lang/elixir/issues/8076#issuecomment-638333947) that changed behaviour of code samples on Programming Elixir 1.6 book by Dave Thomas (see [erratum](https://pragprog.com/titles/elixir16/errata#row-86784)).

Assuming the breaking change will not be reverted, I hope that at least the warning can be brought back in future Elixir versions.

josevalim:
Please read this comment on why we needed this change to fix important bugs in the language: https://github.com/elixir-lang/elixir/issues/7403#issuecomment-368902698

adrianomitre:
Thanks for sharing the reference with the motivation behing the change. The _what_ seems reasonable. The unfortunate was _how_ it was carried out, in particular the combination (i) breaking change in minor version, violating both semantic versioning and stated Elixir deprecations policy plus (ii) no clear warning. Either bumping the language major version or keeping a meaningful warning ("shadowing") would have been much better.

From [Compatibility and Deprecations — Elixir v1.10.3 - Deprecations - Policy](https://hexdocs.pm/elixir/compatibility-and-deprecations.html#deprecations):
> Elixir deprecations happen in 3 steps: [...] 3.The feature is removed. This can only happen on major releases. This means deprecated features in Elixir v1.x shall only be removed by Elixir v2.x.

Considering where we are right now, I can only wish for a meaningful warning to be reinstated.

josevalim:
To quote the same document:

> Bugs: if an API has undesired behaviour, a program that depends on the buggy behaviour may break if the bug is fixed. We reserve the right to fix such bugs.

Having unreliable unused variable warnings or, even worse, disabling tail recursion in certain scenarios, are important bugs that would be impossible to fix without removing this buggy behaviour. We emitted a warning for 2 years before fixing the bug but there is no way to fix the other issues without removing the broken behaviour altogether. Which is why there is no option to reinstate said behaviour. If we did so, we would have to reintroduce bugs previously fixed.

We definitely discussed shipping a major version over this, but we chose to sit on the "bug fix" side of the fence. I understand others would have done difference choices too. I am just pointing out our perspective on this (and mentioning the 2 year long prior warning).

adrianomitre:
> I understand others would have taken a different choice too. I am just pointing out our perspective on this (and mentioning the 2 year long prior warning).

Number of years does not seem to be a relevant metric. The last release of the Programming Elixir 1.6 book is older than 2 years, by the way. The warning shoud have been kept on undefinitely in Elixir 1.x and I have yet to see any user-centric (e.g., usability) reason for doing otherwise.

josevalim:
By the time the book was written, the warning was already there for 1.5 years. If fixing compiler bugs that can directly affect users doesn't count in your definition of usability, we will have to agree to disagree.

PR #8289, opened 2018-10-16
Allow @type, @opaque, @typep, @spec, @callback, and @macrocallback to be used as module attributes

### LINKED ISSUE #8085: @spec module attribute no longer accessible in 1.7
### Current behavior

```elixir
defmodule Foo do
  @spec x() :: integer()
  def x(), do: 5
  def spec(), do: @spec
end

iex> Foo.spec() #=> nil
```

### Expected behavior

```elixir
iex> Foo.spec() #=>
[
  {:spec, {:::, [line: 2], [{:x, [line: 2], nil}, {:integer, [line: 2], []}]},
   {Foo, {2, 1}}}
]
```

fertapric:
Closes #8085

### DIFF
--- a/lib/elixir/lib/kernel/typespec.ex
+++ b/lib/elixir/lib/kernel/typespec.ex
@@ -99,8 +99,9 @@ defmodule Kernel.Typespec do
   Invoked by `Kernel.@/1` expansion.
   """
   def deftypespec(:spec, expr, _line, _file, module, pos) do
-    {_set, bag} = :elixir_module.data_tables(module)
+    {set, bag} = :elixir_module.data_tables(module)
     store_typespec(bag, :spec, {expr, pos})
+    store_module_attribute(set, :spec, {expr, pos})
   end
 
   def deftypespec(kind, expr, line, _file, module, pos)
@@ -150,6 +151,11 @@ defmodule Kernel.Typespec do
     :ets.lookup_element(bag, key, 2)
   catch
     :error, :badarg -> []
+  end
+
+  defp store_module_attribute(set, key, value) do
+    :ets.insert(set, {key, value, nil})
+    :ok
   end
 
   defp store_typespec(bag, key, value) do
--- a/lib/elixir/lib/kernel/typespec.ex
+++ b/lib/elixir/lib/kernel/typespec.ex
@@ -99,9 +99,8 @@ defmodule Kernel.Typespec do
   Invoked by `Kernel.@/1` expansion.
   """
   def deftypespec(:spec, expr, _line, _file, module, pos) do
-    {set, bag} = :elixir_module.data_tables(module)
+    {_set, bag} = :elixir_module.data_tables(module)
     store_typespec(bag, :spec, {expr, pos})
-    store_module_attribute(set, :spec, {expr, pos})
   end
 
   def deftypespec(kind, expr, line, _file, module, pos)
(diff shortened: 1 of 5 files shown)


lib/elixir/lib/kernel/typespec.ex
```diff
@@ -99,8 +99,9 @@ defmodule Kernel.Typespec do
   Invoked by `Kernel.@/1` expansion.
   """
   def deftypespec(:spec, expr, _line, _file, module, pos) do
-    {_set, bag} = :elixir_module.data_tables(module)
+    {set, bag} = :elixir_module.data_tables(module)
     store_typespec(bag, :spec, {expr, pos})
+    store_module_attribute(set, :spec, {expr, pos})
```
josevalim: How did it work on Elixir v1.7? Did `@spec` return only the last `@spec` or would it return all of them?

fertapric: It accumulates, returning all the defined ones until that point.

I also notices it includes everything into a `{:spec, ...}` tuple.

(I guess you meant v1.6, because v1.7 does not allow to use `@spec` as module attribute)

josevalim: Yes, v1.6. :) We should mirror the return type there. I am wondering if instead if we should change `Module.get_attribute` and make it return the old format for `@spec`. We should also probably forbid `Module.put_attribute` with a spec (if we haven't yet).

fertapric: It sounds good to me!

Should `Module.get_attribute` be the one responsible for returning the old format, or should we store the old format directly?

josevalim: I would change Module.get_attribute to compute the old format. Very few projects actually want to read it. So I would "punish" the readers rather than the writers.

lib/elixir/lib/module.ex
```diff
@@ -1726,7 +1726,16 @@ defmodule Module do
   def get_attribute(module, key, line) when is_atom(key) do
     assert_not_compiled!(:get_attribute, module)
     {set, bag} = data_tables_for(module)
+    get_attribute(module, key, line, set, bag)
+  end
+
+  defp get_attribute(_module, :spec, _line, _set, bag) do
+    specs = bag_lookup_element(bag, :spec, 2)
```
josevalim: Hrm, I am wondering. Maybe all we need to do is to add `spec` and friends as an accumulate attribute here? https://github.com/elixir-lang/elixir/blob/343b65c112fc8809d9b7445c92d7a403cdcdab09/lib/elixir/src/elixir_module.erl#L265

josevalim: Nah, it won't work because all accumulate attributes are stored as `{accumulate, Key}`. I am wondering if it is worth to store spec/type/callback and friends as `{accumulate, Key}` but the main issue is that type has a different return type and it won't work transparently for @type, @opaque and @typep.

fertapric: Oh! I was only focusing on `@spec`, but I see it was working previously with the rest of attributes (accumulating). 

> but the main issue is that type has a different return type and it won't work transparently for @type, @opaque and @typep.

I think we can make it work if we change `translate_typespecs_for_module/2` to get `:type`, `:typep` and `:opaque` from `bag` and adjust the rest of the functions to use `{kind, {expr, pos}}` instead of `{:type, {kind, expr, pos}}`. What do you think?

josevalim:
Yeah, I think we can try it. But if we are going in this direction, then I
would always store it as {kind, expr, pos} for all of them. With no
nesting. Then we can do a lookup instead of a lookup_element to return the
values.
-- 


josevalim:

wende:
Awesome! What Elixir version would that be planned to be released with?

josevalim:
v1.8 unless someone backports this to v1.7. :)

fertapric:
Backported in https://github.com/elixir-lang/elixir/pull/8327

PR #8292, opened 2018-10-19
Update README.md

anshul77:


README.md
```diff
@@ -32,7 +32,7 @@ If Elixir fails to build (specifically when pulling in a new version via
 `make clean`, then `make test`.
 
 If tests pass, you are ready to move on to the [Getting Started guide][1]
-or to try Interactive Elixir by running `bin/iex` in your terminal.
+or to try our Interactive Elixir by running `bin/iex` in your terminal.
```
davydog187: I don't think `our` makes sense in this context

whatyouhide:
I agree that in this case "our" doesn't really add anything to the sentence, and we never (as far as I know) talk about Elixir as "our" project or refer to tools as "our" tools. Closing this for now but if others think this is a good change, feel free to reopen it or add a comment here to keep the discussion going :). Thanks for the contribution @anshul77! 💟

PR #8293, opened 2018-10-20
Allways call ExUnit.EventManager.module_finished/2

eksperimental:
While ExUnit.RunnerStats doesn't do anything,
it may be needed by other handlers.

### DIFF
--- a/lib/ex_unit/lib/ex_unit/runner.ex
+++ b/lib/ex_unit/lib/ex_unit/runner.ex
@@ -141,11 +141,14 @@ defmodule ExUnit.Runner do
           nil
       end
 
-    if pending_tests do
-      test_module = %{test_module | tests: Enum.reverse(finished_tests, pending_tests)}
-      EM.module_finished(config.manager, test_module)
-    end
+    test_module =
+      if pending_tests do
+        %{test_module | tests: Enum.reverse(finished_tests, pending_tests)}
+      else
+        test_module
+      end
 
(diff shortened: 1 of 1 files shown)


josevalim:
@eksperimental so we do this by design. Since we haven't run all tests in that module, I am not sure we can consider it to be finished. This also mirrors the behaviour of tests.

To be honest, I can see this going either way. Just pointing out the current behaviour is intentional and mirrors tests, so unless we have precise examples of otherwise, I would like to keep this as is. :)

josevalim:
Maybe we should ping maintaining other formatters, such as the junit one, and ask them to give it a try and provide feedback.

eksperimental:
> Maybe we should ping maintaining other formatters, such as the junit one, and ask them to give it a try and provide feedback.

@josevalim let's just wait until I submit a PR related to a bug fix, because there are changes related to this

eksperimental:
> @eksperimental so we do this by design. Since we haven't run all tests in that module, I am not sure we can consider it to be finished. This also mirrors the behaviour of tests.

But isn't it the exact opposite? There's only one call to `EM.module_finished` in the current code, and it happens when there are pending tests.

eksperimental:
@josevalim just came back to say that I misunderstood the logic of `if pending_tests do`. If pending_tests is `[]` the `if` still gets executed. I will add a comment to avoid other people the trouble

PR #8294, opened 2018-10-20
Touch up max_failures tests in ExUnit

eksperimental:
- Use setup_all instead of "setup all".
- Rename tests from "pass" to "fail" as the names were inverted.

### DIFF
--- a/lib/ex_unit/test/ex_unit_test.exs
+++ b/lib/ex_unit/test/ex_unit_test.exs
@@ -625,7 +625,7 @@ defmodule ExUnitTest do
       assert output =~ "3 tests, 2 failures"
     end
 
-    test ":max_failures on setup all errors" do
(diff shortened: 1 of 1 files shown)


lib/ex_unit/test/ex_unit_test.exs
```diff
@@ -625,7 +625,7 @@ defmodule ExUnitTest do
       assert output =~ "3 tests, 2 failures"
     end
 
-    test ":max_failures on setup all errors" do
+    test ":max_failures on setup_all errors" do
```
fertapric: `setup_all/1,2`?

josevalim: I think it is fine as is, it is not mean to be a complete reference and `setup_all` is enough to indicate the test purpose.

josevalim:

PR #8295, opened 2018-10-20
ExUnit.Runner - Pass config seed and not the whole config when not need

eksperimental:
Avoid passing the whole config when only the seed is needed

### DIFF
--- a/lib/ex_unit/lib/ex_unit/runner.ex
+++ b/lib/ex_unit/lib/ex_unit/runner.ex
@@ -285,15 +285,15 @@ defmodule ExUnit.Runner do
   defp spawn_test(config, test, context) do
     parent_pid = self()
     timeout = get_timeout(config, test.tags)
-    {test_pid, test_ref} = spawn_test_monitor(config, test, parent_pid, context)
+    {test_pid, test_ref} = spawn_test_monitor(test, parent_pid, context, config.seed)
     test = receive_test_reply(test, test_pid, test_ref, timeout)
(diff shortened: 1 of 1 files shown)


lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -285,15 +285,15 @@ defmodule ExUnit.Runner do
   defp spawn_test(config, test, context) do
     parent_pid = self()
     timeout = get_timeout(config, test.tags)
-    {test_pid, test_ref} = spawn_test_monitor(config, test, parent_pid, context)
+    {test_pid, test_ref} = spawn_test_monitor(test, parent_pid, context, config.seed)
```
fertapric: I'm 👎 for this change. I liked better when `spawns_test/3` did not know that the only thing that `spawn_test_monitor/4` needs is the seed configuration.

josevalim:
Agreed with @fertapric :) Thanks @eksperimental !

lib/ex_unit/lib/ex_unit/runner.ex
eksperimental: yeah, it's understandable.
the original idea was the replace generate_test_seed, which only needs the seed.

eksperimental:
@josevalim could you please reopen the PR? I will submit a new one just changing `generate_test_seed`

josevalim:
I would say it is also fine as is. :) thanks!
-- 

PR #8296, opened 2018-10-20
Change Enum.reduce_while/3

eksperimental:
... so the integer in the "if" clause is different than the return value


josevalim:

PR #8305, opened 2018-10-20
Rename Calendar module files according to the convention

eksperimental:


eksperimental:
Date.Range breaks the rule big time since it is placed in a different directory,
The other two are just missing the underscore.

josevalim:
It is fine as is. :) If we were going to follow the convention, they wouldn't even be inside the `calendar` folder. I also think we agreed that we would write `DateTime` as `datetime`. Or did we settle on `date_time`? I can't remember.

wojtekmach:
I'd keep it as `datetime` even if it's arbitrary; GItHub is `github`, not `git_hub`!

eksperimental:
> I also think we agreed that we would write `DateTime` as `datetime`. Or did we settle on `date_time`? I can't remember.

I missed that discussion.

PR #8306, opened 2018-10-20
Make folder path example more obvious

E1337Kat:
"ebin" is not a common path, but a path with "src" is well known.

### DIFF
--- a/lib/iex/lib/iex/helpers.ex
+++ b/lib/iex/lib/iex/helpers.ex
@@ -152,7 +152,7 @@ defmodule IEx.Helpers do
 
   ## Examples
 
-      iex> c(["foo.ex", "bar.ex"], "ebin")
+      iex> c(["foo.ex", "bar.ex"], "src")
       [Foo, Bar]
 
       iex> c("baz.ex")


michalmuskala:
`ebin` is a common place for storing compiled modules, I believe it's even required by the standard release handlers. Since mix compiles in `_build`, you can find the regular "release" structure of the application, including the `ebin` directory, there.

That said, it's not something very common (since we don't usually compile by hand), so maybe we could find some better example - I'm not sure `src` is a good place, though, I don't think compiled code should live with source code.

E1337Kat:
I see, I did not read the full description and assumed it was a path to the file to compile. I can adjust to better show what it means. Probably `build` or `release`, or even `_build/ebin`could be very descriptive. Is there a preference?

whatyouhide:
Maybe go simpler, and have `c(..., "my_compiled_files")`?

josevalim:
@whatyouhide well, the issue is that "my_compiled_files" is really "ebin" for Elixir. Maybe, instead of changing the example, we could update the docs and say something like: "In the example below, we pass a directory to where the `c` helper will write the `.beam` files to. This directory is typically named "ebin" for Erlang/Elixir systems".

E1337Kat:
I really like that. It explains the purpose of the "ebin" directory, and makes it clear what the example will do (compile and place into ebin directory). I will implement that change real quick and change my original change back. I plan to have the sentence injected into the middle of the paragraph right after explaining passing an empty string saves files to current directory. The last sentence though then seems out of place from a grammatical standpoint. Going from and explanatory sentence to then a note about a different area reads a bit... weird. This is how it would read:

```
@doc """
  Compiles the given files.

  It expects a list of files to compile and an optional path to write
  the compiled code to. By default files are in-memory compiled.
  To write compiled files to a current directory use empty string "" for the path. 
  In the example below, we pass a directory to where the c helper will 
  write the .beam files to. This directory is typically named "ebin" for 
  Erlang/Elixir systems. 
  When compiling one file, there is no need to wrap it in a list.

  It returns the names of the compiled modules.

  If you want to recompile an existing module, check `r/1` instead.

  ## Examples

      iex> c(["foo.ex", "bar.ex"], "ebin")
      [Foo, Bar]

      iex> c("baz.ex")
      [Baz]

  """
```

josevalim:
You can change the order of the examples and move the explanatory text to
inside the examples section. :)
-- 

PR #8309, opened 2018-10-20
Update use of numbers as adjectives

eksperimental:
- Replace "two element tuples" with "two-element tuples" for every number in 1..10
- Replace "2-element tuples" with "two-element tuples" for every number
- "1-based" with "one-based"
- "0-based" with "zero-based"


josevalim:

PR #8312, opened 2018-10-21
Fix 2 bugs in ExUnit.Runner

eksperimental:
There were 2 bugs in the current implementation

1. After the introduction of max-failures, the printing of the number of invalid tests went missing.

2. Invalid and Excluded (pending) tests were counted twice if setup_all failed.
A module with 1 skipped test and 1 excluded and 6 tests in it, it would be reported with a total of 8 tests.
This happened before max-failures was introduced.

A test file is provided including the new tests:
https://gist.github.com/eksperimental/30073f1b190fd2d34c75eadb92a41551

This PR takes the following approach:

1. We determine before running any test which are the tests to be skipped and excluded,
therefore:
  A. these tests are always part of the total number of tests in the stats.
  B. these tests are not counted as failures or invalid ones if setup_all fails.

2.
 A. We call `EM.test_started/2` and `EM.test_finished/2` for every test in a module that is run.
 B. We call `EM.module_started/2` and `EM.module_finished/2` for every module that is run.

 This way it is upto the implementation of the Event Manager handler to choose what to do to.
 The approach that we take in `ExUnit.CLIFormatter` is to evaluate the calls for test_started and test_finished,
 and to ignore the ones for `module_started` and `module_finished`.

### DIFF
--- a/lib/ex_unit/lib/ex_unit/runner.ex
+++ b/lib/ex_unit/lib/ex_unit/runner.ex
@@ -376,12 +376,12 @@ defmodule ExUnit.Runner do
   defp process_max_failures(%{max_failures: :infinity}, _), do: :no
 
   defp process_max_failures(config, %ExUnit.TestModule{state: {tag, _}, tests: tests})
-       when tag in [:failed, :invalid] do
+       when tag in [:failed] do
     process_max_failures(config.stats_pid, config.max_failures, length(tests))
   end
 
   defp process_max_failures(config, %ExUnit.Test{state: {tag, _}})
-       when tag in [:failed, :invalid] do
+       when tag in [:failed] do
     process_max_failures(config.stats_pid, config.max_failures, 1)
   end
 
--- a/lib/ex_unit/lib/ex_unit/runner.ex
+++ b/lib/ex_unit/lib/ex_unit/runner.ex
@@ -122,33 +122,34 @@ defmodule ExUnit.Runner do
     tests = prepare_tests(config, test_module.tests)
     {excluded_and_skipped_tests, to_run_tests} = Enum.split_with(tests, & &1.state)
 
+    for excluded_or_skipped_test <- excluded_and_skipped_tests do
+      EM.test_started(config.manager, excluded_or_skipped_test)
+      EM.test_finished(config.manager, excluded_or_skipped_test)
+    end
+
     {test_module, invalid_tests, finished_tests} = spawn_module(config, test_module, to_run_tests)
 
     pending_tests =
       case process_max_failures(config, test_module) do
         :no ->
-          invalid_tests ++ excluded_and_skipped_tests
+          invalid_tests
 
         {:reached, n} ->
-          Enum.take(invalid_tests, n) ++ excluded_and_skipped_tests
+          Enum.take(invalid_tests, n)
 
         :surpassed ->
           nil
       end
 
-    test_module =
-      if pending_tests do
-        for pending_test <- pending_tests do
-          EM.test_started(config.manager, pending_test)
-          EM.test_finished(config.manager, pending_test)
-        end
-
-        %{test_module | tests: Enum.reverse(finished_tests, pending_tests)}
-      else
-        test_module
+    if pending_tests do
+      for pending_test <- pending_tests do
+        EM.test_started(config.manager, pending_test)
+        EM.test_finished(config.manager, pending_test)
       end
 
-    EM.module_finished(config.manager, test_module)
+      test_module = %{test_module | tests: Enum.reverse(finished_tests, pending_tests)}
+      EM.module_finished(config.manager, test_module)
+    end
 
     send(config.runner_pid, {self(), :module_finished})
   end
--- a/lib/ex_unit/lib/ex_unit/runner.ex
+++ b/lib/ex_unit/lib/ex_unit/runner.ex
@@ -378,13 +378,11 @@ defmodule ExUnit.Runner do
 
   defp process_max_failures(%{max_failures: :infinity}, _), do: :no
 
-  defp process_max_failures(config, %ExUnit.TestModule{state: {tag, _}, tests: tests})
-       when tag in [:failed] do
+  defp process_max_failures(config, %ExUnit.TestModule{state: {:failed, _}, tests: tests}) do
     process_max_failures(config.stats_pid, config.max_failures, length(tests))
   end
 
-  defp process_max_failures(config, %ExUnit.Test{state: {tag, _}})
-       when tag in [:failed] do
+  defp process_max_failures(config, %ExUnit.Test{state: {:failed, _}}) do
     process_max_failures(config.stats_pid, config.max_failures, 1)
   end
 
--- a/lib/ex_unit/lib/ex_unit/runner.ex
+++ b/lib/ex_unit/lib/ex_unit/runner.ex
@@ -127,24 +127,28 @@ defmodule ExUnit.Runner do
     pending_tests =
       case process_max_failures(config, test_module) do
         :no ->
-          for test <- excluded_tests do
-            EM.test_started(config.manager, test)
-            EM.test_finished(config.manager, test)
-          end
-
           invalid_tests ++ excluded_tests
 
         {:reached, n} ->
-          Enum.take(invalid_tests, n)
+          Enum.take(invalid_tests, n) ++ excluded_tests
 
         :surpassed ->
           nil
       end
 
-    if pending_tests do
-      test_module = %{test_module | tests: Enum.reverse(finished_tests, pending_tests)}
-      EM.module_finished(config.manager, test_module)
-    end
+    test_module =
+      if pending_tests do
+        for pending_test <- pending_tests do
+          EM.test_started(config.manager, pending_test)
+          EM.test_finished(config.manager, pending_test)
+        end
+
+        %{test_module | tests: Enum.reverse(finished_tests, pending_tests)}
+      else
+        test_module
+      end
+
+    EM.module_finished(config.manager, test_module)
 
     send(config.runner_pid, {self(), :module_finished})
   end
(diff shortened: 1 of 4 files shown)


lib/ex_unit/test/ex_unit_test.exs
```diff
@@ -498,6 +498,77 @@ defmodule ExUnitTest do
     ExUnit.configure(seed: global_seed)
   end
 
+  @doc """
+    Skipped and excluded tests should be included in the stats as well as printed to stdout.
+    On the other hand, invalid tests should be marked as failures in the stats,
+    but still be printed as "invalid" to stdout.
+
+    If setup_all fails, the skipped and excluded tests should not be counted as invalid or failures.
```
josevalim: Please remove indention.

lib/ex_unit/test/ex_unit_test.exs
```diff
@@ -720,4 +799,9 @@ defmodule ExUnitTest do
   defp max_failures_reached_msg() do
     "--max-failures reached, aborting test suite"
   end
+
+  # Strips escape-sequence colors from IO.ANSI strings
+  defp strip(string) do
+    String.replace(string, ~r/\e\[(\d+;?)+m/, "")
+  end
```
josevalim: We don't need this strip function. :)

lib/ex_unit/test/ex_unit_test.exs
```diff
@@ -498,6 +498,77 @@ defmodule ExUnitTest do
...
+      capture_io(fn ->
+        predictable_ex_unit_start(exclude: [:exclude])
+        assert ExUnit.run() == %{total: 6, failures: 4, excluded: 1, skipped: 1}
+      end)
+
+    refute output =~ max_failures_reached_msg()
+    assert strip(output) =~ "6 tests, 0 failures, 1 excluded, 4 invalid, 1 skipped"
+  end
+
+  test "check for invalid and excluded tests in stdout" do
```
josevalim: What is this test doing  that is different from the precious one?

lib/ex_unit/test/ex_unit_test.exs
```diff
@@ -498,6 +498,77 @@ defmodule ExUnitTest do
...
   end
 
+  @doc """
+    Skipped and excluded tests should be included in the stats as well as printed to stdout.
+    On the other hand, invalid tests should be marked as failures in the stats,
+    but still be printed as "invalid" to stdout.
+
+    If setup_all fails, the skipped and excluded tests should not be counted as invalid or failures.
+  """
+  test "skipped and excluded" do
```
josevalim: We already have tests for skipped i this module, I would propose to rename this "accounts skipped and excluded on setup_all errors.

lib/ex_unit/test/ex_unit_test.exs
```diff
@@ -635,23 +710,27 @@ defmodule ExUnitTest do
...
         @tag :skip
         test "skipped", do: assert(true)
+
         test "pass #{__ENV__.line}", do: assert(true)
         test "pass #{__ENV__.line}", do: assert(true)
         test "fail #{__ENV__.line}", do: assert(false)
         test "fail #{__ENV__.line}", do: assert(false)
+
+        @tag :exclude
+        test "exclude me please", do: assert(true)
```
josevalim: Nice! :+1: We should probably add `:exclude` to all other `:max_failures` test cases here, like we do with `@tag :skip`.

lib/ex_unit/test/ex_unit_test.exs
josevalim: Or, even better, this file has no test for "setup_all", so maybe we should call it "setup_all has invalid, skipped and excluded tests". :)

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -127,24 +127,28 @@ defmodule ExUnit.Runner do
...
-          for test <- excluded_tests do
-            EM.test_started(config.manager, test)
-            EM.test_finished(config.manager, test)
-          end
-
           invalid_tests ++ excluded_tests
 
         {:reached, n} ->
-          Enum.take(invalid_tests, n)
+          Enum.take(invalid_tests, n) ++ excluded_tests
```
josevalim: The issue with this approach is that, if we have a formatter that also prints excluded_tests, it means that we will print N excluded tests after max_failures. If the goal is to account `excluded_tests` more reliably, then I would suggest to run all excluded_tests before everything else, before we call `spawn_module`. What do you think?

lib/ex_unit/test/ex_unit_test.exs
eksperimental: the invalid counter shows these codes, how to do check for that?

lib/ex_unit/test/ex_unit_test.exs
eksperimental: I will leave that for another PR once this is merged

lib/ex_unit/lib/ex_unit/runner.ex
eksperimental: We can print the excluded and skipped tests as soon as they are determined. I like that approach. Later in another loop we print the invalid ones.
But I fail to see how now "N excluded tests"  are printed, as I think only "N invalid tests" are printed

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -376,7 +380,8 @@ defmodule ExUnit.Runner do
     process_max_failures(config.stats_pid, config.max_failures, length(tests))
   end
 
-  defp process_max_failures(config, %ExUnit.Test{state: {:failed, _}}) do
+  defp process_max_failures(config, %ExUnit.Test{state: {tag, _}})
+       when tag in [:failed, :invalid] do
```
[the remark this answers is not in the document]
eksperimental: true, and neither for %ExUnit.TestModule. I guess I can remove :invalid for both

lib/ex_unit/test/ex_unit_test.exs
eksperimental: sorry, it is an accident duplicate

lib/ex_unit/test/ex_unit_test.exs
eksperimental: how about "setup_all fails and module has skipped and excluded tests", since its it not setup_all that has invalid test but the cause of them being rendered as invalid

lib/ex_unit/test/ex_unit_test.exs
eksperimental: Are you suggesting doing something like this?
https://github.com/elixir-lang/elixir/blob/b611983d57258a2cff82de48216e9f66bd566268/lib/ex_unit/test/ex_unit_test.exs#L356-L358

the problem that approach is that we are not seeing the full line in the test.
Actually, I am thinking about submitting a new PR after this that checks for the output message until the end of the line (including `\n`). Have we had theses checks, they bug that missed the skipped messages wouldn't have been introduced.

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -127,24 +127,28 @@ defmodule ExUnit.Runner do
...
+          EM.test_started(config.manager, pending_test)
+          EM.test_finished(config.manager, pending_test)
+        end
+
+        %{test_module | tests: Enum.reverse(finished_tests, pending_tests)}
+      else
+        test_module
+      end
+
+    EM.module_finished(config.manager, test_module)
```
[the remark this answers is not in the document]
eksperimental: moved out.
I will reevaluate this as I misunderstood the logic of `if pending_tests do`

lib/ex_unit/lib/ex_unit/runner.ex
eksperimental: Please have a look and tell me if the code now solves the issue that you mentioned regarding the N excluded tests

lib/ex_unit/test/ex_unit_test.exs
josevalim: Can't we disable ansi codes when running the test suite?

lib/ex_unit/test/ex_unit_test.exs
josevalim: Your call. But since we are fixing tests related to the exclusion of exclude, adding those now may help us find more failures.

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -372,11 +377,12 @@ defmodule ExUnit.Runner do
   defp process_max_failures(%{max_failures: :infinity}, _), do: :no
 
   defp process_max_failures(config, %ExUnit.TestModule{state: {tag, _}, tests: tests})
-       when tag in [:failed, :invalid] do
+       when tag in [:failed] do
```
josevalim: Let's pattern match on `:failed` then instead of the guard clause. :)

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -372,11 +377,12 @@ defmodule ExUnit.Runner do
...
     process_max_failures(config.stats_pid, config.max_failures, length(tests))
   end
 
-  defp process_max_failures(config, %ExUnit.Test{state: {:failed, _}}) do
+  defp process_max_failures(config, %ExUnit.Test{state: {tag, _}})
+       when tag in [:failed] do
```
josevalim: Same here.

josevalim:
@eksperimental thanks, this is looking great! Just another round of comments and we should be good to go! :heart:

lib/ex_unit/test/ex_unit_test.exs
eksperimental: Didn't think of it.
fixed!

lib/ex_unit/test/ex_unit_test.exs
eksperimental: I will add it then

eksperimental: done

lib/ex_unit/lib/ex_unit/runner.ex
eksperimental: fixed

lib/ex_unit/lib/ex_unit/runner.ex

eksperimental:
@josevalim all suggestions have been implement.

If you want me to rebase them in 2 or 3 commits that group the changes, I can do that, as well as updating the initial commit message to reflect the final changes

josevalim:

eksperimental:
@josevalim Thank you for your guidance!

PR #8313, opened 2018-10-21
Improvements to the Version module documentation

eksperimental:
- General improvements to docs
- Minor grammar corrections
- Add doc test examples to Version.match?/3 using :allow_pre option
- List all the operators allowed in requirements.
- Mention that the PATH part of the version can be used as operand with ~>
- Make a clarification that "`~>` will never include pre-release versions of its upper bound,
  regardless of the usage of the `allow_pre:` option", and add tests to verify this
- The requirement table has been update:
  - It doesn't use the consecutive PATCH number in the example, so `~> 2.1.2` will not use
    `2.1.3-dev` but `2.1.6-dev` to make it that the upper bound is within the MINOR part
    not the PATCH one
  - Remove "-" and replace with "`true` or `false`" as it is not clear what "-" means in that example,
   and it could be understood that the option is not allowed in that example.
  - Updated to "`true` or `false`" the cases whether both options return the same value.
  - Added an additional example: `>= 2.1.0` | `2.2.0-dev` | `true` | `true`

### DIFF
--- a/lib/elixir/lib/version.ex
+++ b/lib/elixir/lib/version.ex
@@ -69,7 +69,7 @@ defmodule Version do
   `~> 2.0`       | `>= 2.0.0 and < 3.0.0`
   `~> 2.1`       | `>= 2.1.0 and < 3.0.0`
 
-  The requirement operand after the `~>` is allows to omit the PATCH part of the version,
+  The requirement operand after the `~>` is allowed to omit the PATCH part of the version,
   allowing us to express `~> 2.1` or `~> 2.1-dev`, something that wouldn't be allowed
   when using the common comparison operators.
 
--- a/lib/elixir/lib/version.ex
+++ b/lib/elixir/lib/version.ex
@@ -35,9 +35,9 @@ defmodule Version do
   ## Requirements
 
   Requirements allow you to specify which versions of a given
-  dependency you are willing to work against. Requirements support common
-  operators like `>=`, `<=`, `>`, `==`, and friends that
-  work as one would expect:
+  dependency you are willing to work against. Requirements support the common
+  operators such as `>`, `>=`, `<`, `<=`, `==`, `!=` that work as one would expect,
+  and additionally the special operator `~>` described in detail further bellow.
 
       # Only version 2.0.0
       "== 2.0.0"
@@ -55,10 +55,12 @@ defmodule Version do
 
       "~> 2.0.0"
 
-  `~>` will never include pre-release versions of its upper bound.
+  `~>` will never include pre-release versions of its upper bound,
+  regardless of the usage of the `allow_pre:` option, or whether the operand
+  is a pre-release version.
   It can also be used to set an upper bound on only the major
   version part. See the table below for `~>` requirements and
-  their corresponding translation.
+  their corresponding translations.
 
   `~>`           | Translation
   :------------- | :---------------------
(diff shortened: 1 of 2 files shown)


lib/elixir/lib/version.ex
```diff
@@ -31,9 +31,9 @@ defmodule Version do
   ## Requirements
 
   Requirements allow you to specify which versions of a given
-  dependency you are willing to work against. Requirements support common
-  operators like `>=`, `<=`, `>`, `==`, and friends that
-  work as one would expect:
+  dependency you are willing to work against. Requirements support the common
+  operators such as `>`, `>=`, `<`, `<=`, `==`, `!=` that work as one would expect,
+  and additionally the special operator `~>` described in detail further bellow.
```
fertapric: s/bellow/below

lib/elixir/lib/version.ex
```diff
@@ -51,10 +51,12 @@ defmodule Version do
 
       "~> 2.0.0"
 
-  `~>` will never include pre-release versions of its upper bound.
+  `~>` will never include pre-release versions of its upper bound,
+  regardless of the usage of the `allow_pre:` option, or whether the operand
```
fertapric: Maybe "of the `:allow_pre` option"?

lib/elixir/lib/version.ex
```diff
@@ -51,10 +51,12 @@ defmodule Version do
...
+  is a pre-release version.
   It can also be used to set an upper bound on only the major
```
fertapric: Turn this into a single paragraph "...is a pre-release version. It can also..."? 

(or add a new line so they are two separate paragraphs?)

lib/elixir/lib/version.ex
```diff
@@ -64,23 +66,28 @@ defmodule Version do
   `~> 2.0`       | `>= 2.0.0 and < 3.0.0`
   `~> 2.1`       | `>= 2.1.0 and < 3.0.0`
 
-  When `allow_pre: false` is set, the requirement will not match a
-  pre-release version unless the operand is a pre-release version.
+  The requirement operand after the `~>` is allows to omit the PATCH part of the version,
+  allowing us to express `~> 2.1` or `~> 2.1-dev`, something that wouldn't be allowed
+  when using the common operators.
```
fertapric: "when using common operators."?

lib/elixir/lib/version.ex
```diff
@@ -64,23 +66,28 @@ defmodule Version do
...
+
+  When the option `allow_pre: false` is set in `Version.match?/3`, the requirement
```
fertapric: Maybe "When the `:allow_pre` option is set to `false` in `Version.match?/3`, the requirement..."?

lib/elixir/lib/version.ex
```diff
@@ -64,23 +66,28 @@ defmodule Version do
   `~> 2.0`       | `>= 2.0.0 and < 3.0.0`
   `~> 2.1`       | `>= 2.1.0 and < 3.0.0`
 
-  When `allow_pre: false` is set, the requirement will not match a
-  pre-release version unless the operand is a pre-release version.
+  The requirement operand after the `~>` is allows to omit the PATCH part of the version,
```
fertapric: What do you think about?

> `~>` only requires the MAJOR and MINOR parts of the version.

lib/elixir/lib/version.ex
eksperimental: it mentions them at the beginning of the sections. I am referring to them as now "common comparison operators".
Or should we refer to the as "regular" or "standard" comparison operators?

lib/elixir/lib/version.ex
eksperimental: Beautiful!

lib/elixir/lib/version.ex
eksperimental: Yes, it's one paragraph. thank you.

lib/elixir/lib/version.ex
eksperimental: Sweet. I will submit a PR fixing all similar issues. Thanks for pointing it out @fertapric

lib/elixir/lib/version.ex
eksperimental: fixed

josevalim:

PR #8319, opened 2018-10-22
Check for line ending in output messages in ex_unit_tests.exs

eksperimental:
... and beginning of line when applicable.

This makes sure we check for the full line,
avoiding unexpected results appearing at the end of the line.

Related discussion:
https://github.com/elixir-lang/elixir/pull/8312#discussion_r226867015

### DIFF
--- a/lib/ex_unit/test/ex_unit_test.exs
+++ b/lib/ex_unit/test/ex_unit_test.exs
@@ -100,7 +100,7 @@ defmodule ExUnitTest do
       end
     end
 
-    on_exit_reload_config()
+    configure_and_reload_on_exit([])
     ExUnit.start(slowest: 2)
     ExUnit.Server.modules_loaded()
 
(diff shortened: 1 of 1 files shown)


lib/ex_unit/test/ex_unit_test.exs
```diff
@@ -760,15 +762,27 @@ defmodule ExUnitTest do
 
   ##  Helpers
 
-  defp on_exit_reload_config(extra \\ []) do
+  defp on_exit_reload_config() do
```
josevalim: Why are we breaking this in two functions?

eksperimental: the original function was passing extra options, but those were set on_exit. But that extra argument was never used (I don't know if it became unused with the latest changes).

The new function uses for the current module and restores the original configuration on exit.
We could stick with the new one, and replace `on_exit_reload_config/2` calls

josevalim: Good catch. Yes, we should have just one function then. :)

eksperimental: done

PR #8322, opened 2018-10-23
Capitalize Mix

fertapric:


lib/mix/lib/mix/rebar.ex
```diff
@@ -119,7 +119,7 @@ defmodule Mix.Rebar do
     [fun.(config) | subs]
   end
 
-  # Translate a rebar dependency declaration to a mix declaration
+  # Translate a rebar dependency declaration to a Mix declaration
```
eksperimental: "rebar" should be capitalized here as well

josevalim:

PR #8323, opened 2018-10-23
Warn on conflicting .formatter.exs files

fertapric:


lib/mix/lib/mix/tasks/format.ex
```diff
@@ -360,10 +360,18 @@ defmodule Mix.Tasks.Format do
...
       for input <- List.wrap(formatter_opts[:inputs]),
           file <- Path.wildcard(Path.join(prefix ++ [input]), match_dot: true),
-          do: {file, formatter_opts},
+          do: {file, {dot_formatter, formatter_opts}},
           into: %{}
 
-    Enum.reduce(subs, Map.merge(acc, map), fn {sub, formatter_opts_and_subs}, acc ->
+    acc =
+      Map.merge(acc, map, fn file, {dot_formatter1, _}, {dot_formatter2, _} ->
+        Mix.raise(
```
josevalim: Let's go with Mix.shell.error, since raising is likely too much!

lib/mix/lib/mix/tasks/format.ex
```diff
@@ -360,10 +360,20 @@ defmodule Mix.Tasks.Format do
...
-          do: {file, formatter_opts},
+          do: {file, {dot_formatter, formatter_opts}},
           into: %{}
 
-    Enum.reduce(subs, Map.merge(acc, map), fn {sub, formatter_opts_and_subs}, acc ->
+    acc =
+      Map.merge(acc, map, fn file, {dot_formatter1, _}, {dot_formatter2, formatter_opts} ->
+        Mix.shell().error(
+          "Both #{dot_formatter1} and #{dot_formatter2} specify the file #{file} " <>
+            "in their :inputs option"
```
whatyouhide: Should we clearly say in the error message that options from `dot_formatter2` are used when there's a conflict like this?

fertapric: Extra info won’t hurt 👍

fertapric: On other hand, that message might get outdated if there are other conflicting `.formatter.exs` files. It would be quite an edge case, but possible.

Here is an example:

```
.formatter.exs # [inputs: "lib/**/*.ex", subdirectories: ["lib"]]
lib/.formatter.exs # [inputs: ["*.ex", "foo/**/*.ex"], subdirectories: ["foo"]]
lib/foo/.formatter.exs # [inputs: "*.ex"]
lib/foo/a.ex
```

whatyouhide: The last one of all of those will win for a given file, no?

lib/mix/lib/mix/tasks/format.ex
```diff
@@ -360,10 +360,21 @@ defmodule Mix.Tasks.Format do
...
+          do: {file, {dot_formatter, formatter_opts}},
           into: %{}
 
-    Enum.reduce(subs, Map.merge(acc, map), fn {sub, formatter_opts_and_subs}, acc ->
+    acc =
+      Map.merge(acc, map, fn file, {dot_formatter1, _}, {dot_formatter2, formatter_opts} ->
+        Mix.shell().error(
+          "Both #{dot_formatter1} and #{dot_formatter2} specify the file #{file} " <>
+            "in their :inputs option. To resolve the conflict, the configuration in " <>
+            "#{dot_formatter1} will be ignored"
```
josevalim: If we are actually resolving the conflicts, then I would rather document how we solve it and get rid of the warning altogether. For instance, I would say the one closest to the subdirectory always wins. This means the inputs files are simpler. Thoughts?

fertapric: There are some edge cases in which telling which `.formatter.exs` is closer to a file cannot be accurate. They are quite edgy though.

For example:

```
.formatter.exs # [inputs: "lib/**/*.ex", subdirectories: ["lib/bar", "lib/foo"]]
lib/bar/.formatter.exs # [inputs: ["../*.ex", "/*.ex"]]
lib/foo/.formatter.exs # [inputs: "../*.ex"]
lib/a.ex
```

Btw, the example above was not printing errors. I just pushed some changes to deal with `..` and `.` in the paths.

lib/mix/lib/mix/tasks/format.ex
```diff
@@ -360,15 +360,33 @@ defmodule Mix.Tasks.Format do
...
+          do: {expand_relative_to_cwd(file), {dot_formatter, formatter_opts}},
           into: %{}
 
-    Enum.reduce(subs, Map.merge(acc, map), fn {sub, formatter_opts_and_subs}, acc ->
+    acc =
+      Map.merge(acc, map, fn file, {dot_formatter1, _}, {dot_formatter2, formatter_opts} ->
+        Mix.shell().error(
+          "Both #{dot_formatter1} and #{dot_formatter2} specify the file " <>
+            "#{Path.relative_to_cwd(file)} in their :inputs option. To resolve the " <>
+            "conflict, the configuration in #{dot_formatter1} will be ignored"
```
josevalim: We should tell people how to address this warning too: "to avoid this warning, please change the `:inputs` entry in one of the formatter files so only one of them match this file".

lib/mix/lib/mix/tasks/format.ex
josevalim: I see, the wording would have to be the innermost subdirectory that matches it. Although I guess we can start with a warning for now and we can always amend it based on feedback!

josevalim:

vnegrisolo:
hey @fertapric , I just noticed this change today after upgrading my local elixir to 1.8.0-rc.1 and I think that we can avoid some of these warning messages if we consider the `subdirectories` config for this mix task.

Anyway, I opened a short PR to solve that if you are interested: https://github.com/elixir-lang/elixir/pull/8596

PR #8325, opened 2018-10-24
Optimize guard checks in OptionParser.to_argv/2

eksperimental:
just a small optimization


josevalim:
Doesn’t it mean we are not checking the key is an atom on line 496 and
others anymore?

I don’t think the is_atom check will have any perf difference so I would
rather focus on what is clearer.
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

eksperimental:
Sorry, I think I mistook key for value, and since we were passing `true` and `false` we didn't have the need to check for those cases.
It was an old fix that I had in my contribution list sitting for while.
sorry about the trouble

josevalim:
No problem at all. Thanks <3
-- 

PR #8329, opened 2018-10-24
Improve String.split/3 doc

ggcampinho:
- Use uniformly parts instead of substrings/parts
- Improve explanation about graphemes boundaries


josevalim:

PR #8330, opened 2018-10-24
Document --only and --include --exclude difference

VitorTrin:
#7295 made tests raise an error if the option --only was used and it matched no tests, but it isn't mentioned in the documentation and it says `--only tag` and `--include tag --exclude test` are the same when now they are not, as `--include tag --exclude test` doesn't raise on no matches.

### DIFF
--- a/lib/mix/lib/mix/tasks/test.ex
+++ b/lib/mix/lib/mix/tasks/test.ex
@@ -210,7 +210,8 @@ defmodule Mix.Tasks.Test do
 
       mix test --only external
 
-  Which is equivalent to:
+  If no tests match and the `--only` option was used the test will raise an error.
+  The following produces a similar result, except it doesn't raise an error on no matches:
 
       mix test --include external --exclude test
 
--- a/lib/mix/lib/mix/tasks/test.ex
+++ b/lib/mix/lib/mix/tasks/test.ex
@@ -214,7 +214,7 @@ defmodule Mix.Tasks.Test do
 
       mix test --include external --exclude test
 
-  It differs in that the test suite will fail if no tests are executed when the --only option is used.
+  It differs in that the test suite will fail if no tests are executed when the `--only` option is used.
 
   In case a single file is being tested, it is possible to pass a specific
   line number:
--- a/lib/mix/lib/mix/tasks/test.ex
+++ b/lib/mix/lib/mix/tasks/test.ex
@@ -210,10 +210,11 @@ defmodule Mix.Tasks.Test do
 
       mix test --only external
 
-  If no tests match and the `--only` option was used the test suite will fail.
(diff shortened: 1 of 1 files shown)


lib/mix/lib/mix/tasks/test.ex
```diff
@@ -210,7 +210,8 @@ defmodule Mix.Tasks.Test do
 
       mix test --only external
 
-  Which is equivalent to:
+  If no tests match and the `--only` option was used the test will raise an error.
```
fertapric: I think "raise" is only accurate if the `--raise` option is provided.

lib/mix/lib/mix/tasks/test.ex
```diff
@@ -210,7 +210,8 @@ defmodule Mix.Tasks.Test do
 
       mix test --only external
 
-  Which is equivalent to:
```
fertapric: What do you think about?

> For this reason, Mix also provides an `--only` option that excludes all tests and includes only the given ones:
>
>     mix test --only external
>
> Which is similar to:
>
>     mix test --include external --exclude test
>
> It differs in that the test suite will fail if no tests are executed when the `--only` option is used.

lib/mix/lib/mix/tasks/test.ex
```diff
@@ -210,7 +210,8 @@ defmodule Mix.Tasks.Test do
...
+  If no tests match and the `--only` option was used the test suite will fail.
```
eksperimental: How about rephrasing it a bit like this:
If no tests are matched and the `--only` option has been provided, the test suite will fail.

VitorTrin: Both suggestions are way better than what I wrote. I'm going to use @fertapric's

eksperimental:
I'm submitting a PR including tests for this feature and a bug that I just found. Just to let you know guys

lib/mix/lib/mix/tasks/test.ex
VitorTrin: This is much cleaner. The only small downside is that the complete explanation the behaviour of `--only`is only over after the example.

lib/mix/lib/mix/tasks/test.ex
```diff
@@ -210,10 +210,12 @@ defmodule Mix.Tasks.Test do
 
       mix test --only external
 
-  Which is equivalent to:
+  Which is similar to:
 
       mix test --include external --exclude test
 
+  It differs in that the test suite will fail if no tests are executed when the --only option is used.
```
fertapric: Nitpick: backticks in `--only` :)
```suggestion
  It differs in that the test suite will fail if no tests are executed when 
  the `--only` option is used.
```

josevalim:

PR #8332, opened 2018-10-24
Info instead of error on unmatched umbrella tests

TheFirstAvenger:
When running `mix test file_in_one_child_app_but_not_another.exs` an error is logged and an exit/raise happens for every other child app stating `Paths given to 'mix test' did not match any directory/file...`. Since this is a frequent valid use case, this Pull Request changes the message to an `info()` (similar to the `There are no tests to run` message) and doesn't raise or exit when no matching files are found while recursing.

### DIFF
--- a/lib/mix/lib/mix/tasks/test.ex
+++ b/lib/mix/lib/mix/tasks/test.ex
@@ -421,11 +421,16 @@ defmodule Mix.Tasks.Test do
   end
 
   defp raise_or_error_at_exit(message, opts) do
-    if opts[:raise] do
-      Mix.raise(message)
-    else
-      Mix.shell().error(message)
-      System.at_exit(fn _ -> exit({:shutdown, 1}) end)
+    cond do
+      Mix.Project.recursing?() ->
+        Mix.shell().info(message)
+
+      opts[:raise] ->
+        Mix.raise(message)
+
+      true ->
+        Mix.shell().error(message)
+        System.at_exit(fn _ -> exit({:shutdown, 1}) end)
     end
   end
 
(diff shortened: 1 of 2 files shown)


lib/mix/lib/mix/tasks/test.ex
```diff
@@ -412,8 +412,15 @@ defmodule Mix.Tasks.Test do
             Mix.shell().info("There are no tests to run")
 
           true ->
-            message = "Paths given to \"mix test\" did not match any directory/file: "
-            raise_or_error_at_exit(message <> Enum.join(files, ", "), opts)
+            message =
+              "Paths given to \"mix test\" did not match any directory/file: " <>
+                Enum.join(files, ", ")
+
+            if Mix.ProjectStack.recursing() do
```
josevalim: This is private API and I don't think we should be accessing it here - our tasks should not "cheat" just because they are also part of Mix. We would first need to expose it at `Mix.Project`. :)

lib/mix/lib/mix/tasks/test.ex
```diff
@@ -412,8 +412,15 @@ defmodule Mix.Tasks.Test do
...
+              Mix.shell().info(message)
+            else
+              raise_or_error_at_exit(message, opts)
```
josevalim: Btw, I think we should add this logic to inside the `raise_or_error_at_exit` function and maybe rename it to `raise_or_error_at_exit_unless_recursing`. The reason is that the same rationale you have for files would apply to `mix test --only foo` and `foo` is only available in one of the umbrella children.

TheFirstAvenger:
@josevalim Updated per the review comments

josevalim:

PR #8335, opened 2018-10-24
Add tests for "mix test {--only, --include, --exclude}" + Fix bug

eksperimental:
Add tests for "mix test {--only, --include, --exclude}"
Fix bug when test module contains skipped tests when running mix test --only
    
There was a bug that when `mix test --only` didn't return any match,
it would not fail, and it would work exactly as with "--include, --exclude options"

Reviewing #8330 made me realize this feature was not tested, and there was a bug laying behind a corner case.

It's separated in two commits. Please keep them as such to maintain the history

### DIFF
--- a/lib/mix/lib/mix/tasks/test.ex
+++ b/lib/mix/lib/mix/tasks/test.ex
@@ -383,7 +383,7 @@ defmodule Mix.Tasks.Test do
     display_warn_test_pattern(test_files, test_pattern, matched_test_files, warn_test_pattern)
 
     case CT.require_and_run(matched_test_files, test_paths, opts) do
-      {:ok, %{excluded: excluded, failures: failures, total: total}} ->
+      {:ok, %{excluded: excluded, failures: failures, skipped: skipped, total: total}} ->
         cover && cover.()
 
         option_only_present? = Keyword.has_key?(opts, :only)
@@ -395,7 +395,7 @@ defmodule Mix.Tasks.Test do
           failures > 0 ->
             System.at_exit(fn _ -> exit({:shutdown, 1}) end)
 
-          excluded == total and option_only_present? ->
+          total == excluded + skipped and option_only_present? ->
             message = "The --only option was given to \"mix test\" but no test was executed"
             raise_or_error_at_exit(message, opts)
 
--- a/lib/mix/test/mix/tasks/test_test.exs
+++ b/lib/mix/test/mix/tasks/test_test.exs
@@ -297,6 +297,12 @@ defmodule Mix.Tasks.TestTest do
       # --only doesn't match any test; it fails
       output = mix(["test", "--only", "one_hundred", "test/test_passing_test_passed_.exs"])
       assert output =~ only_error_message
+
+      # --only doesn't match any test, but the test_module contains skipped and excluded tests;
+      # it fails
+      output = mix(["test", "--only", "one_hundred"])
+      assert output =~ "5 tests, 0 failures, 4 excluded, 1 skipped"
+      assert output =~ only_error_message
     end)
   end
 
--- a/lib/mix/test/mix/tasks/test_test.exs
+++ b/lib/mix/test/mix/tasks/test_test.exs
@@ -257,6 +257,49 @@ defmodule Mix.Tasks.TestTest do
     end)
   end
 
+  test "--include, --exclude, and --only" do
+    in_fixture("test_passed", fn ->
+      # --include has no effect without --exclude
+      output_no_options = mix(["test"])
+      output_include = mix(["test", "--include", "one"])
+      assert output_include =~ "5 tests, 0 failures, 1 skipped"
+      assert output_no_options =~ "5 tests, 0 failures, 1 skipped"
+
+      # --include + --exclude test
+      output = mix(["test", "--include", "two", "--exclude", "test"])
+      assert output =~ "5 tests, 0 failures, 3 excluded, 1 skipped"
+
+      # --exclude
+      output = mix(["test", "--exclude", "exclude"])
+      assert output =~ "5 tests, 0 failures, 1 excluded, 1 skipped"
+
+      # --only
+      output = mix(["test", "--only", "one"])
+      assert output =~ "5 tests, 0 failures, 4 excluded, 1 skipped"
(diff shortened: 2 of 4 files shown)


eksperimental:
Please review and let me know if I'm missing a case to test

lib/mix/lib/mix/tasks/test.ex
```diff
@@ -395,7 +395,7 @@ defmodule Mix.Tasks.Test do
           failures > 0 ->
             System.at_exit(fn _ -> exit({:shutdown, 1}) end)
 
-          excluded == total and option_only_present? ->
+          total == excluded + skipped and option_only_present? ->
```
josevalim: I would say the current behaviour is fine. The reason why we raise here is in case you give a bad name to --only. However, if there are skipped tests, it means the name was found! Maybe we should change the error message below to say "but no tests were found"?

lib/mix/test/mix/tasks/test_test.exs
```diff
@@ -257,6 +257,55 @@ defmodule Mix.Tasks.TestTest do
     end)
   end
 
+  test "--include, --exclude, and --only" do
```
josevalim: Those tests are very very very expensive to run, that's why we don't run all possible combination. Instead we unit test the `ex_unit_opts` function and make sure we are passing the proper options to ExUnit. :)

lib/mix/lib/mix/tasks/test.ex
eksperimental: Two same test modules with the same options being called by Mix, one has a skipped tag, and one doesn't. and one fails and the other doesn't. Shouldn't it be a bug?

lib/mix/test/mix/tasks/test_test.exs
eksperimental: jeeez.. ididn't know they were THAT expensive. I will make sure how to avoid using them

lib/mix/lib/mix/tasks/test.ex
josevalim: I am not sure I understand. Can you please expand?

lib/mix/test/mix/tasks/test_test.exs
josevalim: The test suite in Mix is getting close to 2 minutes. It is precisely because some tests require setting up a new project. Those in particular set up a new project and a whole new VM too. It is death by a 10000 cuts. :)

lib/mix/lib/mix/tasks/test.ex
eksperimental: well, They are not "the same" modules, because one has a test tagged with `:skip`.

It set this repo up https://github.com/eksperimental/test_only_bug

tests involved:
- https://github.com/eksperimental/test_only_bug/blob/master/test/test_only_bug_test.exs
- https://github.com/eksperimental/test_only_bug/blob/master/test/test_only_bug_skipped_test.exs

**It does fail**
```console
$ mix test --only ten test/test_only_bug_test.exs 
Including tags: [:ten]
Excluding tags: [:test]

Finished in 0.05 seconds
1 doctest, 2 tests, 0 failures, 3 excluded

Randomized with seed 663958
The --only option was given to "mix test" but no test executed
```

**It does not fail**
```console
$ mix test --only ten test/test_only_bug_skipped_test.exs 
Including tags: [:ten]
Excluding tags: [:test]

*

Finished in 0.07 seconds
1 doctest, 2 tests, 0 failures, 2 excluded, 1 skipped

Randomized with seed 291589
```

josevalim:
I see. That’s a bug in ExUnit then. We should only mark a test as skipped
if it was not excluded. :) great find!
-- 


eksperimental:
> I see. That’s a bug in ExUnit then. We should only mark a test as skipped if it was not excluded. :) great find!

So what is the precedence here?

First, we exclude all tests passed in `--exclude` (all `test` if given the `--only` option),
Second, mark all skipped tests if they are not excluded marked with `@tag skip`?
Third,  add all included tests passed in `--include` or `--only`

is this correct? Am I missing anything?

josevalim:
@eksperimental first we should do all of the exclude, include, only, as today, and only then skip. In a nutshell, we should not take a test as skipped if we were not meant to run it in the first place.

eksperimental:
@josevalim I will look into it and submit a PR. Thank you for advice

Let's keep this one open until then

eksperimental:
@josevalim I already have a fix for this.
One last question (I hope), is, given the following test module

```elixir
      @tag :one
      test "one", do assert(true)

      @tag :skip
      @tag :two
      test "two", do assert(true)
```
And I call `mix test --include two` and `mix test --only two`, should the tests skipped or not?

josevalim:
Yes, if the test was not excluded and it was tagged as skipped, then it should be marked as skipped.

eksperimental:
great. PR is on the way

eksperimental:
I said I hope...

one last question. How about this corner case?
`mix test --include skip` and `mix test --only skip`

josevalim:
In this case skip is not excluded, so it should be taken into account. I think it should be a matter of changing the order we process excluded and skip in the runner.

eksperimental:
> In this case skip is not excluded
Do you mean skipped?

josevalim:
To be clear, I think in this case they should be marked as skipped.

josevalim:
Wait, nevermind. According to the current code we should run those tests then. So `mix test --include skip` and `mix test --only skip` should run the tests tagged as skipped.

eksperimental:
> To be clear, I think in this case they should be marked as skipped.

The current behaviour is to run it.
There's a test for this;.

https://github.com/elixir-lang/elixir/blob/master/lib/ex_unit/test/ex_unit/filters_test.exs#L29-L30

but if we allow it, it is kind of like an exception in order to run only skipped tests. I guess kind of useful because if they were maked skipped it was for a reason --they were not passing-- but developer may want to fix them by running `mix test --only skip` and remove the `:skip` tag when done

josevalim:
Yes, I agree, let's keep the current behaviour in this regard.

josevalim:
Closing this in favor of the new PR.

Issue #8283, opened 2018-10-16
`into:` argument to `for` comprehension breaks AST in quote→unquote


am-kantox:
### Environment

```
$ elixir --version
Erlang/OTP 21 [erts-10.0] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [hipe]
Elixir 1.7.3 (compiled with Erlang/OTP 21)
$ uname -a
Linux aleksei 4.15.0-38-generic #41-Ubuntu SMP Wed Oct 10 10:59:38 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
```

### Current behavior

Consider the following `__using__/1` implementation:

```elixir
defmodule Foo do
  defmacro __using__(args) do
    # AST-based map declaration using Keyword
    kw = for arg <- args, do: {arg, Macro.var(arg, __MODULE__)}
    map_kw = {:%{}, [], kw} # ⇐ hack

    # An attempt to build the map directly using `into:` in call to `for`
    map = for arg <- args, do: {arg, Macro.var(arg, __MODULE__)}, into: %{}

    result =
      quote do
        def foo(unquote(args)) do
          [unquote(map_kw), unquote(map)]
        end
      end

    # printing it out since the attempt to return an invalid AST raises
    result
    |> Macro.to_string()
    |> IO.puts()

    result
  end
end
defmodule Test, do: use Foo, [:aa]
```

Basically, there are two identical `for`s building lists out of arguments. The former one builds _the list_, the latter one tries to build _the map_, using `into:` option in the call to `for`.

_Currently:_ the former builds the correct map. while the latter raises:

```
** (CompileError) iex: invalid quoted expression: %{aa: {:aa, [], Foo}}
```

The standard output from `IO.puts`:

```
def(foo([:aa])) do
  [%{aa: aa}, %{aa: {:aa, [], Foo}}]
end
```

### Expected behavior

The results should be the same.

josevalim:
Hi @am-kantox!

That's because a map is not a quoted expression, it is a value. Every time you have a value, you need to make sure you `Macro.escape/1` it so it becomes a AST. We will make sure to improve the error message, because we could absolutely have this information in the error message.

josevalim:
To clarify, you need to do:

```elixir
      quote do
        def foo(unquote(args)) do
          [unquote(map_kw), unquote(Macro.escape(map))]
        end
      end

josevalim:
[...]

PR #8336, opened 2018-10-24
Improve error message for invalid quoted expression

mhanberg:
Addresses #8283.

Let me know what you think!

Thanks

### DIFF
--- a/lib/elixir/src/elixir_expand.erl
+++ b/lib/elixir/src/elixir_expand.erl
@@ -1117,8 +1117,10 @@ format_error({invalid_call, Call}) ->
   io_lib:format("invalid call ~ts", ['Elixir.Macro':to_string(Call)]);
 format_error({invalid_quoted_expr, Expr}) ->
   Message =
-    "invalid quoted expression: ~ts\n"
-    "You may have meant to escape this with Macro.escape/1",
+    "invalid quoted expression: ~ts\n\n"
+    "Please make sure your quoted expressions are made of valid AST nodes."
+    "If you would like to introduce a value into the AST, such as a four-element"
+    "tuple or a map, make sure to call `Macro.escape/1` before",
   io_lib:format(Message, ['Elixir.Kernel':inspect(Expr, [])]);
 format_error({invalid_local_invocation, Context, {Name, _, Args} = Call}) ->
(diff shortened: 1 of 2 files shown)


lib/elixir/test/elixir/kernel/errors_test.exs
```diff
@@ -706,7 +706,7 @@ defmodule Kernel.ErrorsTest do
 
   test "invalid macro" do
     assert_eval_raise CompileError,
-                      "nofile: invalid quoted expression: {:foo, :bar, :baz, :bat}",
+                      "nofile: invalid quoted expression: {:foo, :bar, :baz, :bat}\nYou may have meant to escape this with Macro.escape/1",
```
josevalim: I would propose for us to use a regex so we don't have to change this every time. Something like this:

```suggestion
                      ~r"nofile: invalid quoted expression: {:foo, :bar, :baz, :bat}",
```

lib/elixir/src/elixir_expand.erl
```diff
@@ -1116,7 +1116,10 @@ format_error({invalid_function_call, Expr}) ->
 format_error({invalid_call, Call}) ->
   io_lib:format("invalid call ~ts", ['Elixir.Macro':to_string(Call)]);
 format_error({invalid_quoted_expr, Expr}) ->
-  io_lib:format("invalid quoted expression: ~ts", ['Elixir.Kernel':inspect(Expr, [])]);
+  Message =
+    "invalid quoted expression: ~ts\n"
+    "You may have meant to escape this with Macro.escape/1",
```
josevalim: I think we can be a bit more descriptive. Something like this:

```
"invalid quoted expression: %{foo: :bar}\n\n"
"Please make sure your quoted expressions are made of valid AST nodes."
"If you would like to introduce a value into the AST, such as a four-element"
tuple or a map, make sure to call `Macro.escape/1` before"
```

Thoughts?

mhanberg: Sounds good to me! I'll update the branch.

mhanberg:
@josevalim I have addressed your comments. Thanks for the feedback 👍

josevalim:

PR #8340, opened 2018-10-25
Change the precedence in ExUnit.Filters.eval/4

eksperimental:
Fixes issues mentioned in #8335, particularly in this discussion:
https://github.com/elixir-lang/elixir/pull/8335#pullrequestreview-168148010

Tests should be first excluded, then included, and then skipped (if any left).
The only exception to the rule is when we explicitely include tests with :skip filter,
to which the :ok value is returned.

Fixes return value of `@spec`

### DIFF
--- a/lib/ex_unit/lib/ex_unit/filters.ex
+++ b/lib/ex_unit/lib/ex_unit/filters.ex
@@ -109,12 +109,23 @@ defmodule ExUnit.Filters do
   Evaluates the `include` and `exclude` filters against the given `tags` to
   determine if tests should be skipped or excluded.
 
-  Some filters, like `:line`, may require the whole test collection to
+  Some filters, like `:line`, may require the whole test `collection` to
   find the closest line, that's why it must also be passed as an argument.
 
   Filters can either be a regular expression or any data structure
   that implements the `String.Chars` protocol, which is invoked before comparing
-  the filter with the tag value.
+  the filter with the `:tag` value.
+
+  ## Precedence
+
+  Tests are first excluded, then included, and then skipped (if any left).
+
+  If a `:skip` tag is found in `tags`, `{:skipped, message}` is returned if the test
+  has been left after the `exclude` and `include` filters. Otherwise `{:exclude, message}`
+  is returned.
+
+  The only exception to this rule is that `:skip` is found in the `include` filter,
+  `:ok` is returned regardless of wether the test was excluded or not.
 
   ## Examples
 
@@ -125,25 +136,34 @@ defmodule ExUnit.Filters do
       {:excluded, "due to foo filter"}
 
   """
-  @spec eval(t, t, map, [ExUnit.Test.t()]) :: :ok | {:error, binary}
+  @spec eval(t, t, map, [ExUnit.Test.t()]) ::
+          :ok | {:excluded, String.t()} | {:skipped, String.t()}
   def eval(include, exclude, tags, collection) when is_map(tags) do
-    skip? = not Enum.any?(include, &has_tag(&1, %{skip: true}, collection))
+    excluded = Enum.find_value(exclude, &has_tag(&1, tags, collection))
+    excluded? = not (!excluded)
+    included? = Enum.any?(include, &has_tag(&1, tags, collection))
 
-    case Map.fetch(tags, :skip) do
-      {:ok, msg} when is_binary(msg) and skip? ->
-        {:skipped, msg}
+    skip_tag =
+      case Map.fetch(tags, :skip) do
+        {:ok, value} -> %{skip: value}
+        :error -> %{skip: true}
+      end
 
-      {:ok, true} when skip? ->
-        {:skipped, "due to skip tag"}
+    skip_included_explicitely? = Enum.any?(include, &has_tag(&1, skip_tag, collection))
 
-      _ ->
-        excluded = Enum.find_value(exclude, &has_tag(&1, tags, collection))
+    if included? or not excluded? do
+      case Map.fetch(tags, :skip) do
+        {:ok, msg} when is_binary(msg) and not skip_included_explicitely? ->
+          {:skipped, msg}
 
-        if !excluded or Enum.any?(include, &has_tag(&1, tags, collection)) do
+        {:ok, true} when not skip_included_explicitely? ->
+          {:skipped, "due to skip tag"}
+
+        _ ->
           :ok
-        else
-          {:excluded, "due to #{excluded} filter"}
-        end
+      end
+    else
+      {:excluded, "due to #{excluded} filter"}
     end
   end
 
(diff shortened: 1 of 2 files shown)


eksperimental:
I'm aware that improvements can be done to the docs explaining the PRECEDENCE, so contributions are welcome.

lib/ex_unit/lib/ex_unit/filters.ex
```diff
@@ -125,25 +136,34 @@ defmodule ExUnit.Filters do
       {:excluded, "due to foo filter"}
 
   """
-  @spec eval(t, t, map, [ExUnit.Test.t()]) :: :ok | {:error, binary}
+  @spec eval(t, t, map, [ExUnit.Test.t()]) ::
+          :ok | {:excluded, String.t()} | {:skipped, String.t()}
   def eval(include, exclude, tags, collection) when is_map(tags) do
-    skip? = not Enum.any?(include, &has_tag(&1, %{skip: true}, collection))
+    excluded = Enum.find_value(exclude, &has_tag(&1, tags, collection))
+    excluded? = not (!excluded)
```
josevalim:
```suggestion
    excluded? = excluded != nil
```

lib/ex_unit/lib/ex_unit/filters.ex
```diff
@@ -125,25 +136,34 @@ defmodule ExUnit.Filters do
...
+    included? = Enum.any?(include, &has_tag(&1, tags, collection))
+
+    skip_tag =
+      case Map.fetch(tags, :skip) do
```
josevalim: This assignment can be written as `skip_tag = %{skip: Map.get(tags, :skip, true)}`.

josevalim: Since `skip_tag` and `skip_included_explicitely?` are used only on the `if` branch, it should be moved inside the `if` clause.

lib/ex_unit/test/ex_unit/filters_test.exs
```diff
@@ -5,6 +5,106 @@ defmodule ExUnit.FiltersTest do
 
   doctest ExUnit.Filters
 
+  setup do
```
josevalim: Please move this setup inside `describe`.

lib/ex_unit/test/ex_unit/filters_test.exs
```diff
@@ -5,6 +5,106 @@ defmodule ExUnit.FiltersTest do
...
+  The following setting is given:
+
+      @tag :one
+      test "one", do assert(true)
+
+      @tag :skip
+      @tag :two
+      test "two", do assert(true)
+
+  """
```
josevalim: I think we don't need this comment, because it is duplicating what is written in the docs for the function.

lib/ex_unit/lib/ex_unit/filters.ex
```diff
@@ -109,12 +109,23 @@ defmodule ExUnit.Filters do
...
+  ## Precedence
+
+  Tests are first excluded, then included, and then skipped (if any left).
+
+  If a `:skip` tag is found in `tags`, `{:skipped, message}` is returned if the test
+  has been left after the `exclude` and `include` filters. Otherwise `{:exclude, message}`
+  is returned.
+
+  The only exception to this rule is that `:skip` is found in the `include` filter,
+  `:ok` is returned regardless of wether the test was excluded or not.
```
fertapric: Typo whether

lib/ex_unit/lib/ex_unit/filters.ex
```diff
@@ -125,25 +136,34 @@ defmodule ExUnit.Filters do
...
+    skip_tag =
+      case Map.fetch(tags, :skip) do
+        {:ok, value} -> %{skip: value}
+        :error -> %{skip: true}
+      end
 
-    case Map.fetch(tags, :skip) do
-      {:ok, msg} when is_binary(msg) and skip? ->
-        {:skipped, msg}
+    skip_included_explicitely? = Enum.any?(include, &has_tag(&1, skip_tag, collection))
```
fertapric: Typo: explicitly

lib/ex_unit/lib/ex_unit/filters.ex
eksperimental: what about `false`?

lib/ex_unit/lib/ex_unit/filters.ex
eksperimental: thank you. fixed!

lib/ex_unit/test/ex_unit/filters_test.exs
eksperimental: removed

lib/ex_unit/lib/ex_unit/filters.ex
eksperimental: done. thank you

lib/ex_unit/lib/ex_unit/filters.ex
josevalim: Can it be false? It seems it either returns a tag or nil.

eksperimental: `@tag key: false`

josevalim: I haven't checked this code in a while, but [this line](https://github.com/elixir-lang/elixir/blob/ef24e78632362355bae4e53831d15057bb4885c1/lib/ex_unit/lib/ex_unit/filters.ex#L161) makes me believe `excluded` is the name of the tag (key) and not the value.

josevalim: I don't think `@tag nil` or `@tag false` wuld make sense anyway. So let's go ahead and change it.

eksperimental: It was already fixed.

josevalim:
@eksperimental please run the formatter and we should be good to go! :heart:

eksperimental:
there you go!

lib/ex_unit/test/ex_unit/filters_test.exs
```diff
@@ -5,6 +5,99 @@ defmodule ExUnit.FiltersTest do
...
+      assert ExUnit.Filters.eval([], [:test], tags.two, []) == {:excluded, "due to test filter"}
+
+      assert ExUnit.Filters.eval([], [:numeric], tags.one, []) ==
+               {:excluded, "due to numeric filter"}
+
+      assert ExUnit.Filters.eval([], [:unknown], tags.one, []) == :ok
+      assert ExUnit.Filters.eval([], [:unknown], tags.two, []) == {:skipped, "due to skip tag"}
+    end
+
+    test "explicitly exclude a test with the `:skip` tag", tags do
```
fertapric: Minor nitpick: I would remove the backticks in the test description. 

There are only 4 places in all the codebase with backticks, and I just pushed a [PR](https://github.com/elixir-lang/elixir/pull/8343) to remove them 🙂

eksperimental: It was me who made sure they were not there :-D
check also for comments, they should not be included either

eksperimental: fixed. thank you for pointing that out

josevalim:

PR #8342, opened 2018-10-25
Change order in which filter meesage are printed in  ExUnit.CLIFormatter

eksperimental:
It was printing this when running: mix test --only foo

```
Including tags: [:foo]
Excluding tags: [:test]
```

which is deceiving beucase the order end up rendering useless the include filter.

Swaps the order then.

Additionally swaps the order in ExUnit.Formatter.format_filters/2 and tests,
while not necessary, maintaining the order is a good thing when it's something related
to precedence.


josevalim:

PR #8343, opened 2018-10-25
Remove backticks from error messages and test desc

fertapric:
To keep consistency in the codebase.

### DIFF
--- a/lib/mix/lib/mix/tasks/test.ex
+++ b/lib/mix/lib/mix/tasks/test.ex
@@ -549,7 +549,7 @@ defmodule Mix.Tasks.Test do
   defp failed_opts(opts) do
     if opts[:failed] do
       if opts[:stale] do
-        Mix.raise("Combining `--failed` and `--stale` is not supported.")
+        Mix.raise("Combining --failed and --stale is not supported.")
       end
 
       {allowed_files, failed_ids} = ExUnit.Filters.failure_info(opts[:failures_manifest_file])
--- a/lib/mix/test/mix/task_test.exs
+++ b/lib/mix/test/mix/task_test.exs
@@ -197,11 +197,11 @@ defmodule Mix.TaskTest do
     assert Mix.Task.preferred_cli_env(:no_task) == nil
   end
 
-  test "preferred_cli_env/1 returns nil when task does not have `preferred_cli_env` attribute" do
+  test "preferred_cli_env/1 returns nil when task does not have :preferred_cli_env attribute" do
     assert Mix.Task.preferred_cli_env(:deps) == nil
   end
 
-  test "preferred_cli_env/1 returns specified `preferred_cli_env` attribute" do
+  test "preferred_cli_env/1 returns specified :preferred_cli_env attribute" do
     assert Mix.Task.preferred_cli_env(:test) == :test
   end
 
--- a/lib/mix/test/mix/task_test.exs
+++ b/lib/mix/test/mix/task_test.exs
@@ -197,11 +197,11 @@ defmodule Mix.TaskTest do
     assert Mix.Task.preferred_cli_env(:no_task) == nil
   end
 
-  test "preferred_cli_env/1 returns nil when task does not have :preferred_cli_env attribute" do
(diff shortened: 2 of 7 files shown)


lib/mix/lib/mix/tasks/test.ex
```diff
@@ -549,7 +549,7 @@ defmodule Mix.Tasks.Test do
   defp failed_opts(opts) do
     if opts[:failed] do
       if opts[:stale] do
-        Mix.raise("Combining `--failed` and `--stale` is not supported.")
+        Mix.raise("Combining --failed and --stale is not supported.")
```
eksperimental: IIRC we use double quotes for commands. I think the same should apply for command arguments

fertapric: I saw the double quotes for commands, but I think command arguments are printed without them. For example:

https://github.com/elixir-lang/elixir/blob/master/lib/mix/lib/mix/tasks/xref.ex#L642
https://github.com/elixir-lang/elixir/blob/master/lib/mix/lib/mix/tasks/xref.ex#L559
https://github.com/elixir-lang/elixir/blob/master/lib/mix/lib/mix/tasks/run.ex#L104-L107

lib/mix/test/mix/task_test.exs
```diff
@@ -197,11 +197,11 @@ defmodule Mix.TaskTest do
     assert Mix.Task.preferred_cli_env(:no_task) == nil
   end
 
-  test "preferred_cli_env/1 returns nil when task does not have `preferred_cli_env` attribute" do
+  test "preferred_cli_env/1 returns nil when task does not have :preferred_cli_env attribute" do
```
eksperimental: It should be `@preferred_cli_env`
https://github.com/elixir-lang/elixir/blob/6757fbd9c68d18f627f53d0c631cdacb36905506/lib/mix/lib/mix/task.ex#L29

lib/mix/test/mix/task_test.exs
```diff
@@ -197,11 +197,11 @@ defmodule Mix.TaskTest do
...
     assert Mix.Task.preferred_cli_env(:deps) == nil
   end
 
-  test "preferred_cli_env/1 returns specified `preferred_cli_env` attribute" do
+  test "preferred_cli_env/1 returns specified :preferred_cli_env attribute" do
```
eksperimental: same here: `@preferred_cli_env`

lib/mix/test/mix/task_test.exs
fertapric: Right. I had doubts about this one. Thanks!

lib/mix/test/mix/tasks/test_test.exs
```diff
@@ -178,7 +178,7 @@ defmodule Mix.Tasks.TestTest do
 
       # `--failed` and `--stale` cannot be combined
       output = mix(["test", "--failed", "--stale"])
-      assert output =~ "Combining `--failed` and `--stale` is not supported"
+      assert output =~ "Combining --failed and --stale is not supported"
```
eksperimental: same here with the double-quotes if we decide to go for that apporach

lib/mix/lib/mix/tasks/test.ex
eksperimental: the problem with not using quotes when it is not at the end of the line is that command options can take arguments, so it can be ambiguos.

See https://github.com/elixir-lang/elixir/blob/master/lib/mix/lib/mix/tasks/run.ex#L106

it should read 

    and "--require VAL" for requiring

to avoid ambiguity and for clarity

eksperimental: I guess we can be conservative and use it when needed

eksperimental: But, I would always use double-quotes when a command option with an argument are present

josevalim:
I don’t think the quotes are useful around flags such as —foo. It is clear
from the context what they mean and the quotes would add noise.
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

josevalim:

PR #8344, opened 2018-10-25
Deprecate System.cwd/0 and System.cwd!/0

fertapric:


lib/elixir/lib/path.ex
```diff
@@ -39,7 +39,7 @@ defmodule Path do
   """
   @spec absname(t) :: binary
   def absname(path) do
-    absname(path, System.cwd!())
+    absname(path, File.cwd!())
```
fertapric: `absname/1` and `expand/1,2` could potentially raise a different error than previous Elixir versions. I guess we are ok, but just to double check.

eksperimental:
> cc @eksperimental (you mentioned it [in this mail](https://groups.google.com/forum/#!searchin/elixir-lang-core/cwd%7Csort:date/elixir-lang-core/fHcVOSu5GrI/7_B6GJO0CQAJ))

wow, I sent that email almost exactly 3 years ago. What a coincidence.

Shouldn't we soft deprecate it first, annouce it in the CHANGELOG and in v1.10 hard deprecated emitting the warnings?

josevalim:
@eksperimental we can deprecate it immediately because the alternatives have always existed. Usually the soft deprecation are to give the alternative enough time to exist.

fertapric:
❤️ 💚 💙 💛 💜

😂

PR #8347, opened 2018-10-27
Revise use of "flag" terminology

eksperimental:


eksperimental:
It omits the use of flags in Supervision flags, which is inherited from Erlang.

lib/mix/lib/mix/tasks/deps.clean.ex
```diff
@@ -53,8 +53,8 @@ defmodule Mix.Tasks.Deps.Clean do
           Mix.raise(
             "\"mix deps.clean\" expects dependencies as arguments or " <>
               "a flag indicating which dependencies to clean. " <>
-              "The --all option will clean all dependencies while " <>
-              "the --unused option cleans unused dependencies"
+              "The --all flag will clean all dependencies while " <>
+              "the --unused flag cleans unused dependencies"
```
josevalim: I would stick with `option` or `switch` here. It may be worth standardizing on one of the two.

lib/mix/lib/mix/tasks/deps.unlock.ex
```diff
@@ -67,8 +67,8 @@ defmodule Mix.Tasks.Deps.Unlock do
         Mix.raise(
           "\"mix deps.unlock\" expects dependencies as arguments or " <>
             "a flag indicating which dependencies to unlock. " <>
-            "The --all option will unlock all dependencies while " <>
-            "the --unused option unlocks unused dependencies"
+            "The --all flag will unlock all dependencies while " <>
+            "the --unused flag unlocks unused dependencies"
```
josevalim: Same here.

lib/elixir/lib/option_parser.ex
```diff
@@ -479,7 +479,7 @@ defmodule OptionParser do
       ["--bool", "--no-bool"]
 
   Some switches will output different values based on the switches
```
fertapric: “based on the `:switches` option”?  In this case, `:switches` is one of the options of `to_argv`.

lib/elixir/lib/option_parser.ex
```diff
@@ -479,7 +479,7 @@ defmodule OptionParser do
       ["--bool", "--no-bool"]
```
fertapric: A few lines above, there is a backticked `options` that does not match the argument name `opts`. There is also a backticked `switches` that I guess should be `:switches` (because I think it’s talking about the option).

lib/mix/lib/mix/tasks/deps.clean.ex
fertapric: After reading the `OptionParser` module documentation, I think I would use “switch” for the command line (`--all`) and “option” for the keyword list. So, the `OptionParser`  transforms switches from the command line into options. Does that make sense?

fertapric: Said that, looking at the documentation of several Mix tasks, the term “option” dominates.

josevalim:
Ping. :)

lib/mix/lib/mix/tasks/deps.clean.ex
eksperimental: the thing is that if we do not use flag, there are other instances of options with no arguments where it is referred to as "flag", and we should change it.
What this commit does is to standardize the use of "flag" so when there is an option with no argument we refer to it as "flag" for clarity, but If we decide to call it "option" when we should replace it wherever we call it "flag"

lib/elixir/lib/option_parser.ex
eksperimental: I think it is ok as it is, since switches type is a subset of `:switches` option

eksperimental: in this case if it refering to `:count`  in particular, which is referred as a "type"

lib/elixir/lib/option_parser.ex
eksperimental: fixed. That's why I am a big fan of never naming arguments `opts` but `options`

lib/elixir/lib/option_parser.ex
```diff
@@ -467,9 +467,9 @@ defmodule OptionParser do
   (if the value is `true` or `false`, respectively),
   and all other values are converted using `Kernel.to_string/1`.
 
-  It is advised to pass to `to_argv/2` the same set of `options`
+  It is advised to pass to `to_argv/2` the same set of `:options`
```
fertapric: I think this `options` refers to the argument, so it should be `options` and not `:options`

eksperimental: oops. fixed

lib/mix/lib/mix/tasks/deps.clean.ex
josevalim: I still think having three different names (switch, option, flag) for the same thing is confusing. Of those, I think flag is the least common, so I would completely remove it from the codebase. I definitely do not want to end-up using flag in places we didn't before. :) Thoughts?

eksperimental: I have created a new commit that replaces flag. Please have a look

michalmuskala:
Thank you! :heart:

PR #8361, opened 2018-11-03
Fix ZWJ handling in Unicode grapheme clusters

yzyzsun:
### Problem

In Elixir, `String.length("👨‍👩‍👧‍👦")` returns 2.

"👨‍👩‍👧‍👦" consists of 4 emojis joined with 3 ZWJs (ZERO WIDTH JOINER, U+200D). It should be regarded as a single Unicode grapheme cluster, so its length should be 1.

### Solution

In the module `String.Unicode`, `next_extend_size` handles some combing characters in order to properly separate grapheme clusters. However, it doesn't take account of the cases where more than 2 emojis are joined with ZWJ. This commit will fix that.

### DIFF
--- a/lib/elixir/test/elixir/string_test.exs
+++ b/lib/elixir/test/elixir/string_test.exs
@@ -562,6 +562,7 @@ defmodule StringTest do
     assert String.length("סם ייםח") == 7
     assert String.length("がガちゃ") == 4
     assert String.length("Ā̀stute") == 6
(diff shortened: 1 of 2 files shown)


josevalim:
Great find! Can you please add a test case here too: https://github.com/elixir-lang/elixir/blob/master/lib/elixir/test/elixir/string_test.exs#L556 Thank you!

josevalim:

Issue #8362, opened 2018-11-03
Failing test example in ExUnit register_attribute documentation


edwinthinks:
While trying out ExUnit [register_attribute](https://hexdocs.pm/ex_unit/ExUnit.Case.html#register_attribute/3) I noticed that the example code in the documentation results in a test failure.
```
defmodule MyTest do
  use ExUnit.Case
  ExUnit.Case.register_attribute __ENV__, :foobar

  @foobar hello: "world"
  test "using custom test attribute", context do
    assert context.registered.hello == "world"
  end
end
```

It does work however, if I replace `context.registered.hello` with `context.registered.foobar[:hello]`. I am not sure if the documented example is outdated?

### Environment

* Elixir & Erlang/OTP versions (elixir --version): Elixir 1.7.4 (compiled with Erlang/OTP 21)
* Operating system: Mac OSX 10.13.6

### Current behavior

```
defmodule MyTest do
  use ExUnit.Case
  ExUnit.Case.register_attribute __ENV__, :foobar

  @foobar hello: "world"
  test "using custom test attribute", context do
    assert context.registered.hello == "world"
  end
end
```
Has this assertion failure:
```
  1) test using custom test attribute (MyTest)
     test/my_test.exs:6
     ** (KeyError) key :hello not found in: %{foobar: [hello: "world"]}
     code: assert context.registered.hello == "world"
     stacktrace:
       test/my_test.exs:7: (test)
```

### Expected behavior

I would have expected at first that the assertion would not have failed in the documented example.

Let me know if I can help with anything :) Thanks!

josevalim:
THe documented example is wrong. Great catch!
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

edwinthinks:
Opened https://github.com/elixir-lang/elixir/pull/8363 to address this issue. Hope it works!

josevalim:
Closing in favor of #8363.

PR #8363, opened 2018-11-03
Update documented example for ExUnit register_attribute

edwinthinks:
As per issue mentioned here: https://github.com/elixir-lang/elixir/issues/8362

I think this updated comment + example will make things clearer for the next viewer :). Let me know of any suggestions!

Thanks

### DIFF
--- a/lib/ex_unit/lib/ex_unit/case.ex
+++ b/lib/ex_unit/lib/ex_unit/case.ex
@@ -499,9 +499,9 @@ defmodule ExUnit.Case do
   @doc """
   Registers a new attribute to be used during `ExUnit.Case` tests.
 
-  The attribute values will be available as an accessible key/value
-  pair in `context.registered`. The key/value pairs will be cleared
-  after each `ExUnit.Case.test/3` similar to `@tag`.
+  The attribute values will be available through `context.registered`.
+  Registered values are cleared after each `ExUnit.Case.test/3` similar
+  to `@tag`.
 
   `Module.register_attribute/3` is used to register the attribute,
   this function takes the same options.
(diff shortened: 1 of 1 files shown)


michalmuskala:
I wonder if we should go with an even simpler example:

```elixir
@foobar :foo
assert context.registered.foobar == :foo
```

I think the current keyword example might be what confused the person documenting the behaviour in the first place.

josevalim:
@michalmuskala i think the original use case though wanted to store key-values, so that's why it ended-up being part of the example. So I can also see an argument for keeping the example as is. Thoughts?

edwinthinks:
If the typical or intended use case is for stored key-value but it also enables for simple usage, I suppose including two examples wouldn't hurt? Rewrite the comments to indicate that it stores a value as simple as an atom like `:foo` but mention that it can be used to store key-values like `hello: "world"`.

Hows that sound?

fertapric:
We could use the example of the original use case:

```elixir
defmodule MyTest do
  use ExUnit.Case

  ExUnit.Case.register_attribute __ENV__, :fixtures, accumulate: true

  @fixtures :user
  @fixtures {:post, insert: false}
  test "using custom attribute", context do
    assert context.registered.fixtures == [{:post, insert: false}, :user]
  end
  
  test "custom attributes are removed per test", context do
    assert context.registered.fixtures == []
  end
end
```

edwinthinks:
@fertapric I like that! It shows the capabilities of what you could store but also illustrates the behavior of clearing out after `ExUnit.case.test/3`. Updating the PR :)

lib/ex_unit/lib/ex_unit/case.ex
```diff
@@ -510,11 +510,17 @@ defmodule ExUnit.Case do
 
       defmodule MyTest do
         use ExUnit.Case
-        ExUnit.Case.register_attribute __ENV__, :foobar
 
-        @foobar hello: "world"
-        test "using custom test attribute", context do
-          assert context.registered.hello == "world"
+        ExUnit.Case.register_attribute __ENV__, :fixtures, accumulate: true
```
fertapric: @josevalim Does it make sense that this function accepts both env or module? I think it could be simplified to accept only modules, like `Module.register_attribute/3`.

@edwinthinks I would use `__MODULE__` instead of `__ENV__`.

josevalim:

PR #8372, opened 2018-11-08
Minor docs updates

wojtekmach:


lib/elixir/lib/application.ex
```diff
@@ -22,9 +22,8 @@ defmodule Application do
   lists returned by the `project/0` and `application/0` functions defined in
   `mix.exs`, and others are generated by Mix itself.
 
-  You can learn more about the generation of application resource files in the
-  documentation of `Mix.Tasks.Compile.App`, available as well by running `mix
-  help compile.app`.
+  You can learn more about the generation of application resource files by
```
fertapric: I think the previous documentation works better in IEx:

```elixir
iex> h Application
...some doc

You can learn more about the generation of application resource files in the
documentation of Mix.Tasks.Compile.App, available as well by running mix help
compile.app.

...more doc
iex> h Mix.Tasks.Compile.App

                       Mix.Tasks.Compile.App

Writes an .app file.

An .app file is a file containing Erlang terms that defines your application.
Mix automatically generates this file based on your mix.exs configuration.
```

josevalim: Agreed. Can we pls revert those changes? The others are good to go!

josevalim:

PR #8373, opened 2018-11-08
Fix typos in IEx module

gusaiani:
Hi all, 

this fixes a few assorted apparent typos in `IEx`.

Cheers!

### DIFF
--- a/lib/iex/lib/iex.ex
+++ b/lib/iex/lib/iex.ex
@@ -249,7 +249,7 @@ defmodule IEx do
   and we can access the information and modules defined over there:
 
       iex(foo@HOST)1> Hello.world
-      "it works"
+      "it works!"
 
   In fact, connecting to remote shells is so common that we provide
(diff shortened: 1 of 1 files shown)


lib/iex/lib/iex.ex
```diff
@@ -248,7 +248,7 @@ defmodule IEx do
   Now we are connected into the remote node, as the prompt shows us,
   and we can access the information and modules defined over there:
 
-      rem(foo@macbook)1> Hello.world
+      iex(foo@HOST)1> Hello.world
       "it works"
```
fertapric: This should be `"it works!"` (with exclamation mark)
```suggestion
       "it works!"
```

gusaiani: Thank you for pointing this out, @fertapric 🙇

whatyouhide:
Thanks @gusaiani! 💟

PR #8375, opened 2018-11-08
Expand operands in size*unit bitstring modifier

### LINKED ISSUE #8357: Module attribs multiplied by constant in binaries size throws a CompileError
### Environment
Erlang/OTP 21 [erts-10.1.1] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1] [hipe]
IEx 1.7.3 (compiled with Erlang/OTP 21)

* Operating system: 
Manjaro

### Description
Elixir 1.7.3 doesn't support usage of module attributes multiplied by some unit value.

### Current behavior
Defining module:
```
defmodule A do                
  @attrib 5                     
  def a, do: <<5 :: @attrib>>   
end
```
compiles succesfully, but defining module
```
defmodule B do                
  @attrib 5                     
  def b, do: <<5 :: @attrib*2>>   
end
```

throws
```
(CompileError) iex:4: unknown bitstring specifier: @attrib * 2
    (elixir) src/elixir_bitstring.erl:167: :elixir_bitstring.expand_specs/5
    (elixir) src/elixir_bitstring.erl:41: :elixir_bitstring.expand/7
    (elixir) src/elixir_bitstring.erl:20: :elixir_bitstring.expand/4
    iex:4: (module)
```

BTW, following code works like a charm:
```
defmodule C do                
  @attrib 5                     
  def c, do: <<5 :: size(@attrib) - unit(2)>>  
end
```

### Expected behavior

Second form should also compile.

wojtekmach:
Fixes #8357

### DIFF
--- a/lib/elixir/src/elixir_bitstring.erl
+++ b/lib/elixir/src/elixir_bitstring.erl
@@ -218,9 +218,9 @@ expand_each_spec(_Meta, [], Map, E) ->
 
 unpack_specs({'-', _, [H, T]}, Acc) ->
   unpack_specs(H, unpack_specs(T, Acc));
-unpack_specs({'*', _, [{'_', _, Atom}, Unit]}, Acc) when is_atom(Atom) and is_integer(Unit) ->
+unpack_specs({'*', _, [{'_', _, Atom}, Unit]}, Acc) when is_atom(Atom) ->
   [{unit, [], [Unit]} | Acc];
-unpack_specs({'*', _, [Size, Unit]}, Acc) when is_integer(Size) and is_integer(Unit) ->
+unpack_specs({'*', _, [Size, Unit]}, Acc) ->
   [{size, [], [Size]}, {unit, [], [Unit]} | Acc];
 unpack_specs(Size, Acc) when is_integer(Size) ->
   [{size, [], [Size]} | Acc];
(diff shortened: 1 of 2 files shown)


michalmuskala:
I'm not sure we should allow just about anything in there, in particular `(8 * x) * 2` is definitely not valid. I think we should check what exactly erlang allows in there and follow suit. We don't have the `erl_lint` layer at the moment to protect us, so we need to be careful, the compiler might do some weird things if it receives inputs it doesn't expect.

wojtekmach:
@michalmuskala there's existing validation code, e.g. your example fails with:

```
iex> <<0::(8 * x) * 2>>
** (CompileError) iex:2: size in bitstring expects an integer or a variable as argument, got: :erlang.*(8, x)
    (elixir) src/elixir_bitstring.erl:197: :elixir_bitstring.expand_each_spec/4
    (elixir) src/elixir_bitstring.erl:168: :elixir_bitstring.expand_specs/5
    (elixir) src/elixir_bitstring.erl:41: :elixir_bitstring.expand/7
    (elixir) src/elixir_bitstring.erl:20: :elixir_bitstring.expand/4
```

but I definitely agree we should make it stricter if we need to.

josevalim:
@michalmuskala the code change only takes care of the unpacking, we still validate and expand it later on. So I believe this change is safe. Can you think of other downsides?

michalmuskala:
You're right, I'm sorry for not checking more deeply. 

LGTM

josevalim:

Issue #8370, opened 2018-11-06
Unable to compile large project on slow machine (deadlock between files)


jeffutter:
### Environment

* macOS Mojave (installed via asdf)

```
Erlang/OTP 21 [erts-10.0.5] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [hipe]

Elixir 1.7.2 (compiled with Erlang/OTP 21)
```

* Docker (ubuntu)

```
Erlang/OTP 21 [erts-10.1] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1] [hipe]

Elixir 1.7.3 (compiled with Erlang/OTP 21)
```

### Current behavior

We are getting deadlocks when doing `mix` compile on slower machines. This issue started happening very rarely and has become more pervasive, to the point of being unable to compile on CI (CircleCI).

When compiling locally on my MacBook without any resource restrictions compiling works fine. However, other developers on slower machines have seen the issue pop up more often.

We also see this issue when compiling in Docker when not enough resources are allocated. The problem appears when a user only allocates 2 cores and 4GB of memory. Increasing this to 4 cores and 8GB makes the issue go away. I believe the build boxes on CircleCI that we have all have 2 cores available for compiling.

I have been able to reproduce this outside of docker by instructing the erlang VM to only use 2 cores with `MIX_ENV=prod elixir --erl '+S 2' -S mix compile`. The issue happens with `+S 2` `+S 3` but not `+S 4`

The compile output is fairly long. Here is an abridged version skipping to the OTP app that the errors appear in:

```elixir
==> partner
Compiling 58 files (.ex)

== Compilation error in file lib/partner/job_assignment.ex ==
** (CompileError)  deadlocked waiting on module Partner.Job
    (elixir) lib/kernel/typespec.ex:462: Kernel.Typespec.typespec/3
    (elixir) lib/kernel/typespec.ex:744: anonymous fn/4 in Kernel.Typespec.fn_args/4
    (elixir) lib/enum.ex:1925: Enum."-reduce/3-lists^foldl/2-0-"/3
    (elixir) lib/kernel/typespec.ex:744: Kernel.Typespec.fn_args/4
    (elixir) lib/kernel/typespec.ex:733: Kernel.Typespec.fn_args/5
    (elixir) lib/kernel/typespec.ex:300: Kernel.Typespec.translate_spec/7

== Compilation error in file lib/partner/estimated_rate_period_cost.ex ==
[...]

PR #8377, opened 2018-11-09
Resolve circular struct dependencies in typespecs

josevalim:
Prior to this patch, if a module depends on the struct of
another module and that other module depends on the struct
of the first module inside typespecs, it would lead to a
compiler deadlock, which this deadlock would not happen
in "regular code".

This commit also adds a test to ensure @enforce_keys are
not enforced when expanding typespecs.

For now, we are using a compiler private API, but we should
provide a public version for it. This will be discussed below.

### DIFF
--- a/lib/elixir/lib/kernel/typespec.ex
+++ b/lib/elixir/lib/kernel/typespec.ex
@@ -533,14 +533,9 @@ defmodule Kernel.Typespec do
     module = Macro.expand(name, caller)
 
     struct =
-      if module == caller.module do
-        Module.get_attribute(module, :struct) ||
-          compile_error(caller, "struct is not defined for #{Macro.to_string(name)}")
-      else
-        module.__struct__
-      end
-
-    struct = struct |> Map.from_struct() |> Map.to_list()
+      :elixir_map.load_struct(meta, module, [], caller)
+      |> Map.from_struct()
+      |> Map.to_list()
 
     unless Keyword.keyword?(fields) do
       compile_error(caller, "expected key-value pairs in struct #{Macro.to_string(name)}")
@@ -553,7 +548,10 @@ defmodule Kernel.Typespec do
 
     Enum.each(fields, fn {field, _} ->
       unless Keyword.has_key?(struct, field) do
-        compile_error(caller, "undefined field #{field} on struct #{Macro.to_string(name)}")
+        compile_error(
+          caller,
+          "undefined field #{inspect(field)} on struct #{Macro.to_string(name)}"
+        )
       end
     end)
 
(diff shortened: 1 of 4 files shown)


nietaki:
(If I understand it right, it should fix https://github.com/elixir-lang/elixir/issues/8370)

lib/elixir/lib/kernel/typespec.ex
```diff
@@ -533,14 +533,9 @@ defmodule Kernel.Typespec do
...
     struct =
-      if module == caller.module do
-        Module.get_attribute(module, :struct) ||
-          compile_error(caller, "struct is not defined for #{Macro.to_string(name)}")
-      else
-        module.__struct__
-      end
-
-    struct = struct |> Map.from_struct() |> Map.to_list()
+      :elixir_map.load_struct(meta, module, [], caller)
```
josevalim: I don't like that we are accessing a compiler function here and I would like to provide a proper API for this. The implementation would look like this:

```elixir
def ...(module, env) when is_atom(module) do
  :elixir_map.load_struct([line: env.line], module, [], env)
end
```

The question is: where to introduce it?

1. We could perform this expansion by default in `Macro.expand/2`. I am afraid this can be surprising and undesired in some cases.

2. We could add a specific function to `Macro`, such as `Macro.expand_struct!/2`, that does it

3. Since this is similar to `Code.ensure_compiled`, we could add `Code.ensure_struct/2`, but note that `ensure_struct` receives the environment but `ensure_compiled` does not

Thoughts?

wojtekmach: I like `Macro.struct!/2` because it mirrors `Kernel.struct!/1`

michalmuskala: I like `Macro.struct!/2` as well, though always "just" doing it on `Macro.expand/2` also seems compelling. I'm not sure...

whatyouhide: `Macro.struct!/2` is a bit of a cryptic name IMO, you can't really say what it's doing by its name 😕

fertapric: I would like to take a few steps back: how would users benefit from the `Macro.struct!` function (or whatever we decide)? 🤔 Maybe compile time deadlocks like the ones in the typespecs...but I guess you would also like to use functions from the module. I think you would have deadlocks either way.

On top of that, we would have two `struct!` functions, and that might be confusing.

In summary, could be an option to leave it as is and move it elsewhere if needed in the future?

fertapric: However, if we are planning to `@doc false` it, then `Macro.load_struct!` sounds good to me 🙂

josevalim: @fertapric we have ways to resolve deadlocks from functions by calling `Code.ensure_compiled!`. We don't have a way to do structs resolution though. :(

josevalim: I will go ahead with `Macro.struct!` as I don't think we will find a perfect name. I like it mirrors Kernel.

josevalim:

jeffutter:
@josevalim Thanks so much for digging into this and fixing it so quickly!

PR #8383, opened 2018-11-12
Timezone support

josevalim:
Continuation of #7914.

It removes leap seconds support, so we can have a more fine grained discussion on the topic. It also streamlines the `shift_zone` operation.

We still need to discuss the API for the timezone client.

### DIFF
--- a/lib/elixir/lib/calendar/datetime.ex
+++ b/lib/elixir/lib/calendar/datetime.ex
@@ -172,7 +172,9 @@ defmodule DateTime do
   Converts the given `NaiveDateTime` to `DateTime`.
 
   It expects a time zone to put the NaiveDateTime in.
-  Currently it only supports "Etc/UTC" as time zone.
+
+  It only supports "Etc/UTC" as time zone if a `TimeZoneDatabase`
+  is not provided as a third argument.
 
   ## Examples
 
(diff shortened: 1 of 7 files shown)


lib/elixir/lib/calendar/datetime.ex
```diff
@@ -172,20 +172,129 @@ defmodule DateTime do
...
+        {:ambiguous, first_datetime, second_datetime}
+
+      {:gap, {first_period, first_period_until_wall}, {second_period, second_period_from_wall}} ->
+        # `until_wall` is not valid, but any time just before is.
+        # So by subtracting a second and adding .999999 seconds
+        # we get the last microsecond just before.
+        before_naive =
+          first_period_until_wall
+          |> Map.put(:microsecond, {999_999, 6})
+          |> NaiveDateTime.add(-1)
```
josevalim: I wonder if we should ask the timezone database to return the last valid instant so we can skip this conversion. Thoughts @lau?

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -206,31 +315,35 @@ defmodule DateTime do
...
-  @spec from_naive!(NaiveDateTime.t(), Calendar.time_zone()) :: t
-  def from_naive!(naive_datetime, time_zone) do
-    case from_naive(naive_datetime, time_zone) do
+  @spec from_naive!(
+          NaiveDateTime.t(),
+          Calendar.time_zone(),
+          TimeZoneDatabaseClient.tz_db_or_config()
+        ) :: t
+  def from_naive!(naive_datetime, time_zone, tz_db_or_config \\ :from_config) do
+    case from_naive(naive_datetime, time_zone, tz_db_or_config) do
```
wojtekmach: we also need to handle `{:ambiguous, _, _}` and `{:gap, _, _}` returns from `from_naive/3`, I guess we may have to raise on them.

lib/elixir/lib/calendar/datetime.ex
lau: Documenting this for posterity. José and I just talked about this and agreed to keep the handling here for the following reason:

When the gap starts, the start of the gap is invalid. What we want here is the last possible valid time just before a gap. In this case it is 6 nines (.999999). If the timezone database was to decide it, it would have to know about what precision Elixir is using. Today it is 6 nines. So we would have to change the interface in order to pass a fractional seconds with a lot of nines just for this purpose. And the timezone database or interface would decide how many nines. Instead it is simpler to let Elixir itself decide this.

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -172,20 +172,129 @@ defmodule DateTime do
...
+          NaiveDateTime.t(),
+          Calendar.time_zone(),
+          TimeZoneDatabaseClient.tz_db_or_config()
+        ) ::
+          {:ok, t}
+          | {:ambiguous, t, t}
+          | {:gap, t, t}
+          | {:error, :time_zone_not_found}
+          | {:error, :incompatible_calendars}
+          | {:error, :no_time_zone_database}
```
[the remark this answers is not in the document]
josevalim: It can also be returned for `:from_config`. `:from_config` means we read the global timezone database instead of the one passed as arg.

lib/elixir/lib/calendar/datetime.ex
josevalim: @lau what do you suggest here? Should we raise on ambiguous and gap?

lau: I think raising makes sense on both of those.

Here are some other ideas that are related:

`from_naive!` could raise on `:gap` because the provided naive datetime is not valid in the timezone. For `:ambiguous` though it could just return one of the valid results. In case people just want one of the valid results even though they won't know that there were two results.

I imagine that something that people might want in some cases is "just get me a valid datetime that is close enough if in a gap or ambiguous".  For instance if you get a `:gap` it just returns the second (or first) datetime of  `{:ambiguous, first_datetime, second_datetime}`. And the same thing for `{:gap, first_datetime, second_datetime}`. 

There could be a function where you pass the result of `from_naive/3` and it just returns either the only result if `{:ok, dt}`, or the second datetime of `{:gap, _, dt2}` or `{:ambiguous, _, dt2}`. e.g. `DateTime.from_naive(ndt, tz) |> DateTime.degap_and_disambiguate`. But maybe such a function is simple enough that it should not be in the standard library and could just be an example pattern matching that is in the documentation:
```
from_naive(...) ->
    {:ok, dt} -> {:ok, dt}
    {:gap, _, dt2} -> {:ok, dt2}
    {:ambiguous, _, dt2} -> {:ok, dt2}
    {:error, _} = error -> error
end
```

I'm not sure which of those I prefer or should be in the standard library, but some ideas to consider.

josevalim: Thanks @lau! Since I am not sure any of those choices are reasonable to pick as a default, let's make sure we raise for now and we can add new functions in the future.

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -206,40 +319,167 @@ defmodule DateTime do
...
+      iex> {:ok, datetime} = DateTime.now("Europe/Copenhagen", FakeTimeZoneDatabase)
+      iex> datetime.time_zone
+      "Europe/Copenhagen"
+      iex> DateTime.now("not a real time zone name", FakeTimeZoneDatabase)
+      {:error, :time_zone_not_found}
+
+  """
+  @doc since: "1.8.0"
+  @spec now(Calendar.time_zone(), Calendar.time_zone_database()) ::
+          {:ok, t} | {:error, :time_zone_not_found}
```
fertapric: This might also return `{:error, :utc_only_time_zone_database}`

lib/elixir/lib/calendar.ex
```diff
@@ -90,6 +90,28 @@ defmodule Calendar do
           microsecond: microsecond
         }
 
+  @typedoc """
+  Speficies the time zone database for calendar operations.
```
fertapric:
```suggestion
  Specifies the time zone database for calendar operations.
```

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -206,40 +319,167 @@ defmodule DateTime do
...
+  end
+
+  @doc """
+  Changes the time zone of a `DateTime`.
+
+  Returns a `DateTime` for the same point in time, but instead at
+  the time zone provided. It assumes that `DateTime` is valid and
+  exists in the given timezone and calendar.
+
+  By default, it uses the default time_zone returned by
```
fertapric: I think this should be ``` `time_zone_database` ```.

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -206,40 +319,167 @@ defmodule DateTime do
...
+      iex> cph_datetime = DateTime.from_naive!(~N[2018-07-16 12:00:00], "Europe/Copenhagen", FakeTimeZoneDatabase)
+      iex> {:ok, pacific_datetime} = DateTime.shift_zone(cph_datetime, "America/Los_Angeles", FakeTimeZoneDatabase)
+      iex> pacific_datetime
+      #DateTime<2018-07-16 03:00:00-07:00 PDT America/Los_Angeles>
+
+  """
+  @doc since: "1.8.0"
+  @spec shift_zone(t, Calendar.time_zone(), Calendar.time_zone_database()) ::
+          {:ok, t} | {:error, :time_zone_not_found | :utc_only_time_zone_database}
+  def shift_zone(datetime, time_zone, time_zone_database)
```
fertapric: The documentation says that the `time_zone_database` defaults to one returned by `Calendar.get_time_zone_database/1`, so I think the default value is missed here.

lib/elixir/lib/calendar.ex
```diff
@@ -90,6 +90,28 @@ defmodule Calendar do
           microsecond: microsecond
         }
 
+  @typedoc """
+  Specifies the time zone database for calendar operations.
+
+  Many functions in the `DateTime` module requires a time zone database.
```
michalmuskala: Many functions require~s~

lib/elixir/lib/calendar.ex
```diff
@@ -90,6 +90,28 @@ defmodule Calendar do
...
+  By default, it uses the default time zone database returned by
+  `Calendar.get_time_zone_database/1`, which defaults to
+  `Calendar.UTCOnlyTimeZoneDatabase` which only handles "Etc/UTC"
+  datetimes and returns `{:error, :utc_only_time_zone_database}`
+  for any other time zone.
+
+  Another time zone database, such as one originating from a package,
+  can be passed as argument or set globally, either via configuration:
```
michalmuskala: I think this sentence is a bit awkward. I had to read it twice to understand it. Maybe something like this would be better?

```
Other time zone databases (including ones provided by packages) can be configured
as default either via configuration:

    config :elixir, :time_zone_database, CustomTimeZoneDatabase

or by calling `Calendar.put_time_zone_database/1`
```

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -16,17 +16,21 @@ defmodule DateTime do
...
 
-  Such may be addressed in upcoming versions, meanwhile,
-  use third-party packages to provide `DateTime` building and
-  similar functionality with time zone backing.
+  Another time zone database, such as one originating from a package,
+  can be passed as argument or set globally, either via configuration:
+
+      config :elixir, :time_zone_database, CustomTimeZoneDatabase
+
+  or by calling `Calendar.put_time_zone_database/1`.
```
michalmuskala: Do we really want to duplicate the documentation text like that? I'm pretty sure they will get out of sync pretty quickly. I would forward from one to another - I'm not sure which way, though.

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -171,21 +175,130 @@ defmodule DateTime do
   @doc """
   Converts the given `NaiveDateTime` to `DateTime`.
 
-  It expects a time zone to put the NaiveDateTime in.
-  Currently it only supports "Etc/UTC" as time zone.
+  It expects a time zone to put the `NaiveDateTime` in.
+  If the timezone is "Etc/UTC", it always succeeds. Otherwise,
+  the NaiveDateTime is checked against the time zone database
+  given as `time_zone_database`. See the "Time zone database"
```
michalmuskala: We could have a link to the type with the explanation here. Wdyt?

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -206,40 +319,167 @@ defmodule DateTime do
...
+      iex> cph_datetime = DateTime.from_naive!(~N[2018-07-16 12:00:00], "Europe/Copenhagen", FakeTimeZoneDatabase)
+      iex> {:ok, pacific_datetime} = DateTime.shift_zone(cph_datetime, "America/Los_Angeles", FakeTimeZoneDatabase)
+      iex> pacific_datetime
+      #DateTime<2018-07-16 03:00:00-07:00 PDT America/Los_Angeles>
+
+  """
+  @doc since: "1.8.0"
+  @spec shift_zone(t, Calendar.time_zone(), Calendar.time_zone_database()) ::
+          {:ok, t} | {:error, :time_zone_not_found | :utc_only_time_zone_database}
+  def shift_zone(datetime, time_zone, time_zone_database \\ Calendar.get_time_zone_database())
```
michalmuskala: Should we have an accompanying `force_zone` function that uses the same naive dt, but in a specific zone? For example, if you got time in the wrong zone, but the numeric values are correct. From experience, I know such a function is useful.

If we don't want to provide such a function, maybe in the docs for `shift_zone` we describe how to do it by first converting to native and then from naive to the target zone.

lib/elixir/lib/calendar/time_zone_database.ex
```diff
@@ -0,0 +1,89 @@
...
+
+  For instance one period could be the summer of 2018 in "Europe/London" where summer time /
+  daylight saving time is in effect and lasts from spring to autumn. At autumn the `std_offset`
+  changes along with the `zone_abbr` so a different period is needed during winter.
+  """
+  @type time_zone_period :: %{
+          optional(any) => any,
+          utc_offset: Calendar.utc_offset(),
+          std_offset: Calendar.std_offset(),
+          zone_abbr: Calendar.zone_abbr()
```
michalmuskala: I would suggest that we use a record or a raw tuple here. I know it's much less legible, but it's more performant and in things like this we probably want to squeeze every last bit of performance.

fertapric:
I would go with `Calendar.get_time_zone_database` because it looks more balanced with `Calendar.put_time_zone_database`, but either way would work for me.

michalmuskala:
@josevalim I guess it's pretty common to have a `/0` function for "get" and a `/1` function for "put" with exactly the same name. We could use this convention here. What do you think?

fertapric:
@michalmuskala do we have examples of that convention in the codebase? 

We have examples of `get_value()` and `put_value(value)` in `Application.get_env/put_env`, `module.get_attribute/put_attribute`, and `System.get_env/put_env`.

We also have examples of `value()` and `put_value(value)` in `Registry.meta/put_meta`.

lib/elixir/lib/calendar/time_zone_database.ex
josevalim: I am not worried about eligibility but rather about extensibility. Using a tuple would be very unforgiving in case we need to add new information. A record is a bit better but adding custom fields would be tricky. Btw, if we have many maps, all with the same keys, do they share the keys structure when loading it in memory?

josevalim:
> We also have examples of value() and put_value(value) in Registry.meta/put_meta.

I believe this is the convention @michalmuskala suggestion. :) I am honestly torn on which one to pick. I can say though the ones in Application/System are usually more meta (it is a general key-value) and the ones in Registry is more about setting something, which is more inline with what we have here.

lexmag:
We have `Logger.metadata/0,1` that does what @michalmuskala said.
However, I always wished it have `get` and `put` for enhanced readability. :bowtie:

josevalim:
I went with `Calendar.time_zone_database` and `Calendar.put_time_zone_database` per the rationale above.

michalmuskala:
@fertapric one place where this is used is [`URI.default_port/1,2`](https://github.com/elixir-lang/elixir/blob/master/lib/elixir/lib/uri.ex#L49-L68)

lib/elixir/lib/calendar/time_zone_database.ex
michalmuskala: It depends on how it is loaded. In general, though, if we're loading it from ETS, the table will always copy everything. ETS does not do literal sharing.

lib/elixir/lib/calendar/datetime.ex
michalmuskala: Ok, that's fair. Should we add some comment around the docs that you should change the other place as well?

lib/elixir/lib/calendar.ex
```diff
@@ -90,6 +90,28 @@ defmodule Calendar do
           microsecond: microsecond
         }
 
+  @typedoc """
+  Specifies the time zone database for calendar operations.
+
+  Many functions in the `DateTime` module require a time zone database.
+  By default, it uses the default time zone database returned by
+  `Calendar.get_time_zone_database/1`, which defaults to
```
lexmag: `/0`? (in some places below also)

lib/elixir/lib/calendar/datetime.ex
michalmuskala: Sorry for not being clear. The use case I have in mind would be exactly equivalent to something like:

```elixir
def force_zone(datetime, time_zone) do
  from_naive(to_naive(datetime), time_zone)
end
```

This means basically that if I have a 9 AM CET and I want to have 9 AM UTC or similar. It's generally useful in many cases, where you actually don't want to do proper time zone conversion.

lau: @michalmuskala If you have a DateTime and want the same "wall time" in a different time zone you can pass it to a function only expecting a naive datetime. E.g. 2018-11-13 09:00:00 Europe/Copenhagen and want 9:00:00 for the same date but in UTC instead.

DateTime.from_naive(cph_dt_9oclock, "Etc/UTC") should return the same "wall time" as the one provided, but in UTC instead. Because `from_naive` just looks at the fields required by the Calendar.naive_datetime type and ignores the time zone part.

lib/elixir/lib/calendar/time_zone_database.ex
josevalim: I was wondering about modules. But keep in mind we can always store it in ets as a tuple and then convert it to a map when reading out. :)

josevalim:
Squashed and merged manually, thanks everyone! 

lib/elixir/test/elixir/calendar_test.exs
```diff
@@ -878,4 +887,179 @@ defmodule DateTimeTest do
 
     assert DateTime.diff(dt1, dt2) == 3_281_904_000
   end
+
+  describe "from_naive" do
+    test "uses default time zone database from config" do
+      Calendar.put_time_zone_database(FakeTimeZoneDatabase)
```
lau: This test module is has `async` set to `true` for tests. I assume this and `Calendar.put_time_zone_database` can cause race conditions. Should the tests that call `Calendar.put_time_zone_database` be put in a separate test module? Or just make the existing module `async: false`.

cc @josevalim

josevalim: Pushed!

PR #8388, opened 2018-11-14
Update Kernel sigil macro argument name to match others.

schneiderderek:
Renamed the first argument to `term` to better describe the argument. Additionally, using `date` as an argument name was confusing for `sigil_T` and `sigil_N`.

### DIFF
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -5205,7 +5205,7 @@ defmodule Kernel do
       ~D[2015-01-13]
 
   """
-  defmacro sigil_D(date, modifiers)
+  defmacro sigil_D(date_string, modifiers)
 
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/kernel.ex
```diff
@@ -5205,7 +5205,7 @@ defmodule Kernel do
       ~D[2015-01-13]
 
   """
-  defmacro sigil_D(date, modifiers)
+  defmacro sigil_D(term, modifiers)
```
josevalim: I think we could go with `date_string`, `time_string`, `naive_datetime_string`, etc. Thoughts?

schneiderderek: Works for me 👍

schneiderderek:
@josevalim fixed

josevalim:

PR #8394, opened 2018-11-16
Wrap httpc read in a task to timeout

### LINKED ISSUE #8252: `mix local.hex` hangs indefinately
### Environment

```
Erlang/OTP 21 [erts-10.1] [source] [64-bit] [smp:8:4] [ds:8:4:10] [async-threads:1] [hipe]

Elixir 1.8.0-dev (c483c9c) (compiled with Erlang/OTP 21)
```

* Operating system: 

```
~/elixir> lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 16.04.5 LTS
Release:        16.04
Codename:       xenial
```

### Current behavior

Running `mix local.hex` hangs with no output. I've tracked down the hang to [this line](https://github.com/elixir-lang/elixir/blob/master/lib/mix/lib/mix/utils.ex#L579) by binary-searching using of `IO.puts` and commenting out lines.

Commenting out the line `:httpc.set_option(:ipfamily, :inet6fb4, :mix)` makes the request succeed. I'm unable to find documentation on the `:inet6fb4` option for `set_option` - actually I'm unable to find documentation for `http:set_option` at all. I see that there exists a function called `httpc:set_options` (note plural options vs option). 

### Expected behavior

Mix should not hang. Commenting out the above line fixes the hang.

NJichev:
Add a default timeout of 30 seconds and add it as an option.
Test locally using an unroutable local address 10.0.0.0(it timeouts).

Should fix #8252

Edit: I might have to pass in the `HEX_HTTP_TIMEOUT=120` env option from where this is called.

### DIFF
--- a/lib/mix/lib/mix/utils.ex
+++ b/lib/mix/lib/mix/utils.ex
@@ -487,8 +487,10 @@ defmodule Mix.Utils do
 
     * `:sha512` - checks against the given SHA-512 checksum. Returns
       `{:checksum, message}` in case it fails
+
     * `:timeout` - timeout the request after the given milliseconds.
-      Returns {:remote, :timeout} if it fails. Defaults to 30 seconds.
+      Returns {:remote, "request timed out after ...ms"} if it fails.
+      Defaults to 30 seconds
 
   """
   @spec read_path(String.t(), keyword) ::
@@ -503,12 +505,13 @@ defmodule Mix.Utils do
         task = Task.async(fn -> read_httpc(path) |> checksum(opts) end)
         timeout = Keyword.get(opts, :timeout, 30_000)
 
-        case Task.yield(task, timeout) || Task.shutdown(task, :brutal_kill) do
+        case Task.yield(task, timeout) do
           {:ok, result} ->
             result
 
           _ ->
-            {:remote, :timeout}
+            Task.shutdown(task, :brutal_kill)
+            {:remote, "request timed out after #{timeout}ms"}
         end
 
       file?(path) ->
--- a/lib/mix/lib/mix/utils.ex
+++ b/lib/mix/lib/mix/utils.ex
@@ -487,6 +487,8 @@ defmodule Mix.Utils do
 
     * `:sha512` - checks against the given SHA-512 checksum. Returns
       `{:checksum, message}` in case it fails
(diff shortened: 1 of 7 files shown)


josevalim:
> Edit: I might have to pass in the HEX_HTTP_TIMEOUT=120 env option from where this is called.

I think it is fine to not have an option, we just need to guarantee that, if the install fails, we will tell the user how to do the manual install. Usually the manual install is straight-forward. It is a matter of downloading the archive and calling `mix archive.install path/to/file`. We already show those steps to certain failures, we just need to make sure we will also show them in case of timeouts.

lib/mix/lib/mix/utils.ex
```diff
@@ -498,7 +500,13 @@ defmodule Mix.Utils do
   def read_path(path, opts \\ []) do
     cond do
       url?(path) ->
-        read_httpc(path) |> checksum(opts)
+        task = Task.async(fn -> read_httpc(path) |> checksum(opts) end)
+        timeout = Keyword.get(opts, :timeout, 60_000)
+
+        case Task.yield(task, timeout) || Task.shutdown(task, :brutal_kill) do
+          {:ok, result} -> result
+          _ -> {:remote, :timeout}
```
josevalim: The return here needs to be a string. Maybe we can say `{:remote, "request timed out after #{timeout}ms"}`?

NJichev:
The integration test looks like this so far:

```
test "archive install timeout" do
    message = ~r[request timed out after 10000ms]

    send(self(), {:mix_shell_input, :yes?, true})
    assert_raise Mix.Error, message, fn ->
      Mix.Tasks.Archive.Install.run(["https://10.0.0.0/unlikely-to-exist-0.1.0.ez"])
    end
end
```

But I get an {:error, "https .. : :socket_closed_remotely"} instead of the timeout message.

Setting the default timeout to 0 seconds makes it pass, any insights on this?

josevalim:
@NJichev oh... there is the timeout. One option is to allow the timeout to be a flag. `--timeout INTEGER`. This way it is customizable and we can use it in the test. Thoughts?

Also, I think you have to use http instead of https.

NJichev:
Yeah I was thinking of the same stuff, let me add it then.

josevalim:

PR #8396, opened 2018-11-16
Omit parens on calls with only do/end blocks (Macro.to_string)

fertapric:
Follow up of https://github.com/elixir-lang/elixir/issues/8078#issuecomment-412833448


josevalim:

PR #8402, opened 2018-11-18
Add DateTime.add function

lau:


lib/elixir/lib/system.ex
```diff
@@ -759,6 +759,11 @@ defmodule System do
   `System.convert_time_unit(1, :second, :native)`).
   """
   @spec convert_time_unit(integer, time_unit | :native, time_unit | :native) :: integer
+  # 0 will always convert to 0
+  def convert_time_unit(0, _, _) do
+    0
+  end
```
josevalim: The issue with this is that we won't check the other arguments are valid. And then if we have to check if the other arguments are valid, then I am afraid we won't see any gain in performance, so I would probably not merge this. :) Thoughts?

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -998,6 +998,68 @@ defmodule DateTime do
...
+    datetime
+    |> to_iso_days()
+    # Subtract total offset in order to get UTC and add the integer for the addition
+    |> Calendar.ISO.add_day_fraction_to_iso_days(integer_to_add - total_offset, ppd)
+    |> from_iso_days(
+      %{time_zone: "Etc/UTC", zone_abbr: "UTC", utc_offset: 0, std_offset: 0},
+      calendar,
+      precision
+    )
+    |> shift_zone(datetime.time_zone, time_zone_database)
```
michalmuskala: We should probably extract some common paths between `shift_zone` and this function since we already have iso days. The current implementation will in most cases (other than UTC) go back and forth between the struct and iso days.

lau: Yes I considered doing that. Making a private function that takes iso days (and the other required information) and does the shifting.

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -998,6 +998,68 @@ defmodule DateTime do
...
+        time_zone_database \\ Calendar.get_time_zone_database()
+      ) do
+    ppd = System.convert_time_unit(86400, :second, unit)
+    total_offset = System.convert_time_unit(utc_offset + std_offset, :second, unit)
+
+    datetime
+    |> to_iso_days()
+    # Subtract total offset in order to get UTC and add the integer for the addition
+    |> Calendar.ISO.add_day_fraction_to_iso_days(integer_to_add - total_offset, ppd)
+    |> from_iso_days(
```
josevalim: Here we are converting from iso days and then the first thing shift_zone does is to convert from datetime back to iso_days. My suggestion is to extract a `shift_iso_days` function from `shift_zone`  and make sure this function and `shift_zone` calls it.

josevalim: Ok, apparently we all said the same thing at the same time, so let's do it. :D

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -998,6 +998,68 @@ defmodule DateTime do
...
+      iex> {:ok, dt_added} = dt |> DateTime.add(3600, :second, FakeTimeZoneDatabase)
+      iex> dt_added
+      #DateTime<2018-11-15 11:00:00+01:00 CET Europe/Copenhagen>
+
+      iex> dt = DateTime.from_naive!(~N[2018-11-15 10:00:00], "Etc/UTC")
+      iex> {:ok, dt_added} = dt |> DateTime.add(3600, :second)
+      iex> dt_added
+      #DateTime<2018-11-15 11:00:00Z>
+
+  When adding 3 seconds just before "spring forward" we go from 1:59:59 to 3:00:02
```
josevalim: And what happens when add or remove a second and we go back in time? Does it return an error? Which error?

lau: The steps are all in UTC (pretending leap seconds don't exist) so there should not be any gaps or problems. Converting from UTC to wall time is fine. In real life there is always a valid wall time at any point in time.

lau: The steps are done in UTC and then converted back to the timezone of the passed DateTime.

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -998,6 +998,68 @@ defmodule DateTime do
...
+
+      iex> dt = DateTime.from_naive!(~N[2019-03-31 01:59:59.123], "Europe/Copenhagen", FakeTimeZoneDatabase)
+      iex> {:ok, dt_added} = dt |> DateTime.add(3, :second, FakeTimeZoneDatabase)
+      iex> dt_added
+      #DateTime<2019-03-31 03:00:02.123+02:00 CEST Europe/Copenhagen>
+
+  """
+  @doc since: "1.8.0"
+  @spec add(t, integer, System.time_unit(), Calendar.get_time_zone_database()) ::
+          {:ok, t()} | {:error, atom}
```
josevalim: I have noticed an API consistency issue. The other `add` functions do not return tuples, but the value, and this would be annoyingly inconsistent. We have two options:

1. Name this function something else
2. Raise in case of `{:error, _}`

It seems the only possible error is due to using the UTC-only database, so I would vote to go with 2.

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -998,6 +998,68 @@ defmodule DateTime do
...
+  When adding 3 seconds just before "spring forward" we go from 1:59:59 to 3:00:02
+
+      iex> dt = DateTime.from_naive!(~N[2019-03-31 01:59:59.123], "Europe/Copenhagen", FakeTimeZoneDatabase)
+      iex> {:ok, dt_added} = dt |> DateTime.add(3, :second, FakeTimeZoneDatabase)
+      iex> dt_added
+      #DateTime<2019-03-31 03:00:02.123+02:00 CEST Europe/Copenhagen>
+
+  """
+  @doc since: "1.8.0"
+  @spec add(t, integer, System.time_unit(), Calendar.get_time_zone_database()) ::
```
fertapric:
```suggestion
  @spec add(t, integer, System.time_unit(), Calendar.time_zone_database()) ::
```

lib/elixir/lib/system.ex
lau: I see. So it is consistent and raises on invalid units? Would it make sense to do this pattern matching in a private function in the DateTime module that does not check for the other arguments? For instance in add we can call it for the offsets, but call the System module directly for  `ppd = System.convert_time_unit(86400, :second, unit)` since that is not going to be 0 and then it will raise is the unit is not valid.

josevalim: Doing it inside calendar is fine by ultimately I don't think it will matter as this is supposed to be quite fast anyway. :) Your call.

lib/elixir/lib/calendar/datetime.ex
lau: Yes, I made it use tuples to be consistent with other functions in DateTime where the time zone database returns an error if it does not recognize the time zone or cannot handle it (in the case of the UTC only one).

Raising instead of tuples make sense to me. In some of the other function where a time zone name is provided separately it is more likely that it is an invalid time zone name. If the time zone name is from the field in the `DateTime` struct it is probably from the same time zone database the developer is using so it is less likely to be invalid.

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -427,7 +427,12 @@ defmodule DateTime do
   end
 
   def shift_zone(
-        %{std_offset: std_offset, utc_offset: utc_offset} = datetime,
+        %{
```
fertapric: Minor nitpick: move the variable assignment to the first line?

```elixir
def shift_zone(datetime, time_zone, time_zone_database) do
  %{
    std_offset: std_offset,
    utc_offset: utc_offset,
    calendar: calendar,
    microsecond: {_, precision}
  } = datetime
```

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -998,6 +1005,62 @@ defmodule DateTime do
...
+
+      iex> dt = DateTime.from_naive!(~N[2019-03-31 01:59:59.123], "Europe/Copenhagen", FakeTimeZoneDatabase)
+      iex> dt |> DateTime.add(3, :second, FakeTimeZoneDatabase)
+      #DateTime<2019-03-31 03:00:02.123+02:00 CEST Europe/Copenhagen>
+
+  """
+  @doc since: "1.8.0"
+  @spec add(t, integer, System.time_unit(), Calendar.time_zone_database()) :: t()
+  def add(
+        %DateTime{
```
fertapric: Minor nitpicks: 

1. move variable assignment to the first line
2. rename `integer_to_add` to `integer` (same name variable as `NaiveDateTime.add/3`)

lau: Good idea. I have renamed the variable in both functions to `amount_to_add` for consistency and a better description of what the variable is for.

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -998,6 +1004,73 @@ defmodule DateTime do
     naive_diff + System.convert_time_unit(offset_diff, :second, unit)
   end
 
+  @doc """
+  Adds a specified amount of time to a `DateTime`.
+
+  Accepts an `integer` in any `unit` available from `t:System.time_unit/0`.
```
fertapric: Nitpick: I think `` `integer` `` here was referring to the second argument, which is now `amount_to_add`. We could either remove the backticks or rephrase this paragraph :)

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -998,6 +1004,73 @@ defmodule DateTime do
...
+      |> Calendar.ISO.add_day_fraction_to_iso_days(amount_to_add - total_offset, ppd)
+      |> shift_zone_for_iso_days_utc(calendar, precision, datetime.time_zone, time_zone_database)
+
+    case result do
+      {:ok, result_datetime} ->
+        result_datetime
+
+      {:error, error} ->
+        raise ArgumentError,
+              "Cannot add #{amount_to_add} #{unit} to #{datetime}.\n" <>
```
fertapric: Nipitck: errors in Elixir usually start with lowercase

josevalim:

PR #8403, opened 2018-11-18
Use proper calendar types

fertapric:


josevalim:

michalmuskala:
BTW, shouldn't we have a warning for non-existent types? We probably could do something like that with xref, couldn't we?

josevalim:
@michalmuskala I think it is the job of dialyzer or similar tools to catch things like that. :)

michalmuskala:
If that's the case shouldn't we be running dialyzer on CI? Otherwise we can ship broken code.

josevalim:
@michalmuskala I am ok with doing so to at least detect undefined types. :+1:

Issue #5440, opened 2016-11-11
GenServer.cast/2 different behaviour


michalmuskala:
Currently the implementation of `GenServer.cast/2` and `:gen_server.cast/2` differ slightly in how remote, not connected nodes, are handled. The Elixir version will try connecting synchronously to the node, while the erlang version will do this asynchronously spawning a process to do so.

Elixir: https://github.com/elixir-lang/elixir/blob/master/lib/elixir/lib/gen_server.ex#L796-L803
```elixir
  defp do_send(dest, msg) do
    try do
      send(dest, msg)
      :ok
    catch
      _, _ -> :ok
    end
  end
```

Erlang: https://github.com/erlang/otp/blob/maint/lib/stdlib/src/gen_server.erl#L401-L407
```erlang
do_send(Dest, Msg) ->
    case catch erlang:send(Dest, Msg, [noconnect]) of
	noconnect ->
	    spawn(erlang, send, [Dest,Msg]);
	Other ->
	    Other
    end.
```

Should the elixir version be changed to mirror the erlang one?

fishcakez:
We purposefully used a different behaviour so that semantics were consistent for all naming schemes and so that we had ordering of messages. Unfortunately we do not have that ordering guarantee. I have changed the tag to `bug` because I think it is more important to highlight that we can't provide the guarantee we were trying to than the behaviour is different to `:gen_server`.

fishcakez:
Sorry I misunderstood an email. We do provide this guarantee and it is documented as such. This would be a backwards incompatible change.

michalmuskala:
If we're aiming for different semantics it makes a perfect sense to have different implementation. I wasn't aware of the semantic difference between `GenServer.cast/2` and `:gen_server.cast/2`. Is this something that should be explicitly indicated in the docs?

jvf:
With the new auto-connect implementation in Erlang/OTP 21 (see [otp\_src\_21.0-rc1.readme](http://erlang.org/download/otp_src_21.0-rc1.readme)) the the behaviour of `GenServer.cast/2` and `:gen_server.cast/2` is now identical, neither of them is blocking the caller anymore. See #8407. 

It is my understanding that this also solves the ordering issue for `:gen_server.cast/2`.

josevalim:
@jvf yes, it does. :)

Issue #4251, opened 2016-02-01
GenServer.cast/2 can crash calling process


bharendt:
`&GenServer.cast/2` behaves differently whether event is casted to non-existent name or to non-existent pid. This makes it unsafe to call and make elixir differ significantly from erlang in that regard where it's always safe.

`&GenServer.cast/2` throws `ArgumentError` if no pid is associated with the given name. [This is also documented](http://elixir-lang.org/docs/v1.2/elixir/GenServer.html#cast/2).
That differs from the behaviour if an event is casted to a non-existent pid and also differs from the erlang implementation of [`&:gen_server.cast/2`](http://erlang.org/doc/man/gen_server.html#cast-2).

```elixir
iex> non_existing_pid = pid(0,12345,0)
iex> GenServer.cast(non_existing_pid, :foo)
:ok
iex> GenServer.cast(:non_existing_name, :foo)
** (ArgumentError) argument error
             :erlang.send(:non_existing_name, {:"$gen_cast", :foo})
    (elixir) lib/gen_server.ex:639: GenServer.do_send/2
iex> :gen_server.cast(:non_existing_name, :foo)
:ok
```

 Is that really the desired behaviour? Especially if it's not expected to know whether a casted event is handled in any case which also the documentation says: 
 *"Therefore it is unknown whether the destination server successfully handled the message"*. 
 
 This makes arbitrary processes to crash when they cast to named processes that don't exist or (temporarily) died, e.g. if a named process crashed and is restarted by a supervisor. Example:

```elixir
defmodule NamedServer do
  use GenServer

  def handle_cast({:log, something = <<_,_::binary>>}, state) do
    IO.puts "logged #{something}"
    {:noreply, state}
  end
  
  # crashes on any other event
end


import Supervisor.Spec
children = [
  worker(GenServer, [NamedServer, [], [name: :named_server]])
]
{:ok, _} = Supervisor.start_link(children, strategy: :one_for_one)
is_pid(Process.whereis(:named_server)) == true

arbitrary_pid = Process.spawn fn ->
  GenServer.cast(:named_server, {:log, "hello"}) # everything is fine
  GenServer.cast(:named_server, {:log, :crashing_argument}) # named server crashes ...
  # this process does not notice that, which is ok
  GenServer.cast(:named_server, {:log, "hello again"}) # this process crashes ...
  # even if it does not care so much about whether the message is logged in that case.
  # the :hello message from the cast before was not logged either
end, []

[...]

PR #8407, opened 2018-11-19
Fix GenServer.cast/2 doc

jvf:
### Environment

* Elixir & Erlang/OTP versions (elixir --version): Elixir 1.7.4 (compiled with Erlang/OTP 21)
* Operating system: macOS 10.13.6

### Current behavior

The [docs of `GenServer.cast/2`][0] currently state: 

> `handle_cast/2` will be called on the server to handle the request. In case the `server` is on a node which is not yet connected to the caller one, the call is going to block until a connection happens. This is different than the behaviour in OTP's `:gen_server` where the message is sent by another process in this case, which could cause messages to other nodes to arrive out of order.

### Expected behavior

The above documentation is not correct when running on top of Erlang/OTP 21. In Erlang/OTP 21 the C implementation of `erlang:send/3` was changed to implement "truly asynchronous auto-connect", as described in the [otp\_src\_21.0-rc1.readme]:

> Truly asynchronous auto-connect. Earlier, when  
erlang:send was done toward an unconnected node, the  
function would not return until the connection setup  
had completed (or failed). Now the function returns  
directly after the signal has been enqueued and the  
connection setup started. 

As Elixir is using `erlang:send` this means that the semantics of `GenServer.cast/2` now depend on the underlying version of Erlang OTP. If Elixir is running on top of OTP 21 a cast to a remote, non-connected node is not going to block anymore but is going to return immediately (i.e. the semantics of casts are now identical in Erlang and Elixir). This also means the current documentation of [`GenServer.cast/2`][0] is incomplete.

The pull request changes the documentation to reflect this behaviour.

[0]: https://github.com/elixir-lang/elixir/blob/38f35fd4487748623492d4be477e37c978d8529a/lib/elixir/lib/gen_server.ex#L983
[otp\_src\_21.0-rc1.readme]: http://erlang.org/download/otp_src_21.0-rc1.readme

### DIFF
--- a/lib/elixir/lib/gen_server.ex
+++ b/lib/elixir/lib/gen_server.ex
@@ -983,14 +983,12 @@ defmodule GenServer do
   `c:handle_cast/2` will be called on the server to handle
   the request. In case the `server` is on a node which is
   not yet connected to the caller one, the semantics differ
-  depending on the used Erlang version.
-  Before OTP 21 the call is going to block until a connection
-  happens. This is different than the behaviour in OTP's
-  `:gen_server` where the message is sent by another process
-  in this case, which could cause messages to other nodes to
-  arrive out of order.
-  Starting with OTP 21 the call is not going to block, the
-  behaviour of `GenServer` and `:gen_server` is identical.
+  depending on the used Erlang/OTP version.
+
+  Before Erlang/OTP 21, the call is going to block until a
+  connection happens. This was made to address an ordering
+  bug in Erlang's own `:gen_server.cast/2`. Starting with
+  Erlang/OTP 21, both Erlang and Elixir do not block the call.
   """
   @spec cast(server, term) :: :ok
   def cast(server, request)
--- a/lib/elixir/lib/gen_server.ex
+++ b/lib/elixir/lib/gen_server.ex
@@ -990,7 +990,7 @@ defmodule GenServer do
   in this case, which could cause messages to other nodes to
   arrive out of order.
   Starting with OTP 21 the call is not going to block, the
-  behaviour of `GenServer` and `:gen_server` is idendical.
+  behaviour of `GenServer` and `:gen_server` is identical.
   """
   @spec cast(server, term) :: :ok
   def cast(server, request)
--- a/lib/elixir/lib/gen_server.ex
+++ b/lib/elixir/lib/gen_server.ex
@@ -986,9 +986,9 @@ defmodule GenServer do
   depending on the used Erlang/OTP version.
 
   Before Erlang/OTP 21, the call is going to block until a
-  connection happens. This was done to guarantee ordering
-  in Erlang's own `:gen_server.cast/2`. Starting with
-  Erlang/OTP 21, both Erlang and Elixir do not block the call.
+  connection happens. This was done to guarantee ordering.
+  Starting with Erlang/OTP 21, both Erlang and Elixir do
+  not block the call.
   """
   @spec cast(server, term) :: :ok
   def cast(server, request)
--- a/lib/elixir/lib/gen_server.ex
+++ b/lib/elixir/lib/gen_server.ex
@@ -986,8 +986,8 @@ defmodule GenServer do
   depending on the used Erlang/OTP version.
 
   Before Erlang/OTP 21, the call is going to block until a
-  connection happens. This was made to address an ordering
-  bug in Erlang's own `:gen_server.cast/2`. Starting with
+  connection happens. This was done to guarantee ordering
+  in Erlang's own `:gen_server.cast/2`. Starting with
   Erlang/OTP 21, both Erlang and Elixir do not block the call.
   """
   @spec cast(server, term) :: :ok
--- a/lib/elixir/lib/gen_server.ex
+++ b/lib/elixir/lib/gen_server.ex
@@ -982,11 +982,15 @@ defmodule GenServer do
 
   `c:handle_cast/2` will be called on the server to handle
   the request. In case the `server` is on a node which is
-  not yet connected to the caller one, the call is going to
-  block until a connection happens. This is different than
-  the behaviour in OTP's `:gen_server` where the message
-  is sent by another process in this case, which could cause
-  messages to other nodes to arrive out of order.
+  not yet connected to the caller one, the semantics differ
+  depending on the used Erlang version.
+  Before OTP 21 the call is going to block until a connection
+  happens. This is different than the behaviour in OTP's
+  `:gen_server` where the message is sent by another process
+  in this case, which could cause messages to other nodes to
+  arrive out of order.
+  Starting with OTP 21 the call is not going to block, the
+  behaviour of `GenServer` and `:gen_server` is idendical.
   """
   @spec cast(server, term) :: :ok
   def cast(server, request)


jvf:
# Semantics of `cast` in Elixir and Erlang - Long version

## Situation OTP < 21

The semantics between `GenServer.cast/2` and `:gen_server.cast/2` differ in how casts to remote, non-connected nodes are handled. This is documented in [`GenServer.cast/2`][0]:

>  `c:handle_cast/2` will be called on the server to handle
  the request. In case the `server` is on a node which is
  not yet connected to the caller one, the call is going to
  block until a connection happens. This is different than
  the behaviour in OTP's `:gen_server` where the message
  is sent by another process in this case, which could cause
  messages to other nodes to arrive out of order.


The Elixir implementation will try connecting synchronously to the node, this takes up to 7 seconds to fail if the node is not reachable. 

Elixir: [lib/elixir/lib/gen_server.ex](
https://github.com/elixir-lang/elixir/blob/38f35fd4487748623492d4be477e37c978d8529a/lib/elixir/lib/gen_server.ex#L1037-L1044):

```elixir
defp do_send(dest, msg) do
  try do
    send(dest, msg)
    :ok
  catch
    _, _ -> :ok
  end
end
```

The Erlang implemention will try to send the message directly. If this fails (i.e. if `send` returns `noconnect`) the send will be delegated to a spawned process, making the connection process asynchrones.

Erlang: [lib/stdlib/src/gen_server.erl](https://github.com/erlang/otp/blob/339e4dc3005d15e7fcea61a401b1bda7f5783fdd/lib/stdlib/src/gen_server.erl#L422-L428)

```erlang
do_send(Dest, Msg) ->
    case catch erlang:send(Dest, Msg, [noconnect]) of
	noconnect ->
	    spawn(erlang, send, [Dest,Msg]);
	Other ->
	    Other
    end.
```

This was first noted by @michalmuskala in https://github.com/elixir-lang/elixir/issues/5440, but deemed intentional behaviour.

### Steps to reproduce

Note: This only works with Erlang/OTP < 21. The VM needs to be started in distributed node, i.e. with a name. 

Using Elixir's `GenServer.cast/2`:

```elixir
$ iex --name jvf@127.0.0.1
iex(jvf@127.0.0.1)1> t1 = System.system_time(:second); :ok = GenServer.cast({:fake_name, :"fake_node@1.2.3.4"}, :fake_msg); t2 = System.system_time(:second); t2 - t1
7
```

Using Erlang's `:gen_server.cast/2`:

```elixir
$ iex --name jvf@127.0.0.1
iex(jvf@127.0.0.1)1> t1 = System.system_time(:second); :ok = :gen_server.cast({:fake_name, :"fake_node@1.2.3.4"}, :fake_msg); t2 = System.system_time(:second); t2 - t1
0
```

## OTP >= 21

In OTP 21 the C implementation of `erlang:send/3` was changed to implement "truly asynchronous auto-connect", as described in the [otp\_src\_21.0-rc1.readme]:


> OTP-14370    Application(s): erts
>
> *** POTENTIAL INCOMPATIBILITY ***
> 
> Truly asynchronous auto-connect. Earlier, when  
erlang:send was done toward an unconnected node, the  
function would not return until the connection setup  
had completed (or failed). Now the function returns  
directly after the signal has been enqueued and the  
connection setup started.  
> 
> The same applies to all distributed operations that may  
trigger auto-connect, i.e. '!', send, link, monitor,  
monitor\_node, exit/2 and group\_leader.  
>
> The interface for all these functions are unchanged as  
they do not return connection failures. The only  
exception is erlang:monitor where a *possible  
incompatibility* is introduced: An attempt to monitor a  
process on a primitive node (such as erl_interface or  
jinterface), where remote process monitoring is not  
implemented, will no longer fail with badarg exception.  
Instead a monitor will be created, but it will only  
supervise the connection to the node.  
>>>


For the source changes introducing the new asynchronous auto-connect see [this diff](https://github.com/erlang/otp/commit/f89fb92384280e2939414287a2ecb8f86a199318#diff-190f9182a0dcf029f6e8810c78e2fd1dR467).

With this the implementation `:gen.server.do_send/2` was changed as well and is now identical to the Elixir implemention, see [lib/stdlib/src/gen_server.erl][1]. 

Hower, the asynchronous auto-connect means that the semantics of `GenServer.cast/2` now depend on the underlying version of Erlang OTP. If Elixir is running on top of OTP 21 a cast to a remote, non-connected node is not going to block anymore but is going to return immediately (i.e. the semantics of casts are now identical in Erlang and Elixir). This also means the current documentation of [`GenServer.cast/2`][0] is incomplete.


[0]: https://github.com/elixir-lang/elixir/blob/38f35fd4487748623492d4be477e37c978d8529a/lib/elixir/lib/gen_server.ex#L983

[lib/elixir/lib/gen_server.ex]: 
https://github.com/elixir-lang/elixir/blob/38f35fd4487748623492d4be477e37c978d8529a/lib/elixir/lib/gen_server.ex#L1037-L1044

[lib/stdlib/src/gen_server.erl]: https://github.com/erlang/otp/blob/339e4dc3005d15e7fcea61a401b1bda7f5783fdd/lib/stdlib/src/gen_server.erl#L422-L428

[otp\_src\_21.0-rc1.readme]: http://erlang.org/download/otp_src_21.0-rc1.readme

[this diff]: https://github.com/erlang/otp/commit/f89fb92384280e2939414287a2ecb8f86a199318#diff-190f9182a0dcf029f6e8810c78e2fd1dR467

[1]: https://github.com/erlang/otp/blob/3e580e039a8eb2467c638503bc526116fe829fec/lib/stdlib/src/gen_server.erl#L439-L444

lib/elixir/lib/gen_server.ex
```diff
@@ -982,11 +982,15 @@ defmodule GenServer do
...
-  messages to other nodes to arrive out of order.
+  not yet connected to the caller one, the semantics differ
+  depending on the used Erlang version.
+  Before OTP 21 the call is going to block until a connection
+  happens. This is different than the behaviour in OTP's
+  `:gen_server` where the message is sent by another process
+  in this case, which could cause messages to other nodes to
+  arrive out of order.
+  Starting with OTP 21 the call is not going to block, the
+  behaviour of `GenServer` and `:gen_server` is idendical.
```
fertapric:
```suggestion
  behaviour of `GenServer` and `:gen_server` is identical.
```

josevalim:
Hi @jvf, I have attempted to streamline the docs. I don't think we no longer need to go into why an old version of Erlang's `gen_server.casts/2` did not work properly. Can you please let me know what you think? Thanks!

jvf:
Hi @josevalim. Looks good to me. I added a small change that makes the description more neutral I hope (it sounded a little less defensive in my option).

Out of curiosity: I saw you and @fishcakez discuss the question of ordering in casts to other nodes here in the Elixir issues, namely #4251 and #5440. In 38ac0b1f6979e024646504c906cbc801b9cb114a you called the Erlang behaviour a "bug". What are the Erlang people saying about this? I did find [this old thread][0] in the Erlang mailing list, but nothing else, neither in the mailing list nor in the  [bugs.erlang.org][1]. Do you have any other resources to read up on this you could point me to?

[0]: http://erlang.org/pipermail/erlang-questions/2008-November/039711.html
[1]: https://bugs.erlang.org/

lib/elixir/lib/gen_server.ex
```diff
@@ -982,11 +982,13 @@ defmodule GenServer do
...
-  not yet connected to the caller one, the call is going to
-  block until a connection happens. This is different than
-  the behaviour in OTP's `:gen_server` where the message
-  is sent by another process in this case, which could cause
-  messages to other nodes to arrive out of order.
+  not yet connected to the caller one, the semantics differ
+  depending on the used Erlang/OTP version.
+
+  Before Erlang/OTP 21, the call is going to block until a
+  connection happens. This was done to guarantee ordering
```
fertapric: I think this is not accurate, because the changes were made to guarantee ordering in Elixir’s one, not Erlang’s.

One thing that is not clear is what happens with the ordering in newer Elixir/Erlang versions.

jvf: > I think this is not accurate, because the changes were made to guarantee ordering in Elixir’s one, not Erlang’s.

You are right, adjusted in d8d5a1a97d0ce7c8ffaa4d21a3731433e88efbd9.

josevalim:
Not sure if it was ever reported as a bug (bugs.erlang.org is relatively recent) but other projects also aimed to fix [the exact same issue](https://github.com/rabbitmq/rabbitmq-common/blob/master/src/gen_server2.erl#L15-L17).

josevalim:

PR #8408, opened 2018-11-19
Improve error message for `BadFunctionError`

### LINKED ISSUE #8406: Confusing error message for BadFunctionError
### Environment

* Elixir & Erlang/OTP versions (elixir --version): Elixir 1.7.3, Erlang 21.0

### Current behavior

I'm seeing errors reported to our error catching service in production that look like this:
`(BadFunctionError) expected a function, got: #Function<5.105936317/1 in MyAppWeb.Users.Worker>`

This was pretty confusing because it's saying it expected a function and didn't get it, but it's showing that it got a function.

I'm not 100% sure that this is the case because I haven't actually fixed the bug yet, but I _think_ I'm seeing this because the reference to this anonymous function was passed from one node to another node, and that other node is running a different release than the node on which that function was originally created, so it can't find the correct function (we're running with Kubernetes, clustering with libcluster).

If this is the case, I think we might want a better error message telling folks what's going on here in case they can't figure out this distribution case for themselves and pointing them towards passing a `{M, F, A}` tuple instead of an anonymous function to prevent this issue (which I think is what's going to fix this).

### Expected behavior

A short description on how you expect the code to behave.

I think a better error might look something like:
`(BadFunctionError) `(BadFunctionError) expected an existing function, got: #Function<5.105936317/1 in MyAppWeb.Users.Worker> which could not be found` or something like that.

devonestes:
Now if the term is a function that couldn't be found because it's
referencing an anonymous function from a previous version of a given
module, it says as much instead of saying it expected a function.

Resolves #8406

### DIFF
--- a/lib/elixir/lib/exception.ex
+++ b/lib/elixir/lib/exception.ex
@@ -788,6 +788,10 @@ defmodule BadFunctionError do
   defexception [:term]
 
   @impl true
+  def message(%{term: term}) when is_function(term) do
+    "function #{inspect(term)} is invalid, likely because it points to an old version of the code"
+  end
+
   def message(exception) do
     "expected a function, got: #{inspect(exception.term)}"
   end
--- a/lib/elixir/test/elixir/kernel/raise_test.exs
+++ b/lib/elixir/test/elixir/kernel/raise_test.exs
@@ -361,6 +361,32 @@ defmodule Kernel.RaiseTest do
       assert result == "expected a function, got: :example"
     end
 
+    test "badfun error when the function is gone" do
+      defmodule BadFunction.Missing do
+        def fun, do: fn -> :ok end
+      end
+
+      fun = BadFunction.Missing.fun()
+
+      :code.delete(BadFunction.Missing)
+
+      defmodule BadFunction.Missing do
+        def fun, do: fn -> :another end
+      end
+
+      :code.purge(BadFunction.Missing)
+
+      result =
+        try do
+          fun.()
+        rescue
+          x in [BadFunctionError] -> Exception.message(x)
+        end
+
+      assert result =~
+               ~r/function #Function<[0-9]\.[0-9]*\/0 in Kernel.RaiseTest.BadFunction.Missing> is invalid, likely because it points to an old version of the code/
+    end
+
     test "badmatch error" do
       x = :example
 


lib/elixir/lib/exception.ex
```diff
@@ -787,6 +787,11 @@ end
 defmodule BadFunctionError do
   defexception [:term]
 
+  @impl true
+  def message(%{term: term} = exception) when is_function(term) do
+    "function #{term} does not exist in this application version"
+  end
+
   @impl true
```
fertapric: I think this `@impl` is not needed

lib/elixir/lib/exception.ex
```diff
@@ -787,6 +787,11 @@ end
 defmodule BadFunctionError do
   defexception [:term]
 
+  @impl true
+  def message(%{term: term} = exception) when is_function(term) do
```
fertapric: Unused variable
```suggestion
  def message(%{term: term}) when is_function(term) do
```

lib/elixir/lib/exception.ex
```diff
@@ -788,6 +788,10 @@ defmodule BadFunctionError do
   defexception [:term]
 
   @impl true
+  def message(%{term: term}) when is_function(term) do
+    "function #{term} does not exist in this application version"
```
josevalim: We need a test for this because it is missing `#{inspect(term)}`. Here is a minimal script to reproduce it:

```elixir
defmodule Foo do
  def fun, do: fn -> :ok end
end

fun = Foo.fun

defmodule Foo do
  def fun, do: fn -> :another end
end

:code.purge(Foo)
fun.()
```

Also, this is not related to the application version. It just happens it can be triggered by application versions. My suggestion would be something similar to what you proposed:

> function #{inspect(term)} is invalid, likely because it points an old version of the code

lib/elixir/test/elixir/kernel/raise_test.exs
```diff
@@ -361,6 +361,34 @@ defmodule Kernel.RaiseTest do
...
     end
 
+    test "badfun error when the function is gone" do
+      defmodule Foo do
+        def fun, do: fn -> :ok end
+      end
+
+      fun = Foo.fun()
+
+      :code.delete(Foo)
```
devonestes: This was needed to avoid a compiler warning about redefining module `Foo`.

lib/elixir/test/elixir/kernel/raise_test.exs
```diff
@@ -361,6 +361,34 @@ defmodule Kernel.RaiseTest do
       assert result == "expected a function, got: :example"
     end
 
+    test "badfun error when the function is gone" do
+      defmodule Foo do
```
josevalim: Instead of calling it `Foo`, which has a high chance of conflicts, maybe we should call it `BadFunction`? Thoughts?

lib/elixir/test/elixir/kernel/raise_test.exs
```diff
@@ -361,6 +361,34 @@ defmodule Kernel.RaiseTest do
...
+        try do
+          fun.()
+        rescue
+          x in [BadFunctionError] -> Exception.message(x)
+        end
+
+      regex =
+        ~r/function #Function<[0-9]\.[0-9]*\/0 in Kernel.RaiseTest.Foo> is invalid, likely because it points to an old version of the code/
+
+      assert Regex.match?(regex, result)
```
fertapric: Minor nitpick: 

```elixir
assert result =~ ~r/function #Function<[0-9]\.[0-9]*\/0 in Kernel.RaiseTest.Foo> is invalid, likely because it points to an old version of the code/

# or even 

assert result =~ ~r/function #Function<*\/0 in Kernel.RaiseTest.Foo> is invalid/ 
assert result =~ ~r/likely because it points to an old version of the code/
```

josevalim:

PR #8409, opened 2018-11-20
Always inspect MapSet content as regular lists

michalmuskala:


lib/elixir/test/elixir/map_set_test.exs
```diff
@@ -138,6 +138,10 @@ defmodule MapSetTest do
     refute MapSet.subset?(map_set_v1(7..12), MapSet.new(1..10))
   end
 
+  test "inspect" do
+    assert inspect(MapSet.new([?a])) == "#MapSet<[#{?a}]>"
```
wojtekmach: wdyt about being very explicit here? This mimics `lib/elixir/test/elixir/inspect_test.exs`.

```suggestion
    assert inspect(MapSet.new([?a])) == "#MapSet<[97]>"
```

lib/elixir/lib/map_set.ex
```diff
@@ -398,6 +398,7 @@ defmodule MapSet do
     import Inspect.Algebra
 
     def inspect(map_set, opts) do
+      opts = %Inspect.Opts{opts | charlists: :as_lists}
```
wojtekmach: I think it's better default but can't help but wonder if folks would ever prefer to opt-in to the previous behaviour. E.g. they do `inspect(map_set, chartists: :as_charlists)` and now the option is ignored

PR #8410, opened 2018-11-20
Improve error message on DateTime.add/4

fertapric:
Using `inspect` instead of interpolating directly the date time value (like other error messages in this module).


michalmuskala:
Thank you! :heart:

PR #8413, opened 2018-11-21
Optimize EEx template rendering

josevalim:
Consider the following template:

    one
    <%= arg1 %>
    two
    <% var2 = String.duplicate(arg2, 1) %>
    three
    <%= var2 %>
    four
    <% var3 = String.duplicate(arg3, 1) %>
    five
    <%= var3 %>
    six

Before this patch, it would compile to:

    (
      tmp1 = (
        tmp2 = (
          tmp1 = (
            tmp2 = (
              tmp1 = "" <> "one\n"
              tmp1 <> String.Chars.to_string(arg1)
            ) <> "\ntwo\n"
            var2 = String.duplicate(arg2, 1)
            tmp2
          ) <> "\nthree\n"
          tmp1 <> String.Chars.to_string(var2)
        ) <> "\nfour\n"
        var3 = String.duplicate(arg3, 1)
        tmp2
      ) <> "\nfive\n"
      tmp1 <> String.Chars.to_string(var3)
    ) <> "\nsix\n"

The issue with this approach is that it emits binary append
instructions, frequently copying binaries. This commit changes
it to extract all of the dynamic contents out, as originally
suggested by @tmbb:

    tmp0 = String.Chars.to_string(arg1)
    var2 = String.duplicate(arg2, 1)
    tmp1 = String.Chars.to_string(var2)
    var3 = String.duplicate(arg3, 1)
    tmp2 = String.Chars.to_string(var3)
    <<"one\n", tmp0::binary, "\ntwo\n", "\nthree\n", tmp1::binary, "\nfour\n", "\nfive\n", tmp2::binary, "\nsix\n">>

Which leads to code that compiles faster and also run faster
as there are less instructions.

### DIFF
--- a/lib/eex/lib/eex/engine.ex
+++ b/lib/eex/lib/eex/engine.ex
@@ -139,35 +139,43 @@ defmodule EEx.Engine do
   Returns an empty string as initial buffer.
   """
   def init(_opts) do
-    ""
+    %{
+      binary: [],
+      dynamic: [],
+      vars_count: 0
+    }
   end
 
   @doc """
   Returns an empty string as the new buffer.
   """
-  def handle_begin(_previous) do
-    ""
+  def handle_begin(state) do
+    %{state | binary: [], dynamic: []}
   end
 
   @doc """
   End of the new buffer.
   """
   def handle_end(quoted) do
-    quoted
+    handle_body(quoted)
   end
 
   @doc """
   The default implementation simply returns the given expression.
   """
-  def handle_body(quoted) do
-    quoted
+  def handle_body(state) do
+    %{binary: binary, dynamic: dynamic} = state
+    binary = {:<<>>, [], Enum.reverse(binary)}
+    dynamic = [binary | dynamic]
+    {:__block__, [], Enum.reverse(dynamic)}
   end
 
   @doc """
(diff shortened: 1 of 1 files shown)


lib/eex/lib/eex/engine.ex
```diff
@@ -139,35 +139,43 @@ defmodule EEx.Engine do
...
   @doc """
   The default implementation simply returns the given expression.
   """
-  def handle_body(quoted) do
-    quoted
+  def handle_body(state) do
+    %{binary: binary, dynamic: dynamic} = state
+    binary = {:<<>>, [], Enum.reverse(binary)}
+    dynamic = [binary | dynamic]
+    {:__block__, [], Enum.reverse(dynamic)}
```
michalmuskala: What do you think about `Enum.reverse(dynamic, [binary])`?

tmbb:
Have you benchmarked this? I wonder what the real performance impact is.

michallepicki:
Documentation needs to change or get removed, e.g.:
>   Returns an empty string as the new buffer.
is no longer true after this got merged in.

josevalim:
@michallepicki good catch. Docs updated in another commit.

PR #8419, opened 2018-11-23
Use --module value when setting up filenames in mix new

kevinbader:
The module name was used in the templates, but the app name was used for
the file names. Now the module name is used for both.

For example:

```bash
mkdir foo
cd foo
mix new . --module Bar
```

...creates a `Bar` module as expected, but the module's filename is `lib/foo.ex`. This, of course, should be `lib/bar.ex` instead. The reason for the wrong filename is that the code currently uses the (inferred) app name when creating the file, which in this case is `foo`.

### DIFF
--- a/lib/mix/lib/mix/tasks/new.ex
+++ b/lib/mix/lib/mix/tasks/new.ex
@@ -93,6 +93,8 @@ defmodule Mix.Tasks.New do
       version: get_version(System.version())
     ]
 
+    mod_filename = Macro.underscore(mod)
+
     create_file("README.md", readme_template(assigns))
     create_file(".formatter.exs", formatter_template(assigns))
     create_file(".gitignore", gitignore_template(assigns))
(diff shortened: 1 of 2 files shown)


kevinbader:
I'd love to test this properly, but frankly I don't know how to compile the mix binary :/ Seems it comes with the checkout and `make compile` doesn't update it. Since I cannot find anything related in the README either, I thought why not create a PR first and ask there: how can I recompile Mix? :)

josevalim:
Beautiful! Can we have a test too? Thanks!
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

kevinbader:
@josevalim sure, here you go. I still wonder.. how can I compile Mix into a binary?

josevalim:
@kevinbader I believe you can call `MIX_ENV=prod mix escript.build` inside lib/mix? We were used to do it on the past but we believe nobody used it.

michalmuskala:
Thank you! :heart:

PR #4876, opened 2016-06-24
Generate error for no pid given to Genserver.reply

benfalk:
Prior to this change it is possible to send any two-pair tuple to
`GenServer.reply/2` and it will happily reply with `:ok`.  This
leads to some hard to track-down bugs if you pass it the wrong
data.  With this small check developers may be able to get a bit
more of a heads up to what is going wrong sooner.

For more information on how I came accross this:
https://github.com/benfalk/data_pool/commit/dcf81d1a87a363952715deeed3caa18ec0efccd9

PR #8420, opened 2018-11-25
Remove unnecessary try block in GenServer.reply

gyson:
Based on Erlang reference [here](http://erlang.org/doc/reference_manual/expressions.html#send):

```
Sending a message to a pid never fails, even if the pid identifies a non-existing process.
```

We do not need try block anymore because there is a `is_pid(to)` guard, which was added from #4876.

Therefore, try block within `GenServer.reply` is unnecessary and [confusing](https://elixirforum.com/t/when-would-send-raise-an-error/18177). We should remove it.


josevalim:

PR #8421, opened 2018-11-25
[Discussion] Add Calendar callbacks and Calendar.ISO implementations

kipcole9:
This PR adds 4 of the proposed new `Calendar` callbacks and implementation in the `Calendar.ISO` calendar module.  The `week_in_year/3` and `week_in_month/3` implementations are close, I wasn't happy with my previous implementation.

* day_of_year/3
* quarter_of_year/3
* year_of_era/1
* day_of_era/3

### DIFF
--- a/lib/elixir/lib/calendar.ex
+++ b/lib/elixir/lib/calendar.ex
@@ -137,6 +137,26 @@ defmodule Calendar do
   @callback day_of_week(year, month, day) :: non_neg_integer()
 
   @doc """
+  Calculates the day of the year from the given `year`, `month`, and `day`.
+  """
+  @callback day_of_year(year, month, day) :: non_neg_integer()
+
+  @doc """
+  Calculates the quarter of the year from the given `year`, `month`, and `day`.
+  """
+  @callback quarter_of_year(year, month, day) :: non_neg_integer()
+
+  @doc """
+  Calculates the year and era from the given `year`.
+  """
+  @callback year_of_era(year) :: {year, era :: non_neg_integer()}
+
+  @doc """
+  Calculates the day and era from the given `year`, `month`, and `day`.
+  """
+  @callback day_of_era(year, month, day) :: {day, era :: non_neg_integer()}
+
+  @doc """
   Converts the date into a string according to the calendar.
   """
   @callback date_to_string(year, month, day) :: String.t()
--- a/lib/elixir/lib/calendar/date.ex
+++ b/lib/elixir/lib/calendar/date.ex
@@ -711,6 +711,54 @@ defmodule Date do
     calendar.quarter_of_year(year, month, day)
   end
 
+  @doc """
+  Calculates the year-of-era and era for a given
+  calendar year.
+
+  Returns a tuple `{year, era}` representing the
+  year within the era and the era number.
+
+  ## Examples
+
+      iex> Date.year_of_era(~D[0001-01-01])
+      {1, 1}
+      iex> Date.year_of_era(~D[0000-12-31])
+      {1, 0}
+      iex> Date.year_of_era(~D[-0001-01-01])
+      {2, 0}
+
+  """
+  @doc since: "1.8.0"
+  @spec year_of_era(Calendar.date) :: {Calendar.year, non_neg_integer()}
+  def year_of_era(date)
+
+  def year_of_era(%{calendar: calendar, year: year}) do
+    calendar.year_of_era(year)
+  end
+
+  @doc """
+  Calculates the day-of-era and era for a given
+  calendar `date`.
+
+  Returns a tuple `{day, era}` representing the
+  day within the era and the era number.
+
+  ## Examples
+
+    iex> Date.day_of_era(~D[0001-01-01])
+    {1, 1}
+    iex> Date.day_of_era(~D[0000-12-31])
+    {1, 0}
+
+  """
+  @doc since: "1.8.0"
+  @spec day_of_era(Calendar.date) :: {Calendar.day, non_neg_integer()}
+  def day_of_era(date)
+
+  def day_of_era(%{calendar: calendar, year: year, month: month, day: day}) do
+    calendar.day_of_era(year, month, day)
+  end
+
   ## Helpers
 
   defimpl String.Chars do
--- a/lib/elixir/lib/calendar/date.ex
+++ b/lib/elixir/lib/calendar/date.ex
@@ -659,6 +659,58 @@ defmodule Date do
     calendar.day_of_week(year, month, day)
   end
 
+  @doc """
+  Calculates the day of the year of a given `date`.
+
+  Returns the day of the year as an integer. For the ISO 8601
+  calendar (the default), it is an integer from 1 to 366.
+
+  ## Examples
+
+      iex> Date.day_of_year(~D[2016-01-01])
+      1
+      iex> Date.day_of_year(~D[2016-11-01])
+      306
+      iex> Date.day_of_year(~D[-0015-10-30])
+      303
+      iex> Date.day_of_year(~D[2004-12-31])
+      366
+
+  """
+  @doc since: "1.8.0"
+  @spec day_of_year(Calendar.date()) :: non_neg_integer()
+  def day_of_year(date)
+
+  def day_of_year(%{calendar: calendar, year: year, month: month, day: day}) do
+    calendar.day_of_year(year, month, day)
+  end
+
+  @doc """
+  Calculates the quarter of the year of a given `date`.
+
+  Returns the day of the year as an integer. For the ISO 8601
+  calendar (the default), it is an integer from 1 to 4.
+
+  ## Examples
+
+      iex> Date.quarter_of_year(~D[2016-10-31])
+      4
+      iex> Date.quarter_of_year(~D[2016-01-01])
+      1
+      iex> Date.quarter_of_year(~N[2016-04-01 01:23:45])
+      2
+      iex> Date.quarter_of_year(~D[-0015-09-30])
+      3
+
+  """
+  @doc since: "1.8.0"
+  @spec quarter_of_year(Calendar.date()) :: non_neg_integer()
+  def quarter_of_year(date)
+
+  def quarter_of_year(%{calendar: calendar, year: year, month: month, day: day}) do
+    calendar.quarter_of_year(year, month, day)
+  end
+
   ## Helpers
 
   defimpl String.Chars do
--- a/lib/elixir/lib/calendar/iso.ex
+++ b/lib/elixir/lib/calendar/iso.ex
@@ -383,7 +383,7 @@ defmodule Calendar.ISO do
 
   """
   @doc since: "1.8.0"
-  @spec quarter_of_year(year, month, day) :: Calendar.quarter()
+  @spec quarter_of_year(year, month, day) :: 1..4
   @impl true
   def quarter_of_year(year, month, day)
       when is_integer(year) and is_integer(month) and is_integer(day) do
--- a/lib/elixir/lib/calendar/iso.ex
+++ b/lib/elixir/lib/calendar/iso.ex
@@ -358,7 +358,7 @@ defmodule Calendar.ISO do
 
   """
   @doc since: "1.8.0"
-  @spec day_of_year(year, month, day) :: day
+  @spec day_of_year(year, month, day) :: 1..366
   @impl true
   def day_of_year(year, month, day)
       when is_integer(year) and is_integer(month) and is_integer(day) do
--- a/lib/elixir/lib/calendar/iso.ex
+++ b/lib/elixir/lib/calendar/iso.ex
@@ -37,6 +37,11 @@ defmodule Calendar.ISO do
 
   @months_in_year 12
 
+  # The ISO epoch starts, in this implementation,
+  # with ~D[0000-01-01]. Era "1" starts
+  # on ~D[0001-01-01] which is 366 days later.
+  @iso_epoch 366
+
   @doc false
   def __match_date__ do
     quote do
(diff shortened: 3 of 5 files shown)


kipcole9:
I'm vaguely uncomfortable with years in Calendar.ISO's "0" era being positive for negative time.  But I suppose all calendars face that issue when they move past the last defined era.

lib/elixir/lib/calendar/iso.ex
```diff
@@ -337,6 +342,122 @@ defmodule Calendar.ISO do
...
+      iex> Calendar.ISO.day_of_year(2018, 2, 28)
+      59
+
+  """
+  @doc since: "1.8.0"
+  @spec day_of_year(year, month, day) :: day
+  @impl true
+  def day_of_year(year, month, day)
+      when is_integer(year) and is_integer(month) and is_integer(day) do
+    date_to_iso_days(year, month, day) - date_to_iso_days(year, 1, 1) + 1
```
josevalim: This implementation is a bit inefficient. I think we can "inline" the code in `date_to_iso_days` to something like this:

```elixir
true = day <= days_in_month(year, month)
days_before_month(month) + leap_day_offset(year, month) + day
```

Give or take +1/-1. :)

lib/elixir/lib/calendar/date.ex
```diff
@@ -659,6 +659,58 @@ defmodule Date do
...
+
+  """
+  @doc since: "1.8.0"
+  @spec quarter_of_year(Calendar.date()) :: non_neg_integer()
+  def quarter_of_year(date)
+
+  def quarter_of_year(%{calendar: calendar, year: year, month: month, day: day}) do
+    calendar.quarter_of_year(year, month, day)
+  end
+
```
josevalim: We can also surface the other two functions to the `Date` module. :)

josevalim:
@kipcole9 this is excellent. I have added to comments and we should be good to go. I like that you did not tackle all of them at once, it makes it easier to review. :)

lib/elixir/lib/calendar/date.ex
kipcole9: I just notice that myself - being added to the PR in a few minutes

lib/elixir/lib/calendar/iso.ex
kipcole9: Makes sense, I'll get it done.

lib/elixir/lib/calendar/date.ex
```diff
@@ -659,6 +659,106 @@ defmodule Date do
...
+  def day_of_year(%{calendar: calendar, year: year, month: month, day: day}) do
+    calendar.day_of_year(year, month, day)
+  end
+
+  @doc """
+  Calculates the quarter of the year of a given `date`.
+
+  Returns the day of the year as an integer. For the ISO 8601
+  calendar (the default), it is an integer from 1 to 4.
```
wojtekmach: how about just the following?

```
For the ISO calendar (the default), it is an integer from 1 to 4.
```

lib/elixir/lib/calendar/iso.ex
```diff
@@ -337,6 +342,122 @@ defmodule Calendar.ISO do
...
+      iex> Calendar.ISO.day_of_year(2016, 1, 31)
+      31
+      iex> Calendar.ISO.day_of_year(-99, 2, 1)
+      32
+      iex> Calendar.ISO.day_of_year(2018, 2, 28)
+      59
+
+  """
+  @doc since: "1.8.0"
+  @spec day_of_year(year, month, day) :: day
```
wojtekmach:
```suggestion
  @spec day_of_year(year, month, day) :: 1..366
```

lib/elixir/lib/calendar/iso.ex
```diff
@@ -337,6 +342,122 @@ defmodule Calendar.ISO do
...
+      iex> Calendar.ISO.quarter_of_year(2016, 4, 3)
+      2
+      iex> Calendar.ISO.quarter_of_year(-99, 9, 31)
+      3
+      iex> Calendar.ISO.quarter_of_year(2018, 12, 28)
+      4
+
+  """
+  @doc since: "1.8.0"
+  @spec quarter_of_year(year, month, day) :: Calendar.quarter()
```
wojtekmach: I think leaving as is fine but consider more specific type:

```suggestion
  @spec quarter_of_year(year, month, day) :: 1..4
```

lib/elixir/lib/calendar/iso.ex
```diff
@@ -337,6 +342,122 @@ defmodule Calendar.ISO do
...
+      iex> Calendar.ISO.year_of_era(2018)
+      {2018, 1}
+      iex> Calendar.ISO.year_of_era(0)
+      {1, 0}
+      iex> Calendar.ISO.year_of_era(-1)
+      {2, 0}
+
+  """
+  @doc since: "1.8.0"
+  @spec year_of_era(year) :: {year, era :: non_neg_integer()}
```
wojtekmach: I think leaving as is fine but consider more specific type:

```suggestion
  @spec year_of_era(year) :: {year, era :: 0..1}
```

lib/elixir/lib/calendar/iso.ex
```diff
@@ -337,6 +342,122 @@ defmodule Calendar.ISO do
...
+      iex> Calendar.ISO.day_of_era(0, 12, 31)
+      {1, 0}
+      iex> Calendar.ISO.day_of_era(0, 12, 30)
+      {2, 0}
+      iex> Calendar.ISO.day_of_era(-1, 12, 31)
+      {367, 0}
+
+  """
+  @doc since: "1.8.0"
+  @spec day_of_era(year, month, day) :: {day, era :: non_neg_integer()}
```
wojtekmach:
```suggestion
  @spec day_of_era(year, month, day) :: {day :: pos_integer(), era :: 0..1}
```

lib/elixir/lib/calendar/date.ex
```diff
@@ -659,6 +659,106 @@ defmodule Date do
     calendar.day_of_week(year, month, day)
   end
 
+  @doc """
+  Calculates the day of the year of a given `date`.
+
+  Returns the day of the year as an integer. For the ISO 8601
+  calendar (the default), it is an integer from 1 to 366.
```
wojtekmach: how about just the following?

```
For the ISO calendar (the default), it is an integer from 1 to 366.
```

kipcole9: Thanks for the feedback,  makes sense for the @specs since they are calendar specific.

lib/elixir/lib/calendar/date.ex
kipcole9: I think the nuance here is that although Calendar.ISO is the default, there are other possibilities. The first sentence is saying “all calendars will return an integer”.  THe second sentence say “and for the default here is the range”.

lib/elixir/test/elixir/calendar/date_test.exs
```diff
@@ -53,6 +53,21 @@ defmodule DateTest do
...
+    assert Date.day_of_year(~D[2018-01-01]) == 1
+    assert Date.day_of_year(~D[2018-12-31]) == 365
+  end
+
+  test "quarter_of_year/1" do
+    assert Date.quarter_of_year(~D[2004-03-01]) == 1
+    assert Date.quarter_of_year(~D[2015-04-01]) == 2
+    assert Date.quarter_of_year(~D[2018-09-01]) == 3
+    assert Date.quarter_of_year(~D[2004-12-31]) == 4
+  end
```
josevalim: Is there a reason why we test those two but not the other ones? :)

lib/elixir/lib/calendar/iso.ex
```diff
@@ -337,6 +342,122 @@ defmodule Calendar.ISO do
...
+  @doc since: "1.8.0"
+  @spec quarter_of_year(year, month, day) :: 1..4
+  @impl true
+  def quarter_of_year(year, month, day)
+      when is_integer(year) and is_integer(month) and is_integer(day) do
+    cond do
+      month <= 3 -> 1
+      month <= 6 -> 2
+      month <= 9 -> 3
+      true -> 4
```
michalmuskala: it should be slightly more efficient to do something like:
```elixir
cond do
  month in [1, 2, 3] -> 1
  month in [4, 5, 6] -> 2
  month in [7, 8, 9] -> 3
  month in [10, 11, 12] -> 4
end
```
it will be compiled into an optimised multi-way integer branch in the bytecode

lib/elixir/lib/calendar/iso.ex
```diff
@@ -337,6 +342,122 @@ defmodule Calendar.ISO do
...
+
+  """
+  @doc since: "1.8.0"
+  @spec year_of_era(year) :: {year, era :: 0..1}
+  @impl true
+  def year_of_era(year) when is_integer(year) and year > 0 do
+    {year, 1}
+  end
+
+  def year_of_era(year) when is_integer(year) and year < 1 do
```
michalmuskala: We could probably leave out the year check in the other clause since it will always be true. Unfortunately, currently the compiler is not able to make that deduction, so it leaves in the check in the final code.

lib/elixir/lib/calendar/iso.ex
josevalim: Or `div(month - 1, 3)`.

I just realized now though that quarter is something very specific to the ISO calendar possibly? Do other calendars have the concept of quarters? Maybe it would be best to drop the function for now.

michalmuskala: As far as I remember, we settled on if after going though the spec for i18n of dates and it was there, so I'm not sure.

josevalim:
> Integers in the documentation should preferable by surrounded by backticks

Are we sure about that? I think it is ok to refer to general programming terms without the backtick, especially if we are not referring to a particular variable or particular Elixir construct. If we just surround everything by backtick, then it makes it harder to distinguish the other cases.

lib/elixir/lib/calendar/iso.ex
josevalim: This check will hardly matter though in terms of performance though. We can leave it, it is fine.

lib/elixir/lib/calendar.ex
```diff
@@ -136,6 +136,26 @@ defmodule Calendar do
...
+
+  @doc """
+  Calculates the quarter of the year from the given `year`, `month`, and `day`.
+  """
+  @callback quarter_of_year(year, month, day) :: non_neg_integer()
+
+  @doc """
+  Calculates the year and era from the given `year`.
+  """
+  @callback year_of_era(year) :: {year, era :: non_neg_integer()}
```
fertapric: What do you think about defining the `era` type? It could be reused in the rest of the implementations

lib/elixir/lib/calendar/date.ex
```diff
@@ -659,6 +659,106 @@ defmodule Date do
...
+      iex> Date.day_of_year(~D[2016-11-01])
+      306
+      iex> Date.day_of_year(~D[-0015-10-30])
+      303
+      iex> Date.day_of_year(~D[2004-12-31])
+      366
+
+  """
+  @doc since: "1.8.0"
+  @spec day_of_year(Calendar.date()) :: non_neg_integer()
```
fertapric: Maybe?

```suggestion
  @spec day_of_year(Calendar.date()) :: Calendar.day()
```

fertapric: Ooops! Nop. But maybe `pos_integer`, or this could return 0?

lib/elixir/lib/calendar/iso.ex
kipcole9: There are lots of calendars that are derivatives of the Gregorian calendar so whilst not all calendars may have that concept, many do.  Perhaps it reasonable to also permit a `false` return?

Great on the `div` version.  I spend 30 minutes trying to derive that :-)

kipcole9: And yes, in i18n and [CLDR](https://cldr.unicode.org) each locale includes quarter information for all calendars that apply it (I'll check the non-gregorian based ones to confirm).

That said, CLDR also includes the concept of work day and weekend but I recall there wasn't much (any?) support for that being included.

lib/elixir/lib/calendar/date.ex
kipcole9: I'm not aware of any calendars that are 0-based for days (I'll check again) so I think `pos_integer` is probably fine.  `Calendar.day` is defined as `integer`.  I wonder if that should be `pos_integer`?

lib/elixir/test/elixir/calendar/date_test.exs
kipcole9: Apart from oversight, no.  Will add them.

lib/elixir/lib/calendar/iso.ex
josevalim: @kipcole9 the current improvements were based on the APIs needed for ICU. We will be glad to improve the feature set on other areas, we just need a proposal. :)

eksperimental:
> > Integers in the documentation should preferable by surrounded by backticks
> 
> Are we sure about that? I think it is ok to refer to general programming terms without the backtick, especially if we are not referring to a particular variable or particular Elixir construct. If we just surround everything by backtick, then it makes it harder to distinguish the other cases.

Sorry, that was poorly written (from my mobile BTW)
What I meant is that integers that are terms in Elixir should surrounded by backticks (maybe we are talking about the same thing here).
Suchs as:

```
  Returns the day of the year as an integer. For the ISO 8601
   calendar (the default), it is an integer from `1` to `366`.
```
8601 shouldn't be surrounded by backticks, but `1` and `366` should.

lib/elixir/lib/calendar/date.ex
josevalim: @kipcole9 I think so, yes. :+1:

josevalim:
I have merged this locally to tackle whatever is left and I will push soon. Thank you!

josevalim:
Merged manually, thank you! :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

PR #8425, opened 2018-11-26
Replace calls to :lists Erlang module in modules not used during bootstrapping

eksperimental:


josevalim:
Thanks @eksperimental but I would like not to do this because this is very fickle. Maybe a single change in the compiler or in the parallel compiler would make those be used, and then we will need to spend minutes hunting bootstrapping issues.

So except by the changes outside of lib/elixir, I really think we should continue using `:lists`.

lib/iex/lib/iex/config.ex
```diff
@@ -157,7 +157,7 @@ defmodule IEx.Config do
   end
 
   def handle_configure(tab, options) do
-    options = :lists.ukeysort(1, options)
+    options = List.keysort(options, 0) |> Enum.uniq_by(fn {k, _} -> k end)
```
eksperimental: I tried to look for a way to do this in only one pass, but I wasn't able to find it, which leads me to think if we need to add a function to the API, or extend the existing one (which is kind of difficult becuase we already have List.keysort/2 and /3) in case we want to add an `:options` field.
If there is interest, I can think of something and write a proposal.

lib/elixir/lib/kernel/cli.ex
```diff
@@ -480,7 +480,8 @@ defmodule Kernel.CLI do
   defp filter_patterns(pattern) do
     pattern
     |> Path.wildcard()
-    |> :lists.usort()
+    |> Enum.sort()
+    |> Enum.uniq()
```
eksperimental: Same here: I tried to look for a way to do this in only one pass, but I wasn't able to find it, which leads me to think if we need to add a function to the API. If there is interest, I can think of something and write a proposal to extend Enum.sort

eksperimental:
> So except by the changes outside of lib/elixir, I really think we should continue using `:lists`.

I will submit a new PR excluding everything under `lib/elixir/lib/`

michalmuskala:
@eksperimental regarding `lists:ukeysort/2` and `lists:usort/1` the consensus was that those are very specific and are primary an optimisation over more naive two-pass algorithms - for such cases it's entirely fine to drop down into things provided by Erlang.

eksperimental:
Thank you for your answer @michalmuskala

PR #8427, opened 2018-11-26
Fix link to fn/1 special form in Kernel docs

eksperimental:


eksperimental:
I think on the ExDoc side, we should emit a warning if the function doesn't exist. This way it would be easier to spot this kind of silly mistakes.

wojtekmach:
> On the ExDoc side I think we should emit a warning if the function doesn't exist. This way it would be easier to spot this kind of silly mistakes.

this may trigger false positives, e.g. someone writes:

```markdown
E.g. if you define function `foo/1`.
```

we definitely don't want to warn there. We could however warn if the module exists but the function doesn't, so e.g. exactly this: `Kernel.SpecialForms.fn/2`.

eksperimental:
> we definitely don't want to warn there. We could however warn if the module exists but the function doesn't, so e.g. exactly this: `Kernel.SpecialForms.fn/2`.

yes, that's exactly what I meant.

josevalim:

PR #8429, opened 2018-11-26
add special fn case to signature_to_binary

### LINKED ISSUE #8428: IEx doc header broken for `fn/1`
### Environment

* Operating system: macOS Mojave

### Current behavior

screenshot 2018-11-26 at 12 04 21

Hanspagh:
Handles the special fn case in the signature_to_binary function. As of now I hardcoded the signature of fn, I am not sure this is the intended behavior? 
Fixes #8428

### DIFF
--- a/lib/elixir/src/elixir_erl.erl
+++ b/lib/elixir/src/elixir_erl.erl
@@ -531,7 +531,7 @@ get_type_docs(Set, Types) ->
 signature_to_binary(_Module, Name, _Signature) when Name == '__aliases__'; Name == '__block__' ->
   <<(atom_to_binary(Name, utf8))/binary, "(args)">>;
 
-signature_to_binary(_Module, Name, _Signature) when Name == fn -> 
+signature_to_binary(_Module, fn, _Signature) -> 
   <<"fn">>;
 
 signature_to_binary(_Module, Name, _Signature)
--- a/lib/elixir/src/elixir_erl.erl
+++ b/lib/elixir/src/elixir_erl.erl
@@ -531,6 +531,9 @@ get_type_docs(Set, Types) ->
 signature_to_binary(_Module, Name, _Signature) when Name == '__aliases__'; Name == '__block__' ->
   <<(atom_to_binary(Name, utf8))/binary, "(args)">>;
 
+signature_to_binary(_Module, Name, _Signature) when Name == fn -> 
+  <<"fn">>;
+
 signature_to_binary(_Module, Name, _Signature)
     when Name == '__CALLER__'; Name == '__DIR__'; Name == '__ENV__';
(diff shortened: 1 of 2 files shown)


lib/elixir/test/elixir/code_test.exs
```diff
@@ -293,6 +293,14 @@ defmodule CodeTest do
       Code.compiler_options(debug_info: :not_a_boolean)
     end
   end
+
+  describe "fecth docs handles special cases correct" do
```
fertapric: There is a [`docs_test.exs` file](https://github.com/elixir-lang/elixir/blob/master/lib/elixir/test/elixir/kernel/docs_test.exs), I would put this test there :)

lib/elixir/src/elixir_erl.erl
```diff
@@ -531,6 +531,9 @@ get_type_docs(Set, Types) ->
 signature_to_binary(_Module, Name, _Signature) when Name == '__aliases__'; Name == '__block__' ->
   <<(atom_to_binary(Name, utf8))/binary, "(args)">>;
 
+signature_to_binary(_Module, Name, _Signature) when Name == fn -> 
+  <<"fn [clauses] end">>;
```
josevalim: I think we can do `fn (clauses) end` or `fn ... end` or simply `fn`.

lib/elixir/test/elixir/code_test.exs
Hanspagh: Makes sense, don't know why I didn't see that file before.

lib/elixir/src/elixir_erl.erl
Hanspagh: Would like make sense to do something like this `atom_to_binary(Name, utf8);` then?

josevalim: In this case we know that the Name is atom, so we can skip the indirection. So we should rather write it like this:

```erlang
signature_to_binary(_Module, fn, _Signature) -> 
  <<"fn">>;
```

Hanspagh: Fixed

lib/elixir/test/elixir/kernel/docs_test.exs
```diff
@@ -307,4 +307,12 @@ defmodule Kernel.DocsTest do
              {{:fuz, 0}, :none}
            ] = Enum.sort(function_docs)
   end
+
+  describe "fecth docs handles special cases correct" do
```
maennchen:
```suggestion
  describe "fetch docs handles special cases correct" do
```

fertapric: Maybe?
 
```elixir
describe "special signatures" do
  test "fn" do
    ...
  end
end
```

PR #8430, opened 2018-11-26
Fix a typo in the Mix.Task.Deps module

pallix:
Fix a typo for the `runtime` option documentation of the Mix.Task.Deps module


whatyouhide:
Thanks @pallix! 💟

PR #8431, opened 2018-11-27
Eliminate warning for an unused variable

bjorng:
Fix a warning that has started to occur because of a recent bug fix in
Erlang/OTP (using `fun M:F/A` could suppress warnings for unused
variables).


josevalim:

PR #8438, opened 2018-11-29
Simplify Makefile by introducing --path option to compile.app task

michalmuskala:


lib/mix/lib/mix/tasks/compile.app.ex
```diff
@@ -82,6 +82,8 @@ defmodule Mix.Tasks.Compile.App do
   ## Command line options
 
     * `--force` - forces compilation regardless of modification times
+    * `--path path` - where to find `.beam` files and write the resulting
+      `.app` file, defaults to `Mix.Project.compile_path/0`
```
josevalim: Because many things call "compile" passing all args, I am afraid of accidentally passing a --path. Maybe we can call it `--compile-path`?

josevalim:
Ping!

michalmuskala:
Updated

lib/mix/lib/mix/tasks/compile.app.ex
```diff
@@ -82,6 +82,8 @@ defmodule Mix.Tasks.Compile.App do
   ## Command line options
 
     * `--force` - forces compilation regardless of modification times
+    * `--compile-path path` - where to find `.beam` files and write the
```
fertapric:
```suggestion
    * `--compile-path` - where to find `.beam` files and write the
```

I think that's the current format. Here are two examples: [`mix format`](https://github.com/elixir-lang/elixir/blob/479d5f9893b6537becb81a89f6a1d7cd8706d81c/lib/mix/lib/mix/tasks/format.ex#L66) and [`mix profile.cprof`](https://github.com/elixir-lang/elixir/blob/479d5f9893b6537becb81a89f6a1d7cd8706d81c/lib/mix/lib/mix/tasks/profile.cprof.ex#L34)

PR #8439, opened 2018-11-30
iex port helper

### LINKED ISSUE #8437: Provide port/1,2 helpers in IEx
Those should be similar to the `pid` and `ref` helpers but work for ports instead. It should be possible to use the `:erlang.list_to_port/1` function to do this.

cas27:
closes #8437 

@michalmuskala I didn't see a use case for /2 lemme know what you were thinking there and I can change this.


michalmuskala:
The first integer would be different then `0` for remote ports - ports from other nodes in the cluster.

lib/iex/lib/iex/helpers.ex
```diff
@@ -1209,6 +1210,34 @@ defmodule IEx.Helpers do
...
+      iex> port(8080)
+      #Port<0.8080>
+      iex> port(443)
+      #Port<0.443>
+
+  """
+  def port(x) when is_integer(x) and x >= 0 do
+    :erlang.list_to_port('#Port<0.#{x}>')
+  end
+
```
whatyouhide: Having two clauses of the same function with two different documentation is not going to work, only the last one (I think? or the first one) is going to be used. I would go with something like:

```elixir
@doc """
Docs for both string and int
"""
def port(port)

def port(string) when is_binary(string), do: ...
def port(int) when is_integer(int) and int >= 0, do: ...
```

josevalim: To be fair I think we should have two differnet functions. `port(string)` and `port(major, minor)`.

whatyouhide: Yep that's fair, let's go with two!

lib/iex/lib/iex/helpers.ex
```diff
@@ -1209,6 +1210,34 @@ defmodule IEx.Helpers do
...
+  @doc """
+  Creates a Port from `string`.
+
+  ## Examples
+
+      iex> port("0.4")
+      #Port<0.4>
+
+  """
+  def port(string) when is_binary(string) do
```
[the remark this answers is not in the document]
josevalim: IF we are going to do that we should do it for PIDs too. So I would treat that as a separate discussion and focus this PR on having feature parity to the existing feature set.

lib/iex/lib/iex/helpers.ex
cas27: Will do.  I was looking for an example of declaring a major minor port and am coming up blank.  Point me in the right direction?

```
iex(27)> p = :erlang.list_to_port '#Port<10.10>'
    :erlang.list_to_port('#Port<10.10>')
iex(27)> p = :erlang.list_to_port '#Port<0.10.10>'
#Port<0.10>
```

josevalim: It will be unfortunately hard to test. For testing, you should test with major always being 0. For implementation, you can take a look at `pid/3`.

cas27: I see that `pid/3` is specifically using ` '<' ++ Integer.to_charlist(x) ++ `.  Any specific reason to use that over what im doing with `'#Port<#{major}.#{minor}>'`

josevalim: `pid/3` is likely more efficient. Your approach has two calls to a protocol (from the interpolations) to then build a list while `pid/3` just builds the list at once. But it is unlikely to matter for this code. The only reason I would stick to the same shape as `pid/3` is consistency, but nothing more than that.

PR #8443, opened 2018-11-30
Introduce Mix.target

josevalim:


josevalim:
Could it be a difference in defaults? We always default to :host. Can it be
an issue for Nerves?
-- 


ConnorRigby:
It's definitely an issue in Nerves, not Elixir. I didn't look much into it, but it seems strange to me that something configured for a different target is loaded. I didn't look at how the Nerves.Env module is getting that list of loaded packages however.

josevalim:
@ConnorRigby did you try master as well before this branch? Maybe the error is being caused by another commit in master and not this branch, so I want to rule that out. :)

lib/mix/lib/mix/dep/converger.ex
```diff
@@ -59,7 +59,7 @@ defmodule Mix.Dep.Converger do
     apps = Enum.map(main, & &1.app)
 
     lock_given? = !!lock
-    env = opts[:env]
+    env_target = {opts[:env], opts[:target]}
```
whatyouhide: By looking at this code, I'd go with `env_and_target`. It's slightly longer but `env_target` makes me thing of the target for the environment, plus the functions are called stuff like `split_by_env_and_target` so I think it would make sense.

lib/mix/lib/mix/project.ex
```diff
@@ -472,14 +472,25 @@ defmodule Mix.Project do
   end
 
   defp env_path(config) do
-    build = config[:build_path] || "_build"
+    dir = config[:build_path] || "_build"
+    subdir = build_target() <> build_per_environment(config)
+    Path.expand(dir <> "/" <> subdir)
+  end
+
+  defp build_target() do
```
whatyouhide: Sorry for the nitpick :P, but I think we go without parens here. :)

mobileoverlord:
I've tested this locally and things are going well for a demo project. Here is the example project I have generated and used to test this: https://github.com/mobileoverlord/elixir_18. I have also updated `nerves` and `nerves_bootstrap` to support this PR. You can see more about those at these locations:
[nerves](https://github.com/nerves-project/nerves/compare/elixir-18)
[nerves_bootstrap](https://github.com/nerves-project/nerves_bootstrap/compare/elixir-18)

I have been able to successfully build, burn, and boot a Raspberry Pi Zero using this. This looks good to me.

josevalim:
Thanks @mobileoverlord!

It would be nice to get rid of this too: https://github.com/mobileoverlord/elixir_18/blob/fb18425a6b221cee3d80d0b55bbe6c83d6bab337/mix.exs#L39

Otherwise you won't reap all of the benefit of this feature. But I guess you want to avoid listing all targets? We could allow a way to do `except: :host` but I am not quite sure how to support it. We could do `targets: [except: :host]`, but should we?

mobileoverlord:
@josevalim I updated the new project generator and example to have a single deps section. Here is what it looks like. https://github.com/mobileoverlord/elixir_18/blob/master/mix.exs#L35-L54

It works well.

josevalim:
Thanks @mobileoverlord! :heart:

josevalim:

PR #8444, opened 2018-12-01
Add Calendar.datetime type and allow it in DateTime functions

lau:
Instead of requiring a DateTime struct, allow a map containing at least
those fields. In the same way Calendar.naive_datetime is used in
NaiveDateTime (and Calendar.time/Time, Calendar.date/Date).

This way custom types can be used as long as they have the necessary fields required by the DateTime functions.

cc @josevalim

### DIFF
--- a/lib/elixir/lib/calendar/datetime.ex
+++ b/lib/elixir/lib/calendar/datetime.ex
@@ -527,10 +527,10 @@ defmodule DateTime do
       -17412508655
 
   """
-  @spec to_unix(t, System.time_unit()) :: integer
+  @spec to_unix(Calendar.datetime(), System.time_unit()) :: integer
   def to_unix(datetime, unit \\ :second)
 
-  def to_unix(%DateTime{utc_offset: utc_offset, std_offset: std_offset} = datetime, unit) do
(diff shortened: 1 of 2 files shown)


whatyouhide:
Thanks @lau! 💟

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -1096,8 +1097,8 @@ defmodule DateTime do
 
   """
   @doc since: "1.6.0"
-  @spec truncate(t(), :microsecond | :millisecond | :second) :: t()
-  def truncate(%DateTime{microsecond: microsecond} = datetime, precision) do
+  @spec truncate(Calendar.datetime(), :microsecond | :millisecond | :second) :: t()
```
michalmuskala: The return type is not valid anymore - it's not guaranteed to always return a `DateTime` struct. Similar situation exists in couple other functions.

lau: Good catch. I'll go through them and make sure that `DateTime`s will be returned even if a non-`DateTime` is passed.

lau: See https://github.com/elixir-lang/elixir/pull/8448

PR #8448, opened 2018-12-02
Make sure DateTime.truncate/2 returns DateTime when passed map.

lau:
With tests for Calendar.datetime maps being passed to DateTime functions.


michalmuskala:
@lau the same issue exists in the `convert/2` function, do you think it would make sense to fix it in the same or a separate PR?

lau:
@michalmuskala Thanks. Yes, I had first made a test that tested conversion to a different calendar, but the issue was when "converting to" the same calendar. I have added a test for conversion to same calendar and a fix for that.

michalmuskala:
Thank you @lau for working on this.

PR #8451, opened 2018-12-02
Update GitHub issue template with mailing list link

arturoherrero:
It could be useful to link directly to the mailing list in the GitHub issue template.

The same linking as we do in the Elixir website, see https://elixir-lang.org/development.html

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,15 +1,15 @@
 ### Precheck
 
 * Do not use the issues tracker for help or support (try Elixir Forum, Stack Overflow, IRC, etc.)
-* For proposing a new feature, please start a discussion on the Elixir Core mailing list
+* For proposing a new feature, please start a discussion on the Elixir Core mailing list: https://groups.google.com/group/elixir-lang-core
 * For bugs, do a quick search and make sure the bug has not yet been reported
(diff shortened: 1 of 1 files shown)


ISSUE_TEMPLATE.md
```diff
@@ -1,7 +1,7 @@
 ### Precheck
 
 * Do not use the issues tracker for help or support (try Elixir Forum, Stack Overflow, IRC, etc.)
-* For proposing a new feature, please start a discussion on the Elixir Core mailing list
+* For proposing a new feature, please start a discussion on the [Elixir Core mailing list][1]
```
josevalim: Because this is not rendered as markdown, I recommend listing the link right after. Something like:

> on the Elixir Core mailing list: https://groups.google.com/group/elixir-lang-core

arturoherrero: It's rendered as reference-style link in Markdown: https://daringfireball.net/projects/markdown/syntax#link
We can also use the Markdown inline link or just write the URL as you proposed.

You can see that it's working in the GitHub rich diff:

screenshot 2018-12-02 at 20 46 30

josevalim:
I understand that. :) My point is that by default it is shown in raw format in the issues:

image

So I think we should keep it inline, as most people won't click preview when reading the template.

PR #8458, opened 2018-12-03
Add `Range.disjoint?/2`

fxn:
Wanted to use `Range.disjoint?/2` for [today's Advent of Code](https://adventofcode.com/2018/day/3), so here's a proposal.

Doctests seem to provide good coverage and didn't implement separate unit tests... is that fine? Otherwise please just tell me and I'll add them.

### DIFF
--- a/lib/elixir/lib/range.ex
+++ b/lib/elixir/lib/range.ex
@@ -61,6 +61,33 @@ defmodule Range do
             "got: #{inspect(first)}..#{inspect(last)}"
   end
 
+  @doc """
+  Checks if two ranges are disjoint.
+
+  ## Examples
+
+      iex> Range.disjoint?(1..5, 6..9)
+      true
+      iex> Range.disjoint?(5..1, 6..9)
+      true
+      iex> Range.disjoint?(1..5, 5..9)
+      false
+      iex> Range.disjoint?(1..5, 2..7)
+      false
+
+  """
+  @doc since: "1.8.0"
+  @spec disjoint?(t, t) :: boolean
+  def disjoint?(first1..last1, first2..last2) do
+    {first1, last1} = normalize(first1, last1)
+    {first2, last2} = normalize(first2, last2)
+    last2 < first1 or last1 < first2
+  end
+
+  @compile inline: [normalize: 2]
+  defp normalize(first, last) when first > last, do: {last, first}
+  defp normalize(first, last), do: {first, last}
+
   # TODO: Remove by 2.0
   @doc false
   @deprecated "Pattern match on first..last instead"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
   * [List] Add `List.myers_difference/3`
   * [Macro] Add `Macro.struct!/2` for proper struct resolution during compile time
   * [Map] Optimize and merge nested maps `put` and `merge` operations
+  * [Range] Add `Range.disjoint?/2`
   * [Registry] Allow associating a value on `:via` tuple
   * [String] Add `String.bag_distance/2`
 
--- a/lib/elixir/test/elixir/range_test.exs
+++ b/lib/elixir/test/elixir/range_test.exs
@@ -5,6 +5,29 @@ defmodule RangeTest do
 
   doctest Range
 
+  defp reverse(first..last) do
+    last..first
+  end
+
+  defp assert_disjoint(r1, r2) do
+    disjoint_assertions(r1, r2, true)
+  end
+
+  defp assert_overlap(r1, r2) do
+    disjoint_assertions(r1, r2, false)
+  end
+
+  defp disjoint_assertions(r1, r2, expected) do
+    # The caller should choose pairs of representative ranges, and we take care
+    # here of commuting them.
+    Enum.each([[r1, r2], [r2, r1]], fn [a, b] ->
+      assert Range.disjoint?(a, b) == expected
+      assert Range.disjoint?(reverse(a), b) == expected
+      assert Range.disjoint?(a, reverse(b)) == expected
+      assert Range.disjoint?(reverse(a), reverse(b)) == expected
+    end)
+  end
+
   test "precedence" do
     assert Enum.to_list(1..(3 + 2)) == [1, 2, 3, 4, 5]
     assert 1..3 |> Enum.to_list() == [1, 2, 3]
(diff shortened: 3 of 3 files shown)


lib/elixir/lib/range.ex
```diff
@@ -56,6 +56,41 @@ defmodule Range do
...
+      iex> Range.overlap?(1..1, 1..1)
+      true
+      iex> Range.overlap?(1..5, 6..9)
+      false
+      iex> Range.overlap?(9..7, 3..1)
+      false
+  """
+  @spec overlap?(t, t) :: boolean
+  def overlap?(first1..last1, first2..last2) when first1 <= last1 and first2 <= last2 do
+    !(last2 < first1 || last1 < first2)
```
josevalim:
```suggestion
    not(last2 < first1 or last1 < first2)
```

Since they are all booleans, we should prefer to use the strict boolean operators. :)

lib/elixir/lib/range.ex
```diff
@@ -56,6 +56,41 @@ defmodule Range do
...
+      true
+      iex> Range.overlap?(1..1, 1..1)
+      true
+      iex> Range.overlap?(1..5, 6..9)
+      false
+      iex> Range.overlap?(9..7, 3..1)
+      false
+  """
+  @spec overlap?(t, t) :: boolean
+  def overlap?(first1..last1, first2..last2) when first1 <= last1 and first2 <= last2 do
```
josevalim: Given `String.contains?`, should this be `overlaps?`? Or are we asking if "both ranges overlap"?

fxn: Good question.

To me `overlap?` sounds more natural, similar to `MapSet.disjoint?` in a way, in which both arguments have the same importance, and the function is commutative. I think of them as "do the ranges overlap?", "are the sets disjoint?", "do the sets intersect?".

But I guess both options are equally good, you have a better perspective of naming conventions in the API, whatever you prefer 😄.

lib/elixir/lib/range.ex
```diff
@@ -56,6 +56,41 @@ defmodule Range do
...
+      iex> Range.overlap?(7..1, 9..5)
+      true
+      iex> Range.overlap?(0..1, 1..2)
+      true
+      iex> Range.overlap?(1..1, 1..1)
+      true
+      iex> Range.overlap?(1..5, 6..9)
+      false
+      iex> Range.overlap?(9..7, 3..1)
+      false
```
fertapric: Minor nitpicks: I would add `@doc since: "1.8.0"` and include a new line here :)

fxn: Oh yes, pushed!

fxn:
Yeah, I have implemented `Range.disjoint?/2` too.

lib/elixir/lib/range.ex
josevalim: Good points about `overlap?` being commutative.

However, I am still a bit worried about the `overlap?` not being plural and that being a possible inconsistency. Now that you added `disjoint?`, it is clear we are using it as a noun, that's why we can keep it singular. So I propose to either:

1. Figure out a name that is mostly seen as a noun
2. Keep only disjoint for now

Thoughts?

fxn: I believe `overlap` is the idiomatic name for this, can't think of a good alternative. The variant that matches `disjoint?` would be `overlapping?` I believe ("these are disjoint intervals", "these are overlapping intervals"). But not sure I like how `Range.overlapping?(...)` reads.

Let's leave it at disjoint maybe?

michalmuskala:
@fxn what if we changed the implementation slightly to avoid the deconstruction and reconstruction of the ranges and the recursive calls? I think this is a bit clearer.

```elixir
def disjoint?(first1..last1, first2..last2) do
  {first1, last1} = normalize(first1, last1)
  {first2, last2} = normalize(first2, last2)
  last2 < first1 or last1 < first2
end

@compile inline: [normalize: 2]
defp normalize(first, last) when first > last, do: {last, first}
defp normalize(first, last), do: {first, last}
```

I also like the name `overlaps` better than `overlap` but given we're going with the set-theoretical name `disjoint` for the other operation, perhaps `intersecting?` would be better?

fxn:
@michalmuskala that's perfect, had written it that way at some point indeed, but wasn't sure if the approach with guards was more idiomatic, and whether building tuples inconditionally was a good trade-off (don't have intuitions for this).

Cool, going to write it that way and also going to remove `overlap?`, please shout if any of you'd like it otherwise!

fxn:
Ah! will wait for the removal to see what people think about `intersecting?`.

fertapric:
`MapSet` provides a `disjoint?/2` and an `intersection/2` functions, but not `intersecting?/2` nor `overlap?/2`. I would keep only `disjoint?/2` for now.

fxn:
@josevalim awesome, thanks for the explanation!

michalmuskala:
I suggested the tuples because I was pretty sure the compiler would be able to optimise the allocation away given the inlining. It seems though it can't (I'll make sure OTP 22 can 🙂). I still think the implementation is a bit clearer because we avoid recursive calls.

fxn:
@michalmuskala good! My main doubt was that probably the majority of ranges have `first <= last` and optimized for them. But I learned about this thanks to this discussion, and maybe that `normalize/2` could be used in other functions as well (present or future).

michalmuskala:
Some of the structs are public some are private. It's always documented. `Range` is public, but for example `MapSet` is private.

wojtekmach:
Something to consider: to achieve parity should there be a `Date.Range.disjoint?/2`? Although for time ranges, [Allen's Interval Algebra](https://en.wikipedia.org/wiki/Allen%27s_interval_algebra) is more comprehensive so perhaps we shouldn't add `Date.Range.disjoint?/2` after all.

lib/elixir/lib/range.ex
```diff
@@ -56,6 +56,39 @@ defmodule Range do
...
+      iex> Range.disjoint?(1..5, 9..6)
+      true
+      iex> Range.disjoint?(5..1, 9..6)
+      true
+      iex> Range.disjoint?(1..5, 5..9)
+      false
+      iex> Range.disjoint?(1..5, 3..7)
+      false
+      iex> Range.disjoint?(1..5, 1..4)
+      false
```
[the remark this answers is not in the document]
eksperimental: @fxn I feel the same way.  I guess we can add value to the tests, by using comments and creating different test groups within a describe block, that way we are not "just" duplicating code.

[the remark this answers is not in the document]
fxn: @eksperimental excellent, let me have a stab at it.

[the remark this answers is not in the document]
josevalim: Doctests are documentation and not tests. :) It is completely fine if they overlap.

The only place where we use doctests in lieu of tests is in the calandar suite, which I mostly regret.

fxn:
Pushed another revision.

In this one there is a smaller set of examples, a proper test suite, and have squashed previous work to have one clean commit.

WDYT?

michalmuskala:
Thank you! :heart:

PR #8462, opened 2018-12-04
Improve error message for missing dots in multiline iex doctest

### LINKED ISSUE #8461: Indentation level mismatch on iex doctest block
### Environment

Erlang/OTP 21 [erts-10.0.8] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [hipe] [dtrace]

Elixir 1.7.3 (compiled with Erlang/OTP 21)
* Operating system: 
Mac OSX

### Current behavior
When running the following:
```
defmodule HelloWorld do

  @doc ~S"""
  Multiply all the things by 2.

  ## Example
    iex> outputs = HelloWorld.hello([
     1,
     2,
     3,
    ])
  """
  def hello(inputs) do
    Enum.map(inputs, fn x -> x * 2 end)
  end

end

ExUnit.start()

defmodule HelloWorldTest do
  use ExUnit.Case, async: true
  doctest HelloWorld
end
```

The error message is:
```
** (ExUnit.DocTest.Error) lib/broken.ex:12: indentation level mismatch: "         1,", should have been 4 spaces
```
because the doctest should be using the `...>` syntax for continuing blocks of code.

### Expected behavior

Ideally the error message would be more helpful, telling us that doctest is expecting `...>` for indented code blocks.

ecly:
Closes #8461 

Perhaps the error message can be better than my initial suggestion.


josevalim:
Thanks @ecly !

In the scenario of the stream that failed, I didn't have a `>` either, so I think we shuold just improve the "catch-all" error message without changing the code.

So my suggestion is to change the error message here:

https://github.com/elixir-lang/elixir/blob/3ac54ce9be790b3034460229e78cf57bb53cebd8/lib/ex_unit/lib/ex_unit/doc_test.ex#L540

To this:

```
message: """
indentation level mismatch on doctest line: #{inspect(line)}

If you are planning to assert on the result of an `iex>` expression, \
make sure the result is indented at the beginning of `iex>`, which \
in this case is exactly #{n_spaces} spaces.

If instead you have a `iex>` expression that spans over multiple lines, \
please make sure that each line after the first one begins with `...>`.
"""
```

And then update the tests accordingly. What do you think?

ecly:
@josevalim that's a good point!
I like the catch-all idea. Certainly less error prone. Shouldn't be too inconvenient with the added text for the actual indentation mismatch cases.
I'll change the PR using your suggested message.

ecly:
Updated according to @josevalim's suggestion.

lib/ex_unit/lib/ex_unit/doc_test.ex
```diff
@@ -531,7 +531,16 @@ defmodule ExUnit.DocTest do
...
           module: module,
-          message: "indentation level mismatch: #{inspect(line)}, should have been #{n_spaces}"
+          message: """
+          indentation level mismatch on doctest line: #{inspect(line)}
+
+          If you are planning to assert on the result of an `iex>` expression, \
+          make sure the result is indented at the beginning of `iex>`, which \
+          in this case is exactly #{n_spaces} spaces.
+
+          If instead you have a `iex>` expression that spans over multiple lines, \
```
fertapric:
```suggestion
          If instead you have an `iex>` expression that spans over multiple lines, \
```

lib/ex_unit/lib/ex_unit/doc_test.ex
```diff
@@ -531,7 +531,16 @@ defmodule ExUnit.DocTest do
         raise Error,
           line: line_no,
           module: module,
-          message: "indentation level mismatch: #{inspect(line)}, should have been #{n_spaces}"
+          message: """
+          indentation level mismatch on doctest line: #{inspect(line)}
+
+          If you are planning to assert on the result of an `iex>` expression, \
+          make sure the result is indented at the beginning of `iex>`, which \
+          in this case is exactly #{n_spaces} spaces.
```
fertapric:
```suggestion
          in this case is exactly #{n_spaces}.
```

lib/ex_unit/test/ex_unit/doc_test_test.exs
```diff
@@ -640,8 +640,9 @@ defmodule ExUnit.DocTestTest do
   end
 
   test "fails in indentation mismatch" do
-    message =
-      ~r[test/ex_unit/doc_test_test\.exs:\d+: indentation level mismatch: "   iex> bar = 2", should have been 2 spaces]
```
fertapric: I would keep the regex:

`~r[test/ex_unit/doc_test_test\.exs:\d+: indentation level mismatch on doctest line: "   iex> bar = 2".*is exactly 2 spaces]s`

lib/ex_unit/lib/ex_unit/doc_test.ex
ecly: I find it less clear without ‘spaces’. And having the ‘spaces’ there certainly doesn’t seem to add confusion.

fertapric: @ecly The variable `n_spaces` contains the word "space" or "spaces". Right now is rendering "is exactly 2 spaces spaces"

ecly: @fertapric haha, sorry. That went past me on my phone. Cheers

lib/ex_unit/test/ex_unit/doc_test_test.exs
ecly: Inclined to agree. I’ll be at a computer in an hour, then I’ll tidy that up.

josevalim: Yeah, let's keep the regexes so we don't have to change tests on every small change. :)

PR #8464, opened 2018-12-05
Improve the documentation of Range

fxn:
Couple of proposals for the documentation of `Range`.

### DIFF
--- a/lib/elixir/lib/range.ex
+++ b/lib/elixir/lib/range.ex
@@ -2,7 +2,7 @@ defmodule Range do
   @moduledoc """
   Defines a range.
 
-  A range represents a discrete number of integers.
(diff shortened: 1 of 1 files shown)


michalmuskala:
Thank you! :heart:

lib/elixir/lib/range.ex
```diff
@@ -2,7 +2,7 @@ defmodule Range do
   @moduledoc """
   Defines a range.
 
-  A range represents a discrete number of integers.
+  A range represents a discrete number of consecutive integers.
```
eksperimental: This is not true for ranges with the same start and end such as `1..1`

fxn: Good catch, what do you think about #8473?

PR #8465, opened 2018-12-05
Remove `=>` operator from documentation

CrowdHailer:
I don't think this belongs in the Operators documentation.

- It is not a standalone operator, i.e. valid only in map syntax
- List of operators does not include `->`

```
iex(1)> quote do: a => b
** (SyntaxError) iex:1: syntax error before: '=>'
```

### DIFF
--- a/lib/elixir/pages/Operators.md
+++ b/lib/elixir/pages/Operators.md
@@ -22,7 +22,6 @@ Operator                                                                        
 `\|\|` `\|\|\|` `or`                                                                     | Left to right
 `&`                                                                                      | Unary
 `=`                                                                                      | Right to left
-`=>`                                                                                     | Right to left
 `\|`                                                                                     | Right to left
 `::`                                                                                     | Right to left
 `when`                                                                                   | Right to left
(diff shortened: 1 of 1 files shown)


josevalim:
Hi @CrowdHailer, good catch!

The downside of removing it is that precedence information is still important. For example, it is important to know how `|` and `=` work inside maps alongside the `=>` operator. Should we instead add a note that says "map-only" or something similar?

josevalim:
Ping!

CrowdHailer:
So it's a good questions. I hadn't considered precedence information being needed.

Probably adding a note is the right way forward. in which case we should also add `->` with the same kind of note

josevalim:
@CrowdHailer good point. Will you add the notes or should we go ahead? :)

CrowdHailer:
Probably quicker without the back and forth.

Also I don't know where `->` in order of precedence

josevalim:
@CrowdHailer as a "template", what do you think of the changes as is right now? Then I can add `->` in another commit.

CrowdHailer:
I'd say `valid syntax only within %{}`. available to me sounds like it could be implemented if you felt like it, but that's not the case.

josevalim:
Perfect. Let's get more feedback and then we can merge it.

josevalim:

eksperimental:
Beautiful! I will be sharing a PR improving this page as well

PR #8466, opened 2018-12-05
Add missing `@doc since: "1.8.0"` to new functions

wojtekmach:


wojtekmach:
btw, does `@doc since: ...` make sense for types? Neither IEx nor ExDoc currently handles it.

lib/elixir/lib/application.ex
```diff
@@ -314,6 +314,7 @@ defmodule Application do
   application environment. `new` is a keyword list with all new keys
   and their values. `removed` is a list with all removed keys.
   """
+  @doc since: "1.8.0"
```
josevalim: The callback was always there but not documented, so I think we can remove this one. :)

josevalim:

PR #8467, opened 2018-12-06
Replace calls to :lists Erlang module in files outside elixir/lib

eksperimental:
Related to #8425

### DIFF
--- a/lib/iex/lib/iex/autocomplete.ex
+++ b/lib/iex/lib/iex/autocomplete.ex
@@ -349,7 +349,8 @@ defmodule IEx.Autocomplete do
 
   defp match_modules(hint, root) do
     get_modules(root)
-    |> :lists.usort()
+    |> Enum.sort()
+    |> Enum.dedup()
     |> Enum.drop_while(&(not String.starts_with?(&1, hint)))
     |> Enum.take_while(&String.starts_with?(&1, hint))
   end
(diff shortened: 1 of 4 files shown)


josevalim:
@eksperimental it still has changes that are under lib/elixir. Can you please revert them? In the future something may cause us to compile Unicode right after bootstrap and then we would need to undo them.

(the one change in test can be kept though). :)

eksperimental:
sorry, it thought it was only lib/elixir/lib that we cared about. I will change it.

lib/iex/lib/iex/autocomplete.ex
```diff
@@ -349,7 +349,8 @@ defmodule IEx.Autocomplete do
 
   defp match_modules(hint, root) do
     get_modules(root)
-    |> :lists.usort()
+    |> Enum.sort()
+    |> Enum.uniq()
```
michalmuskala: You can use `Enum.sort() |> Enum.dedup()` as a slightly less expensive alternative to `uniq` on an already sorted data.

eksperimental: @michalmuskala great advice!
I am applying the same principle here https://github.com/elixir-lang/elixir/pull/8467/files#diff-0a5aeba395ef65ae5e0061fc5b4bbf16R163

josevalim: I have reverted this other one though because it is completely fine to call functions in `:lists` when they don't have an equivalent in our List module. ;)

josevalim:

PR #8468, opened 2018-12-06
Enum.reverse/2 returns an improper list when tail is one

eksperimental:


lib/elixir/test/elixir/enum_test.exs
```diff
@@ -1234,6 +1238,10 @@ defmodule EnumTest.Range do
     assert Enum.reverse(1..3, [4, 5, 6]) == [3, 2, 1, 4, 5, 6]
     assert Enum.reverse(-3..5, MapSet.new([-3, -2])) == [5, 4, 3, 2, 1, 0, -1, -2, -3, -3, -2]
     assert Enum.reverse(5..5, [5]) == [5, 5]
+
+    assert_raise ArgumentError, fn ->
+      Enum.reverse([1 | 2], [3])
+    end
```
eksperimental: I don't know if it is possible to solve since we rely on :lists.reverse/1,2. Passing an improper list to `Enum.reverse/1` raises `FunctionClauseError` and passing it to `Enum.reverse/2` returns `ArgumentErrror`

josevalim: That's totally fine. We don't guarantee compatibility of errors.

lib/elixir/lib/enum.ex
```diff
@@ -2030,6 +2030,8 @@ defmodule Enum do
   Reverses the elements in `enumerable`, appends the `tail`, and returns
   it as a list.
 
+  If `tail` is an improper-list, it returns an improper-list.
```
josevalim: This is the same as #8469. We expect an enumerable here and an improper list is not an enumerable. It works only for optimization purposes but it wouldn't work like if the language was typed, for example.

PR #8469, opened 2018-12-06
Emum.to_list/1 supports improper-lists

eksperimental:


josevalim:
Good catch!

I don't think this is an actual behaviour though because an improper list is not enumerable. It happens only because of optimization purposes. For example, if we add types in the future, then I wouldn't expect it to compile (at all). So I believe this should stay as is.

eksperimental:
Good to know this. I am working on a PR that adds a special-case improper lists in the `i` helper in IEx.
So we should remove Enumerable from the Implemented protocols list.
How about Collectable?

josevalim:
@eksperimental unfortunately the protocol mechanism cannot differenciate between lists and improper lists, because doing so would be `O(N)` depending on the list size. It is like a protocol may only work for non-empty maps but we do not have enough information to say so. It is something that is known only at runtime, so I am afraid there isn't much that can be done.

eksperimental:
well, we are traversing the lists to determine if they are ascii_printable or keyword lists. I had to implement the function `List.improper?/1`. I though it is OK to traverse the list once just for printing the right information in the IEx session, since we are doing the same approach for the other list subtypes

josevalim:
What I am saying is that the *protocol dispatch* cannot do that as that would be expensive. And I don't think we should special case improper lists and the Enumerable protocol in the IEx code. If someone else has the same issue in their protocol, they would have to live with it, as they can't do the same. So we shouldn't either. The same rules should apply to everyone. :)

josevalim:
Perfect, I will close this one for now and I am looking forward to the next PR!

PR #8471, opened 2018-12-06
IEx helpers: port and reference should be in lowercase

eksperimental:


josevalim:
Why should they be in lowercase? Not that I disagree, just trying to understand the rationale. :)

josevalim:
So we keep it as "PID" above because it is an abbreviation?

eksperimental:
> So we keep it as "PID" above because it is an abbreviation?
yes, plus it is the standard across the code base. `pid` is only used for arguments.

josevalim:

PR #8472, opened 2018-12-06
IEx helper "i" lists Implemented Protocols sorted alphabetically

eksperimental:
Before:

```iex
iex(1)> i [1 | 2]
Term
  [1 | 2]
Data type
  List
Reference modules
  List
Implemented protocols
  IEx.Info, String.Chars, Enumerable, Inspect, List.Chars, Collectable
```

Now:

```iex
iex(1)> i [1 | 2]
Term
  [1 | 2]
Data type
  List
Reference modules
  List
Implemented protocols
  Collectable, Enumerable, IEx.Info, Inspect, List.Chars, String.Chars
```

### DIFF
--- a/lib/iex/lib/iex/helpers.ex
+++ b/lib/iex/lib/iex/helpers.ex
@@ -493,6 +493,7 @@ defmodule IEx.Helpers do
     |> Protocol.extract_protocols()
     |> Enum.uniq()
     |> Enum.reject(fn protocol -> is_nil(protocol.impl_for(term)) end)
+    |> Enum.sort()
     |> Enum.map_join(", ", &inspect/1)
   end
 


lib/iex/lib/iex/helpers.ex
```diff
@@ -491,6 +491,7 @@ defmodule IEx.Helpers do
   defp all_implemented_protocols_for_term(term) do
     :code.get_path()
     |> Protocol.extract_protocols()
+    |> Enum.sort()
     |> Enum.uniq()
```
josevalim: We can use `:lists.usort` here too, which both sorts and removes duplicates. If we want to keep it this way, then I think the fastest thing to do here is to sort after we reject the protocols, because that would be less data to sort. :)

eksperimental: > We can use `:lists.usort` here too, which both sorts and removes duplicates
I am trying to get rid of :lists calls as much possible as you can see in my other PR

josevalim: `:get_path` is misleading. It actually returns a list of paths.

josevalim:

PR #8473, opened 2018-12-06
Cover degenerate case in the description of ranges

fxn:
Covers degenerate case `a..a`.

### DIFF
--- a/lib/elixir/lib/range.ex
+++ b/lib/elixir/lib/range.ex
@@ -2,7 +2,8 @@ defmodule Range do
   @moduledoc """
   Defines a range.
 
-  A range represents a discrete number of consecutive integers.
+  A range represents a sequence of one or many,
+  ascending or descending, consecutive integers.
 
   Ranges can be either increasing (`first <= last`) or
   decreasing (`first > last`). Ranges are also always
@@ -34,7 +35,7 @@ defmodule Range do
       false
       iex> Enum.member?(range, 8)
       true
-      
+
   Such function calls are efficient memory-wise no matter the
   size of the range. The implementation of the `Enumerable`
   protocol uses logic based solely on the endpoints and does


fxn:
Trying to find a wording here that covers the degenerate case but keeps a casual style.

lib/elixir/lib/range.ex
```diff
@@ -2,7 +2,8 @@ defmodule Range do
   @moduledoc """
   Defines a range.
 
-  A range represents a discrete number of consecutive integers.
+  A range represents a series of one or several consecutive
+  integers.
```
eksperimental: "a series of one integer" doesn't sound right to me.

I like how it was before your last edit.
"A range represents an integer or a series of consecutive integers."

Since we are not in the summary, do you think we could elaborate on how an integer could be represeted, or we could use a comment and explain in the exàmples section

eksperimental: How about something like this?

>> A range represents a series of consecutive integers or just one integer.

fxn: The problem I had with that first one, is that a `a..a` is not really an integer, but a singleton, a set of one element, a series of one element. I believe that if we say "integer" we are still not quite there.

josevalim: I don't think saying it represents an integer would be good. My suggestion would be:

> A range represents a discrete number of one or many consecutive integers.

Or if we want to add more details:

> A range represents a discrete number of one or many, ascending or descending, consecutive integers.

There is some ambiguity here if the "consecutive" word applies to both "one" and "many" but that can be ruled out by logic.

fxn: A range would be a closed interval in **𝕫**.

eksperimental: > > A range represents a discrete number of one or many, ascending or descending, consecutive integers.

Actually, dependeding on how you read it, the ambiguity dissapears, becuase consecutive may only apply to the second half of the statement. So I would go with this one

fxn: I would like to revise "discrete number" since we are on it.

To me "discrete number" doesn't have a lot of sense. Yoy may have a finite or infinite numer of integers, "discrete" would be more something like "non-continuum". But talking sets of integers discrete is kind of redundant.

That is why I chose "series", because "series" or "sequence" is a concept that captures well what's a range, and also is informal enough.

fxn: Pushed that one.

josevalim:

PR #8475, opened 2018-12-06
Change wording from "change" to "rebind"

inoas:
not?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ The language has been improved semantically and includes new types and APIs. Let
 
 ### Deprecation of imperative assignment
 
-Elixir will now warn if constructs like `if`, `case` and friends assign to a variable that is accessed in an outer scope. As an example, imagine a function called `format` that receives a message and some options and it must return a path alongside the message:
+Elixir will now warn if constructs like `if`, `case` and friends rebind a variable that is accessed in an outer scope. As an example, imagine a function called `format` that receives a message and some options and it must return a path alongside the message:
 
 ```elixir
 def format(message, opts) do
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -8,7 +8,7 @@ The language has been improved semantically and includes new types and APIs. Let
 
 ### Deprecation of imperative assignment
 
-Elixir will now warn if constructs like `if`, `case` and friends assign to a variable that is accessed in an outer scope. As an example, imagine a function called `format` that receives a message and some options and it must return a path alongside the message:
+Elixir will now warn if constructs like `if`, `case` and friends rebind a variable that is accessed in an outer scope. As an example, imagine a function called `format` that receives a message and some options and it must return a path alongside the message:
```
josevalim: This is `assign` because it will also warn even if the variable did not previously exist.

CHANGELOG.md
```diff
@@ -23,7 +23,7 @@ def format(message, opts) do
 end
 ```
 
-The `if` block above is implicitly changing the value in `message`. Now imagine we want to move the `if` block to its own function to clean up the implementation:
+The `if` block above is implicitly rebinding the outer scope variable `message` to a new value. Now imagine we want to move the `if` block to its own function to clean up the implementation:
```
josevalim: Yes, this is better. :)

CHANGELOG.md
```diff
@@ -8,7 +8,7 @@ The language has been improved semantically and includes new types and APIs. Let
 
 ### Deprecation of imperative assignment
 
-Elixir will now warn if constructs like `if`, `case` and friends assign to a variable that is accessed in an outer scope. As an example, imagine a function called `format` that receives a message and some options and it must return a path alongside the message:
+Elixir will now warn if constructs like `if`, `case` and friends assign a value to a variable that is accessed in an outer scope. As an example, imagine a function called `format` that receives a message and some options and it must return a path alongside the message:
```
josevalim: It is a bit redundant though, no? We can always only assign a value to a variable, nothing else. :)

CHANGELOG.md
inoas: ~~to a new value~~ ?

josevalim:

PR #8476, opened 2018-12-06
Add List.product function

SofaKing18:
Hello :)
I found, that there are not analog of Ruby's Array#product (https://apidock.com/ruby/Array/product), so I've decided to fix this mistake :)

I guess that this is a useful function, and hope, you accept PR

### DIFF
--- a/lib/elixir/lib/list.ex
+++ b/lib/elixir/lib/list.ex
@@ -920,6 +920,31 @@ defmodule List do
   end
 
   @doc """
+  Returns an list of all combinations of elements from given lists.
+
+  ## Examples
+
+      iex> List.product([1, 2, 3], ["A"])
+      [[[1, "A"]], [[2, "A"]], [[3, "A"]]]
+
+      iex> List.product([1, 2, 3], ["A"], fn a, b -> {a, b} end)
+      [[{1, "A"}], [{2, "A"}], [{3, "A"}]]
+  """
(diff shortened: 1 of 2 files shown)


josevalim:
Hi @SofaKing18! As your code shows, this can be done quite simply with comprehensions, so that's what we would recommend folks to do. :+1: Thanks!

SofaKing18:
Honestly, `List.wrap` can be done simply too, but still, it is in the STD

There is the Cartesian product in math's world (https://en.wikipedia.org/wiki/Cartesian_product), and there is no any implementation in elixir STD.

MapSet or List should have this

michalmuskala:
You can get Cartesian product of any two enumerables using `for`:

```elixir
for x1 <- enumerable2, x2 <- enumerable2, do: {x1, x2}
```

josevalim:
> Honestly, List.wrap can be done simply too, but still, it is in the STD

List.wrap is used 47 times in the Elixir code base. So although it is straight-forward, it is common enough to warrant its inclusion. `product` is simply, not really uncommon, and it can be done with a comprehension, which is very expressive and also provides other features while achieving the same result.

eksperimental:
I don't see any advantage over `for`, on the contrary you are missing the niceties of `for`.
But if you still want to implement such a function, you should go with `Enum.product/`, not with `List`

PR #8478, opened 2018-12-06
Remove references to Erlang types

devonestes:
I think it's nicer to link to Elixir types instead of Erlang types where
we can, and these are both places where we definitely can.

### DIFF
--- a/lib/elixir/lib/io.ex
+++ b/lib/elixir/lib/io.ex
@@ -36,7 +36,7 @@ defmodule IO do
 
   @type device :: atom | pid
   @type nodata :: {:error, term} | :eof
-  @type chardata() :: :unicode.chardata()
+  @type chardata :: String.t() | maybe_improper_list(char | chardata, String.t() | [])
 
   defmacrop is_iodata(data) do
     quote do
--- a/lib/elixir/lib/path.ex
+++ b/lib/elixir/lib/path.ex
@@ -12,7 +12,7 @@ defmodule Path do
   that require it (like `wildcard/2` and `expand/1`).
   """
 
-  @type t :: :unicode.chardata()
+  @type t :: IO.chardata()
 
   @doc """
   Converts the given path to an absolute one. Unlike


lib/elixir/lib/io.ex
```diff
@@ -36,7 +36,7 @@ defmodule IO do
 
   @type device :: atom | pid
   @type nodata :: {:error, term} | :eof
-  @type chardata() :: :unicode.chardata()
+  @type chardata :: charlist | binary
```
josevalim: Those are not equivalent. A chardata can contain a list of binaries, for example. :)

lib/elixir/lib/path.ex
```diff
@@ -12,7 +12,7 @@ defmodule Path do
   that require it (like `wildcard/2` and `expand/1`).
   """
 
-  @type t :: :unicode.chardata()
+  @type t :: IO.chardata
```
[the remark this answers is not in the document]
whatyouhide: But it seems like it's not formatted :P

lib/elixir/lib/io.ex
devonestes: @josevalim Oops - I got a little confused reading the Erlang docs because they have a `unicode.charlist` type!

lib/elixir/lib/io.ex
```diff
@@ -36,7 +36,7 @@ defmodule IO do
 
   @type device :: atom | pid
   @type nodata :: {:error, term} | :eof
-  @type chardata() :: :unicode.chardata()
+  @type chardata :: binary | maybe_improper_list(char | chardata, binary | [])
```
fertapric:
```suggestion
  @type chardata :: String.t() | maybe_improper_list(char | chardata, String.t() | [])
```

What do you think about using `String.t()` instead of `binary` here to represent Erlang's `unicode_binary()`?

josevalim:

PR #8480, opened 2018-12-07
Make the type opaque to deal with dialyzer

### LINKED ISSUE #8463: Opaque MapSet type causes Dialyzer error
### Environment

* Elixir & Erlang/OTP versions (elixir --version): Elixir 1.7.4 (compiled with Erlang/OTP 21)
* Operating system: macOS 10.13.6

### Current behavior

```
def public_field?(field, %MapSet{} = public_fields), do: MapSet.member?(public_fields, field)
```

gives Dialyzer error `The call 'Elixir.MapSet':'member?'(_public_fields@1::#{'__struct__':='Elixir.MapSet', _=>_},_field@1::any()) does not have an opaque term of type 'Elixir.MapSet':t(_) as 1st argument`

### Expected behavior

Dialyzer not to give any error.

Similar to https://github.com/elixir-ecto/ecto/issues/2693, fixed by making the type not opaque.

treble37:
Closes #8463

### DIFF
--- a/lib/elixir/lib/map_set.ex
+++ b/lib/elixir/lib/map_set.ex
@@ -38,7 +38,11 @@ defmodule MapSet do
 
   @type value :: term
 
-  @opaque t(value) :: %__MODULE__{map: %{optional(value) => []}}
+  @opaque internal_data :: %{optional(value) => []}
+
+  @typep internal_key :: :map
+  @typep internal_value(value) :: %{optional(value) => []}
+  @type t(value) :: %{required(internal_key) => internal_value(value), __struct__: __MODULE__}
   @type t :: t(term)
 
   # TODO: Remove version key on Elixir 2.0
--- a/lib/elixir/lib/map_set.ex
+++ b/lib/elixir/lib/map_set.ex
@@ -38,8 +38,6 @@ defmodule MapSet do
 
   @type value :: term
 
-  @opaque internal_data :: %{optional(value) => []}
-
   @typep internal_key :: :map
   @typep internal_value(value) :: %{optional(value) => []}
   @type t(value) :: %{required(internal_key) => internal_value(value), __struct__: __MODULE__}


josevalim:
Thanks @treble37 for the PR!

However, the issue is the opposite. The type is already considered opaque, because it calls `t(term)` and sometimes we are accessing its fields, which violates the opaque boundary. The solution would be to make the types non-opaque but the struct contents opaque.

treble37:
thank you for the kind words @josevalim, I think this is what you meant?...if it is open to suggestions on renaming `internal_data` (or anything else)...

lib/elixir/lib/map_set.ex
```diff
@@ -38,7 +38,9 @@ defmodule MapSet do
 
   @type value :: term
 
-  @opaque t(value) :: %__MODULE__{map: %{optional(value) => []}}
+  @opaque internal_data :: %{optional(value) => []}
+
+  @type t(internal_data) :: %__MODULE__{map: internal_data}
```
josevalim: That's much better. But we also need to hide the key. So I think it would be this:

```elixir
@typep internal_key :: :map 
@typep internal_value(value) :: %{optional(value) => []}
@type t(value) :: %__MODULE__{required(internal_key) => interval_value(value)}
```

josevalim: Actually, this looks so complex that I am thinking about doing what @michalmuskala suggests and just not support it. :(

treble37: Oh sure that makes sense @josevalim, I updated with your suggestion(s), but understand if we want to take a different direction. Look forward to hearing what you decide.

josevalim: Thank you @treble37! Your solution is correct :heart: but it indeed looks very complex for reasons beyond our (yours and mine) control. We will discuss this a bit and follow up accordingly.

treble37: 👍 thank you

josevalim:
For now, we will won't merge this because if we have to add this to all of our structs, then it becomes unmanageable. This is really something we need to make more accessible at a base level, maybe within typespecs/dialyzer itself. Thanks @treble37!

Fire-Dragon-DoL:
@josevalim I stumbled upon the same issue recently. Would it make sense for now **not** to suggest in the docs to pattern match against `%MapSet{}`, dialyxir is already checking with `@spec`, but would prevent always getting errors. Ideally we'd have a new `@opaquedata` or something along those lines.

PR #8482, opened 2018-12-07
Add style remark in String.next_codepoint/1

fxn:
I believe this recommendation might be good from some chats with @josevalim this week, but I am not 100% sure.

Let me optimistically propose it. This week in AoC I used this function and would have benefited from this remark when reading the documentation of `String.next_codepoint/1`. And if it is not good, I'll learn something.

Regarding the code fences in the examples, I have seen docs don't generally use them, but seemed appropriate in this case. Of course, let me know if you'd prefer a different doctestable example.

### DIFF
--- a/lib/elixir/lib/string.ex
+++ b/lib/elixir/lib/string.ex
@@ -1504,6 +1504,23 @@ defmodule String do
       iex> String.next_codepoint("olá")
       {"o", "lá"}
 
+  ## Comparison with binary pattern matching
+
+  Binary pattern matching provides a similar way to decompose
+  a string:
+
+      iex> <<codepoint::utf8, rest::binary>> = "Elixir"
+      "Elixir"
+      iex> codepoint
+      69
+      iex> rest
+      "lixir"
+
+  though not entirely equivalent because `codepoint` comes as
+  an integer, and the pattern won't match invalid UTF-8.
+
+  Binary pattern matching, however, is simpler and more efficient,
+  so pick the option that better suits your use case.
   """
   @compile {:inline, next_codepoint: 1}
   @spec next_codepoint(t) :: {codepoint, t} | nil


lib/elixir/lib/string.ex
```diff
@@ -1504,6 +1504,24 @@ defmodule String do
...
       {"o", "lá"}
 
+  Note, however, that getting the next codepoint via pattern
+  matching is generally preferred. For example
+
+  ```elixir
+  def process(<<cp::utf8, rest::binary>>) do
+    # ...
+  end
+  ```
```
michalmuskala: A style nitpick: we use the 4-space indentation for code blocks instead of the fences.

michalmuskala: Oh, I just read the end of your comment. Yeah, I think using the indentation would be better. We have those more complex examples indented like that in other parts of the codebase as well.

fxn: Aaah, I did a project search and found [this occurrence](https://github.com/elixir-lang/elixir/blob/master/lib/mix/lib/mix/tasks/new.ex#L243-L248).

Let me revise it.

fxn: Up!

lib/elixir/lib/string.ex
```diff
@@ -1504,6 +1504,20 @@ defmodule String do
       iex> String.next_codepoint("olá")
       {"o", "lá"}
 
+  Note, however, that getting the next codepoint via pattern
+  matching is generally preferred. For example
```
josevalim: I like this but I am not sure if we should say "preferred". Maybe we should just say it is possible? What does everyone think?

fxn: I wanted to say something like "preferred if possible", "preferred if available"... something like that, but didn't find the right words (for my taste) and left it short.

fxn: Although... it is always possible now that I think, since that is what the function does really.

The use case I had in mind was the one in the examples mainly.

eksperimental: "can be simplified"

eksperimental: "can be simplified"

michalmuskala: @fxn to be honest, I'm not entirely sure. I suspect the goal was to offer similar & complete API for both codepoints & graphemes.

fxn: Cool!

What about a remark that says something in the line of

> This function is provided for symmetry with the API for graphemes, but doing the same with pattern matching is simpler and more performant. For example... < function definitions in both styles >

?

josevalim: I am not sure if that is relevant for users. Let's go with the simplest change here and just mention it can be implemented more efficiently with pattern matching too. :)

fxn: In my view it is relevant because if you put yourself in the shoes of a user reading:

> Function _F_ does this, but the same can be done more efficiently this way.

and the presented alternative it is not only more performant, but also looks simpler, and idiomatic, the user will inevitably wonder what is the point of the function, and more importantly _when_ should they use this apparently "inferior" option.

josevalim: Alright, I am convinced. :) Let's ship what you proposed!

fertapric: However, both examples are not "exactly" the same: in the pattern matching version the value assigned to the `cp` variable is an integer codepoint, and in the `String.next_codepoint/1` version is a string codepoint. 

Maybe we should mention that? or update the pattern matching example to turn the integer codepoint into a string one?

fxn: @fertapric I'll take that into account. Let me come with a reworded patch.

fxn:
I have pushed a revised patch.

I think there is conflicting vocabulary that difficults a bit being precise: As far as I know a _code point_ in Unicode jargon is a number. The entity that code maps to is a _character_, and one particular rendition a _glyph_. Characters may _combine_. Have double checked with [_Unicode Explained_](http://shop.oreilly.com/product/9780596101213.do), that I have here.

The documentation of `String`, however, refers to characters as "codepoints". In Elixir codepoints are binaries. To avoid confusion, I used `code` in the pattern matching examples, but if you can think of a better name please just tell me and I'll change it.

fxn:
I have realized that with the new approach, the point is driven home by the IEx session and we no longer need the precious space taken by the function definitions.

The new version, more brief, does not need two variables number/binary, so it just uses `codepoint` in line with the existing terminology.

Have rebased too.

josevalim:

PR #8485, opened 2018-12-08
Add ExUnit.Iex

Olshansk:
A helper that can be used to run test and iterate on code from within a shell.

I can add proper documentation if this is something you think is worthwhile adding.


josevalim:
In this case, I would suggest for folks to rather run it from IEX using mix:

    Mix.Task.rerun "test", ["test/foo.exs:6"]

This can be encapsulated in different ways in your app, so we likely don't need to add this functionality to Elixir itself. :) Thanks!

Olshansk:
Makes sense, thanks @josevalim!

PR #8486, opened 2018-12-08
[WIP] IEx: Special case "i" helper for improper lists

eksperimental:
Related discussion: https://github.com/elixir-lang/elixir/pull/8469#issuecomment-444784727

I had to create a List.improper?/1 in the same fashion of Keyword.keyword? and List.ascii_printable?

Please do not squash these two commits when merging.

### DIFF
--- a/lib/iex/lib/iex/info.ex
+++ b/lib/iex/lib/iex/info.ex
@@ -122,6 +122,7 @@ defimpl IEx.Info, for: List do
         list == [] -> info_list(list)
         List.ascii_printable?(list) -> info_charlist(list)
         Keyword.keyword?(list) -> info_kw_list(list)
+        List.improper?(list) -> info_improper_list(list)
         true -> info_list(list)
       end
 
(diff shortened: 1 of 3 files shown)


lib/iex/lib/iex/info.ex
```diff
@@ -153,6 +154,20 @@ defimpl IEx.Info, for: List do
     [{"Description", description}, {"Reference modules", "Keyword, List"}]
   end
 
+  defp info_improper_list(_improper_list) do
+    description = """
+    This is what is referred to as an "improper list". An improper list is a
+    list which the last element in its internal representation is not a list.
+    For example: [1, 2, 3] is a proper list, as opposed to [1, 2 | 3] which is
+    an improper one.
+    """
```
eksperimental: Feedback is welcome here, as there is a lot of room for improvement.

eksperimental: I initially took that path, but then, I realized that `[1, 2 | [3]]` is a proper list

josevalim: @eksperimental that's why I included "last tail", because that list has another list so the last tail requires further traversing. We could make the definition recursive: "a proper list is a list where its tail is a non-empty proper list or the empty list" but I think that is less helpful. :)

eksperimental: I have included your version

eksperimental:
Please have a look again

michalmuskala:
Thank you! :heart:

PR #8487, opened 2018-12-08
Check for function arity in guards for aggregates with empty handling

josevalim:
In the past, we had decided to not use is_function/1 or
is_function/2 in guards in Enum because the underlying
BadFunctionError was clearer than a FunctionClauseError.
However, today this is no longer necessarily true, due
to exception blaming that is capable of diffing the clauses
in a FunctionClauseError.

Furthermore, not checking for the argument can be an issue
on arguments that are rarely invoked, such as the empty
clause default function passed to Enum.min/2 friends, which
makes some mistakes impossible to catch. For example, someone
may write:

    Enum.min(non_empty_list, &elem(&1, 1))

The user intent here was actually to call `Enum.min_by/2`,
but without the guard, everything appears to work and the
error can go undetected.

Therefore this commit adds function guards to the min and
max aggregate functions in order to provide more feedback.
Depending on how this goes, we can add more guards to the
remaining functions in the future.

### DIFF
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -1479,7 +1479,8 @@ defmodule Enum do
 
   """
   @spec max(t, (() -> empty_result)) :: element | empty_result when empty_result: any
-  def max(enumerable, empty_fallback \\ fn -> raise Enum.EmptyError end) do
+  def max(enumerable, empty_fallback \\ fn -> raise Enum.EmptyError end)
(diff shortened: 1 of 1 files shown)


josevalim:
See here as an example: https://medium.com/@JPYamamoto9/day-06-advent-of-code-2018-5ae3a60662c3

I also did the same mistake when live streaming but someone caught my back. :)

josevalim:

PR #8491, opened 2018-12-08
Revamp List.ascii_printable?/2

eksperimental:
- Simplify logic
- Add specs
- Validate specs with guards, raising on non-lists
- Add tests
- Correct documentation about improper lists
- Rename argument `counter` to `limit`

### DIFF
--- a/lib/elixir/lib/list.ex
+++ b/lib/elixir/lib/list.ex
@@ -449,7 +449,7 @@ defmodule List do
   end
 
   @doc """
-  Checks if a list is a charlist made only of printable ASCII characters.
+  Checks if `list` is a charlist made only of printable ASCII characters.
 
   A printable charlist in Elixir contains only ASCII characters.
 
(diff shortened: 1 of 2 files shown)


eksperimental:
Please let me know if the tests are missing an edge case

josevalim:
Can we please keep the implementation the same? The compiler can optimize
multiple clauses like it was before better. The current version will do
linear lookups.
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

lib/elixir/lib/list.ex
```diff
@@ -467,62 +467,44 @@ defmodule List do
...
+        when limit: :infinity | non_neg_integer
+  def ascii_printable?(list, limit \\ :infinity)
+      when is_list(list) and (limit == :infinity or (is_integer(limit) and limit >= 0)) do
+    ascii_printable_with_limit?(list, limit)
   end
 
-  def ascii_printable?([?\b | rest], counter) do
-    ascii_printable?(rest, decrement(counter))
-  end
+  @compile {:inline, ascii_printable_with_limit?: 2}
```
michalmuskala: inlining recursive functions usually is not very beneficial

michalmuskala: in general, it's highly specific from case to case, but overall, it generally doesn't provide huge benefits. You'd need to measure it

eksperimental: thank you

lib/elixir/lib/list.ex
```diff
@@ -467,62 +467,43 @@ defmodule List do
...
-  end
+  defp ascii_printable_with_limit?(_list, 0),
+    do: true
 
-  def ascii_printable?([?\e | rest], counter) do
-    ascii_printable?(rest, decrement(counter))
+  for control_char <- [?\a, ?\b, ?\e, ?\f, ?\n, ?\r, ?\t, ?\v] do
+    defp ascii_printable_with_limit?([unquote(control_char) | rest], limit),
+      do: ascii_printable_with_limit?(rest, decrement(limit))
   end
```
eksperimental: It is not compiling (it was because I didn't run `make clean`).
How can I call a list comprehension in Erlang, from Elixir? or any why to bypass this limitation?

josevalim: This is a bootstrap issue. We can't use comprehensions on the List module because we likely need the List module for comprehensions to work (or something else that depends on the list). We can't be "smart" here.

But as I said in another comment, let's please keep the implementation as is in this commit and only change the minimum necessary to add guards (i.e. only rename the def/calls, no change to variables or syntax).

eksperimental: done

lib/elixir/lib/list.ex
```diff
@@ -467,58 +467,66 @@ defmodule List do
       iex> List.ascii_printable?('abc' ++ [0], 2)
       true
 
-  Improper lists are not printable, even if made only of ascii characters:
+  Last element in an improper list renders them not printable, even if made only of ASCII characters:
```
josevalim: I am having trouble understanding the rewritten sentence. I will let others give their feedback too. :)

fertapric: Same here. I think the previous version was easier to understand. I would only change `ascii` -> `ASCII`

eksperimental: Is there a specific term to refer to the last element in an improper list?
What I am trying to say is that if we reach this one element within `limit`, this the the one that renders it non-printable

eksperimental: Note that I am correct the previous sentence, since it is not correct.
Improper lists are printable provided limit doesn't reach their last element

josevalim: It is the same as the other discussion. :) They are not printable. If you pass one, then it may work by accident, but it is not a guarantee our API provides.

lib/elixir/test/elixir/list_test.exs
```diff
@@ -257,4 +257,39 @@ defmodule ListTest do
...
+      assert List.ascii_printable?('abc' ++ [0], 2)
+    end
+
+    test "improper lists" do
+      refute List.ascii_printable?('abc' ++ ?d)
+      assert List.ascii_printable?('abc' ++ ?d, 3)
+      assert List.ascii_printable?([1, 2 | 3], 0)
+    end
+
+    test "function clause errors" do
```
josevalim: We generally don't test function clause errors.

eksperimental: should i spread those tests around the other test blocks? or remove the tests

josevalim:
Regardless of being the same case or not, we don’t document things that
work by accident. The spec should be clear, it only works with lists, we
are not required to check it, which is why we also don’t check for function
clause errors.
-- 


eksperimental:
I was reviewing the List module, and a lot of those functions break when passed an improper list. I will revert those changes. thank you

josevalim:
@michalmuskala i don't see any merge conflicts in the UI? :D

josevalim:
Oh, doh, I resolved them apparently.

michalmuskala:
Thank you! :heart:

PR #8492, opened 2018-12-08
Elaborate on which characters are considered printable

eksperimental:
It extends #8491

### DIFF
--- a/lib/elixir/lib/list.ex
+++ b/lib/elixir/lib/list.ex
@@ -451,10 +451,24 @@ defmodule List do
   @doc """
   Checks if a list is a charlist made only of printable ASCII characters.
 
-  A printable charlist in Elixir contains only ASCII characters.
-
   Takes an optional `limit` as a second argument. `ascii_printable?/2` only
   checks the printability of the list up to the `limit`.
+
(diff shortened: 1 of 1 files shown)


josevalim:
Please use the ~S sigil so we don’t have to double escape. Lists should
also be written with asterisks and be indented with two spaces (our
convention), thanks!
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

lib/elixir/lib/list.ex
```diff
@@ -448,14 +448,28 @@ defmodule List do
...
   Checks if a list is a charlist made only of printable ASCII characters.
 
-  A printable charlist in Elixir contains only ASCII characters.
-
   Takes an optional `limit` as a second argument. `ascii_printable?/2` only
   checks the printability of the list up to the `limit`.
 
+  A printable charlist in Elixir contains only the printable characters in the
+  standard seven-bit ASCII character encoding, which are characters ranging from
+  32 to 126 in decimal notation, plus  the following control characters:
```
fertapric:
```suggestion
  32 to 126 in decimal notation, plus the following control characters:
```

josevalim:

PR #8494, opened 2018-12-09
Stricter link check in IO.ANSI.Docs module

entertainyou:
This is a naive fix of formatting "`http_bar`", but will not fix "`http://elixir.org/foo_bar`".

Closes #8449

PR #8497, opened 2018-12-10
IO.ANSI.Docs: Make regex in escape_underlines_in_link/1 handle all schemes

eksperimental:
It follows the same rules as URI.parse/1
Related to #8494

### DIFF
--- a/lib/elixir/lib/io/ansi/docs.ex
+++ b/lib/elixir/lib/io/ansi/docs.ex
@@ -524,7 +524,8 @@ defmodule IO.ANSI.Docs do
   end
 
   defp escape_underlines_in_link(text) do
-    ~r{https?://\S*}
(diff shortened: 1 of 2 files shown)


lib/elixir/lib/io/ansi/docs.ex
```diff
@@ -524,7 +524,8 @@ defmodule IO.ANSI.Docs do
   end
 
   defp escape_underlines_in_link(text) do
-    ~r{https?://\S*}
+    # Regular expression adapted from https://tools.ietf.org/html/rfc3986#appendix-B
+    ~r{[a-z][a-z0-9\+\-\.]*://\S*}i
```
josevalim: Just a note, beca use +-. are actually reserved characters, except inside `[` and `]`, I would rather train the eye to always escape them, which is why I have added the escapes here. Removing the escape for `/` was a great call though.

josevalim:

PR #8501, opened 2018-12-11
Use System.os_time/1 in the Erlang compiler as well

michalmuskala:
Follow up on 499446737ea0008bbb1c69a3e2f953fc7e1fedae that already did it for the Elixir compiler.


michalmuskala:
I found that the xref compiler was also using `:calendar.universaltime`, so I changed it to `os_time` as well.

PR #8502, opened 2018-12-11
use System.os_time

schnittchen:
OS time must be used when comparing against file system timestamps.


josevalim:
The changes to compilers/elixir only works in master, so I have reverted it, and keep only the minimum patch, thanks!

josevalim:

schnittchen:
Thank you. I wasn't sure about the second commit.

PR #8505, opened 2018-12-11
Replace quotation marks

eksperimental:


lib/elixir/lib/supervisor.ex
```diff
@@ -97,10 +97,12 @@ defmodule Supervisor do
   Supervisors support different strategies; in the example above, we
   have chosen `:one_for_one`. Furthermore, each supervisor can have many
   workers and/or supervisors as children, with each one having its own
-  configuration (as outlined in the “Child specification” section).
+
+  configuration (as outlined in the "Child specification" section).
 
   The rest of this document will cover how child processes are started,
-  how they can be specified, different supervision strategies and more.
+  how they
```
fertapric: I think this was not intended.

eksperimental: oops.

PR #8506, opened 2018-12-11
Touch-ups to String.jaro_distance/2

eksperimental:


lib/elixir/lib/string.ex
```diff
@@ -2343,22 +2343,22 @@ defmodule String do
...
 
-  Returns a float value between 0 (equates to no similarity) and 1
+  Returns a float value between `0.0` (equates to no similarity) and `1.0`
   (is an exact match) representing [Jaro](https://en.wikipedia.org/wiki/Jaro–Winkler_distance)
   distance between `string1` and `string2`.
 
   The Jaro distance metric is designed and best suited for short
   strings such as person names. Elixir itself uses this function
-  to provide "did you mean?" functionality. For instance, when you
+  to provide the "did you mean?" functionality. For instance, when you
```
josevalim: I think it was fine as it was because we are not referring to any implementation of "did you mean?" in particular.

eksperimental: IMO that wouldn't matter. I think it doesn't sound right. We should ask an English native speaker.

OvermindDL1: As a native english speaker I'd choose based on:

* If "did you mean?" is title of functionality then put `the` before it.
* If "did you mean?" is a descriptor of functionality then don't put `the` before it.

josevalim: I meant it as the second one, which is why I skipped the "the". :)

OvermindDL1: The more I look at it the more 'correct' `the` looks...  I'm unsure...  English (as most latin or anglo-saxon languages) is a stupid language.  ^.^;

josevalim: Let's go with your gut feeling then.

eksperimental: Thank you @OvermindDL1

PR #8510, opened 2018-12-11
Add $callers tracking to Task

### LINKED ISSUE #7995: Track Task callers in process dictionary
Elixir tasks makes it really straight-forward for developers to execute chunks of code concurrently. This is usually done with `Task.async/1`:

```elixir
Task.async(fn ->
  ...
end) |> Task.await()
```

However, in practice, developers should prefer to use `Task.Supervisor.async`, as the supervision tree gives more visibility over spawned tasks:

```elixir
Task.Supervisor.async(MySup, fn ->
  ...
end) |> Task.await()
```

Unfortunately, moving to `Task.Supervisor` has one big downside: we lose the relationship between the process who started the task and the task itself. That's because in the first example, the parent of the task is the caller process, but in the second example, the parent is the supervisor.

This means instrumentation/monitoring/logging tools and even testing tools like Mox and the Ecto's SQL Sandbox are incapable of detecting the relationship between them.

I propose that we add a new field to the Task process dictionary, called `$callers`, that tracks exactly  the caller of the task. This will allow tools to properly build the relationship between processes and allow testing tools that rely on ownership mechanisms to propagate those effectively without requiring tests to run synchronously.

josevalim:
This allows us to track which process invoked the Task
besides its ancestors in the supervision tree.

Closes #7995

### DIFF
--- a/lib/elixir/lib/task/supervised.ex
+++ b/lib/elixir/lib/task/supervised.ex
@@ -2,20 +2,21 @@ defmodule Task.Supervised do
   @moduledoc false
   @ref_timeout 5000
 
-  def start(owner, fun) do
-    {:ok, :proc_lib.spawn(__MODULE__, :noreply, [owner, fun])}
+  def start(owner, callers, fun) do
+    {:ok, :proc_lib.spawn(__MODULE__, :noreply, [owner, callers, fun])}
   end
 
-  def start_link(owner, fun) do
-    {:ok, :proc_lib.spawn_link(__MODULE__, :noreply, [owner, fun])}
+  def start_link(owner, callers, fun) do
+    {:ok, :proc_lib.spawn_link(__MODULE__, :noreply, [owner, callers, fun])}
   end
 
-  def start_link(owner, monitor, fun) do
-    {:ok, :proc_lib.spawn_link(__MODULE__, :reply, [owner, monitor, fun])}
+  def start_link(owner, callers, monitor, fun) do
+    {:ok, :proc_lib.spawn_link(__MODULE__, :reply, [owner, callers, monitor, fun])}
   end
 
-  def reply({_, _, owner_pid} = owner, monitor, mfa) do
+  def reply({_, _, owner_pid} = owner, callers, monitor, mfa) do
     initial_call(mfa)
+    put_callers(callers)
 
     case monitor do
       :monitor ->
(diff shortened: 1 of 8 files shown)


lib/elixir/lib/task/supervised.ex
```diff
@@ -183,6 +193,13 @@ defmodule Task.Supervised do
     )
   end
 
+  defp get_callers do
+    case :erlang.get(:"$callers") do
+      [_ | _] = list -> [self() | list]
```
michalmuskala: Why add the task itself to the list?

josevalim: That's not the task, it is the caller of the task. :)

josevalim: Or to be precise, it is the caller of the task that will be created.

michalmuskala: Ah, makes sense. Sorry.

lib/logger/lib/logger/translator.ex
```diff
@@ -379,8 +380,14 @@ defmodule Logger.Translator do
           ["Process ", crash_name(pid, name), " terminating", format(kind, reason, stack)] ++
             [crash_info(min_level, extra ++ crashed, [?\n]), crash_linked(min_level, linked)]
 
-        crash_metadata = [crash_reason: exit_reason(kind, reason, stack)] ++ registered_name(name)
-        {:ok, msg, crash_metadata ++ extra ++ user_metadata}
+        extra =
+          if ancestors = crashed[:ancestors], do: [{:ancestors, ancestors} | extra], else: extra
```
fertapric:
```suggestion
          if ancestors = crashed[:"$ancestors"], do: [{:ancestors, ancestors} | extra], else: extra
```
?

fertapric: Nevermind. I just checked the format of the crash report.

josevalim:

greg-rychlewski:
@josevalim Sorry for resurrecting this really old PR. I had an idea for an extension of this and was curious about your opinion. 

I was trying to think of how to allow tests using the Ecto Sandbox with shared GenServers, Agents, etc... to run their tests concurrently. 

Would it be a terrible idea to utilize `$callers` for this purpose as well? I was thinking the `$callers` value of the GenServer/Agent can be set to the `$callers` value of the process that sent the message (+ add the calling pid as well) and then reset it before the response is returned

josevalim:
We considered this a lot but it can still break very easily. Maybe the GenServer changes in the future to process the message asynchronously and use GenServer.reply, queries on init/terminate, etc. So I think it would help in some cases but not enough to cover everything in practice. Plus if you are doing queries in the GenServer, one-to-one compared to the caller, it may be better to do it in the caller and send the results.

greg-rychlewski:
Ah I didn't think about those complications. Thanks for explaining.

PR #8512, opened 2018-12-12
Review of Operators (part 1)

eksperimental:
This is part one (of two hopefully), broken into several commits to ease the reviewing

### DIFF
--- a/lib/elixir/pages/Operators.md
+++ b/lib/elixir/pages/Operators.md
@@ -14,14 +14,15 @@ Operator                                                                        
 `*` `/`                                                                                  | Left to right
 `+` `-`                                                                                  | Left to right
 `++` `--` `..` `<>`                                                                      | Right to left
+`^^^`                                                                                    | Left to right
 `in` `not in`                                                                            | Left to right
 `\|>` `<<<` `>>>` `~>>` `<<~` `~>` `<~` `<~>` `<\|>`                                     | Left to right
 `<` `>` `<=` `>=`                                                                        | Left to right
 `==` `!=` `=~` `===` `!==`                                                               | Left to right
 `&&` `&&&` `and`                                                                         | Left to right
 `\|\|` `\|\|\|` `or`                                                                     | Left to right
-`&`                                                                                      | Unary
 `=`                                                                                      | Right to left
+`&`                                                                                      | Unary
 `=>` (valid syntax only inside `%{}`)                                                    | Right to left
 `\|`                                                                                     | Right to left
 `::`                                                                                     | Right to left
--- a/lib/elixir/pages/Operators.md
+++ b/lib/elixir/pages/Operators.md
@@ -42,7 +42,7 @@ Elixir provides the following built-in comparison operators:
   * [`<=`](`Kernel.<=/2`) - less than or equal
   * [`>=`](`Kernel.>=/2`) - greater than or equal
 
-The only difference between [`==`](`Kernel.==/2`) and [`===`](`Kernel.===/2`) is that [`===`](`Kernel.===/2`) is stricter when it comes to comparing integers and floats:
+The only difference between [`==`](`Kernel.==/2`) and [`===`](`Kernel.===/2`) is that [`===`](`Kernel.===/2`) is strict when it comes to comparing integers and floats:
 
 ```elixir
 iex> 1 == 1.0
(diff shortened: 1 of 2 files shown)


lib/elixir/pages/Operators.md
```diff
@@ -67,12 +69,20 @@ The reason we can compare different data types is pragmatism. Sorting algorithms
 number < atom < reference < function < port < pid < tuple < map < list < bitstring
 ```
 
-When comparing two numbers of different types (a number is either an integer or a float), a conversion to the type with greater precision will always occur, unless the comparison operator used is either `===` or `!==`. A float will be considered more precise than an integer, unless the float is greater/less than +/-9007199254740992.0, at which point all the significant figures of the float are to the left of the decimal point. This behavior exists so that the comparison of large numbers remains transitive.
+When comparing two numbers of different types (a number being either an integer
+or a float), a conversion to the type with greater precision will always occur,
+unless the comparison operator used is either [`===`](`===/2`) or [`!==`](`!===/2`).
+A float will be considered more precise than an integer, unless the float is
+greater/less than `+/-9007199254740992.0` respectively, at which point all the
```
eksperimental: backticks added to `+/-9007199254740992.0`
"respectively" added for clarity

lib/elixir/pages/Operators.md
```diff
@@ -67,12 +69,20 @@ The reason we can compare different data types is pragmatism. Sorting algorithms
...
+exists so that the comparison of large numbers remains transitive.
 
 The collection types are compared using the following rules:
 
-* Tuples are compared by size then element by element.
-* Maps are compared by size then by keys in ascending term order then by values in key order. In the specific case of maps' key ordering, integers are always considered to be less than floats.
+* Tuples are compared by size, then element by element.
+* Maps are compared by size, then by keys in ascending term order, then by values
+  in key order. In the specific case of maps' key ordering, integers are always
+  considered to be less than floats.
```
eksperimental: comas added before "then"

lib/elixir/pages/Operators.md
fertapric: Why the backticks?

lib/elixir/pages/Operators.md
```diff
@@ -67,12 +69,20 @@ The reason we can compare different data types is pragmatism. Sorting algorithms
 number < atom < reference < function < port < pid < tuple < map < list < bitstring
 ```
 
-When comparing two numbers of different types (a number is either an integer or a float), a conversion to the type with greater precision will always occur, unless the comparison operator used is either `===` or `!==`. A float will be considered more precise than an integer, unless the float is greater/less than +/-9007199254740992.0, at which point all the significant figures of the float are to the left of the decimal point. This behavior exists so that the comparison of large numbers remains transitive.
+When comparing two numbers of different types (a number being either an integer
```
josevalim: We don't break text in our guides line by line like this, so please revert this change and keep everything as a long line. I would honestly prefer to have it line by line but this is not the time to change it. :)

lib/elixir/pages/Operators.md
eksperimental: True. They shouldn't have been used in this case.

eksperimental:
suggestions implemented!

lib/elixir/pages/Operators.md
```diff
@@ -67,12 +68,12 @@ The reason we can compare different data types is pragmatism. Sorting algorithms
 number < atom < reference < function < port < pid < tuple < map < list < bitstring
 ```
 
-When comparing two numbers of different types (a number is either an integer or a float), a conversion to the type with greater precision will always occur, unless the comparison operator used is either `===` or `!==`. A float will be considered more precise than an integer, unless the float is greater/less than +/-9007199254740992.0, at which point all the significant figures of the float are to the left of the decimal point. This behavior exists so that the comparison of large numbers remains transitive.
+When comparing two numbers of different types (a number being either an integer or a float), a conversion to the type with greater precision will always occur, unless the comparison operator used is either [`===`](`===/2`) or [`!==`](`!===/2`). A float will be considered more precise than an integer, unless the float is greater/less than +/-9007199254740992.0 respectively, at which point all the significant figures of the float are to the left of the decimal point. This behavior exists so that the comparison of large numbers remains transitive.
```
fertapric:
```suggestion
When comparing two numbers of different types (a number being either an integer or a float), a conversion to the type with greater precision will always occur, unless the comparison operator used is either [`===`](`===/2`) or [`!==`](`!==/2`). A float will be considered more precise than an integer, unless the float is greater/less than +/-9007199254740992.0 respectively, at which point all the significant figures of the float are to the left of the decimal point. This behavior exists so that the comparison of large numbers remains transitive.
```

The link had an extra `=` (`` [`!==`](`!===/2`) ``)

fertapric: And maybe use `Kernel.` like the links above?

eksperimental: Im planing another PR removing all mentions to Kernel and Kernel.SpecialForms

eksperimental: link to `!==/2` has been fixed

josevalim:

Issue #8513, opened 2018-12-12
Provide more hints for UndefinedFunctionError in case there is a behaviour


michalmuskala:
If there's a behaviour declared on a module that expects a particular function and we get an `UndefinedFunctionError` on that function - the `blame` callback should add information about this.

The information about behaviours is available in the return of the `module_info/1` function and the information about callbacks of a particular behaviour is available in the return of the `behaviour_info/1` function.

dnsbty:
I would love to help with this one if no one has picked it up by this afternoon.

michalmuskala:
Please, go ahead 🙂

dnsbty:
Any ideas on what those hints should actually say? I think I've got some basic ideas on this so I'll try to get a first pass on a PR up in a couple minutes

josevalim:
I am really not sure what is the benefit here. We already warn it at
compile time, so why do we need an extra information on runtime? It can
even be confusing. What if we missing function is not in a module that I
own? So what is the use case here? :)
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

josevalim:
Closing in favor of #8517

josevalim:
Thanks for expanding @michalmuskala !

PR #8517, opened 2018-12-12
add hint for behaviours on UndefinedFunctionErrors

dnsbty:
This is the first pass on #8513. I would love any feedback you have on this.

### DIFF
--- a/lib/elixir/lib/exception.ex
+++ b/lib/elixir/lib/exception.ex
@@ -945,6 +945,10 @@ defmodule UndefinedFunctionError do
           behaviour -> ", but the behaviour #{inspect(behaviour)} expects it to be present"
         end
     end
+  rescue
+    # In case the module was removed while we are computing this
+    UndefinedFunctionError ->
+      []
   end
 
   defp expects_callback?(behaviour, function, arity) do
@@ -1018,17 +1022,9 @@ defmodule UndefinedFunctionError do
   end
 
   defp behaviours_for(module) do
-    attributes =
-      case function_exported?(module, :__info__, 1) do
-        true -> module.__info__(:attributes)
-        false -> module.module_info(:attributes)
-      end
-
-    Keyword.get(attributes, :behaviour, [])
-  rescue
-    # In case the module was removed while we are computing this
-    UndefinedFunctionError ->
-      []
+    :attributes
+    |> module.module_info()
+    |> Keyword.get(:behaviour, [])
   end
 
   defp exports_for(module) do
--- a/lib/elixir/lib/exception.ex
+++ b/lib/elixir/lib/exception.ex
@@ -924,12 +924,32 @@ defmodule UndefinedFunctionError do
 
   defp message(:"function not exported", module, function, arity) do
     formatted_fun = Exception.format_mfa(module, function, arity)
-    {"function #{formatted_fun} is undefined or private", true}
+    fun_message = "function #{formatted_fun} is undefined or private"
+    behaviour_hint = behaviour_hint(module, function, arity)
+    {fun_message <> behaviour_hint, true}
   end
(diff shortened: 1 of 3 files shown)


lib/elixir/lib/exception.ex
```diff
@@ -997,6 +1017,20 @@ defmodule UndefinedFunctionError do
     ["      * ", Code.Identifier.inspect_as_function(fun), ?/, Integer.to_string(arity), ?\n]
   end
 
+  defp behaviours_for(module) do
+    attributes =
+      case function_exported?(module, :__info__, 1) do
+        true -> module.__info__(:attributes)
+        false -> module.module_info(:attributes)
```
josevalim: We don't need the conditional, we can just call `module.module_info(:attributes)` that should work for both.

lib/elixir/lib/exception.ex
```diff
@@ -997,6 +1017,20 @@ defmodule UndefinedFunctionError do
...
+      end
+
+    Keyword.get(attributes, :behaviour, [])
+  rescue
```
josevalim: I would move this rescue to `behaviour_hint` since there are other operations there that may fail.

lib/elixir/lib/exception.ex
```diff
@@ -924,14 +924,34 @@ defmodule UndefinedFunctionError do
...
+        case Enum.find(behaviours, &expects_callback?(&1, function, arity)) do
+          nil -> ""
+          behaviour -> ", but the behaviour #{inspect(behaviour)} expects it to be present"
+        end
+    end
+  end
+
+  defp expects_callback?(behaviour, function, arity) do
+    callbacks = behaviour.behaviour_info(:callbacks)
+    Enum.member?(callbacks, {function, arity})
```
josevalim: We also need to make sure the callback is not in the optional callbacks. In a nutshell:

```elixir
pair = {function, arity}
pair in behaviour.behaviour_info(:callbacks) and pair not in behaviour.behaviour_info(:optional_callbacks)
```

josevalim: Good call, let's keep it as is then!

dnsbty:
The other question I have is around testing this. Should I define both a behaviour and its implementation inside a test?

lib/elixir/test/elixir/exception_test.exs
```diff
@@ -727,6 +727,17 @@ defmodule ExceptionTest do
 
       assert %UndefinedFunctionError{module: nil, function: :bar, arity: 0}
              |> message == "function nil.bar/0 is undefined"
+
+      defmodule(Behaviour, do: @callback(required() :: :ok))
```
josevalim: Let's define it as an optional callback too, so we don't get compilation warnings. :) I would also define a separate test, instead of changing an existing one.

dnsbty: That's a great idea. Done.

lib/elixir/test/elixir/record_test.exs
```diff
@@ -6,8 +6,6 @@ defmodule RecordTest do
   require Record
   doctest Record
 
-  import ExUnit.CaptureIO
```
dnsbty: I removed this because I was getting
```elixir
warning: unused import ExUnit.CaptureIO
  test/elixir/record_test.exs:9
```
when I ran the full test suite

michalmuskala:
Thank you! :heart:

dnsbty:
Thank you for creating a great issue @michalmuskala and for your guidance with getting the PR up to snuff @josevalim! I've been really intimidated by contributing to Elixir, and I felt like the two of you made it super simple to get this in and understand some of the internals better. I really appreciate it!

PR #8524, opened 2018-12-15
Improve specs for List.improper?/1

eksperimental:


michalmuskala:
Wouldn't dialyzer complain in here that it doesn't currently (so probably never will) support functions with overlapping domains?

josevalim:
@michalmuskala you are correct. I always forget about this detail. I think we should keep things as is then.

josevalim:
Ping.

eksperimental:
so basically Dialyzer will treat `nonempty_improper_list2` as a `list` and it will overlap. Is that what you are saying @michalmuskala ?

josevalim:
Yes. It will automatically become `list :: true | false` once it merges.

PR #8525, opened 2018-12-15
Make List.ascii_printable?/2 specs more specific

eksperimental:


josevalim:
I prefer as it was. :) Specs for us are mostly for documentation purposes and I don't think it becomes more helpful in the documentation department to break it into many clauses. Thanks!

PR #8526, opened 2018-12-15
Improve specs for List.wrap/1

eksperimental:


lib/elixir/lib/list.ex
```diff
@@ -410,9 +410,11 @@ defmodule List do
       []
 
   """
+  @spec wrap(improper_list) :: improper_list
+        when improper_list: nonempty_improper_list(any(), any())
+  @spec wrap(list) :: list
   @spec wrap(nil) :: []
-  @spec wrap(list) :: list when list: maybe_improper_list()
-  @spec wrap(term) :: nonempty_list(term) when term: any()
+  @spec wrap(term) :: [term] when term: any()
```
josevalim: Can you please explain the improvements? As far as I understand the previous version was more concise and also correct.

josevalim:
Ping.

lib/elixir/lib/list.ex
```diff
@@ -410,9 +410,11 @@ defmodule List do
       []
 
   """
+  @spec wrap(improper_list) :: improper_list
+        when improper_list: nonempty_improper_list(any(), any())
```
eksperimental: Here we are more precise than in the previous spec. Given an improper-list, it will return an improper list.

lib/elixir/lib/list.ex
```diff
@@ -410,9 +410,11 @@ defmodule List do
...
+  @spec wrap(list) :: list
```
eksperimental: Given a list, it will return a list

lib/elixir/lib/list.ex
eksperimental: This is the same as the previous spec line, but it mimics what wrap does, which is to wrap the `term` in list brackets

josevalim: So as @michalmuskala said in another comment, dialyzer does not support overlapping clauses, for lists, it will match both the `wrap(term)` clause  and the `wrap(list)` clause. Even worse, `improper_list` will match all three clauses. So we can't accept those changes.

eksperimental: do you have any documentation you can refer me to about these limitations?

josevalim: I am not sure if it is documented because the expectation that clauses in a spec work like function clauses is simply not true. It doesn't work like it tries one clause, then the other and then the other. I am not sure if the behaviour in this case is documented. It may either be unspecified or it just gets the most general form (the `term` clause, which would be wrong here).

josevalim:
Closing due to overlapping domains.

PR #8527, opened 2018-12-15
Add shortcut built-in type nonempty_improper_list/0

eksperimental:
Basic list types have built-in shortcuts:

  * `list(type)` has `list()` which is defined as `[any()]`
  * `nonempty_list(type)` has `nonempty_list()` defined as `nonempty_list(any())`
  * `maybe_improper_list(type1, type2)` has `maybe_improper_list()` defined as  `maybe_improper_list(any(), any())`
  * `nonempty_maybe_improper_list(type1, type2)` has `nonempty_maybe_improper_list()` defined as `nonempty_maybe_improper_list(any(), any())`

but
* `nonempty_improper_list(type1, type2)` doesn't have `nonempty_improper_list()` which should be  defined as `nonempty_improper_list(any(), any())`

This PR adds it.

### DIFF
--- a/lib/elixir/lib/code/typespec.ex
+++ b/lib/elixir/lib/code/typespec.ex
@@ -382,6 +382,10 @@ defmodule Code.Typespec do
     typespec_to_quoted({:type, line, :as_boolean, [arg]})
   end
 
+  defp remote_type(line, {:atom, _, :elixir}, {:atom, _, :nonempty_improper_list}, []) do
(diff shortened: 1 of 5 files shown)


josevalim:
In this case, I would recommend adding those to Erlang/OTP, as all other types are provided by them and we just delegate. :+1:

josevalim:
Thanks for the PR anyway!

PR #8528, opened 2018-12-15
Simplify custom links when linking to Kernel and SpecialForms

eksperimental:
Now that ExDoc automatically links to functions in
Kernel and Kernel.SpecialForms, when linking to them in custom links,
we can leave out the module name.


josevalim:

PR #8529, opened 2018-12-15
Format optional macrocallbacks in IEx.Helpers.b/1

fertapric:
Macrocallbacks are internally stored prefixing them with "MACRO-". However, that prefix should be removed when presenting the callback information to the user.

Before:

```elixir
@optional_callbacks ["MACRO-foo": 3]
```

After:

```elixir
@optional_callbacks [foo: 3]
```


michalmuskala:
Thank you! :heart:

PR #8531, opened 2018-12-16
Add :reduce to for comprehensions

josevalim:


OvermindDL1:
This looks super useful!

The syntax change in the body kind of rubs me in an inconsistent way, but the functionality boon and the inconsistent existing syntax still fits in well enough to be super useful.  :-)

michalmuskala:
We've discussed couple different ways to implement this and that one was the least awkward.

OvermindDL1:
See the TL;DR at the end if you just want to see the proposed style.

> We've discussed couple different ways to implement this and that one was the least awkward.

Yeah.  Just for documentation reasons I think I find everything currently about this awkward because:

1. The `for` expression is Variable-Arg, which is extremely inconsistent with the rest of Elixir.
2. The `reduce: ...` 'argument' implies it can be passed multiple times to `for` for some reason, and I'm wondering what would happen if it were, like would the body increase in the `acc0, acc1, acc2, ... -> ...` or so?  What if they are interspersed among other `... <- ...` or filtering calls?
3. Can't just get rid of the `reduce: ...` as it is because you need some 'base case' for the reduction.
4. The `->` construct forces requiring a block, so the common `, do: ...` usage will fail pretty hard if someone tries to do `for <<x <- "AbCabCABc">>, x in ?a..?z, reduce: %{}, do: acc -> Map.update(acc, <<x>>, 1, & &1 + 1)` as the `->` wraps *everything* that comes before as the 'lead argument' (you need parenthesis to fix it), which no doubt *will* cause some errors for people who try to adjust existing `for`'s (or even new ones), regardless of how well document it is.

The only main idea I really have that might help will only help on numbers 2 and 3, and that is if you default it to `nil` and allow people to handle cases inside it:
```elixir
      iex> for <<x <- "AbCabCABc">>, x in ?a..?z do
      ..>    nil -> %{}
      ...>   acc -> Map.update(acc, <<x>>, 1, & &1 + 1)
      ...> end
```
This is more consistent with cases like in `case` and heads, and it is more obvious where the initial value comes from, and it is trivially to optimize by testing for an explicit `nil` match.  An alternative is defaulting it to whatever type the `for` is already returning, so an `[]` for a normal `for`, a `""` for a binary for, and maybe just passing in straight whatever is passed into `into: ...` (which is another argument I find questionable as implies being able to be used multiple times like `<-` and filters, ditto with `unit: ...`).

However, 2 and 3 aren't even the biggest issues, but rather 1 and 4 are, and fixing 1 is not backwards compatible, and fixing 4 is troublesome, but if fixed then it would more easily open the way to 'reduce' over multiple things, perhaps something like:
```elixir
      iex> for\
      ...>   <<x <- "AbCabCABc">>,
      ...>   x in ?a..?z,
      ...>   reduce: something,
      ...>   do: Map.update(acc, <<x>>, 1, & &1 + 1)
```
Where `something` as just an initial example could be `acc` if it is defaulted as per prior section, so like:
```elixir
      iex> for\
      ...>   <<x <- "AbCabCABc">>,
      ...>   x in ?a..?z,
      ...>   reduce: acc,
      ...>   do: Map.update(acc, <<x>>, 1, & &1 + 1)
```
Or perhaps something could be like `acc <- %{}`:
```elixir
      iex> for\
      ...>   <<x <- "AbCabCABc">>,
      ...>   x in ?a..?z,
      ...>   reduce: acc <- %{},
      ...>   do: Map.update(acc, <<x>>, 1, & &1 + 1)
```
However, since `into: ...` already contains a 'base case' then why not just: specify the 'base reduce' case in that (defaulting to the normal type of the `for`), so if you want to reduce into a map then just do:
```elixir
      iex> for\
      ...>   <<x <- "AbCabCABc">>,
      ...>   x in ?a..?z,
      ...>   reduce: acc,
      ...>   into: %{},
      ...>   do: Map.update(acc, <<x>>, 1, & &1 + 1)
```
So you instead just keep using `into: ...` as what it is already for (the base case to inject into) and if you specify a `reduce: ...` then that argument becomes the body binding (thus consistent with the rest of the expressions), thus everything remains consistent with how it is currently inconsistent with the rest of the language.  :-)

TL;DR:  I prefer this version for consistency with the rest of the `for` style:
```elixir
      iex> for <<x <- "AbCabCABc">>, x in ?a..?z, reduce: acc, into: %{} do
      ...>   Map.update(acc, <<x>>, 1, & &1 + 1)
      ...> end
```

ggcampinho:
I agree that the new syntax is a bit confusing, especially because I would expect more something like `reduce: true` combined with the `into` option (for me, the reduce is just giving more flexibility to the into option).

About the change in syntax when using `reduce` is somehow strange, but it's also more powerful since you could match things on the left side that you can't do if you have just a binding. So, I prefer the first.

OvermindDL1:
> About the change in syntax when using reduce is somehow strange, but it's also more powerful since you could match things on the left side that you can't do if you have just a binding. So, I prefer the first.

You can already match by using a `case`, which them becomes unambiguous (you can even use it in a `, do: case acc do ... end`) as it still matches the prior syntax.

Personally I would prefer if `for` used matchers built in from the start (thus always have been mandating `do ... end`) but that doesn't fit with the current styles of usage.

What if someone uses the `acc -> ...` without `reduce: true` as well, you could give an error message but this could also be a legit use-case as well that is already covered by just always using `case ... do ... end`, plus which comprehension generator expression `... <- ...` would it bind to, or should it be multi-arg with one for each?  It's all just unambiguous and user-specified by just using `case ... do ... end`.

josevalim:
> The for expression is Variable-Arg, which is extremely inconsistent with the rest of Elixir.

Note `:into` already behaves like this although I agree, I think `for` is the only special form that behaves like that.

> The reduce: ... 'argument' implies it can be passed multiple times to for for some reason

That's not necessarily true. You can't pass multiple "do"s to `if`, `case`, etc. Most APIs will actually just pick the first of multiple options. Being a keyword does not imply multiple options are ok.

> The -> construct forces requiring a block, so the common , do: ... usage will fail pretty hard i

Right. We consider the failing hard a feature though, because you need to explicitly opt into it or not. Although I will leave the judgement of `->` being good or bad up to your tastes. :D

josevalim:
Btw, the other syntax we discussed at the time was:

```elixir
reduce for <<x <- "AbCabCABc">>, x in ?a..?z, into: %{} do
  acc -> Map.update(acc, <<x>>, 1, & &1 + 1)
end
```

lib/elixir/src/elixir_expand.erl
```diff
@@ -1045,6 +1079,12 @@ format_error({invalid_args, Construct}) ->
   io_lib:format("invalid arguments for \"~ts\"", [Construct]);
 format_error({for_invalid_uniq, Value}) ->
   io_lib:format(":uniq option for comprehensions only accepts a boolean, got: ~ts", ['Elixir.Macro':to_string(Value)]);
+format_error(for_duplicate_reduce_into_uniq) ->
+  "cannot use :reduce alongside :into/:uniq in comprehension";
+format_error(for_with_reduce_bad_args) ->
+  "when using :reduce with comprehensions, the do block must be written using acc -> expr clauses, where each clause expects the accumulator as a single argument";
+format_error(for_without_reduce_with_arg) ->
+  "the do block was written using acc -> expr clauses but there :reduce option was not given";
```
ericmj:
```suggestion
  "the do block was written using acc -> expr clauses but the :reduce option was not given";
```

ggcampinho:
I like the `as` readability. I don't know if it's too much, but it came to
my mind that we could use the `reduce` option  with the block like the
`else` syntax on the `with`, it's readable and it looks similar to the
existing syntax.

lib/elixir/src/elixir_expand.erl
```diff
@@ -272,6 +272,22 @@ expand({for, Meta, [_ | _] = Args}, E) ->
         form_error(Meta, ?key(E, file), ?MODULE, {missing_option, for, [do]})
     end,
 
+  IsReducing = lists:keymember(reduce, 1, Opts),
+  IsClauses = is_list(Expr) andalso lists:any(fun({'->', _, _}) -> true; (_) -> false end, Expr),
+
+  case IsReducing andalso (lists:keymember(into, 1, Opts) orelse lists:keymember(uniq, 1, Opts)) of
+    true ->
+      form_error(Meta, ?key(E, file), ?MODULE, for_duplicate_reduce_into_uniq);
```
fertapric: Maybe `for_with_conflicting_reduce_into_uniq` or `for_conflicting_reduce_into_uniq`?

lib/elixir/src/elixir_expand.erl
```diff
@@ -272,6 +272,22 @@ expand({for, Meta, [_ | _] = Args}, E) ->
...
+    false ->
+      ok
+  end,
+
+  case {IsReducing, IsClauses} of
+    {true, false} -> form_error(Meta, ?key(E, file), ?MODULE, for_with_reduce_bad_args);
```
fertapric: And maybe these two `for_with_reduce_bad_args -> for_with_reduce_bad_do_block` and `for_without_reduce_with_arg -> for_bad_do_block`?

josevalim: Since those names are only used internally, I think it is fine to keep them as is. :+1:

lib/elixir/src/elixir_expand.erl
```diff
@@ -272,6 +272,22 @@ expand({for, Meta, [_ | _] = Args}, E) ->
         form_error(Meta, ?key(E, file), ?MODULE, {missing_option, for, [do]})
     end,
 
+  IsReducing = lists:keymember(reduce, 1, Opts),
+  IsClauses = is_list(Expr) andalso lists:any(fun({'->', _, _}) -> true; (_) -> false end, Expr),
```
michalmuskala: Should this check `lists:all` and not `any`?

michalmuskala:
There's a lot of checks and parsing and validating of the `for` options, which seems quite complex and also requires passing through the options multiple times. What if we defined a record for the options, parsed the options into the record using `lists:foldl` and then validated the record? I think it will make the implementation much simpler and obvious into what's happening - I find the current code is starting to be hard to follow.

lib/elixir/src/elixir_expand.erl
josevalim: It is always all or none.  So theoretically we need to check only the first.

josevalim:

PR #8533, opened 2018-12-16
Perform {type_syntax, _} checks in Kernel.Typespec 

### LINKED ISSUE #8242: Perform {type_syntax, _} checks in Kernel.Typespec
Today `erl_lint` performs checks for syntax in typespecs (search for `{type_syntax` [in this file](https://github.com/erlang/otp/blob/master/lib/stdlib/src/erl_lint.erl)). Since we will no longer run `erl_lint`, we need to perform those checks ourselves.

They are:

  * [x] Check that the (... -> ...) syntax is a product or any on the left side
  * [ ] Check that both sides of `x..y` are integers after expansion
  * [ ] Check that all `x` and `y` in `<<x * y>>` are non negative integers
  * [ ] Check that the record name in `record(name, ...)` is an atom

This is a leftover from #5800.

fertapric:
Closes #8242

### DIFF
--- a/lib/elixir/lib/kernel/typespec.ex
+++ b/lib/elixir/lib/kernel/typespec.ex
@@ -588,9 +588,12 @@ defmodule Kernel.Typespec do
   end
 
   # Handle ranges
-  defp typespec({:.., meta, args}, vars, caller, state) do
-    {args, state} = Enum.map_reduce(args, state, &typespec(&1, vars, caller, &2))
-    {{:type, line(meta), :range, args}, state}
+  defp typespec({:.., meta, [left, right]}, vars, caller, state) do
+    {left, state} = typespec(left, vars, caller, state)
+    {right, state} = typespec(right, vars, caller, state)
+    :ok = validate_range(left, right, caller)
+
+    {{:type, line(meta), :range, [left, right]}, state}
   end
 
   # Handle special forms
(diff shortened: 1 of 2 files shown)


lib/elixir/lib/kernel/typespec.ex
```diff
@@ -587,10 +591,22 @@ defmodule Kernel.Typespec do
...
   # Handle ranges
-  defp typespec({:.., meta, args}, vars, caller, state) do
-    {args, state} = Enum.map_reduce(args, state, &typespec(&1, vars, caller, &2))
-    {{:type, line(meta), :range, args}, state}
+  defp typespec({:.., meta, [left, right]}, vars, caller, state) do
+    {left, state} = typespec(left, vars, caller, state)
+    {right, state} = typespec(right, vars, caller, state)
+
+    case {:erl_eval.partial_eval(left), :erl_eval.partial_eval(right)} do
+      {{:integer, _, first}, {:integer, _, last}} when first < last ->
```
fertapric: I included this because [erl_lint forces the "from/first" part of the range to be lower than the "to/last" part](https://github.com/erlang/otp/blob/master/lib/stdlib/src/erl_lint.erl#L2848).

lib/elixir/lib/kernel/typespec.ex
```diff
@@ -587,10 +591,22 @@ defmodule Kernel.Typespec do
...
+
   # Handle ranges
-  defp typespec({:.., meta, args}, vars, caller, state) do
-    {args, state} = Enum.map_reduce(args, state, &typespec(&1, vars, caller, &2))
-    {{:type, line(meta), :range, args}, state}
+  defp typespec({:.., meta, [left, right]}, vars, caller, state) do
+    {left, state} = typespec(left, vars, caller, state)
+    {right, state} = typespec(right, vars, caller, state)
+
+    case {:erl_eval.partial_eval(left), :erl_eval.partial_eval(right)} do
```
josevalim: I would prefer if we do not depend on erl_eval. Is there a reason why we need it?

josevalim:
Looks great @fertapric! We should use this opportunity to improve the error messages. For example, "invalid range type, expected both sides to be integer with the left side less than or equal to the right side" and so on.

lib/elixir/lib/kernel/typespec.ex
fertapric: I was currently writing a comment related to this.

In short words, it's required for negative values. But the logic is quite simple, so we can roll our own implementation to "expand" the values.

josevalim: Let's roll our own then, especially because this seems to be private API (it is not publicly documented). :)

josevalim:

PR #8534, opened 2018-12-16
Add option to inspect tuples as records

wojtekmach:
Proposal: https://groups.google.com/forum/#!topic/elixir-lang-core/UlQIfOgNdK0

* Add `:records` key to `Inspect.Opts`

* Update `Inspect.Tuple` to format tuples as records
  if they are found in `opts.records`

* `defrecord` macro now defines a `__record__/1` function

Elixir example:

```elixir
# grab module from a manifest
iex -S mix
iex> [manifest] = Mix.Tasks.Compile.Elixir.manifests()
iex> module = Mix.Compilers.Elixir.read_manifest(manifest, "") |> hd()

iex> module
{:module, String.Chars.Decimal, {:impl, String.Chars}, ["lib/decimal.ex"],
 "Elixir.String.Chars.Decimal.beam", nil, nil}

iex> IEx.configure(inspect: [records: [module: Mix.Compilers.Elixir]])
iex> module
#module(
  module: String.Chars.Decimal,
  kind: {:impl, String.Chars},
  sources: ["lib/decimal.ex"],
  beam: "Elixir.String.Chars.Decimal.beam",
  binary: nil,
  struct: nil
)
```

Erlang example:

```elixir
# grab `file_info` record fields
iex> records = for {name, fields} <- Record.extract_all(from_lib: "kernel/include/file.hrl"), into: %{}, do: {{name, length(fields)}, {nil, Keyword.keys(fields)}
iex> inspect_opts = [records: records]
iex> {:ok, file_info} = :file.read_file_info("README.md")
iex> IO.inspect(file_info, inspect_opts)
#file_info(
  size: 4712,
  type: :regular,
  access: :read_write,
  atime: {{2018, 11, 27}, {13, 56, 3}},
  mtime: {{2018, 11, 23}, {22, 42, 6}},
  ctime: {{2018, 11, 23}, {22, 42, 6}},
  mode: 33188,
  links: 1,
  major_device: 16777223,
  minor_device: 0,
  inode: 5936255,
  uid: 501,
  gid: 20
)
{:file_info, 4712, :regular, :read_write, {{2018, 11, 27}, {13, 56, 3}},
 {{2018, 11, 23}, {22, 42, 6}}, {{2018, 11, 23}, {22, 42, 6}}, 33188, 1,
 16777223, 0, 5936255, 501, 20}
```

Possible future work:

  * [x] Make it easier to extract record fields from Elixir modules

  * Support `:only` and `:except` options, similarly to `@derive {Inspect, ...}`, e.g.:

    ```elixir
    iex> inspect_opts = %{module: {Mix.Compilers.Elixir, only: [:module]}}
    iex> IO.inspect(module)
    #module(module: String.Chars.Decimal, ...)
    ```

### DIFF
--- a/lib/elixir/test/elixir/record_test.exs
+++ b/lib/elixir/test/elixir/record_test.exs
@@ -127,6 +127,16 @@ defmodule RecordTest do
     refute match?(user(_: "other"), user())
   end
 
+  test "__record__/1" do
+    assert __record__(:timestamp) == [:date, :time]
+    assert __record__(:user) == [:name, :age]
+
+    # private record
+    assert_raise FunctionClauseError, fn ->
+      __record__(:certificate)
+    end
+  end
+
   Record.defrecord(
     :defaults,
     struct: ~D[2016-01-01],
--- a/lib/elixir/lib/inspect.ex
+++ b/lib/elixir/lib/inspect.ex
@@ -235,13 +235,28 @@ defimpl Inspect, for: Tuple do
     list = Tuple.to_list(tuple)
     [record_name | values] = list
 
-    with {:ok, fields} <- Map.fetch(opts.records, record_name),
-         true <- length(fields) == tuple_size(tuple) - 1 do
-      inspect_record(record_name, fields, values, opts)
-    else
+    case record_fields(opts.records, record_name, tuple) do
+      {:ok, fields} ->
+        inspect_record(record_name, fields, values, opts)
+
       _ ->
         inspect_tuple(list, opts)
     end
+  end
+
+  defp record_fields(records, record_name, tuple) do
+    with {:ok, fields} <- Map.fetch(records, record_name),
+         validate_fields(fields, records),
+         true <- length(fields) == tuple_size(tuple) - 1,
+         do: {:ok, fields}
+  end
+
+  defp validate_fields(fields, _records) when is_list(fields), do: true
+
+  defp validate_fields(fields, records) do
+    raise ArgumentError,
+          "expected a list of field names, got #{Kernel.inspect(fields)}" <>
+            " in #{Kernel.inspect(records)}"
   end
 
   defp inspect_tuple(list, opts) do
--- a/lib/elixir/lib/inspect.ex
+++ b/lib/elixir/lib/inspect.ex
@@ -236,7 +236,7 @@ defimpl Inspect, for: Tuple do
     [record_name | values] = list
 
     case record_fields(opts.records, record_name, tuple) do
-      {:ok, fields} ->
+      {:ok, {_module, fields}} ->
         inspect_record(record_name, fields, values, opts)
 
       _ ->
@@ -245,18 +245,8 @@ defimpl Inspect, for: Tuple do
   end
 
   defp record_fields(records, record_name, tuple) do
-    with {:ok, fields} <- Map.fetch(records, record_name),
-         validate_fields(fields, records),
-         true <- length(fields) == tuple_size(tuple) - 1,
-         do: {:ok, fields}
-  end
-
-  defp validate_fields(fields, _records) when is_list(fields), do: true
-
-  defp validate_fields(fields, records) do
-    raise ArgumentError,
-          "expected a list of field names, got #{Kernel.inspect(fields)}" <>
-            " in #{Kernel.inspect(records)}"
+    fields_size = tuple_size(tuple) - 1
+    Map.fetch(records, {record_name, fields_size})
   end
 
   defp inspect_tuple(list, opts) do
--- a/lib/elixir/lib/inspect.ex
+++ b/lib/elixir/lib/inspect.ex
@@ -229,12 +229,36 @@ defimpl Inspect, for: List do
 end
 
 defimpl Inspect, for: Tuple do
+  def inspect({}, opts), do: inspect_tuple([], opts)
+
   def inspect(tuple, opts) do
-    open = color("{", :tuple, opts)
+    list = Tuple.to_list(tuple)
+    [record_name | values] = list
+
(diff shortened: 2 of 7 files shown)


josevalim:
Hi @wojtekmach, I believe this is a very good starting point but I believe we should discuss how to get this information from modules and how we will pass to IEx and ExUnit before merging it. Since this PR is still quite small, we can continue building on it here. :)

OvermindDL1:
I'm curious about the `#module(` printed syntax.  Records are representable by their constructor so why not just use that constructor syntax as normal so it can be copy pasted around like most inspectable types (assuming the right import)?

wojtekmach:
> can be copy pasted around like most inspectable types (assuming the right import)

exactly, sometimes we don't have the import so I went on the safer side and added `#`. I'd be OK with dropping it though as the record constructor is a macro. What do others think?

michalmuskala:
Should the records be identified by a `{tag, size}` tuple? You can easily have multiple records with the same size but different tag.

wojtekmach:
@michalmuskala could you elaborate? Currently we identify records as `{tag, field_names}` so we have the size already and we need to know the field names to construct `#tag(field1: value1, field2: value2, ...)`.

michalmuskala:
I mean the option you pass to inspect should be `records: %{{:foo, 3} => [:a, :b, :c]}` - the store should distinguish a `foo` record with 3 fields and one with 2.

wojtekmach:
@michalmuskala oh, I get it now, thank you.

I'd personally prefer to leave it up to the user to handle these "conflicts" themselves but I don't have a strong opinion about it.

If they have:

```elixir
defmodule A do
  defrecord :person, [:name]
end

defmodule B do
  defrecord :person, [:name, :email]
end
```

I'd have them do either `records: %{person: A}` or `records: %{person: B}` depending which one they expect in a particular context. If they use global config (e.g. using `IEx.configure/1`) they simply can't have it both ways.

michalmuskala:
You can totally use both records in the same code by requiring the macros and not importing - this allows you to use them qualified and call `A.person(name: ...)` and `B.person(name: ..., email: ...)`. Maybe we should use that syntax when printing?

josevalim:
@michalmuskala while it is possible, I wonder how common it is and what should the API for specifying records then be? The most efficient is `{name, length}` as keys but that requires manual book-keeping. On the value, we could have only the module name and do `__record__`  lookups, but that means remote calls on every matching record. In any case, for the most efficient way possible, we would need this:

```elixir
%{
  {record_name, record_length} => {record_module, [record_field]}
}
```

But that feels like too much?

The other approach is to remove the length, but then we need to use a list for specifying multiple entries per module.

Thoughts?

michalmuskala:
Here are some thoughts on the matter.

I think the API proposed by José is the most flexible and most "accurate".

We could also have this exact format returned from the `__info__(:records)` callback on each module. This means that to extract the data from multiple modules we could use something like this, which seems rather straightforward:

```elixir
modules = [A, B]
Enum.reduce(modules, %{}, &Map.merge(&1.__info__(:records), &2))
```

With that - maybe in `Kernel.inspect` and `IO.inspect` we should support an option called `record_modules` and inspect itself would extract this data from all the modules? It's convenient, but on the other hand it might be expensive to do that on each call.

josevalim:
Unfortunately to add __info__(:records) or similar we would need to add
“use Record” or change the compiler, and I don’t believe we should change
the user APIs or the compiler for this feature. So we need to at least
provide the record name and record module.
-- 


wojtekmach:
Okay, I'll look into supporting:

```elixir
records: %{
  {record_name, record_length} => {record_module, [record_field]}
}
```

representation which will be stored in `Inspect.Opts`.

FWIW I believe we could still make a very convenient API in IEx (and similar):

```elixir
IEx.configure(inspect: [records: [person: A]])
```

that translates to the representation above. Note, records here is a keywords list to be able to do: `records: [person: A, person: B]`.

WDYT?

> With that - maybe in Kernel.inspect and IO.inspect we should support an option called record_modules and inspect itself would extract this data from all the modules?

Yeah, I wanted something like this too!  What I ended up with on a project was:

```elixir
defmodule MyApp do
  def inspect_opts() do
    [records: ...]
  end

  def inspect(term, opts \\ inspect_opts()), do: inspect(term, opts)

  def io_inspect(term, opts \\ inspect_opts()), do: IO.inspect(term, opts)
end

IEx.configure(inspect: MyApp.inspect_opts())
```

and honestly, that wasn't so bad after all.

wojtekmach:
Updated.

We now support `%{{record_name, record_length} => {record_module, [record_field]}}`. We also have a shortcut: `IEx.configure(inspect: [records: [person: A]])`.

I started working on ExUnit diffs support and it's a bit more involved so perhaps its better to have a separate PR for it (I'm about 80% done there ;-))

lib/elixir/test/elixir/record_test.exs
```diff
@@ -127,6 +127,16 @@ defmodule RecordTest do
     refute match?(user(_: "other"), user())
   end
 
+  test "__record__/1" do
+    assert __record__(:timestamp) == [:date, :time]
```
wojtekmach: perhaps the `__record__` function would be more useful if it returns fields and default values:

```elixir
assert __record__(:timestamp) == [date: nil, time: nil]
```

I don't need such functionality at the moment, however.

josevalim:
After a conversation with @wojtekmach we agreed the footprint of this feature is too big for something that still needs to be configured manually. Therefore, we won't merge it.

One possible approach is to have an inspect_fun or callback in `Inspect.Opts` that will be invoked, allowing someone to do anything they want and plug-in custom behaviour.

PR #8536, opened 2018-12-17
Add missing typespecs to Macro.underscore/1 and Macro.do_underscore/2

MainShayne233:
Was poking around in the `Macro` module and realized these functions were missing `@spec`s.


josevalim:

josevalim:
We use `@spec` mostly for documentation, so it wouldn't apply for private functions.

PR #8537, opened 2018-12-17
Deprecate `%{key => value}` in typespecs

fertapric:
`%{required(foo) => bar}` and `%{optional(foo) => bar}` should be used instead.


josevalim:
Let's hold this for 1.9 since it drops Erlang 18 support on more recent versions.

PR #8539, opened 2018-12-18
Remove deprecated Mix.Dep.loaded/1

fertapric:
As specified by the comment `# TODO: Remove this on v1.8`


josevalim:
Unfortunately it is too late for that now. We should do it early so
everyone running on master can find bad dependencies and catch up. The RCs
are also for that but cause a large impact. So we will merge this soon
after we branch out.
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

PR #8543, opened 2018-12-20
Minor doc changes in Collectable protocol moduledoc

spejman:
In the example, add the type the protocol is implemented for (`MapSet`).

When reading the documentation is not obvious that this code is extracted from the `MapSet` module and therefore the `for: MapSet` is redundant.

### DIFF
--- a/lib/elixir/lib/collectable.ex
+++ b/lib/elixir/lib/collectable.ex
@@ -41,7 +41,7 @@ defprotocol Collectable do
   implementation for `MapSet`. In this implementation "collecting" elements
   simply means inserting them in the set through `MapSet.put/2`.
 
-      defimpl Collectable do
+      defimpl Collectable, for: MapSet do
         def into(original) do
           collector_fun = fn
             set, {:cont, elem} -> MapSet.put(set, elem)


lib/elixir/lib/collectable.ex
```diff
@@ -41,7 +41,7 @@ defprotocol Collectable do
   implementation for `MapSet`. In this implementation "collecting" elements
   simply means inserting them in the set through `MapSet.put/2`.
 
-      defimpl Collectable do
+      defimpl Collectable, for: MapSet do
         def into(original) do
           collector_fun = fn
             set, {:cont, elem} -> MapSet.put(set, elem)
```
fertapric: I just noticed there is an extra `end` at the end, do you mind removing it too?

fertapric: Nevermind, that was GitHub messing up with the indentation

wojtekmach:
FWIW its ok to skip `for: ` if we are inside module, like exactly in MapSet 
https://github.com/elixir-lang/elixir/blob/v1.7.4/lib/elixir/lib/map_set.ex#L385, but I guess its fine as is too.

```
defmodule MapSet do
  # ...

  defimpl Collectable do
  end
end
```

would be preferred? We’d both match existing code as well as show inner defimpl.

josevalim:
@wojtekmach good point. Although it is probably clearer to keep the `:for`.

josevalim:

PR #6641, opened 2017-10-07
Show the multiple clauses for the same callback on IEx.Introspection.b/1

kelvinst:
Fixes #6546

Issue #6546, opened 2017-09-05
Multiple @callback with the same name/arity are discarded


michalmuskala:
### Environment

* Elixir & Erlang versions (elixir --version): 1.6.0-dev, OTP 20
* Operating system: macOS

### Current behavior

Given the following module:

```elixir
defmodule Test do
  @callback test(:foo) :: String.t
  @callback test(:bar) :: [String.t]
end
```

When later getting callback info from iex, we only get the last clause:

```elixir
iex(1)> b Test
@callback test(:bar) :: [String.t()]
```

### Expected behavior

I'm not 100% sure what the expected behaviour is, two possible solutions I see are:

* return all the clauses in `b`.
* warn/raise when defining multiple callbacks like that in the module.

sorentwo:
I've wondered about this recently as well, and didn't find anything conclusive in the Erlang docs. Hopefully somebody more knowledgeable will provide the correct answer.

josevalim:
The first step is to see if Erlang supports multiple callbacks with the same name and, if so, does it return the same result?

alco:
Erlang/OTP 20.0 refuses to compile such code:

```erlang
-module(test).

-callback test(foo) -> string.
-callback test(bar) -> [string].
```

```
$ erlc test.erl
test.erl:4: callback test/1 already defined
```

kelvinst:
So let's add a compile error for it right?

Just one more thing: Erlang/OTP 20 also refuses to compile duplicated spec definitions.

```erlang
-module(test).

-spec test(foo) -> string.
-spec test(bar) -> [string].
```

```erlang
$ erlc test.erl
test.erl:4: spec for test/1 already defined
```

Let's add the compile check for both?

josevalim:
The way multiple clauses work in Erlang is that they are separated by comma:

```erlang
-module(foo).
-callback test(foo) -> string;
              (bar) -> [string].
```

In Elixir we don't have the same syntax so we allow the same one to specified multiple times and then we merge them together in a later pass.

We do this later pass for specs but apparently we don't do it for callbacks, so that's what needs to be done. :)

kelvinst:
> The way multiple clauses work in Erlang is that they are separated by comma:

Well, living and learning. 🤣

> We do this later pass for specs but apparently we don't do it for callbacks, so that's what needs to be done. :)

If no one else is interested, I can do it! :)

josevalim:
Please go ahead :)

PR #8546, opened 2018-12-21
Add test for showing docs of callback with multiple clauses

wojtekmach:
We've introduced showing of multiple callback clauses in https://github.com/elixir-lang/elixir/pull/6641.

I noticed that docs are not attached to these multiple clauses though, see inline comments.

### DIFF
--- a/lib/iex/test/iex/helpers_test.exs
+++ b/lib/iex/test/iex/helpers_test.exs
@@ -659,14 +659,17 @@ defmodule IEx.HelpersTest do
              """
     end
 
-    test "lists callback with multiple clauses" do
+    test "callback with multiple clauses" do
       filename = "multiple_clauses_callback.ex"
 
       content = """
       defmodule MultipleClauseCallback do
-        @doc "callback"
+        @doc "A callback"
         @callback test(:foo) :: integer
         @callback test(:bar) :: [integer]
+
+        @doc "Another callback"
+        @callback test2(:foo) :: integer
       end
       """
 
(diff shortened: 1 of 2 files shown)


lib/iex/test/iex/helpers_test.exs
```diff
@@ -676,6 +679,25 @@ defmodule IEx.HelpersTest do
         assert capture_io(fn -> b(MultipleClauseCallback) end) =~ """
                @callback test(:foo) :: integer()
                @callback test(:bar) :: [integer()]
+
+               @callback test2(:foo) :: integer()
+               """
+
+        IO.inspect IEx.Introspection.get_docs(MultipleClauseCallback, [:callback])
```
wojtekmach: This prints:

```elixir
[
  {{:callback, :test, 1}, 4, [], :none, %{}},
  {{:callback, :test, 1}, 4, [], :none, %{}},
  {{:callback, :test2, 1}, 6, [], %{"en" => "Another callback"}, %{}}
]
```

I'd expect to see "A callback" docs attached, but not sure how exactly that would look like, not sure if we should duplicate it for all clauses? Or it should be represented in a different way altogether?

josevalim:
Can we actually have multiple causes for the same callback? I would
validate it such is possible. If it is possible, then the next step is to
fix the bug and merge them together.

josevalim:
To be fair, we likely already have a test for multiple clauses of a callback. We probably don't have a test for multiple callbacks with docs. :)

wojtekmach:
@josevalim Erlang has multiple clauses https://github.com/elixir-lang/elixir/issues/6546#issuecomment-333922301 and we do some work to be compatible.

The test is not final, I just wanted to showcase the bug that we're not attaching the docs to the callback if it has multiple clauses, and I'm looking into fixing that bug.

PR #8553, opened 2018-12-21
Show docs for callback with multiple clauses

wojtekmach:


lib/elixir/test/elixir/kernel/docs_test.exs
```diff
@@ -211,6 +215,8 @@ defmodule Kernel.DocsTest do
         callback_bar,
         callback_baz,
         callback_foo,
+        callback_qux,
+        callback_qux,
```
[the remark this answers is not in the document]
josevalim: According to my understand of the code this should be correct. So maybe we have something in `elixir_erl.erl` that is duplicating it again?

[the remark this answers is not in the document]
wojtekmach: @josevalim Good call. I noticed the issue in elixir_erl and fixed it, but I couldn't come up with a nicer solution yet. Will keep at it.

lib/elixir/lib/kernel/typespec.ex
```diff
@@ -108,8 +108,11 @@ defmodule Kernel.Typespec do
 
     case spec_to_signature(expr) do
       {name, arity} ->
-        {line, doc} = get_doc_info(set, :doc, line)
-        store_doc(set, kind, name, arity, line, :doc, doc, %{})
+        # store doc only once in case callback has multiple clauses
+        unless :ets.member(set, {kind, name, arity}) do
```
josevalim: Just checking: is this still necessary with the new fix?

lib/elixir/src/elixir_erl.erl
```diff
@@ -318,7 +318,23 @@ typespecs_form(Map, TranslatedTypespecs, MacroNames) ->
   Forms1 = types_form(Types, Forms0),
   Forms2 = callspecs_form(spec, Specs, [], MacroNames, Forms1, Map),
   Forms3 = callspecs_form(callback, AllCallbacks, OptionalCallbacks, MacroCallbackNames, Forms2, Map),
-  {Types, AllCallbacks, Forms3}.
+  AllCallbacks2 = typespecs_back_to_callbacks(Forms3),
```
josevalim: Are we doing this because of duplicates in AllCallbacks? If so, maybe it would be better to call `lists:uniq` or `lists:usort` in callbacks?

lib/elixir/lib/kernel/typespec.ex
wojtekmach: Yes, still needed.

lib/elixir/src/elixir_erl.erl
wojtekmach: Good call, I'm now stripping spec information from callbacks and usorting them, check out latest commit.

lib/elixir/src/elixir_erl.erl
```diff
@@ -318,23 +318,15 @@ typespecs_form(Map, TranslatedTypespecs, MacroNames) ->
...
-typespec_back_to_callback({Name, Arity}, Line, Rest) ->
-  {Kind, FA} =
-    case atom_to_list(Name) of
-      "MACRO-" ++ Name2 -> {macrocallback, {list_to_atom(Name2), Arity - 1}};
-      _ -> {callback, {Name, Arity}}
-    end,
 
-  {Kind, FA, Line, Rest}.
+  AllCallbacksWithoutSpecs = [
+    {Kind, NameArity, Line} || {Kind, NameArity, Line, _Spec} <- AllCallbacks
```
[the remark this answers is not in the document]
wojtekmach: Good call, fixed.

lib/elixir/src/elixir_erl.erl
```diff
@@ -318,7 +318,12 @@ typespecs_form(Map, TranslatedTypespecs, MacroNames) ->
   Forms1 = types_form(Types, Forms0),
   Forms2 = callspecs_form(spec, Specs, [], MacroNames, Forms1, Map),
   Forms3 = callspecs_form(callback, AllCallbacks, OptionalCallbacks, MacroCallbackNames, Forms2, Map),
-  {Types, AllCallbacks, Forms3}.
+
+  AllCallbacksWithoutSpecs = lists:usort([
+    {Kind, NameArity} || {Kind, NameArity, _Line, _Spec} <- AllCallbacks
```
josevalim: I was thinking we could already return `{Kind, Name, Arity}` here. So in the line 524, we would do `Callback <- Callbacks`.

wojtekmach: Right. I saw that in other place we had `{Kind, {Name, Arity}, _, _}` (specs I believe) but this sounds good too, I'll make the change.

wojtekmach: ok, now I see why it's nicer :)

michalmuskala:
Thank you! :heart:

PR #8557, opened 2018-12-21
Regex: Use "regular expression"

eksperimental:
...instead of the more informal term "regex"

Related to bd7f914075655a717a33f089af77cbae2a742c8f


josevalim:
Given that the module is called Regex and that’s a common abbreviation, I
would like to continue to use “regex”, even if it requires us to formally
define it in the module docs.
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

josevalim:
Closing as per the rationale above. Thanks! :heart:

PR #8558, opened 2018-12-21
Support non zero decimals string in String.to_float/1

schabou:
Adds capability for `String.to_float/1` to convert string represented decimals that don't contain leading digits such as ".767" which neither `String.to_float/1` or `Float.parse/1` currently.

### DIFF
--- a/lib/elixir/lib/string.ex
+++ b/lib/elixir/lib/string.ex
@@ -2276,12 +2276,12 @@ defmodule String do
   def to_float(string) do
     case String.at(string, 0) == "." do
       true ->
-        ("0" <> string)
-        |> :erlang.binary_to_float()
+        "0" <> string
 
       false ->
-        :erlang.binary_to_float(string)
+        string
     end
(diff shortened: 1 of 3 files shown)


eksperimental:
`.7` is not a valid Elixir syntax, therefore we should not support it.

schabou:
I should clarify that the resulting float is Elixir compatible, for example the string `".6"` would be converted here to `0.6`.

I went on a search to find a popular language that didn't handle this case and I came up empty handed.  As far as I can tell most programming languages in common usage support converting that format into floats.  

But if you're firm in your decision would you be open to a pull request adding a note to the documentation?

josevalim:
Hi @schabou! This has been proposed before and the consensus is indeed that we shouldn't support ".789" because we can't input floats like that into the language. Therefore, a patch for the documentation would most definitely be appreciated. Of course, you can detect for a float starting with a period yourself and add the zero accordingly before invoking those functions. Thanks!

PR #8560, opened 2018-12-23
Halt IEx when it receives EOF from stdin

josevalim:
Prior to this patch, IEx would just hang, as stdin
was dead. This aligns the behaviour with the `erl`
shell.


josevalim:

PR #8561, opened 2018-12-23
Suggest quotes in the executables help where appropriate 

josevalim:
The rationale is that those options should almost always be given in quotes so by adding the quotes in the documentation we can signal the intent.


josevalim:

PR #8563, opened 2018-12-23
#8562 Fix: added IEx warning when using --remsh with 'dumb' terminal

### LINKED ISSUE #8562: IEx should print a warning when using --remsh with "dumb" terminal
`--remsh` only works with `user_drv`, so we should warn if `--remsh` is used with a "dumb" terminal:

      TERM=dumb iex --sname bar --remsh foo@josemac

sfusato:
Fixes #8562: IEx should print a warning when using --remsh with "dumb" terminal


josevalim:
@sfusato thanks! I have rewritten the message, can you please let me know what you think? :+1:

sfusato:
@josevalim it's much better. Thank you!

josevalim:

PR #8567, opened 2018-12-26
Use a descentralized mode computation for Logger

josevalim:
Previous logger versions would always compute the mode on the Logger handler which meant that, if any other Logger handler blocked, either because it was waiting on IO, the system is overloaded or due to long computations, a burst of messages could make it to the Logger message queue grow quite large until it notices it should change its mode.

This commit changes the computation to use two counters, one that is kept by the Logger and another updated by the client. The mode is computed by adding those two counters together. On every message, the Logger handler resets the counter and updates the message queue one. This makes the following script go from tens of thousands messages in the Logger inbox to only hundreds (assuming a `sync_threshold` of 100) when writing to stdio is rate limited with `mix run foo.exs | pv -p -L 1k 1>/dev/null`:

```elixir
require Logger

# monitor memory
# :observer.start()

defmodule Loop do
  def run() do
    Logger.info(inspect(:crypto.strong_rand_bytes(1024)))
    Process.sleep(10)
    run()
  end
end

for _i <- 1..100 do
  spawn(fn ->
    Loop.run()
  end)
end

defmodule OtherLoop do
  def run() do
    IO.inspect(:stderr, Process.info(Process.whereis(Logger), :message_queue_len), [])
    IO.inspect(:stderr, :ets.tab2list(Logger.Config), [])
    Process.sleep(100)
    run()
  end
end

OtherLoop.run()
```


lib/logger/lib/logger.ex
```diff
@@ -133,15 +133,16 @@ defmodule Logger do
...
+      Defaults to 20 messages. `:sync_threshold` can be set to `0` to
+      force *sync mode*.
 
     * `:discard_threshold` - if the `Logger` manager has more than
       `:discard_threshold` messages in its queue, `Logger` will change
       to *discard mode* and messages will be discarded directly in the
       clients. `Logger` will return to *sync mode* once the number of
-      messages in the queue is reduced to `discard_threshold * 0.75`
-      messages. Defaults to 500 messages.
+      messages in the queue is reduced to one below `discard_threshold`.
```
audiolion: The new comment above on https://github.com/elixir-lang/elixir/pull/8567/files#diff-e595abf4d143c9f920f1d59b61df7c35R136 uses `one below the` so suggesting adding `the` here to match

```suggestion
      messages in the queue is reduced to one below the `discard_threshold`.
```

lib/logger/mix.exs
```diff
@@ -20,7 +20,7 @@ defmodule Logger.MixProject do
         backends: [:console],
         translators: [{Logger.Translator, :translate}],
         sync_threshold: 20,
-        discard_threshold: 500,
+        discard_threshold: 10_000,
```
fishcakez: This change could lead to 9980 processes blocking. If we are to provide this stronger guarantee on message queue length should we also set sync threshold to `10_000` by default? I don't think we really need `sync` any more unless desired? Also we don't want to block for the time it takes to handle `9999` events. This would be very bad buffer bloat for events not discarded.

josevalim:
Thanks @fishcakez! I will reply trying to address all of your comments at once.

First, regarding `read_concurrency`, given that we have multiple processes reading and only one writing, I thought that `read_concurrency` would still be the best approach. Moving the message queue len to the write table is not a bad idea though, because we would be able to move to a counter (if we keep the current implementation). The cost is that we have to perform one extra read every time we log as we have to read the log data, the message queue len and the counter.

Regarding logging a message when we start discard, the tricky part is to guarantee that it only happens from time to time: i.e. if for some reason our log rate makes our counter close the discard threshold, then we would log that we started discarding multiple times. So my idea was to have a separate process that receives a message every time when there is discarding and it emits the discard log message only from to time, for example, every 30 seconds. If we have this timer process, then we can also use this process to solve the "discard deadlock" you mentioned, where we could stay on discard state forever. So this discard timer process would 1. receive a message on every discard, 2. log whenever it enters into discard mode, 3. ignore all other messages for the next 30 seconds, 4. emit a log event with the current status (no longer discarding, still discarding, etc) after the 30 seconds.

However, we still need to  discuss which implementation we want to use for the write table. We can use `ets+update_counter`, and that would be atomic + linear, potentially simplifying some things. However, it will be a lock for all processes. Using counters will increase concurrency but there is a chance of occurring the problem [you describe here](https://github.com/elixir-lang/elixir/pull/8567#discussion_r244647509) - which I believe is minor. For this reason, I am more inclined to go with counters.

Thoughts?

lib/logger/mix.exs
josevalim: I  think we need to bump the current threshold but maybe `10_000` is way too hight. Would 2000 alleviate the issue?

fishcakez:
> However, it will be a lock for all processes.

Doesn't the current implementation require locking all processes? It acts on an `atomics` counter twice? Without knowing how counters is implemented it may not be more efficient? Also the current implementation *may* have some overhead when not overloaded as the switch between read and write is more expensive with read concurrency (assuming 1 read, 1 write, 1 read, 1 write, due to caller/logger switch, the worst case for read concurrency). It seems like we should measure the different approaches.

michalmuskala:
Counters are implemented using hardware atomic operations, so there's no locking involved.

lib/logger/lib/logger/config.ex
```diff
@@ -21,187 +17,239 @@ defmodule Logger.Config do
...
 
-  def code_change(_old, state, _extra) do
-    {:ok, state}
+  def code_change(_old, counter, _extra) do
+    {:ok, counter}
+  end
+
+  # It is very important to reset the counter first and then update
+  # the message queue length because that's the pessimistic approach.
+  # If the system is overloaded, then we will overestimate the overall
```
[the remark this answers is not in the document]
josevalim: Done!

lib/logger/mix.exs
josevalim: Reverted.

josevalim:
I have pushed some changes so we no longer write to the table on every read. Instead, we keep the read table only for reads and the write table (or counter) for writes (i.e. both counter and message queue len). I will run some benchmarks next.

josevalim:
Here are the values for bumping the counter vs update counters in ETS. The first parameter is the number of processes and followed by how far we want to count. It prints the time in microseconds followed by the value of the counter at the end.

```
~/OSS/elixir[jv-better-logger %]$ elixir foo.exs 1 1000000
BumpCounter
{91819, 1000000}
BumpETS
{139681, 1000000}
~/OSS/elixir[jv-better-logger %]$ elixir foo.exs 2 1000000
BumpCounter
{70487, 1000000}
BumpETS
{119023, 1000001}
~/OSS/elixir[jv-better-logger %]$ elixir foo.exs 4 1000000
BumpCounter
{45005, 1000003}
BumpETS
{309280, 1000003}
~/OSS/elixir[jv-better-logger %]$ elixir foo.exs 16 1000000
BumpCounter
{38368, 1000012}
BumpETS
{1375083, 1000015}
```

This is a machine with four cores. Contention increases considerably with ETS. Here is the code:

```elixir
[instances, limit] = System.argv() |> Enum.map(&String.to_integer/1)
wait = 0

defmodule BumpCounter do
  start =
    for _ <- 1..instances do
      quote do
        bumper(self(), var!(counter), unquote(limit), unquote(wait))
      end
    end

  stop =
    for _ <- 1..instances do
      quote do
        receive do
          :done -> :ok
        end
      end
    end

  def run do
    counter = :counters.new(1, [:atomics])

    fn ->
      unquote_splicing(start)
      unquote_splicing(stop)
      :counters.get(counter, 1)
    end
    |> :timer.tc()
    |> IO.inspect()
  end

  defp bumper(parent, counter, limit, wait) do
    spawn_link(fn -> bumper_loop(parent, counter, limit, wait) end)
  end

  defp bumper_loop(parent, counter, limit, wait) do
    :counters.add(counter, 1, 1)

    if :counters.get(counter, 1) >= limit do
      send(parent, :done)
    else
      Process.sleep(wait)
      bumper_loop(parent, counter, limit, wait)
    end
  end
end

defmodule BumpETS do
  start =
    for _ <- 1..instances do
      quote do
        bumper(self(), var!(counter), unquote(limit), unquote(wait))
      end
    end

  stop =
    for _ <- 1..instances do
      quote do
        receive do
          :done -> :ok
        end
      end
    end

  def run do
    counter = :ets.new(:counter, [:set, :public, write_concurrency: true])
    :ets.insert(counter, {:counter, 0})

    fn ->
      unquote_splicing(start)
      unquote_splicing(stop)
      :ets.lookup_element(counter, :counter, 2)
    end
    |> :timer.tc()
    |> IO.inspect()
  end

  defp bumper(parent, counter, limit, wait) do
    spawn_link(fn -> bumper_loop(parent, counter, limit, wait) end)
  end

  defp bumper_loop(parent, counter, limit, wait) do
    if :ets.update_counter(counter, :counter, {2, 1}) >= limit do
      send(parent, :done)
    else
      Process.sleep(wait)
      bumper_loop(parent, counter, limit, wait)
    end
  end
end

IO.puts "BumpCounter"
Task.async(fn -> BumpCounter.run() end) |> Task.await(:infinity)
IO.puts "BumpETS"
Task.async(fn -> BumpETS.run() end) |> Task.await(:infinity)
```

I will benchmark now one writing and multiple reading.

josevalim:
Here are the results for read and writes with a fixed number of processes:

```
~/OSS/elixir[jv-better-logger %]$ elixir foo.exs 1 100000
ReadWriteCounter
{77914, 100000}
ReadWriteETS (read concurrency)
{89361, 100000}
ReadWriteETS (write concurrency)
{88667, 100000}
ReadWriteETS (none)
{79059, 100000}
~/OSS/elixir[jv-better-logger %]$ elixir foo.exs 2 100000
ReadWriteCounter
{162649, 100001}
ReadWriteETS (read concurrency)
{141792, 100001}
ReadWriteETS (write concurrency)
{152847, 100001}
ReadWriteETS (none)
{147481, 100001}
~/OSS/elixir[jv-better-logger %]$ elixir foo.exs 4 100000
ReadWriteCounter
{186416, 100002}
ReadWriteETS (read concurrency)
{204912, 100002}
ReadWriteETS (write concurrency)
{192689, 100001}
ReadWriteETS (none)
{178800, 100002}
~/OSS/elixir[jv-better-logger %]$ elixir foo.exs 16 100000
ReadWriteCounter
{151358, 100010}
ReadWriteETS (read concurrency)
{153701, 100001}
ReadWriteETS (write concurrency)
{148644, 100012}
ReadWriteETS (none)
{145476, 100009}
~/OSS/elixir[jv-better-logger %]$ elixir foo.exs 32 100000
ReadWriteCounter
{160449, 100018}
ReadWriteETS (read concurrency)
{149312, 100030}
ReadWriteETS (write concurrency)
{151108, 100028}
ReadWriteETS (none)
{145465, 100030}
```

The ETS table with read/write concurrency is more efficient but the difference is not large because the writer process is the bottleneck.

Here is the code:

```elixir
[instances, limit] = System.argv() |> Enum.map(&String.to_integer/1)

defmodule ReadWriteCounter do
  start =
    for _ <- 1..instances do
      quote do
        reader(self(), var!(incrementer), var!(counter), unquote(limit))
      end
    end

  stop =
    for _ <- 1..instances do
      quote do
        receive do
          :done -> :ok
        end
      end
    end

  def run do
    counter = :counters.new(1, [:atomics])
    incrementer = incrementer(counter)

    fn ->
      unquote_splicing(start)
      unquote_splicing(stop)
      :counters.get(counter, 1)
    end
    |> :timer.tc()
    |> IO.inspect()
  end

  defp incrementer(counter) do
    spawn_link(fn -> increment_loop(counter, 0) end)
  end

  defp increment_loop(counter, value) do
    receive do
      {:bump, pid} ->
        :counters.put(counter, 1, value)
        send(pid, :bumped)
        increment_loop(counter, value + 1)
    end
  end

  defp reader(parent, incrementer, counter, limit) do
    spawn_link(fn -> read_loop(parent, incrementer, counter, limit) end)
  end

  defp read_loop(parent, incrementer, counter, limit) do
    if :counters.get(counter, 1) >= limit do
      send(parent, :done)
    else
      send(incrementer, {:bump, self()})

      receive do
        :bumped -> read_loop(parent, incrementer, counter, limit)
      end
    end
  end
end

defmodule ReadWriteETS do
  start =
    for _ <- 1..instances do
      quote do
        reader(self(), var!(incrementer), var!(counter), unquote(limit))
      end
    end

  stop =
    for _ <- 1..instances do
      quote do
        receive do
          :done -> :ok
        end
      end
    end

  def run(modes) do
    counter = :ets.new(:counter, [:public] ++ modes)
    :ets.insert(counter, {:counter, 0})
    incrementer = incrementer(counter)

    fn ->
      unquote_splicing(start)
      unquote_splicing(stop)
      :ets.lookup_element(counter, :counter, 2)
    end
    |> :timer.tc()
    |> IO.inspect()
  end

  defp incrementer(counter) do
    spawn_link(fn -> increment_loop(counter, 0) end)
  end

  defp increment_loop(counter, value) do
    receive do
      {:bump, pid} ->
        :ets.update_element(counter, :counter, {2, value})
        send(pid, :bumped)
        increment_loop(counter, value + 1)
    end
  end

  defp reader(parent, incrementer, counter, limit) do
    spawn_link(fn -> read_loop(parent, incrementer, counter, limit) end)
  end

  defp read_loop(parent, incrementer, counter, limit) do
    if :ets.lookup_element(counter, :counter, 2) >= limit do
      send(parent, :done)
    else
      send(incrementer, {:bump, self()})

      receive do
        :bumped -> read_loop(parent, incrementer, counter, limit)
      end
    end
  end
end

IO.puts("ReadWriteCounter")
Task.async(fn -> ReadWriteCounter.run() end) |> Task.await(:infinity)
IO.puts "ReadWriteETS (read concurrency)"
Task.async(fn -> ReadWriteETS.run([read_concurrency: true]) end) |> Task.await(:infinity)
IO.puts "ReadWriteETS (write concurrency)"
Task.async(fn -> ReadWriteETS.run([write_concurrency: true]) end) |> Task.await(:infinity)
IO.puts "ReadWriteETS (none)"
Task.async(fn -> ReadWriteETS.run([]) end) |> Task.await(:infinity)
```

I will try the burst approach next.

josevalim:
Here is the burst results (spawning multiple processes were each does only one operation):

```elixir
~/OSS/elixir[jv-better-logger %]$ elixir foo.exs 16 10000
BurstCounter
{92745, 0}
BurstETS (read concurrency)
{694400, 0}
BurstETS (write concurrency)
{731023, 0}
BurstETS (none)
{742579, 0}
~/OSS/elixir[jv-better-logger %]$ elixir foo.exs 4 40000
BurstCounter
{107563, 0}
BurstETS (read concurrency)
{589730, 0}
BurstETS (write concurrency)
{94679, 0}
BurstETS (none)
{572116, 0}
~/OSS/elixir[jv-better-logger %]$ elixir foo.exs 2 80000
BurstCounter
{160183, 0}
BurstETS (read concurrency)
{149105, 0}
BurstETS (write concurrency)
{142469, 0}
BurstETS (none)
{139383, 0}
~/OSS/elixir[jv-better-logger %]$ elixir foo.exs 1 160000
BurstCounter
{265164, 0}
BurstETS (read concurrency)
{245035, 0}
BurstETS (write concurrency)
{234930, 0}
BurstETS (none)
{230173, 0}
```

Here is the code:

```elixir
[instances, limit] = System.argv() |> Enum.map(&String.to_integer/1)

start =
  for _ <- 1..instances do
    quote do
      spawn(fn -> burst_loop(var!(incrementer), var!(counter), unquote(limit)) end)
    end
  end

defmodule BurstCounter do
  def run do
    counter = :counters.new(1, [:atomics])
    incrementer = incrementer(self(), counter, unquote(instances * limit))

    fn ->
      unquote_splicing(start)

      receive do
        :done -> :counters.get(counter, 1)
      end
    end
    |> :timer.tc()
    |> IO.inspect()
  end

  defp incrementer(parent, counter, limit) do
    spawn_link(fn -> increment_loop(parent, counter, limit) end)
  end

  defp increment_loop(parent, counter, 0) do
    :counters.put(counter, 1, 0)
    send(parent, :done)
  end

  defp increment_loop(parent, counter, value) do
    receive do
      :bump ->
        :counters.put(counter, 1, value)
        increment_loop(parent, counter, value - 1)
    end
  end

  defp burst_loop(_incrementer, _counter, 0) do
    :ok
  end

  defp burst_loop(incrementer, counter, limit) do
    spawn(fn ->
      :counters.get(counter, 1)
      send(incrementer, :bump)
    end)

    burst_loop(incrementer, counter, limit - 1)
  end
end

defmodule BurstETS do
  def run(modes) do
    counter = :ets.new(:counter, [:public] ++ modes)
    :ets.insert(counter, {:counter, 0})
    incrementer = incrementer(self(), counter, unquote(instances * limit))

    fn ->
      unquote_splicing(start)

      receive do
        :done -> :ets.lookup_element(counter, :counter, 2)
      end
    end
    |> :timer.tc()
    |> IO.inspect()
  end

  defp incrementer(parent, counter, limit) do
    spawn_link(fn -> increment_loop(parent, counter, limit) end)
  end

  defp increment_loop(parent, counter, 0) do
    :ets.update_element(counter, :counter, {2, 0})
    send(parent, :done)
  end

  defp increment_loop(parent, counter, value) do
    receive do
      :bump ->
        :ets.update_element(counter, :counter, {2, value})
        increment_loop(parent, counter, value - 1)
    end
  end

  defp burst_loop(_incrementer, _counter, 0) do
    :ok
  end

  defp burst_loop(incrementer, counter, limit) do
    spawn(fn ->
      :ets.lookup_element(counter, :counter, 2)
      send(incrementer, :bump)
    end)

    burst_loop(incrementer, counter, limit - 1)
  end
end

IO.puts("BurstCounter")
Task.async(fn -> BurstCounter.run() end) |> Task.await(:infinity)
IO.puts("BurstETS (read concurrency)")
Task.async(fn -> BurstETS.run(read_concurrency: true) end) |> Task.await(:infinity)
IO.puts("BurstETS (write concurrency)")
Task.async(fn -> BurstETS.run(write_concurrency: true) end) |> Task.await(:infinity)
IO.puts("BurstETS (none)")
Task.async(fn -> BurstETS.run([]) end) |> Task.await(:infinity)
```

Counters behave consistently better under load.

josevalim:
TL;DR - counters perform better.

josevalim:
@fishcakez the last commit adds the logging back and has a periodic check, every 30 seconds, that reports the status of the system and heals the counters. Once we enter in discard mode, the next status update is only on the next periodic check, this guarantees we won't be checking all the time.

josevalim:
@fishcakez I think we can do a single counter after all. Good call.

josevalim:
I pushed a single counter.

lib/logger/lib/logger/config.ex
```diff
@@ -139,69 +143,147 @@ defmodule Logger.Config do
...
+  defp schedule_update_counter({_, _, _, discard_period}) do
+    Process.send_after(self(), {__MODULE__, :update_counter}, discard_period)
+  end
+
+  ## Counter Helpers
+
+  # TODO: Use counters exclusively when we require Erlang/OTP 22+.
+  defp new_counter() do
+    if Code.ensure_loaded?(:counters) do
+      {:counters, :counters.new(1, [:atomics])}
```
michalmuskala: Have you explored using the `write_concurrency` mode? As far as I understand the counters are not perfectly consistent anyway and provide more of a heuristic value rather than a hard limit.

josevalim: I did in the benchmarks. Because it is per row and we have a single row, write_concurrency actually make things worse.e

josevalim:
```iex
iex> :math.pow(2, 64)
1.8446744073709552e19
```

Yeah... we are fine.

lib/logger/lib/logger/config.ex
```diff
@@ -139,69 +143,147 @@ defmodule Logger.Config do
...
+    cond do
+      total >= discard_threshold ->
+        if log == :log or periodic_check? do
+          warn("Attempted to log #{total} messages, which is above :discard_threshold")
+        end
+
+        {counter, :discard, discard_threshold, discard_period}
+
+      log == :discard and periodic_check? ->
+        warn("Attempted to log #{total} messages, which is below :discard_threshold")
```
fishcakez: There isnt a test for this feature

josevalim: I tried but it is really tricky to test it without adding more complexity and/or adding long timers to the suite.

lib/logger/lib/logger/config.ex
josevalim: We tested it here: https://github.com/elixir-lang/elixir/pull/8567#issuecomment-450852031

josevalim: @fishcakez there is no relevant  difference in the benchmarks linked above with the write_concurrency option compared to atomics.

lib/logger/lib/logger/config.ex
josevalim: I think I can test it by manually sending  a `handle_info`. Not nice but it should do it. 🤷‍♂️

fertapric:
According to the [Compatibility and Deprecations page](https://github.com/elixir-lang/elixir/blob/master/lib/elixir/pages/Compatibility%20and%20Deprecations.md#compatibility-between-elixir-and-erlangotp), Elixir v1.9 will support Erlang/OTP 20. I believe `:counters` was introduced in Erlang/OTP 21, should we update that page?

fishcakez:
> should we update that page? or should we use ets?

We are falling back to `ets` when `counters` is not available.

lib/logger/lib/logger/config.ex
fishcakez: Sounds good

josevalim:
Test added, I think this is ready to go!

lib/logger/lib/logger/config.ex
```diff
@@ -21,50 +19,74 @@ defmodule Logger.Config do
...
   def deleted_handlers(handlers) do
     :gen_event.call(Logger, @name, {:deleted_handlers, handlers})
   end
 
-  def new() do
-    tab = :ets.new(@table, [:named_table, :public, {:read_concurrency, true}])
-    true = :ets.insert_new(@table, [{:data, nil}, {:deleted_handlers, []}])
-    tab
+  def new do
+    {new_data(), new_counter()}
```
fertapric: Why are we passing the "data" here to always pattern match against `@table` later? (taking into account this is a private module).

josevalim: Yeah, we are just being assertive the correct data is being passed around. We could also just use it without checking. Either way is fine.

PR #8572, opened 2018-12-27
Deprecate atom-based supervisor child spec

josevalim:
This is for discussion.

The fact we support both `MyApp.Repo` and `{MyApp.Repo, arg}`, where `MyApp.Repo` is  the same as `{MyApp.Repo, []}`, have been historically confusing. For instance, folks when using `MyApp.Repo` believe `start_link/0` will be invoked. Other times, developers are just unaware of the relationship between both formats.

The argument for supporting the atom-based ones is readability. For example, in a Phoenix app:

    children = [
      MyApp.Repo,
      MyApp.Endpoint
    ]

However, having a unique format is more consistent without a large impact to readability:

    children = [
      {MyApp.Repo, []},
      {MyApp.Endpoint, []}
    ]

Finally, because Elixir v1.9 (the one coming in July/2019) will deprecate the old child specification, we want to make sure the current child specification is where we want it to be.

### DIFF
--- a/lib/elixir/lib/supervisor.ex
+++ b/lib/elixir/lib/supervisor.ex
@@ -223,15 +223,7 @@ defmodule Supervisor do
 
       use GenServer, restart: :transient
 
-  Finally, note it is also possible to simply pass the `Stack` module as
-  a child:
-
-      children = [
-        Stack
-      ]
-
-  When only the module name is given, it is equivalent to `{Stack, []}`.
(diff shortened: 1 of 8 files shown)


josevalim:
@yordis the description has been updated.

yordis:
🤦‍♂️ I am an idiot

josevalim:
For now we will continue improving the docs: 1ba9859e9e7afa898069092f2972556cadacb4e7

But the PR is here for documentation/discussion purposes. :+1:

PR #8573, opened 2018-12-27
Make KeyError.message/2 private

wojtekmach:
It was introduced in https://github.com/elixir-lang/elixir/pull/7803, looks like it was accidentally made public (no docs, no specs)

It shipped in Elixir v1.7 so technically this would be a breaking change. I'm happy to deprecate/doc false it instead.


eksperimental:
what is the protocol here? should this be backported to 1.7?

PR #8574, opened 2018-12-29
Added additional guards for Keywords.merge/3.

janderland:
Added guards to ensure only lists can be passed to `Keywords.merge/3`.

I thought about adding this change after realizing the `merge/3` method had inconsistent behavior...

```elixir
Keyword.merge([a: 1, b: 2], nil)
# ** (FunctionClauseError)

Keyword.merge([], nil)
# nil
```

What do you think?

### DIFF
--- a/lib/elixir/lib/keyword.ex
+++ b/lib/elixir/lib/keyword.ex
@@ -704,8 +704,8 @@ defmodule Keyword do
   @spec merge(t, t) :: t
   def merge(keywords1, keywords2)
 
-  def merge(keywords1, []), do: keywords1
-  def merge([], keywords2), do: keywords2
+  def merge(keywords1, []) when is_list(keywords1), do: keywords1
+  def merge([], keywords2) when is_list(keywords2), do: keywords2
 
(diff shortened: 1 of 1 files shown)


yordis:
@janderland it seems that it should always expect two Keywordlist instead of simply return one or the other one.

This is an odd behavior from API perspective to be honest.

eksperimental:
I agree with the changes, but if you want to be correct, the first two clauses should be removed all together, and let the third clause deal with it.

The current implementation and the proposed one, will not pass this test:

```iex
iex> Keyword.merge([:this, :is, :not, :a, :kw, :list], [])
[:this, :is, :not, :a, :kw, :list]
```

michalmuskala:
Those two clauses are an important optimisation for when merging with an empty keyword. In general, the `Keyword` module does not guarantee in only accepts valid keywords - it might accept other things as well, but it must accept valid keywords. It checks when it is practical or does not impose performance issues, but not otherwise.

josevalim:

PR #8575, opened 2018-12-29
Improve spec and documentation for System.build_info/0

eksperimental:
- Give more detailed spec
- Correct mention to keyword list when it is a map that it returns.
- Compile a list of keys returned and what each one is
- Add example


josevalim:

PR #8580, opened 2018-12-30
Module.__info__/1 docs and specs improvements

wojtekmach:


eksperimental:
I stumbled upon this situation yesterday. I was looking for information about @moduledoc and I  thought __info__ would have it, So a mention to `Code.fetch_docs/1` would be a big hint I think

josevalim:
@eksperimental I don't think we should mention `fetch_docs` on `__info__`. There are many things people could expect to be `__info__` and it is not its job to document what is not in there (unless it is a very common confusion). However, we could mention `fetch_docs` when we mention the doc related attributes in the `Module` module doc.

michalmuskala:
Thank you! :heart:

PR #8581, opened 2018-12-31
Improve documentation for Enum.take/2

eksperimental:
It was confusing to say in the summary that:
"Takes the first `amount` items from the `enumerable`."
when it can be from the end as well, depending on the sign of `amount`.


michalmuskala:
Thank you! :heart:

PR #8582, opened 2018-12-31
Improve specs for Enum.take/2, Enum.take_every/2, Enum.take_random/2

eksperimental:
It is specific about the behaviour when 2nd arg is 0,
as it always return [].


josevalim:
Those are overlapping domains. It won't work. :) Remember, specs do not work like function clauses. Or they should be completely distinct, otherwise they will get merged.

Issue #8590, opened 2019-01-03
EEx multiple callbacks with `-> %>…<% end` format


LostKobrakai:
### Environment

* Elixir & Erlang/OTP versions (elixir --version): 1.7.3 / 21
* Operating system: MacOS X

### Current behavior

```elixir
<%= component(ComponentsWeb.GridListView, [list: ["a", "b"]], fn assigns -> %>
  <div style="display: inline-block">
    <h1><%= @item %></h1>
  </div>
<% end, fn assigns -> %>
  <div style="display: inline-block">
    <h1><%= @item %></h1>
  </div>
<% end) %>
```

Having this in an eex template I get an syntax error: `** (SyntaxError) lib/components_web/templates/page/index.html.eex:5: unexpected token: end`

### Expected behavior

I'm not sure if this is purposefully not possible, but having the possiblity to define multiple `-> %>…<% end` callbacks for a single function call would enable the possibiltiy to build components with multiple slots to be filled.

LostKobrakai:
I've just found another case (notice the list around the callbacks):
```
<%= component(ComponentsWeb.GridListView, [list: ["a", "b"]], [fn assigns -> %>
  <div style="display: inline-block">
    <h1><%= @item %></h1>
  </div>
<% end, fn assigns -> %>
  <div style="display: inline-block">
    <h1><%= @item %></h1>
  </div>
<% end]) %>
```

I'd just do a PR handling trailing `]` like trailing `)`. This should be just fine, right?

PR #8591, opened 2019-01-03
Fix tokenize bug of #8590

LostKobrakai:


lib/eex/lib/eex/tokenizer.ex
```diff
@@ -123,24 +125,28 @@ defmodule EEx.Tokenizer do
...
+    if starts_with_end_expr?(rest) do
+      :middle_expr
+    else
+      # Tokenize the remaining passing check_terminators as
+      # false, which relax the tokenizer to not error on
+      # unmatched pairs. Then, we check if there is a "fn"
+      # token and, if so, it is not followed by an "end"
+      # token. If this is the case, we are on a start expr.
+      case :elixir_tokenizer.tokenize(rest, 1, file: "eex", check_terminators: false) do
+        {:ok, tokens} ->
```
josevalim: Hi @LostKobrakai!

I have cleaned up your commit a bit and then I noticed that your implementation actually has a bug. It will match if the code starts with a variable named `endgame`, for example.

I think that instead of looking at the buffer, you should look if the first token is the end token. It will be the same result, except you will rely on the Elixir tokenizer to handle it properly.

LostKobrakai: Great, I'll take a look at it. I wasn't really sure how the elixir_tokenizer behaves, so I tried to stay clear of it :D

LostKobrakai: I've gone and replace the check with looking at the results of the tokenizer and added the same improvements to tokens ending in `do %>`

lib/eex/lib/eex/tokenizer.ex
```diff
@@ -154,6 +158,13 @@ defmodule EEx.Tokenizer do
     :expr
   end
 
+  # Tokenize the remaining passing check_terminators as false,
+  # which relax the tokenizer to not error on unmatched pairs.
+  # If the tokens start with an "end" we have a middle expr.
```
fertapric: @josevalim I would remove the last comment line (`If the tokens start with an "end" we have a middle expr.`) or move it before the pattern matching (although it's a bit redundant).

josevalim: I would say it is still relevant info for the callers of this function. :)

josevalim:

PR #8594, opened 2019-01-04
Improve eex tokenizer further

LostKobrakai:
As further discussed here: https://github.com/elixir-lang/elixir/issues/8590#issuecomment-451407465

### DIFF
--- a/lib/eex/lib/eex/tokenizer.ex
+++ b/lib/eex/lib/eex/tokenizer.ex
@@ -10,6 +10,8 @@ defmodule EEx.Tokenizer do
 
   @spaces [?\s, ?\t]
 
+  @closing_brackets [?), ?], ?}]
+
   @doc """
   Tokenizes the given charlist or binary.
 
@@ -112,13 +114,14 @@ defmodule EEx.Tokenizer do
   #
   # Start tokens finish with "do" and "fn ->"
   # Middle tokens are marked with "->" or keywords
-  # End tokens contain only the end word and optionally ")"
+  # End tokens contain only the end word and optionally
+  # combinations of ")", "]" and "}".
 
-  defp token_name([h | t]) when h in [?\s, ?\t, ?)] do
+  defp token_name([h | t]) when h in @spaces or h in @closing_brackets do
     token_name(t)
   end
 
-  defp token_name('od' ++ [h | rest]) when h in [?\s, ?\t, ?)] do
+  defp token_name('od' ++ [h | rest]) when h in @spaces or h in [?)] do
     case tokenize_rest(rest) do
       {:ok, [{:end, _} | _]} -> :middle_expr
       _ -> :start_expr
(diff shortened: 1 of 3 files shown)


lib/eex/lib/eex/tokenizer.ex
```diff
@@ -112,13 +114,14 @@ defmodule EEx.Tokenizer do
...
+  # End tokens contain only the end word and optionally
+  # combinations of ")", "]" and "}".
 
-  defp token_name([h | t]) when h in [?\s, ?\t, ?)] do
+  defp token_name([h | t]) when h in @spaces or h in @closing_brackets do
     token_name(t)
   end
 
-  defp token_name('od' ++ [h | rest]) when h in [?\s, ?\t, ?)] do
+  defp token_name('od' ++ [h | rest]) when h in @spaces or h in [?)] do
```
josevalim: Shouldn't we consider closing brackets here?

josevalim: For instance, we could have this: `foo{bar}do`. In fact, we could even have this: `foo"bar"do`.

josevalim:
I will push some fixes too. :)

lib/eex/lib/eex/tokenizer.ex
fertapric: Should `;` also be considered in some cases of`check_spaces` to cover cases like: `something();end`?

josevalim:
Pushed! 

lib/eex/lib/eex/tokenizer.ex
josevalim: The existing code is confusing because we are removing terminators for both do and end but they should happen at distinct phases. This line is about `do`, so we don't have to worry about `;` since you can't have that before `do`.

My new code fixes exactly this. In any case, you are right, we should likely check for a `;` after `end`. I will push another commit.

josevalim: Actually, ignore me. EEx does not allow you to interpolate everywhere. For example, you can't interpolate it here:

    <%= foo( %>bar<% ) %>

Having a semicolon changes an expression from end expression to a middle expression and i don't think we should allow interpolating within `;` expressions.

fertapric: I was thinking on cases like this one:

```elixir
iex(1)> EEx.Tokenizer.tokenize('
...(1)>   <%= for a <- [1,2,3] do %>
...(1)>     foo
...(1)>   <% IO.puts "hello";end %>
...(1)> ', 1)
{:ok,
 [
   {:text, '\n  '},
   {:start_expr, 2, '=', ' for a <- [1,2,3] do '},
   {:text, '\n    foo\n  '},
   {:expr, 4, [], ' IO.puts "hello";end '},
   {:text, '\n'}
 ]}
```

PR #8595, opened 2019-01-04
Revamp Elixir CLI

josevalim:
This is a proposal to add five new options to Elixir's CLI with the goal of having releases into Elixir in the future.

In order to start an Elixir application in a release, there is a lot of work that has to be done by the underlying script: setting up the `HEART_COMMAND`, configuring boot files and boot vars, defining `run_erl` and making sure both pipe and log directories exist.

Because we want the scripts in an Elixir release to be minimal, we will move the complexity of managing those steps directly to the elixir/iex executables. Besides making releases more manageable, it means features like `heart` and `run_erl` become easily accessible to Elixir developers even outside of releases.

## The options

This PR adds 5 new options to elixir CLI:

  * `--pipe-to PIPEDIR LOGDIR` - invokes `run_erl` with daemon pipe and logs. This option is only for UNIX-like systems. Releases on Windows use a separate schema.
  * `--rpc-eval NODE COMMAND` - evaluates the given expression on the given node
  * `--boot FILE`, `--boot-var VAR VALUE` and `--vm-args FILE` - which are equivalent to Erlang's `-boot`, `-boot_var` and `-args_file`

By adding those options, we can build a script that would be part of a release and allows you to start, enable a remote console, perform rpcs (which includes stop, restart, pids, etc) in 70 LOC:

```bash
#!/bin/sh
set -e

SELF=$(readlink "$0" || true)
if [ -z "$SELF" ]; then SELF="$0"; fi
REL_ROOT="$(cd "$(dirname "$SELF")/.." && pwd -P)"
REL_NAME="demo"
REL_VSN="$(cut -d' ' -f2 "$REL_ROOT/releases/start_erl.data")"
export RELEASE_DIR="${RELEASE_DIR:-"$REL_ROOT/releases/$REL_VSN"}"
export COOKIE=${COOKIE:-$(cat "$REL_ROOT/releases/COOKIE")}

gen_id () {
  od -t x -N 4 /dev/urandom | head -n1 | awk '{print $2}'
}

rpc () {
  exec "$RELEASE_DIR/elixir" \
       --hidden --name "rpc-$(gen_id)@127.0.0.1" --cookie "$COOKIE" \
       --boot "${RELEASE_DIR}/start" \
       --boot-var RELEASE_ROOT "$REL_ROOT" \
       --rpc-eval "$REL_NAME@127.0.0.1" "$1"
}

start () {
  exec "$RELEASE_DIR/$1" --no-halt \
       --werl --cookie "$COOKIE" \
       --boot "${RELEASE_DIR}/${REL_NAME}" \
       --boot-var RELEASE_ROOT "$REL_ROOT" \
       --vm-args "${RELEASE_DIR}/vm.args" "${@:2}"
}

case $1 in
  start)
    start ${2:-elixir}
    ;;

  daemon)
    export RELEASE_TMP="${RELEASE_TMP:-"$REL_ROOT/tmp"}"
    start ${2:-elixir} --pipe-to "${RELEASE_TMP}/pipe" "${RELEASE_TMP}/log"
    ;;

  remote)
    exec "$RELEASE_DIR/iex" \
         --werl --hidden --name "remsh-$(gen_id)@127.0.0.1" --cookie "$COOKIE" \
         --boot "${RELEASE_DIR}/start" \
         --boot-var RELEASE_ROOT "$REL_ROOT" \
         --remsh "$REL_NAME@127.0.0.1"
    ;;

  rpc)
    if [ -z "$2" ]; then
      echo "BAD RPC" >&2
      exit 1
    fi
    rpc "$2"
    ;;

  restart|stop)
    rpc "System.$1"
    ;;

  pid)
    rpc "IO.puts System.pid"
    ;;

  *)
    echo "BAD COMMAND" >&2
    exit 1
    ;;
esac
```

This PR also reorganizes the CLI help instructions to make understanding all of those options slightly more manageable. It also trims down the `iex` CLI and makes it point to `elixir -h`. **I am looking for feedback on the new options as well as on the new help structure**.

## Summary

We are not quite sure yet how Elixir releases will look like but this is the first step. Given we have 6 months until the next Elixir release, we have plenty of time to remove or add any new options if necessary. In fact, we know we will need at least one extra option for Windows support, but that will be tackled in another PR.


josevalim:
I have implemented `--rpc-eval` and also added these features to Windows. This should be good to go (although I am waiting for CI on Windows too).

tsloughter:
Just to have it here, adding on what Fred said, aside from performance issues with `run_erl` it also doesn't seem to have much use today. Even if you aren't running in a container environment (like kubernetes or heroku where you output logs to stdout) you likely have systemd or similar for handling log rotation.

ConnorRigby:
> Do you ever use -detached / —detached flag?

Not that i know of. I'll let someone else answer for sure though.

> Are you using run_erl? run_erl requires you to have a mutable dir, which is where your logs go to, so what does Nerves do in this case?

Frank just told me about `run_erl` pretty recently. I don't think we are using it quite yet, but i at least would like to use it in the future. Logs would likely go to either `/root` (the read/write data partition) or to `/tmp`.

> Do you start the release using elixir or iex?

Nerves doesn't use Elixir at all to start a release _on_ the device itself. [erlinit](https://github.com/nerves-project/erlinit) starts the OTP release similarly to how one would manually start a distillery release. You can see _how_ it is started [here](https://github.com/nerves-project/erlinit/blob/master/src/erlinit.c#L642)

> Do you always connect to the device remotely ...

In Nerves we will have a UART console on the device itself, or a console on HDMI on devices that support that just like it were a regular machine. We also allow connecting via `ssh` and vie erlang distribution with [nerves_init_gadget](https://github.com/nerves-project/nerves_init_gadget). Both of those mechanisms work more or less the same as on say a cloud server. You connect with 
`ssh nerves.local` or `ssh 192.168.x.x` for ssh or with `iex --sname console --cookie democookie --remsh app@nerves.local`

josevalim:
> Nerves doesn't use Elixir at all to start a release on the device itself. erlinit starts the OTP release similarly to how one would manually start a distillery release. You can see how it is started here

So if I understand this correctly... you don't even use shell scripts to start it, which means that whatever we are doing here at the CLI level does not make a difference to Nerves. Is this correct?

Thanks @ConnorRigby!

ConnorRigby:
@josevalim Correct. Nerves starts an OTP release with `erlexec` on the device. On the host side (on your development PC for example), a Nerves app is started like any other Elixir application. 
ie `iex -S mix`

fhunleth:
I think Connor answered most of the questions, and the most important one, that we don't use the shell scripts. For completeness, here's info on the parts he skipped.

1. Does Nerves ever start `--detached`?

No. Nerves' erlinit only supports starting the OTP release in the foreground. It waits for exits or crashes to restart the device or take some other action.

2. run_erl

Very few people use run_erl with Nerves. Our console frequently runs at 115,200 baud and loses data, so we advise people to avoid using it for logging. I recently caught a NIF outputting useful debug data to the startup console and that's the only reason I use run_erl. run_erl logs are stored under /tmp which is writable on Nerves.

ferd:
> @ferd btw, was the fsync behaviour ever brought to discussion with the OTP team? Any chance we can put it behind a flag and/or disable it by default instead?

Never discussed it with them. Mostly the disterl solution was always available (including on windows) and faster to us so we just made the switch. Plus in some scenarios, we'd start before the root disk partition was mounted in write mode.

bin/elixir.bat
```diff
@@ -10,30 +10,50 @@ goto parseopts
...
+echo   --rpc-eval NODE "COMMAND"    Evaluates the given command on the given remote node (*)
+echo   --sname NAME                 Makes and assigns a short name to the distributed node
+echo.
+echo ## Release options
+echo.
+echo The following options are generally used under releases.
+echo.
+echo   --boot "FILE"                Uses the given FILE.boot to start the system
+echo   --boot-var VAR "VALUE"       Makes $VAR available as VALUE to FILE.boot (*)
+echo   --pipe-to "PIPEDIR" "LOGDIR" Starts the Erlang VM as a named PIPEDIR and LOGDIR
```
eksperimental:
```suggestion
echo   --pipe-to "PIPE_DIR" "LOG_DIR" Starts the Erlang VM as a named PIPE_DIR and LOG_DIR
```

bin/elixir
```diff
@@ -56,63 +83,111 @@ while [ $I -le $# ]; do
...
+        eval "RUN_ERL_PIPE=\${$I}"
+        if [[ "$RUN_ERL_PIPE" == "-"* ]]; then
+          echo "--pipe-to : PIPEDIR cannot be a switch" >&2
+          exit 1
+        fi
+
+        I=$(($I + 1))
+        eval "RUN_ERL_LOG=\${$I}"
+        if [[ "$RUN_ERL_LOG" == "-"* ]]; then
+          echo "--pipe-to : LOGDIR cannot be a switch" >&2
```
eksperimental:
```suggestion
          echo "--pipe-to : LOG_DIR cannot be a switch" >&2
```

bin/elixir
```diff
@@ -35,84 +63,152 @@ readlink_f () {
...
+          exit 1
+        fi
+        ;;
     --werl)
         USE_WERL=true
         ;;
```
eksperimental: All these options in this sections can be sorted alphabetically

bin/iex.bat
```diff
@@ -9,34 +9,13 @@ goto run
...
+echo   --remsh NAME        Connects to a node using a remote shell
 echo.
-echo   --dot-iex PATH              Overrides default .iex.exs file and uses path instead;
-echo                               path can be empty, then no file will be loaded
-echo   --remsh NAME                Connects to a node using a remote shell
-echo.
-echo ** Options marked with (*) can be given more than once
-echo ** Options given after the .exs file or -- are passed down to the executed code
-echo ** Options can be passed to the Erlang VM using ELIXIR_ERL_OPTIONS or --erl
+echo The remaining options are the same as in the elixir executable. See: elixir -h
```
eksperimental:
```suggestion
echo The remaining options are the same as in the "elixir" executable. Run "elixir --help" to see them
```

It's advisable to refer to the long switch, not the short cut in the documentation. `--help` instead of `-h`.

bin/iex
```diff
@@ -1,35 +1,14 @@
...
+  --remsh NAME        Connects to a node using a remote shell
 
-  --dot-iex PATH              Overrides default .iex.exs file and uses path instead;
-                              path can be empty, then no file will be loaded
-  --remsh NAME                Connects to a node using a remote shell
-
-** Options marked with (*) can be given more than once
-** Options given after the .exs file or -- are passed down to the executed code
-** Options can be passed to the VM using ELIXIR_ERL_OPTIONS or --erl" >&2
+The remaining options are the same as in the elixir executable. See: elixir -h" >&2
```
eksperimental:
```suggestion
The remaining options are the same as in the \"elixir\" executable. Run \"elixir --help\" to see them" >&2
```

bin/elixir
```diff
@@ -1,31 +1,59 @@
...
+
+--pipe-to starts Elixir detached from console (Unix-like only).
+It will attempt to create PIPEDIR and LOGDIR if they don't exist.
+See run_erl to learn more. To reattach, run: to_erl PIPEDIR.
+
+--pipe-to automatically sets \$HEART_COMMAND if none is set.
+See the -heart mode of the Erlang VM for more information.
+
+** Options marked with (*) can be given more than once." >&2
   exit 1
```
fertapric: Should we just `exit`? I think at least `elixir --help`, or `elixir -h` should return 0.

josevalim: I believe it was because if Elixir is accidentally called without argument or something of sorts.  So we probably need to check why we are invoking this. Let's do that in a separate occasion.

bin/elixir
josevalim: Erlang refers to those and other environment variables without snake case so I am going with their convention for now.

PR #8596, opened 2019-01-05
Respect `subdirectories` config on mix format warn

vnegrisolo:
For context: the related warning message was introduced by this PR: https://github.com/elixir-lang/elixir/pull/8323
For reference: https://hexdocs.pm/mix/master/Mix.Tasks.Format.html#module-formatting-options

There is a mix format option called `subdirectories` that should be considered to avoid unnecessary warning message as this option defines folders to be ignored.

Check this out:

```shell
$ cat .formatter.exs
[
  import_deps: [:ecto, :phoenix],
  inputs: ["*.{ex,exs}", "{config,lib,priv,test}/**/*.{ex,exs}"],
  subdirectories: ["priv/*/migrations"]
]
```

```shell
$ cat priv/repo/migrations/.formatter.exs
[
  import_deps: [:ecto_sql],
  inputs: ["*.exs"]
]
```

And here is the message that we can avoid as the config seems consistent with the documentation:

```shell
$ mix format
Both .formatter.exs and priv/repo/migrations/.formatter.exs specify the file priv/repo/migrations/.formatter.exs in their :inputs option. To resolve the conflict, the configuration in .formatter.exs will be ignored. Please change the list of :inputs in one of the formatter files so only one of them matches priv/repo/migrations/.formatter.exs
Both .formatter.exs and priv/repo/migrations/.formatter.exs specify the file priv/repo/migrations/20181105133308_create_users.exs in their :inputs option. To resolve the conflict, the configuration in .formatter.exs will be ignored. Please change the list of :inputs in one of the formatter files so only one of them matches priv/repo/migrations/20181105133308_create_users.exs
```

### DIFF
--- a/lib/mix/lib/mix/tasks/format.ex
+++ b/lib/mix/lib/mix/tasks/format.ex
@@ -358,9 +358,23 @@ defmodule Mix.Tasks.Format do
       Mix.raise("Expected :inputs or :subdirectories key in #{dot_formatter}")
     end
 
+    subdirectories =
+      formatter_opts[:subdirectories]
+      |> List.wrap()
+      |> Enum.map(&Path.join(prefix ++ [&1]))
+      |> Enum.map(&Path.wildcard(&1))
+
+    files =
+      formatter_opts[:inputs]
+      |> List.wrap()
+      |> Enum.map(&Path.join(prefix ++ [&1]))
+      |> Enum.flat_map(&Path.wildcard(&1, match_dot: true))
+      |> Enum.reject(fn file ->
+        Enum.any?(subdirectories, &String.starts_with?(file, &1))
+      end)
+
     map =
-      for input <- List.wrap(formatter_opts[:inputs]),
-          file <- Path.wildcard(Path.join(prefix ++ [input]), match_dot: true),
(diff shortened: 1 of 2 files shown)


vnegrisolo:
By the way, this is the new entry in this project's [CHANGELOG](https://github.com/elixir-lang/elixir/blob/v1.8/CHANGELOG.md) and I think that is still valid, so it may not be needed to add anything new I guess.

> [mix format] Warn when the same file may be formatted by multiple .formatter.exs

josevalim:
Hi @vnegrisolo ! That's *exactly* the case we want to warn about. Elixir only reads a single `.formatter.exs`, so if there are any conflicts, it is exactly because of child dependencies. You can fix this by removing `priv` from your inputs. Here is what Phoenix does:

https://github.com/phoenixframework/phoenix/blob/master/installer/templates/phx_single/formatter.exs

I will add it to the CHANGELOG. Good catch!

vnegrisolo:
Hi @josevalim thanks for the response. I though that the way to remove was adding a folder into the `subdirectories`. Thanks for the link, very useful!

Issue #8507, opened 2018-12-11
ExUnit.CaptureLog.remove_capture fails with error


samphilipd:
### Environment

```
Erlang/OTP 21 [erts-10.0.7] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1] [hipe]

Elixir 1.7.3 (compiled with Erlang/OTP 21)
```

### Current behavior

I changed this line:

`ExUnit.start()`

to 

`ExUnit.start(capture_log: true)`

in order to silence log output during tests. However, some tests are now failing with this error message:

```
     ** (EXIT from #PID<0.1226.0>) exited in: ExUnit.CaptureLog.remove_capture(#PID<0.1241.0>)
         ** (EXIT) {:error, :module_not_found}
```

Any ideas?

josevalim:
So I cannot reproduce this. Here is a snippet that defines 100 test modules with 100 tests each:

```elixir
require Logger
ExUnit.start(capture_log: true)

for i <- 1..100 do
  defmodule Module.concat(["Test#{i}"]) do
    use ExUnit.Case, async: :rand.uniform > 0.5

    for i <- 1..100 do
      test :"#{i}" do
        Logger.debug "ok"
      end
    end
  end
end
```

It passes just fine.

Looking at the code, the only situation I could see this happening is if the Logger supervisor restarts, maybe because a custom backend is failing? Are you using any custom backends? Does removing them reflect the failures?

samphilipd:
We do not use any custom logger backends. Our setup looks like this:

```
config :logger, :console, metadata: :all
config :logger, :console, level: :error
```

samphilipd:
@josevalim I'm not really sure what else I can give you. The failures are intermittent - sometimes a test will fail and sometimes it will pass. But it is always the same tests that flicker.

Is there anything else I can share with you that will help us to debug this?

josevalim:
Maybe one option is for you to add a file logger backend and then see if the file backend has logged any useful information because the issue in cases like this is that something else may have been logged but we wouldn't know because there is no logger backend.

fertapric:
I managed to reproduce it:

```elixir
require Logger

Application.put_env(:logger, :console, [
  format: "\n$time $metadata[$level] $levelpad$message\n",
  metadata: :all
])

ExUnit.start(capture_log: true)

defmodule Module.CaptureLogError do
  use ExUnit.Case

  test "boom!" do
[...]

PR #8600, opened 2019-01-06
Handle non-printable arguments in StringIO gracefully

fertapric:
Related to #8507 

Non-printable arguments were killing the StringIO process, which was killing the Logger handler associated to it. That was causing some errors when users were running their test suites with `capture_log: true`, because the proxy (logger handler) was not there when `ExUnit.CaptureLog` was trying to remove it.


josevalim:

PR #8460, opened 2018-12-04
Add week_of_year/3

josevalim:

PR #8601, opened 2019-01-07
Provide better errors with invalid dates in Calendar.ISO

fertapric:
As discussed in https://github.com/elixir-lang/elixir/pull/8460/files#r238752220

Provides more information than the previous `MatchError`:

```
** (MatchError) no match of right hand side value: false
```

### DIFF
--- a/lib/elixir/lib/calendar/iso.ex
+++ b/lib/elixir/lib/calendar/iso.ex
@@ -213,7 +213,7 @@ defmodule Calendar.ISO do
   end
 
   def date_to_iso_days(year, month, day) when year in -9999..9999 do
-    ensure_day_in_month(year, month, day)
+    ensure_day_in_month!(year, month, day)
 
     days_in_previous_years(year) + days_before_month(month) + leap_day_offset(year, month) + day -
       1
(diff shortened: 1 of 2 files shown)


lib/elixir/lib/calendar/iso.ex
```diff
@@ -366,7 +366,7 @@ defmodule Calendar.ISO do
   @impl true
   def day_of_year(year, month, day)
       when is_integer(year) and is_integer(month) and is_integer(day) do
-    true = day <= days_in_month(year, month)
+    if day > days_in_month(year, month), do: raise(ArgumentError, "invalid date")
```
josevalim: Should we try to include the date in the error message?

Also, we should probably move this to a private function.

josevalim: We can emulate it with `invalid date: ~D[day: #{day}, month: #{month}, year: #{year}]`

fertapric: I think that mix of `~D` sigil and struct might be confusing, what do you think about?

```elixir
raise ArgumentError, "invalid date: #{date_to_string(year, month, day)}"
#=> "invalid date: 2000-12-53"
```

lib/elixir/lib/calendar/iso.ex
```diff
@@ -213,7 +213,7 @@ defmodule Calendar.ISO do
   end
 
   def date_to_iso_days(year, month, day) when year in -9999..9999 do
-    true = day <= days_in_month(year, month)
+    ensure_day_in_month(year, month, day)
```
fertapric: Should it be `ensure_day_in_month!` (with bang)? I'm not sure what's the convention when there is only the "raise" version.

josevalim: Yup, with bang!

josevalim:

Issue #8604, opened 2019-01-08
Add File.rename!


josevalim:
All other functions in File have bang variants, this one should have to.

eksperimental:
I can take this one, if nobody else ia working on it already

PR #8606, opened 2019-01-08
Add File.rename!/2

Nitrino:
Add rename function with a bang, according to #8604

### DIFF
--- a/lib/elixir/lib/file.ex
+++ b/lib/elixir/lib/file.ex
@@ -731,6 +731,26 @@ defmodule File do
   end
 
   @doc """
+  The same as `rename/2` but raises a `File.RenameError` exception if it fails.
+  Returns `:ok` otherwise.
+  """
+
+  @spec rename!(Path.t(), Path.t()) :: :ok | {:error, posix}
+  def rename!(source, destination) do
+    case rename(source, destination) do
+      :ok ->
+        :ok
+
+      {:error, reason} ->
+        raise File.RenameError,
+          reason: reason,
+          action: "rename",
+          source: maybe_to_string(source),
+          destination: maybe_to_string(destination)
+    end
+  end
+
+  @doc """
   Copies the contents in `source` to `destination` preserving its mode.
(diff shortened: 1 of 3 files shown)


lib/elixir/lib/file.ex
```diff
@@ -730,6 +730,26 @@ defmodule File do
     :file.rename(source, destination)
   end
 
+  @doc """
+  The same as `rename/2` but raises a `File.RenameError` exception if it fails.
+  Returns `:ok` otherwise.
+  """
+
```
fertapric:
```suggestion
  @doc since: "1.9.0"
```

Nitrino: Thanks!

lib/elixir/lib/file.ex
```diff
@@ -730,6 +730,26 @@ defmodule File do
...
+  def rename!(source, destination) do
+    case rename(source, destination) do
+      :ok ->
+        :ok
+
+      {:error, reason} ->
+        raise File.RenameError,
+          reason: reason,
+          action: "rename",
+          source: maybe_to_string(source),
```
fertapric: I think this can use `IO.chardata_to_string/1` since the function only accepts `Path.t()`

Nitrino: Yes, it will be better. 
Fixed.

lib/elixir/lib/file.ex
```diff
@@ -730,6 +730,26 @@ defmodule File do
     :file.rename(source, destination)
   end
 
+  @doc """
+  The same as `rename/2` but raises a `File.RenameError` exception if it fails.
+  Returns `:ok` otherwise.
+  """
+  @doc since: "1.9.0"
+  @spec rename!(Path.t(), Path.t()) :: :ok | {:error, posix}
```
fertapric:
```suggestion
  @spec rename!(Path.t(), Path.t()) :: :ok
```

PR #8608, opened 2019-01-08
.gitignore improvements

Virviil:
## The problem

I've found some kind of growing trend of increasing artifact files in Elixir's repositories [(for example here)](https://github.com/imeraj/Phoenix_Playground/tree/master/1.4/rumbl)  and even Hex packages [(here in repo also)](https://hex.pm/packages/asura) that are more or less used and views by the community.

While this kind of a "problem" is a result of carelessness and haste, it seems that automation can help avoiding this kind of problem.

## Solution

The solution seems to be more or less cleared: `.gitignore` file, that is generated during calling `mix new` task can be expanded.

## Possible disadvantages

It's definitely impossible to predict every possible artifact, that can appear in someone's code.
Thus, every line in ".gitignore" should be kindly discussed by the community.

## Proposal
I'm trying to start the discussion with **PR** instrument in order to concretize the problem and directly point out the place of possible code modifications. I will appreciate any comments here, with any possible additions to .gitignore, some arguments to extend it or not to do this.

I'm definitely don't insist on concrete modifications, but I believe that **.gitignore** modification can improve development experience for Elixir developers.

### DIFF
--- a/lib/mix/lib/mix/tasks/new.ex
+++ b/lib/mix/lib/mix/tasks/new.ex
@@ -273,6 +273,9 @@ defmodule Mix.Tasks.New do
   # The directory Mix will write compiled artifacts to.
   /_build/
 
+  # The directory for Elixir Language Server temporary files
+  **/.elixir_ls
+
   # If you run "mix test --cover", coverage assets end up here.
   /cover/
 


josevalim:
@Virviil editor information should not be in the project gitignore but it should be in the user local gitignore. Imagine how big Elixir's .gitignore would be if we had to include editor information from everyone who already contributed to Elixir.

However, I do understand your concern regarding artifacts.

My suggestion, in this particular case, would be to open up an issue in `elixir_ls` itself and have them suggest users from time to time how to configure their local .gitignore so `.elixir_ls` is not listed as a dependency.

Another suggestion is for `elixir_ls` to put its contents inside `_build`, as that is git ignored by default.

Virviil:
@josevalim Yep, you are seems so to be right. Probably here is not the best place to try to resolve the problem.

I'll try to create the discussion with the same prerequisite in Elixir Forum - and then, create a community guide for gitignore improvements.

PR #8613, opened 2019-01-11
Allow maps with required fields for to_date and to_time functions.

lau:
`DateTime`/`NaiveDateTime.to_date` functions do not need any information
about time to convert to a date. Just a map. The same goes for `to_time`
and date or time zone information.

This allows for instance passing in a list of mixed data: e.g. both
Date and DateTime to the same function.

The functions do the same thing already except for NaiveDateTime "artificially" requiring a NaiveDateTime struct. The other functions in the module only require a map instead of a struct. So only requiring a map with be more consistent in that respect.

It would be cool to have in 1.8. This way we can say that all of the functions just require a map and does not arbitrarily require certain structs for certain functions.

### DIFF
--- a/lib/elixir/lib/calendar/datetime.ex
+++ b/lib/elixir/lib/calendar/datetime.ex
@@ -583,6 +583,9 @@ defmodule DateTime do
   Because `Date` does not hold time nor time zone information,
   data will be lost during the conversion.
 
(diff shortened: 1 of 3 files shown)


lau:
Closed in favor of https://github.com/elixir-lang/elixir/pull/8615

PR #8615, opened 2019-01-11
Allow maps with required fields for to_date and to_time functions.

lau:
`DateTime`/`NaiveDateTime.to_date` functions do not need any information
about time to convert to a date. Just a map. The same goes for `to_time`
and date or time zone information.

This allows for instance passing in a list of mixed data: e.g. both
Date and DateTime to the same function.

The functions do the same thing already except for NaiveDateTime "artificially" requiring a NaiveDateTime struct. The other functions in the module only require a map instead of a struct. So only requiring a map with be more consistent in that respect.

It would be cool to have in 1.8. This way we can say that all of the functions just require a map and does not arbitrarily require certain structs for certain functions.

### DIFF
--- a/lib/elixir/lib/calendar/datetime.ex
+++ b/lib/elixir/lib/calendar/datetime.ex
@@ -583,6 +583,9 @@ defmodule DateTime do
   Because `Date` does not hold time nor time zone information,
   data will be lost during the conversion.
 
+  In addition to a `DateTime` any map containing the required fields can be
+  passed to this function. This includes `Date` or a `NaiveDateTime`.
+
   ## Examples
 
       iex> dt = %DateTime{year: 2000, month: 2, day: 29, zone_abbr: "CET",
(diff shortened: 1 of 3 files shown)


josevalim:
The reason why we didn’t do it using only the required fields is because we
lose an important aspect which is writing assertive code. When I call
NaiveDateTime.to_date today, I want the guarantee that we have something
that looks like a NaiveDateTime. We should always downcast, never upcast.

If we want something that works with anything that has Date fields, then we
should probably add a new function to Date. Date.to_date is redundant....
maybe Date.new/1?
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

lau:
> The reason why we didn’t do it using only the required fields is because we
lose an important aspect which is writing assertive code. When I call
NaiveDateTime.to_date today, I want the guarantee that we have something
that looks like a NaiveDateTime. We should always downcast, never upcast.

That makes sense. Then for consistency maybe `DateTime.to_date` should require a time_zone field etc. even if it is ignored?

> If we want something that works with anything that has Date fields, then we
should probably add a new function to Date. Date.to_date is redundant....
maybe Date.new/1?

Yes I think it makes sense to have such a function in the Date module. So that if you have a mix of data that could be NaiveDateTime, Date or DateTime to be downcast to a Date they can all be passed there. I think of it as a `from_map` functions. But adding the functionality to `Date.new` makes sense. And the same for the other modules. At least Date.new, Time.new, NaiveDateTime.new should be simple enough.
It also makes sense to me to have the validation that the `new` functions have if people pass "home made" maps instead of the structs created in the standard lib.

josevalim:
> That makes sense. Then for consistency maybe DateTime.to_date should
require a time_zone field etc. even if it is ignored?

Definitely.
-- 


josevalim:
Closing this according to feedback. I will push a fix soon that clarifies DateTime and NaiveDateTime signatures since I consider the current behaviour a bug. :)

However, a PR that adds `from_map` to all four modules will be very welcome! I think `from_map` you mentioned is better than overloading `new`. :+1:

Issue #8611, opened 2019-01-11
PIDPartition grows on :already_registered errors from Registry.register


fhunleth:
### Environment

* Elixir & Erlang/OTP versions (elixir --version): Elixir 1.7.4, Erlang 21
* Operating system: Linux

### Current behavior

If I run the following and watch the size of the ETS UniqueTest.PIDPartition0 table, it'll grow a row on every call to `register` even though all but the first one return an error.
```
iex(3)> Registry.start_link(keys: :unique, name: UniqueTest)
{:ok, #PID<0.128.0>}
iex(4)> Registry.register(UniqueTest, "hello", [])
{:ok, #PID<0.129.0>}
iex(5)> Registry.register(UniqueTest, "hello", [])
{:error, {:already_registered, #PID<0.107.0>}}
iex(6)> Registry.register(UniqueTest, "hello", [])
{:error, {:already_registered, #PID<0.107.0>}}
iex(7)> Registry.register(UniqueTest, "hello", [])
{:error, {:already_registered, #PID<0.107.0>}}
iex(8)> :ets.i()
 id              name              type  size   mem      owner
 ----------------------------------------------------------------------------
 #Ref<0.3405087786.1607073797.255116> 'Elixir.UniqueTest.KeyPartition0' set   1      1340     'Elixir.UniqueTest.PIDPartition0'
 #Ref<0.3405087786.1607073797.255117> 'Elixir.UniqueTest.PIDPartition0' duplicate_bag 4      1392     'Elixir.UniqueTest.PIDPartition0'
```

It looks like this behavior is intentional since the `:already_registered` error case is called out here: https://github.com/elixir-lang/elixir/blob/v1.7.4/lib/elixir/lib/registry.ex#L895. 

I'd like to keep PIDPartition from growing since it got quite large on a long running system. The code that blindly calls `Registry.register` is unfortunately not too smart and it's hard to make it know beforehand whether it has already registered a key or not. Calling `Registry.unregister` on `:already_registered` errors cleans up too much so I can't do that.

### Expected behavior

The ETS PIDPartition doesn't grow unbounded on `:already_registered` errors.

Sadly, simply adding `:ets.delete_object` on this error case didn't work since it deletes all matching rows (the table is a duplicate_bag).

josevalim:
[...]

PR #8616, opened 2019-01-12
Fix ETS leak in Registry.register (#8611)

fhunleth:
This fixes an issue where a registry's PIDPartition table would gain an
entry each time after the first call to Registry.register on a unique
registry.

The handler for :already_registered errors did not clean up the entry
that was initially added to the PIDPartition table. The fix is to delete
the new entry. Since the ets table is a duplicate_bag, it's not possible
to delete only the one entry. However, since the registry is :unique,
there was exactly one entry before the call so the code deletes all
matching entries and adds one back.

### DIFF
--- a/lib/elixir/lib/registry.ex
+++ b/lib/elixir/lib/registry.ex
@@ -922,6 +922,11 @@ defmodule Registry do
         ok
 
       {:error, {:already_registered, ^self}} = error ->
+        # This is an error for the :unique case, so there are two entries.
+        # :ets.delete_object deletes both since pid_ets is a :duplicate_bag so
+        # add one back.
+        true = :ets.delete_object(pid_ets, {self, key, key_ets})
+        true = :ets.insert(pid_ets, {self, key, key_ets})
         error
 
       {:error, _} = error ->
--- a/lib/elixir/test/elixir/registry_test.exs
+++ b/lib/elixir/test/elixir/registry_test.exs
@@ -261,6 +261,41 @@ defmodule RegistryTest do
         {:ok, pid} = Agent.start_link(fn -> 0 end, name: name)
         assert Registry.lookup(registry, "hello") == [{pid, :value}]
       end
+
+      test "doesn't grow ets on already_registered",
+           %{registry: registry, partitions: partitions} do
+        assert sum_pid_entries(registry, partitions) == 0
+
+        {:ok, pid} = Registry.register(registry, "hello", :value)
+        assert is_pid(pid)
+        assert sum_pid_entries(registry, partitions) == 1
+
+        {:ok, pid} = Registry.register(registry, "world", :value)
+        assert is_pid(pid)
+        assert sum_pid_entries(registry, partitions) == 2
+
+        assert {:error, {:already_registered, pid}} = Registry.register(registry, "hello", :value)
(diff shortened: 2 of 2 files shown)


lib/elixir/lib/registry.ex
```diff
@@ -922,6 +922,11 @@ defmodule Registry do
         ok
 
       {:error, {:already_registered, ^self}} = error ->
+        # This is an error for the :unique case, so there are two entries.
+        # :ets.delete_object deletes both since pid_ets is a :duplicate_bag so
+        # add one back.
+        true = :ets.delete_object(pid_ets, {self, key, key_ets})
```
josevalim: Unfortunately this is a race condition. Because we have a key registered under `self()`, if this process happens to crash right after deleting here due to a link, its key will never be removed. I believe the proper fix is to do what @michalmuskala suggested and replace [this line](https://github.com/elixir-lang/elixir/blob/77e36f5b98d33cc2d1cbb02fbda572576cb84bce/lib/elixir/lib/registry.ex#L914) by `insert_new` and return `{:error, {:already_registered, self()}}` straight away.

fhunleth:
Update: I've push some minor fixes to the unit tests, but nothing to fix the handling of the core issue. I started down the counter path as described in the #8611, but that needs modifications in a few other places an is failing quite a few tests. Will pick up on this tomorrow.

lib/elixir/lib/registry.ex
adkron: I tried to use `insert_new` to create a cleanup callback that would only delete when `insert_new` returned true. I ran into a lot of failing tests and didn't get a chance to jump back to it. 

@fhunleth, did you have failures like:

~~~bash
 23) test unique with 8 partitions allows process unregistering (RegistryTest)
     test/elixir/registry_test.exs:184
     Assertion with == failed
     code:  assert Registry.keys(registry, self()) |> Enum.sort() == ["hello", "world"]
     left:  ["hello"]
     right: ["hello", "world"]
     stacktrace:
       test/elixir/registry_test.exs:189: (test)

.............

 24) test duplicate with 1 partition compares using matches (RegistryTest)
     test/elixir/registry_test.exs:300
     Assertion with == failed
     code:  assert Registry.keys(registry, self()) |> Enum.sort() == [1, 1.0]
     left:  [1.0]
     right: [1, 1.0]
     stacktrace:
       test/elixir/registry_test.exs:303: (test)
~~~

That is just the last two because I thought putting them all in here would be a terrible plan.

fhunleth: Yes, I ran into that. There's some more conversation on the linked issue. The current plan of attack is to add a unique number to the tuple that gets put in ETS so that it can be deleted on fail (José's original idea). I got waylaid by work or I'd be continuing down that path. It needs a bunch of updates to other code to make it work, though.

josevalim:
Ping. :)

fhunleth:
Hey - so sorry for not getting back on this. I ended up removing the calls that caused the leak since I was in a crunch. I'm too busy this week and next to make progress. Feel free to close and then when I get more time, I can re-open.

josevalim:
Thanks @fhunleth! I will close this for now then, in case someone else wants to tackle it. Feel free too reopen or repush any time!

PR #8621, opened 2019-01-14
Default to UTC in DateTime.from_naive,from_naive!

wojtekmach:
This is especially useful in tests where all we have to do now is write:

```elixir
DateTime.from_naive!(~N[2019-01-01 00:00:00])
```

Of course all we're doing here is typing less characters than explicitly setting UTC, so I think this PR is more of an opening to a discussion if making this more convenient is worth pursuing. If this is annoying to people, I believe a suggestion at some point was to have users write their own wrappers for this function (e.g. a custom ~Z sigil)

### DIFF
--- a/lib/elixir/lib/calendar/datetime.ex
+++ b/lib/elixir/lib/calendar/datetime.ex
@@ -247,7 +247,7 @@ defmodule DateTime do
 
   def from_naive(
         naive_datetime,
-        time_zone,
+        time_zone \\ "Etc/UTC",
         time_zone_database \\ Calendar.get_time_zone_database()
       )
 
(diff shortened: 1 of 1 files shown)


josevalim:
The counter-argument is that timezone is essential to DateTime, so hiding it is not very helpful.

/cc @lau

wojtekmach:
@josevalim good point.

Btw, I was looking at other functions that might take this default and I thought there were none, however I missed the new `DateTime.now/1,2`. So, if this PR is accepted, should there be a `DateTime.now/0` with the same default? That's less desirable as there's already a `DateTime.utc_now`. I'll keep this open for now but I gotta say I'm a bit less convinced now myself :)

lau:
The time zone is very important. I think default data can be dangerous and lead to insidious bugs. What if you are writing code and want to create DateTimes and you accidentally forget to pass the time zone? Then it defaults to UTC instead of the time zone it was supposed to be. This can lead to hidden bugs.

I wrote about something related here: http://www.creativedeletion.com/2019/01/08/data-types-assumptions.html

wojtekmach:
Thanks for feedback @lau!

PR #8624, opened 2019-01-15
V1.8

azados:
Hello hello! Congratulations with new version. Thank you guys for doing brilliant work and contribution.

I'm very new to GitHub and I came for Elixir here.

I wonder if Elixir can migrate (V2) from GitHub (Microsoft) to GitLab? There are many reasons not to be parented by MS Corp, also GitHub use in their status only blonde emojis. 

Exclusion? Also time for new trends.

Privet from Tebet!


josevalim:
Thanks @azados. I recommend moving this discussion to the Elixir mailing list, as  we focus the issues tracker on bugs.

PR #8625, opened 2019-01-15
Use consistent variable naming in DynamicSupervisor docs

derekkraan:
I saw this while browsing the DynamicSupervisor code and was reminded of my previous PR #8226

### DIFF
--- a/lib/elixir/lib/dynamic_supervisor.ex
+++ b/lib/elixir/lib/dynamic_supervisor.ex
@@ -47,12 +47,12 @@ defmodule DynamicSupervisor do
         # Automatically defines child_spec/1
         use DynamicSupervisor
 
-        def start_link(arg) do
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/dynamic_supervisor.ex
```diff
@@ -78,20 +78,20 @@ defmodule DynamicSupervisor do
       defmodule MySupervisor do
         use Supervisor
 
-        def start_link(arg) do
-          Supervisor.start_link(__MODULE__, arg, name: __MODULE__)
+        def start_link(extra_arg) do
```
josevalim: Why  are we not using  `init_arg` here too? 🤔

derekkraan: Check, I pushed an updated commit.

josevalim:

PR #8627, opened 2019-01-15
Make documentation accurate in File.cp* functions

eksperimental:


lib/elixir/lib/file.ex
```diff
@@ -886,8 +894,6 @@ defmodule File do
     end
   end
 
-  # src may be a file or a directory, dest is definitely
-  # a directory. Returns nil unless an error is found.
```
eksperimental: This comment is wrong, either it belongs to a different function, or it is outdated

lib/elixir/lib/file.ex
```diff
@@ -791,35 +795,38 @@ defmodule File do
...
+  `destination`. If the `source` is a directory, it copies
+  the contents inside source into the `destination` directory.
 
   If a file already exists in the destination, it invokes `callback`.
   `callback` must be a function that takes two arguments: `source` and `destination`.
   The callback should return `true` if the existing file should be overwritten and `false` otherwise.
 
-  If a directory already exists in the destination
-  where a file is meant to be (or vice versa), this
-  function will fail.
```
eksperimental: I'm still not able to understand what the "vice versa" translates back to plain simple English,
but in any case the information about the directory existing in the destination and the function failing is wrong.

lib/elixir/lib/file.ex
```diff
@@ -791,35 +795,38 @@ defmodule File do
...
-          destination: IO.chardata_to_string(destination)
+          source: IO.chardata_to_string(source_file),
+          destination: IO.chardata_to_string(destination_file)
     end
   end
 
   @doc ~S"""
-  Copies the contents in source to destination.
+  Copies the contents in `source` to `destination` recursively, maintaining the
+  source directory structure and modes.
```
eksperimental: Here I talk about "modes" which was omitted before, but since this function relies on `do_cp_file/4` which is the same function `cp/3` relies on, it is assumed to keep modes, with only one mention bellow (which is a bug IMO)

lib/elixir/lib/file.ex
```diff
@@ -712,8 +712,8 @@ defmodule File do
 
   Returns `:ok` in case of success, `{:error, reason}` otherwise.
 
-  Note: The command `mv` in Unix systems behaves differently depending
-  if `source` is a file and the `destination` is an existing directory.
+  Note: The command `mv` in Unix systems behaves differently depending on
+  whether `source` is a file and the `destination` is an existing directory.
   We have chosen to explicitly disallow this behaviour.
```
josevalim: We can probably get rid of this note altogether given the discussion on the forum. Up to you if you prefer to do it later though.

lib/elixir/lib/file.ex
josevalim: If there is a file in the source and a directory in the destination OR if there is a directory in the source and a file in the destination, it fails.

lib/elixir/lib/file.ex
eksperimental: I will make it part of a new PR

lib/elixir/lib/file.ex
whatyouhide: Personally it feels like this is clear enough (not a native speaker). Should we put it back?

lib/elixir/lib/file.ex
```diff
@@ -751,30 +751,34 @@ defmodule File do
...
   If you want to copy contents from an IO device to another device
   or do a straight copy from a source to a destination without
   preserving modes, check `copy/3` instead.
 
-  Note: The command `cp` in Unix systems behaves differently depending
-  if `destination` is an existing directory or not. We have chosen to
-  explicitly disallow this behaviour. If destination is a directory, an
-  error will be returned.
+  Note: The command `cp` in Unix systems behaves differently depending on
+  wether destination is an existing directory or not. We have chosen to
```
whatyouhide:
```suggestion
  whether destination is an existing directory or not. We have chosen to
```

lib/elixir/lib/file.ex
josevalim: We can expand it without the vice-versa. Being precise doesn't hurt. :)

lib/elixir/lib/file.ex
```diff
@@ -791,35 +795,38 @@ defmodule File do
...
     end
   end
 
   @doc ~S"""
-  Copies the contents in source to destination.
+  Copies the contents in `source` to `destination` recursively, maintaining the
+  source directory structure and modes.
+
+  If `source` is a file or a symbolic link to it, `destination` must be a path
+  to an existent file, a sysmbolic link to one, or a path to a non-existent file.
```
whatyouhide:
```suggestion
  to an existent file, a symbolic link to one, or a path to a non-existent file.
```

lib/elixir/lib/file.ex
eksperimental: It think we can still remove it, because it is covered by the two opening paragraphs

> 
>   If `source` is a file or a symbolic link to it, `destination` must be a path
>   to an existent file, a symbolic link to one, or a path to a non-existent file.
> 
>   If `source` is a directory, or a symbolic link to it, then `destination` must
>   be an existent `directory` or a symbolic link to one, or a path to a non-existent directory.
>

lib/elixir/lib/file.ex
```diff
@@ -886,8 +894,6 @@ defmodule File do
     end
   end
 
-  # src may be a file or a directory, dest is definitely
-  # a directory. Returns nil unless an error is found.
   defp do_cp_r(src, dest, callback, acc) when is_list(acc) do
```
eksperimental: In this function, we we create a directory, because source has it,
https://github.com/elixir-lang/elixir/pull/8627/files#diff-65687140d0fd6e1c1505681b6fa5f679R911

we are omitting the modes.
I'm also investigating about symbolic link modes, because Macn and Linux behave differently, but all these changes can be part of another Pr

lib/elixir/lib/file.ex
```diff
@@ -751,30 +751,34 @@ defmodule File do
...
   If you want to copy contents from an IO device to another device
   or do a straight copy from a source to a destination without
   preserving modes, check `copy/3` instead.
 
-  Note: The command `cp` in Unix systems behaves differently depending
-  if `destination` is an existing directory or not. We have chosen to
-  explicitly disallow this behaviour. If destination is a directory, an
-  error will be returned.
+  Note: The command `cp` in Unix systems behaves differently depending on
+  whether destination is an existing directory or not. We have chosen to
```
sneako:
```suggestion
  whether the destination is an existing directory or not. We have chosen to
```

eksperimental: Thank you!

lib/elixir/lib/file.ex
```diff
@@ -829,20 +836,21 @@ defmodule File do
...
+  is a directory, `{:error, :eisdir}` will be returned.
 
   ## Examples
 
       # Copies file "a.txt" to "b.txt"
       File.cp_r("a.txt", "b.txt")
 
       # Copies all files in "samples" to "tmp"
-      File.cp_r("samples", "tmp")
+      File.cp_r("samples/", "tmp/")
```
josevalim: I think your goal here is to make the intent clearer but `/` is not strictly necessary, so it can also be confusing. So I would suggest removing the trailing `/` OR have an example with both.

eksperimental: I'd rather remove the slashes than have two examples meaning the same.

eksperimental: done

PR #8635, opened 2019-01-16
Improve docs in Module functions and raise errors

eksperimental:
It offers alternatives to every possible function that has an equivalent
function for compiled modules.

It shows the mfa of the called function in the error message,
since before it only showed the function name.

### DIFF
--- a/lib/elixir/lib/module.ex
+++ b/lib/elixir/lib/module.ex
@@ -986,8 +986,8 @@ defmodule Module do
   @doc """
   Returns all functions and macros defined in `module`.
 
-  It returns a list with all functions and macros, public and private
-  defined, in the shape of `[{name, arity}, ...]`.
+  It returns a list with all defined functions and macros, public and private,
+  in the shape of `[{name, arity}, ...]`.
 
   This function can only be used on modules that have not yet been compiled.
   Use the `c:Module.__info__/1` callback to get the public functions and macros in
--- a/lib/elixir/lib/module.ex
+++ b/lib/elixir/lib/module.ex
@@ -899,6 +899,8 @@ defmodule Module do
   # Otherwise, returns a generic guess
   defp merge_signature({_, meta, _}, _newer, i), do: {:"arg#{i}", meta, Elixir}
 
+  @extra_error_msg_defines? "Use Kernel.function_exported?/3 and Kernel.macro_exported?/3 " <>
+                              "to check for public functions and macros instead"
   @doc """
   Checks if the module defines the given function or macro.
 
@@ -920,7 +922,7 @@ defmodule Module do
   @spec defines?(module, definition) :: boolean
   def defines?(module, {name, arity} = tuple)
       when is_atom(module) and is_atom(name) and is_integer(arity) and arity >= 0 and arity <= 255 do
-    assert_not_compiled!(__ENV__.function, module)
+    assert_not_compiled!(__ENV__.function, module, @extra_error_msg_defines?)
     {set, _bag} = data_tables_for(module)
     :ets.member(set, {:def, tuple})
   end
@@ -948,7 +950,8 @@ defmodule Module do
   def defines?(module, {name, arity} = tuple, def_kind)
       when is_atom(module) and is_atom(name) and is_integer(arity) and arity >= 0 and arity <= 255 and
              def_kind in [:def, :defp, :defmacro, :defmacrop] do
-    assert_not_compiled!(__ENV__.function, module)
+    assert_not_compiled!(__ENV__.function, module, @extra_error_msg_defines?)
+
     {set, _bag} = data_tables_for(module)
 
     case :ets.lookup(set, {:def, tuple}) do
@@ -979,6 +982,7 @@ defmodule Module do
     Kernel.Typespec.spec_to_callback(module, definition)
   end
 
+  @extra_error_msg_definitions_in "Use the Module.__info__/1 callback to get public functions and macros instead"
   @doc """
   Returns all functions and macros defined in `module`.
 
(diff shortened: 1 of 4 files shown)


lib/elixir/lib/module.ex
```diff
@@ -1985,9 +2000,29 @@ defmodule Module do
...
+      raise ArgumentError, assert_not_compiled_message(function_name_arity, module)
+  end
+
+  defp assert_not_compiled_message({function_name, arity} = function_name_arity, module) do
+    mfa = "Module.#{function_name}/#{arity}"
+
+    "could not call #{mfa} with module argument #{inspect(module)} because the module is already compiled" <>
+      case function_name_arity do
+        {:defines?, arity} when arity in [2, 3] ->
+          ". Use Kernel.function_exported?/3 and Kernel.macro_exported?/3 to " <>
```
josevalim: Good job! However, I think it would be better to pass this extra information when we call `assert_not_compiled!`. If there are duplicated messages, you can put it in a module attribute.

lib/elixir/lib/module.ex
```diff
@@ -1985,9 +2000,29 @@ defmodule Module do
...
     open?(module) ||
-      raise ArgumentError,
-            "could not call #{fun} with argument #{inspect(module)} because the module is already compiled"
+      raise ArgumentError, assert_not_compiled_message(function_name_arity, module)
+  end
+
+  defp assert_not_compiled_message({function_name, arity} = function_name_arity, module) do
+    mfa = "Module.#{function_name}/#{arity}"
+
+    "could not call #{mfa} with module argument #{inspect(module)} because the module is already compiled" <>
```
fertapric: I think the previous "argument" was more accurate than "module argument". Besides, it doesn't fit with `Module.eval_quoted` :)

josevalim:
```suggestion
    "could not call #{mfa} because the module #{inspect(module)} is already compiled" <>
```

lib/elixir/lib/module.ex
eksperimental: done

lib/elixir/lib/module.ex
eksperimental: @fertapric you are correct.
@josevalim that's one beautiful concise message.

lib/elixir/lib/module.ex
```diff
@@ -919,7 +922,7 @@ defmodule Module do
   @spec defines?(module, definition) :: boolean
   def defines?(module, {name, arity} = tuple)
       when is_atom(module) and is_atom(name) and is_integer(arity) and arity >= 0 and arity <= 255 do
-    assert_not_compiled!(:defines?, module)
+    assert_not_compiled!(__ENV__.function, module, @extra_error_msg_defines?)
```
fertapric: It seems this module attribute is only used once, so I would just put the message inside the function:

```elixir
extra_message = 
  "Use Kernel.function_exported?/3 and Kernel.macro_exported?/3 " <>
    "to check for public functions and macros instead"

assert_not_compiled!(__ENV__.function, module, extra_message)
```

lib/elixir/lib/module.ex
```diff
@@ -978,19 +984,27 @@ defmodule Module do
   end
 
   @doc """
-  Returns all functions defined in `module`.
+  Returns all functions and macros defined in `module`.
+
+  It returns a list with all functions and macros, public and private
```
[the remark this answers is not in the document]
eksperimental: Yes. that's better.
What about  "with all defined functions and macros" (I removed the "the")
Does that sound good to you?

lib/elixir/lib/module.ex
eksperimental: Actually, It is used twice. I misplaced the `@extra_error_msg_definitions_in` above `defines?/2`, so it may have given you the impression it was only used once.
It has been rebased, so you cannot see this in the commit history.

lib/elixir/lib/module.ex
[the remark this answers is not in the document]
fertapric: Yeah. I think "all defined" is correct (but I'm not native)

lib/elixir/lib/module.ex
```diff
@@ -977,20 +982,29 @@ defmodule Module do
     Kernel.Typespec.spec_to_callback(module, definition)
   end
 
+  @extra_error_msg_definitions_in "Use the Module.__info__/1 callback to get public functions and macros instead"
```
josevalim: This is used just once?

lib/elixir/lib/module.ex
```diff
@@ -899,13 +899,16 @@ defmodule Module do
   # Otherwise, returns a generic guess
   defp merge_signature({_, meta, _}, _newer, i), do: {:"arg#{i}", meta, Elixir}
 
+  @extra_error_msg_defines? "Use Kernel.function_exported?/3 and Kernel.macro_exported?/3 " <>
+                              "to check for public functions and macros instead"
```
josevalim: Let's move to the top of the file as it is a bit unusual to have such attributes in the middle of the code? Unsure...

eksperimental: I initially put them there and changed my mind.
I'll move them back

eksperimental: fixed

lib/elixir/lib/module.ex
eksperimental: sorry, there was something wrong with my commit history. it is fixed

lib/elixir/lib/module.ex
eksperimental: there was some issue with my commit history, please have a look again

lib/elixir/lib/module.ex
```diff
@@ -1,4 +1,9 @@
 defmodule Module do
+  @extra_error_msg_defines? "Use Kernel.function_exported?/3 and Kernel.macro_exported?/3 " <>
+                              "to check for public functions and macros instead"
+
+  @extra_error_msg_definitions_in "Use the Module.__info__/1 callback to get public functions and macros instead"
```
josevalim: Can it be after the `@moduledoc`? Sorry for the nitpicking, as we really have no rules for this.

whatyouhide:
Thanks @eksperimental! 💟

eksperimental:
Cool!, You are welcome @whatyouhide

PR #8641, opened 2019-01-16
Fix parsing Markdown links for h function (fix #8629)

### LINKED ISSUE #8629: Bug in IEx's MarkDown parser
IEx parses wrongly the following MarkDown

```
In Elixir, an enumerable is any data type that implements the
`Enumerable` protocol. [Lists](`List`) (`[1, 2, 3]`), [maps](`Map`) (`%{foo: 1, bar: 2}`)
and [ranges](`Range`) (`1..3`) are common data types used as enumerables:
```

it is shown as:

```
In Elixir, an enumerable is any data type that implements the Enumerable
protocol. Lists (List) (1, 2, 3]), [maps (Map) (%{foo: 1, bar: 2}) and ranges
(Range) (1..3) are common data types used as enumerables:
```

it should be
```
In Elixir, an enumerable is any data type that implements the Enumerable
protocol. Lists (List) (1, 2, 3]), maps (Map) (%{foo: 1, bar: 2}) and ranges
(Range) (1..3) are common data types used as enumerables:
```

romul:
I've nailed down the bug described in #8629
The case was that `.*?` could be more eager than expected when it is followed by other statement.

### DIFF
--- a/lib/elixir/lib/io/ansi/docs.ex
+++ b/lib/elixir/lib/io/ansi/docs.ex
@@ -531,7 +531,7 @@ defmodule IO.ANSI.Docs do
   end
 
   defp remove_square_brackets_in_link(text) do
-    ~r{\[(.*?)\]\((.*?)\)}
+    ~r{\[([^\]]*?)\]\((.*?)\)}
     |> Regex.recompile!()
     |> Regex.replace(text, "\\1 (\\2)")
   end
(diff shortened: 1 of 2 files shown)


josevalim:

eksperimental:
Awesome @romul!
Thanks a lot.

PR #8643, opened 2019-01-17
Improve IEx.Introspection.h/1 error message

gusaiani:
Hi all, 

this removes an extraneous `to` from a `puts_error` message:

`would like to more information` → `would like more information`

(I also considered changing it to `would like to see more information`...if that makes more sense, please let me know.

Cheers!


josevalim:

PR #8648, opened 2019-01-18
Update Kernel rounding documentation

TylerWitt:
Updates Kernel.round/1 to describe what happens on half.

### DIFF
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -835,7 +835,7 @@ defmodule Kernel do
   end
 
   @doc """
-  Rounds a number to the nearest integer. Rounds half away from 0.
+  Rounds a number to the nearest integer, half away from 0.
 
   Allowed in guard tests. Inlined by the compiler.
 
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/kernel.ex
```diff
@@ -835,7 +835,7 @@ defmodule Kernel do
   end
 
   @doc """
-  Rounds a number to the nearest integer.
+  Rounds a number to the nearest integer. Rounds half away from 0.
```
```suggestion
  Rounds a number to the nearest integer half away from 0.
```

?

TylerWitt: How about `Rounds a number to the nearest integer using half away from 0.`?

josevalim: @TylerWitt :+1:

eksperimental: Same here

fertapric: What do you think about?

>  If the number is equidistant to two integers, rounds away from zero.

TylerWitt: Updated. Thanks for the suggestions.

josevalim:

PR #8653, opened 2019-01-20
Augment doctest errors with misplaced opaque types

### LINKED ISSUE #8523: Doctest of a Map with a MapSet value fails
### Environment

Erlang/OTP 21 [erts-10.1.3]
Elixir 1.7.4 (compiled with Erlang/OTP 21)

### Current behavior
The following Doctest fails
```
  @doc """
  ## Examples

      iex> TestMapSet.test_map_set()
      %{set: #MapSet<["a"]>}
  """
  def test_map_set() do
    %{set: MapSet.new(["a"])}
  end
```
with the error:
```
Doctest did not compile, got: (TokenMissingError) lib/test_map_set.ex:9: missing terminator: } (for "{"starting at line 9)
     code: %{set: #MapSet<["a"]>}
     stacktrace:
       lib/test_map_set.ex:9: TestMapSet (module)
```

fertapric:
Closes #8523

### DIFF
--- a/lib/ex_unit/lib/ex_unit/doc_test.ex
+++ b/lib/ex_unit/lib/ex_unit/doc_test.ex
@@ -147,6 +147,8 @@ defmodule ExUnit.DocTest do
   suite run.
   """
 
+  @opaque_type_regex ~r{#.*<.*>}
+
   defmodule Error do
     defexception [:message]
 
@@ -389,6 +391,19 @@ defmodule ExUnit.DocTest do
       e ->
         ex_message = "(#{inspect(e.__struct__)}) #{Exception.message(e)}"
         message = "Doctest did not compile, got: #{ex_message}"
+
+        message =
+          if e.__struct__ == TokenMissingError and expr =~ Regex.recompile!(@opaque_type_regex) do
+            message <>
+              """
+              . If you are planning to assert on the result of an iex> expression \
+              using an opaque type (with the format #Name<...>), please make sure \
+              the opaque type is placed at the beginning of the expression; \
+              otherwise Elixir will treat it as a comment due to the leading sign #.\
+              """
+          else
+            message
+          end
 
         opts =
           if String.valid?(expr) do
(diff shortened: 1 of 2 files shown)


lib/ex_unit/lib/ex_unit/doc_test.ex
```diff
@@ -390,6 +392,19 @@ defmodule ExUnit.DocTest do
         ex_message = "(#{inspect(e.__struct__)}) #{Exception.message(e)}"
         message = "Doctest did not compile, got: #{ex_message}"
 
+        message =
+          if e.__struct__ == TokenMissingError and expr =~ Regex.recompile!(@opaque_type_regex) do
+            message <>
+              """
+              . If you are planning to assert on the result of an iex> expression \
+              using an opaque type (with the format #Name<...>), please make sure \
```
josevalim:
```suggestion
              using an opaque type in the format #Name<...>, please make sure \
```

?

lib/ex_unit/lib/ex_unit/doc_test.ex
```diff
@@ -390,6 +392,19 @@ defmodule ExUnit.DocTest do
...
+              the opaque type is placed at the beginning of the expression; \
```
josevalim: I would try to avoid the term `opaque_type`. Maybe something like:

```suggestion
              which contains a value inspected as #Name<...>, please make sure \
```

lib/ex_unit/lib/ex_unit/doc_test.ex
josevalim: I would try to avoid the term `opaque_type`. Maybe something like:

```suggestion
              which contains a value inspected as #Name<...>, please make sure \
```

lib/ex_unit/lib/ex_unit/doc_test.ex
josevalim:
```suggestion
              the inspected value is placed at the beginning of the expression; \
```

josevalim:

PR #8654, opened 2019-01-20
Fix indentation in doctests

eksperimental:
ExDoc is not formatting then as Elixir code


eksperimental:
Can be merged into v1.8 as well.

PR #8658, opened 2019-01-20
Use "operating system" instead of OS

eksperimental:


josevalim:
My thought process was that operating system is vague so putting them in title case would make it clear we are talking about THE operating system. But I don't know if this is only me.

Another operation is "operating system (OS)".

Thoughts?

eksperimental:
> Another option is "operating system (OS)".
I like this, but only for the documentation for functions with the "os" acronym

josevalim:
Sounds good to me then! :D

eksperimental:
pushed, also check the last commit

PR #8663, opened 2019-01-21
Update Module.spec_to_callback to keep existing specs and copy docs

wojtekmach:
Ref: https://github.com/elixir-lang/ex_doc/issues/942

### DIFF
--- a/lib/elixir/lib/kernel/typespec.ex
+++ b/lib/elixir/lib/kernel/typespec.ex
@@ -75,12 +75,21 @@ defmodule Kernel.Typespec do
 
   def spec_to_callback(module, {name, arity} = signature)
       when is_atom(module) and is_atom(name) and arity in 0..255 do
-    {_set, bag} = :elixir_module.data_tables(module)
+    {set, bag} = :elixir_module.data_tables(module)
 
     filter = fn {:spec, expr, pos} ->
       if spec_to_signature(expr) == signature do
-        delete_typespec(bag, :spec, expr, pos)
-        store_typespec(bag, :callback, expr, pos)
+        kind = :callback
+        store_typespec(bag, kind, expr, pos)
+
+        case :ets.lookup(set, {:function, name, arity}) do
+          [{{:function, ^name, ^arity}, line, _, doc, doc_meta}] ->
+            store_doc(set, kind, name, arity, line, :doc, doc, doc_meta)
+
+          _ ->
+            nil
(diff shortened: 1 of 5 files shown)


lib/elixir/lib/kernel/typespec.ex
```diff
@@ -75,12 +75,21 @@ defmodule Kernel.Typespec do
...
+    {set, bag} = :elixir_module.data_tables(module)
 
     filter = fn {:spec, expr, pos} ->
       if spec_to_signature(expr) == signature do
-        delete_typespec(bag, :spec, expr, pos)
-        store_typespec(bag, :callback, expr, pos)
+        kind = :callback
+        store_typespec(bag, kind, expr, pos)
+
+        case :ets.lookup(set, {:function, name, arity}) do
```
fertapric: I tried locally just by removing the line `delete_typespec(bag, :spec, expr, pos)` and the `@spec` is shown on IEx, why is this part needed? 🙂

josevalim:
That part is needed to add the docs to the callback. Btw Wojtek, we need a
test that the spec was not deleted. It could be one extra assertion in the
test you added. Although we should probably add unit tests to
typespec_test.exs, at least where specs/callbacks are concerned.
-- 


fertapric:
@josevalim Ah! Gotcha. Thanks :)

Btw, I would update update this comment: 

https://github.com/elixir-lang/elixir/blob/985bfe53b4074db34d548e6856ac432e699726c0/lib/elixir/lib/protocol.ex#L45-L48

lib/elixir/lib/module.ex
```diff
@@ -974,9 +974,9 @@ defmodule Module do
   end
 
   @doc """
-  Converts the given spec to a callback.
+  Copies the given spec as a callback.
```
fertapric: Maybe mention that the docs are also copied to the callback?

lib/elixir/lib/module.ex
```diff
@@ -974,9 +974,11 @@ defmodule Module do
...
 
-  Returns `true` if there is such a spec and it was converted to a callback.
+  Returns `true` if there is such a spec and it was copied as a callback.
+  The function associated to the spec has documentation defined prior to
```
fertapric:
```suggestion
  If the function associated to the spec has documentation defined prior to
```

josevalim:

PR #8668, opened 2019-01-22
Name args in Calendar.UTCOnlyTimeZoneDatabase callbacks

eksperimental:


josevalim:
@eksperimental what is the benefit? :)

josevalim:
@eksperimental in this case I would change the inference system in ExDoc. If the type is a remote signature, we can use the name of the type invoked remotely. :)

fertapric:
> If the type is a remote signature, we can use the name of the type invoked remotely. :)

And maybe the name of the module in case of `MyModule.t()`? (with the proper conversion to `snake_case`)

PR #8673, opened 2019-01-22
Add ability to filter by more than one line number

devonestes:
As [discussed in the mailing list](https://groups.google.com/d/msg/elixir-lang-core/7yejSvr0RMM/Cfv5JXDnBgAJ), here's a PR that will allow a
shorthand for filtering ExUnit tests by more than one line number at a
time.

For now this doesn't address the question of running single tests in
more than one file since that would be a much larger change. If we do
want to allow that, we can add it later on, but it should be an addition
to the work here, not a replacement.

### DIFF
--- a/lib/ex_unit/lib/ex_unit/filters.ex
+++ b/lib/ex_unit/lib/ex_unit/filters.ex
@@ -14,16 +14,31 @@ defmodule ExUnit.Filters do
   on the command line) includes a line number filter, and if so returns the
   appropriate ExUnit configuration options.
   """
-  @spec parse_path(String.t()) :: {String.t(), any}
+  @spec parse_path(String.t()) :: {String.t(), Keyword.t()}
   def parse_path(file) do
-    {paths, [line]} = file |> String.split(":") |> Enum.split(-1)
+    {paths, lines, _, _} =
+      file
+      |> String.split(":")
+      |> Enum.reduce({[], [], false, false}, &reduce_line_numbers/2)
 
-    case Integer.parse(line) do
-      {_, ""} ->
-        {Enum.join(paths, ":"), exclude: [:test], include: [line: line]}
+    path = Enum.join(paths, ":")
 
-      _ ->
-        {file, []}
+    case lines do
+      [] -> {path, []}
+      _ -> {path, exclude: [:test], include: Enum.map(lines, &{:line, &1})}
+    end
+  end
+
+  defp reduce_line_numbers(part, {paths, _, false, false}) do
+    {paths ++ [part], [], false, String.match?(part, ~r/\.exs/)}
+  end
+
+  defp reduce_line_numbers(part, {paths, _, true, _}), do: {paths ++ [part], [], true, false}
+
+  defp reduce_line_numbers(part, {paths, parts, _, true}) do
+    case Integer.parse(part) do
+      {_, ""} -> {paths, parts ++ [part], false, true}
+      _ -> {paths ++ parts ++ [part], [], true, true}
     end
   end
 
--- a/lib/ex_unit/lib/ex_unit/filters.ex
+++ b/lib/ex_unit/lib/ex_unit/filters.ex
@@ -16,29 +16,34 @@ defmodule ExUnit.Filters do
   """
   @spec parse_path(String.t()) :: {String.t(), Keyword.t()}
   def parse_path(file) do
-    {paths, lines, _, _} =
-      file
-      |> String.split(":")
-      |> Enum.reduce({[], [], false, false}, &reduce_line_numbers/2)
-
-    path = Enum.join(paths, ":")
-
-    case lines do
-      [] -> {path, []}
-      _ -> {path, exclude: [:test], include: Enum.map(lines, &{:line, &1})}
+    case extract_line_numbers(file) do
+      {path, []} -> {path, []}
+      {path, line_numbers} -> {path, exclude: [:test], include: line_numbers}
     end
   end
 
-  defp reduce_line_numbers(part, {paths, _, false, false}) do
-    {paths ++ [part], [], false, String.match?(part, ~r/\.exs/)}
-  end
+  defp extract_line_numbers(file) do
+    case String.split(file, ":") do
+      [part] ->
+        {part, []}
 
-  defp reduce_line_numbers(part, {paths, _, true, _}), do: {paths ++ [part], [], true, false}
+      parts ->
+        {reversed_line_numbers, reversed_path_parts} =
+          parts
+          |> Enum.reverse()
+          |> Enum.split_while(&match?({_, ""}, Integer.parse(&1)))
 
-  defp reduce_line_numbers(part, {paths, parts, _, true}) do
-    case Integer.parse(part) do
-      {_, ""} -> {paths, parts ++ [part], false, true}
-      _ -> {paths ++ parts ++ [part], [], true, true}
+        line_numbers =
+          reversed_line_numbers
+          |> Enum.reverse()
+          |> Enum.map(&{:line, &1})
+
+        path =
+          reversed_path_parts
+          |> Enum.reverse()
+          |> Enum.join(":")
+
+        {path, line_numbers}
     end
   end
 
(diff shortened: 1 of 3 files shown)


lib/ex_unit/lib/ex_unit/filters.ex
```diff
@@ -14,16 +14,31 @@ defmodule ExUnit.Filters do
   on the command line) includes a line number filter, and if so returns the
   appropriate ExUnit configuration options.
   """
-  @spec parse_path(String.t()) :: {String.t(), any}
+  @spec parse_path(String.t()) :: {String.t(), Keyword.t()}
   def parse_path(file) do
-    {paths, [line]} = file |> String.split(":") |> Enum.split(-1)
```
fertapric: What do you think about abstracting the concept of "extract line numbers"? so the code here would be something like:

```elixir
case extract_line_numbers(file) do
  {path, []} ->  {path, []}
  {path, line_numbers} -> {path, exclude: [:test], include: Enum.map(lines, &{:line, &1})}
end
```

lib/ex_unit/lib/ex_unit/filters.ex
```diff
@@ -14,16 +14,31 @@ defmodule ExUnit.Filters do
...
-      _ ->
-        {file, []}
+    case lines do
+      [] -> {path, []}
+      _ -> {path, exclude: [:test], include: Enum.map(lines, &{:line, &1})}
+    end
+  end
+
+  defp reduce_line_numbers(part, {paths, _, false, false}) do
+    {paths ++ [part], [], false, String.match?(part, ~r/\.exs/)}
```
josevalim: I would prefer to avoid using regexes. Here is a general algorithm I would propose:

1. Split the path into ":"
2. Revert the split path and call Enum.split_while so you take trailing information while `match?({_, ""}, Integer.parse(part))` returns true
3. Once that is done, reverse the path back, join the remaining parts with `:` and compute the line numbers

This should be cleaner and rely less on stateful reduce.

Thoughts?

josevalim: To be more precise, something like this:

```elixir
path
|> String.split(path, ":")
|> extract_line_numbers()
|> case do
  {[], _} ->
    {path, []}

  {reverse_lines, reverse_parts} ->
    include = reverse_lines |> Enum.reverse() |> Enum.map(&{:line, String.to_integer(&1)})
    path = reverse_parts |> Enum.reverse() |> Enum.join(":")
    {path, exclude: [:test], include: include}
end

defp extract_line_numbers([part]), do: {[], [part]}

defp extract_line_numbers(parts) do
  parts
  |> Enum.reverse()
  |> Enum.split_while(&match?({_, ""}, Integer.parse(&1)))
end
```

fertapric: What about? (a little refactor) :)

```elixir
case extract_line_numbers(path) do
  {path, []} -> {path, []}
  {path, line_numbers} -> {path, exclude: [:test], include: line_numbers}
end

defp extract_line_numbers(path) do
  case String.split(path, ":") do
    [part] -> 
      {path, []}

    parts ->
      {reversed_line_numbers, reversed_path_parts} =
        parts
        |> Enum.reverse()
        |> Enum.split_while(&match?({_, ""}, Integer.parse(&1)))

      line_numbers = 
        reversed_line_numbers 
        |> Enum.reverse() 
        |> Enum.map(&{:line, String.to_integer(&1)})
        
      path = 
        reversed_path_parts 
        |> Enum.reverse() 
        |> Enum.join(":")

      {path, line_numbers}
  end
end
```

devonestes: I did consider something like that, but that would change the semantics of the feature. If we were do to that, the following test would change from:

```elixir
    assert ExUnit.Filters.parse_path("test/some/path.exs:123notalinenumber123:456") ==
             {"test/some/path.exs:123notalinenumber123:456", []}
```

to

```elixir
    assert ExUnit.Filters.parse_path("test/some/path.exs:123notalinenumber123:456") ==
             {"test/some/path.exs:123notalinenumber123", [exclude: [:test], include: [line: "456"]]}
```

That would create some weird behavior since then the following test (with those same line numbers but in different order) would remain the same as:

```elixir
    assert ExUnit.Filters.parse_path("test/some/path.exs:123:456notalinenumber456") ==
             {"test/some/path.exs:123:456notalinenumber456", []}
```

whatyouhide:
Should have suggested this in the mailing list, but what if instead of `myfile:1:2:3`, we use the syntax `myfile:1,2,3`? It would simplify the parsing and I think it's easier to understand as well. If it's too late for this discussion, ignore me 😄

josevalim:
@whatyouhide I think multiple `:123:123:123` is the most natural but that's a personal opinion. ALso I am not sure if it facilitates parsing, because we still need to split the comma and verify each part of the comma is a number, so we don't accidentally match on a path with commas in it and consider it to be line numbers. May even be slightly more complex.

josevalim:
> We should execute the tests in the order they have been passed, because when combined with the --max-failures feature,

Filter have no relation with ordering. Ordering is ultimately decided by the `seed` and I believe it to be unrelated.

In any case, :+1: for the documentation. The current docs are in `mix test`.

eksperimental:
> > We should execute the tests in the order they have been passed, because when combined with the --max-failures feature,
> 
> Filter have no relation with ordering. Ordering is ultimately decided by the `seed` and I believe it to be unrelated.

when passing `--seed 0`, the order is not maintained.

josevalim:
> when passing --seed 0, the order is not maintained.

Yes, --seed 0 sorts them by line number I believe?

eksperimental:
My suggestion is that when passing `--seed 0`, it should executes first line 14, and then line 8  
Including tags keep the order, but when it is executed, it is ignored.

```
$ mix test test/some_test.exs:14:8 --seed 0
Excluding tags: [:test]
Including tags: [line: "14", line: "8"]

9
.15
.

Finished in 2.1 seconds
3 tests, 0 failures, 1 excluded

Randomized with seed 0
```

eksperimental:
@josevalim you are absolutely right.
Having the pre-load all the tests to determine the the order is the big limitation here.

devonestes:
@josevalim I intended to document this further once the behavior was settled. If we think what's here now is what's going to stay, then I'll gladly add some docs & examples. I just wanted to make sure what was implemented was actually accepted before documenting.

josevalim:
@devonestes beautiful. So let's add the docs so we can ship it :D

eksperimental:
It works with two files and no line numbers
```
$ mix test test/testing_test.exs test/testing2_test.exs
Elixir.TestingTest: 9
.Elixir.TestingTest: 21
.Elixir.TestingTest: 15
.Elixir.Testing2Test: 9
.Elixir.Testing2Test: 21
.Elixir.Testing2Test: 15
.

Finished in 4.0 seconds
6 tests, 0 failures

Randomized with seed 324618
```


It ignores the first file, when this one uses line numbers and second one doesn't
```
$ mix test test/testing_test.exs:14 test/testing2_test.exs
Elixir.Testing2Test: 9
.Elixir.Testing2Test: 21
.Elixir.Testing2Test: 15
.

Finished in 2.0 seconds
3 tests, 0 failures

Randomized with seed 222014
```

and similar to the previous one, it ignores the second file when this one uses line numbers and the first one doesn't
```
$ mix test test/testing_test.exs test/testing2_test.exs:14
Elixir.TestingTest: 15
.Elixir.TestingTest: 21
.Elixir.TestingTest: 9
.

Finished in 2.0 seconds
3 tests, 0 failures

Randomized with seed 860631
```

It does not recognize the second file as a valid one, when both files use line numbers
```
$ mix test test/testing_test.exs:14 test/testing2_test.exs:14
Paths given to "mix test" did not match any directory/file: test/testing_test.exs:14, test/testing2_test.exs:14
```

josevalim:
@eksperimental is this a bug related to this enhancement though? The feature already exists today, do you mean it is buggy today?

eksperimental:
> @eksperimental is this a bug related to this enhancement though? The feature already exists today, do you mean it is buggy today?

Sorry, I didn't realize it is not related to this PR

fertapric:
@eksperimental it's on the PR description :)

eksperimental:
> @eksperimental it's on the PR description :)

It is not related to this PR. Reading the docs from `test`
> In case a single file is being tested, it is possible to pass a specific line number:

I have created a PR https://github.com/elixir-lang/elixir/issues/8678 for this.

josevalim:

eksperimental:
great addition to ExUnit!

PR #6331, opened 2017-07-14
Minor updates to Calendar.ISO and test Calendar.Holocene

wojtekmach:

PR #8676, opened 2019-01-22
Added missing doc and spec to Calendar.ISO.time_to_string/5

amatalai:
It was missing since 
```elixir
def time_to_string(hour, minute, second, microsecond, format \\ :extended)
```
 was split into `time_to_string/4` and `time_to_string/5` without default value in #6331

### DIFF
--- a/lib/elixir/lib/calendar/iso.ex
+++ b/lib/elixir/lib/calendar/iso.ex
@@ -482,6 +482,24 @@ defmodule Calendar.ISO do
     time_to_string(hour, minute, second, microsecond, :extended)
   end
 
+  @doc """
+  Converts the given time into a string in the given format.
+
+  ## Examples
+
+      iex> Calendar.ISO.time_to_string(2, 2, 2, {2, 6}, :basic)
+      "020202.000002"
+      iex> Calendar.ISO.time_to_string(2, 2, 2, {2, 6}, :extended)
+      "02:02:02.000002"
+
+  """
+  @spec time_to_string(
+          Calendar.hour(),
+          Calendar.minute(),
+          Calendar.second(),
+          Calendar.microsecond(),
+          :basic | :extended
+        ) :: String.t()
   def time_to_string(hour, minute, second, {_, 0}, format) do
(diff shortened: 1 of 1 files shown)


fertapric:
`Time.to_iso8601/2` provides a similar functionality. I think this was supposed to be private  (like `date_to_string(year, month, day, format)`), and accessed via the `to_iso8601` functions.

josevalim:
I think @fertapric is correct. It is meant to be private, so adding `@doc false` is the best way to go.

In any case, good catch @amatalai! Can you please amend?

fertapric:
@josevalim I would use `defp` in this case. The functions are publicly accessible using `date_to_iso8601/4`:

https://github.com/elixir-lang/elixir/blob/9014d4f3653f0266acfcaf976b58d7a87b387f0e/lib/elixir/lib/calendar/iso.ex#L745-L748

amatalai:
Are you sure it won't be breaking change? It was documented in older version of elixir, for example: https://hexdocs.pm/elixir/1.5.0/Calendar.ISO.html#time_to_string/5

josevalim:
You are correct. So I would merge the two definitions together again by using default values and improve the docs. WDYT?

fertapric:
@josevalim it seems `@impl true` was complaining about that https://github.com/elixir-lang/elixir/pull/6331#discussion_r127476966

fertapric:
@josevalim what do you think about adding `@doc deprecated: "Use Time.to_iso8601/2 instead"`?

josevalim:
@fertapric Calendar.ISO is meant to be a bunch  of functionality that works without hte data structure, so they are not quite equivalent. Plus it doesn't hurt to keep it public since we need to keep the functionality anyway.

josevalim:
To be clear, I am saying no to deprecating it. Let's keep it. We will be fie. :)

fertapric:
@josevalim ok. I was thinking about consistency. There is also `date_to_string`, `datetime_to_string`, and `naive_datetime_to_string` :)

josevalim:

Issue #8612, opened 2019-01-11
Releases


josevalim:
This is a meta issue to track the progress of releases in Elixir. The goal is to include minimum support for releases, then add user conveniences, and then finally work on hot code upgrades with relups and appups.

## Milestone 1

The goal here is to add releases to core without many bells and whistles:

  * [x] Add new options to Elixir CLI to trim the size of release executables (#8595)
  * [x] Add basic `mix release` support to Elixir (#8677)

## Milestone 2

The goal here is implement user facing features:

  * [x] Add eval, version, start_iex, daemon_iex commands
  * [x] Custom steps
  * [x] Custom vm.args (and possibly rel/vm.args.eex) and bin/start (and possibly rel/start.eex) files (possibly with a generator?)
  * [x] Release configuration (via config providers) and a rewrite of the existing `Mix.Config`. The default config provider will look a config at `config_path |> Path.dirname() |> Path.join("release.exs")` and always log if a release config is being used or not. We also need to raise a good error message if `Mix.Config` is used instead of `Config`

josevalim:
Hi folks,

I have spent most of the week studying hot code upgrades and building prototypes and I have decided to not include them as part of Elixir Core for now (at least for Elixir v1.9). The rationale is that they are still very complex, error prone and opinionated in a way that whoever is performing them needs to be aware of many of the decisions taken. For example, the [distillery guides on appups](https://github.com/bitwalker/distillery/blob/master/docs/guides/appups.md) (which is a part of hot code upgrades) covers many of those topics.

We are not saying they will never be part of Elixir but *not for now*. This can be a good opportunity for the community to build on top of what Elixir provides.

I wrote an extensive section on the docs about hot code upgrades which I reproduce below for convenience that explains this rationale and guides users.

---

Erlang and Elixir are sometimes known for the capability of upgrading
a node that is running in production without shutting down that node.
However, this feature is not supported out of the box by Elixir releases.

The reason we don't provide hot code upgrades is because they are very
complicated to perform in practice, as they require careful coding of
your processes and applications as well as extensive testing. Given most
teams can use other techniques that are language agnostic to upgrade
their systems, such as Blue/Green deployments, Canary deployments,
Rolling deployments, and others, hot upgrades are rarely a viable
option. Let's understand why.

In a hot code upgrade, you want to update a node from version A to
version B. To do so, the first step is to write recipes for every application
that changed between those two releases, telling exactly how the application
[...]

PR #8677, opened 2019-01-22
[WIP] mix release support

josevalim:
For the complete roadmap for releases, see #8612.


josevalim:
Hi everyone, this PR is feature complete for the goals it aims to achieve (for a complete roadmap see #8612). The missing parts are:

  * Windows tooling
  * Improve the test suite

This means it would be really appreciated if we start to get some reviews in. There are two ways you can help:

  1. Review the code in the PR
  2. Review the docs. The docs are all here: https://github.com/elixir-lang/elixir/pull/8677/files#diff-53c844525c23fd1a21c03113baf7e781R3

Thanks!

lib/mix/lib/mix/release.ex
```diff
@@ -0,0 +1,389 @@
...
+    destination
+    |> Path.join("bin/erl")
+    |> File.write!(~S"""
+    #!/bin/sh
+    SELF=$(readlink "$0" || true)
+    if [ -z "$SELF" ]; then SELF="$0"; fi
+    BINDIR="$(cd "$(dirname "$SELF")" && pwd -P)"
+    ROOTDIR="$(dirname "$(dirname "$BINDIR")")"
+    EMU=beam
+    PROGNAME=$(echo $0 | sed 's/.*\///')
```
lukaszsamson:
```suggestion
    PROGNAME=$(echo "$0" | sed 's/.*\///')
```

lib/mix/lib/mix/tasks/release.ex
```diff
@@ -0,0 +1,913 @@
...
+  ##-env ERL_FULLSWEEP_AFTER 10
+  """)
+
+  embed_template(:start, ~S"""
+  #!/bin/sh
+  set -e
+  # Feel free to edit this file in anyway you want
+  # To start your system using IEx: . $(dirname "$0")/<%= @name %> start iex
+  # To start it as a daemon using IEx: . $(dirname "$0")/<%= @name %> daemon iex
+  . $(dirname "$0")/<%= @name %> start
```
lukaszsamson: Will `<%= @name %>` be a valid unix path?

lib/mix/lib/mix/tasks/release.ex
```diff
@@ -0,0 +1,913 @@
...
+    restart|stop)
+      rpc "System.$1"
+      ;;
+
+    pid)
+      rpc "IO.puts System.pid"
+      ;;
+
+    *)
+      echo "Usage: $(basename $0) COMMAND [ARGS]
```
lukaszsamson:
```suggestion
      echo "Usage: $(basename "$0") COMMAND [ARGS]
```

lib/mix/lib/mix/tasks/release.ex
```diff
@@ -0,0 +1,913 @@
...
+  . $(dirname "$0")/<%= @name %> start
+  """)
+
+  embed_template(:cli, ~S"""
+  #!/bin/sh
+  set -e
+
+  SELF=$(readlink "$0" || true)
+  if [ -z "$SELF" ]; then SELF="$0"; fi
+  export RELEASE_ROOT="$(cd "$(dirname "$SELF")/.." && pwd -P)"
```
lukaszsamson:
```suggestion
  export RELEASE_ROOT
  RELEASE_ROOT="$(cd "$(dirname "$SELF")/.." && pwd -P)"
```

lib/mix/lib/mix/tasks/release.ex
```diff
@@ -0,0 +1,913 @@
...
+         --rpc-eval "$RELEASE_NAME@127.0.0.1" "$1"
+  }
+
+  start () {
+    exec "$REL_VSN_DIR/$1" --no-halt \
+         --werl --name "$RELEASE_NAME@127.0.0.1" --cookie "$COOKIE" \
+         --erl-config "${REL_VSN_DIR}/sys" \
+         --boot "${REL_VSN_DIR}/start" \
+         --boot-var RELEASE_LIB "$RELEASE_ROOT/lib" \
+         --vm-args "${REL_VSN_DIR}/vm.args" "${@:2}"
```
lukaszsamson: string indexing is undefined in POSIX sh
https://wiki.ubuntu.com/DashAsBinSh#A.24.7Bfoo:3.5B:1.5D.7D

lib/mix/lib/mix/tasks/release.ex
```diff
@@ -0,0 +1,913 @@
...
+  }
+
+  case $1 in
+    start)
+      start ${2:-elixir}
```
lukaszsamson:
```suggestion
      start "${2:-elixir}"
```

lib/mix/lib/mix/tasks/release.ex
```diff
@@ -0,0 +1,913 @@
...
+      ;;
+
+    daemon)
+      export RELEASE_TMP="${RELEASE_TMP:-"$RELEASE_ROOT/tmp"}"
+      start ${2:-elixir} --pipe-to "${RELEASE_TMP}/pipe" "${RELEASE_TMP}/log"
```
lukaszsamson:
```suggestion
      start "${2:-elixir}" --pipe-to "${RELEASE_TMP}/pipe" "${RELEASE_TMP}/log"
```

lib/mix/lib/mix/release.ex
```diff
@@ -0,0 +1,389 @@
...
+    File.mkdir_p!(destination)
+    File.cp_r!(release.erts_source, destination)
+
+    _ = File.rm(Path.join(destination, "bin/erl"))
+    _ = File.rm(Path.join(destination, "bin/erl.ini"))
+
+    destination
+    |> Path.join("bin/erl")
+    |> File.write!(~S"""
+    #!/bin/sh
```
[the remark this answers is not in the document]
josevalim: Yeah, I noticed this but this script comes from the Erlang Distribution and I wanted to avoid doing changes to it. :)

lib/mix/lib/mix/tasks/release.ex
josevalim: I will validate this. Good catch.

lib/mix/lib/mix/tasks/release.ex
josevalim: Can you please explain the difference? Do we also need to do this for the environment variables defined below? Thanks!

lib/mix/lib/mix/release.ex
```diff
@@ -0,0 +1,389 @@
...
+  Copies the cookie to the given path.
+
+  If a cookie option was given, we compare it with
+  the contents of the file (if any), and as the user
+  if they want to override.
+
+  If there is no option, we generate a random one
+  the first time.
+
+  Returns a boolean saying if the cookie was copied
```
saulecabrera: I might be wrong, but I feel it reads better and it's more explicit:

```
  Returns true if the cookie was copied, false otherwise.
```

lib/mix/lib/mix/tasks/release.ex
josevalim: I see... what do you suggest? To assign $1, then call shift, and then use $@?

lib/mix/lib/mix/release.ex
josevalim: Tks, fixed locally!

lib/mix/lib/mix/tasks/release.ex
lukaszsamson: Exporting a value in this way will mask exit code (e.g. when we cannot `cd` for some reason). IMO only worth if the commands can fail.

lib/mix/lib/mix/tasks/release.ex
josevalim: The above worked. Thanks for the review on the scripts, really appreciated!

lukaszsamson: Good idea

lib/mix/lib/mix/tasks/release.ex
```diff
@@ -0,0 +1,916 @@
+defmodule Mix.Tasks.Release do
+  @moduledoc """
+  Assembles a release for the current project:
+
+      MIX_ENV=prod mix release
+      MIX_ENV=prod mix release NAME
+
+  Once a release is assembled, it can be packaged and deployed to a
+  remote host, as long as the remote host runs on the same operating
+  system (OS) and versions as the machine running the `mix release`
```
sneako: "versions" is a bit ambiguous to me here. Maybe it should clarify "Elixir and Erlang/OTP versions"?

josevalim: Definitely ambiguous as it is the exact same OS version. :D How to better phrase this? "The same operating system (OS) distribution and version"?

sneako: I think that is a bit more clear. Should we also mention that the `ERTS` version must match if you have chosen not to include it in the release?

josevalim: We do mention that but in a separate section (when we talk about include_erts: false).

josevalim: Done!

lib/mix/lib/mix/tasks/release.ex
```diff
@@ -0,0 +1,916 @@
...
+  If you open up `bin/start`, you will also see there is an option to
+  run the release in daemon mode. In daemon mode, the system is started
+  on the background via [run_erl](http://erlang.org/doc/man/run_erl.html)
+  on Unix-like systems and [erlsrv](http://erlang.org/doc/man/erlsrv.html)
+  on Windows systems.
+
+  ## Deployments
+
+  It is very important that a release is assembled on a machine with the
+  same operating system (OS) version and tooling as the machine the system
```
bitwalker: I think it is better to be more specific about the requirements here, specifically what requirements need to be met to build a release and deploy to a target. Something like the following:

A release is built on a **host**, a machine which contains Erlang, Elixir, and any other dependencies needed to compile your application. A release is then deployed to a **target**, potentially the same machine as the host, but usually separate, and often there are many targets (either multiple instances, or the release is deployed to heterogeneous environments).

To deploy straight from a host to a separate target without cross-compilation, the following must be the same between the host and the target:

- Target architecture (e.g. x86_64 vs ARM)
- Target Vendor+OS (e.g. Windows, Linux, Darwin/macOS)
- Target ABI (e.g. musl, gnu)

This is often represented in the form of target triples, e.g. `x86_64-unknown-linux-gnu`, `x86_64-unknown-linux-musl`, `x86_64-apple-darwin`. 

So to be more precise, to deploy straight from a host to a separate target, the Erlang Runtime System (ERTS), and any native dependencies (NIFs), must be compiled for the same target triple. If you are building on a MacBook (`x86_64-apple-darwin`) and trying to deploy to a typical Ubuntu machine (`x86_64-unknown-linux-gnu`), the release will not work. Instead you should build the release on a `x86_64-unknown-linux-gnu` host; trivially done by building on the target itself, but most commonly done by building the release in a virtual machine or container, which can be automated as part of your release process.

In addition to matching the target triple, it is also important that the target has all of the system packages that your application will need at runtime. A common one is the need for OpenSSL when building an application that uses `:crypto` or `:ssl`, which is dynamically linked to ERTS. The other common source for native dependencies like this comes from dependencies containing NIFs (natively-implemented functions) which may expect to dynamically link to libraries they use.

These system packages are typically managed using the system package manager, but if necessary, you can also bundle the compiled object files in the release, as long as they were compiled for the same target. If doing so, you need to update `LD_LIBRARY_PATH` with the paths containing the bundled objects.

Currently, there is no official way to cross-compile a release from one target triple to another, however it is possible to build ERTS on a given target once, copy it to your host, and then reference that "cross-compiled" ERTS rather than the host ERTS. However, this does not account for NIFs your release contains, which would need to be cross-compiled as well, and included in the release. Due to the complexity, it is generally recommended to use the approaches mentioned earlier if your target can be emulated in a virtual machine or container.

lib/mix/lib/mix/tasks/release.ex
```diff
@@ -0,0 +1,916 @@
...
+    * `:applications` - a keyword list that configures and adds new applications
+      to the release. The key is the application name and the value is one of:
+
+        * `:permanent` - the application is started and the node shuts down
+          if the application terminates, regardless of reason
+        * `:transient` - the application is started and the node shuts down
+          if the application terminates abnormally
+        * `:temporary` - the application is started and the node does not
+          shut down if the application terminates
+        * `:load` - the application is only loaded
```
bitwalker: One thing that bears mentioning here is that if an application is set to `:load` or `:none`, but is a dependency of other applications in the release, then those dependents will not start, even if declared `:permanent`; instead the release will just get stuck halfway through init, waiting for an application to start that never will.

The result of the above is that a release will appear to have started normally, but upon inspection will seem to be missing functionality (for example, perhaps a Phoenix endpoint isn't available, because the application containing isn't started yet due to waiting on some other application).

In my opinion, neither of these options should be made available without some static analysis which guarantees that the applications marked load/none will not produce a broken release.

lib/mix/lib/mix/tasks/release.ex
```diff
@@ -0,0 +1,916 @@
...
+
+    * `:include_executables_for` - a list of atoms with executables to include on
+      the release
+
+    * `:cookie` - a string representing the Erlang Distribution cookie. If no cookie
+      is set, one is automatically generated when the first release is assembled. The
+      cookie will be written to `releases/COOKIE` and shared across multiple release
+      versions. If you are setting this option manually, we recommend the cookie option
+      to be a long and randomly generated, such as:
+      `Base.url_encode64(:crypto.strong_rand_bytes(40))`
```
bitwalker: The presence of `-` in the cookie can cause some problems in my experience, all of the following actually:

- `!, ~, -, +, ', ", \, #`

Obviously not all of these are produced by `url_encode64`, but they _are_ characters that are interpreted specially in `vm.args` or on the command line, and had to be filtered out in Distillery at various points. I'd make sure they aren't produced by the tooling here, but should also be documented so that people don't use them in cookies they produce on their own.

lib/mix/lib/mix/tasks/release.ex
josevalim: Good call. I will have added a TODO for the static check.

lib/mix/lib/mix/tasks/release.ex
josevalim: This is a beautiful description, thank you. I have incorporated most of it into an explicit section called "Requirements".

beardedeagle: It could be possible to set this up similar to the rustup tool where you can download these target triples using something like `mix release.toolchain download x86_64-unknown-linux-gnu` and use something like `mix release.toolchain set x86_64-unknown-linux-gnu` to set prior to building your release, along with commands for updating, listing, deleting, etc. These target triples could be downloaded from hex similar to hex tooling itself. Just an idea.

lib/mix/lib/mix/tasks/release.ex
```diff
@@ -0,0 +1,959 @@
...
+      to the release. The key is the application name and the value is one of:
+
+        * `:permanent` - the application is started and the node shuts down
+          if the application terminates, regardless of reason
+        * `:transient` - the application is started and the node shuts down
+          if the application terminates abnormally
+        * `:temporary` - the application is started and the node does not
+          shut down if the application terminates
+        * `:load` - the application is only loaded
+        * `:none` - the application is part of the release but it isn't loaded nor
```
wojtekmach:
```suggestion
        * `:none` - the application is part of the release but it is neither loaded nor
```

lib/mix/lib/mix/tasks/release.ex
```diff
@@ -0,0 +1,959 @@
...
+
+  In other words, there are multiple ways systems can be deployed and
+  releases can be automated and incorporated into all of them as long
+  as you follow the requirements outlined above.
+
+  Once a system is deployed, shutting down said systems can be done by
+  sending SIGINT/SIGTERM to the system, which is what most containers,
+  platforms and tools do, or by explicitly invoking `bin/RELEASE_NAME stop`.
+  Shutting down the system consists of stopping each application and
+  their respective supervisiont trees, one by one, in the opposite order
```
wojtekmach:
```suggestion
  their respective supervision trees, one by one, in the opposite order
```

lib/mix/lib/mix/tasks/release.ex
josevalim: @beardedeagle the VM part is doable but the problem is also every NIF available which also needs to consider the Erlang runtime version if precompiled (so it becomes a quadruple and at this point probably too many possibilities). Others will be able explain it better than me. :)

lib/mix/lib/mix/tasks/release.ex
```diff
@@ -0,0 +1,959 @@
...
+
+  Each key represents the release name. Releasing a certain name is done
+  with:
+
+      MIX_ENV=prod mix release demo
+
+  If the given name does not exist, an error is raised.
+
+  If `mix release`, without a name, is invoked and there are multiple names,
+  the first name is used. If `mix release` is invoked and there are no names,
```
AndrewDryga: Why do we want to add this edge case? Isn't it's better to ask the developer (eg. by returning an error) which release to use when we are uncertain?

lib/mix/lib/mix/tasks/release.ex
```diff
@@ -0,0 +1,959 @@
...
+
+      It defaults to the current application and then we proceed to include all
+      applications the current application depends on recursively. You can
+      include new applications or change the mode of existing ones by listing
+      them here.
+
+      Releases assembled from an umbrella project require this configuration
+      to be explicitly given.
+
+    * `:strip_beams` - a boolean that controls if BEAM files should have their debug
```
AndrewDryga: I guess `strip_beam_files` can more clear here

lib/mix/lib/mix/tasks/release.ex
bitwalker: The default value for that config option chooses the first one in the config, so you are correct that Distillery does behave that way if you make no changes to that setting with multiple releases present :)

josevalim:
Windows support is in. Now only integration tests are missing.

josevalim:
Integration tests are in. It is just a matter of becoming green on Travis+Appveyor.

eksperimental:
@josevalim Are you planning one merging this soon, or in the near future? I'm asking this because I'm started working on converting the bin scripts to a POSIX compliant, but I worked master. I see many of the non-POSIX compliant code have been eliminated in this PR.
So far I have only worked on bin/elixir but it's looking good.

josevalim:
@eksperimental I want to merge it really soon. Regarding POSIX, CI got a bunch of failures, that's why they have been fixed. :) But I would appreciate if the new PR is based off this branch.

eksperimental:
Right on. I will submit a PR once this gets merged.

josevalim:
@eksperimental tests are green on travis, which means we should no longer change the script files. :)

josevalim:
@eksperimental btw, if you can also review the release scripts for POSIX standards, it would be awesome. Although we should be good on this front too.

eksperimental:
OK, I will do that then.

josevalim:

bin/elixir
```diff
@@ -186,14 +206,18 @@ else
   if [ $USE_WERL ]; then ERL_EXEC="werl"; fi
 fi
 
-set -- "$ERTS_BIN$ERL_EXEC" -pa "$SCRIPT_PATH"/../lib/*/ebin $ELIXIR_ERL_OPTIONS $ERL -extra "$@"
+ERTS_BIN=""
```
eksperimental: `ERTS_BIN` is not doing anything in this script, as it will always be `""`

josevalim:
It is replaced when we assemble the release.
-- 

PR #8686, opened 2019-01-23
Use defguard to define Kernel.is_nil/1

wojtekmach:


josevalim:
We probably won't be able to bootstrap. Let's see what travis-ci says. :)

wojtekmach:
oh no! It worked for me on an incremental compilation. Closing!

PR #8688, opened 2019-01-24
Make tests pass in 2020

bmwiedemann:
using only 2038 to keep 32-bit UNIX systems happy.

Background:
As part of my work on reproducible builds for openSUSE, I check that software still gives identical build results in the future.
The usual offset is +15 years, because that is how long I expect some software will be used in some places.

### DIFF
--- a/lib/mix/test/mix/tasks/compile.elixir_test.exs
+++ b/lib/mix/test/mix/tasks/compile.elixir_test.exs
@@ -123,7 +123,7 @@ defmodule Mix.Tasks.Compile.ElixirTest do
       Mix.shell().flush
       purge([A, B])
 
-      future = {{2020, 1, 1}, {0, 0, 0}}
+      future = {{2038, 1, 1}, {0, 0, 0}}
       File.touch!("lib/a.ex", future)
       Mix.Tasks.Compile.Elixir.run(["--verbose"])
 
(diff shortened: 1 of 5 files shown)


josevalim:
@eksperimental since we delegate to Erlang exclusively, that's really a conversation we need to have with the Erlang team. So can you please have this discussion there since our hands are tied on the matter? For now, we should document the behaviour though (in case it isn't).

josevalim:

eksperimental:
what's the best way to proceed? mailing list, issue tracker, or anything else..

josevalim:
@eksperimental for Erlang? I am not sure. Mailing list or a report on bugs.erlang.org.

OvermindDL1:
@eksperimental It would be nice to have a link to the erlang-side of the bug report here.  :-)

josevalim:
@eksperimental oh so it works today for 64 bit issues. So if it only fails on 32-bit OSes, then I don't even think there is anything Erlang/OTP can do, we just need to make sure it fails accordingly (i.e. with `{:error, :something}`).

eksperimental:
Erlang in the typespecs it says the year can be from 1970 to 10000 or something like that with more 0s. but it depends on the OS. That's what I wanted to check.

OvermindDL1:
'stat' results used to be 32-bit timestamps on 32-bits, and 64-bits on 64-bits, however there is a field on the extended stat call that has expanded timestamp range, if OTP is not using that one then it should be.

PR #8691, opened 2019-01-24
Ensure we're testing the actual assignment metadata

joeyates:
The existing test was not actually testing anything, as
the default argument to Keyword.get/3, 0, matched one of the
expected values.

### DIFF
--- a/lib/eex/test/eex/smart_engine_test.exs
+++ b/lib/eex/test/eex/smart_engine_test.exs
@@ -35,10 +35,13 @@ defmodule EEx.SmartEngineTest do
     Macro.prewalk(result, fn
       {_left, meta, [_, :hello]} ->
         assert Keyword.get(meta, :line) == 2
+        send(self(), :found)
 
-      _ ->
-        :ok
+      node ->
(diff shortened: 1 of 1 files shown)


lib/eex/test/eex/smart_engine_test.exs
```diff
@@ -29,16 +29,25 @@ defmodule EEx.SmartEngineTest do
...
-    end)
+  test "preserves line numbers in assignments" do
+    result = EEx.compile_string("foo\n<%= @hello %>", engine: EEx.SmartEngine)
+
+    found =
+      try do
+        Macro.prewalk(result, fn
+          {_left, meta, [_, :hello]} ->
+            assert Keyword.get(meta, :line) == 2
+            throw :found
```
josevalim: instead of throw, you can send a message to yourself and then assert it was received: `send self(), :found` and then `assert_received :found`.

josevalim:
Good catch! I dropped a comment to clean up the code a bit!

josevalim:

Issue #8689, opened 2019-01-24
.beam files differ between builds


bmwiedemann:
### Environment

* Elixir & Erlang/OTP versions (elixir --version): 
elixir-1.7.4 erlang-21.1.4
* Operating system: openSUSE Tumbleweed 20190124

### Current behavior

Somewhat similar to #4814 building the openSUSE elixir package produces various .beam files that differ for every build, e.g. `/usr/lib/elixir/lib/mix/ebin/Elixir.Mix.Tasks.Deps.Unlock.beam`

Some of the diff is even there when doing builds as similar as possible (e.g. disabling ASLR, 1-core VM without parallelism)

The erlang package itself does not have this kind of variations in its .beam files (but others from timestamps)

Note: the elixir-ex_doc package (in the best case) only has 1 beam file with 1 bit diff so might be easier to debug.

### Expected behavior

Builds should produce deterministic results.
See https://reproducible-builds.org/ for why this matters.

josevalim:
Thanks @bmwiedemann. Do you have steps to reproduce this? How are you going the builds? Which Erlang/OTP version are you using?

I am getting consistent results with Elixir master and Erlang/OTP 21. Note that earlier Erlang/OTP versions would include timestamps, which would make them not reproducible.

josevalim:
I also could not reproduce this in ExDoc:

```
~/OSS/ex_doc[master]$ mix compile --force
Compiling 18 files (.ex)
Generated ex_doc app
~/OSS/ex_doc[master]$ md5 _build/dev/lib/ex_doc/ebin/*
MD5 (_build/dev/lib/ex_doc/ebin/Elixir.ExDoc.CLI.beam) = 48e25a8df33d27d27f9f556db1b52a41
MD5 (_build/dev/lib/ex_doc/ebin/Elixir.ExDoc.Config.beam) = 28887a5beed69bcecfc76074dd7a9e3d
MD5 (_build/dev/lib/ex_doc/ebin/Elixir.ExDoc.Formatter.EPUB.Assets.beam) = cb365ebd1781bf25bd36d6e095b5eed9
MD5 (_build/dev/lib/ex_doc/ebin/Elixir.ExDoc.Formatter.EPUB.Templates.beam) = c7589f848178d992a962b2d488c86b0a
MD5 (_build/dev/lib/ex_doc/ebin/Elixir.ExDoc.Formatter.EPUB.beam) = a27fa4cfe66a6c20f71e74fced081456
[...]

PR #8694, opened 2019-01-24
Support SOURCE_DATE_EPOCH for reproducible builds

josevalim:
See #8689.

### DIFF
--- a/lib/elixir/lib/system.ex
+++ b/lib/elixir/lib/system.ex
@@ -129,8 +129,21 @@ defmodule System do
   defp revision, do: get_revision()
 
   # Get the date at compilation time.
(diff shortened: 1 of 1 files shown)


bmwiedemann:
Using `SOURCE_DATE_EPOCH` is one common approach, that will work fine.
Another is using the date of last git commit and falling back to ChangeLog file mtime if the git tree is unavailable:
https://github.com/ntop/nDPI/pull/662

eksperimental:
Changelog will not work, because it gets updated in chunks

josevalim:
Plus not all changes go to the CHANGELOG, such as documentation improvements. Also, if git is not available, we would need a fallback, so i think SOURCE_DATE_EPOCH is the way to go.

josevalim:

PR #8696, opened 2019-01-24
ExUnit.CaseTemplate moduledoc example improvement

amatalai:


josevalim:

amatalai:
@fertapric It's half-automated. I wrote few scripts (very ugly and sometimes violating private api like `Code.Typespec`; no way I would share something like this with anyone 😱  ) which returns list of possible improvements (also very ugly) for given otp_app. Then I just copy them to `bin/iex` and review outputs manually because it contains a lot of false positives (I would say 50% of all found examples)

PR #8701, opened 2019-01-25
Reproducible build

eksperimental:


eksperimental:
I'm uploading it to see if it works with Travis.

It did locally with this Bash script
https://gist.github.com/eksperimental/d44698b69066958d7bd3554e60566342

josevalim:
@eksperimental I like this a lot. I would just propose to run those changes at the end of the suite, this way we don't need to remove the trailing ebin_reproducible directories and we don't block the test results on this. However, in order to do that, we would need to run `make compile` after we remove `.git` the first time around too (which I also think is a good thing to test).

fertapric:
Would it make sense to move this into a script so you can run it locally?

bmwiedemann:
The travis.yml is already running shell commands like `make compile`, so you can add + call a `path/to/new/script`

josevalim:
We wanted to avoid adding assorted scripts to the repository but I believe the answer is that we can simply add this to the makefile for now. :D

josevalim:
@eksperimental another suggestion, that came up on #8702, is for us to have a specific entry in the test matrix that checks for reproducible long-term builds. In this matrix entry we would:

1. Do the diff check you perform in this commit right now
2. Use libfaketime to change the time and run the build "in the future"

This way we keep those concerns isolated in the CI instead of running them on every OTP version. We can use the latest OTP version on this build.

eksperimental:
@josevalim can you have a look at this and tell me what would be the cause of the error
https://travis-ci.org/elixir-lang/elixir/jobs/484534189#L755
could it be because we are not running  `make clear` ?

eksperimental:
Alright, I 'm done with all the requests. Please have a look.

The new command is `make check_replicable`
I wasn't really happy with the idea that we had to pass `SOURCE_DATE_EPOCH` so I came up with a solution.
We could even run `make test check_replicable` and it will work.

.gitignore
```diff
@@ -1,5 +1,6 @@
 /doc/
 /lib/*/ebin/
+/lib/*/ebin_reproducible/
```
josevalim: Given that we move it to the makefile, then I believe it is best for us to remove the directories later. Another idea is to write the ebins to `lib/*/tmp/ebin_reproducible`. Thoughts?

eksperimental: still the task can be interrupted, and the files are going to be left behind.
I will look into the lib/*/tmp/ebin_reproducible option...

josevalim: Sorry, the comment above doesn't make sense. I wanted to say I don't like to have them potentially stay around in case of failures, which is why I would prefer to have them in `tmp`. :)

Makefile
```diff
@@ -113,11 +113,40 @@ install: compile
...
+                      lib/eex/ebin_reproducible/ \
+	              lib/iex/ebin_reproducible/ \
+	              lib/logger/ebin_reproducible/ \
+	              lib/mix/ebin_reproducible/
+	$(Q) mv lib/elixir/ebin/* lib/elixir/ebin_reproducible/
+	$(Q) mv lib/eex/ebin/* lib/eex/ebin_reproducible/
+	$(Q) mv lib/iex/ebin/* lib/iex/ebin_reproducible/
+	$(Q) mv lib/logger/ebin/* lib/logger/ebin_reproducible/
+	$(Q) mv lib/mix/ebin/* lib/mix/ebin_reproducible/
+	SOURCE_DATE_EPOCH="$(call SOURCE_DATE_EPOCH)" $(MAKE) compile
```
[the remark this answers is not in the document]
eksperimental: I cannot make it work on Travis.
I had pushed a commit right before you mention this, exporting it.\
if it works, i will see If I can avoid passing the variable explicitely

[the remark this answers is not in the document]
josevalim: Yeah, i saw the new commit, I think you have to remove the call here for the export to work. I just think though... :D  not sure!

.travis.yml
```diff
@@ -4,16 +4,17 @@ sudo: false
 env:
   global:
     - ELIXIR_ASSERT_TIMEOUT=2000
+    - LATEST_OTP_RELEASE=false
```
josevalim: I would suggest not setting this default value here. And then you can do this in the script:

    - if [ -n CHECK_REPRODUCIBLE ]; then make check_reproducible; fi

See that I am also proposing to rename the variable, as I think that better reveals the intent. Thoughts?

eksperimental: sure. I think I chose this one when we were going to check if it was reproducible and also the future test..

josevalim:
> @josevalim can you have a look at this and tell me what would be the cause of the error
https://travis-ci.org/elixir-lang/elixir/jobs/484534189#L755
could it be because we are not running make clear ?

I am not sure why. If I had to guess, it changed the time for the VM but not for the filesystem operations.

.gitignore
eksperimental: > > @josevalim can you have a look at this and tell me what would be the cause of the error
> > https://travis-ci.org/elixir-lang/elixir/jobs/484534189#L755
> > could it be because we are not running make clear ?
> 
> I am not sure why. If I had to guess, it changed the time for the VM but not for the filesystem operations.

don't worry about it for now, was we are not getting that anymore.

Makefile
```diff
@@ -113,6 +113,35 @@ install: compile
 	done
 	$(MAKE) install_man
 
+check_reproducible: compile
+	$(Q) echo "==> Checking for reproducible builds..."
+	export SOURCE_DATE_EPOCH="$(shell bin/elixir -e \
```
josevalim: I think you don't need export if you pass it explicitly later :D

Makefile
```diff
@@ -113,6 +113,35 @@ install: compile
...
+	$(Q) mv lib/logger/ebin/* lib/logger/tmp/ebin_reproducible/
+	$(Q) mv lib/mix/ebin/* lib/mix/tmp/ebin_reproducible/
+	SOURCE_DATE_EPOCH="${SOURCE_DATE_EPOCH}" $(MAKE) compile
+	$(Q) echo "Diffing..."
+	$(Q) diff -r lib/elixir/ebin/ lib/elixir/tmp/ebin_reproducible/
+	$(Q) diff -r lib/eex/ebin/ lib/eex/tmp/ebin_reproducible/
+	$(Q) diff -r lib/iex/ebin/ lib/iex/tmp/ebin_reproducible/
+	$(Q) diff -r lib/logger/ebin/ lib/logger/tmp/ebin_reproducible/
+	$(Q) diff -r lib/mix/ebin/ lib/mix/tmp/ebin_reproducible/
+	$(Q) $(MAKE) clean_reproducible_files
```
josevalim: We don't need this call now (nor the task) as we do leave dirty stuff in `tmp`. That's why it exists! :D

Makefile
eksperimental: the variable is not getting stored, and on the next line I end up with 
`SOURCE_DATE_EPOCH=""`

eksperimental: there you go:
https://travis-ci.org/elixir-lang/elixir/jobs/484682208#L664-L669

josevalim: Oh, it probably executes each line as a separate script? So it is not like a bash script? That makes sense.

eksperimental: then we can either run the whole chunk  in one command, or store in a file in /tmp the SOURCE_DATE_EPOCH

eksperimental: I will try to finish this later, as I've got to go now. Thank you

josevalim: Nah, it is perfectly fine as is. My comment is more me trying to wrap my head around makefiles (it has been a while) than anything else, sorry!

Makefile
```diff
@@ -113,6 +121,29 @@ install: compile
 	done
 	$(MAKE) install_man
 
+check_reproducible: compile
+	$(Q) echo "==> Checking for reproducible builds..."
+	$(Q) rm -rf lib/*/tmp/ebin_reproducible
+	$(Q) echo "Moving files to ebin_reproducible/ ..."
```
fertapric: Minor nitpick
```suggestion
	$(Q) echo "Moving files to lib/*/tmp/ebin_reproducible/ ..."
```

josevalim:
@eksperimental I pushed a commit where we can compute the epoch correctly but make is failing for some reason.

Makefile
eksperimental: done! thank you

eksperimental:
@josevalim I'm running ubuntu, but a later version than the one in CI, and this is working differently. exporting variables work, maybe it is the GNUMake version that Travis is running.

The problem with your solution is that we need to determine SOURCE_DATE_EPOCH before moving the files from ebin to ebin_reproducible, and then use this value again when compiling the last time. 

As I wrote this the new implementation seems to work. So, have a look please

Makefile
eksperimental: Make is a beast hard to tame. If anybody can recommend a good reading I would greatly appreciate  it

eksperimental:
This is ready to be merged!

Makefile
```diff
@@ -46,6 +48,18 @@ test_$(1): compile $(1)
 	$(Q) cd lib/$(1) && ../../bin/elixir -r "test/test_helper.exs" -pr "test/**/*_test.exs";
 endef
 
+define write_SOURCE_DATE_EPOCH
```
josevalim: It seems that our convention here is to have those all in uppercase, so what about WRITE_SOURCE_DATE_EPOCH and READ_SOURCE_DATE_EPOCH?

Makefile
```diff
@@ -113,6 +127,30 @@ install: compile
 	done
 	$(MAKE) install_man
 
+check_reproducible: compile
+	$(Q) echo "==> Checking for reproducible builds..."
+	$(Q) rm -rf lib/*/tmp/ebin_reproducible/
+	$(Q) echo "Moving files lib/*/tmp/ebin_reproducible/ ..."
```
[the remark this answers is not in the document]
josevalim: Actually, I would remove this echo. Not adding much information. :)

josevalim:
Two minor comments and we should be good to go!

eksperimental:
> Two minor comments and we should be good to go!

done!

Issue #8706, opened 2019-01-25
Allow to specify a `format` argument to `date_to_string`, `datetime_to_string` and `naive_datetime_to_string` 


fertapric:
**Note: this is blocked by https://github.com/elixir-lang/elixir/issues/8705**

`Calendar.ISO.time_to_string/3,4` allows to specify a `format` argument but similar functions (`date_to_string`, `datetime_to_string` and `naive_datetime_to_string` ) in that module do not.

PR #8707, opened 2019-01-25
Use default args in Calendar.ISO.time_to_string/*

### LINKED ISSUE #8705: Do not emit a warning if @impl is used in a function with default arguments
### Environment

```
Erlang/OTP 21 [erts-10.2.1] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [hipe] [dtrace]

Elixir 1.9.0-dev (0c89a94c6) (compiled with Erlang/OTP 21)
```
* Operating system: OS X

### Current behavior

As stated by [this comment](https://github.com/elixir-lang/elixir/pull/6331#discussion_r127476966), if `@impl` is used in a function with default arguments, a warning will be emitted because a callback might not be specified for some of the arities.

In the case of `Calendar.ISO.time_to_string/4,5` the warning was:

```
warning: module attribute @impl was not set for function time_to_string/4 callback (specified in Calendar). This either means you forgot to add the "@impl true" annotation before the definition or that you are accidentally overriding this callback
  lib/calendar/iso.ex:509: Calendar.ISO (module)
```

### Expected behavior

No warnings

fertapric:
Closes #8705 and #8706

### DIFF
--- a/lib/elixir/lib/calendar/iso.ex
+++ b/lib/elixir/lib/calendar/iso.ex
@@ -494,11 +494,12 @@ defmodule Calendar.ISO do
         ) :: String.t()
   def time_to_string(hour, minute, second, microsecond, format \\ :extended)
 
-  def time_to_string(hour, minute, second, {_, 0}, format) do
+  def time_to_string(hour, minute, second, {_, 0}, format) when format in [:basic, :extended] do
     time_to_string_format(hour, minute, second, format)
   end
 
-  def time_to_string(hour, minute, second, {microsecond, precision}, format) do
+  def time_to_string(hour, minute, second, {microsecond, precision}, format)
+      when format in [:basic, :extended] do
     time_to_string_format(hour, minute, second, format) <>
       "." <> (microsecond |> zero_pad(6) |> binary_part(0, precision))
   end
@@ -514,6 +515,10 @@ defmodule Calendar.ISO do
   @doc """
   Converts the given date into a string.
 
+  By default, returns dates formatted in the "extended" format,
+  for human readability. It also supports the "basic" format
+  through passing the `:basic` option.
+
   ## Examples
 
       iex> Calendar.ISO.date_to_string(2015, 2, 28)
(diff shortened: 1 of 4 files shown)


lib/elixir/lib/calendar/iso.ex
```diff
@@ -464,6 +464,10 @@ defmodule Calendar.ISO do
   @doc """
   Converts the given time into a string.
 
+  By default, returns times formatted in the "extended" format,
+  for human readability. It also supports the "basic" format
+  through passing the `:basic` option.
```
eksperimental:
```suggestion
  by passing the `:basic` option.
```

lib/elixir/lib/calendar/iso.ex
```diff
@@ -525,6 +515,10 @@ defmodule Calendar.ISO do
   @doc """
   Converts the given date into a string.
 
+  By default, returns dates formatted in the "extended" format,
+  for human readability. It also supports the "basic" format
+  through passing the `:basic` option.
```
eksperimental:
```suggestion
  by passing the `:basic` option.
```

lib/elixir/lib/calendar/iso.ex
```diff
@@ -534,32 +528,42 @@ defmodule Calendar.ISO do
...
+  def date_to_string(year, month, day, :basic) do
     zero_pad(year, 4) <> zero_pad(month, 2) <> zero_pad(day, 2)
   end
 
   @doc """
   Converts the datetime (without time zone) into a string.
 
+  By default, returns datetimes formatted in the "extended" format,
+  for human readability. It also supports the "basic" format
+  through passing the `:basic` option.
```
eksperimental:
```suggestion
  by passing the `:basic` option.
```

lib/elixir/lib/calendar/iso.ex
```diff
@@ -570,15 +574,31 @@ defmodule Calendar.ISO do
...
+    date_to_string(year, month, day, format) <>
+      " " <> time_to_string(hour, minute, second, microsecond, format)
   end
 
   @doc """
   Converts the datetime (with time zone) into a string.
 
+  By default, returns datetimes formatted in the "extended" format,
+  for human readability. It also supports the "basic" format
+  through passing the `:basic` option.
```
eksperimental:
```suggestion
  by passing the `:basic` option.
```

lib/elixir/lib/calendar/iso.ex
fertapric: This was copied from here:

https://github.com/elixir-lang/elixir/blob/3bd3f88d57d7fff6cab7b171294b89fb08eedfe7/lib/elixir/lib/calendar/datetime.ex#L651-L652

We should probably change those in a different PR

PR #8736, opened 2019-01-29
Make bin/elixir POSIX compliant

eksperimental:


bin/elixir
```diff
@@ -69,8 +69,17 @@ ERL=""
...
+  E=$((E + 1))
+}
+
+# Checks if a line starts with char
+# starts_with "$LINE", "$CHAR"
+starts_with () {
+case $1 in
+  "$2"*) true;;
+  *) false;;
+esac
```
josevalim: I think the previous version clearer. Can we please revert this one?

bin/elixir
```diff
@@ -159,44 +168,44 @@ while [ $I -le $LENGTH ]; do
...
-  I=$(($I - 1))
+  I=$((I - 1))
 done
 
 SELF=$(readlink_f "$0")
 SCRIPT_PATH=$(dirname "$SELF")
 
+# $OSTYPE is not POSIX compliant
+# https://github.com/pstadler/metrics.sh/blob/master/lib/utils/os.sh
+if [ -z "$OSTYPE" ]; then OSTYPE=$(uname | tr '[:upper:]' '[:lower:]'); fi
```
josevalim: I don't think it matters because we are only doing decisions for cygwin and windows, where those are certainly set. :)

bin/elixir
```diff
@@ -210,7 +219,8 @@ set -- "$ERTS_BIN$ERL_EXEC" -pa "$SCRIPT_PATH"/../lib/*/ebin $ELIXIR_ERL_OPTIONS
 if [ -n "$RUN_ERL_PIPE" ]; then
   ESCAPED=""
   for PART in "$@"; do
-    ESCAPED=""$ESCAPED" $(echo "$PART" | sed 's/[^a-zA-Z0-9_\-\/]/\\&/g')"
+    PART_ESCAPED="$(echo "$PART" | sed 's/[^a-zA-Z0-9_\-\/]/\\&/g')"
+    ESCAPED="$(printf "%s %s" "$ESCAPED" "$PART_ESCAPED")"
```
josevalim: Can you please explain the difference?

eksperimental: The linter complains about the quotes at the beggining. I would prefer if we could stick to it.

bin/elixir
yordis: I prefer this change since it is more clear to me and easier to contribute to, subjectively speaking.

eksperimental:
Ideally I would like to verify the bin scripts with shellcheck in the CI.
What's the opinion of the core team on the matter?

bin/elixir
josevalim: I don't think we need the double quotes in $ESCAPED. Would this be enough: `ESCAPED="$ESCAPED $(echo "$PART" | sed 's/[^a-zA-Z0-9_\-\/]/\\&/g')"`?

josevalim:
@eksperimental how can we install shellcheck on CI?

bin/elixir
```diff
@@ -69,8 +69,16 @@ ERL=""
...
+  eval "E${E}=\$1"
+  E=$((E + 1))
+}
+
+# Checks if a line starts with char starts_with "$LINE", "$CHAR"
+starts_with () {
+  case $1 in
+    "$2"*) true;;
+    *) false;;
+  esac
```
josevalim: I indented this and one-lined the comment.

bin/elixir
```diff
@@ -132,7 +140,7 @@ while [ $I -le $LENGTH ]; do
         ;;
     --sname|--name)
         S=2
-        erl $(echo $1 | cut -c 2-)
+        erl "$(echo "$1" | cut -c 2-)"
```
josevalim: Good catch!

eksperimental:
> @eksperimental how can we install shellcheck on CI?

from the official docs https://github.com/koalaman/shellcheck#travis-ci

> Travis CI
> Travis CI has now integrated ShellCheck by default, so you don't need to manually install it.
> 
> If you still want to do so in order to upgrade at your leisure or ensure the latest release, follow the steps to install the shellcheck binary, bellow.

I didn't know!

bin/elixir
eksperimental: yes, this is what shellcheck is really good at.

bin/elixir
eksperimental: ok! great, and thank you @yordis for your input

josevalim:
Beautiful. Let’s add a flag to CI then, like CHECK_REPRODUCIBLE but for the
scripts and run them together. Let’s do it in another PR though!
-- 


bin/elixir
eksperimental: True. But if we are going to stick with the shellcheck linter, I would recommend to leave it. We could ignore it by passing the error code to the command line as we already do, but if in the future, somebody decides to improve the script and uses the variable it. it will not work in strictly POSIX shells.
Regardless of the linter warnings, by adding this line we can claim our script is POSIX compliant

josevalim:
I don’t agree. :) then we are adhering to the rules for the rules sake,
even if they don’t have a purpose. Our script will be POSIX compliant even
without this change. Can we rewrite this to something that checks if the
variable is defined and then assert on the value? Would that make the
linter happy?
-- 


bin/elixir
eksperimental: I think so.
I have pushed this suggestion, but tests are failing locally (most likely unrelated to this change)

josevalim: @eksperimental it is unrelated but it may have left some VMs hanging around, so you need to kill them before. I will fix them!

bin/elixir
josevalim: But someone else can set any variable externally, no? how do we handle variables that are set externally then? To be clear, `if [ -n $OSTYPE ] && [ "$OSTYPE" = "cygwin" ]` is not enough?

josevalim: I just checked locally and it is not enough indeed. :(

josevalim: So I checked shellcheck source and skipping 2039 will only skip reserved variables, so i think it is safe to skip it! :+1:

eksperimental: this one works too
```
# shellcheck disable=SC2039
if [ -z $OSTYPE ] && [ "$OSTYPE" = "cygwin" ]; then SCRIPT_PATH=$(cygpath -m "$SCRIPT_PATH"); fi
```

eksperimental: @josevalim I have pushed a last commit. It's three comments, as we save the trouble of bypassing error codes that may skip useful checks. It think it's a good trade-off

bin/elixir
```diff
@@ -159,58 +168,57 @@ while [ $I -le $LENGTH ]; do
...
-  I=$(($I - 1))
+  I=$((I - 1))
 done
 
 SELF=$(readlink_f "$0")
 SCRIPT_PATH=$(dirname "$SELF")
 
-if [ "$OSTYPE" = "cygwin" ]; then SCRIPT_PATH=$(cygpath -m "$SCRIPT_PATH"); fi
+# shellcheck disable=SC2039,SC2086
+if [ -n "$OSTYPE" ] && [ "$OSTYPE" = "cygwin" ]; then SCRIPT_PATH=$(cygpath -m "$SCRIPT_PATH"); fi
```
josevalim: Let's keep this one as it was then, as the `-n` is not buying us anything. :)

josevalim:
Sorry, by disabling, i meant on the command line when invoking shellcheck. I am really not a big fan of adding comments to the code to appease linters/compilers. :) So can we please revert those?

bin/elixir
eksperimental: done and ready to be merged

bin/elixir
```diff
@@ -2,29 +2,29 @@
 set -e
 
 if [ $# -eq 0 ] || [ "$1" = "--help" ] || [ "$1" = "-h" ]; then
-  echo "Usage: $(basename $0) [options] [.exs file] [data]
-
+  printf "Usage: %s [options] [.exs file] [data]\n" "$(basename "$0")"
+  printf '
```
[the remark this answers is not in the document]
eksperimental: It was to make the doc look as it is going to look on screen. IMO it is a bit messy with all the unquoting, and things don't look aligned.
I think using printf to inject variables and output from commands is less error prone that using double quotes which which hold other pair of double quotes and go inside the string.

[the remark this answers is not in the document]
eksperimental: I have pushed a new commit that replaces `printf '...'` with `echo '...'`

josevalim:
I see two concerns:

1. I would say that single quotes are not very straightforward either when
it comes to escaping.

2. The commit says it is about POSIX but those changes are not POSIX
related. So it may be confusing on a future review. In fact, they add new
warnings on shellcheck, which is a bit counter intuitive with the commit
goals

Thoughts?
-- 


eksperimental:
I reverted the changes related to the USAGE message.

now these are the only messages by `shellcheck`

```
$ shellcheck bin/elixir

In bin/elixir line 204:
if [ "$OSTYPE" = "cygwin" ]; then SCRIPT_PATH=$(cygpath -m "$SCRIPT_PATH"); fi
      ^-----^ SC2039: In POSIX sh, OSTYPE is undefined.


In bin/elixir line 212:
set -- "$ERTS_BIN$ERL_EXEC" -pa "$SCRIPT_PATH"/../lib/*/ebin $ELIXIR_ERL_OPTIONS $ERL "$@"
                                                             ^-----------------^ SC2086: Double quote to prevent globbing and word splitting.
                                                                                 ^--^ SC2086: Double quote to prevent globbing and word splitting.

For more information:
  https://www.shellcheck.net/wiki/SC2039 -- In POSIX sh, OSTYPE is undefined.
  https://www.shellcheck.net/wiki/SC2086 -- Double quote to prevent globbing ...
```

josevalim:
Absolutely beautiful. I learned a lot, thanks @eksperimental! 

PR #8742, opened 2019-01-30
Include field names in generated type for records

wojtekmach:
For this module:

    defmodule M do
      import Record
      @type t :: record(:r, foo: atom(), bar: integer())
      defrecord :r, [:foo, :bar]
    end

`@type t` is defined as the following:

Before this patch:

    @type t() :: {:r, atom(), integer()}

After this patch:

    @type t() :: {:r, foo :: atom(), bar :: integer()}

This is similar to loading record definition in Erlang shell:

    1> rr(file).
    [file_descriptor,file_info]
    2> rl(file_descriptor).
    -record(file_descriptor,{module :: module(),data :: term()}).
    ok


josevalim:

PR #8745, opened 2019-01-30
Add function specs to Mix public interface

saulecabrera:
This change adds function specs to `Mix`.

I'm not completely sure if the specs for `Mix.shell/0` and `Mix.shell/1` are correct; I've used `module` as return type and parameter type respectively given that I'm not sure if it's possible to express the contract between a module and `Mix.Shell` in a more specific way.


josevalim:

PR #8751, opened 2019-02-01
Reverse order of OTP releases in .travis.yml

eksperimental:
The rationale is that the last version checks for
reproducible build and will also do other linting checks in the future
while the rest of the OTP releases don't.
So if it's going to fail, we don't need to wait until the end of the CI suite.


josevalim:

PR #8752, opened 2019-02-01
CI: Check for .travis.yml validity

eksperimental:


.travis.yml
```diff
@@ -35,3 +36,5 @@ script:
   - dialyzer -pa lib/elixir/ebin --build_plt --output_plt elixir.plt --apps lib/elixir/ebin/elixir.beam lib/elixir/ebin/Elixir.Kernel.beam
   # Check for reproducible builds only in the latest OTP release
   - if [ -n "$CHECK_REPRODUCIBLE" ]; then make check_reproducible; fi
+  # Check for .travis.yml validity
+  - if [ -n "$CHECK_TRAVIS_YML" ]; then echo "y" | travis lint .travis.yml; fi
```
fertapric: I think we need to use the `-x` or `--exit-code` option so the exit code is 1 if there are warnings:

```
-x, --[no-]exit-code             sets the exit code to 1 if there are warning
```

Why is the `echo "y"` needed? I couldn't manage to get a prompt while using the `travis lint` command.

josevalim:
Thanks @eksperimental but I think we are fine without this. We only had one minor mistake after all this time and it made no impact. If there is something really wrong, I am hoping that travis won't even run our suite. :)

PR #8761, opened 2019-02-03
Improve List.ascii_printable?/2 spec

eksperimental:


josevalim:
This also has overlapping domains. :) Specs do not work like functions, a
clause in one spec cannot have a subset or superset of another clause.
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

eksperimental:
I was aware of this, and I and the same dialyze command we use in travis, and not error.
Which one is the overlapping domain?

josevalim:
The dialyze command in Travis is just a smoke test. And even if dialyzer
doesn’t warn, we shouldn’t merge something we know is semantically
incorrect.
-- 


eksperimental:
I have run it in a test project, and I see no errors
https://github.com/eksperimental/test_overlapping

If I change `pos_integer` to `nonneg_integer` I see the overlapping domain errors. If you please be more specific about the case that is overlapping, I can look into it.

josevalim:

eksperimental:
Awesomeness!

PR #8762, opened 2019-02-03
User "true or false" instead of the other way around

eksperimental:
While "false or true" is alphabetically ordered, the most common
way it is used in the English language is "true or false" and this
is the only one reference in Elixir core that we say "false or true".


josevalim:

PR #8774, opened 2019-02-06
Use "code point" instead of "codepoint"

eksperimental:
That's the term it is used in the Unicode standard.

/cc @fertapric


josevalim:

PR #8775, opened 2019-02-06
Reword docs about Collectable.into/1

eksperimental:
Be explicit that we are talking about Collectable.into/1
and Enumerable.reduce/3

### DIFF
--- a/lib/elixir/lib/collectable.ex
+++ b/lib/elixir/lib/collectable.ex
@@ -21,9 +21,9 @@ defprotocol Collectable do
   shape where just the range limits are stored.
 
   The `Collectable` module was designed to fill the gap left by the
-  `Enumerable` protocol. `into/1` can be seen as the opposite of
-  `Enumerable.reduce/3`. If `Enumerable` is about taking values out,
-  `Collectable.into/1` is about collecting those values into a structure.
+  `Enumerable` protocol. `Collectable.into/1` can be seen as the opposite of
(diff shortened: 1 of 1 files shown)


eksperimental:
This one has been in my repo for about 2 and half years

lib/elixir/lib/collectable.ex
```diff
@@ -21,9 +21,9 @@ defprotocol Collectable do
   shape where just the range limits are stored.
 
   The `Collectable` module was designed to fill the gap left by the
-  `Enumerable` protocol. `into/1` can be seen as the opposite of
-  `Enumerable.reduce/3`. If `Enumerable` is about taking values out,
-  `Collectable.into/1` is about collecting those values into a structure.
+  `Enumerable` protocol. `Collectable.into/1` can be seen as the opposite of
+  `Enumerable.reduce/3`. The former is about taking values out, while the
```
whatyouhide: I think using former and latter tends to make things more confusing especially for non-native speakers. Case in point, you switched former and latter up because the former is `Collectable.into/1` which is about collecting values into a structure, and the latter is `Enumerable.reduce/3`. Tbh, I think the documentation we had before was clearer than this so my take is that we can leave it like it was before.

eksperimental: have a second look. I have addressed those issues

josevalim:

PR #8776, opened 2019-02-06
Clearer documentation for GenServer timeouts

nathanl:
I thought it was unclear what would happen if a timeout of `0` were set
but messages were already waiting.

### DIFF
--- a/lib/elixir/lib/gen_server.ex
+++ b/lib/elixir/lib/gen_server.ex
@@ -414,7 +414,7 @@ defmodule GenServer do
 
   Returning `{:ok, state, timeout}` is similar to `{:ok, state}`,
   except that it also sets a timeout. See the "Timeouts" section
-  in the module documentation.
+  in the module documentation for more information.
 
   Returning `{:ok, state, :hibernate}` is similar to `{:ok, state}`
   except the process is hibernated before entering the loop. See
--- a/lib/elixir/lib/gen_server.ex
+++ b/lib/elixir/lib/gen_server.ex
@@ -251,16 +251,17 @@ defmodule GenServer do
   ## Timeouts
 
   The return value of `c:init/1` or any of the `handle_*` callbacks may include
-  a timeout value in milliseconds; if not, `:infinity` is assumed. Each time a
-  timeout is set, it replaces the previously set timeout.
+  a timeout value in milliseconds; if not, `:infinity` is assumed.
+  The timeout can be used to detect a lull in incoming messages.
 
   If the process has no messages waiting when the timeout is set and the
   number of given milliseconds pass without any message arriving,
   then `handle_info/2` will be called with `:timeout` as the first argument.
+  The timeout is cleared if any message is waiting or arrives before the timeout.
 
-  Because the timeout is cleared by each received message, even a timeout of `0` milliseconds
-  is not guaranteed to execute. To take another action immediately and
-  unconditionally, use a `:continue` instruction.
+  Because a message may arrive before the timeout is set, even a timeout of `0`
+  milliseconds is not guaranteed to execute. To take another action immediately
+  and unconditionally, use a `:continue` instruction.
 
   ## When (not) to use a GenServer
 
--- a/lib/elixir/lib/gen_server.ex
+++ b/lib/elixir/lib/gen_server.ex
@@ -260,7 +260,7 @@ defmodule GenServer do
 
   Because the timeout is cleared by each received message, even a timeout of `0` milliseconds
   is not guaranteed to execute. To take another action immediately and
-  unconditionally, use a `continue` instruction.
+  unconditionally, use a `:continue` instruction.
 
   ## When (not) to use a GenServer
 
--- a/lib/elixir/lib/gen_server.ex
+++ b/lib/elixir/lib/gen_server.ex
@@ -464,7 +464,7 @@ defmodule GenServer do
 
   Returning `{:reply, reply, new_state, timeout}` is similar to
   `{:reply, reply, new_state}` except that it also sets a timeout.
-  See the "Timeouts" section in the module documentation.
+  See the "Timeouts" section in the module documentation for more information.
 
   Returning `{:reply, reply, new_state, :hibernate}` is similar to
   `{:reply, reply, new_state}` except the process is hibernated and will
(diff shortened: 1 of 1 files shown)


nathanl:
Here's a little `GenServer` process I used to test how this works.

```elixir
defmodule Timeout do
  use GenServer

  def start_link(_) do
    GenServer.start_link(__MODULE__, 0, name: __MODULE__)
  end

  def init(count) do
    {:ok, count, 1_000}
  end

  def yo(server \\ __MODULE__) do
    GenServer.cast(server, :yo)
  end

  def handle_cast(:yo, count) do
    IO.puts("processing a message - count is #{count}")
    # simulate work
    :timer.sleep(1_000)

    case count do
      0 ->
        IO.puts("setting timeout of :infinity")
        {:noreply, count + 1, :infinity}

      1 ->
        IO.puts("setting timeout of 2 seconds")
        {:noreply, count + 1, 2_000}

      _n ->
        IO.puts("setting a timeout of 0")
        {:noreply, count + 1, 0}
    end
  end

  def handle_info(:timeout, count) do
    IO.puts("timeout!")
    {:noreply, 0}
  end
end
```
With that in a supervision tree, in `iex -S mix`, call `Timeout.yo` 0 times, once, or repeatedly, then wait to see when the `handle_info` is called.

lib/elixir/lib/gen_server.ex
```diff
@@ -248,6 +248,16 @@ defmodule GenServer do
         end
       end
 
+  ## Timeouts
+
+  The return value of `c:init/1` or any of the `handle_` callbacks may include
+  a timeout value in milliseconds; if not, `:infinity` is assumed. Each time a
+  timeout is set, it replaces the previously-set timeout.
```
eksperimental:
```suggestion
  timeout is set, it replaces the previously set timeout.
```

lib/elixir/lib/gen_server.ex
```diff
@@ -248,6 +248,16 @@ defmodule GenServer do
         end
       end
 
+  ## Timeouts
+
+  The return value of `c:init/1` or any of the `handle_` callbacks may include
```
eksperimental:
```suggestion
  The return value of `c:init/1` or any of the `handle_*` callbacks may include
```

nathanl:
@eksperimental Thanks for the feedback. I added one more paragraph to discourage using a timeout when `handle_continue` is better.

lib/elixir/lib/gen_server.ex
```diff
@@ -248,6 +248,20 @@ defmodule GenServer do
         end
       end
 
+  ## Timeouts
+
+  The return value of `c:init/1` or any of the `handle_*` callbacks may include
+  a timeout value in milliseconds; if not, `:infinity` is assumed. Each time a
+  timeout is set, it replaces the previously set timeout.
+
+  If the process has no messages waiting when the timeout is set, and if the
```
whatyouhide:
```suggestion
  If the process has no messages waiting when the timeout is set and the
```

lib/elixir/lib/gen_server.ex
```diff
@@ -248,6 +248,20 @@ defmodule GenServer do
...
+  number of millseconds specified pass without any message arriving,
```
whatyouhide:
```suggestion
  number of given milliseconds pass without any message arriving,
```

lib/elixir/lib/gen_server.ex
```diff
@@ -248,6 +248,20 @@ defmodule GenServer do
...
+  `handle_info(:timeout, new_state)` will be called.
```
whatyouhide:
```suggestion
  then `handle_info/2` will be called with `:timeout` as the first argument.
```

lib/elixir/lib/gen_server.ex
```diff
@@ -248,6 +248,20 @@ defmodule GenServer do
...
+
+  Because the timeout is cleared by each message received, even a timeout of `0`
```
whatyouhide:
```suggestion
  Because the timeout is cleared by each received message, even a timeout of `0` milliseconds
```

lib/elixir/lib/gen_server.ex
```diff
@@ -248,6 +248,20 @@ defmodule GenServer do
...
+  is not guaranteed to execute. To take another action immediately and
+  unconditionally, use a `continue` instruction.
```
whatyouhide:
```suggestion
  unconditionally, use a `:continue` instruction.
```

nathanl:
I pushed another commit to clarify what a timeout is actually useful for:

> The timeout can be used to detect a lull in incoming messages.

... and to clarify the reason that a `0` timeout isn't reliable:

> Because a message may arrive before the timeout is set, even a timeout of `0`
  milliseconds is not guaranteed to execute.

lib/elixir/lib/gen_server.ex
```diff
@@ -248,6 +248,21 @@ defmodule GenServer do
...
+  ## Timeouts
+
+  The return value of `c:init/1` or any of the `handle_*` callbacks may include
+  a timeout value in milliseconds; if not, `:infinity` is assumed.
+  The timeout can be used to detect a lull in incoming messages.
+
+  If the process has no messages waiting when the timeout is set and the
+  number of given milliseconds pass without any message arriving,
+  then `handle_info/2` will be called with `:timeout` as the first argument.
+  The timeout is cleared if any message is waiting or arrives before the timeout.
```
eksperimental: before given timeout.

lib/elixir/lib/gen_server.ex
```diff
@@ -396,9 +412,9 @@ defmodule GenServer do
   Returning `{:ok, state}` will cause `start_link/3` to return
   `{:ok, pid}` and the process to enter its loop.
 
-  Returning `{:ok, state, timeout}` is similar to `{:ok, state}`
-  except `handle_info(:timeout, state)` will be called after `timeout`
-  milliseconds if no messages are received within the timeout.
+  Returning `{:ok, state, timeout}` is similar to `{:ok, state}`,
+  except that it also sets a timeout. See the "Timeouts" section
+  in the module documentation.
```
fertapric:
```suggestion
  in the module documentation for more information.
```

(same suggestion for the ones below)

nathanl:
@fertapric 👍 Updated

josevalim:

PR #8778, opened 2019-02-07
Use as_boolean/1 type in Enum.split_with/1

eksperimental:


lib/elixir/lib/enum.ex
```diff
@@ -1761,7 +1761,7 @@ defmodule Enum do
 
   """
   @doc since: "1.4.0"
-  @spec split_with(t, (element -> any)) :: {list, list}
+  @spec split_with(t, (element -> as_boolean(term))) :: {list, list}
```
eksperimental: or should it be `as_boolean(element)` ?

josevalim: Not really since the return value may have no relationship with the element.

josevalim:

PR #8779, opened 2019-02-07
Improvements to the URI module

eksperimental:
This one has been laying around for over 3 years!

### DIFF
--- a/lib/elixir/lib/uri.ex
+++ b/lib/elixir/lib/uri.ex
@@ -297,8 +297,8 @@ defmodule URI do
   """
   @spec encode_www_form(binary) :: binary
   def encode_www_form(string) when is_binary(string) do
-    for <<char <- string>>, into: "" do
-      case percent(char, &char_unreserved?/1) do
+    for <<byte <- string>>, into: "" do
+      case percent(byte, &char_unreserved?/1) do
         "%20" -> "+"
         percent -> percent
       end
--- a/lib/elixir/lib/uri.ex
+++ b/lib/elixir/lib/uri.ex
@@ -280,10 +280,10 @@ defmodule URI do
       "a str%69ng"
 
   """
-  @spec encode(binary, (char -> as_boolean(term))) :: binary
+  @spec encode(binary, (byte -> as_boolean(term))) :: binary
   def encode(string, predicate \\ &char_unescaped?/1)
       when is_binary(string) and is_function(predicate, 1) do
-    for <<char <- string>>, into: "", do: percent(char, predicate)
+    for <<byte <- string>>, into: "", do: percent(byte, predicate)
   end
 
   @doc """
--- a/lib/elixir/lib/uri.ex
+++ b/lib/elixir/lib/uri.ex
@@ -616,7 +616,6 @@ defmodule URI do
 end
 
 defimpl String.Chars, for: URI do
-  @spec to_string(URI.t()) :: binary
   def to_string(%{scheme: scheme, port: port, path: path, query: query, fragment: fragment} = uri) do
     uri =
       case scheme && URI.default_port(scheme) do
--- a/lib/elixir/lib/uri.ex
+++ b/lib/elixir/lib/uri.ex
@@ -17,13 +17,13 @@ defmodule URI do
             port: nil
 
   @type t :: %__MODULE__{
-          scheme: nil | String.t(),
-          path: nil | String.t(),
-          query: nil | String.t(),
-          fragment: nil | String.t(),
-          authority: nil | String.t(),
-          userinfo: nil | String.t(),
-          host: nil | String.t(),
+          scheme: nil | binary,
+          path: nil | binary,
+          query: nil | binary,
+          fragment: nil | binary,
+          authority: nil | binary,
+          userinfo: nil | binary,
+          host: nil | binary,
           port: nil | :inet.port_number()
         }
 
@@ -45,7 +45,7 @@ defmodule URI do
       nil
 
   """
-  @spec default_port(String.t()) :: nil | non_neg_integer
+  @spec default_port(binary) :: nil | non_neg_integer
   def default_port(scheme) when is_binary(scheme) do
     :elixir_config.get({:uri, scheme}, nil)
   end
@@ -62,7 +62,7 @@ defmodule URI do
   application's start callback in case you want to register
   new URIs.
   """
-  @spec default_port(String.t(), non_neg_integer) :: :ok
+  @spec default_port(binary, non_neg_integer) :: :ok
   def default_port(scheme, port) when is_binary(scheme) and is_integer(port) and port >= 0 do
     :elixir_config.put({:uri, scheme}, port)
   end
@@ -92,7 +92,7 @@ defmodule URI do
       ** (ArgumentError) encode_query/1 values cannot be lists, got: [:a, :list]
 
   """
-  @spec encode_query(Enum.t()) :: String.t()
+  @spec encode_query(Enum.t()) :: binary
   def encode_query(enumerable) do
     Enum.map_join(enumerable, "&", &encode_kv_pair/1)
   end
@@ -128,9 +128,7 @@ defmodule URI do
       %{"percent" => "oh yes!", "starting" => "map"}
 
   """
-  @spec decode_query(String.t(), %{optional(String.t()) => String.t()}) :: %{
-          optional(String.t()) => String.t()
-        }
+  @spec decode_query(binary, %{optional(binary) => binary}) :: %{optional(binary) => binary}
   def decode_query(query, map \\ %{})
 
   def decode_query(query, %_{} = dict) when is_binary(query) do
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/uri.ex
```diff
@@ -92,7 +92,7 @@ defmodule URI do
       ** (ArgumentError) encode_query/1 values cannot be lists, got: [:a, :list]
 
   """
-  @spec encode_query(term) :: binary
+  @spec encode_query(term) :: String.t()
```
fertapric:
```suggestion
  @spec encode_query(Enumerable.t()) :: String.t()
```

lib/elixir/lib/uri.ex
```diff
@@ -76,7 +76,7 @@ defmodule URI do
   values are URL encoded as per `encode_www_form/1`.
 
   Keys and values can be any term that implements the `String.Chars`
-  protocol, except lists which are explicitly forbidden.
+  protocol, except lists – which are explicitly forbidden.
```
whatyouhide: Do you have resources on why this is a good change? Not a native speaker but it feels like this was fine before?

lib/elixir/lib/uri.ex
```diff
@@ -265,10 +265,11 @@ defmodule URI do
   See `encode_www_form` if you are interested in escaping reserved
   characters too.
 
-  This function also accepts a `predicate` function as an optional
-  argument. If passed, this function will be called with each byte
-  in `string` as its argument and should return `true` if the given
-  byte should be left as is.
+  It also accepts a function as an optional argument. If passed, `fun` will be
```
whatyouhide:
```suggestion
  This function also accepts a predicate function as an optional argument. If passed, `fun` will be
```

lib/elixir/lib/uri.ex
```diff
@@ -279,10 +280,10 @@ defmodule URI do
       "a str%69ng"
 
   """
-  @spec encode(binary, (byte -> boolean)) :: binary
-  def encode(string, predicate \\ &char_unescaped?/1)
-      when is_binary(string) and is_function(predicate, 1) do
-    for <<char <- string>>, into: "", do: percent(char, predicate)
+  @spec encode(binary, as_boolean(char)) :: binary
```
fertapric:
```suggestion
  @spec encode(String.t(), (byte -> as_boolean(term)) :: String.t()
```

lib/elixir/lib/uri.ex
```diff
@@ -279,10 +280,10 @@ defmodule URI do
...
+  def encode(string, fun \\ &char_unescaped?/1)
```
whatyouhide: I really liked the name "predicate". If you absolutely wanna stick "fun" in there, maybe `predicate_fun`?

lib/elixir/lib/uri.ex
whatyouhide: Wait, doesn't this take a fun? 🤔

lib/elixir/lib/uri.ex
josevalim: Note that we generally avoid such constructs, exactly because this increases the barrier for writing the docs. Should we use `-` or should we not? Same for `;` vs `.`/`,`.

lib/elixir/lib/uri.ex
```diff
@@ -324,7 +325,7 @@ defmodule URI do
       "https://elixir-lang.org"
 
   """
-  @spec decode(binary) :: binary
+  @spec decode(binary) :: String.t()
```
fertapric: I think the argument could also be `String.t()` (same for other functions in this module)

lib/elixir/lib/uri.ex
eksperimental: It should be actually `(char -> as_boolean(term))`

lib/elixir/lib/uri.ex
eksperimental: > Keys and values can be any term that implements the `String.Chars` protocol, except lists which are explicitly forbidden.

The original sentence may be understood as 
> Keys and values can be any term that implements the `String.Chars` protocol, except lists **that** are explicitly forbidden.

Which reads as some lists in particular are forbidden

I think we need a pause, to strees that any list is forbidden
My take is

> Keys and values can be any term that implements the `String.Chars` protocol with the exception of lists, which are explicitly forbidden.

or

> Keys and values can be any term that implements the `String.Chars` protocol with the exception of lists (which are explicitly forbidden).

lib/elixir/lib/uri.ex
eksperimental: Well, I also like that, but I had to change it, because it doesn't accept a predicate function, but any function. That's what this PR fixes. If we want to stick to the predicate function we need to error when no boolean is returned then.

lib/elixir/lib/uri.ex
eksperimental: Well, I also like that, but I had to change it, because it doesn't accept a predicate function, but any function. That's what this PR fixes. If we want to stick to the predicate function we need to error when no boolean is returned then.

lib/elixir/lib/uri.ex
eksperimental: See my comment bellow for my answer
https://github.com/elixir-lang/elixir/pull/8779#discussion_r254778179

lib/elixir/lib/uri.ex
eksperimental: I am submitting a commit that replaces all `binary` specs with `String.t()`
thank you

lib/elixir/lib/uri.ex
eksperimental: yes. I made the same mistake before. thank you for noting it.

eksperimental:
Some suggestions applied, other comments answered

lib/elixir/lib/uri.ex
whatyouhide: I don't think "predicate" is necessarily strictly boolean. A predicate is a function that is used as a predicate, that is, to tell if something is true or false, which in Elixir is pretty much truthy vs falsey. You could call the condition of an "if" a predicate. Let's call it `predicate` or `predicate_fun` and keep the rest of the code like it is, without checking for strictly boolean.

lib/elixir/lib/uri.ex
whatyouhide: Let's go with

> Keys and values can be any term that implements the `String.Chars` protocol with the exception of lists, which are explicitly forbidden.

whatyouhide:
I am not convinced about the change from `binary/0` to `String.t/0`. It's a lot of churn, but what do we gain? I'm afraid we will tend to do this change everywhere then.

eksperimental:
> I am not convinced about the change from `binary/0` to `String.t/0`. It's a lot of churn, but what do we gain? I'm afraid we will tend to do this change everywhere then.

We gain clarity. If the strings are to be meants UTF-8, then we should use `String.t()`.
I 'm for replacing it everywhere in core where this concept applies.
Also, I wouldn't want to promote to ways of doing the same thing, so to be honest, I would replace `binary` with `String.t()` wherever suitable.

lib/elixir/lib/uri.ex
fertapric: I think it's `byte` 🙂 

```elixir
iex(1)> URI.encode("🙂", &IO.inspect/1)
240
159
153
130
"🙂"
```

josevalim:
URIs do not have a predefined encoding though and many operations here work
rather at the byte level. I believe binary() is the best choice in most
cases here.

lib/elixir/lib/uri.ex
eksperimental: reverted

lib/elixir/lib/uri.ex
eksperimental: done

eksperimental:
please see latest commits

lib/elixir/lib/uri.ex
```diff
@@ -279,14 +280,14 @@ defmodule URI do
       "a str%69ng"
 
   """
-  @spec encode(binary, (byte -> boolean)) :: binary
+  @spec encode(binary, (char -> as_boolean(term))) :: binary
```
josevalim:
```suggestion
  @spec encode(binary, (byte -> as_boolean(term))) :: binary
```

Below we are taking bytes out of the binary (although we incorrectly named the var `char`).

lib/elixir/lib/uri.ex
```diff
@@ -617,6 +616,7 @@ defmodule URI do
 end
 
 defimpl String.Chars, for: URI do
+  @spec to_string(URI.t()) :: binary
```
josevalim: No need to define specs for protocol implementations.

eksperimental: oh yeah, that's true

eksperimental: fixed

lib/elixir/lib/uri.ex
eksperimental: I am renaming everything `byte` then.

josevalim:

eksperimental:
Thank you guys for your feedback

PR #8782, opened 2019-02-08
Fix module inspect on case insensitive filesystems

### LINKED ISSUE #8780: Running `i :Kernel` in IEx seems to crash
### Environment

```
Erlang/OTP 21 [erts-10.2.3] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1] [hipe]

```

### Current behavior

```
ex(15)> i :Kernel
Term

10:46:53.462 [error] beam/beam_load.c(1428): Error loading module 'Kernel':
  module name in object code is kernel


  :Kernel

10:46:53.462 [error] Loading of /usr/local/Cellar/erlang/21.2/lib/erlang/lib/kernel-6.2/ebin/Kernel.beam failed: :badfile

Data type
  Atom
Reference modules
  Atom
Implemented protocols
  IEx.Info, Inspect, String.Chars, List.Chars
```

### Expected behavior

The current behavior lets the user think a crash happened somewhere.

I suppose this is not desired, and the normal behavior is expected as for other non-special modules.

Perhaps it is needed to treat some modules as special for the `i` command to behave normally.

hauleth:
Earlier this outputted error messages when loaded module that existed
with different casing, ex.:

    iex> i Datetime

Caused errors to be shown. Now this checks if there is module file
defines module with the same name as the provided atom to prevent such
cases.

Close #8780

### DIFF
--- a/lib/iex/lib/iex/info.ex
+++ b/lib/iex/lib/iex/info.ex
@@ -51,7 +51,7 @@ defimpl IEx.Info, for: Atom do
       {^atom, beam, _path} ->
         info = :beam_lib.info(beam)
 
-        {:ok, atom} == Keyword.fetch(info, :module)
+        Keyword.fetch(info, :module) == {:ok, atom}
     end
   end
 
--- a/lib/iex/lib/iex/info.ex
+++ b/lib/iex/lib/iex/info.ex
@@ -49,13 +49,9 @@ defimpl IEx.Info, for: Atom do
         Code.ensure_loaded?(atom)
 
       {^atom, beam, _path} ->
-        module_name =
-          beam
-          |> :beam_lib.info()
-          # Prevent match on `nil`
-          |> Keyword.get(:module, 0)
+        info = :beam_lib.info(beam)
 
-        module_name == atom
+        match?({:ok, ^atom}, Keyword.fetch(info, :module))
     end
   end
 
--- a/lib/iex/lib/iex/info.ex
+++ b/lib/iex/lib/iex/info.ex
@@ -18,7 +18,7 @@ defimpl IEx.Info, for: Atom do
   def info(atom) do
     specific_info =
       cond do
-        Code.ensure_loaded?(atom) ->
+        module?(atom) ->
           info_module(atom)
 
         match?("Elixir." <> _, Atom.to_string(atom)) ->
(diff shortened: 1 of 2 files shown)


hauleth:
I am not sure how we could test that on case-sensitive systems so there are no tests. If someone have any idea then I am open to it.

whatyouhide:
@hauleth could we write a test for this that would just do nothing on case-sensitive systems so that at least when the test is running on case-insensitive systems we are testing the correct behaviour?

hauleth:
@whatyouhide I can write one like that.

josevalim:
Looks great, just a test and it should be good to go. I don't think the test needs to check for the filesystem because for case sensitive ones we will never show the warning either.

hauleth:
@josevalim done

lib/iex/lib/iex/info.ex
```diff
@@ -43,6 +43,22 @@ defimpl IEx.Info, for: Atom do
     [{"Data type", "Atom"}] ++ description ++ specific_info
   end
 
+  defp module?(atom) do
+    case :code.get_object_code(atom) do
+      :error ->
+        Code.ensure_loaded?(atom)
+
+      {^atom, beam, _path} ->
+        module_name =
```
fertapric: What do you think about using `Keyword.fetch` to avoid the comment and the default `0` value?

```elixir
beam_info = :beam_lib.info(beam)

case Keyword.fetch(beam_info, :module) do
  {:ok, ^atom} -> true
  _ -> false
end
```

hauleth: Changed to `match?/2`.

josevalim:
This match is equivalent to == {:ok, atom}.
-- 


hauleth:
@josevalim right, I have missed that.

lib/iex/lib/iex/info.ex
```diff
@@ -43,6 +43,18 @@ defimpl IEx.Info, for: Atom do
...
 
+  defp module?(atom) do
+    case :code.get_object_code(atom) do
+      :error ->
+        Code.ensure_loaded?(atom)
+
+      {^atom, beam, _path} ->
+        info = :beam_lib.info(beam)
+
+        {:ok, atom} == Keyword.fetch(info, :module)
```
whatyouhide: Personally, I feel more comfortable with this check being the other way around for some reason. If you also think it's a good change, here's a nice little change that you can just commit otherwise no problem! 😄 

```suggestion
        Keyword.fetch(info, :module) == {:ok, atom}
```

lib/iex/lib/iex/info.ex
```diff
@@ -43,6 +43,18 @@ defimpl IEx.Info, for: Atom do
...
   end
 
+  defp module?(atom) do
+    case :code.get_object_code(atom) do
+      :error ->
+        Code.ensure_loaded?(atom)
+
+      {^atom, beam, _path} ->
+        info = :beam_lib.info(beam)
+
```
whatyouhide:
```suggestion
```

hauleth: I personally always separate variables and "main body" of the function, but this is up to you.

josevalim:

PR #8786, opened 2019-02-10
(WIP) Do not allow to override macros with functions and vice-versa

fertapric:
Overriding a macro with a function (and vice-versa) might produce undesired effects.

For example, it's not clear if the following code should or should not raise:

```elixir
defmodule Foo do
  def foo, do: bar()
  defmacro bar, do: :ok
  defoverridable bar: 0
  def bar, do: :ok
end
```

On top of that, the current implementation of `super` does not work for such cases.


fertapric:
I'll open a new PR with a new approach.

PR #8797, opened 2019-02-15
String.Casing.Guards.{is_lowercase/1, is_uppercase/1}

am-kantox:
To parse strings with full Unicode support it’s useful to have guards `is_lowercase/1` and `is_uppercase/1` for codepoints.

I could produce my own library that would parse `"SpecialCasing.txt"` on its own but this would be a code duplication. Also, `String.Casing` module does not export anything to make it possible to reuse. I am not sure if the existence of these guards makes sense in the Elixir core, but it sounds like where it actually belongs.

It would be easier to explain the reasons with the code, hence this PR. If it’ll be found of any value, I could provide docs etc.

### DIFF
--- a/lib/elixir/test/elixir/string_test.exs
+++ b/lib/elixir/test/elixir/string_test.exs
@@ -185,6 +185,36 @@ defmodule StringTest do
     assert String.downcase("OLÁ", :ascii) == "olÁ"
   end
 
+  require String.Casing.Guards
+
+  test "assert is_lowercase/1" do
+    "olá"
+    |> String.normalize(:nfc)
+    |> String.codepoints()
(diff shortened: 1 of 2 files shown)


josevalim:
The proposed implementation has a very big issue in that every time those guards are used, it will inject many checks, considerably increasing the size of the bytecode. I really advise to not implement those as guards/macros but a regular function that returns the casing if appropriate:

    casing("é") -> :lower
    casing("E") -> :upper
    ...

That's also why we don't include this functionality in Elixir, it will increase the language size when not everyone needs it. It is better as a library that exports the casing information.

eksperimental:
I was 2 days ago working on a function to optimize guards and regular functions, and this looks like a really good case for it. It can get back to you if you are interested wheneer I have some functioning code

PR #8799, opened 2019-02-15
Store overridable information more efficiently

fertapric:
This is part of an ongoing effort to forbid overriding macros with functions and vice-versa. New checks will require the overridable information to be stored more efficiently.

### DIFF
--- a/lib/elixir/src/elixir_overridable.erl
+++ b/lib/elixir/src/elixir_overridable.erl
@@ -19,12 +19,12 @@ overridable_for(Module, Tuple) ->
 record_overridable(Module, Tuple, Def, Neighbours) ->
   {Set, Bag} = elixir_module:data_tables(Module),
 
-  case ets:lookup(Set, {overridable, Tuple}) of
-    [{_, Count, _, _, _}] ->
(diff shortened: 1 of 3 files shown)


lib/elixir/src/elixir_overridable.erl
```diff
@@ -1,80 +1,89 @@
...
 
-store_pending(Module) ->
+record_overridable(Module, Tuple, Def, Neighbours) ->
+  {Set, Bag} = elixir_module:data_tables(Module),
+
+  case ets:insert_new(Set, {{overridable, Tuple}, 1, Def, Neighbours, false}) of
+    true ->
+      ets:insert(Bag, {overridable, Tuple});
+    false ->
+      [{_, Count, _, _, _}] = ets:lookup(Set, {overridable, Tuple}),
```
josevalim:
```suggestion
      Count = ets:lookup_element(Set, {overridable, Tuple}, 2),
```

I did not test it though!

josevalim:
I dropped one last comment. Then we just need to update this doc and we should be good to go: https://github.com/elixir-lang/elixir/blob/master/lib/elixir/src/elixir_module.erl#L222

josevalim:

PR #8804, opened 2019-02-17
Adding missing List functions tests

rauann:
I noticed that there are missing tests for existing functions in the List module.

### DIFF
--- a/lib/elixir/test/elixir/list_test.exs
+++ b/lib/elixir/test/elixir/list_test.exs
@@ -22,6 +22,11 @@ defmodule ListTest do
     assert (&[&1, &2 | &3]).(1, 2, 3) == [1, 2 | 3]
   end
 
+  test "delete/2" do
+    assert List.delete([:a, :b, :c], :a) == [:b, :c]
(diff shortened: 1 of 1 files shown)


lib/elixir/test/elixir/list_test.exs
```diff
@@ -22,6 +22,11 @@ defmodule ListTest do
     assert (&[&1, &2 | &3]).(1, 2, 3) == [1, 2 | 3]
   end
 
+  test "delete/2" do
+    assert List.delete([:a, :b, :c], :a) == [:b, :c]
```
amatalai: These tests are already generated from documentation examples as doctests (more detailed information on how doctests work can be found here: https://hexdocs.pm/ex_unit/ExUnit.DocTest.html)

josevalim: Although originally we used the doctests for coverage, we tend to not rely on doctests for tests, since they are about documentation first and foremost. :)

josevalim:

lib/elixir/test/elixir/list_test.exs
rauann: :+1:

PR #8808, opened 2019-02-18
Update docs to consistently use "keyword list"

trptcolin:
As I understand it, "keyword" only refers to the [atom] keys of this data structure, and a "keyword list" refers to the data structure itself. This change makes the documentation consistent.

Arguably changing the module name to `KeywordList` would make sense too, since `Keyword.new` [perhaps unintuitively] creates a keyword list. But I definitely wouldn't argue for a breaking change! 😄 

Feel free to close if my understanding is wrong, and we should be using "keyword" and "keyword list" interchangeably!


josevalim:

PR #8814, opened 2019-02-21
Raise error when `@doc true`

### LINKED ISSUE #8806: Help for function crashes when `@doc true` for the function
### Environment

```
$ iex -S mix
Erlang/OTP 21 [erts-10.2.3] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1] [hipe]

```

### Current behavior

Given the file `p.ex` with the content

```
defmodule P do
  @doc true
  def foo, do: :foo
end
```

the following happens in IEx:

```
$ iex -S mix
Erlang/OTP 21 [erts-10.2.3] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1] [hipe]

iex(1)> h P.foo

                                   def foo()

** (FunctionClauseError) no function clause matching in String.split/3

    The following arguments were given to String.split/3:

        # 1
        true

        # 2
        ["\r\n", "\n"]

        # 3
        [trim: false]

    Attempted function clauses (showing 4 out of 4):

        def split(string, %Regex{} = pattern, options) when is_binary(string)
        def split(string, "", options) when is_binary(string)
        def split(string, pattern, []) when is_tuple(pattern) or is_binary(string)
        def split(string, pattern, options) when is_binary(string)

    (elixir) lib/string.ex:407: String.split/3
    (elixir) lib/io/ansi/docs.ex:109: IO.ANSI.Docs.print/2
    (iex) lib/iex/introspection.ex:369: IEx.Introspection.h_mod_fun_arity/3
    (iex) lib/iex/introspection.ex:295: anonymous fn/5 in IEx.Introspection.h/1
    (elixir) lib/enum.ex:1940: Enum."-reduce/3-lists^foldl/2-0-"/3
    (iex) lib/iex/introspection.ex:289: IEx.Introspection.h/1
iex(1)>
```

It seems that the value of `@doc` is not being checked for a string before `String.split/3` is called.

Note that this issue does not happen when the module is defined directly inside IEx, as in

```
iex(1)> defmodule P do
...(1)>   @doc true
...(1)>   def foo, do: :foo
...(1)> end
{:module, P,
 <<70, 79, 82, 49, 0, 0, 4, 16, 66, 69, 65, 77, 65, 116, 85, 56, 0, 0, 0, 119,
   0, 0, 0, 13, 8, 69, 108, 105, 120, 105, 114, 46, 80, 8, 95, 95, 105, 110,
   102, 111, 95, 95, 7, 99, 111, 109, 112, ...>>, {:foo, 0}}
iex(2)> h P.foo
P was not compiled with docs
iex(3)>
```

### Expected behavior

The expected behavior should be

```
iex(1)> h P.foo

                                   def foo()
```

DiodonHystrix:
fixes #8806

### DIFF
--- a/lib/elixir/lib/module.ex
+++ b/lib/elixir/lib/module.ex
@@ -1851,7 +1851,7 @@ defmodule Module do
 
   defp preprocess_attribute(key, value) when key in [:moduledoc, :typedoc, :doc] do
     case value do
-      {line, doc} when is_integer(line) and (is_binary(doc) or is_boolean(doc) or is_nil(doc)) ->
+      {line, doc} when is_integer(line) and (is_binary(doc) or doc == false or is_nil(doc)) ->
         value
 
       {line, [{key, _} | _]} when is_integer(line) and is_atom(key) ->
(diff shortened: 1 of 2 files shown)


josevalim:

PR #8816, opened 2019-02-21
Do not allow to override macros as functions (and vice-versa)

fertapric:
Overriding a macro as a function (and vice-versa) might produce undesired effects.

For example, it's not clear if the following code should or should not raise:

```elixir
defmodule Foo do
  def foo, do: bar()
  defmacro bar, do: :ok
  defoverridable bar: 0
  def bar, do: :ok
end
```

On top of that, `super` does not currently work for such cases.


josevalim:

PR #8817, opened 2019-02-22
Comparable protocol

tim2CF:
# Comparable

Elixir protocol which describes how two Elixir terms can be compared. There are cases when we want to compare two terms of some data type not just by term value according standard Erlang/Elixir [ordering rules](https://hexdocs.pm/elixir/operators.html#term-ordering) but to use some meaningful business logic to do comparison. Main purpose of this pull request is to provide extended versions of standard kernel functions like `==/2`, `!=/2`, `>/2`, `</2`, `>=/2`, `<=/2` which will rely on Comparable protocol implementation for given type. Protocol itself is very similar to [Ord](http://hackage.haskell.org/package/base-4.12.0.0/docs/Data-Ord.html) type class in Haskell.

## Real example

Kernel Elixir comparison functions work pretty fine with standard numeric types like `integer` or `float` (and it works even in nested terms like `map`):

```elixir
iex> %{a: 1} == %{a: 1.0}
true
```

But if we try to apply Kernel equality function to terms containing custom [Decimal](https://hex.pm/packages/decimal) numbers it will not work so good:

```elixir
iex(1)> %{a: Decimal.new("1")} == %{a: Decimal.new("1.0")}
false
```

This is because **the same** decimal number can be presented **as different Elixir term**:

```elixir
iex> Decimal.new("1") |> Map.from_struct
%{coef: 1, exp: 0, sign: 1}
iex> Decimal.new("1.0") |> Map.from_struct
%{coef: 10, exp: -1, sign: 1}
```

And here `Comparable` protocol can help us, let's implement it for `Decimal` type using existing `Decimal.compare/2` helper:

```elixir
defimpl Comparable, for: Decimal do
  require Comp

  def compare(left, %Decimal{} = right) do
    left
    |> Decimal.compare(right)
    |> case do
      %Decimal{coef: 1, sign: 1} ->
        Comp.gt()

      %Decimal{coef: 1, sign: -1} ->
        Comp.lt()

      %Decimal{coef: 0} ->
        Comp.eq()

      %Decimal{coef: :qNaN} ->
        raise(
          "can't apply Comparable protocol to left = #{inspect(left)} and right = #{
            inspect(right)
          }"
        )
    end
  end

  Comp.deffallback()
end
```

where `Comp.deffallback/0` is simple helper to auto-generate clauses where **right** term is not a `Decimal` (it's just fallback to standard Kernel implementation).

And when protocol for `Decimal` type is implemented, we can use `Comp.equal?/2` utility function instead of Kernel `==/2`:

```elixir
iex> Comp.equal?(%{a: Decimal.new("1")}, %{a: Decimal.new("1.0")})
true
```

which works as expected according **meaning** of `Decimal` numbers instead of just term **values**. Comparison based on  `Comparable` protocol is very useful when for example we compare big nested structures which contain `Decimals` or other custom types (like `Date`, `Time`, `NaiveDateTime`, `URI` etc) in nested collections like lists, maps, tuples or other data types:

```elixir
iex> x0 = %{a: [%{b: Decimal.new("1")}]}
%{a: [%{b: #Decimal<1>}]}
iex> x1 = %{a: [%{b: Decimal.new("1.0")}]}
%{a: [%{b: #Decimal<1.0>}]}
iex> x0 == x1
false
iex> Comp.equal?(x0, x1)
true
```

## Utilities

`Comp` module provides utilities not only for equality, but for other comparison operations as well. Also it provides infix shortcuts for these utilities:

| Kernel.fn/2 | Comp.fn/2 | Comp infix shortcut |
|-------------|-----------|---------------------|
| x == y | Comp.equal?(x, y) | x <~> y |
| x != y | Comp.not_equal?(x, y) | x <&#124;> y |
| x > y | Comp.greater_than?(x, y) | x >>> y |
| x < y | Comp.less_than?(x, y) | x <<< y |
| x >= y | Comp.greater_or_equal?(x, y) | x ~>> y |
| x <= y | Comp.less_or_equal?(x, y) | x <<~ y |
| max(x, y) | Comp.max(x, y) |
| min(x, y) | Comp.min(x, y) |

Example of infix shortcuts usage:

```elixir
iex> use Comp
Comp
iex> Decimal.new("1") <~> Decimal.new("1.0")
true
iex> Decimal.new("1") <|> Decimal.new("2")
true
iex> Decimal.new("2") >>> Decimal.new("1")
true
iex> Decimal.new("1") <<< Decimal.new("2")
true
iex> Decimal.new("1") ~>> Decimal.new("1.0")
true
iex> Decimal.new("1") <<~ Decimal.new("1.0")
true
```

Also there is simple `Comp.compare/2` function if you want to work directly with `Ord` enum values:

```elixir
iex> Comp.compare(1, 1)
:eq
iex> Comp.compare(1, 2)
:lt
iex> Comp.compare(2, 1)
:gt
```

## DIscussion 

`Comparable` protocol features discussion in Elixir mailing list
https://groups.google.com/forum/#!topic/elixir-lang-core/eE_mMWKdVYY

### DIFF
--- a/lib/elixir/lib/comp.ex
+++ b/lib/elixir/lib/comp.ex
@@ -1,0 +1,599 @@
+defprotocol Comparable do
+  @moduledoc """
+  Protocol that describes comparison of two terms
+  """
+
+  @fallback_to_any true
+  @type t :: Comparable.t()
+  @type ord :: :gt | :lt | :eq
+
+  @doc """
+  Function to compare term or given type with any other term
+
+  ## Examples
+
+  ```
+  iex> 2 > 1
+  true
+  iex> Comparable.compare(2, 1)
+  :gt
+  iex> 1 < 2
+  true
+  iex> Comparable.compare(1, 2)
+  :lt
+  iex> 1 == 1
+  true
+  iex> Comparable.compare(1, 1)
+  :eq
+
+  iex> {1, 2, 3} > {1, 2}
+  true
+  iex> Comparable.compare({1, 2, 3}, {1, 2})
+  :gt
+  iex> {1, 2} < {1, 2, 3}
+  true
+  iex> Comparable.compare({1, 2}, {1, 2, 3})
+  :lt
+  iex> {1, 2} == {1, 2}
+  true
+  iex> Comparable.compare({1, 2}, {1, 2})
+  :eq
+
+  iex> {1, 2, 3} > {1, 1000}
+  true
+  iex> Comparable.compare({1, 2, 3}, {1, 1000})
+  :gt
+  iex> {1, 1000} < {1, 2, 3}
+  true
+  iex> Comparable.compare({1, 1000}, {1, 2, 3})
+  :lt
+  iex> {1, 2} < {1, 1000}
+  true
+  iex> Comparable.compare({1, 2}, {1, 1000})
+  :lt
+
+  iex> [1, 2, 3] > [1, 2]
+  true
+  iex> Comparable.compare([1, 2, 3], [1, 2])
+  :gt
+  iex> [1, 2] < [1, 2, 3]
+  true
+  iex> Comparable.compare([1, 2], [1, 2, 3])
+  :lt
+  iex> [1, 2] == [1, 2]
+  true
+  iex> Comparable.compare([1, 2], [1, 2])
+  :eq
+
+  iex> [1, 2, 3] < [1, 1000]
+  true
+  iex> Comparable.compare([1, 2, 3], [1, 1000])
+  :lt
+  iex> [1, 1000] > [1, 2, 3]
+  true
+  iex> Comparable.compare([1, 1000], [1, 2, 3])
+  :gt
+  iex> [1, 2] < [1, 1000]
+  true
+  iex> Comparable.compare([1, 2], [1, 1000])
+  :lt
+
+  iex> %{z: 1000} < %{a: 1, b: 1}
+  true
+  iex> Comparable.compare(%{z: 1000}, %{a: 1, b: 1})
+  :lt
+  iex> %{a: 1, z: 1} > %{a: 1000, b: 1000}
+  true
+  iex> Comparable.compare(%{a: 1, z: 1}, %{a: 1000, b: 1000})
+  :gt
+  iex> %{a: 1, b: 1} == %{a: 1, b: 1}
+  true
+  iex> Comparable.compare(%{a: 1, b: 1}, %{a: 1, b: 1})
+  :eq
+
+  iex> %{a: 1, b: 1, c: 1000} < %{a: 1, b: 1000, c: 1}
+  true
+  iex> Comparable.compare(%{a: 1, b: 1, c: 1000}, %{a: 1, b: 1000, c: 1})
+  :lt
+
+  iex> {:ok, dt} = NaiveDateTime.new(2000, 1, 1, 0, 0, 0)
+  {:ok, ~N[2000-01-01 00:00:00]}
+  iex> NaiveDateTime.compare(NaiveDateTime.add(dt, 1, :microsecond), NaiveDateTime.add(dt, 1, :second))
+  :lt
+  iex> Comparable.compare(NaiveDateTime.add(dt, 1, :microsecond), NaiveDateTime.add(dt, 1, :second))
+  :lt
+
+  iex> Comparable.compare(%URI{host: "1"}, %URI{host: "2"})
+  :lt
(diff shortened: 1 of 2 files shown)


josevalim:
Hi @tim2CF! Thank you for this pull request. :heart:

This particular feature has been proposed a couple times on the Elixir mailing list, so I would recommend reading those discussions if you haven't yet. If after reading those discussions you believe that your implementation addresses the concerns raised, then you can either "resurrect" the thread with further information or send a new proposal linking to your PR as a reference implementation.

From the top of my head, the biggest concern was how to handle associativity, so `1 < Decimal.new("2")` and `Decimal.new("2") >= 1` emits the same result. Insight on how this would be handled with and the performance consequences would be welcome.

Therefore, please bring this up for [discussion in the mailing list](https://groups.google.com/forum/#!forum/elixir-lang-core) so we can get feedback from the community.

Also, if you would be so kind, please edit the PR description with a link to the mailing list thread, so we center the discussion there. Thank you!

tim2CF:
Hi @josevalim ! 
Thanks for feedback, I edited PR description (added link to mailing list). Looks like problem of associativity is real, but I have a plan how I can solve it. Will update PR ASAP. About usage of this functionality as library - yes, it is possible, but Elixir structs ideologically is only way to create new types, extend standard type system, and `Ord` (or maybe just more simple `Eq`) type class is very basic abstractions which allow to work comfortable with these new non-erlang types, so I (and I hope other people as well) would like to see such basic functionality in language itself

OvermindDL1:
If this replaced all the basic operators than wouldn't it slow all the basic operators, in addition prevent them from being used in guards, both of which seem to be pretty major issues (both of which is not an issue with a statically typed system but Elixir is not able to do that yet)?  Any such protocol should instead be called via function names on the protocol itself only where wanted, not operators that are easy to accidentally misuse; that would at least fix the guard and visibility issues though it still wouldn't be terribly efficient as it would decompose to, at best, a set of linearly tested matchers.

josevalim:
Hi @tim2CF, any news in this front?

As I previously mentioned, this issue is a bit more complex than it looks at the surface, so I propose for us to close this PR until we have thoroughly discussed how we want to address its implementation concerns.

Thank you!

tim2CF:
@josevalim yes, it is more complex. I actually solved issue of symmetry - now protocol implementation now is made not for type, but for pair of types. Solution is here, it's done for more simple `Equalable` protocol which describes not ordering, but just equivalence. Take a look, there are examples in readme file:

https://github.com/timCF/equalable

I'm pretty sure that `Comparable` / `Orderable` protocol can be done using the same strategy. I also tried to generalize this pattern and write extended versions of `defprotocol/2` and `defimpl/2` where protocol and implementation are not unary but binary (defined for pair of types) - but finally I did manual implementation for particular `Equalable` protocol because looks like generalization here not decreasing amount of code.

josevalim:
Thanks @tim2CF!

Equalable looks really nice and it does solve some of the problems mentioned but at the cost of combinatory implementations. If I use Decimal and someone uses SuperDecimal in the same project, the comparison will fail unless I explicitly implement Decimal<->SuperDecimal and SuperDecimal<->Decimal.

Unfortunately, the only way I know to solve this issue is to restrict comparisons to only members of the same type. It does solve the immediate issue of comparing decimals with decimals or dates with dates,  but without any of the transitivity. It would solve the most obvious concerns but at this point, we need to choose between a limited solutions or a solution that is too complex, and we are not sure any of them is a good fit. Which is similar to previous conclusions we have reached in the past on the topic.

tim2CF:
Well, symmetry property is here. With Equalable only one implementation required, let's say `Decimal<->SuperDecimal`. Other one, `SuperDecimal<->Decimal` is inferred here automatically (because `left` type is not equal to `right` type)

https://github.com/timCF/equalable/blob/1010df0745f04856b13fd62028661e19be9554c1/lib/equalable.ex#L264-L278

But no transitivity, that is true :( 
That is definitely no good, but looks like transitivity is mostly theoretical issue - I don't see common practical code patterns in Elixir which rely on transitivity.

josevalim:
Oh, nice, I misspoke about symmetry.

I think the overall conclusion still applies though: it is bringing a lot of complexity (for example, a whole new way of declaring protocols) while there is a simpler and more limited solution available (which may not be good enough). If we are going by the more complex route, then we should at least make non-unary protocols part of the language and that already is a large effort (especially if we want to take protocol consolidation into account).

tim2CF:
Generic protocol abstraction with unary, binary and N-ary protocol support will be fatastic, but I don't have enough time to develop it right now. Custom `Equalable` and `Comparable`/`Orderable` binary protocols are enough to solve my problems with `Decimal` type right now, I close PR and will use them as libraries.

PR #8820, opened 2019-02-23
Fix uncommented `...` in periodic task GenServer documentation

notactuallypagemcconnell:
Howdy!

I mentor a lot of folks in Elixir day to day and find myself using this doc with a lot of introductions to GenServers. I recently noticed it stopped working when I would paste directly into the REPL. It looks like these were added in commit b230541fae but before then the only note was commented. In the interest of copy-pasteability for the documentation (especially in IEx) I figured this would be an improvement.


josevalim:

PR #8823, opened 2019-02-24
Set DateTime.from_unix precision to microseconds for units not divisible by 10

wojtekmach:
Before:

    DateTime.from_unix!(143_256_036_886_856, 1024).microsecond
    {320312, 3}

After:

    DateTime.from_unix!(143_256_036_886_856, 1024).microsecond
    {320312, 6}

---

Alternatively, should we truncate it to `{320, 3}`?


josevalim:

PR #8826, opened 2019-02-24
Ignore protocols when computing coverage data

### LINKED ISSUE #8825: mix test --cover: total is less than the mathematical average
### Environment

* Elixir & Erlang/OTP versions (elixir --version): Erlang/OTP 21 [erts-10.2.1]
* Operating system: macOS 10.14.3

### Current behavior

run mix test --cover

Percentage | Module
   100.00% | Cover.BitString
   100.00% | CoverFixture
    33.33% | Total

The total is less than expected. I think the problem is with the protocol. The file "Cover" itself is not listet.

https://github.com/Helpys/coverfixture

### Expected behavior

Percentage | Module
   100.00% | Cover.BitString
   100.00% | CoverFixture
    100.00% | Total

PragTob:
Protocols have no "code" to run, hence shouldn't be considered
for coverage data. We could also just mark them as covered, but
I feel like discarding them is the better way as they really
don't have relevant code.

fixes #8825

As usual, thanks for elixir and all your hard work! :green_heart: :green_heart: :green_heart:

### DIFF
--- a/lib/mix/lib/mix/tasks/test.ex
+++ b/lib/mix/lib/mix/tasks/test.ex
@@ -40,6 +40,8 @@ defmodule Mix.Tasks.Test do
       # Each line is only considered once.
       #
       # We use ETS for performance, to avoid working with nested maps.
+      results = discard_protocol_definitions(results)
+
       table = :ets.new(__MODULE__, [:set, :private])
 
       try do
@@ -61,6 +63,14 @@ defmodule Mix.Tasks.Test do
       after
         :ets.delete(table)
       end
+    end
+
+    # protocol definitions have no testable code they shouldn't show up in coverage results
+    # see #8825
+    defp discard_protocol_definitions(coverage_results) do
+      Enum.reject(coverage_results, fn {{module, _}, _} ->
+        function_exported?(module, :__protocol__, 1)
+      end)
     end
 
     defp read_module_cover_results(table, module) do
--- a/lib/mix/test/fixtures/protocol_cover/lib/protocol_cover.ex
+++ b/lib/mix/test/fixtures/protocol_cover/lib/protocol_cover.ex
@@ -1,0 +1,7 @@
+defprotocol ProtocolCover do
+  def to_uppercase(string)
+end
+
+defimpl ProtocolCover, for: BitString do
+  def to_uppercase(string), do: String.upcase(string)
+end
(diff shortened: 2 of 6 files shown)


lib/mix/lib/mix/tasks/test.ex
```diff
@@ -40,6 +40,8 @@ defmodule Mix.Tasks.Test do
       # Each line is only considered once.
       #
       # We use ETS for performance, to avoid working with nested maps.
+      results = discard_protocol_definitions(results)
```
PragTob: This _could_ be piped into `Enum.each` at the very beginning of the `try` block - I felt like keeping it outside of the `try` and table interactions was best. We could also filter before passing the results in here, as `:cover.modules()` is passed directly I was hesitant  but we could also do it right there.

lib/mix/lib/mix/tasks/test.ex
```diff
@@ -63,6 +65,14 @@ defmodule Mix.Tasks.Test do
       end
     end
 
+    # protocol definitions have no testable code they shouldn't show up in coverage results
+    # see #8825
+    defp discard_protocol_definitions(coverage_results) do
+      Enum.reject(coverage_results, fn {{module, _}, _} ->
+        function_exported?(module, :__protocol__, 1)
```
PragTob: Not sure if this check suffices here of it should rather be like `Module.protocol?/1`:

```
  defp protocol?(module) do
    Code.ensure_loaded?(module) and function_exported?(module, :__protocol__, 1) and
      module.__protocol__(:module) == module
  end
```

Speaking of which, as other parts of the code base also check for `function_exported?(module, :__protocol__)` (info.ex and xref.ex) it might be worth considering (from my point of views) to make `Module.protocol?/1` public - would have also kind of hoped to find it in a public function somewhere :)

lib/mix/lib/mix/tasks/test.ex
josevalim: Thanks @PragTob! However, I am afraid ignoring protocols is going to hide the root cause. What if someone implements something like a protocol that does not have any code either?

josevalim:
Hi @PragTob! I decided to explore this a bit further and I ended-up working a different solution that is a bit more general here: #8828.

PragTob:
@josevalim I see, well that looks like a better solution. Thanks!

PR #8828, opened 2019-02-25
Do not consider modules that are no longer cover compiled

### LINKED ISSUE #8825: mix test --cover: total is less than the mathematical average

josevalim:
The main issue is that Cover no longer considers protocols to
be cover compiled at the end of suite, which was causing protocols
to be taken into account when considering Total but they were not
listed individually. 

By only considering cover compiled modules when computing the results,
it automatically removes protocols and it should also apply for future cases.
The umbrella --cover test was improved to be more assertive and also
list protocols.

Closes #8825

### DIFF
--- a/lib/mix/lib/mix/tasks/test.ex
+++ b/lib/mix/lib/mix/tasks/test.ex
@@ -33,6 +33,8 @@ defmodule Mix.Tasks.Test do
     end
 
     defp gather_coverage(results, keep) do
+      keep_set = MapSet.new(keep)
+
       # When gathering coverage results, we need to skip any
       # entry with line equal to 0 as those are generated code.
       #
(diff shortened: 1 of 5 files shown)


lib/mix/lib/mix/tasks/test.ex
```diff
@@ -43,11 +45,12 @@ defmodule Mix.Tasks.Test do
       table = :ets.new(__MODULE__, [:set, :private])
 
       try do
-        Enum.each(results, fn
-          {{module, 0}, _} -> :ets.insert(table, {{module, 0}, :dummy})
-          {{module, line}, {1, 0}} -> :ets.insert(table, {{module, line}, true})
-          {{module, line}, {0, 1}} -> :ets.insert_new(table, {{module, line}, false})
-        end)
+        for {{module, line}, cov} <- results, module in keep_set, line != 0 do
```
PragTob: :sparkles:

lib/mix/test/mix/tasks/test_test.exs
```diff
@@ -123,24 +123,32 @@ defmodule Mix.Tasks.TestTest do
...
+
+      # For foo, we do regular --cover and test it does not include bar
+      assert output =~ """
+             Generating cover results ...
+
+             Percentage | Module
+             -----------|--------------------------
+                100.00% | Foo
+             -----------|--------------------------
+                100.00% | Total
```
PragTob: I like this test a lot better than the original :tada:

josevalim:

PR #8833, opened 2019-02-26
Remove StreamData-related stuff from the formatter configuration

whatyouhide:
We're not including StreamData (https://github.com/whatyouhide/stream_data) in the standard library anymore. It doesn't make sense that we have StreamData functions in the default locals without parens of the formatter. Instead, we solve the problem on the StreamData side with  https://github.com/whatyouhide/stream_data/pull/124.

### DIFF
--- a/lib/elixir/test/elixir/code_formatter/calls_test.exs
+++ b/lib/elixir/test/elixir/code_formatter/calls_test.exs
@@ -415,33 +415,33 @@ defmodule Code.Formatter.CallsTest do
     end
 
     test "without parens on unique argument" do
-      assert_same "foo(all 1, 2, 3)"
-      assert_same "foo(bar, all(1, 2, 3))"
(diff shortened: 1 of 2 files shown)


lib/elixir/test/elixir/code_formatter/calls_test.exs
```diff
@@ -414,37 +414,6 @@ defmodule Code.Formatter.CallsTest do
       assert_same "foo(bar)", locals_without_parens: [foo: :*]
     end
 
-    test "without parens on unique argument" do
```
josevalim: We need to rewrite those assertions, they are not specific for check and all. Replace `check` by `assert` and `all` by `for` and I think we should be good.

whatyouhide: Ah, got it. Updated!

PR #8834, opened 2019-02-26
Mix version task

kenichi:
Apologies if this is horribly naive, but I searched and couldn't find any discussions on the topic. All the examples I've found generally awk the version string out of mix.exs, or do a similar [thing](https://github.com/aforward/version_tasks/blob/master/lib/mix/tasks/current.ex). I'm using this during a distillery release build to know what path the output tarball will be under. Super appreciative of any feedback!

### DIFF
--- a/lib/mix/lib/mix/tasks/version.ex
+++ b/lib/mix/lib/mix/tasks/version.ex
@@ -1,0 +1,9 @@
+defmodule Mix.Tasks.Version do
+  use Mix.Task
+
+  @impl Mix.Task
+  def run(_args) do
+    Mix.shell.info Mix.Project.config()[:version]
+  end
+
+end
--- a/lib/mix/lib/mix/tasks/version.ex
+++ b/lib/mix/lib/mix/tasks/version.ex
@@ -3,7 +3,6 @@ defmodule Mix.Tasks.Version do
 
   @impl Mix.Task
   def run(_args) do
-    Mix.shell.info Mix.Project.config()[:version]
+    Mix.shell().info(Mix.Project.config()[:version])
   end
-
 end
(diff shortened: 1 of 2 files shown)


josevalim:
Hi @kenichi, thanks for the PR but do note we do support `mix -v`, which is the same as `elixir -v` and `iex -v`.

kenichi:
Hi @josevalim, thanks for looking! Please note this is slightly different than `mix -v`, as this task outputs the current app's version:

```
$ grep version mix.exs
      version: "0.1.0",
$ mix -v
Erlang/OTP 21 [erts-10.2.3] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1] [hipe] [dtrace]

Mix 1.8.1 (compiled with Erlang/OTP 21)
$ mix version
0.1.0
```

josevalim:
Hi @kenichi, apologies for the confusion.

If you need to get the app version, you can do this: 

    $ mix run --no-deps-check --no-compile -e "IO.puts Mix.Project.config[:version]"

I believe this approach is better then adding a new task to Mix with very restricted use cases. :)

Thoughts?

eksperimental:
Maybe create a "project-config" task where you can pass the "--version" option.

kenichi:
@josevalim that line will work, thanks! I do understand about the limited utility of this PR, but found at least one precedent in `npm version`, not to say that is the best idea by any means.

I'm using this task in my deploy scripts to find the path to:

```
_build/prod/rel/[app]/releases/#{version}/[app].tar.gz
```

@eksperimental thanks, I thought of something similar, but thought there might be too much clash/resemblance to the main `--version` option. We've added this to our project's `lib/mix/tasks/` so it is currently working for our needs.

PR #8837, opened 2019-02-27
Raise ArgumentError on GenServer.call/3 invalid timeout

dnlserrano:
While using a GenServer I was debugging an issue for some hours together with a colleague (shout out @dcaixinha). Eventually, I got to understand the issue was due to passing `nil` as the timeout to `GenServer`, implicitly. `poolboy` was being used, and we were configuring the timeout via an environment variable (the value of which was being fetched with a typo from `Application.get_env(:my_app, :confyg)`). This yielded `nil` and thus caused the following somewhat cryptic error:

```elixir
iex> defmodule Stack do
  use GenServer

  def init(args) do
    {:ok, args}
  end

  def handle_call(:pop, _from, [h | t]) do
    {:reply, h, t}
  end
end

iex> {:ok, pid} = GenServer.start_link(Stack, [])

iex> GenServer.call(pid, :pop, nil)
** (FunctionClauseError) no function clause matching in :gen.call/4

    The following arguments were given to :gen.call/4:

        # 1
        #PID<0.157.0>

        # 2
        :"$gen_call"

        # 3
        :pop

        # 4
        nil

    (stdlib) gen.erl:152: :gen.call/4
    (elixir) lib/gen_server.ex:986: GenServer.call/3
```

Discussion here: https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!topic/elixir-lang-core/t8W9sM84eJY

### DIFF
--- a/lib/elixir/lib/gen_server.ex
+++ b/lib/elixir/lib/gen_server.ex
@@ -989,6 +989,18 @@ defmodule GenServer do
   element.
   """
   @spec call(server, term, timeout) :: term
+  def call(_server, _request, timeout)
+      when (not is_integer(timeout) or timeout < 0) and timeout != :infinity do
+    raise ArgumentError, """
+    expected timeout to be one of the following:
+
+    * integer greater than 0
+    * atom :infinity
+
+    Got: #{inspect(timeout)}
+    """
+  end
+
   def call(server, request, timeout \\ 5000) do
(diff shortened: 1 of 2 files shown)


josevalim:
Hi @dnlserrano. I pushed a simpler approach to master that should improve the current error messages as it blames clauses. Thank you!

PR #8840, opened 2019-02-28
Add typespec for Macro.generate_arguments/2

amatalai:


lib/elixir/lib/macro.ex
```diff
@@ -206,6 +206,8 @@ defmodule Macro do
 
   """
   @doc since: "1.5.0"
+  @spec generate_arguments(0, term) :: []
+  @spec generate_arguments(pos_integer, context) :: [{atom, [], context}] when context: atom
```
whatyouhide: We should specify that this is going to be a non-empty list:

```suggestion
  @spec generate_arguments(pos_integer, context) :: [{atom, [], context}, ...] when context: atom
```

PR #8843, opened 2019-03-01
Enable Access.at/1 to handle negative index

carlo-colombo:
Enabling negative index as Enum.at/2 to keep it consistent, as Enum.at/2 returns nil if the index is out of bounds counting from the end.

### DIFF
--- a/lib/elixir/lib/access.ex
+++ b/lib/elixir/lib/access.ex
@@ -610,10 +610,16 @@ defmodule Access do
       iex> list = [%{name: "john"}, %{name: "mary"}]
       iex> get_in(list, [Access.at(1), :name])
       "mary"
+      iex> get_in(list, [Access.at(-1), :name])
+      "mary"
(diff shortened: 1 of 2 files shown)


lib/elixir/lib/access.ex
```diff
@@ -669,7 +670,17 @@ defmodule Access do
     end
   end
 
-  defp get_and_update_at([head | rest], index, next, updates) do
+  defp get_and_update_at(list, index, next, updates)
+       when index < 0 and length(list) + index >= 0 do
```
josevalim: This is computing the length twice (here and below). I would recommend to move this to inside the function body and  merge this clause with the one below so we have to do it only once.

carlo-colombo: hi, thanks for the quick feedback. Update the PR  merging the the functions and removing the length calculation from the guard

josevalim:

PR #8846, opened 2019-03-01
Add System.fetch_env/1 and System.fetch_env!/1

whatyouhide:
This PR adds the `System.fetch_env/1` and `System.fetch_env!/1` functions as per the discussion in https://groups.google.com/forum/#!topic/elixir-lang-core/QGxEibK4y7Q.

I chose to keep using the same underlying implementation for `fetch_env/1` as the one for `get_env/1` because the number of lines would have been the same but we would have added one function call.

### DIFF
--- a/lib/elixir/lib/system.ex
+++ b/lib/elixir/lib/system.ex
@@ -421,6 +421,15 @@ defmodule System do
   The returned value of the environment variable
   `varname` is a string, or `nil` if the environment
   variable is undefined.
+
+  ## Examples
+
+      iex> System.get_env("PORT")
+      "4000"
+
+      iex> System.get_env("NOT_DEFINED")
+      nil
+
   """
   @spec get_env(String.t()) :: String.t() | nil
   def get_env(varname) when is_binary(varname) do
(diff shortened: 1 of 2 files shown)


lib/elixir/lib/system.ex
```diff
@@ -430,6 +439,51 @@ defmodule System do
...
+      "4000"
+
+      iex> System.fetch_env!("NOT_DEFINED")
+      ** (RuntimeError) environment variable "NOT_DEFINED" is not defined
+
+  """
+  @doc since: "1.9.0"
+  @spec fetch_env!(String.t()) :: String.t()
+  def fetch_env!(varname) when is_binary(varname) do
+    get_env(varname) || raise "environment variable #{inspect(varname)} is not defined"
```
thiamsantos: WDYT about raising an error similar to the one raise by `Application.fetch_env!/1`? https://github.com/elixir-lang/elixir/blob/master/lib/elixir/lib/application.ex#L490

whatyouhide: Good call, couldn't remember we used `ArgumentError` in `Application.fetch_env!/1`. Updated!

spencerdcarlson:
Added default values to this idea - https://github.com/elixir-lang/elixir/pull/8897

PR #8847, opened 2019-03-02
WIP: Move struct special form into kernel

blatyo:
This moves the `%/2` special form to be implemented as a macro on `Kernel`. Most of this code is a direct port of the Erlang code in `elixir_map.erl` into Elixir. The change that allows this to work is that the expansion of `%/2` also expands it into a map now in addition to the other work it was doing. Information about the struct is added to meta for the compile phase. Currently, that's just the struct atom, if that is known. Not sure what would be useful in cases where the struct name is a pattern or pinned pattern. One test is failing currently, which I have yet to look into.


lib/elixir/lib/kernel/struct.ex
```diff
@@ -0,0 +1,199 @@
+defmodule Kernel.Struct do
+  @moduledoc false
+
+  @doc false
+  def expand(struct_or_match, {:%{}, map_meta, map_args}, %{context: context} = env) do
```
blatyo: effectively expand_struct from `elixir_map.erl`

lib/elixir/lib/kernel/struct.ex
```diff
@@ -0,0 +1,199 @@
...
+
+        case extract_struct_assocs(meta, map, env) do
+          # Expand
+          {:expand, map_meta, assocs} when context != :match ->
+            struct = load_struct(meta, struct_or_match, [assocs], env)
+            assert_struct_keys(meta, struct_or_match, struct, assocs, env)
+            keys = [:__struct__] ++ for {k, _} <- assocs, do: k
+            without_keys = :maps.to_list(:maps.without(keys, struct))
+            struct_assocs = :elixir_quote.escape(without_keys, :default, false)
+            expand_to_map(meta, struct_or_match, {:%{}, map_meta, struct_assocs ++ assocs})
```
blatyo: expands to map instead of struct

lib/elixir/lib/kernel/struct.ex
```diff
@@ -0,0 +1,199 @@
...
+
+          # Update or match
+          {_, _, assocs} ->
+            struct = load_struct(meta, struct_or_match, [], env)
+            assert_struct_keys(meta, struct_or_match, struct, assocs, env)
+            expand_to_map(meta, struct_or_match, map)
```
blatyo: same

lib/elixir/lib/kernel/struct.ex
```diff
@@ -0,0 +1,199 @@
...
+        end
+
+      true ->
+        IO.puts("2 - 3")
+
+        {:%, meta, [struct_or_match, map]}
+        |> Macro.to_string()
+        |> IO.puts()
+
+        expand_to_map(meta, struct_or_match, map)
```
blatyo: same

lib/elixir/lib/kernel/struct.ex
```diff
@@ -0,0 +1,193 @@
+defmodule Kernel.Struct do
+  @moduledoc false
+
+  @doc false
+  def expand(struct_or_match, {:%{}, map_meta, map_args}, %{context: context} = env) do
+    meta = [file: env.file, line: env.line]
+    map_args = clean_struct_key_from_map_args(meta, map_args, env)
+
+    {[struct_or_match, map], env} =
+      :elixir_expand.expand_args([struct_or_match, {:%{}, map_meta, map_args}], env)
```
josevalim: We can't really call this here. Because macros cannot return the updated environment, if you expand anything that changes the environment here, such as an import or alias or require, it will be lost.

We can only really expand the first argument (the struct name) and by using Macro.expand. But you have to be careful with bootstrapping. So what I would do is to expand the first argument on kernel before calling this module.

Then the keys have to be expanded manually to with Macro.expand/2.

josevalim:
Great work! I have dropped a comment and we also need to take a look at CI.

josevalim:
> Not sure what would be useful in cases where the struct name is a pattern or pinned pattern. 

This will actually be a blocker on this work. :(

`%x{}` needs to be expanded to `%{__struct__: x} when is_atom(x)` but there is no way to add extra guards during expansion in Elixir (there was a proposal to allow `when/2` anywhere in a match but it was not accepted).

lib/elixir/lib/kernel/struct.ex
blatyo: Thanks!

blatyo:
How does that work currently?

``` elixir
%x{} = %{__struct__: Foo}
%{__struct__: Foo}
x #=> Foo
%x{} = %{__struct__: 1}
# ** (MatchError) no match of right hand side value: %{__struct__: 1}
```

josevalim:
IIRC we emit a case that raises match error in case of failures.
-- 


blatyo:
I was pretty happy that I got pretty much everything to work. Then I removed the struct stuff from `elixir_map.erl` and everything broke, because it was falling back to that implementation until the `%/2` macro became available. 🤪I'll be looking into the bootstrapping stuff. The thing I don't get currently is, what do you do when something isn't bootstrapped? For example, why is this safe? https://github.com/elixir-lang/elixir/blob/master/lib/elixir/lib/kernel.ex#L1845

josevalim:
On bootstrap only special forms are available. We compile a subset of the
language, with limited functionality, and then we recompile it all again
with full functionality. There is usually dependency issues (for example,
you can’t use pipe early on).

You are welcome to play with this further but note that I really think that
%x{} requires it to be a special form, unless we allow when/2 anywhere in a
pattern (which is a longer discussion).

Also note that you will have to add your new file to the list of
bootstrapped files in src/elixir_compiler.erl, as you need to define
structs early in the bootstrap stage.
-- 


blatyo:
Wouldn't this be able to handle `%x{}`, by adding a case statement?

``` elixir
%x{} = %{__struct__: 1}
```

becomes

``` elixir
struct = %{__struct__: x} = %{__struct__: 1}
case struct do
  %{__struct__: module} when is_atom(module) -> struct
  _ -> raise MatchError, "..."
end
```

I guess the case where this wouldn't work is in a function head, but in that case, it could be a normal guard on the function. I'm not sure if there would be a good way to distinguish between the two situations though.

Thanks for the insight into bootstrapping.

josevalim:
That’s only a match. You have to handle case, receive, function clauses,
etc...
-- 


blatyo:
Fair

PR #8848, opened 2019-03-03
Add docs for truthy/falsy values

nielsbom:
In a number of places in the docs (`!/`, `&&/2`, `Enum.all?`) the
concept of a truthy/falsy value is mentioned and briefly explained.
I propose to have single place for that explanation and point to it
from those other places.

I'm not sure if this is the right location though :-)

### DIFF
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -173,6 +173,42 @@ defmodule Kernel do
 
   Those functions will be explicitly marked in their docs as
   "inlined by the compiler".
+  
+  ## Truthy and falsy values
+  
+  Besides the booleans `true` and `false` Elixir also has the
+  concept of a "truthy" or "falsy" value.
+
+    *  a value is truthy when it is neither `false` nor `nil`
+    *  a value is falsy when it is `false` or `nil`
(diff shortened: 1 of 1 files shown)


whatyouhide:
Seems like tests are failing because of the formatter. Can you run `mix format` on the Elixir codebase and see if anything changes?

whatyouhide:
Thanks @nielsbom!💟

lib/elixir/lib/kernel.ex
```diff
@@ -173,6 +173,42 @@ defmodule Kernel do
...
   Those functions will be explicitly marked in their docs as
   "inlined by the compiler".
+
+  ## Truthy and falsy values
+
+  Besides the booleans `true` and `false` Elixir also has the
+  concept of a "truthy" or "falsy" value.
+
+    *  a value is truthy when it is neither `false` nor `nil`
+    *  a value is falsy when it is `false` or `nil`
```
eksperimental:
```
    *  a value is falsy when it is either `false` or `nil`
```

nielsbom: Good point => #8859

PR #8853, opened 2019-03-03
Avoid hyphen at EOL

eksperimental:
Avoid splitting words that contain a hyphen into two lines,
since a space is added when docs are rendered in IEx and ExDoc.

Command used to detect this bug:

$ ag -- "\w-\n"

### DIFF
--- a/lib/mix/lib/mix/tasks/escript.build.ex
+++ b/lib/mix/lib/mix/tasks/escript.build.ex
@@ -24,7 +24,7 @@ defmodule Mix.Tasks.Escript.Build do
   developers and not as a deployment mechanism. For running live
   systems, consider using `mix run` or building releases. See
   the `Application` module for more information on systems
-  life- cycles.
+  life-cycles.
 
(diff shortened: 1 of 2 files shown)


lib/mix/lib/mix/tasks/escript.build.ex
```diff
@@ -23,8 +23,8 @@ defmodule Mix.Tasks.Escript.Build do
   Escripts should be used as a mechanism to share scripts between
   developers and not as a deployment mechanism. For running live
   systems, consider using `mix run` or building releases. See
-  the `Application` module for more information on systems life-
-  cycles.
+  the `Application` module for more information on systems
+  life- cycles.
```
fertapric:
```suggestion
  life-cycles.
```

josevalim:

lib/mix/lib/mix/tasks/escript.build.ex
eksperimental: oops

PR #8855, opened 2019-03-04
Merge paragraphs in Enum moduledoc

eksperimental:
Avoid splitting a paragraph due a code that fits as inline-code.


eksperimental:
It's a single function call, after a few more characters with hit `&(&1 * 2)` and the story is the same but we don't split the paragraph again

fertapric:
Agree with @whatyouhide. I would leave it as is.

PR #8856, opened 2019-03-04
Improvements in the documentation

eksperimental:


lib/elixir/pages/Syntax Reference.md
```diff
@@ -295,7 +295,7 @@ end
 #=> {:{}, [], [1, 2, 3]}
 ```
 
-Binaries have a representation similar to tuples, except they are tagged with `:<<>>` instead of `:{}`:
+Binaries have a representation similar to tuples. They are tagged with with `:<<>>` instead of `:{}`:
```
fertapric:
```suggestion
Binaries have a representation similar to tuples. They are tagged with `:<<>>` instead of `:{}`:
```

josevalim:

lib/elixir/pages/Syntax Reference.md
josevalim: Sorry, merged prematurely. I will update in master.

Also, I don't think a break is appropriate here, it is important to say they are tagged differenty before we finish the sentence. :D

eksperimental: I agree

PR #8858, opened 2019-03-04
Improve documentation in test task cmd line options

eksperimental:


lib/mix/lib/mix/tasks/test.ex
```diff
@@ -153,15 +153,16 @@ defmodule Mix.Tasks.Test do
...
-    * `--formatter` - formatter module
+    * `--formatter` - sets the formatter modules that will print the results.
+      Defaults to `ExUnit.CLIFormatter`
     * `--include` - includes tests that match the filter
     * `--listen-on-stdin` - runs tests, and then listens on stdin. Receiving a newline will
       result in the tests being run again. Very useful when combined with `--stale` and
-      external commands which produce output on stdout upon file system modification
-    * `--max-cases` - sets the maximum number of tests running async. Only tests from
+      external commands which produce output on stdout upon file system modifications
+    * `--max-cases` - sets the maximum number of tests running asyncronously. Only tests from
```
```suggestion
    * `--max-cases` - sets the maximum number of tests running asynchronously. Only tests from
```

lib/mix/lib/mix/tasks/test.ex
```diff
@@ -153,15 +153,16 @@ defmodule Mix.Tasks.Test do
...
       different modules run in parallel. Defaults to twice the number of cores
     * `--max-failures` - the suite stops evaluating tests when this number of test
-      failures is reached. When not given, it always runs all tests
+      failures is reached. It runs all test if omitted
```
```suggestion
      failures is reached. It runs all tests if omitted
```

lib/mix/lib/mix/tasks/test.ex
```diff
@@ -153,15 +153,16 @@ defmodule Mix.Tasks.Test do
     * `--exclude` - excludes tests that match the filter
     * `--failed` - runs only tests that failed the last time they ran
     * `--force` - forces compilation regardless of modification times
-    * `--formatter` - formatter module
+    * `--formatter` - sets the formatter modules that will print the results.
```
fertapric: I'll keep the singular here even if you can pass the `--formatter` option more than once, because with the plural it seems you should pass a list.

```suggestion
    * `--formatter` - sets the formatter module that will print the results.
```

eksperimental: I was in doubt when I edited this. Thank you

josevalim:

Issue #8800, opened 2019-02-15
Inconsistency between %Mod{} and Mod.__struct__() when delegating


mguimas:
### Environment

```
$ elixir --version
Erlang/OTP 21 [erts-10.2.3] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1] [hipe]

Elixir 1.8.1 (compiled with Erlang/OTP 21)
```

### Current behavior

Given the file `xx.ex`

```
defmodule XX do
  defstruct [:field]
end
```

the following holds for `XX`

```
iex(2)> %XX{}
%XX{field: nil}
iex(3)> XX.__struct__()
%XX{field: nil}
iex(4)> struct(XX)
%XX{field: nil}
iex(5)> %XX{} == XX.__struct__()
true
```

but given the file `yy.ex`

```
defmodule YY do
  defdelegate __struct__, to: XX
  defdelegate __struct__(x), to: XX
end
```

the follwing happens for `YY`

```
iex(6)> %YY{}
%YY{field: nil}
iex(7)> YY.__struct__()
%XX{field: nil}
iex(8)> struct(YY)
%XX{field: nil}
iex(9)> %YY{} == YY.__struct__()
false
```

in which different idioms that are expected to be equivalent return inconsistent results.

### Expected behavior

The expected behavior is for `%Mod{}` to be equivalent to `Mod.__struct__()`, consequently the function `Kernel.SpecialForms.%/2` must be fixed to return the value of `Mod.__struct__/0` (and of `Mod.__struct__/1` as applicable).

The `defdelegate` construct is working properly because it returns the result of `XX.__struct__/{0,1}` as the result of `YY.__struct__/{0,1}`, as expected for delegation.

This means that somehow `Kernel.SpecialForms.%/2` is overwritting the result returned by `YY.__struct__/{0,1}`, and it must not, and instead pass it "as is".

Note that fixing `Kernel.SpecialForms.%/2` additionally solves the case when `XX` is an Ecto schema. For example, given

```
defmodule XX do
  use Ecto.Schema
  schema "table" do
    field :field
  end
end
```

the following is happening now

```
iex(11)> %YY{}
%YY{__meta__: #Ecto.Schema.Metadata<:built, "table">, field: nil, id: nil}
iex(12)> %YY{} |> IO.inspect(structs: false, pretty: true)
%{
  __meta__: %{
    __struct__: Ecto.Schema.Metadata,
    context: nil,
    prefix: nil,
[...]

PR #8864, opened 2019-03-07
Validate __struct__ key in map returned by __struct__/0,1

fertapric:
Related to #8800

### DIFF
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -2178,8 +2178,11 @@ defmodule Kernel do
   end
 
   defp validate_struct!(expr, module, arity) do
-    error_message = :elixir_map.format_error({:invalid_struct_return_value, module, arity, expr})
-    :erlang.error(ArgumentError.exception(:erlang.iolist_to_binary(error_message)))
+    error_message =
+      "expected #{inspect(module)}.__struct__/#{arity} to return a map with a :__struct__ " <>
+        "key that holds the name of the struct (atom), got: #{inspect(expr)}"
+
+    :erlang.error(ArgumentError.exception(error_message))
   end
 
   @doc """
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -2144,7 +2144,7 @@ defmodule Kernel do
   def struct!(struct, fields \\ [])
 
   def struct!(struct, fields) when is_atom(struct) do
-    struct.__struct__(fields)
+    validate_struct!(struct.__struct__(fields), struct, 1)
   end
 
   def struct!(struct, fields) when is_map(struct) do
@@ -2158,11 +2158,11 @@ defmodule Kernel do
   end
 
   defp struct(struct, [], _fun) when is_atom(struct) do
-    struct.__struct__()
+    validate_struct!(struct.__struct__(), struct, 0)
   end
 
   defp struct(struct, fields, fun) when is_atom(struct) do
-    struct(struct.__struct__(), fields, fun)
+    struct(validate_struct!(struct.__struct__(), struct, 0), fields, fun)
   end
 
   defp struct(%_{} = struct, [], _fun) do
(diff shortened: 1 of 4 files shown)


lib/elixir/lib/kernel.ex
```diff
@@ -2173,6 +2173,15 @@ defmodule Kernel do
     Enum.reduce(fields, struct, fun)
   end
 
+  defp validate_struct!(%{__struct__: module} = struct, _module, _arity) when is_atom(module) do
+    struct
+  end
+
+  defp validate_struct!(expr, module, arity) do
+    error_message = :elixir_map.format_error({:invalid_struct_return_value, module, arity, expr})
```
josevalim: This looks like too much work! 😅 I would just duplicate the error message. Thoughts?

lib/elixir/src/elixir_map.erl
```diff
@@ -217,8 +217,10 @@ format_error({non_map_after_struct, Expr}) ->
   io_lib:format("expected struct to be followed by a map, got: ~ts",
                 ['Elixir.Macro':to_string(Expr)]);
 format_error({invalid_struct_return_value, Module, Arity, Expr}) ->
-  io_lib:format("expected ~ts.__struct__/~p to return a map, got: ~ts",
-                ['Elixir.Macro':to_string(Module), Arity, 'Elixir.Macro':to_string(Expr)]);
+  Message =
+    "expected ~ts.__struct__/~p to return a map with a :__struct__ key that holds the "
+    "name of the struct (atom), got: ~ts",
+  io_lib:format(Message, ['Elixir.Macro':to_string(Module), Arity, 'Elixir.Macro':to_string(Expr)]);
```
josevalim:
```suggestion
  io_lib:format(Message, [elixir_aliases:inspect(Module), Arity, 'Elixir.Macro':to_string(Expr)]);
```

fertapric: I will change other occurrences in a different PR.

lib/elixir/lib/kernel.ex
fertapric: I also had doubts.

If I duplicate the error message I would need to `:elixir_aliases.inspect(module)` and `Macro.to_string(expr)`, which more or less balances with the `:elixir_map.format_error` and `:erlang.iolist_to_binary`. And this way the error message are in sync.

But I don't have a strong opinion here. Your call 🙂

lib/elixir/src/elixir_map.erl
josevalim: Although I can see we used `Elixir.Macro:to_string` before .

lib/elixir/lib/kernel.ex
josevalim: On this side of things, it is ok to use inspect + Macro.to_string. We can keep this more Elixir looking than the rest.

lib/elixir/lib/kernel.ex
```diff
@@ -2173,6 +2173,18 @@ defmodule Kernel do
...
   end
 
+  defp validate_struct!(%{__struct__: module} = struct, _module, _arity) when is_atom(module) do
+    struct
+  end
+
+  defp validate_struct!(expr, module, arity) do
+    error_message =
+      "expected #{inspect(module)}.__struct__/#{arity} to return a map with a :__struct__ " <>
+        "key that holds the name of the struct (atom), got: #{inspect(expr)}"
```
fertapric: I finally didn't use `Macro.to_string/1` because I think `__struct__/0,1` is not meant to return an AST.

I see on the Erlang side we are using `'Elixir.Macro':to_string`, when I think `'Elixir.Kernel':inspect` should be used, because it might lead to misinterpretations. Here is an example:

```elixir
iex(1)> defmodule Foo do
...(1)> def __struct__, do: {:ok, :one, :two}
...(1)> def __struct__(_), do: {:ok, :one, :two}
...(1)> end
{:module, Foo,
 <<70, 79, 82, 49, 0, 0, 4, 148, 66, 69, 65, 77, 65, 116, 85, 56, 0, 0, 0, 128,
   0, 0, 0, 13, 10, 69, 108, 105, 120, 105, 114, 46, 70, 111, 111, 8, 95, 95,
   105, 110, 102, 111, 95, 95, 7, 99, 111, ...>>, {:__struct__, 1}}
iex(2)> %Foo{}
** (CompileError) iex:2: expected Foo.__struct__/1 to return a map with a :__struct__ key that holds the name of the struct (atom), got: ok
```

I will fix that (and other occurrences) in a different PR 🙂

josevalim: Good catch.

josevalim:

PR #8865, opened 2019-03-07
More examples for Kernel.function_exported?/3

shmink:
I personally found the current docs for [function_exported/3](https://hexdocs.pm/elixir/Kernel.html#function_exported?/3) a bit odd. 
```elixir
iex> function_exported?(Enum, :member?, 2)
true
```
Initially I thought the example given with the second parameter was an option. I read it as if the following. "Is there a function in Enum with 2 arity?". I later realised that it was referring to `Enum.member?/2` thanks to @Gazler so I added some more examples that I believe are more memorable functions that aren't close to common option names.

### DIFF
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -3405,9 +3405,14 @@ defmodule Kernel do
 
   ## Examples
 
-      iex> function_exported?(Enum, :member?, 2)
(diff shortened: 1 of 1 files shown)


josevalim:

josevalim:
200 is too big and almost unexpected as a function arity. I think 10 is enough as "too large".

PR #8869, opened 2019-03-08
Use :elixir_aliases.inspect/1 with module aliases

fertapric:
The `Module` variable in these cases is guaranteed to be a module alias.

Follow up of https://github.com/elixir-lang/elixir/pull/8864#discussion_r263304032


josevalim:

PR #8870, opened 2019-03-08
Add guard to Macro.to_string/2

fertapric:
Three-element tuples with an atom at the end raise an `ArgumentError` instead of using inspect (fallback):

```elixir
iex(1)> Macro.to_string({1, 2, :hello})
    :erlang.atom_to_binary(1, :utf8)
    (elixir) lib/macro.ex:621: Macro.to_string/2
iex(1)> Macro.to_string({1, 2, :hello, 1})
"{1, 2, :hello, 1}"
```


josevalim:

PR #8872, opened 2019-03-08
Ensure inspect returns valid ~r// expressions

### LINKED ISSUE #8860: A bug in regex pretty-printer
### Environment
```
$ elixir --version
Erlang/OTP 21 [erts-10.2.3] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [hipe] [dtrace]

Elixir 1.8.1 (compiled with Erlang/OTP 21)
```
The same behaviour is observed in v1.7.3

* Operating system: Linux and MacOS

### Current behaviour
```
$ iex
Erlang/OTP 21 [erts-10.2.3] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [hipe] [dtrace]

iex(1)> re = ~r/a\/b/
~r/a\/b/
iex(2)> {:ok, re2} = Regex.compile("a\\/b")
{:ok, ~r/a\\/b/}
iex(3)> {:ok, re3} = Regex.compile("a\/b")
{:ok, ~r/a\/b/}
iex(4)> Regex.match?(re, "a/b")
true
iex(5)> Regex.match?(re2, "a/b")
true
iex(6)> Regex.match?(re3, "a/b")
true
iex(7)> Regex.match?(~r/a\/b/, "a/b")
true
iex(8)> Regex.match?(~r/a\\/b/, "a/b")
** (SyntaxError) iex:8: syntax error before: ','
```

### Expected behaviour

Pretty printing of `re2` seems to be broken. I believe it should be either 
```
~r/a\/b/
```
or
```
~r/a\\\/b/
```
because `~r/a\\/b/` is syntactically incorrect, see the `SyntaxError` exception.

Given that `re2` matches string `"a/b"` it should be the former - `~r/a\/b/`.

fertapric:
This is an attempt to fix #8860

### DIFF
--- a/lib/elixir/lib/inspect.ex
+++ b/lib/elixir/lib/inspect.ex
@@ -307,10 +307,17 @@ end
 
 defimpl Inspect, for: Regex do
   def inspect(regex, opts) do
-    {escaped, _} = Identifier.escape(regex.source, ?/, :infinity, &escape_map/1)
+    {escaped, _} =
+      regex.source
+      |> :elixir_interpolation.unescape_chars(&unescape_map/1)
+      |> Identifier.escape(?/, :infinity, &escape_map/1)
+
     source = IO.iodata_to_binary(['~r/', escaped, ?/, regex.opts])
     color(source, :regex, opts)
   end
+
+  defp unescape_map(?/), do: ?/
+  defp unescape_map(_), do: false
 
   defp escape_map(?\a), do: '\\a'
   defp escape_map(?\f), do: '\\f'
--- a/lib/elixir/lib/inspect.ex
+++ b/lib/elixir/lib/inspect.ex
@@ -309,15 +309,17 @@ defimpl Inspect, for: Regex do
   def inspect(regex, opts) do
     {escaped, _} =
       regex.source
-      |> :elixir_interpolation.unescape_chars(&unescape_map/1)
+      |> normalize(<<>>)
       |> Identifier.escape(?/, :infinity, &escape_map/1)
 
     source = IO.iodata_to_binary(['~r/', escaped, ?/, regex.opts])
     color(source, :regex, opts)
   end
 
-  defp unescape_map(?/), do: ?/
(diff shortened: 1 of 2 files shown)


lib/elixir/lib/inspect.ex
```diff
@@ -307,11 +307,18 @@ end
 
 defimpl Inspect, for: Regex do
   def inspect(regex, opts) do
-    {escaped, _} = Identifier.escape(regex.source, ?/, :infinity, &escape_map/1)
+    {escaped, _} =
+      regex.source
+      |> :elixir_interpolation.unescape_chars(&unescape_map/1)
```
josevalim: Unless I am missing something, isn't this the same as:

```suggestion
      |> String.replace("/", "\\/")
```

fertapric: It would be the other way around, replacing `"\\/"` with `"/"`, but only for single backslashes `"\\"`:

```elixir
iex(1)> :elixir_interpolation.unescape_chars("a\\/b", fn ?/ -> ?/; _ -> false end)
"a/b"
iex(2)> :elixir_interpolation.unescape_chars("a\\\\/b", fn ?/ -> ?/; _ -> false end)
"a\\\\/b"
iex(3)> :elixir_interpolation.unescape_chars("a\\\\\\/b", fn ?/ -> ?/; _ -> false end)
"a\\\\/b"
iex(4)> String.replace("a\\/b", "\\/", "/")
"a/b"
iex(5)> String.replace("a\\\\/b", "\\/", "/")
"a\\/b"
iex(6)> String.replace("a\\\\\\/b", "\\/", "/")
"a\\\\/b"
```

josevalim: Yes, I see what you mean. I would do a loop by hand then, using unescape_chars may be a bit too much. Thoughts?

fertapric: I’ll give it a try 🙂

PR #8873, opened 2019-03-08
Use inspect/1 with values returned by __struct__/*

fertapric:
`__struct__/0,1` is not meant to return an AST. Using `Macro.to_string/1` might lead to incorrect error messages:

```elixir
    defmodule MyStruct do
      def __struct__, do: {:ok, :one, :two}
      def __struct__(_), do: {:ok, :one, :two}
    end

    iex> %MyStruct{}
    ** (CompileError) iex:2: expected MyStruct.__struct__/1 to return a map with a :__struct__ key that holds the name of the struct (atom), got: ok
```

Follow up of https://github.com/elixir-lang/elixir/pull/8864#discussion_r263334936 (I thought I was going to find more occurrences 😅)


josevalim:

PR #8874, opened 2019-03-10
Fix small typo for ceil behavior docs

codebycaleb:
I got caught off-guard by the existing documentation, so I just wanted to submit a quick fix for it. Would be happy to switch to ```Returns the smallest integer greater than or equal to `number`.``` if that sounds better!

### DIFF
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -354,7 +354,7 @@ defmodule Kernel do
   end
 
   @doc """
-  Returns the smallest integer not greater than `number`.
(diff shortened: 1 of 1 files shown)


josevalim:
Wouldn’t greater be better to avoid the negation? How are the docs for
floor? I would just mirror that.
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

codebycaleb:
Good question! Floor docs also contain the negation:

`Returns the largest integer not greater than number`

I've removed both negations and squashed my commits together. 🙂

josevalim:

PR #8890, opened 2019-03-13
Quote :: to avoid ambiguity

### LINKED ISSUE #8832: Invalid formatting of :: op as keyword
### Environment

* Elixir & Erlang/OTP versions (elixir --version): Elixir 1.8.1 (Erlang/OTP 21.1)
* Operating system: 

### Current behavior

If a keyword list contains a key with the atom consisting of two colons (with the sugar syntax) the formatter emits non-compilable code.

Example in the shell
```
iex(2)> kwlist = ["::": 2]
[::: 2]
iex(3)> kwlist |> Macro.escape() |> Macro.to_string() |> Code.string_to_quoted()
{:error, {1, "syntax error before: ", "\"2\""}}
```
The quotes are removed when formatting the key which is ambiguous.
(note that quotes are only removed in case of 2 colons, because it is an operator)
```
iex(19)> kwlist = for i <- 0..4, do: {List.duplicate(?:, i)|>List.to_atom, i}
["": 0, ":": 1, ::: 2, ":::": 3, "::::": 4]
```

The formatter also behaves the same way ie it removes the quotes.

This is not a practical issue, only breaks the guarantee that any random program that can be compiled should also be compilable after auto-formatting.

### Expected behavior

Formatter should not break comilability.

josevalim:
Closes #8832.

### DIFF
--- a/lib/elixir/lib/code/identifier.ex
+++ b/lib/elixir/lib/code/identifier.ex
@@ -34,7 +34,7 @@ defmodule Code.Identifier do
     cond do
       op in [:<-, :\\] -> {:left, 40}
       op in [:when] -> {:right, 50}
-      op in [:::] -> {:right, 60}
+      op in [:"::"] -> {:right, 60}
       op in [:|] -> {:right, 70}
       op in [:=] -> {:right, 100}
       op in [:||, :|||, :or] -> {:left, 130}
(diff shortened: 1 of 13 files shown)


lib/elixir/lib/code/identifier.ex
```diff
@@ -80,6 +80,9 @@ defmodule Code.Identifier do
       atom in [:%, :%{}, :{}, :<<>>, :..., :.., :., :->] ->
         :not_callable
 
+      atom in [:"::"] ->
+        :not_atomable
```
fertapric: Should `:not_atomable` be documented?

josevalim: Good catch, done!

josevalim:

Issue #8888, opened 2019-03-13
@type type_a(_a) :: any() compile failed


slepher:
https://github.com/elixir-lang/elixir/pull/8234

this pr will cause this code compile fail
@type type_a(_a) :: any()
variable name begins with _ should ignore singleton_typevar check

slepher:
by the way
-type type_a(_) :: any(). in erlang compile fails with "bad type variable"
@type type_a(_) :: any() should fail which fails at v1.8.1
@type type_a(_, _) :: any() should also fail. which passes at v1.8.1

slepher:
are you going to do any fix?
I will create a pr for this.

slepher:
pr https://github.com/elixir-lang/elixir/pull/8891 created

PR #8891, opened 2019-03-14
Handle underscore variables (_, _foo) in typespecs

slepher:
fix issue
https://github.com/elixir-lang/elixir/issues/8888
and add three test case

### DIFF
--- a/lib/elixir/lib/kernel/typespec.ex
+++ b/lib/elixir/lib/kernel/typespec.ex
@@ -962,7 +962,7 @@ defmodule Kernel.Typespec do
   end
 
   defp ensure_no_underscore_local_vars!(caller, var_names) do
-    case :lists.any(&(&1 == :_), var_names) do
+    case :lists.member(:_, var_names) do
       true ->
         compile_error(caller, "type variable '_' is invalid")
 
--- a/lib/elixir/lib/kernel/typespec.ex
+++ b/lib/elixir/lib/kernel/typespec.ex
@@ -962,7 +962,15 @@ defmodule Kernel.Typespec do
 
   defp ensure_no_unused_local_vars!(caller, local_vars) do
     fun = fn
-      {name, :used_once} -> compile_error(caller, "type variable #{name} is unused")
+      {:_, _used} ->
+        compile_error(caller, "type variable '_' is invalid")
+      {name, :used_once} ->
+        case :erlang.atom_to_list(name) do
+          [?_|_] ->
+            :ok
+          _ ->
+            compile_error(caller, "type variable #{name} is unused")
+        end
       _ -> :ok
     end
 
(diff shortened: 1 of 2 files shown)


lib/elixir/lib/kernel/typespec.ex
```diff
@@ -962,8 +962,20 @@ defmodule Kernel.Typespec do
 
   defp ensure_no_unused_local_vars!(caller, local_vars) do
     fun = fn
-      {name, :used_once} -> compile_error(caller, "type variable #{name} is unused")
-      _ -> :ok
+      {:_, _used} ->
+        compile_error(caller, "type variable '_' is invalid")
```
fertapric: I would move this check to `update_local_vars`

slepher: ok, thanks.

lib/elixir/test/elixir/typespec_test.exs
```diff
@@ -128,6 +128,28 @@ defmodule TypespecTest do
       end
     end
 
+    test "type variable unused ignore variable started with _" do
+      test_module do
+        assert @type(foo(_hello) :: integer) == :ok
+      end
+    end
+
+    test "type variable _ should be invalid" do
```
fertapric: Nitpick: I think we can join there two tests

```elixir
    test "type variable _ should be invalid" do
      assert_raise CompileError, ~r"type variable '_' is invalid", fn ->
        test_module do
          @type foo(_) :: integer
        end
      end

      assert_raise CompileError, ~r"type variable '_' is invalid", fn ->
        test_module do
          @type foo(_, _) :: integer
        end
      end
    end
```

lib/elixir/test/elixir/typespec_test.exs
```diff
@@ -128,6 +128,28 @@ defmodule TypespecTest do
       end
     end
 
+    test "type variable unused ignore variable started with _" do
```
fertapric:
```suggestion
    test "@type with a variable starting with underscore" do
```

fertapric:
Thanks @slepher! Overall looks good 👌

lib/elixir/lib/kernel/typespec.ex
slepher: I noticed ensure_no_unused_local_vars! is also used in translate_spec
it's not works as expected.
I will move this check to new function ensure_no_underscore_local_vars! and use it only in translate_type.

lib/elixir/lib/kernel/typespec.ex
```diff
@@ -960,10 +961,29 @@ defmodule Kernel.Typespec do
     end
   end
 
+  defp ensure_no_underscore_local_vars!(caller, var_names) do
+    case :lists.any(&(&1 == :_), var_names) do
```
josevalim:
```suggestion
    case :lists.member(:_, var_names) do
```

fertapric:
Thanks @slepher!

PR #8895, opened 2019-03-15
Extract types from GenServer callbacks

devonestes:
The possible returns for the asynchronous callbacks in a GenServer
(`handle_cast`, `handle_info` and `handle_continue`) were all the same,
but that wasn't easy to see as they all had their own type definition
for the return value in the callback definition.

Also, this will make adding `@spec` annotations to modules implemeting
the `GenServer` behaviour much easier, as they'll now have public types
in `GenServer` that can be used in the same way that
`GenServer.on_start` is often used now.

### DIFF
--- a/lib/elixir/lib/gen_server.ex
+++ b/lib/elixir/lib/gen_server.ex
@@ -444,12 +444,8 @@ defmodule GenServer do
   `{:error, reason}` and the process to exit with reason `reason` without
   entering the loop or calling `c:terminate/2`.
   """
-  @callback init(init_arg :: term) ::
-              {:ok, state}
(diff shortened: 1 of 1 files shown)


josevalim:
The downside is that documentation is worse because now the docs and all
return types are apart, requiring back and forth.

That’s why I never mind repeating the callbacks. I prefer that. And if you
have your own callbacks, it helps you explicitly list that you support all
cases. For example, once we added continue, you may not directly handle it
yet.

With all that said, I would prefer to keep the code as is. :) thanks!
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

PR #3798, opened 2015-09-29
Add support for a default value in System.get_env/1

guilleiguaran:
Similarly to Erlang `os:getenv/2` this allow to return a default value in `System.get_env` when the environment variable is undefined.

PR #8897, opened 2019-03-16
Adding System.fetch/2

spencerdcarlson:
Fetch env by `varname ` and return a default if the `varname` is not set. Handle `String.to_integer` inferred from default type

### DIFF
--- a/lib/elixir/lib/system.ex
+++ b/lib/elixir/lib/system.ex
@@ -447,13 +447,37 @@ defmodule System do
     end
   end
 
+  @doc """
+  Returns the value of the given environment variable or the supplied default value.
+
+  If the environment variable `varname` is set, then the value is returned.
+  If `varname` is not set, then the supplied `default` value is returned.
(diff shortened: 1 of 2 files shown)


josevalim:
The default value should be added to System.get_env/3, as to mirror the
Map/Keyword API
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

josevalim:
Sorry, System.get_env/2. You should be able to add a default argument to
the existing implementation. Then you just need to update the docs and add
tests. :)
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

guilleiguaran:
@josevalim I sent a PR for `get_env` some time ago: https://github.com/elixir-lang/elixir/pull/3798 and was rejected, I can rebase it if you changed your mind about the reasons 😅

josevalim:
Perfect, I will merge it then. Thank you @spencerdcarlson and @guilleiguaran !

PR #8750, opened 2019-02-01
Check for POSIX compliant shell scripts in CI

eksperimental:
WIP because I need to run it against the CI to see if it works

PR #8898, opened 2019-03-17
Add OTP 21.3 to the build matrix

PragTob:
Wasn't sure if the posix compliant check should be added to
only the newest or 21.2+ so I kept it with both - #8750 wasn't
explicit about this.


whatyouhide:
Thanks @PragTob! 💟

eksperimental:
Yes, it is meant to be only in the latest stable OTP release

PR #8899, opened 2019-03-17
Struct updating docs

PragTob:
I felt that the option to use `update_in/2`/`put_in/2` & `get_and_update_in/2` with structs wasn't really prominent in the docs while I believe it's very important and a great feature.

Tried to highlight it a bit more in all of these that it also works.

Happy for feedback about wording, example change or what not.

I wanted to add doc tests but my doctests always ended up telling me `Foooo.__struct__/1 is undefined, cannot expand struct Foooo` (double checked module/struct definition and usage of course)

so I felt just adding them to the tests was best.

Thanks for all your work :clap:

### DIFF
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -2435,7 +2435,7 @@ defmodule Kernel do
       put_in(opts, [:foo, :bar], :baz)
 
   This also works with nested structs and the `struct.path.to.value` way to specify
-  paths"
+  paths:
 
       put_in(struct.foo.bar, :baz)
 
@@ -2512,7 +2512,7 @@ defmodule Kernel do
       update_in(opts, [:foo, :bar], &(&1 + 1))
 
   This also works with nested structs and the `struct.path.to.value` way to specify
-  paths"
+  paths:
 
       update_in(struct.foo.bar, &(&1 + 1))
 
@@ -2555,7 +2555,7 @@ defmodule Kernel do
       get_and_update_in(opts, [:foo, :bar], &{&1, &1 + 1})
 
   This also works with nested structs and the `struct.path.to.value` way to specify
-  paths"
+  paths:
 
       get_and_update_in(struct.foo.bar, &{&1, &1 + 1})
 
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -2434,6 +2434,11 @@ defmodule Kernel do
 
       put_in(opts, [:foo, :bar], :baz)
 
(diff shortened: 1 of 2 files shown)


lib/elixir/test/elixir/kernel_test.exs
```diff
@@ -547,6 +547,10 @@ defmodule KernelTest do
   end
 
   describe "access" do
+    defmodule StructAccess do
+      defstruct [:foo, :bar]
+    end
```
PragTob: Not sure if it was best to put here or if it should be inlined in the tests

lib/elixir/test/elixir/kernel_test.exs
```diff
@@ -584,6 +588,11 @@ defmodule KernelTest do
 
       assert put_in(users["john"].age, 28) == %{"john" => %{age: 28}, "meg" => %{age: 23}}
 
+      struct = %StructAccess{foo: %StructAccess{}}
+
+      assert put_in(struct.foo.bar, :baz) ==
+               %StructAccess{bar: nil, foo: %StructAccess{bar: :baz, foo: nil}}
```
PragTob: I added this to the tests of each individual function, if this is more considered an axillary concern I can see how putting all the three examples into separate tests also makes total sense.

lib/elixir/lib/kernel.ex
```diff
@@ -2434,6 +2434,11 @@ defmodule Kernel do
 
       put_in(opts, [:foo, :bar], :baz)
 
+  This also works with nested structs and the `struct.path.to.value` way to specify
+  paths"
```
whatyouhide:
```suggestion
  paths:
```

lib/elixir/lib/kernel.ex
```diff
@@ -2506,6 +2511,11 @@ defmodule Kernel do
 
       update_in(opts, [:foo, :bar], &(&1 + 1))
 
+  This also works with nested structs and the `struct.path.to.value` way to specify
+  paths"
```
whatyouhide:
```suggestion
  paths:
```

lib/elixir/lib/kernel.ex
```diff
@@ -2544,6 +2554,11 @@ defmodule Kernel do
 
       get_and_update_in(opts, [:foo, :bar], &{&1, &1 + 1})
 
+  This also works with nested structs and the `struct.path.to.value` way to specify
+  paths"
```
whatyouhide:
```suggestion
  paths:
```

PragTob: argh what a stupid mistake, thank you :green_heart:

josevalim:

PR #8900, opened 2019-03-17
Ensure Erlang-based Mix compilers (erlang, leex, yecc) set valid position on diagnostics

JakeBecker:
The `Mix.Task.Compiler.Diagnostic` spec allows position to be `nil`, but the yecc compiler was sometimes setting it to `:none` which violates the spec.


whatyouhide:
Thanks @JakeBecker! 💟

PR #8905, opened 2019-03-21
Ensure guard checks when expanding string interpolations

fertapric:
During the bitstring expansion, guard checks were avoided due to a special case for inlining binaries during interpolation. These changes ensure guards are checked.

Related to #8867

### DIFF
--- a/lib/elixir/test/elixir/kernel/expansion_test.exs
+++ b/lib/elixir/test/elixir/kernel/expansion_test.exs
@@ -595,6 +595,20 @@ defmodule Kernel.ExpansionTest do
         expand(quote(do: fn arg when make_ref() -> arg end))
       end
     end
+
+    test "in guards with bitstrings" do
+      message = ~r"cannot invoke remote function String.Chars.to_string/1 inside guards"
+
+      assert_raise CompileError, message, fn ->
+        expand(
+          quote do
+            fn arg when <<:"Elixir.Kernel".to_string(arg)::binary, "foo">> == "argfoo" ->
+              arg
+            end
+          end
+        )
+      end
+    end
   end
 
   describe "comprehensions" do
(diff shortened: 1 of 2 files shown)


fertapric:
In the issue I said:

> However, I think string interpolations in the form of `"#{arg} string"` should raise, because it implies calling the `String.Chars` protocol. If we make it work for some cases, users might expect this to work:
>
> ```elixir
> defmodule Foo do
>   def foo(a, b) when a in ["#{b}.market"] do
>     a
>   end
> end
> 
> Foo.foo("1.market", 1)
> ```

But, in fact, it does work for some cases: if `arg` is a literal or is expanded to a literal. For example:

```elixir
defmodule Foo do
  @foo "foo"
  defmacro bar, do: "bar"

  def foo(a) when a in ["#{@foo}.foo"] do
    a
  end

  def bar(a) when a in ["#{bar()}.bar"] do
    a
  end

  def wadus(a) when a in ["#{"wadus"}.wadus"] do
    a
  end
end
```

Raising in those cases would be a backwards incompatible change.

lib/elixir/test/elixir/kernel/expansion_test.exs
```diff
@@ -595,6 +595,24 @@ defmodule Kernel.ExpansionTest do
         expand(quote(do: fn arg when make_ref() -> arg end))
       end
     end
+
+    test "in guards with bitstrings" do
+      message = ~r"cannot invoke remote function String.Chars.to_string/1 inside guards"
+
+      assert_raise CompileError, message, fn ->
```
fertapric: I did not include this before because I thought string interpolation in guards was going to be treated differently.

josevalim: Should we explicitly add a test for the cases that *work*? Which is literal and the results from expansion?

fertapric: The test suite has these ones (`bar/0` is defined at the top`defmacro bar, do: "bar"`):

https://github.com/elixir-lang/elixir/blob/7aaf02a28d3ace8bfc19c9960997efe82c108df4/lib/elixir/test/elixir/kernel/expansion_test.exs#L1993-L2007

I think that's enough, wdyt?

josevalim: Perfect, yes. Ship it!

PR #8908, opened 2019-03-24
Raise when retrieving capitalized module attribute

### LINKED ISSUE #8907: Capital word in module attributes
### Precheck

[origin proposal](https://groups.google.com/forum/#!topic/elixir-lang-core/2RHCgsoOnAw)

### Environment

Erlang/OTP 21 [erts-10.3] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [hipe]

Elixir 1.8.1 (compiled with Erlang/OTP 20)

### Current behavior

```ex
defmodule M do
  
  Module.put_attribute(__MODULE__, :A, 1)

  IO.inspect @A
end
```

iex -S mix
```shell
Compiling 1 files (.ex)
:ok
warning: module attribute @__aliases__ was set but never used
  lib/m.ex:5
```

### Expected behavior

Raising an error when trying to use capital word in module attribute (maybe better error type and message):
```
** (ArgumentError) module attributes cannot start with an uppercase letter
```

or support capital word in module attributes, could just ignore alias.

edisonywh:
Fixes #8907

### DIFF
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -2867,6 +2867,10 @@ defmodule Kernel do
   """
   defmacro @expr
 
+  defmacro @{:__aliases__, _meta, _args} do
+    raise ArgumentError, "module attributes set via @ cannot start with an uppercase letter"
+  end
+
   defmacro @{name, meta, args} do
     assert_module_scope(__CALLER__, :@, 1)
     function? = __CALLER__.function != nil
--- a/lib/elixir/test/elixir/kernel_test.exs
+++ b/lib/elixir/test/elixir/kernel_test.exs
@@ -361,6 +361,16 @@ defmodule KernelTest do
       _ = a
     end
 
+    test "setting attribute with uppercase" do
+      message = "module attributes set via @ cannot start with an uppercase letter"
+
+      assert_raise ArgumentError, message, fn ->
+        defmodule UpcaseAttrSample do
+          @Upper
+        end
+      end
+    end
+
     test "in module body" do
       defmodule InSample do
         @foo [:a, :b]


edisonywh:
I have a question though @josevalim - is it safe to assume that `__aliases__` means capitalized letter?

lib/elixir/test/elixir/kernel_test.exs
```diff
@@ -361,6 +361,18 @@ defmodule KernelTest do
...
+    test "setting attribute with uppercase" do
+      message = "module attributes cannot start with an uppercase letter"
+
+      assert_raise ArgumentError, message, fn ->
+        defmodule InSample do
+          @Upper
+        end
+      end
+    after
+      purge(InSample)
```
josevalim: You don't need to purge it here because the module was never defined in this case. So you can remove the whole `after` bit.

lib/elixir/test/elixir/kernel_test.exs
```diff
@@ -361,6 +361,18 @@ defmodule KernelTest do
       _ = a
     end
 
+    test "setting attribute with uppercase" do
+      message = "module attributes cannot start with an uppercase letter"
+
+      assert_raise ArgumentError, message, fn ->
+        defmodule InSample do
```
josevalim:
```suggestion
        defmodule UpcaseAttrSample do
```

lib/elixir/lib/kernel.ex
```diff
@@ -2867,6 +2867,10 @@ defmodule Kernel do
   """
   defmacro @expr
 
+  defmacro @{:__aliases__, _meta, _args} do
+    raise ArgumentError, "module attributes cannot start with an uppercase letter"
```
```suggestion
    raise ArgumentError, "module attributes set via @ cannot start with an uppercase letter"
```

edisonywh:
Updated as per suggestion @josevalim

josevalim:

PR #8912, opened 2019-03-27
Make List.last/1 deal with improper lists and improve specs

eksperimental:
Now List.last/1 deals with improper lists same way as Kernel.tl/1 does.

Spec is more specific when list is empty.

### DIFF
--- a/lib/elixir/lib/list.ex
+++ b/lib/elixir/lib/list.ex
@@ -235,6 +235,8 @@ defmodule List do
   @doc """
   Returns the last element in `list` or `nil` if `list` is empty.
 
+  It accepts improper lists.
+
   ## Examples
 
       iex> List.last([])
@@ -246,11 +248,18 @@ defmodule List do
       iex> List.last([1, 2, 3])
       3
 
+      iex> List.last([1, 2, 3 | 4])
+      4
   """
+  @spec last([]) :: nil
   @spec last([elem]) :: nil | elem when elem: var
-  def last([]), do: nil
-  def last([head]), do: head
-  def last([_ | tail]), do: last(tail)
+  @spec last(nonempty_improper_list(any, tail)) :: tail when tail: var
+  def last(list) when is_list(list), do: last_guarded(list)
+
+  defp last_guarded([]), do: nil
+  defp last_guarded([head]), do: head
+  defp last_guarded([_ | tail]), do: last_guarded(tail)
+  defp last_guarded(improper_list_tail), do: improper_list_tail
 
   @doc """
   Receives a list of tuples and returns the first tuple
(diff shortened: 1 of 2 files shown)


lib/elixir/lib/list.ex
```diff
@@ -246,11 +248,18 @@ defmodule List do
       iex> List.last([1, 2, 3])
       3
 
+      iex> List.last([1, 2, 3 | 4])
+      4
```
whatyouhide: Please add an empty line after this and before the `"""`.

lib/elixir/lib/list.ex
```diff
@@ -246,11 +248,18 @@ defmodule List do
...
   """
+  @spec last([]) :: nil
   @spec last([elem]) :: nil | elem when elem: var
-  def last([]), do: nil
-  def last([head]), do: head
-  def last([_ | tail]), do: last(tail)
+  @spec last(nonempty_improper_list(any, tail)) :: tail when tail: var
+  def last(list) when is_list(list), do: last_guarded(list)
```
whatyouhide: Since we're using brackets in all clauses except one, do we really need this additional function with the guard? Couldn't we add `is_list` to the last clause and achieve the same?

lib/elixir/lib/list.ex
eksperimental: @whatyouhide this is a great coincidence. Do we remember I have fixed these exact thing in a PR in the past. Do you know which commit was that or PR? I have been looking for it for the past 20 minutes to no avail, and now you send this comment.

lib/elixir/lib/list.ex
eksperimental: the last clause is not a list, is an improper list tail. If we don't check with `is_list` in the main function, we cannot tell whether the user passed a non-list term from the tail of an improper list.

lib/elixir/lib/list.ex
whatyouhide: I don't recall, sorry!

lib/elixir/lib/list.ex
whatyouhide: Oh I see what you mean. Then this maybe:

```elixir
def last([]), do: nil
def last([head]), do: head
def last([_ | tail]) when is_list(tail), do: last(tail)
def last([_ | improper_list_tail]), do: improper_list_tail
```

? Not sure this would work but it would save us the additional function.

lib/elixir/lib/list.ex
eksperimental: well. I will use this comment for reference. thank you anyway

josevalim:
`last` may not mean anything if a list is improper. If we go by this definition, then the `last` of a proper list is always `[]`. So I don't believe we should merge this. Thanks!

lib/elixir/lib/list.ex
eksperimental: you will run `is_list/0` the length of the list minus 2.
with my implementation it is run once.

PR #8913, opened 2019-03-27
List.to_string/1 improve error message and docs

eksperimental:
* Make List.to_string/1 error msg non-ambiguous and corrects it mentioning empty lists are valid
* Add elements allowed in List.to_string/1 to @doc
* add tests for empty lists

### DIFF
--- a/lib/elixir/lib/list.ex
+++ b/lib/elixir/lib/list.ex
@@ -870,11 +870,11 @@ defmodule List do
         raise ArgumentError, """
         cannot convert the given list to a string.
 
-        To be converted to a string, a list must contain only:
+        To be converted to a string, a list must only contain one of the following:
 
           * strings
           * integers representing Unicode code points
-          * or a list containing one of these three elements
+          * a list containing one of these three elements
 
         Please check the given list or call inspect/1 to get the list representation, got:
 
--- a/lib/elixir/lib/list.ex
+++ b/lib/elixir/lib/list.ex
@@ -845,6 +845,13 @@ defmodule List do
   Converts a list of integers representing code points, lists or
   strings into a string.
 
+  To be converted to a string, a list must either be empty or only
+  contain the following elements:
+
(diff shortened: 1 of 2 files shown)


lib/elixir/lib/list.ex
```diff
@@ -872,9 +872,9 @@ defmodule List do
 
         To be converted to a string, a list must contain only:
 
-          * strings
-          * integers representing Unicode code points
-          * or a list containing one of these three elements
+          * strings;
+          * integers representing Unicode code points;
+          * and/or a list containing one of these three elements.
```
whatyouhide: Could we achieve the same result by removing "or" from what we had here before the change?

eksperimental: Nope. because that would be that it must contain all three items

eksperimental: We could remove "or" and in the header say "a list must contain either:"

whatyouhide: I'm not really sure how so, but then what if we make the first sentence "To be converted to a string, a list must only contain one of the following:"?

eksperimental: I like it. let's go with it.

eksperimental:
Ok. I see other issues. a list must be empty and it is valid, and it doesn't follow any of the three items that this error message mention.

lib/elixir/lib/list.ex
eksperimental: fixed,

eksperimental: actually it is wrong. "one of the following" is wrong. it can contain a string and a number

eksperimental: I'm changing it for "To be converted to a string, a list must either be empty or only contain the following elements:"

eksperimental:
I have created a second commit that improves the docs and the tests.

PR #8915, opened 2019-03-27
Improve specs for List.first/1 and List.last/1 when list is empty

eksperimental:


fertapric:
@eksperimental what do you think about doing the same changes in other similar functions like `first`?

eksperimental:
great idea. done!

josevalim:

PR #8916, opened 2019-03-27
List.foldl/1 and List.foldr/1 docs and specs for empty lists

eksperimental:


josevalim:
Thanks Eksperimental!

As mentioned in other cases, we use typespecs mostly for docs and this is
one of the cases where I think that, although the specs are more precise,
the readability of them are worse. :(
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

eksperimental:
as a last resort to make it more readable, 
`@spec foldl([], acc, (any, any -> any)) :: acc when initial_acc: var`

Your call

josevalim:
Let's leave them as is for now, thank you!

PR #8919, opened 2019-03-30
`Keyword.pop/3` must return all values associated with key

milmazz:
The description from `Keyword.pop/3` mentioned that it will return
**all** the values associated with `key` in the keyword list, but this
wasn't happening, at least when you have duplicate keys in the keyword
list `Keyword.pop/3` was returning just the first match, even one of
examples showed that behavior.

This is my attempt to fix that situation based on my understanding from
the initial description from `Keyword.pop/3`:

> Returns and removes all values associated with key in the keyword list.
>
> All duplicated keys are removed. See pop_first/3 for removing only the first
entry.

### DIFF
--- a/lib/elixir/lib/keyword.ex
+++ b/lib/elixir/lib/keyword.ex
@@ -960,17 +960,20 @@ defmodule Keyword do
       iex> Keyword.pop([a: 1], :b, 3)
       {3, [a: 1]}
       iex> Keyword.pop([a: 1, a: 2], :a)
-      {1, []}
+      {[1, 2], []}
 
   """
   @spec pop(t, key, value) :: {value, t}
   def pop(keywords, key, default \\ nil) when is_list(keywords) do
-    case fetch(keywords, key) do
-      {:ok, value} ->
+    case get_values(keywords, key) do
+      [] ->
+        {default, keywords}
+
+      [value | []] ->
         {value, delete(keywords, key)}
 
-      :error ->
-        {default, keywords}
+      values ->
+        {values, delete(keywords, key)}
     end
   end
 
--- a/lib/elixir/lib/keyword.ex
+++ b/lib/elixir/lib/keyword.ex
@@ -946,10 +946,15 @@ defmodule Keyword do
   end
 
   @doc """
-  Returns and removes all values associated with `key` in the keyword list.
+  Returns the value for `key` and removes all associated entries in the keyword list.
 
-  All duplicated keys are removed. See `pop_first/3` for
-  removing only the first entry.
+  It returns a tuple where the first value for `key` is the first element and the
+  second element is a keyword list with all values associated with `key` are removed.
+  If the `key` is not present in the keyword list, `{default, keyword_list}` is
+  returned.
+
+  If you don't want to remove all the values associated with key use `pop_first/3`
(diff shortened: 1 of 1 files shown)


josevalim:
Good find. We should only return the first though and delete everything
else. :) otherwise it can be a breaking change as some may now get a list
where they wouldn’t before.

so pretty much how it was done before, but with an extra delete call.
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

milmazz:
@josevalim So, are you implying that the behavior as is now implemented is correct and we should just adjust the docs? In that case, what do you think about this description:

> Returns and removes all values associated with `key` in the keyword list.
> 
>  If `key` is present in the keyword list with `value`,
>  `{value, new_keyword_list}` is returned where `new_keyword_list` is the result of
>  removing **all** the keys (a keyword list can have duplicated keys). If the
>  `key` is not present in the keyword list, `{default, keyword_list}` is
>  returned. See `pop_first/3` for removing only the first entry.

josevalim:
Oh, I see. Yes, the current behaviour is correct. I have pushed your docs suggestion to your branch with some amendments, what do you think?

milmazz:
@josevalim LGTM. Thanks!

lib/elixir/lib/keyword.ex
```diff
@@ -946,10 +946,15 @@ defmodule Keyword do
   end
 
   @doc """
-  Returns and removes all values associated with `key` in the keyword list.
+  Returns the value for `key` and removes all associated entries in the keyword list.
```
fertapric:
```suggestion
  Returns the first value for `key` and removes all associated entries in the keyword list.
```

lib/elixir/lib/keyword.ex
```diff
@@ -946,10 +946,15 @@ defmodule Keyword do
...
 
-  All duplicated keys are removed. See `pop_first/3` for
-  removing only the first entry.
+  It returns a tuple where the first value for `key` is the first element and the
```
fertapric:
```suggestion
  It returns a tuple where the first element is the first value for `key` and the
```

lib/elixir/lib/keyword.ex
```diff
@@ -946,10 +946,15 @@ defmodule Keyword do
...
+  second element is a keyword list with all values associated with `key` are removed.
```
fertapric: The combination of "with + are" feels weird to me.

```suggestion
  second element is a keyword list with all entries associated with `key` removed.
```

Or, as an alternative, "a keyword list in which all entried associated with `key` are removed."

lib/elixir/lib/keyword.ex
```diff
@@ -946,10 +946,15 @@ defmodule Keyword do
...
+  If the `key` is not present in the keyword list, `{default, keyword_list}` is
+  returned.
+
+  If you don't want to remove all the values associated with key use `pop_first/3`
```
fertapric: Nitpick

```suggestion
  If you don't want to remove all the values associated with `key` use `pop_first/3`
```

fertapric:
Thanks @milmazz! 💜

PR #8920, opened 2019-03-30
Make Version.Requirement public

fertapric:
The `Version.Requirement` struct is mentioned in several places of the `Version` documentation, used as an argument of `Version.compile_requirement/1` and `Version.match?/3`, and returned as a result of `Version.parse_requirement/1` and `Version.parse_requirement!/1`.

In addition, ExDoc is currently linking to modules even if they are marked with `@moduledoc false` (that issue is being tracked in https://github.com/elixir-lang/ex_doc/issues/951), and produces "PAGE NOT FOUND" when clicking on the [`Version.Requirement` links.](https://hexdocs.pm/elixir/Version.html#parse_requirement/1)

### DIFF
--- a/lib/elixir/lib/version.ex
+++ b/lib/elixir/lib/version.ex
@@ -110,11 +110,15 @@ defmodule Version do
     @moduledoc """
     A struct that holds version requirement information.
 
+    The struct fields are private and should not be accessed.
+
     See the "Requirements" section in the `Version` module
     for more information.
     """
 
     defstruct [:source, :matchspec, :compiled]
+
+    @typedoc false
     @type t :: %__MODULE__{source: String.t(), matchspec: :ets.match_spec(), compiled: boolean}
   end
 
--- a/lib/elixir/lib/version.ex
+++ b/lib/elixir/lib/version.ex
@@ -107,7 +107,13 @@ defmodule Version do
   @type t :: %__MODULE__{major: major, minor: minor, patch: patch, pre: pre, build: build}
 
   defmodule Requirement do
-    @moduledoc false
+    @moduledoc """
+    A struct that holds version requirement information.
+
+    See the "Requirements" section in the `Version` module
+    for more information.
+    """
+
     defstruct [:source, :matchspec, :compiled]
     @type t :: %__MODULE__{source: String.t(), matchspec: :ets.match_spec(), compiled: boolean}
   end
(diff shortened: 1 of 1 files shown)


josevalim:
Good catch. I think we need to say though that all struct fields are private and `@doc false` `defstruct` just in case. Thoughts?

lib/elixir/lib/version.ex
```diff
@@ -107,8 +107,18 @@ defmodule Version do
...
+
+    The struct fields are private and should not be accessed.
+
+    See the "Requirements" section in the `Version` module
+    for more information.
+    """
+
     defstruct [:source, :matchspec, :compiled]
+
+    @typedoc false
```
fertapric: Links to `Version.Requirement.t` will open the module documentation. The anchor `#t:t/0` won't be find, but I think that's ok.

josevalim: We need to either do that or make the type as opaque, but that will lead to dialyzer errors.

fertapric: Other alternative is to make the structure public, and maybe hide the inner types?

```elixir
@typep source :: String.t()
@typep compiled :: boolean
@typep matchspec :: :ets.match_spec()

@type t :: %__MODULE__{source: source, matchspec: matchspec, compiled: compiled}
```

whatyouhide: @fertapric if we make it public, I'm afraid we're still allowing people to reply on the fields in the struct.

josevalim: I would go with `@opaque t` and let dialyzer warn (as we do in other places).

eksperimental:
As far as I know, exdoc ia not linking to modules with doc false

josevalim:
It links, we don’t check today. :)
-- 


eksperimental:
I will look into it in that PR THEN

umamaistempo:
In the original message of the PR it's said:
> used as an argument of Version.compile_requirement/1 and Version.match?/3

Considering the type now is not exported won't this potentially produce a dialyzer warning ?

fishcakez:
Yes, the Requirement struct should be created in the Requirement module, not Version module to fix the warnings about breaking opaque contract. This function could be `@doc false`.

umamaistempo:
I can submit a PR for that

fertapric:
Thanks @umamaistempo! 💜

PR #8924, opened 2019-03-31
Add a better error message when overriding an umbrella dependency

AnilRedshift:
I wasn't able to get the tests working for this - If someone with a better understanding of how to set up a proper umbrella fixture wants to add a test, this would be great.

I did verify manually that the changes work as expected:

### Output in an umbrella project

> ~/scratch/my_umbrella (master)$ mix deps.get
> Dependencies have diverged:
> * a (apps/a)
>   the dependency a in mix.exs is overriding a child dependency:
> 
>   > In mix.exs:
>     {:a, [path: "apps/a", from_umbrella: true, env: :dev]}
> 
>   > In apps/b/mix.exs:
>     {:a, [path: "../a", in_umbrella: true, env: :test, override: true]}
> 
> It is not possible to override a dependency within an umbrella app. Either ensure the above dependencies match, or move them out of the umbrella
> ** (Mix) Can't continue due to errors on dependencies

### DIFF
--- a/lib/mix/lib/mix/dep.ex
+++ b/lib/mix/lib/mix/dep.ex
@@ -368,10 +368,17 @@ defmodule Mix.Dep do
       "\n  Ensure they match or specify one of the above in your deps and set \"override: true\""
   end
 
-  def format_status(%Mix.Dep{app: app, status: {:overridden, other}} = dep) do
-    "the dependency #{app} in #{Path.relative_to_cwd(dep.from)} is overriding a child dependency:\n" <>
-      "#{dep_status(dep)}#{dep_status(other)}" <>
-      "\n  Ensure they match or specify one of the above in your deps and set \"override: true\""
+  def format_status(%Mix.Dep{app: app, opts: opts, status: {:overridden, %Mix.Dep{opts: other_opts} = other}} = dep) do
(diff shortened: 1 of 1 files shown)


lib/mix/lib/mix/dep.ex
```diff
@@ -368,10 +368,17 @@ defmodule Mix.Dep do
...
 
-  def format_status(%Mix.Dep{app: app, status: {:overridden, other}} = dep) do
-    "the dependency #{app} in #{Path.relative_to_cwd(dep.from)} is overriding a child dependency:\n" <>
-      "#{dep_status(dep)}#{dep_status(other)}" <>
-      "\n  Ensure they match or specify one of the above in your deps and set \"override: true\""
+  def format_status(%Mix.Dep{app: app, opts: opts, status: {:overridden, %Mix.Dep{opts: other_opts} = other}} = dep) do
+    problem = "the dependency #{app} in #{Path.relative_to_cwd(dep.from)} is overriding a child dependency:\n" <>
+    "#{dep_status(dep)}#{dep_status(other)}"
+
+    resolution = case Keyword.get(opts, :in_umbrella) || Keyword.get(other_opts, :in_umbrella) do
```
josevalim: You need to check that one of them is from_umbrella and the other is in_umbrella.

Also, I think this error happens when they *diverge* not when they override, so it should be in another branch of format_status.

Finally, please submit a test too. :) Thank you!

PR #8925, opened 2019-04-01
add to mix elixirc_options: [ export_all: true ]

mogorman:
this is very useful for testing private functions. follows same logic as the already present debug_opts.


josevalim:
Can you please submit a proposal to the mailing list. `export_all` is discouraged in Erlang and we would need very strong reasons to add this behaviour to Elixir. Thank you.

mogorman:
https://groups.google.com/forum/#!topic/elixir-lang-core/Z_CZ8_GOAQ8 notes here for future searchers

josevalim:
Thank you!

PR #8928, opened 2019-04-02
Add functions Map.with/2 and Map.without/2

rutaka-n:
Hi.
There are functions with/2 and without/2 in erlang stdlib, but there are not in Elixir, so I add wrappers for them.
It would be great to use them in Elixir style, for example: use them with pipe operator.

What do you think?
I am looking forward feedback.
Have a nice day!

### DIFF
--- a/lib/elixir/lib/map.ex
+++ b/lib/elixir/lib/map.ex
@@ -897,4 +897,21 @@ defmodule Map do
   def size(map) do
     map_size(map)
   end
+
+  @doc """
+  Returns a new map without keys through list and their associated values from map.
+
+  Any key in list that does not exist in map is ignored
+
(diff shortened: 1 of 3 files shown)


radar:
`Map.with/2` sounds like `Map.take/2`: https://hexdocs.pm/elixir/Map.html#take/2

rutaka-n:
I'm sorry, I miss it function, maybe because I more familiar with erlang, then elixir.
I'll remove this commit.

And what do you think about Map.without/2 ?

josevalim:
There are Map.take/2 and Map.drop/2.
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

PR #8929, opened 2019-04-02
Make Version.Requirement opaque

umamaistempo:
Fix dialyzer warning introduced on #8920

### DIFF
--- a/lib/elixir/lib/version.ex
+++ b/lib/elixir/lib/version.ex
@@ -133,7 +133,7 @@ defmodule Version do
     @doc false
     def match?(%__MODULE__{matchspec: spec, compiled: true}, matchable_pattern) do
       matches = :ets.match_spec_run([matchable_pattern], spec)
-      Enum.any?(matches)
+      matches != []
     end
 
     def match?(%__MODULE__{matchspec: spec, compiled: false}, matchable_pattern) do
--- a/lib/elixir/lib/version.ex
+++ b/lib/elixir/lib/version.ex
@@ -119,6 +119,26 @@ defmodule Version do
     defstruct [:source, :matchspec, :compiled]
 
     @opaque t :: %__MODULE__{source: String.t(), matchspec: :ets.match_spec(), compiled: boolean}
+
+    @doc false
+    def new(source, spec) do
+      %__MODULE__{source: source, matchspec: spec, compiled: false}
+    end
+
+    @doc false
+    def compile(%__MODULE__{matchspec: spec} = req) do
+      %{req | matchspec: :ets.match_spec_compile(spec), compiled: true}
+    end
+
+    @doc false
+    def compiled?(%__MODULE__{compiled: compiled?}) do
+      compiled?
+    end
+
+    @doc false
+    def matchspec(%__MODULE__{matchspec: spec}) do
+      spec
+    end
   end
 
   defmodule InvalidRequirementError do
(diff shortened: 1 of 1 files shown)


umamaistempo:
wait a moment, we're no longer running any check on PRs ?

lib/elixir/lib/version.ex
```diff
@@ -192,15 +212,16 @@ defmodule Version do
...
-  end
-
-  def match?(version, %Requirement{matchspec: spec, compiled: true}, opts) do
+  def match?(version, requirement, opts) do
+    spec = Requirement.matchspec(spec)
     allow_pre = Keyword.get(opts, :allow_pre, true)
-    :ets.match_spec_run([to_matchable(version, allow_pre)], spec) != []
+    if Requirement.compiled?(requirement) do
+      matches = :ets.match_spec_run([to_matchable(version, allow_pre)], spec)
+      matches != []
```
umamaistempo: Consideration: this is this way because it was based on the original implementation. Particularly I'd prefer to use `Enum.any?(matches)` because it sounds nicer :>

lib/elixir/lib/version.ex
```diff
@@ -192,15 +212,16 @@ defmodule Version do
...
+    else
+      {:ok, result} = :ets.test_ms(to_matchable(version, allow_pre), spec)
+      result != false
+    end
```
umamaistempo: Maybe move the this whole function to `Version.Requirement` as it's mostly based on the requirement spec which belongs to that module :thinking:

josevalim: Good call. I think we can move the whole function in there as that would allow us to cut the amount of intermediate functions.

umamaistempo: Going to do that (right now fixing wrong var name and formatting) :)

umamaistempo: Done

lib/elixir/lib/version.ex
```diff
@@ -119,6 +119,27 @@ defmodule Version do
...
+
+    @doc false
+    def compile(%__MODULE__{matchspec: spec} = req) do
+      %{req | matchspec: :ets.match_spec_compile(spec), compiled: true}
+    end
+
+    @doc false
+    def match?(%__MODULE__{matchspec: spec, compiled: true}, matchable_pattern) do
+      matches = :ets.match_spec_run([matchable_pattern], spec)
+      Enum.any?(matches)
```
josevalim:
```suggestion
      matches != []
```

Since the result can only be lists, I prefer a `!= []` instead of Enum which can be polymorphic. :)

umamaistempo: :+1:

I did this because of how it sounds "Is there any matches?" instead of "is not an empty list". But I can see the benefits of sticking to simple

lib/elixir/lib/version.ex
```diff
@@ -119,6 +119,27 @@ defmodule Version do
     defstruct [:source, :matchspec, :compiled]
 
     @opaque t :: %__MODULE__{source: String.t(), matchspec: :ets.match_spec(), compiled: boolean}
```
fertapric: I think the correct typespec for `:matchspec` is `:ets.match_spec() | :ets.comp_match_spec()`

lib/elixir/lib/version.ex
```diff
@@ -119,6 +119,27 @@ defmodule Version do
...
+
+    @doc false
+    def new(source, spec) do
```
fertapric: I think we need to include `@spec` here to avoid warnings too:

```elixir
@spec new(String.t(), :ets.match_spec()) :: t
@spec compile(t) :: t
@spec match?(t, tuple) :: boolean # or expanding "tuple" to the "{non_neg_integer, non_neg_integer, non_neg_integer, [String.t() | non_neg_integer], boolean}"
```

or expanding `tuple` in the last one to `{non_neg_integer, non_neg_integer, non_neg_integer, [String.t() | non_neg_integer], boolean}`

umamaistempo: Those won't cause warning because of the success typing :)

But I can surely do that (and would like to)

lib/elixir/lib/version.ex
umamaistempo: Good catch :+1: will update

lib/elixir/lib/version.ex
umamaistempo: I'll simplify to tuple because otherwise I'd like that 5-tuple to be a type defined and exported at Version.

fertapric:
Thanks @umamaistempo! 💜

umamaistempo:
Note about last commit: That's awkward, I have run `mix format` and it left whitespaces at the end of the line :thinking: was this some missconfiguration ? (yep, clean install of IDE, so didn't configure it to strip those things)

PR #8932, opened 2019-04-04
No longer generate config for mix new

### LINKED ISSUE #8811: Config from `mix run --config` invocations doesn't get deeply merged
### Environment

```
$ elixir --version
Erlang/OTP 21 [erts-10.2.2] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [hipe]

Elixir 1.8.0 (compiled with Erlang/OTP 20)
```

### Current behavior

When I put, e.g.:

```
config :my_app, MyApp.Endpoint,
  http: [port: 7435]
```

into a `some_config.exs` file and then `mix run --config some_config.exs`, I'll get:

```
IO.inspect Application.get_env(:my_app, MyApp.Endpoint)

# [http: [port: 7435]]
```

despite having entries in the "default" config files loaded in a rather standard manner (i.e. `config/config.exs` doing `import_config config/dev.exs` etc.).

### Expected behavior

I'd expect the config from `some_config.exs` be deeply merged as are the `config/config.exs` and `config/dev.exs` loaded by default.

That is, if without the override in `--config some_config.exs` I have:

```
IO.inspect Application.get_env(:my_app, MyApp.Endpoint)

# [
#   secret_key_base: ...
# <snip>,
#   http: [port: 7434],
# <snip>
# ]
```

**with the override `--config some_config.exs`**, I'd expect to have 

```
# [
#   secret_key_base: ...
# <snip>,
#   http: [port: 7435],
# <snip>
# ]
```

which would effectively be `config/config.exs` with `config/dev.exs` merged in with `some_config.exs` merged in, deeply

### LINKED ISSUE #8815: Do not generate config/config.exs for mix new
`mix new` is mostly used for libraries and libraries should be actively discouraged of defining (because it is global) or using entries (as it only applies to the library) in `config`.

josevalim:
The fact we explicitly told developers to configure their
application in the config file and access it with
`Application.get_env/2`, probably led many to rely on
the application environment when better options are
available.

Furthermore, `mix new` is mostly used for libraries,
where config is even less important, as configuration
from libraries is not transitive.

Developers that need the config, can simply add it back
by calling:

    echo "use Mix.Config" > config/config.exs

This commit removes config for new apps and deprecate
the `--config` flag, which cannot perform deep merges.
Note we do keep the config for umbrellas, as umbrellas
are about applications and not libraries (plus we need
the umbrella children to share and point to an existing
config).

Closes #8815.
Closes #8811.

### DIFF
--- a/lib/mix/test/mix/tasks/run_test.exs
+++ b/lib/mix/test/mix/tasks/run_test.exs
@@ -11,17 +11,6 @@ defmodule Mix.Tasks.RunTest do
     Mix.Project.push(MixTest.Case.Sample)
   end
 
-  test "loads configuration", context do
-    in_tmp(context.test, fn ->
-      config = fixture_path("configs/good_config.exs")
-      expr = "IO.puts Application.get_env(:my_app, :key)"
-      output = capture_io(fn -> Mix.Task.run("run", ["--config", config, "--eval", expr]) end)
-      assert output == "value\n"
-    end)
-  after
-    Application.delete_env(:my_app, :key)
-  end
-
   test "run requires files before evaling commands", context do
     git_repo = fixture_path("git_repo/lib/git_repo.ex")
 
--- a/lib/mix/test/mix/tasks/run_test.exs
+++ b/lib/mix/test/mix/tasks/run_test.exs
@@ -11,6 +11,23 @@ defmodule Mix.Tasks.RunTest do
     Mix.Project.push(MixTest.Case.Sample)
   end
 
+  test "loads configuration", context do
+    in_tmp(context.test, fn ->
+      config = fixture_path("configs/good_config.exs")
+      expr = "IO.puts Application.get_env(:my_app, :key)"
+
+      output =
+        capture_io(:stderr, fn ->
+          assert capture_io(fn -> Mix.Task.run("run", ["--config", config, "--eval", expr]) end) ==
+                   "value\n"
+        end)
+
+      assert output =~ "deprecated"
+    end)
+  after
+    Application.delete_env(:my_app, :key)
+  end
+
   test "run requires files before evaling commands", context do
     git_repo = fixture_path("git_repo/lib/git_repo.ex")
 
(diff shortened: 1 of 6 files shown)


lib/mix/test/mix/tasks/run_test.exs
```diff
@@ -11,17 +11,6 @@ defmodule Mix.Tasks.RunTest do
     Mix.Project.push(MixTest.Case.Sample)
   end
 
-  test "loads configuration", context do
```
fertapric: What do you think about keeping this test and add a `# TODO: Remove on v2.0.` instead?

josevalim: Done.

perrycate:
Pardon my ignorance, but what are the aforementioned "better options"? Based on [this guide](https://elixir-lang.org/getting-started/mix-otp/config-and-releases.html#custom-configuration) it seems the alternative to using a config.exs file is to add config parameters to env in application/0 in mix.exs.

There clearly are differences between the two, as I am unable to set module attributes using Application.fetch_env!/2 using the latter config option. I'm having trouble groking what those differences are, though, and what the best practices for configuration in general are.

(If it makes a difference, I have been learning elixir by working through the book "Programming Elixir >= 1.6 by Dave Thomas. It seems it was written before this decision was made, as the example I'm working through assumes config/config.exs exists and is the best place to put application configuration. Is there another resource that would be better for a new user such as myself to consult?)

vans163:
> Pardon my ignorance, but what are the aforementioned "better options"? Based on [this guide](https://elixir-lang.org/getting-started/mix-otp/config-and-releases.html#custom-configuration) it seems the alternative to using a config.exs file is to add config parameters to env in application/0 in mix.exs.
> 
> There clearly are differences between the two, as I am unable to set module attributes using Application.fetch_env!/2 using the latter config option and using `iex -S mix`. I'm having trouble groking what those differences are, though, and what the best practices for configuration in general are.
> 
> (If it makes a difference, I have been learning elixir by working through the book "Programming Elixir >= 1.6 by Dave Thomas. It seems it was written before this decision was made, as the example I'm working through assumes config/config.exs exists and is the best place to put application configuration. Is there another resource that would be better for a new user such as myself to consult?)

Create a .env file and source it or pass env vars directly on commandline.

Issue #8818, opened 2019-02-22
mix: follow XDG base dir specification


NobbZ:
### Environment

* Elixir & Erlang/OTP versions (elixir --version): currently 1.7 and 1.8 (depends on the host) but affects earlier versions as well.
* Operating system: Linux, but maybe affects Mac OS as well

### Current behavior

Mix saves it configuration, downloaded data, archives, etc to `~/.mix`

### Expected behavior

It should follow the [XDG Base Dir spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) and use appropriate folders from the users configuration to save aforementioned stuff.

This makes it easier for a user to wipe caches, backup config or archives, etc.

This is issue is explicitely about global cache/data/config, not their per-project equivalents.

josevalim:
Can you please send a proposal to the mailing list? In particular, I would
love to hear why we should accept this particular spec, given that it is
not 1.0 yet, how it compares to other specs, and who is already adopting
it. Thank you.
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

NobbZ:
https://groups.google.com/forum/#!topic/elixir-lang-core/7nyDfkgRsI0

maennchen:
Is the idea here to only change the `Mix.Utils.mix_home/1`to consider the `XDG_DATA_HOME` or would a clean implementation also change the location of the following files / folders (those are the ones I found, there may be more):

- `archives` - `${XDG_DATA_HOME:-/usr/local/share/}mix/archives`
- `config.exs` - `${XDG_CONFIG_HOME:-$HOME/.config/}mix/config.exs`
- `rebar` - `${XDG_DATA_DIRS:-/usr/local/share/}mix/bin/rebar`
- `rebar3` - `${XDG_DATA_DIRS:-/usr/local/share/}mix/bin/rebar3`
- `escripts` - `${XDG_DATA_HOME:-/usr/local/share/}mix/bin/escripts`

michalmuskala:
When it comes to implementation, the Erlang's [filename:basedir/3](http://erlang.org/doc/man/filename.html#basedir-3) function might be of interest, that generally does "the right thing" on each OS.

maennchen:
If we'd use the directories as specified, this would also change existing file locations for people that do neither specify `XDG_*` nor `MIX_*` variables.

Would that be considered a breaking change and has to wait for v2?

[...]

PR #8933, opened 2019-04-04
mix: follow XDG base dir specification

maennchen:
Implementation of #8818 

The documentation is not as good as I'd expect. Since I'm not a native speaker some help would be appreciated.

### DIFF
--- a/lib/mix/lib/mix/utils.ex
+++ b/lib/mix/lib/mix/utils.ex
@@ -4,8 +4,8 @@ defmodule Mix.Utils do
   @doc """
   Gets the Mix home.
 
-  It defaults to `~/.mix` unless the `MIX_HOME`
-  environment variable is set.
+  It uses the the locations `MIX_HOME`, `XDG_DATA_HOME/mix`,
+  `~/.mix` with decreasing priority.
 
   Developers should only store entries in the
   `MIX_HOME` directory which are guaranteed to
@@ -15,7 +15,29 @@ defmodule Mix.Utils do
   stored there.
   """
   def mix_home do
-    System.get_env("MIX_HOME") || Path.expand("~/.mix")
+    case {System.get_env("MIX_HOME"), System.get_env("XDG_DATA_HOME")} do
+      {directory, _} when is_binary(directory) -> directory
+      {nil, directory} when is_binary(directory) -> :filename.basedir(:user_data, "mix")
+      {nil, nil} -> Path.expand("~/.mix")
+    end
+  end
+
+  @doc """
+  Gets possible location of global Mix configuration.
+
+  Possible locations:
+
+   * `~/.mix`
+   * `MIX_HOME`
+   * `XDG_CONFIG_HOME/mix`
+
+  """
+  def mix_config_home do
+    case {System.get_env("MIX_HOME"), System.get_env("XDG_CONFIG_HOME")} do
+      {directory, _} when is_binary(directory) -> directory
+      {nil, directory} when is_binary(directory) -> :filename.basedir(:user_config, "mix")
+      {nil, nil} -> Path.expand("~/.mix")
+    end
   end
 
   @doc """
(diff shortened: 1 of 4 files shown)


lib/mix/lib/mix/utils.ex
```diff
@@ -15,7 +15,31 @@ defmodule Mix.Utils do
...
+  Gets possible locations of global Mix configuration.
+
+  Possible locations:
+
+   * `~/.mix`
+   * `MIX_HOME`
+   * `XDG_DATA_HOME/mix`
+
+  """
+  def mix_config_homes do
```
josevalim: I think this can be the same  as `mix_home`, i.e. we will just lookup at one of the directories, depending the variable is set. In fact, this change is even backwards incompatible as it can  load multiple configs thaat wouldn't be loaded today.

lib/mix/test/mix/utils_test.exs
```diff
@@ -7,6 +7,36 @@ defmodule Mix.UtilsTest do
...
+    System.delete_env("XDG_CONFIG_HOME")
+
+    # Reset Env Variables
+    on_exit(fn ->
+      case mix_home do
+        nil -> System.delete_env("MIX_HOME")
+        _ -> System.put_env("MIX_HOME", mix_home)
+      end
+
+      case xdg_data_home do
```
josevalim: No need to revert this, we know from the test_helper they should be nil.

lib/mix/test/mix/utils_test.exs
```diff
@@ -7,6 +7,36 @@ defmodule Mix.UtilsTest do
...
+        nil -> System.delete_env("XDG_DATA_HOME")
+        _ -> System.put_env("XDG_DATA_HOME", xdg_data_home)
+      end
+
+      case xdg_config_home do
```
josevalim: No need to revert this, we know from the test_helper they should be nil.

lib/mix/test/mix/utils_test.exs
```diff
@@ -7,6 +7,36 @@ defmodule Mix.UtilsTest do
...
+
+    # Clear all variables to get a reproducible test
+    System.delete_env("MIX_HOME")
+    System.delete_env("XDG_DATA_HOME")
+    System.delete_env("XDG_CONFIG_HOME")
+
+    # Reset Env Variables
+    on_exit(fn ->
+      case mix_home do
+        nil -> System.delete_env("MIX_HOME")
```
josevalim: It should never be nil from the test helper.

maennchen:
@josevalim I did some fixes. Is the general idea that a test cleans up the env variables it set itself or should I leave the `on_exit` as is?

josevalim:
If used once, I do in the same test. Otherwise I move to on_exit on setup.
-- 


maennchen:
@josevalim Ok, then I think all comments are adressed :)

lib/mix/lib/mix/utils.ex
```diff
@@ -15,7 +15,29 @@ defmodule Mix.Utils do
   stored there.
   """
   def mix_home do
-    System.get_env("MIX_HOME") || Path.expand("~/.mix")
+    case {System.get_env("MIX_HOME"), System.get_env("XDG_DATA_HOME")} do
+      {directory, _} when is_binary(directory) -> directory
+      {nil, directory} when is_binary(directory) -> :filename.basedir(:user_data, "mix")
```
josevalim:
```suggestion
      {nil, directory} when is_binary(directory) -> Path.join(directory, "mix")
```

lib/mix/lib/mix/utils.ex
```diff
@@ -15,7 +15,29 @@ defmodule Mix.Utils do
...
+
+   * `~/.mix`
+   * `MIX_HOME`
+   * `XDG_CONFIG_HOME/mix`
+
+  """
+  def mix_config_home do
+    case {System.get_env("MIX_HOME"), System.get_env("XDG_CONFIG_HOME")} do
+      {directory, _} when is_binary(directory) -> directory
+      {nil, directory} when is_binary(directory) -> :filename.basedir(:user_config, "mix")
```
josevalim:
```suggestion
      {nil, directory} when is_binary(directory) -> Path.join(directory, "mix")
```

josevalim:

PR #8934, opened 2019-04-04
Fix doc example for Protocol.derive/3

dsdshcym:
- Sample was not a defined Protocol in this example
- run this example in an iex session would cause `ArgumentError`

    ** (ArgumentError) Sample is not available

This PR also made some small changes to make the example more consistent and intuitive IMO.

A question: do we need to turn this example into a doctest? If so, I can also do that.

### DIFF
--- a/lib/elixir/lib/protocol.ex
+++ b/lib/elixir/lib/protocol.ex
@@ -351,7 +351,7 @@ defmodule Protocol do
   ## Examples
 
       defprotocol Derivable do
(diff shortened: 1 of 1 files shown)


josevalim:
> A question: do we need to turn this example into a doctest? If so, I can also do that.

We generally avoid doctests with side-effects. :)

fertapric:
Thanks @dsdshcym! 💜

Issue #8936, opened 2019-04-04
Error in Enum.max/2 documentation


Maultasche:
As of Elixir 1.8.1, the documentation for Enum.max/2 is incorrect. It states that "IIf multiple elements are considered maximal, the first one that was found is returned." This is incorrect. The *last* maximal element is returned.

Here's an example from iex.

iex> max(1, 1.0)
1
iex> max(1.0, 1)
1.0
iex> Enum.max([1, 1.0])
1.0
iex> Enum.max([1.0, 1])
1

Maultasche:
Working on a PR to fix this issue.

Maultasche:
This has been fixed in pull request #8937

whatyouhide:
Closing in favour of the PR.

Issue #8935, opened 2019-04-04
Error in Enum.min/2 documentation


Maultasche:
As of Elixir 1.8.1, the documentation for Enum.min/2 is incorrect. It states that "If multiple elements are considered minimal, the first one that was found is returned." This is incorrect. The *last* minimal element is returned.

Here's an example from iex.

iex> min(1, 1.0)
1
iex> min(1.0, 1)  
1.0
iex> Enum.min([1, 1.0])
1.0
iex> Enum.min([1.0, 1])  
1

josevalim:
Yes, you are correct. Good catch. Can you  please submit a PR that fixes the docs? Thank you.

Maultasche:
Yes, I'm working on that now. :)

Maultasche:
This has been fixed in pull request #8937

whatyouhide:
Closing in favour of the PR.

PR #8937, opened 2019-04-04
Corrected the docs on Enum.min/2 and Enum.max/2

Maultasche:
This is a fix for issues #8936 and #8935, which involve a minor error in the documentation of Enum.min/2 and Enum.max/2.

### DIFF
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -1457,7 +1457,7 @@ defmodule Enum do
   Returns the maximal element in the `enumerable` according
   to Erlang's term ordering.
 
-  If multiple elements are considered maximal, the first one that was found
+  If multiple elements are considered maximal, the last one that was found
   is returned.
 
   Calls the provided `empty_fallback` function and returns its value if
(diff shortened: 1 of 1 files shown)


eksperimental:
Is there a test that verifies this?

eksperimental:
Also doctests exemplifying this would be welcome

josevalim:
Thanks for the PR @Maultasche!

I took a deeper look at the code and it seems we document to return the first in other functions, such as min_by  and max_by, and those are  behaving accordingly.  Therefore, instead of having functions behaving differently, I decided to fix those two.

Sorry for misleading you, I asked you to send a PR for the docs but in this case the proper fix  is in the code. I have pushed a fix on bcac90e3a64fd85dfcacd7807b84b852a0e556dd. In any case, thanks! 

Maultasche:
Hi @josevalim. Thanks for following up with that comment! I completely agree that it is even better to make sure that min and max are consistent with min_by and max_by. I took a look at the changes you made and got to learn a bit more about how the Enum module works and what the tests look like.

I appreciate the great contribution experience, and I plan to make some more contributions. As I've been learning Elixir and reading over the docs, I've made notes of a few places where I could make some small documentation improvements, so you may see some more PRs from me in the future.

PR #8941, opened 2019-04-06
Added documentation to slice/3 indicating that start_index can be negative

Maultasche:
There was nothing in the Enum.slice/3 documentation indicating that start_index can be a negative number, so I updated the documentation to mention that fact, using similar language to the Enum.at/2 documentation. I also added an iex example to the docs as well.


whatyouhide:
Thanks @Maultasche! 💟

PR #8942, opened 2019-04-07
Fix Kernel.in/2

eksperimental:
Some big improvements to the in/2 operator.

1)  Optimize code when context is nil. Previously it was optimized as if for guards,
when it was not needed, ending up it code that was unnecessarily verbose and unoptimized.

previously:

    iex> quote do 1 in [1, 2, 3] end |> Macro.expand(__ENV__) |> Macro.to_string()
    "(\n  var = 1\n  :erlang.orelse(:erlang.\"=:=\"(var, 2), :erlang.orelse(:erlang.\"=:=\"(var, 3), :erlang.\"=:=\"(var, 1)))\n)"

now:

    iex>  quote do 1 in [1, 2, 3] end |> Macro.expand(__ENV__) |> Macro.to_string()
    "(\n  var = 1\n  :lists.member(var, [1, 2, 3])\n)"

2) Keep order of elements in Kernel.in/2:
The order of elements was not being kept (not doing as it was said in the docs).
Now it is guaranteed that the elements are compared in the same order they were given.

previously:

    iex> quote do 1 in [1, 2, 3] end |> Macro.expand(Map.put(__ENV__, :context, :guard)) |> Macro.to_string()
    ":erlang.orelse(:erlang.\"=:=\"(1, 2), :erlang.orelse(:erlang.\"=:=\"(1, 3), :erlang.\"=:=\"(1, 1)))"

now:

    iex>  quote do 1 in [1, 2, 3] end |> Macro.expand(Map.put(__ENV__, :context, :guard)) |> Macro.to_string()
    ":erlang.orelse(:erlang.orelse(:erlang.\"=:=\"(1, 1), :erlang.\"=:=\"(1, 2)), :erlang.\"=:=\"(1, 3))"

A special case is optimized when using the cons operator, and the last cell cannot be determined at compile time, and the order is still maintained:

    iex> quote do :x in [1, 2 | some_call.([4, 3])] end |> Macro.expand(__ENV__) |> Macro.to_string()
    "(\n  {arg0} = {some_call.([4, 3])}\n  (\n    var = :x\n    :erlang.orelse(:lists.member(var, [1, 2]), :lists.member(var, arg0))\n  )\n)"

3) Error message improved from: invalid args for operator "in"...
to: invalid right argument for operator "in"...

4) Multiple tests added to check for generated code when using guards and when not

### DIFF
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -3543,17 +3543,28 @@ defmodule Kernel do
       [] when not in_module? ->
         false
 
-      [head | tail] = list when not in_module? ->
-        in_var(in_module?, left, &in_list(&1, head, tail, expand, list, in_module?))
+      [] ->
+        quote(do: :lists.member(unquote(left), []))
+
+      [head | tail] when not in_module? ->
+        in_var(
+          in_module?,
+          left,
+          &in_list(&1, head, tail, expand, right, in_module?, true, [])
+        )
 
       [_ | _] = list when in_module? ->
         case ensure_evaled(list, {0, []}, expand) do
           {[head | tail], {_, []}} ->
-            in_var(in_module?, left, &in_list(&1, head, tail, expand, list, in_module?))
+            in_var(
+              in_module?,
+              left,
+              &in_list(&1, head, tail, expand, list, in_module?, false, [])
+            )
 
           {[head | tail], {_, vars_values}} ->
             {vars, values} = :lists.unzip(:lists.reverse(vars_values))
-            is_in_list = &in_list(&1, head, tail, expand, list, in_module?)
+            is_in_list = &in_list(&1, head, tail, expand, list, in_module?, false, [])
 
             quote do
               {unquote_splicing(vars)} = {unquote_splicing(values)}
@@ -3571,11 +3582,7 @@ defmodule Kernel do
         raise ArgumentError, "non-literal range in guard should be escaped with Macro.escape/2"
 
       right ->
-        raise ArgumentError, <<
-          "invalid args for operator \"in\", it expects a compile-time proper list ",
-          "or compile-time range on the right side when used in guard expressions, got: ",
-          Macro.to_string(right)::binary
-        >>
+        raise_on_invalid_args_in_2(right)
     end
   end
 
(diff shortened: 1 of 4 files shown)


lib/elixir/lib/kernel.ex
```diff
@@ -3669,48 +3676,117 @@ defmodule Kernel do
     end
   end
 
-  defp in_list(left, head, tail, expand, right, in_module?) do
-    fun = fn elem, acc ->
+  defp in_list(left, {:|, _, [head, tail]}, [], expand, right, in_module?, _use_right?, acc) do
```
eksperimental: This is an optimization: new argument `use_right?` keeps track of when `right` can be used, so we don't need to reverse `acc` and unquote `right` directly

lib/elixir/lib/kernel.ex
```diff
@@ -3669,48 +3676,117 @@ defmodule Kernel do
...
+        )
       end
     end
 
-    :lists.foldr(fun, comp(left, head, expand, right, in_module?), tail)
+    :lists.foldl(fun, comp(left, head, expand, right, false, false, acc), tail)
   end
 
-  defp comp(left, {:|, _, [head, tail]}, expand, right, in_module?) do
+  defp in_list(left, _head, [], _expand, right, true, true, _acc) do
```
eksperimental: And here is where we make use of the `use_right? = true`  optimization

lib/elixir/lib/kernel.ex
```diff
@@ -3543,17 +3543,28 @@ defmodule Kernel do
       [] when not in_module? ->
         false
 
-      [head | tail] = list when not in_module? ->
-        in_var(in_module?, left, &in_list(&1, head, tail, expand, list, in_module?))
+      [] ->
+        quote(do: :lists.member(unquote(left), []))
```
eksperimental: Can we just merge this clause with the previous one? and do

```elixir
      [] ->
￼	        false
```

josevalim: Yes, we could.

josevalim:
> previously:
>
> iex> quote do 1 in [1, 2, 3] end |> Macro.expand(Map.put(__ENV__, :context, :guard)) |> Macro.to_string()
":erlang.orelse(:erlang.\"=:=\"(1, 2), :erlang.orelse(:erlang.\"=:=\"(1, 3), :erlang.\"=:=\"(1, 1)))"
now:
>
> iex>  quote do 1 in [1, 2, 3] end |> Macro.expand(Map.put(__ENV__, :context, :guard)) |> Macro.to_string()
":erlang.orelse(:erlang.orelse(:erlang.\"=:=\"(1, 1), :erlang.\"=:=\"(1, 2)), :erlang.\"=:=\"(1, 3))"

It is important that we keep the orselse always as the second argument. Many passes in the compiler are optimized for handling `orelse` as right associative and this change can make the compiler or dialyzer quite slow.

We need to make it compile to:

    :erlang.orelse(:erlang.\"=:=\"(1, 1), :erlang.orelse(:erlang.\"=:=\"(1, 2), :erlang.\"=:=\"(1, 3)))

It seems the previous way is correct, it is just the first argument that was being weird misplaced?

lib/elixir/lib/kernel.ex
josevalim: Actually, we cannot. It would raise unused variable warnings. And not execute the left side at all. We can probably compile it to this though:

    quote do
      _ = unquote(left)
    end

Then it should safely apply. But yeah, they can't be merged, sorry.

lib/elixir/lib/kernel.ex
```diff
@@ -3543,17 +3543,28 @@ defmodule Kernel do
...
+
+      [head | tail] when not in_module? ->
+        in_var(
+          in_module?,
+          left,
+          &in_list(&1, head, tail, expand, right, in_module?, true, [])
+        )
 
       [_ | _] = list when in_module? ->
```
josevalim: We don't need to check the guard again, so we could do:

```suggestion
      [_ | _] = list ->
```

But possibly better is to move the clause above down and keep the guard in this one.

lib/elixir/lib/kernel.ex
josevalim: > keeps track of when right can be used

What do you mean by "use"? :)

josevalim:
Awesome job @eksperimental, especially on the tests. I have added some comments!

michalmuskala:
>     1. Optimize code when context is nil. Previously it was optimized as if for guards,
>        when it was not needed, ending up it code that was unnecessarily verbose and unoptimized.
> 
> 
> previously:
> 
> ```
> iex> quote do 1 in [1, 2, 3] end |> Macro.expand(__ENV__) |> Macro.to_string()
> "(\n  var = 1\n  :erlang.orelse(:erlang.\"=:=\"(var, 2), :erlang.orelse(:erlang.\"=:=\"(var, 3), :erlang.\"=:=\"(var, 1)))\n)"
> ```
> 
> now:
> 
> ```
> iex>  quote do 1 in [1, 2, 3] end |> Macro.expand(__ENV__) |> Macro.to_string()
> "(\n  var = 1\n  :lists.member(var, [1, 2, 3])\n)"
> ```

Are you sure `lists:members` is actually faster? The nested comparisons, especially on integers, will compile into a very efficient jump table that is O(1) to execute, while `lists:members` is always O(n) - even in other cases, the compiler can optimise the repeated comparisons to some sort of tree making the time logarithmic. `lists:members` calls are completely opaque to the compiler.

josevalim:
Good point @michalmuskala. I remember measuring at the time and the code is indeed faster by emitting `orelse`. So we should revert it and continue emitting the same code as in guards.

> ending up it code that was unnecessarily verbose and unoptimized.

So to clarify, the code is indeed more verbose but it is definitely more optimized. And we don't care about the verbosity of the compiled code really, so we should optimize for perf.

josevalim:
Thanks @eksperimental! I brought your improvements here (08829b3709a0b7445b5a33f8d1491a9e735fdf15) but kept the orelse form since it is more efficient!

PR #8943, opened 2019-04-07
Add a section about iodata to the IO docs

### LINKED ISSUE #8940: Document IO lists and IO data
We should have a section in the IO module about them.

whatyouhide:
Closes #8940.

### DIFF
--- a/lib/elixir/lib/io.ex
+++ b/lib/elixir/lib/io.ex
@@ -34,8 +34,8 @@ defmodule IO do
   IO data is a data type that can be used as a more efficient alternative to binaries
   in certain situations.
 
-  A term of type **IO data** is a binary or a list containing bytes (integers in `0..255`),
-  binaries, or nested iolists. The type is recursive. Let's see an example of one of
+  A term of type **IO data** is a binary or a list containing bytes (integers in `0..255`)
+  or nested IO data. The type is recursive. Let's see an example of one of
   the possible IO data representing the binary `"hello"`:
 
       [?h, "el", ["l", [?o]]]
--- a/lib/elixir/lib/io.ex
+++ b/lib/elixir/lib/io.ex
@@ -1,5 +1,5 @@
 defmodule IO do
-  @moduledoc """
+  @moduledoc ~S"""
   Functions handling input/output (IO).
 
   Many functions in this module expect an IO device as an argument.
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/io.ex
```diff
@@ -32,6 +29,109 @@ defmodule IO do
   was last accessed. The position of files can be changed using the
   `:file.position/2` function.
 
+  ## Iolists and iodata
```
fertapric: What do you think about "IO Data/IO Lists" or "IOData/IOLists"? I've seen the latter [in the Erlang/OTP codebase](https://github.com/erlang/otp/blob/master/lib/kernel/src/raw_file_io_list.erl#L74) 🙂

lib/elixir/lib/io.ex
```diff
@@ -32,6 +29,109 @@ defmodule IO do
...
+
+      def email(username, domain) do
+        username <> "@" <> domain
+      end
+
+  In this function, creating the email address will copy the `username` and `domain`
+  binaries. Now imagine you want to use the resulting email inside another binary:
+
+      def welcome_message(name, username, domain) do
+        "Welcome #{name}, your email is: #{email(username, domain)}"
```
fertapric: This heredoc allows interpolation.

lib/elixir/lib/io.ex
```diff
@@ -32,6 +29,109 @@ defmodule IO do
   was last accessed. The position of files can be changed using the
   `:file.position/2` function.
 
+  ## Iolists and iodata
+
+  Iolists and iodata are data structures that can be used as a sometimes
+  more efficient alternative to binaries.
```
josevalim: "that can be used as a more efficient alternative to binaries under certain situations".

lib/elixir/lib/io.ex
josevalim: Now that I think about it, we don't expose iolists anywhere in Elixir except in the typespecs. Most places where we say iolists informally we actually mean iodata. So my suggestion is to call it "IO data" and focus on that. We can have one note on iolists saying that iolists are the same as iodata except they don't accept a binary in the top level.

whatyouhide: Good call on dropping the iolists. We should go with "IO data" for the naming IMO.

whatyouhide:
@fertapric @josevalim updated with your suggestions. :)

lib/elixir/lib/io.ex
```diff
@@ -1,19 +1,16 @@
...
-  lists of characters and strings. In case another type is given,
-  functions will convert to string via the `String.Chars` protocol
-  (as shown in typespecs).
-
-  The functions starting with `bin` expect iodata as an argument,
-  i.e. binaries or lists of bytes and binaries.
+  The majority of the functions expect chardata. In case another type is given,
+  functions will convert those types to string via the `String.Chars` protocol
+  (as shown in typespecs). For more information on chardata, see the
+  "Iolists and iodata" section below.
```
fertapric:
```suggestion
  "IO data" section below.
```

whatyouhide: Good catch!

PR #8944, opened 2019-04-07
Define what a "switch" is

nielsbom:
The term "switch" is used exclusively here, "flag" is more common I think. I added some prose in the beginning to try and make this more clear.

Wikipedia has some (hopefully) default language for this: https://en.wikipedia.org/wiki/Command-line_interface#Command-line_option

### DIFF
--- a/lib/elixir/lib/option_parser.ex
+++ b/lib/elixir/lib/option_parser.ex
@@ -3,13 +3,13 @@ defmodule OptionParser do
   Functions for parsing command line arguments.
 
   When calling a command, it's possible to pass command line options
-  to modify what that command does. In this documentation those are
+  to modify what the command does. In this documentation, those are
   called "switches", in other situations they may be called "flags"
   or simply "options". A switch can be given a value, also called an
   "argument".
-  
+
   The main function in this module is `parse/2`, which parses a list
-  of command line arguments into a keyword list:
+  of command line options and arguments into a keyword list:
 
       iex> OptionParser.parse(["--debug"], strict: [debug: :boolean])
       {[debug: true], [], []}
--- a/lib/elixir/lib/option_parser.ex
+++ b/lib/elixir/lib/option_parser.ex
@@ -2,7 +2,7 @@ defmodule OptionParser do
   @moduledoc """
   Functions for parsing command line arguments.
 
-  When calling a command, it's possible to pass commandline options
+  When calling a command, it's possible to pass command line options
   to modify what that command does. In this documentation those are
   called "switches", in other situations they may be called "flags"
   or simply "options". A switch can be given a value, also called an


lib/elixir/lib/option_parser.ex
```diff
@@ -1,9 +1,15 @@
 defmodule OptionParser do
   @moduledoc """
-  Functions for parsing command line options.
-
-  The main function in this module is `parse/2`, which allows
-  developers to parse a list of arguments into options:
+  Functions for parsing command line arguments.
+
+  When calling a command, it's possible to pass commandline options
```
josevalim:
```suggestion
  When calling a command, it's possible to pass command line options
```

lib/elixir/lib/option_parser.ex
```diff
@@ -1,9 +1,15 @@
...
+  to modify what that command does. In this documentation those are
+  called "switches", in other situations they may be called "flags"
+  or simply "options". A switch can be given a value, also called an
```
josevalim: I don't think use "arguments" to refer to switch values in these docs though, do we?

nielsbom: Looks like we do that here: https://hexdocs.pm/elixir/OptionParser.html#parse/2-types
"Switches parsed by OptionParser may take zero or one arguments."

nielsbom:
I think there's quite a lot of (mixed) terminology in these docs, which is understandable, but also confusing :-)
We parse "command line arguments" into "options" but we call those "switches"... right?

Thinking about this again, we (I) could add an example in the beginning to define terminology...

josevalim:
@nielsbom that's a good point. By trying to define what it isn't, it makes it more confusing. Maybe we should just focus on the definition:

> When calling a command, it's possible to modify its behaviour by
> passing "switches" via the command line. A switch starts with two
> dashes (`-`) and is followed by alphanumeric characters, dashes,
> and underscore, such as `--foobar`. After a switch, zero, one or
> multiple arguments may be given. In other literate, switches may
> be referred to as "flags" or simply "options".

Thoughts?

nielsbom:
I like examples so I took your suggestion and added some examples:

> When calling a command, it's possible to modify its behaviour by
passing one or more "switches" via the command line. A switch starts with two
dashes (-) and the name of the switch. Instead of the full name it's also possible to define and use a one-character alias of that switch, preceded by a single dash (-). 
For example, in `elixir --version` the `version` switch is used. This switch has an alias that is used like so: `elixir -v`.
After a switch, zero, one or multiple arguments may be given.
For example, in `iex -S mix` we use the `S` switch and pass it the value of `mix`.
In other systems, switches may be referred to as "flags" or simply "options".

I removed the "alphanumeric characters, dashes, and underscore" because I think that's too much detail at this time in the text.

josevalim:

PR #8945, opened 2019-04-07
Support diff on pattern matching

ggcampinho:
This is the preparation for #6248

This PR adds the ability to diff a pattern match doing:
1. `ExUnit.Diff` can compare an extended AST with a value
2. `ExUnit.Diff` transforms the diff result into algebra
3. `ExUnit.Assertions` and `ExUnit.Formatter` use the new methods

### 1. `ExUnit.Diff` can compare an extended AST with a value

The function `compare_quoted/4` is added to compare an extended AST (`left`) to a value (`right`), returning a `ExUnit.Diff` struct with `equivalent?`, `left` and `right` attributes. The `left` and `right` attributes of the struct contain also an extended AST with the meta `diff: true` when something on that side has no equivalent on the other side, and a block with meta `diff_container` for diffs inside numbers and strings.

To use the `compare_quoted/4` function, the `left` side has to be prepared first, where macros need to be expanded and added as `expanded` meta on their original nodes, to keep the original AST and to be able to compare the expanded AST. If the `left` side is a value, it needs to be transformed into an AST with pids and references being untouched.

To simplify the list comparison, the lists are parsed and `++`/`|` operators are removed with their indices, so the list can be rebuilt properly later.

The result of the comparison should be equivalent to the current implementation.

### 2. `ExUnit.Diff` transforms the diff result into algebra

The `to_algebra/2` is provided to transform the `ExUnit.Diff` struct into algebra docs. It receives a wrapper that is called when a meta `diff: true` is found.

### 3. `ExUnit.Assertions` and `ExUnit.Formatter` use the new methods

`ExUnit.Assertions` now sends the `left` side as an extended AST for every case, and also adds the pins and the context (`:match` or `nil`) of the comparison. `ExUnit.Formatter` uses the new `ExUnit.Diff.compare_quoted/4`.


ggcampinho:
Screenshot 2019-04-09 at 09 55 46
Screenshot 2019-04-09 at 09 55 53
Screenshot 2019-04-09 at 09 56 08
Screenshot 2019-04-09 at 09 56 17
Screenshot 2019-04-09 at 09 58 21
Screenshot 2019-04-09 at 09 58 27
Screenshot 2019-04-09 at 09 58 37

josevalim:
This has been merged, sorry for the delay! Thank you so much @jeremyowensboggs and @ggcampinho! :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

There are two next steps left:

1. Use the new diff on `assert_receive`

2. Also compute the diff when guards are involved. In other words, if the pattern matches, then it likely because of guards, so we need to evaluate it. We only need to evaluate the root nodes of every `when`, `or`, and `and`. So if we have something like: `condition and tuple_size(foo) == tuple_size(bar)`, we extract both sides of `or`, and if it fails, we paint it in red. The unfortunate complexity in this approach is that it requires us to implement a subset of the language (basically function calls and the creation of lists, tuples, maps).

Let me know if you want to contribute these remaining changes!

ggcampinho:
Thank you! I'll work on the remaining changes and I'll send it soon! :)

ericmj:
I think there is some issue on re-using the same code for values and quoted code for the left side of `ExUnit.Diff.compute/3`. I haven't looked into much detail but as an example of code that fails would be `assert_diff({:a, :b, :c} == {:a, :b, :c}, [])` because `{:a, :b, :c}` is interpreted as an AST variable.

josevalim:
@ericmj Good catch. And this is non-trivial to solve because the tuple could be inside a struct and escaping all of that can be complicated. The best option is most likely to have two distinct entry points for the diffing and make sure they re-use most functions. @ericmj can you please open up a new issue?

PR #8947, opened 2019-04-09
Check that the first letter of alias is upcase

ggcampinho:
The current approach discards the `__aliases__` tuple and do a simple check that allows some strange behavior like:

```elixir
iex(1)> alias :lists, as: :"Elixir.foobar"
:lists
iex(2)> :"Elixir.foobar"
:"Elixir.foobar"
iex(3)> alias :"Elixir.foobar", as: Baz
:"Elixir.foobar"
iex(3)> Baz
:lists
```

Using `__aliases__` tuple is more restrict and not backwards compatible, so a warning was introduced.

```elixir
iex(1)> alias :lists, as: :"Elixir.foobar"
warning: invalid value for option :as, expected an alias, got: :"Elixir.foobar"
  iex:1

:lists
```

### DIFF
--- a/lib/elixir/test/elixir/kernel/warning_test.exs
+++ b/lib/elixir/test/elixir/kernel/warning_test.exs
@@ -562,24 +562,6 @@ defmodule Kernel.WarningTest do
     purge(Sample)
   end
 
-  test "expands args and warn for atoms beginning with `Elixir.`" do
-    message = ~r"invalid value for option :as, expected an alias, got: :\"Elixir.foobar\""
-
-    assert capture_err(fn ->
-             Code.eval_string("""
-               alias :lists, as: :"Elixir.foobar"
-             """)
-           end) =~ message
-
-    message = ~r"invalid value for option :as, expected an alias, got: Foobar"
-
-    assert capture_err(fn ->
-             Code.eval_string("""
-               alias :lists, as: :"Elixir.Foobar"
-             """)
-           end) =~ message
-  end
-
   test "unused inside dynamic module" do
     import List, only: [flatten: 1], warn: false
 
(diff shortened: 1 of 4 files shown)


lib/elixir/test/elixir/kernel/warning_test.exs
```diff
@@ -562,6 +562,24 @@ defmodule Kernel.WarningTest do
...
+             Code.eval_string("""
+               alias :lists, as: :"Elixir.foobar"
+             """)
+           end) =~ message
+
+    message = ~r"invalid value for option :as, expected an alias, got: Foobar"
+
+    assert capture_err(fn ->
+             Code.eval_string("""
+               alias :lists, as: :"Elixir.Foobar"
```
josevalim: It is actually fine for it to be an atom. Sometimes you may generate an alias name dynamically and use `unquote(generated)` when building the alias and that would break it, since `generated` is most likely an atom.

My suggestion is to keep today's code as is but check that, right after we split on ".", than the first letter of the alias is in upcase.

ggcampinho-travis: I still think that is a bit strange because the following is confusing:
```
iex(1)> alias :lists, as: :"Elixir.Foobar"
:lists
iex(2)> :"Elixir.Foobar"
Foobar
iex(3)> Foobar          
:lists
```
It feels that the best thing would be to have some kind of `Macro.alias/1` like there is `Macro.var/2`, but I imagine that it will break too many libraries.

josevalim: I agree it is weird if you implement it by hand but most likely that would be generated through a macro or Macro.concat and forcing the user to step back to an alias would be a bit awkward and a backwards incompatible change. :)

josevalim:

PR #8951, opened 2019-04-11
Improve typespecs for Enum.*reduce functions

whatyouhide:
Using `acc` and `element` to make it easier to understand the order of the terms when reading the typespec. We have an `Enum.acc/0` type, should we use that instead of `when acc: any`?

### DIFF
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -1079,7 +1079,7 @@ defmodule Enum do
       {[[1], [2], [3], [4], [5]], 15}
 
   """
-  @spec flat_map_reduce(t, acc, fun) :: {[any], any}
+  @spec flat_map_reduce(t, acc, fun) :: {[any], acc}
         when fun: (element, acc -> {t, acc} | {:halt, acc}),
              acc: any
   def flat_map_reduce(enumerable, acc, fun) do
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/enum.ex
```diff
@@ -1438,7 +1438,8 @@ defmodule Enum do
       {[2, 4, 6], 6}
 
   """
-  @spec map_reduce(t, any, (element, any -> {any, any})) :: {any, any}
+  @spec map_reduce(t, acc, (element, acc -> {element, acc})) :: {list, acc}
+        when acc: any
```
josevalim: The thing to remember is that `acc: any` is not necessary the same as not having `acc: any`, because `when acc: any` may imply that, if one `acc` is integer, then all of them is integer.

I don't think it matters though and dialyzer applies a broader semantic anyway, so I would remove the `when` and just rely on the type.

whatyouhide: Awesome, agreed. I'll do that in all instances I can find in `Enum`.

whatyouhide: @josevalim done.

lib/elixir/lib/enum.ex
```diff
@@ -1079,9 +1079,8 @@ defmodule Enum do
       {[[1], [2], [3], [4], [5]], 15}
 
   """
-  @spec flat_map_reduce(t, acc, fun) :: {[any], any}
-        when fun: (element, acc -> {t, acc} | {:halt, acc}),
-             acc: any
+  @spec flat_map_reduce(t, acc, fun) :: {[any], acc}
```
umamaistempo: OOC: why `[any]` instead of `list` like in `map_reduce/3` ?

whatyouhide: Virtually the same thing, I just left it as it was here. We could have changed it for consistency, but now that the PR has been merged the change is probably not worth it :)

umamaistempo: agreed

PR #8952, opened 2019-04-13
Introduce Config module

josevalim:
With the addition of releases to Elixir, we need to move configuration out of Mix and into Elixir per-se. This pull-request is a proposal for the new Config API.

I have decided to split the current `Mix.Config` API in two modules: `Config` and `Config.Reader`. A new function was also added to `Application`. Here is the new mapping:

| Old                  | New                                |
|:---------------------|:-----------------------------------|
| use Mix.Config       | import Config                      |
| Mix.Config.merge/2   | Config.Reader.merge/2              |
| Mix.Config.read!/2   | Config.Reader.read!/2              |
| Mix.Config.eval!/2   | Config.Reader.read_with_imports!/2 |
| Mix.Config.persist/1 | Application.put_all_env/2          |

The motivation for having two modules is to allow us to move from `use Mix.Config` to `import Config`, as it helps promote imports as more idiomatic. Therefore, the `Config` module contains the actual configuration API (`config/2`, `config/3`, and `import_config/1`). While config manipulation functions are in `Config.Reader`. However, as you can see in the code, this is a bit awkward, as all the `Config.Reader` module does is to delegate to underscored functions in `Config`.

I am opening this PR as a WIP (for example, we have no tests), mostly so we can discuss the API. We have two options:

  1. Split `Mix.Config` into `Config` and `Config.Reader`, exactly as done in this PR

  2. Simply move `Mix.Config` into `Config`. This means that `import Config` will also import `merge/2`, `read!/2` and `read_with_imports!/2`, which are not really useful when defining a configuration itself, therefore they are unlikely to be used (but could be seen as confusing)

Furthermore, since I will work on config providers next, we may actually have strong reasons for keeping two modules or unifying them, so we may have more arguments in the future. But currently we have enough content to push the discussion forward.


josevalim:
As suspected, as I worked on the Config.Provider functionality, we do have enough reason to have a Config.Reader, so I will keep things as is. A more complete PR will be sent soon.

PR #8955, opened 2019-04-13
Fix the documentation for ExUnit.Assertions.assert/2

smadeja:
The documentation for `assert/2` suggests that it behaves differently than `assert/1`, which is not the case.


fertapric:
Thanks @smadeja! 💜First contribution 🙌

PR #8956, opened 2019-04-13
Implement static_atom_encoder option

arjan:
This PR introduces the `static_atom_encoder` option to `Code.string_to_quoted/2`, as described in [this elixir-lang-core thread](https://groups.google.com/forum/#!topic/elixir-lang-core/1A0kOswSB7U).

I think it is ready for review; code, docs and 2 added tests.

### DIFF
--- a/lib/elixir/lib/code.ex
+++ b/lib/elixir/lib/code.ex
@@ -669,7 +669,7 @@ defmodule Code do
 
   When `static_atom_encoder: &my_encoder/2` is passed as an argument,
   `my_encoder/2` is called every time the tokenizer needs to create a
-  'static' atom. Static atoms are atoms in the AST that function as
+  "static" atom. Static atoms are atoms in the AST that function as
   aliases, remote calls, local calls, variable names, regular atoms
   and keyword lists.
 
--- a/lib/elixir/lib/code.ex
+++ b/lib/elixir/lib/code.ex
@@ -651,7 +651,8 @@ defmodule Code do
       Defaults to `false`.
 
     * `:static_atom_encoder` - The static atom encoder function, see
-      the section below. The `:static_atom_encoder` option overrides
+      the ["The `:static_atom_encoder` function" section](module-the-static_atom_encoder-function)
+      below. The `:static_atom_encoder` option overrides
       the `:existing_atoms_only` option.
 
     * `:warn_on_unnecessary_quotes` - when `false`, does not warn
--- a/lib/elixir/lib/code.ex
+++ b/lib/elixir/lib/code.ex
@@ -652,7 +652,9 @@ defmodule Code do
 
     * `:static_atom_encoder` - The static atom encoder function, see
       "The `:static_atom_encoder` function" section below. This option
-      overrides the `:existing_atoms_only` option.
+      overrides the `:existing_atoms_only` behaviour for static atoms
+      but `:existing_atoms_only` is still used for dynamic atoms, such
+      as atoms with interpolations.
 
     * `:warn_on_unnecessary_quotes` - when `false`, does not warn
       when atoms, keywords or calls have unnecessary quotes on
--- a/lib/elixir/lib/code.ex
+++ b/lib/elixir/lib/code.ex
@@ -651,9 +651,9 @@ defmodule Code do
       Defaults to `false`.
 
     * `:static_atom_encoder` - The static atom encoder function, see
-      the ["The `:static_atom_encoder` function" section](module-the-static_atom_encoder-function)
-      below. The `:static_atom_encoder` option overrides
-      the `:existing_atoms_only` option.
+      ["The `:static_atom_encoder`
+      function"](#string_to_quoted/2-the-static_atom_encoder-function) section
+      below. This option overrides the `:existing_atoms_only` option.
 
     * `:warn_on_unnecessary_quotes` - when `false`, does not warn
       when atoms, keywords or calls have unnecessary quotes on
@@ -685,11 +685,15 @@ defmodule Code do
   if you want to use the Elixir parser in a user-facing situation, but
   you don't want to exhaust the atom table.
 
-  When the encoder fails for some reason, it must return `{:error,
-  reason :: binary}`. The atom encoder is not called for *all* atoms
-  that are present in the AST: it won't be invoked for operators,
-  syntax keywords (fn, do, etc), and neither for atoms containing
-  interpolations.
+  The atom encoder is not called for *all* atoms that are present in
+  the AST. It won't be invoked for the following atoms:
+
+    * operators (`:+`, `:-`, and so on)
+
+    * syntax keywords (`fn`, `do`, `else`, and so on)
+
+    * atoms containing interpolation (`:"\#{1 + 1} is two"`), as these
+      atoms are constructed at runtime.
 
   """
   @spec string_to_quoted(List.Chars.t(), keyword) ::
--- a/lib/elixir/lib/code.ex
+++ b/lib/elixir/lib/code.ex
@@ -651,9 +651,8 @@ defmodule Code do
       Defaults to `false`.
 
     * `:static_atom_encoder` - The static atom encoder function, see
-      ["The `:static_atom_encoder`
-      function"](#string_to_quoted/2-the-static_atom_encoder-function) section
-      below. This option overrides the `:existing_atoms_only` option.
+      "The `:static_atom_encoder` function" section below. This option
+      overrides the `:existing_atoms_only` option.
 
     * `:warn_on_unnecessary_quotes` - when `false`, does not warn
       when atoms, keywords or calls have unnecessary quotes on
--- a/lib/elixir/lib/code.ex
+++ b/lib/elixir/lib/code.ex
@@ -684,10 +684,11 @@ defmodule Code do
   if you want to use the Elixir parser in a user-facing situation, but
   you don't want to exhaust the atom table.
 
-  When the encoder fails for some reason, it should return `{:error,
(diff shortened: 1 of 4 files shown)


lib/elixir/src/elixir_tokenizer.erl
```diff
@@ -807,6 +809,16 @@ is_unnecessary_quote([Part], #elixir_tokenizer{warn_on_unnecessary_quotes=true}
 is_unnecessary_quote(_Parts, _Scope) ->
   false.
 
+unsafe_to_atom(Part, Line, Column, #elixir_tokenizer{static_atoms_encoder=StaticAtomsEncoder} = Scope) when
+    is_function(StaticAtomsEncoder) ->
```
whatyouhide: We can specify the arity of the `StaticAtomsEncoder` function:

```suggestion
    is_function(StaticAtomsEncoder, 2) ->
```

lib/elixir/test/elixir/code_test.exs
```diff
@@ -153,6 +153,36 @@ defmodule CodeTest do
                {:error, {1, "unsafe atom does not exist: ", "there_is_no_such_atom"}}
     end
 
+    test "supports static_atoms_encoder" do
+      ref = make_ref()
+
+      encoder = fn "there_is_no_such_atom", metadata ->
```
whatyouhide: Nitpick, but maybe for better assertion errors we can have:

```elixir
encoder = fn atom, meta ->
  assert atom == "ther_is_no_such_atom"
  assert meta[:line] == =1
  # ...same
end
```

lib/elixir/src/elixir_tokenizer.erl
```diff
@@ -807,6 +809,16 @@ is_unnecessary_quote([Part], #elixir_tokenizer{warn_on_unnecessary_quotes=true}
...
+  Metadata = [{line, Line}, {column, Column}, {file, Scope#elixir_tokenizer.file}],
+  Value = elixir_utils:characters_to_binary(Part),
+  case StaticAtomsEncoder(Value, Metadata) of
+    {ok, Term} ->
+      {ok, Term};
+    {error, Reason} ->
+      {error, {Line, Column, Reason, Part}}
+  end;
```
whatyouhide: Does having this clause before the next one mean that `:statis_atoms_encoder` will let atoms that are bigger than 255 codepoints work? Since the tokenizer now raises an error for atoms that go over the system limit, maybe it's better to keep that behaviour?

lib/elixir/src/elixir_tokenizer.erl
josevalim: Since this is called during the operation and not during input validation, I think it should stay as is, otherwise you can pass a function with arity 1 and never know why it was not called.

lib/elixir/test/elixir/code_test.exs
```diff
@@ -153,6 +153,36 @@ defmodule CodeTest do
...
+          ArgumentError ->
+            {:ok, {:user_atom, string}}
+        end
+      end
+
+      assert {:ok, {:try, _, [[do: {:test, _, [{{:user_atom, "atom_does_not_exist"}, _, []}]}]]}} =
+               Code.string_to_quoted("try do: test(atom_does_not_exist())",
+                 static_atoms_encoder: encoder
+               )
+    end
```
josevalim: We also need a test for the `{:error, reason}` branch. For instance, I think we will have to convert the Reason to a charlist and the current code does not do it!

lib/elixir/src/elixir_tokenizer.erl
arjan: I actually did this on purpose because when you pass an atom encoder, I think it should be up to yourself to check limits like these. I can imagine an encoder function that takes too long atom names and transforms them into atoms that fit the system limit.

lib/elixir/test/elixir/code_test.exs
arjan: Yep you are right. Sorry I forgot, will add this. The return value for the error was not entirely clear for me (the spec just says `{:error, {line :: pos_integer, term, term}}`).

lib/elixir/test/elixir/code_test.exs
arjan: Addressed this in 23e3b95

lib/elixir/test/elixir/code_test.exs
arjan: Fix this in 282675e0570a1e0f6e7be0579f7495e11ba4bf9d

Note that it adds `": "` to the error message. Which is kind of weird but it is in line with the other error messages. Alternatively it could not do that but then the encoder callback function is supposed to add it, which is also questionable..

lib/elixir/src/elixir_tokenizer.erl
whatyouhide: @arjan I am still not sure about this. I think it depends on what we want this functionality to address. If we want this to let people parse random code safely, then I would go with keeping the hard limit on 255 characters.

lib/elixir/lib/code.ex
```diff
@@ -650,6 +650,10 @@ defmodule Code do
       when non-existing atoms are found by the tokenizer.
       Defaults to `false`.
 
+    * `:static_atom_encoder` - The static atom encoder function, see
+      the section below. The `:static_atom_encoder` option overrides
```
whatyouhide: Try to avoid pointing to just "the section below" since this quickly becomes outdated if we move documentation around. Use this instead:

```suggestion
      the ["The `:static_atom_encoder` function" section](module-the-static_atom_encoder-function)
      below. The `:static_atom_encoder` option overrides
```

lib/elixir/lib/code.ex
```diff
@@ -659,6 +663,33 @@ defmodule Code do
   The opposite of converting a string to its quoted form is
   `Macro.to_string/2`, which converts a quoted form to a string/binary
   representation.
+
+  ## The `:static_atom_encoder` function
+
+  When `static_atom_encoder: &my_encoder/2` is passed as an argument,
+  `my_encoder/2` is called every time the tokenizer needs to create a
+  'static' atom. Static atoms are atoms in the AST that function as
```
whatyouhide: Nitpick, but we don't use single quotes `'` anywhere in the Elixir docs that I know of.

```suggestion
  "static" atom. Static atoms are atoms in the AST that function as
```

lib/elixir/lib/code.ex
```diff
@@ -659,6 +663,33 @@ defmodule Code do
...
+  however, in that case the AST is no longer "valid" in that it cannot
+  be used to compile or evaluate Elixir code. A use case for this is
+  if you want to use the Elixir parser in a user-facing situation, but
+  you don't want to exhaust the atom table.
+
+  When the encoder fails for some reason, it must return `{:error,
+  reason :: binary}`. The atom encoder is not called for *all* atoms
+  that are present in the AST: it won't be invoked for operators,
+  syntax keywords (fn, do, etc), and neither for atoms containing
+  interpolations.
```
whatyouhide: This is an important note, so I would write it in a more formal way:

```
The atom encoder is not called for *all* atoms that are present in
the AST. It won't be invoked for the following atoms:

  * operators (`:+`, `:-`, and so on)

  * syntax keywords (`fn`, `do`, and so on)

  * atoms containing interpolation (`:"#{1 + 1} is two"`)
```

Wdyt?

arjan: Yep, looks good, done in f9ae8e7

lib/elixir/src/elixir_tokenizer.erl
arjan: I'm fine with changing this behaviour around so that any atoms-to-be are first length-checked. It's an edge case anyway.

josevalim: I think 255 is sane limit to have anyway, so let's move the limit check clause up.

lib/elixir/test/elixir/code_test.exs
josevalim: :+1:

lib/elixir/lib/code.ex
```diff
@@ -650,6 +650,10 @@ defmodule Code do
       when non-existing atoms are found by the tokenizer.
       Defaults to `false`.
 
+    * `:static_atom_encoder` - The static atom encoder function, see
+      "The `:static_atom_encoder` function" section below. This option
+      overrides the `:existing_atoms_only` option.
```
josevalim: I removed the link because it doesn't work nicely when read on the terminal. :(

lib/elixir/src/elixir_tokenizer.erl
umamaistempo: 🤔  IMO, such case should be treated as a different clause but that would not be terse and would be a bit too defensive 😅

whatyouhide: Oh good catch @josevalim, this would only make the clause fail and cascade. Agreed, better to have the actual call fail with wrong number of arguments then.

umamaistempo:
Adding that 255 check (and potentially pointing that limitation on the doc) and LGTM :+1:

lib/elixir/lib/code.ex
umamaistempo: :(

lib/elixir/src/elixir_tokenizer.erl
arjan: Fixed in 970ba29

whatyouhide:
Great job, thanks @arjan! 💟

arjan:
Thanks for merging! 😁

PR #8957, opened 2019-04-15
Add config providers

josevalim:
This PR adds config providers to Elixir releases (see #8612).

It is a large-ish PR broken in two parts:

  1. Adding the `Config`, `Config.Reader` and `Config.Provider` modules
  2. Adding config provider support to releases

I will explore the rationale for each part next. Note this is still WIP (docs and tests missing) but I have submitted it so we can get the CI running and open up for feedback.

## Adding the `Config`, `Config.Reader` and `Config.Provider` modules

Since we need to use config files during releases and releases do not necessarily have Mix available, we have moved `Mix.Config` to `Config`. We also used this opportunity to refactor the API.

The first change is to use `import Config` instead of `use Mix.Config`, since `import` is more idiomatic. We also kept only the user config in `Config` and move all of the operations to manipulate config files to the `Config.Reader` module. Here is a summary table:

| Old                  | New                                |
|:---------------------|:-----------------------------------|
| use Mix.Config       | import Config                      |
| Mix.Config.merge/2   | Config.Reader.merge/2              |
| Mix.Config.read!/2   | Config.Reader.read!/2              |
| Mix.Config.eval!/2   | Config.Reader.read_imports!/2      |
| Mix.Config.persist/1 | Application.put_all_env/2          |

We have also added the `Config.Provider` module which outlines the behaviour to be implemented by config providers. It is simply two functions. The `Config.Reader` module itself is a provider that reads a single configuration file.

## Adding config provider support to releases

We have added config providers support to releases using a restart-once approach. This is different from `distillery` and `relx` that boot the VM twice. Instead, we only boot the VM once and restart the Erlang system once it has been configured. The trick is to rewrite the sys.config during after running config providers.

Therefore, to use config providers, we need writable permissions somewhere (which can be customizable with `RELEASE_TMP`). If you don't use providers, you don't need write permission. Also, since config providers run when only a small part of the VM is started, they need to explicitly start any application they may need.

By default there is a single config provider which checks for `config/releases.exs`. If `config/releases.exs` is present, it will be automatically copied to your release and executed at runtime, the first time your release boots.


lib/elixir/lib/application.ex
```diff
@@ -523,10 +523,32 @@ defmodule Application do
...
+
+  It receives the same options as `put_env/4`. Returns `:ok`.
+  """
+  @spec put_all_env([{app, [{key, value}]}], timeout: timeout, persistent: boolean) :: :ok
+  def put_all_env(config, opts \\ []) when is_list(config) and is_list(opts) do
+    # TODO: Remove function exported? check when we require Erlang/OTP 22+
+    if function_exported?(:application, :set_env, 2) do
+      :application.set_env(config, opts)
+    else
+      for {app, kw} <- config, {k, v} <- kw do
```
umamaistempo: kw ?

lib/elixir/lib/application.ex
```diff
@@ -523,10 +523,32 @@ defmodule Application do
     :application.set_env(app, key, value, opts)
   end
 
+  @doc """
+  Puts the environment for multiple apps at the same time.
+
+  It will fail if any of the applications appear more than once
```
umamaistempo: Opinion: change to `It _might_ fail` as it is an undefined behaviour right now and only happens on the newest OTP as to avoid anyone relying on this behaviour

lib/elixir/lib/application.ex
```diff
@@ -523,10 +523,32 @@ defmodule Application do
...
+  or if any of the keys in the application are repeated.
```
umamaistempo: I think it would be clearer if it was written like

a) "or if repeated keys are found on an application config"
b) "or if any application contains repeated keys"
c) "or if any of the keys are repeated within an application"

On those suggestions I'm trying to make it clear that it's okay to have equal keys in different applications but not on the same applications :b

josevalim:
Unfortunately the current approach doesn't work on Windows because Windows only allows one `-config` file. One approach is to override to copy the `sys.config` itself to another location and override it. I have opened an issue in [Erlang issues tracker](https://bugs.erlang.org/browse/ERL-912) and we will resolve accordingly based on that.

lib/elixir/lib/application.ex
whatyouhide: It means "keyword" as in "keyword list". @josevalim if we wanted to be nitpicky I'd probably suggest to call `kw` `keyword` and `k`/`v` `key`/`value` instead.

michalmuskala:
I like this very much, except for one thing - caching of the provider results. I think it will be very surprising and error prone.

Let's imagine a situation where you deploy to heroku with the new releases and providers that reads from system env, and the `FOO` var in particular. This means that when you run the command `heroku config:set FOO=new_value`, heroku will restart the app with the new value in system env, but because the evaled config file already exists, providers won't execute again and the new value won't be picked up. This is in no way a desired behaviour. Not only that, but even after some googling I didn't find a reliable way to tell heroku to discard this file in all dynos so that a new version could be generated.

sorentwo:
Caching the compiled configuration seems like a no-go to me as well. 
Outside of the tmp folder Heroku provides a read only file system, so I’m not sure you could even get the release to boot.

josevalim:
@sorentwo not though we need to have write permissions to something, regardless if caching or not.

sorentwo:
@josevalim I double checked, there are write permissions inside the application directory, so that isn't an actual problem.

Caching the initial config and then not overwriting it when the environment changes, as @michalmuskala mentioned, would be enough to prevent Heroku users from using releases though.

josevalim:
> Caching the initial config and then not overwriting it when the environment changes, as @michalmuskala mentioned, would be enough to prevent Heroku users from using releases though.

Yes, I agree. I was just commenting on the second part. :D

josevalim:
So I pushed a new commit that uses a new random file per boot. The following commit adds automatic pruning of those config files (although it means you can't call another System.restart or perform hot code upgrades unless you disable the pruning).

josevalim:
And thank you @umamaistempo, @michalmuskala, and @sorentwo for the feedback!

lib/elixir/lib/application.ex
```diff
@@ -523,10 +523,38 @@ defmodule Application do
...
+
+  It receives the same options as `put_env/4`. Returns `:ok`.
+  """
+  @spec put_all_env([{app, [{key, value}]}], timeout: timeout, persistent: boolean) :: :ok
+  def put_all_env(config, opts \\ []) when is_list(config) and is_list(opts) do
+    # TODO: Remove function exported? check when we require Erlang/OTP 22+
+    if function_exported?(:application, :set_env, 2) do
+      :application.set_env(config, opts)
+    else
+      for {app, keyword} <- config, {key, value} <- keyword do
```
fertapric: Should we raise instead of ignore if `keyword` is not a keyword list (like `Config.config/2`)?

lib/elixir/lib/config.ex
```diff
@@ -0,0 +1,255 @@
...
+      import_config "#{Mix.env()}.exs"
+
+  `import Config` will import the functions `config/2`, `config/3`
+  and `import_config/1` to help you manage your configuration.
+
+  Once Mix starts, it will automatically evaluate the configuration
+  file persist it into `:some_app`'s application environment, which
+  can be accessed in as follows:
+
+      "value" = Application.fetch_env!(:my_app, :key1)
```
fertapric:
```suggestion
      "value1" = Application.fetch_env!(:my_app, :key1)
```

lib/mix/lib/mix/tasks/release.ex
```diff
@@ -437,6 +445,9 @@ defmodule Mix.Tasks.Release do
     * `RELEASE_NODE` - the release node name, in the format `name@host`.
       It can be overridden on `bin/start` to a custom value
 
+    * `RELEASE_TMP` - the directory in the release to write temporary
+      files to. Defaults to the `$RELEASE_ROOT/tmp`
```
```suggestion
      files to. It can be overridden on `bin/start` to a custom value.
      Defaults to the `$RELEASE_ROOT/tmp`
```

lib/mix/lib/mix/tasks/release.ex
```diff
@@ -404,10 +409,12 @@ defmodule Mix.Tasks.Release do
           ebin/
           include/
           priv/
-      releases
+      releases/
         RELEASE_VSN/
+          consolidated/
           elixir
           iex
+          releases.exs
```
fertapric: I'm not sure about the term "releases" (in plural) inside the directory of a specific release. Maybe `runtime_config.exs`?

josevalim: I had `runtime_config.exs` but given the user file is named `config/releases.exs`, I decided to keep the same name as it is less confusing once you know the context.

josevalim:
I pushed docs for reader and provider. Now I need to:

  * [ ] Tests for provider
  * [ ] Tests for application.put_all_env
  * [ ] Tests for Mix.Release

lib/elixir/lib/config/provider.ex
```diff
@@ -0,0 +1,203 @@
...
+
+  @typedoc """
+  A path pointing to a configuration file.
+
+  Since configuration files are often accessed on target machines,
+  it can be expressed either as:
+
+    * a binary representing an absolute path
+
+    * a tuple {:system, system_var, path} where the config is the
```
fertapric: could `{:system, "RELEASE_SYS_CONFIG", ".config"}` be replaced by a `:init` (or similar) that would be translated into reading the config path from `:init.get_argument/1`? Or the ENV var has other use cases?

josevalim: That's an interesting idea but the user could pass an extra `-config` and then which one we pick could be a bit too confused. We could aim for the sys.config one but I would prefer to keep those decoupled.

josevalim:
Pushed docs for configuration here: https://github.com/elixir-lang/elixir/pull/8957/files#diff-53c844525c23fd1a21c03113baf7e781R335

josevalim:
> The configuration can have sensitive data and this could be exposed by some malicious code in some cases, I believe.

I am not really worried about this. A config provider is regular Elixir code and as such it has access to absolutely everything. if you are worried about a bad config provider, the current config is the least of your concerns. :)

lib/elixir/lib/config.ex
```diff
@@ -0,0 +1,255 @@
...
+  end
+
+  defp delete_files() do
+    Process.delete(@files_key)
+  end
+
+  defp raise_improper_use!() do
+    raise "could not set configuration via Config. " <>
+            "This usually means you are trying to execute a configuration file " <>
+            "directly, instead of reading it with Config.Reader or the proper mix API"
```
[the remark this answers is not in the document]
josevalim: Fixed!

josevalim:
Tests are done, now let's wait for CI and once it is green it will be merged!

josevalim:

Qqwy:
I just took some time to read through the code and documentation, and would like to say: this is really cool stuff. I think the configuration story in Elixir will drastically improve with this. Awesome job! :heart_eyes:

PR #8959, opened 2019-04-16
Fix @spec for `GenServer.code_change/3`

NJichev:
As we can see [here](http://erlang.org/doc/man/gen_server.html#Module:code_change-3),
the function returns `{:ok, new_state} | {:error, reason}` while the
old version can either be old version or `{:down, current version}`


whatyouhide:
Thanks @NJichev! 💟

PR #8962, opened 2019-04-18
Optimize &super/arity and &super(&1)

### LINKED ISSUE #8610: Optimize &super/arity
Today we end-up generating an intermediate function but such is not necessary.

fertapric:
Closes #8610 

The previous implementation was generating an unnecessary intermediate function when using `&super/arity`.

In addition, capture expressions using `super` with sequential value placeholders (`&n`) were not optimized like other functions. For example, the capture expression `&foo(&1, &2, &3)` is optimized to `&foo/3` because the value placeholders (`&n`) are sequential `&1, &2, &3`.

### DIFF
--- a/lib/elixir/src/elixir_expand.erl
+++ b/lib/elixir/src/elixir_expand.erl
@@ -204,29 +204,29 @@ expand({quote, Meta, [_, _]}, E) ->
 
 %% Functions
 
-expand({'&', Meta, [{super, SuperMeta, Args}]}, E) when is_list(Args) ->
-  {_Kind, Name, _} = resolve_super(Meta, length(Args), E),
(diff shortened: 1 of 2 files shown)


fertapric:
The current PR does not work with macros, and unfortunately those cases are not covered by the test suite. I'll update the PR with macro support and tests.

josevalim:
Btw, I thought if we should move the `resolve_super` to defoverridable and/or handle the super logic inside elixir_fn but since `super` requires many of the helpers in `elixir_expand`, I am OK with keeping things as is in the PR and matching three times on `&` in elixir_expand.

fertapric:
@josevalim my thoughts exactly 🙂

PR #8966, opened 2019-04-19
Add :steps to releases

josevalim:


lib/mix/lib/mix/release.ex
```diff
@@ -263,6 +270,21 @@ defmodule Mix.Release do
     |> Keyword.put(:kernel, :permanent)
   end
 
+  defp validate_steps!(steps) do
+    if not is_list(steps) or Enum.any?(steps, &(&1 != :assemble and not is_function(&1, 1))) do
+      Mix.raise(
+        "The :steps option must be a list of anonymous functions that receive one argument " <>
```
fertapric: To remove ambiguity, what do you think about?

```
The :steps option must be a list of:

  * anonymous function that receives one argument
  * the atom :assemble

Got: #{inspect(steps)}
```

lib/mix/lib/mix/release.ex
```diff
@@ -263,6 +270,21 @@ defmodule Mix.Release do
...
+          "or the atom :assemble, got: #{inspect(steps)}"
+      )
+    end
+
+    if :assemble not in steps do
```
fertapric: Is it ok if the list contains the `:assemble` atom more than once?

lib/mix/lib/mix/release.ex
josevalim: Excellent! :+1:

lib/mix/lib/mix/release.ex
josevalim: Excellent! :+1:

josevalim:

PR #8968, opened 2019-04-19
Support ISO8601 timestamps with reduced accuracy

efcasado:
The ISO 8601 in its **"Representations with reduced accuracy"** section specifies that the `hh:mm` is a valid time representation.

The [DateTime.from_iso8601/1](https://hexdocs.pm/elixir/DateTime.html#from_iso8601/2) function is expected to **"Parse the extended 'Date and time of day' format described by ISO 8601:2004."**. However, this function fails to parse `ISO 8601` date and time strings if no seconds are provided (e.g. `2019-04-19T13:30Z`).

```elixir
iex(1)> DateTime.from_iso8601("2019-04-19T13:30Z")
{:error, :invalid_format}
```
This PR updates the `DateTime`, `Calendar.ISO` and `Time` modules to add support for ISO 8601 representations with reduced accuracy.

```elixir
iex(1)> DateTime.from_iso8601("2019-04-19T13:30Z")
{:ok, ~U[2019-04-19 13:30:00Z], 0}
```

Note that the `2019-04-19T13:30` string used to return an `invalid format` error, but the same date is now returning a `missing offset` error instead.

We came across this issue when we connected our Scala and Elixir services together. Our Scala service is using [LocalDateTime.toString()](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/time/LocalDateTime.html#toString()) to convert ISO8601 timestamps to strings.

> Outputs this date-time as a String, such as 2007-12-03T10:15:30.
> The output will be one of the following ISO-8601 formats:
>
> * uuuu-MM-dd'T'HH:mm
> * uuuu-MM-dd'T'HH:mm:ss
> * uuuu-MM-dd'T'HH:mm:ss.SSS
> * uuuu-MM-dd'T'HH:mm:ss.SSSSSS
> * uuuu-MM-dd'T'HH:mm:ss.SSSSSSSSS
> 
> The format used will be the shortest that outputs the full value of the time where the omitted parts  are implied to be zero.

### DIFF
--- a/lib/elixir/lib/calendar/datetime.ex
+++ b/lib/elixir/lib/calendar/datetime.ex
@@ -799,10 +799,11 @@ defmodule DateTime do
   defp raw_from_iso8601(string, calendar, is_year_negative) do
     with <<unquote(match_date), sep, unquote(match_time), rest::binary>> <- string,
          true <- unquote(guard_date) and sep in @sep and unquote(guard_time),
+         {second, rest} <- Calendar.ISO.parse_second(rest),
          {microsecond, rest} <- Calendar.ISO.parse_microsecond(rest),
          {offset, ""} <- Calendar.ISO.parse_offset(rest) do
       {year, month, day} = unquote(read_date)
-      {hour, minute, second} = unquote(read_time)
+      {hour, minute} = unquote(read_time)
       year = if is_year_negative, do: -year, else: year
 
       cond do
--- a/lib/elixir/lib/calendar/iso.ex
+++ b/lib/elixir/lib/calendar/iso.ex
@@ -63,13 +63,12 @@ defmodule Calendar.ISO do
   def __match_time__ do
     quote do
       [
-        <<h1, h2, ?:, i1, i2, ?:, s1, s2>>,
+        <<h1, h2, ?:, i1, i2>>,
         h1 >= ?0 and h1 <= ?9 and h2 >= ?0 and h2 <= ?9 and i1 >= ?0 and i1 <= ?9 and i2 >= ?0 and
-          i2 <= ?9 and s1 >= ?0 and s1 <= ?9 and s2 >= ?0 and s2 <= ?9,
+          i2 <= ?9,
         {
           (h1 - ?0) * 10 + (h2 - ?0),
-          (i1 - ?0) * 10 + (i2 - ?0),
-          (s1 - ?0) * 10 + (s2 - ?0)
+          (i1 - ?0) * 10 + (i2 - ?0)
         }
       ]
     end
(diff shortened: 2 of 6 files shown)


wojtekmach:
This is tricky. ISO 8601 reduced accuracy representation allows to skip any component, so not only 2019-01-01T09:00 is allowed but so is 2019-01-01T09 etc. The choice of just skipping seconds, while practical, is somewhat arbitrary.

A datetime/time with _no_ seconds and _zero_ seconds are arguably different, they convey different information, so automatically casting the former to the latter is dangerous.

When skipping seconds (or any date component for that matter), the idea might be that _any_ second would do, so a datetime with reduced accuracy might serve as a kind of interval. Shameless plug, I explore this in https://github.com/wojtekmach/calendar_interval. This distinction is already present with microseconds handling, `09:00:00`, `09:00:00.0`, `09:00:00.000000` are all distinct values.

FWIW, in Ecto 3 we are _casting_ time without seconds into time with zero seconds but thats just because casting is somewhat liberal and its practical to allow such there, but imho its not appropriate for stdlib (per my understanding of reduced accuracy)

efcasado:
> This is tricky. ISO 8601 reduced accuracy representation allows to skip any component, so not only 2019-01-01T09:00 is allowed but so is 2019-01-01T09 etc. The choice of just skipping seconds, while practical, is somewhat arbitrary.

Whilst `ISO 8601` reduced accuracy allows to skip various components of the local time segment, its extended format is more restrictive on what can be skipped. The example you shared (i.e. `2019-01-01T09`) wouldn't be a valid `Extended ISO 8601` timestamp. See section [4.2.2.3 (page 18)](http://www.loc.gov/standards/datetime/iso-tc154-wg5_n0038_iso_wd_8601-1_2016-02-16.pdf).

> a) A	specific	hour	and	minute
> Basic	format: hhmm Example: 2320
> Extended	format: hh:mm Example: 23:20
>
> b) A	specific	hour
> Basic	format: hh Example: 23
> Extended	format: not	applicable

> A datetime/time with no seconds and zero seconds are arguably different, they convey different information, so automatically casting the former to the latter is dangerous.
>
> When skipping seconds (or any date component for that matter), the idea might be that any second would do, so a datetime with reduced accuracy might serve as a kind of interval.

I tend to agree with you here when you talk about the difference in semantics. However, I deliberately chose to use `0` when no seconds are provided because I considered `0` to be the zero value for seconds. Using `0` when no seconds are provided allows us to keep things backwards compatible (to some extent). The accuracy of using `0` or any other value should be irrelevant given that the developer has explicitly chosen to give up on accuracy by not providing seconds in the timestamp.

Actually, I believe the current implementation is making similar assumptions when working with microseconds. The example below shows how we are assuming microseconds to be `0` when no microseconds are provided.

```elixir
iex(1)> {:ok, ts, _} = DateTime.from_iso8601("2019-04-19T13:30:00Z")
{:ok, ~U[2019-04-19 13:30:00Z], 0}
iex(2)> ts.microsecond
{0, 0}
```

josevalim:
Unfortunately the spec makes it clear that if the "seconds" part is missing, the representation is *not complete*. 13:00:00 and 13:00 are not equivalent and removing the seconds means a less accurate result.

In particular, the Java/Scala representation says:

> The format used will be the shortest that outputs the full value of the time where the omitted parts are implied to be zero.

But there is no where in ISO that specifies that to be the case and I think the Elixir implementation would be wrong (in regards to ISO) if we assume so.

@efcasado I am aware this is a very annoying situation for you because you are likely stuck between two places that won't change. The best we can do here in my opinion is to document we only parse complete specifications and we do not support reduced accuracy. There are other extensions of the ISO8601 standard that we don't support, such as decimal precision.

Other than this, thanks for the PR. The code is perfect but unfortunately we can't merge it for the reasons above.

PR #8977, opened 2019-04-23
Use :persistent_term for Logger config

### LINKED ISSUE #8592: Use persistent_term in Logger
We can load the module at the beginning of Logger.Config and use `:erlang.module_loaded(:persistent_term)` to choose between persistent_term or ETS.

josevalim:
Closes #8592.

### DIFF
--- a/lib/logger/lib/logger/config.ex
+++ b/lib/logger/lib/logger/config.ex
@@ -233,6 +233,8 @@ defmodule Logger.Config do
   #
   defp new_data do
     if Code.ensure_loaded?(:persistent_term) do
+      :persistent_term.put({@table, :log_data}, nil)
+      :persistent_term.put({@table, :translation_data}, nil)
       :persistent_term.put({@table, :deleted_handlers}, [])
     else
       entries = [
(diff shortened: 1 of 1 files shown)


lib/logger/lib/logger/config.ex
```diff
@@ -284,7 +305,19 @@ defmodule Logger.Config do
     end
   end
 
+  defp read_data(key) do
+    if :erlang.module_loaded(:persistent_term) do
```
michalmuskala: Have you checked if maybe `function_exported?` will be faster? The module will always be loaded if available since it's a preloaded module.

michalmuskala:
The general advise for `peristent_term` is to use a single key with a complex value instead of multiple keys. Storing all the settings with a record under one key might be a good solution here - this also might avoid constructing the tuple for the key lookup which should be beneficial.

josevalim:
@michalmuskala yeah, that's what I originally attempted but it makes it harder to work with both persistent term and ETS. In the future, I would migrate to a single key. I guess I can at least merge translation data and log data now.

josevalim:

PR #8978, opened 2019-04-24
Support :force_do_end_blocks in formatter

### LINKED ISSUE #8520: Add :force_do_end_blocks configuration to formatter
The :force_do_end_blocks configuration will convert any inline keyword syntax to the block format. Therefore:

    if something?, do: this, else: that

Will become:

    if something? do
      this
    else
      that
    end

## Rationale

Adding a new option may sound counter-intuitive to the formatter goals but the reason this option makes sense is because it makes the formatter more consistent and not less. When developers ask for formatter options, it is usually because they want to use different styles, while this option is about enforcing a particular style in a situation where we respect the user's choice today.

Furthermore, this option is convergent. Once you add it to the codebase and format it, removing the option will not make all of the code roll back to its previous state. So this option doesn't have the big downside in that flipping it on and off would cause large diffs in the codebase.

josevalim:
Closes #8520.

### DIFF
--- a/lib/elixir/lib/code.ex
+++ b/lib/elixir/lib/code.ex
@@ -264,6 +264,13 @@ defmodule Code do
       expects a valid `Version` which is usually the minimum Elixir
       version supported by the project.
 
+    * `:force_do_end_blocks` (since v1.9.0) - when `true`, converts all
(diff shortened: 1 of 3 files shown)


josevalim:

PR #8979, opened 2019-04-24
Configurable inspect function

wojtekmach:
Before this patch, `Kernel.inspect`, `IO.inspect` etc were using `Inspect.Algebra.to_doc/2` to build algebra document. This commit introduces an ability to swap out this function via `Inspect.Algebra.get_inspect_fun/0` and `Inspect.Algebra.put_inspect_fun/1`.

Let's see some use cases. First we configure Elixir to use a custom function:

   ```elixir
    iex> Inspect.Algebra.put_inspect_fun(&Pretty.to_doc/2)
    :ok
   ```

1. Pretty-printing integers (#4916)

     ```elixir
     iex> :math.pow(2, 32) |> trunc()
     4_294_967_296
     ```

2. Inspect tuples as records (#8534)

     ```elixir
     iex> {:foo, :bar}
     {:foo, :bar}

     iex> {:person, "Alice", 30}
    #person(name: "Alice", age: 30)
   ```

3. Overwrite struct's Inspect implementation

     ```elixir
     iex> URI.parse("https://elixir-lang.org")
     #URI<https://elixir-lang.org>
   ```

---

Code for the `Pretty.to_doc` function:

<details>

```elixir
    defmodule Pretty do
      def to_doc(integer, opts) when is_integer(integer) do
        PrettyIntegers.inspect(integer, opts)
      end

      def to_doc(tuple, opts) when is_tuple(tuple) do
        records = %{
          {:person, 2} => [:name, :age]
        }

        PrettyTuples.inspect(tuple, records, opts)
      end

      def to_doc(%URI{} = uri, _opts) do
        Inspect.Algebra.concat(["#URI<", URI.to_string(uri), ">"])
      end

      def to_doc(term, opts) do
        Inspect.Algebra.to_doc(term, opts)
      end
    end

    defmodule PrettyIntegers do
      def inspect(term, %Inspect.Opts{base: :decimal}) do
        pretty_decimal(term, "_")
      end

      def inspect(term, %Inspect.Opts{base: base}) do
        Integer.to_string(term, base_to_value(base))
        |> prepend_prefix(base)
      end

      def pretty_decimal(n, separator) when n < 0 do
        "-" <> pretty_decimal(abs(n), separator)
      end

      def pretty_decimal(n, separator) when n >= 1000 do
        left = div(n, 1_000)
        right = rem(n, 1_000)
        right_str = :io_lib.format("~3..0B", [right]) |> IO.iodata_to_binary
        pretty_decimal(left, separator) <> separator <> right_str
      end

      def pretty_decimal(n, _) do
        Integer.to_string(n)
      end

      defp base_to_value(base) do
        case base do
          :binary  -> 2
          :octal   -> 8
          :hex     -> 16
        end
      end

      defp prepend_prefix(value, base) do
        prefix = case base do
          :binary -> "0b"
          :octal  -> "0o"
          :hex    -> "0x"
        end
        prefix <> value
      end
    end

    defmodule PrettyTuples do
      import Inspect.Algebra

      def inspect({}, _records, opts), do: inspect_tuple([], opts)

      def inspect(tuple, records, opts) do
        list = Tuple.to_list(tuple)
        [record_name | values] = list

        case record_fields(records, record_name, tuple) do
          {:ok, fields} ->
            inspect_record(record_name, fields, values, opts)

          _ ->
            inspect_tuple(list, opts)
        end
      end

      defp record_fields(records, record_name, tuple) do
        fields_size = tuple_size(tuple) - 1
        Map.fetch(records, {record_name, fields_size})
      end

      defp inspect_tuple(list, opts) do
        inspect("{", list, "}", &to_doc/2, :flex, opts)
      end

      defp inspect_record(record_name, fields, values, opts) do
        kwlist = Enum.zip(fields, values)
        inspect("##{record_name}(", kwlist, ")", &Inspect.List.keyword/2, :strict, opts)
      end

      defp inspect(open, list, close, fun, break, opts) do
        open = color(open, :tuple, opts)
        sep = color(",", :tuple, opts)
        close = color(close, :tuple, opts)
        container_opts = [separator: sep, break: break]
        container_doc(open, list, close, opts, fun, container_opts)
      end
    end
```

</details>

### DIFF
--- a/lib/elixir/lib/inspect/algebra.ex
+++ b/lib/elixir/lib/inspect/algebra.ex
@@ -250,11 +250,27 @@ defmodule Inspect.Algebra do
 
   # Elixir + Inspect.Opts conveniences
 
+  @doc """
+  Gets the current inspect function.
+
+  See `put_inspect_fun/1` for more information.
+  """
+  @doc since: "1.9.0"
+  @spec get_inspect_fun() :: (any, Inspect.Opts.t() -> t)
   def get_inspect_fun() do
     # using :application instead of Application due to bootstrapping
     :application.get_env(:elixir, :inspect_fun, &to_doc/2)
   end
 
+  @doc """
+  Sets the current inspect function.
+
+  By default, functions like `Kernel.inspect/2`, `IO.inspect/2` etc
+  use `Inspect.Algebra.to_doc/2` to build an algebra document. This
+  function allows to use a custom function instead.
+  """
+  @doc since: "1.9.0"
+  @spec put_inspect_fun((any, Inspect.Opts.t() -> t)) :: :ok
   def put_inspect_fun(inspect_fun) when is_function(inspect_fun, 2) do
     :application.set_env(:elixir, :inspect_fun, inspect_fun)
   end
--- a/lib/elixir/lib/inspect/algebra.ex
+++ b/lib/elixir/lib/inspect/algebra.ex
@@ -250,6 +250,15 @@ defmodule Inspect.Algebra do
 
   # Elixir + Inspect.Opts conveniences
 
+  def get_inspect_fun() do
+    # using :application instead of Application due to bootstrapping
+    :application.get_env(:elixir, :inspect_fun, &to_doc/2)
+  end
+
+  def put_inspect_fun(inspect_fun) when is_function(inspect_fun, 2) do
+    :application.set_env(:elixir, :inspect_fun, inspect_fun)
+  end
+
   @doc """
   Converts an Elixir term to an algebra document
   according to the `Inspect` protocol.
--- a/lib/elixir/lib/inspect.ex
+++ b/lib/elixir/lib/inspect.ex
@@ -206,14 +206,14 @@ defimpl Inspect, for: List do
         container_doc(open, term, close, opts, &keyword/2, separator: sep, break: :strict)
 
       true ->
-        container_doc(open, term, close, opts, &to_doc/2, separator: sep)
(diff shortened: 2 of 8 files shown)


lib/elixir/lib/inspect/algebra.ex
```diff
@@ -250,6 +250,15 @@ defmodule Inspect.Algebra do
 
   # Elixir + Inspect.Opts conveniences
 
+  def get_inspect_fun() do
```
wojtekmach: let me know if we want to move forward with this approach and then I'll add docs etc.

kipcole9:
I really like this.  My use case is related to dates.  `Date.t` allows a form of polymorphism through the `:calendar` key but `Date`'s implementation of the `inspect` protocol only provides for `Calendar.ISO` (understandably).  I have briefly discussed with @josevalim the thought of having `Date` delegate `inspect/2` to calendar modules.  This approach seems better.

Would it make sense to return the current `inspect_fun` from `Inspect.Algebra.put_inspect_fun/1` so that it can be reverted if/when required?  Or alternatively `Inspect.Algebra.reset_inspect_fun/0`

wojtekmach:
@kipcole9 I don't see a problem with that, although I chose to return `:ok` to follow a `Calendar.put_time_zone_database/1`.

I think if users need the previous value, they'd call the `get_inspect_fun` before setting a new one, that could be a building block for some `with_inspect_fun/1` if you want to reset after calling some code (although I wouldn't put it in core)

wojtekmach:
Oops, this is not ready yet because it needs to recursively handle nested data.

wojtekmach:
Closing in favour of #8980.

PR #8982, opened 2019-04-25
Remove link from README to empty Windows wiki page

fertapric:
[ci skip]


josevalim:
I am bringing the page back; It should not have been removed and someone removed it without a reason why.

PR #8988, opened 2019-04-26
Replace bin/start by release/RELEASE_VSN/env

### LINKED ISSUE #8987: Place for setting environment variables in releases
Today, the place for setting environment variables in a release is inside `bin/start`. The issue with this approach is that, if you set something like `RELEASE_NODE` or `RELEASE_COOKIE` inside `bin/start`, and then you run a command such as `bin/release eval Foo` or `bin/release remote`, those environment variables won't be set, leading to weird behaviour or even a failure to connect in the second case.

This can be addressed in a couple ways. One option is to introduce a new script, `release/RELEASE_VSN/env.sh` or `release/RELEASE_VSN/env.bat`, that will be invoked early in the release process. The goal of this script is to set up environment variables.

This has the advantage of those variables being set per version, instead of being shared across all releases, which gives us more independence between versions.

By doing this change, we effectively remove the need for `bin/start`. It is important to note that hot code swapping requires `bin/start`, but those performing hot code swapping can add it manually.

josevalim:
Today, the place for setting environment variables in a release
is inside bin/start. The issue with this approach is that, if you
set something like RELEASE_NODE or RELEASE_COOKIE inside
bin/start, and then you run a command such as  `bin/release remote`,
we won't be able to connect to the running node.

This is addressed by introducing `release/RELEASE_VSN/env`
that will be invoked early in the release process. The goal of
this script is to set up environment variables.

This has the advantage of those variables being set per version,
instead of being shared across all releases, which gives us more
independence between versions.

By doing this change, we effectively remove the need for bin/start.
It is important to note that hot code swapping requires bin/start,
but those performing hot code swapping can add it manually.

Closes #8987.

### DIFF
--- a/lib/mix/lib/mix/tasks/release.ex
+++ b/lib/mix/lib/mix/tasks/release.ex
@@ -549,7 +549,7 @@ defmodule Mix.Tasks.Release do
         start_erl.data
       tmp/
 
-  ## Environment variablees
+  ## Environment variables
 
   The system sets different environment variables. The following variables
   are set early on and can only be read by `env.sh` and `env.bat`:
--- a/lib/mix/lib/mix/tasks/release.ex
+++ b/lib/mix/lib/mix/tasks/release.ex
@@ -1000,9 +1000,9 @@ defmodule Mix.Tasks.Release do
     Release created at #{path}!
 
         # To start your system
-        #{path}/bin/start
+        #{cmd} start
 
-    Check bin/start for more information. Once the release is running:
+    Once the release is running:
 
         # To connect to it remotely
         #{cmd} remote
@@ -1048,17 +1048,15 @@ defmodule Mix.Tasks.Release do
     File.mkdir_p!(bin_path)
 
     for os <- include_executables_for do
-      {start, start_fun, clis} = cli_for(os, release)
-      start_path = Path.join(bin_path, start)
-      start_template_path = Path.join("rel", start <> ".eex")
+      {env, env_fun, clis} = cli_for(os, release)
+      env_path = Path.join(release.version_path, env)
+      env_template_path = Path.join("rel", env <> ".eex")
 
-      if File.exists?(start_template_path) do
-        copy_template(start_template_path, start_path, [release: release], force: true)
+      if File.exists?(env_template_path) do
+        copy_template(env_template_path, env_path, [release: release], force: true)
       else
-        File.write!(start_path, start_fun.(release))
+        File.write!(env_path, env_fun.(release))
       end
-
-      executable!(start_path)
 
       for {filename, contents} <- clis do
         path = Path.join(bin_path, filename)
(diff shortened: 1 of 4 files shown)


josevalim:
This is good to go now.

lib/mix/test/mix/tasks/release.init_test.exs
```diff
@@ -7,21 +7,22 @@ defmodule Mix.Tasks.Release.InitTest do
...
       assert File.exists?("rel/vm.args.eex")
-      assert File.read!("rel/start.eex") =~ "<%= @release.name %>"
-      assert File.read!("rel/start.bat.eex") =~ "<%= @release.name %>"
+      assert File.exists?("rel/env.sh.eex")
+      assert File.exists?("rel/env.bat.eex")
     end)
   end
 
-  test "can be set to --force and --quiet" do
+  test "can be set to --ovewrite and --quiet" do
```
fertapric:
```suggestion
  test "can be set to --overwrite and --quiet" do
```

lib/mix/lib/mix/tasks/release.ex
```diff
@@ -550,32 +549,39 @@ defmodule Mix.Tasks.Release do
         start_erl.data
       tmp/
 
-  Furthermore, the system sets the following environment variables.
-  Many of them can be set, either directly in your production nodes
-  or inside the `bin/start` script:
+  ## Environment variablees
```
```suggestion
  ## Environment variables
```

josevalim: Thanks Apple.

lib/mix/lib/mix/tasks/release.init.ex
```diff
@@ -217,11 +219,9 @@ defmodule Mix.Tasks.Release.Init do
     """
 
   @doc false
-  def start_bat_text,
+  def env_bat_text,
```
fertapric: Why the `env.bat` does not contain the heart instructions? 🙂

josevalim: Heart is not available on Windows. Instead one would use Windows Services.

fertapric: Oh! I thought it was supported because the documentation says:

> To reboot on Windows, `HEART_COMMAND` can be set to `heart -shutdown` (included in the Erlang delivery) or to any other suitable program that can activate a reboot.

jesseshieh:
Looks great. Thanks for the heads up. I'm curious though, why "it is important to note that hot code swapping requires bin/start".

josevalim:
Somewhere in the erlang docs they say that if hot code swapping fails, it
may restart the system using HEART_COMMAND, falling back to bin/start. I
can try to find the exact reference later.
-- 


josevalim:
@jesseshieh here you go: http://erlang.org/doc/design_principles/release_handling.html#restart_new_emulator--low-level-

josevalim:

jesseshieh:
@josevalim ah I see. Thanks!

PR #8995, opened 2019-04-29
Fix columns in tokenize of dot_call_op

liskin:
`handle_dot` already gets the column of `(` from `strip_dot_space`.


josevalim:

PR #8997, opened 2019-04-29
Set ticktime and autohost on IEx --remsh

josevalim:
Thanks to @tsloughter for the feedback.


josevalim:

PR #8998, opened 2019-04-29
Use $RELEASE_NODE as the base for --remsh/--rpc

josevalim:
This guarantees that custom hostnames and similar
propagate properly.

Thanks @tsloughter for the tip.


josevalim:
@fertapric because those are atom names, i am trying to make them smaller, to make sure we won't hit the 255 chars limit.

josevalim:

PR #8999, opened 2019-04-29
Deprecate Enumerable keys in Map.{take|drop|split}

whatyouhide:
`Keyword.{take|drop|split}` don't offer support for an `Enumerable` of keys to take/drop/split based on. However, the `Map` counterparts do. With this PR, we hard-deprecate support for `Enumerable`s of keys in `Map.{take|drop|split}`. Only lists of keys are allowed.

### DIFF
--- a/lib/elixir/lib/map.ex
+++ b/lib/elixir/lib/map.ex
@@ -384,13 +384,20 @@ defmodule Map do
       %{a: 1, c: 3}
 
   """
-  @spec take(map, Enumerable.t()) :: map
+  @spec take(map, [key]) :: map
   def take(map, keys)
 
+  def take(map, keys) when is_map(map) and is_list(keys) do
+    take(keys, map, _acc = [])
+  end
+
   def take(map, keys) when is_map(map) do
-    keys
-    |> Enum.to_list()
-    |> take(map, [])
+    IO.warn(
+      "Map.take/2 with an Enumerable of keys that is not a list is deprecated. " <>
+        " Use a list of keys instead."
+    )
+
+    take(map, Enum.to_list(keys))
   end
 
   def take(non_map, _keys) do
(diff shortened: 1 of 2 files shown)


lib/elixir/lib/map.ex
```diff
@@ -384,13 +384,20 @@ defmodule Map do
       %{a: 1, c: 3}
 
   """
-  @spec take(map, Enumerable.t()) :: map
+  @spec take(map, [key]) :: map
   def take(map, keys)
 
+  def take(map, keys) when is_map(map) and is_list(keys) do
+    take(keys, map, _acc = [])
```
eksperimental: ❤️ loved the way you annotated the variable `_acc = []`

lib/elixir/lib/map.ex
```diff
@@ -384,13 +384,20 @@ defmodule Map do
       %{a: 1, c: 3}
 
   """
-  @spec take(map, Enumerable.t()) :: map
+  @spec take(map, [key]) :: map
   def take(map, keys)
```
eksperimental:
```suggestion
```
the head definition can be removed here. and same with the other functions.

Is it desirable to have that in the Elixir code.
It reads better. I'm wondering if new code should be like that.

whatyouhide: I disagree on this. I think when you have multiple clauses with different argument names, it's definitely better to have a bodiless clause. This allows you to see the function signature at a glance and ensures the correct name of the arguments in the documentation, instead of relying on the first clause. Many times I've seen the first clause get changed and screw up with the arguments in the docs because one forgets about updating that.

eksperimental: Sorry, I expressed it wrong. I was saying it read better with the bodiless clause, but i think we have been using mostly without thia clause. Should we start relying on it?

whatyouhide: It's a case by case situation. I don't want to make it a "rule" in any way. If there are a few clauses and the argument names change between the clauses, I tend to go with a bodiless clause.

PR #9002, opened 2019-04-30
Fix some typos in Mix.Tasks.Release

joladev:
Noticed some typos in the new release documentation, plus a link that wasn't formatted as one.

Really good documentation! I especially liked the part about "Why releases", since it's such a common question.


whatyouhide:
Thanks @joladev! 💟

PR #9004, opened 2019-04-30
Update docs to make them easier to read in html format.

pdgonzalez872:
I'm a bit torn with this one. One one hand, I don't want to set the precedent of `style` changes. On the other, the first thing I thought when reading the docs [here](https://hexdocs.pm/mix/master/Mix.Tasks.Release.html) was that there was some type of emphasis missing.

I think the `LD_LIBRARY_PATH ` change was indeed a typo (since there are other instances of it using ``) but I'm torn about the other changes I made.

Anyways, here are some changes, if you'd like, please use them! It not, no hard feelings. ❤️ 💜 

(If you'd like any changes, please let me know and I'll make them)


whatyouhide:
Thanks @pdgonzalez872! 💟

PR #9009, opened 2019-05-02
Unify error handling and compiler deadlock checking

josevalim:
This makes it so we also include the `function/arity` in the
warnings stacktraces and makes the compiler more deterministic.

### DIFF
--- a/lib/elixir/lib/kernel/parallel_compiler.ex
+++ b/lib/elixir/lib/kernel/parallel_compiler.ex
@@ -133,7 +133,7 @@ defmodule Kernel.ParallelCompiler do
     schedulers = max(:erlang.system_info(:schedulers_online), 2)
 
     result =
-      spawn_workers(files, 0, [], [], [], [], %{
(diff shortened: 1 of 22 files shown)


josevalim:

lib/elixir/lib/kernel/parallel_compiler.ex
```diff
@@ -363,13 +338,12 @@ defmodule Kernel.ParallelCompiler do
         Process.monitor(process)
         wait_for_messages(queue, spawned + 1, waiting, files, result, warnings, state)
 
-      {:struct_available, module} ->
+      {:available, kind, module} ->
```
fertapric: Minor: why `:struct_available` is changed to `{:available, :struct}` but `:module_available` not to `{:available, :module}`? 🙂

josevalim: The module_available passes a bunch of extra information which makes it unique. The :struct mechanism though is simpler and it has been generalized for other use cases. I will effectively use it in an upcoming PR. :)

PR #9010, opened 2019-05-02
Add example of using a global named Agent to docs.

pdgonzalez872:
Hi!

I played around with the `global` option when dealing with Agents recently.
Unfortunately there is not a lot of documentation/literature on this topic at
the moment. I think there is a lot of value in exploring this technique
especially when pairing it with something like
[Horde](https://github.com/derekkraan/horde) for distributed supervision.
I believe this is a step in making Docker and OTP be friends, by leveraging
both technologies and having an output greater than each of them alone.

As usual, if you think this adds value, feel free to use it. If you want
changes, please let me know. If it doesn't add value, no hard feelings.

Thanks!

Hope this can shed some light into some cool tools we have available within
 the BEAM.

### DIFF
--- a/lib/elixir/lib/agent.ex
+++ b/lib/elixir/lib/agent.ex
@@ -160,6 +160,47 @@ defmodule Agent do
   The best solution is to simply use the explicit module, function, and arguments
   APIs when working with distributed agents.
 
+  ## An example using named `global` Agents:
+
+  You can create named Agents that can be called from any connected node in a
+  cluster. This technique can be helpful when you have a cluster, don't
+  necessarily know which node will handle the work and you want to ensure
+  consistency between nodes by avoiding duplication between nodes. If you
+  don't, you can run into a situation where a you have multiple Agents with
+  different states in different nodes.
+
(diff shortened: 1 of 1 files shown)


josevalim:
Hi @pdgonzalez872, this is a great example but I also think it is too much for the Agent docs and it belongs more to something like a guide. For example, the guides on the website do talk about Distributed Erlang and how to connect nodes. Thanks for the PR!

pdgonzalez872:
No worries @josevalim. I had that feeling as well but thought this could start a conversation. Yes, we do have a Distributed Erlang section in the guides, but we don't show how GenServers/Agents can be used in that setting. Do you think it would add value to add it there? Same comment as above, no hard feelings.

josevalim:
We do cross node requests in there and use the naming mechanism, but not the global registry per se. I will do a pass on those docs for Elixir v1.9 and add things about releases, I will make sure to include some extra references.

pdgonzalez872:
That'd be great. For some reason, I'm very excited about the global registry. At this time, I'm not sure why, but I'm happy! Maybe it will pass. Thanks @josevalim as usual!

PR #9013, opened 2019-05-03
Add example for Path.dirname/1

sn3p:
Found this out while debugging, thought it would be nice to mention this in the docs.


josevalim:

eksperimental:
Interesting.
It should be documented IMO.

PR #9015, opened 2019-05-05
Document Inspect.inspect/2

wojtekmach:


josevalim:
Actually, we should add a note  that this function shouldn't be invoked directly. If you  want the result of inspect, one should prefer Inspect.Algebra.to_doc, since that one will handle Inspect.Opts properly.

wojtekmach:
Good point. I guess `opts` should be `Inspect.Opts.t`, not `keyword`, too, right? I'll make both changes shortly.

josevalim:
Yes, indeed.

>

josevalim:

PR #9018, opened 2019-05-06
Use parens in IO.puts/1 calls around the codebase

whatyouhide:
There were instances of `IO.puts/1` calls without parens throughout the whole codebase (in docs, code, tests). I removed most of them where it made sense to do so, so that users will find documentation and other things consistent with how the formatter behaves.


eksperimental:
Why doesn't the formatter catch those in the code?

wojtekmach:
These occurrences all happened in comments and other strings, formatter doesn’t work on them.

eksperimental:
I saw them in my phone and they looked like code, but you are right, they are code lines in comments

PR #9020, opened 2019-05-06
Translate process crash on node

mitchellhenke:
Background on this is that a user of the `sentry` library was seeing a difference in behaviour of reported process crashes when running an application as a Distillery release vs. `iex -S mix` (https://github.com/getsentry/sentry-elixir/issues/337)

Sentry implements a Logger backend, but wasn't receiving events with `:crash_reason` metadata for the process crash running a Distillery release.  After some digging and reaching some of the limits of my current understanding, it appears that Erlang prints a different message format with erts, as seen here?
https://github.com/erlang/otp/blob/master/erts/emulator/beam/beam_emu.c#L1591-L1594

```c
int alive = erts_is_alive;
erts_dsprintf_buf_t *dsbufp = erts_create_logger_dsbuf();

erts_dsprintf(dsbufp, "Error in process ~p ");
if (alive)
    erts_dsprintf(dsbufp, "on node ~p ");
erts_dsprintf(dsbufp, "with exit value:~n~p~n");
```

I think ultimately it leads to not matching in the `Logger.Translator`, since it ends up with an extra formatted argument, and a format of:

```erlang
 "Error in process ~p on node ~p with exit value:~n~p~n"
```

I tested my patched version and then things behaved as expected, but I was not sure on the best way to test with the erts stuff involved.

### DIFF
--- a/lib/logger/lib/logger/translator.ex
+++ b/lib/logger/lib/logger/translator.ex
@@ -126,6 +126,19 @@ defmodule Logger.Translator do
 
         {:ok, msg, metadata}
 
+      {'Error in process ' ++ _, [pid, node, {reason, stack}]} ->
+        reason = Exception.normalize(:error, reason, stack)
+
+        msg = [
+          "Process ",
+          inspect(pid),
+          " on node ",
+          inspect(node),
+          " raised an exception" | format(:error, reason, stack)
+        ]
+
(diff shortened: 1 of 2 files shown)


whatyouhide:
Thanks @mitchellhenke! 💟

PR #9024, opened 2019-05-07
Allow developers to choose between sname and name in releases

josevalim:
We default to --sname which provides safer security defaults.

/cc @tsloughter @wojtekmach

### DIFF
--- a/lib/mix/lib/mix/tasks/release.ex
+++ b/lib/mix/lib/mix/tasks/release.ex
@@ -626,8 +626,8 @@ defmodule Mix.Tasks.Release do
       Using `name` (long names) or `sname` (short names). Defaults to
       `sname` which allows access only within the current system.
       `name` allows external connections. If `name` is used and you are
-      running on a version earlier than Erlang/OTP 22+, you must set
-      `RELEASE_NODE` to `RELEASE_NAME@127.0.0.1` or another appropriate host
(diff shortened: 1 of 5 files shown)


lib/mix/lib/mix/tasks/release.ex
```diff
@@ -622,6 +622,13 @@ defmodule Mix.Tasks.Release do
       files to. It can be set to a custom directory. It defaults to
       `$RELEASE_ROOT/tmp`
 
+    * `RELEASE_DISTRIBUTION` - how do we want to run the distribution.
+      Using `name` (long names) or `sname` (short names). Defaults to
+      `sname` which allows access only within the current system.
+      `name` allows external connections. If `name` is used and you are
```
tsloughter: Might be worth making it clear that `name` requires a hostname that resolves through DNS or an ip. While short name you can pretty much give anything, as long as it doesn't have a `.` in it.

lib/mix/lib/mix/tasks/release.ex
```diff
@@ -622,6 +622,13 @@ defmodule Mix.Tasks.Release do
...
+      running on a version earlier than Erlang/OTP 22+, you must set
```
whatyouhide: Nitpick, but I think either we say "on a version earlier than Erlang/OTP 22", or we say "on Erlang/OTP 22+" (my preference is the former).

josevalim:

PR #9025, opened 2019-05-07
Improved archive.uninstall to find any version of specified file

TheFirstAvenger:
Added logic to search for `archive-name-*` if `archive-name` is not found when running `mix archive.uninstall archive-name`.
For example, `mix archive.uninstall ironman` will find `ironman-0.4.2` archive.
Still checks for exact match first, so only modifies cases where legacy would report "no archives found".

### DIFF
--- a/lib/mix/lib/mix/local/installer.ex
+++ b/lib/mix/lib/mix/local/installer.ex
@@ -244,7 +244,7 @@ defmodule Mix.Local.Installer do
       found =
         cond do
           File.exists?(Path.join(root, name)) -> Path.join(root, name)
-          matching = matching_package(root, name) -> Path.join(root, matching)
+          matching = matching_package(root, name) -> matching
           true -> nil
         end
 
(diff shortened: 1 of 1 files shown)


lib/mix/lib/mix/local/installer.ex
```diff
@@ -241,26 +241,45 @@ defmodule Mix.Local.Installer do
...
+          Mix.shell().error("Could not find a local artifact named #{inspect(name)}. We found:")
+          Mix.Task.rerun(listing)
           nil
       end
     else
       Mix.raise("No argument was given to uninstall command")
     end
   end
 
+  defp matching_package(root, name) do
```
josevalim: I think you can use Path.wildcard here. Then you can rewrite this function to: `Path.wildcard(Path.join(root, name <> "-*") |> List.first()`

josevalim:
Looks great! I dropped a comment and then we just need a test and we should be good to go!

lib/mix/lib/mix/local/installer.ex
TheFirstAvenger: 👍 done

TheFirstAvenger:
Ok I reworked the matching_package function as suggested. There isn't currently a test for `Installer.uninstall`. Since it touches the file system multiple times, what strategy would you suggest for testing it?

josevalim:
You can add a new test to mix/tasks/archive_test. Here is a sample test you can use and minimize: https://github.com/elixir-lang/elixir/blob/master/lib/mix/test/mix/tasks/archive_test.exs#L150

josevalim:
I want to slightly refactor that test file, so I merged it and I will add the test as I refactor. :) Thanks for the PR!  :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

PR #9029, opened 2019-05-08
Make sure Logger v1.8 does not get stuck in discard mode

### LINKED ISSUE #9028: Logger never gets out of :discard mode
### Environment

Erlang 21
Elixir 1.8.1

### Current behavior

When the message queue size exceeds threshold Logger enters into a discard mode with a message like that:
```
Logger has 661 messages in its queue, which is above :discard_threshold. Messages will be discarded until the message queue goes back to 75% of the threshold size
```

From that moment on all log messages are discarded until the app is restarted

```
> pid = :erlang.whereis(Logger)           
> Process.info(pid, :message_queue_len)   

{:message_queue_len, 0}

> :sys.get_state(pid)                  
[
  {LoggerFileBackend, :file,
   %{...}},
  {Logger.Config, false,
   {%{
      level: :debug,
      mode: :discard,
      translators: [{Logger.Translator, :translate}],
      truncate: 8096,
      utc_log: false
    },
    %{
      async_threshold: 15,
      discard_threshold: 500,
      keep_threshold: 375,
      sync_threshold: 20
    }}}
]

> require Logger
> Logger.info("test")

# nothing is logged
# manually change logger mode

> st =  :ets.lookup_element(Logger.Config, :data, 2)
%{
level: :debug,
mode: :discard,
translators: [{Logger.Translator, :translate}],
truncate: 8096,
utc_log: false
}

> st = %{st | mode: :async}
> :ets.update_element(Logger.Config, :data, {2, st})

# now logging resumes
```

I believe I found what the problem is.
Logger's mode is updated when a new event arrives here:
https://github.com/elixir-lang/elixir/blob/v1.8/lib/logger/lib/logger/config.ex#L78-L102

But when Logger enters the `:discard` mode, all log messages are dropped and no event get generated:
https://github.com/elixir-lang/elixir/blob/v1.8/lib/logger/lib/logger.ex#L672-L678

So it seems like entering a `:discard` mode Logger can't get out of it and all the logs are dropped.

### Expected behavior

After the Logger's message queue get's under the discard threshold Logger's mode should be updated appropriately.

josevalim:
Closes #9028.

### DIFF
--- a/lib/logger/lib/logger.ex
+++ b/lib/logger/lib/logger.ex
@@ -555,6 +555,7 @@ defmodule Logger do
     :level,
     :utc_log,
     :discard_threshold,
+    :discard_threshold_periodic_check,
     :translator_inspect_opts
   ]
   @spec configure(keyword) :: :ok
--- a/lib/logger/lib/logger/config.ex
+++ b/lib/logger/lib/logger/config.ex
@@ -4,6 +4,7 @@ defmodule Logger.Config do
   @behaviour :gen_event
   @name __MODULE__
   @table __MODULE__
+  @check_discard {__MODULE__, :check_discard}
 
   def start_link do
     GenServer.start_link(__MODULE__, :ok, name: @name)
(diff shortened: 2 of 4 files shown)


lib/logger/lib/logger.ex
```diff
@@ -555,6 +555,7 @@ defmodule Logger do
     :level,
     :utc_log,
     :discard_threshold,
+    :discard_threshold_periodic_check,
```
fertapric: Should this option be documented at the module documentation?

josevalim: I don't want to document it for now, because I don't think it actually has any use. Especially since it is v1.8 only. Our default is likely good enough in all cases.

fertapric: Oh! Ok. I mentioned it because it’s documented in `master`

https://github.com/elixir-lang/elixir/blob/dc154f5db7a54173decd00f7971e9d8d41aa59af/lib/logger/lib/logger.ex#L152

PR #9031, opened 2019-05-09
Make it possible to disable regex precompilation.

hairyhum:
## Current state of things

By default regex strings are compiled on code compilation stage.
This makes it harder to produce packages which would work with different
erlang versions and on different platforms.
Also the runtime prints a warning message, when starting, if code is running
on a system with different endianness. This can make users anxious.

This can be solved on the code level by recompiling regexes in runtime, but
that requires code hygiene, which may be harder to maintain and also does not
protect from regexes in dependencies.

## Proposal

This pull requests proposes to have a build configuration, which will
disable regex precompilation by having a plain binary in the `re_pattern` field of regex structure.

Binary works well with `re` functions, all except `inspect`, so there is not
that much changes required.

## Open questions

I wonder if there is a way to mark the code package containing precompiled
regexes to print or suppress the endiannes warning?

I specify the setting in `Mix.Project.config`, but this may not be the best place to do that. Is there a way to pass build configuration to dependencies?

I'm fully open to any corrections and requests regarding the code, as long as the solution works.

TODO:

Tests and docs.

### DIFF
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -4995,13 +4995,26 @@ defmodule Kernel do
 
   defmacro sigil_r({:<<>>, _meta, [string]}, options) when is_binary(string) do
     binary = :elixir_interpolation.unescape_chars(string, &Regex.unescape_map/1)
-    regex = Regex.compile!(binary, :binary.list_to_bin(options))
+    bin_options = :binary.list_to_bin(options)
+
+    regex = case Mix.Project.config()[:plain_regex] do
+      true ->
+        Regex.compile_plain!(binary, bin_options)
+      _ ->
+        Regex.compile!(binary, bin_options)
+    end
     Macro.escape(regex)
   end
 
   defmacro sigil_r({:<<>>, meta, pieces}, options) do
     binary = {:<<>>, meta, unescape_tokens(pieces, &Regex.unescape_map/1)}
-    quote(do: Regex.compile!(unquote(binary), unquote(:binary.list_to_bin(options))))
+    bin_options = :binary.list_to_bin(options)
+    case Mix.Project.config()[:plain_regex] do
+      true ->
+        quote(do: Regex.compile_plain!(unquote(binary), unquote(bin_options)))
+      _ ->
+        quote(do: Regex.compile!(unquote(binary), unquote(bin_options)))
+    end
   end
 
   @doc ~S"""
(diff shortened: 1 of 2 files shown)


josevalim:
Hi @hairyhum, I agree this is a problem.

Also note that we can't invoke `Mix` inside `kernel`, as the Elixir application does not depend on Mix. You can use Elixir without Mix, and that would make it fail (you can see CI fails).

I am not sure a Mix or project compilation would work in this case either because Elixir itself may have regular expressions and those will only be fixed if you *recompile* Elixir (and setting a flag won't recompile it).

There are also other issues with precompiled packages, such as supporting different Erlang versions and so on.

For all of those reasons, we have generally stepped away from precompiled packages and we actually precompile them locally via `mix archive.install` and `mix escript.install`. So can you please tell us a little bit more about your workflow?

hairyhum:
Hi @josevalim, 
Our main issue is that we in RabbitMQ distribute built packages (our CLI escript), which may run on different systems with different endianness. And those packages should support multiple versions of Erlang as well. We build them on the oldest supported (currently 20.3).

We currently try to be careful with regexes and recompile them in runtime, but this means that we have to carefully review all the code and remember to check regexes all the time, which is not that productive in the long run.

Also we're not sure about dependencies. We don't have that many dependencies and they should not contain regexes, but that's another thing to check and keep track of.

We would prefer to have a build flag of some sort, which we can use to build packages (escripts) which do not contain precompiled regexes.

josevalim:
Thanks @hairyhum!

So there are two things to be done:

1. allow the endianess check to be skipped. This needs to be a applpication environomenet configuration flag for Elixir, it is totally doable.

2. guarantee all of your regexes are properly precompiled. For example, at some point we wrapped all regexes in the Elixir codebase in a Regex.recompile call. I am not sure we can automate this somehow but one option is to have a Credo lint check that checks for this? So at least you are confident it is not falling through the cracks?

hairyhum:
That' basically what we have right now. And what we would like to avoid doing by using the build flag.

josevalim:
I have no idea how said build flag can be implemented though. So any suggestions are welcome. We can't depend on Mix from Kernel.

josevalim:
To clarify, I have some ideas, but none that would cleanly fit with the mechanisms that we have today. So my suggestion is to:

* Send a PR for 1) above
* Perhaps move the discussion about 2) to the core mailing list to see if other can chip-in alternatives

josevalim:
Thanks @hairyhum, I am closing this for now. Could you please send a PR for 1?

michaelklishin:
FTR, here's the [elixir-lang-core thread](https://groups.google.com/forum/#!topic/elixir-lang-core/CgFdxIONvGg).

Issue #9030, opened 2019-05-09
Do not allow `defmodule true | false`


rodrigues:
### Environment

* Elixir & Erlang/OTP versions (elixir --version): Elixir 1.8, OTP 21
* Operating system: Mac OS Mojave

### Current behavior

Please close this issue if it doesn't make sense, it just seems to me this could be improved at the language level. 👍🏼

As [this tweet](https://twitter.com/gregvaughn/status/1126219559441244160) points out, you can call `defmodule` passing `true` or `false` as a module name:

```elixir
defmodule false, do: def foo, do: "bar"
> false.foo
"bar"
```

The point is not about `$handle_undefined_function` – an erlang feature mentioned in the tweet but beyond the scope of this ticket – but about allowing these module names to be defined.

As much as `true` and `false` are just atoms in the end of the day, I think `defmodule` would be a bit more robust if it doesn't allow these to be defined as module names, just as it does with `nil`.

### Expected behavior

The same when you call `defmodule nil`, fail:

```
iex(5)> defmodule nil, do: def a, do: 1
** (ArgumentError) cannot invoke def/2 outside module
    (elixir) lib/kernel.ex:5362: Kernel.assert_module_scope/3
    (elixir) lib/kernel.ex:4101: Kernel.define/4
    (elixir) expanding macro: Kernel.def/2
    iex:5: (file)
```

whatyouhide:
The way `defmodule nil` fails is also terrible. If we want to forbid special atoms to be module names I think we should do it by returning a nicer error.

josevalim:
Agree, we should forbid those three and raise nice error messages.

whatyouhide:
@rodrigues do you want to work on a PR that adds this? :)

rodrigues:
@whatyouhide yes sirrr, just don't know if I can do it, hehehe, gonna take a look at it tonight 👍🏼

whatyouhide:
@rodrigues of course, no rush. Please feel free to ping us here or on IRC or the Elixir Slack if you need some help or guidance. :)

josevalim:
Closing in favor of the PR.

PR #9032, opened 2019-05-09
Do not allow defmodule with special atoms

rodrigues:
Follow up on #9030

```elixir
 defmodule true do end
** (CompileError) iex:1: invalid module name: true
    (stdlib) erl_eval.erl:680: :erl_eval.do_apply/6
    (iex) lib/iex/evaluator.ex:257: IEx.Evaluator.handle_eval/5
    (iex) lib/iex/evaluator.ex:237: IEx.Evaluator.do_eval/3
    (iex) lib/iex/evaluator.ex:215: IEx.Evaluator.eval/3
    (iex) lib/iex/evaluator.ex:103: IEx.Evaluator.loop/1
    (iex) lib/iex/evaluator.ex:27: IEx.Evaluator.init/4
```

### DIFF
--- a/lib/elixir/src/elixir_module.erl
+++ b/lib/elixir/src/elixir_module.erl
@@ -54,6 +54,8 @@ next_counter(Module) ->
 
 %% Compilation hook
 
+compile(Module, _Block, _Vars, #{line := Line, file := File}) when Module == nil; is_boolean(Module) ->
+  elixir_errors:form_error([{line, Line}], File, ?MODULE, {invalid_module, Module});
 compile(Module, Block, Vars, #{line := Line} = Env) when is_atom(Module) ->
   %% In case we are generating a module from inside a function,
   %% we get rid of the lexical tracker information as, at this
--- a/lib/elixir/test/elixir/kernel_test.exs
+++ b/lib/elixir/test/elixir/kernel_test.exs
@@ -627,6 +627,20 @@ defmodule KernelTest do
     end
   end
 
+  describe "defmodule" do
+    test "does not accept special atoms as module names" do
+      special_atoms = [nil, true, false]
+
+      Enum.each(special_atoms, fn special_atom ->
+        msg = ~r"invalid module name: #{inspect(special_atom)}"
+
+        assert_raise CompileError, msg, fn ->
+          defmodule special_atom, do: :ok
+        end
+      end)
+    end
+  end
+
   describe "access" do
     defmodule StructAccess do
       defstruct [:foo, :bar]


whatyouhide:
I think a better place to do this would be https://github.com/elixir-lang/elixir/blob/0229358f882519362d7e9e831514f2a4a8ad2a40/lib/elixir/lib/kernel.ex#L3871, in the `defmodule` macro.

lib/elixir/src/elixir_module.erl
```diff
@@ -54,6 +54,8 @@ next_counter(Module) ->
 
 %% Compilation hook
 
+compile(Module, _Block, _Vars, #{line := Line, file := File}) when Module == nil; is_boolean(Module) ->
+  elixir_errors:form_error([{line, Line}], File, ?MODULE, {invalid_module, Module});
```
josevalim: Beautiful!

lib/elixir/test/erlang/module_test.erl
```diff
@@ -100,6 +100,11 @@ nested_ref_test() ->
 module_with_elixir_as_a_name_test() ->
   ?assertError(#{'__struct__' := 'Elixir.CompileError'}, eval("defmodule Elixir do\nend")).
 
+module_with_special_atom_as_a_name_test() ->
```
josevalim: We rarely add new tests to the Erlang suite. Can you please add those to lib/elixir/test/elixir/kernel_test.exs instead? Thanks!

rodrigues: Moved it to [kernel_test.exs](https://github.com/elixir-lang/elixir/pull/9032/files#diff-8899a6e958f0940b9f6b6e83ca6184ddR630)

lib/elixir/src/elixir_module.erl
rodrigues: @josevalim would you prefer it there, or in `kernel.ex` as @whatyouhide had suggested?

josevalim: I prefer here, you win @rodrigues! Sorry @whatyouhide. :P

Seriously though, where you put also forbids `Module.create(true, ...)`. Where @whatyouhide suggested, it wouldn't.

lib/elixir/test/elixir/kernel_test.exs
```diff
@@ -627,6 +627,25 @@ defmodule KernelTest do
...
 
+  describe "defmodule" do
+    test "does not accept special atoms as module names" do
+      special_atoms = [nil, true, false]
+
+      Enum.each(special_atoms, fn special_atom ->
+        msg = "nofile:1: invalid module name: #{inspect(special_atom)}"
+
+        assert_raise CompileError, msg, fn ->
+          Code.eval_quoted(
```
josevalim: You don't need eval_quoted! I think you can do this:

```
assert_raise CompileError, msg, fn ->
  defmodule special_atom, do: :ok
end
```

lib/elixir/src/elixir_module.erl
whatyouhide: Ah, didn't think of `Module.create/3`, good call.

fertapric:
Thanks @rodrigues! 💜

PR #9033, opened 2019-05-10
Add String.replace/3 with fun

gkpacker:
When the `pattern` is a function, it will return a new string, changing all matching ocurrencies to the `value`

```elixir
def replace(subject, value, replacement) when is_function(replacement) do
  {_status, pattern} =
    subject
    |> String.graphemes
    |> Enum.filter(replacement)
    |> Enum.join("|")
    |> Regex.compile

  replace(subject, pattern, value)
end
```

It usage would be something like this:
```elixir
iex> String.replace("ELIXIR", " ", fn letter -> letter == "I" end)
"EL X R"
```

### DIFF
--- a/lib/elixir/lib/string.ex
+++ b/lib/elixir/lib/string.ex
@@ -1278,7 +1278,7 @@ defmodule String do
   Returns a new string created by replacing occurrences of `pattern` in
   `subject` with `replacement`.
 
-  The `pattern` may be a string, a regular expression, or a compiled pattern.
+  The `pattern` may be a string, a regular expression, a compiled pattern or a function.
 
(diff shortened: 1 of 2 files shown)


josevalim:
Hi @gkpacker, thank you for the PR!

The goal of supporting functions is for the function to return the value that is being replaced once `value` is found. So something like `String.replace("foo", "o", fn _ -> "a" end) == "faa"`. I am sorry for not expanding much on it but the reason I didn't expand is because I am already working on it - which is why it is assigned to me. I will make it clearer though to avoid further confusion. :)

So this code looks good but not quite what we are expecting. Thanks again!

PR #9036, opened 2019-05-10
Add `--app` opt to `mix test`

TheFirstAvenger:
When included, only runs tests for specified apps.
Supports multiple entries.
Modeled after `mix cmd --app`.

Open to suggestions for how to test this functionality.

### DIFF
--- a/lib/mix/lib/mix/tasks/test.ex
+++ b/lib/mix/lib/mix/tasks/test.ex
@@ -148,6 +148,8 @@ defmodule Mix.Tasks.Test do
 
   ## Command line options
 
+    * `--app` - runs test only on specified app(s). May specify multiple apps with
+      multiple `--app` flags.
     * `--color` - enables color in the output
     * `--cover` - runs coverage tool. See "Coverage" section below
     * `--exclude` - excludes tests that match the filter
@@ -296,6 +298,7 @@ defmodule Mix.Tasks.Test do
   """
 
   @switches [
+    app: :keep,
     force: :boolean,
     color: :boolean,
     cover: :boolean,
@@ -326,7 +329,16 @@ defmodule Mix.Tasks.Test do
   @impl true
   def run(args) do
     {opts, files} = OptionParser.parse!(args, strict: @switches)
+    apps = Keyword.get_values(opts, :app)
 
+    if apps == [] or Atom.to_string(Mix.Project.config()[:app]) in apps do
+      run(args, opts, files)
+    else
+      :ok
+    end
+  end
+
+  defp run(args, opts, files) do
     if opts[:listen_on_stdin] do
       System.at_exit(fn _ ->
         IO.gets(:stdio, "")
--- a/lib/mix/test/fixtures/umbrella_app/apps/bar/lib/bar.ex
+++ b/lib/mix/test/fixtures/umbrella_app/apps/bar/lib/bar.ex
@@ -1,0 +1,13 @@
+defmodule Bar do
+  def hello do
+    :world
+  end
+end
+
+defprotocol Bar.Protocol do
+  def to_uppercase(string)
+end
+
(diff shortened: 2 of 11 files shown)


josevalim:
So today you can do `mix cmd --app foo mix test`, it is a handful, but you can provide aliases in your `mix.exs` for convenience. In general, I would just like to avoid adding `--app` flags to many commands, as it is an orthogonal concern. In any case, thanks for the PR!

TheFirstAvenger:
Currently when you do `mix cmd --app`, you lose the output coloring of mix test. Is there another approach that could retain that helpful information?

TheFirstAvenger:
Alternatively, is it possible to make `--app` something that is handled higher up in the mix stack? Such that it would be available on all `@recursive` calls by default and therefore not need to be added to each task?

josevalim:
t's a good point. There are work-arounds but this should maybe be made more convenient. Another concern with running from the umbrella root is that paths in test reports are relative to the root and not the umbrella and that can look a little awkward.

> Alternatively, is it possible to make --app something that is handled higher up in the mix stack? Such that it would be available on all @recursive calls by default and therefore not need to be added to each task?

Not as a flag but it would be possible as an environment variable.

lib/mix/lib/mix/tasks/test.ex
```diff
@@ -148,6 +148,8 @@ defmodule Mix.Tasks.Test do
 
   ## Command line options
 
+    * `--app` - runs test only on specified app(s). Must be first option(s) given.
+      May specify multiple apps with multiple `--app` flags.
```
josevalim: It doesn't need to be the first option given. You can first call `OptionParser.parse!(args, strict: @switches)` and then handle apps accordingly.

josevalim: There is a `:keep` option in OptionParser that would allow the app flag to be given multiple times.

TheFirstAvenger:
>Another concern with running from the umbrella root is that paths in test reports are relative to the root and not the umbrella and that can look a little awkward.

This is another issue that I have on my todo list to look at addressing. We always run mix test from the umbrella root, and it breaks cmd-clicking on the test output. If tests could recognize that they were recursively being called and tack on the extra `apps/app_name/` it would be super helpful.

josevalim:
For testing, it would be added to test_test.exs, you can see an example using umbrella here: https://github.com/elixir-lang/elixir/blob/master/lib/mix/test/mix/tasks/test_test.exs#L126

lib/mix/lib/mix/tasks/test.ex
TheFirstAvenger: I tested this out and it seemed to only remember the first one. The other issue I ran into with handling it via OptionParser is that the entire logic of the test function would need to be wrapped in an `if` statement, but breaking it out as done here and in the `cmd` task results in a cleaner code path.

josevalim: You can break the testing function in two today:

```
def run(args) do
  {args, opts} = OptionParser....
  run(args, opts)
end
```

Then you don't need to wrap it all in an `if`.

TheFirstAvenger:
I rewrote it to use OptionParser from the beginning, removing the requirement for `--app` to be first. Also added a test and corresponding fixture.

josevalim:
Hi @TheFirstAvenger,

I have been thinking a bit more about this feature and I wonder that, if instead of supporting the --app flag, if we should instead allow autocomplete to the path app. So we would do:

    mix test apps/foo/test

The reason for this is the following: if we want to fix how the paths work in umbrellas, which you mentioned you want, then we need to understand paths like "apps/foo", so we might as well use that as our API.

Thoughts?

josevalim:
Closing in favor of the new PR!

PR #9041, opened 2019-05-14
Support default value when reading inexistent module attribute

mguimas:
This PR permits to provide a default value as the third argument to `Module.get_attribute`.

### DIFF
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -2957,7 +2957,7 @@ defmodule Kernel do
     case function? do
       true ->
         value =
-          case Module.get_attribute(env.module, name, line) do
+          case Module.__get_attribute__(env.module, name, line) do
             {_, doc} when doc_attr? -> doc
             other -> other
           end
@@ -2973,7 +2973,7 @@ defmodule Kernel do
 
       false when doc_attr? ->
         quote do
-          case Module.get_attribute(__MODULE__, unquote(name), unquote(line)) do
+          case Module.__get_attribute__(__MODULE__, unquote(name), unquote(line)) do
             {_, doc} -> doc
             other -> other
           end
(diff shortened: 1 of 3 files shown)


lib/elixir/lib/kernel.ex
```diff
@@ -2973,15 +2973,15 @@ defmodule Kernel do
...
+          case Module.__get_attribute__(__MODULE__, unquote(name), unquote(line)) do
             {_, doc} -> doc
             other -> other
           end
         end
 
       false ->
         quote do
-          Module.get_attribute(__MODULE__, unquote(name), unquote(line))
+          Module.__get_attribute__(__MODULE__, unquote(name), unquote(line))
```
josevalim: I like this a lot. :heart: For consistency, can you please rename the private `put_attribute` to `__put_attribute__` too?

mguimas: I renamed it in next commit. Thanks !

lib/elixir/test/elixir/module_test.exs
```diff
@@ -436,4 +436,30 @@ defmodule ModuleTest do
...
+      in_module do
+        Module.register_attribute(__MODULE__, :value, accumulate: true)
+        Module.put_attribute(__MODULE__, :value, 1)
+        assert Module.get_attribute(__MODULE__, :value) == [1]
+        Module.put_attribute(__MODULE__, :value, 2)
+        assert Module.get_attribute(__MODULE__, :value) == [2, 1]
+      end
+    end
+
+    test "returns the value of the attribute if this exists" do
```
fertapric:
```suggestion
    test "returns the value of the attribute if it exists" do
```

lib/elixir/test/elixir/module_test.exs
```diff
@@ -436,4 +436,30 @@ defmodule ModuleTest do
       Module.get_attribute(Enum, :moduledoc)
     end
   end
+
+  describe "get_attribute/3" do
+    test "returns a list when the attribute is marked as `accummulate`" do
```
fertapric:
```suggestion
    test "returns a list when the attribute is marked as `accummulate: true`" do
```

PR #9044, opened 2019-05-15
Improve List.duplicate/2

eksperimental:


lib/elixir/lib/list.ex
```diff
@@ -137,16 +137,27 @@ defmodule List do
   @doc """
   Duplicates the given element `n` times in a list.
 
+  `n` is an integer greater or equal than `0`.
```
fertapric:
```suggestion
  `n` is an integer greater than or equal to `0`.
```

josevalim:

PR #9046, opened 2019-05-16
Fix overlapping domains in Dialyzer

eksperimental:


lib/elixir/lib/io.ex
```diff
@@ -415,8 +415,9 @@ defmodule IO do
   See `IO.getn/3` for a description of return values.
 
   """
-  @spec getn(chardata | String.Chars.t(), pos_integer) :: chardata | nodata
-  @spec getn(device, chardata | String.Chars.t()) :: chardata | nodata
```
josevalim: I don't think this one has overlapping domains as devices are either PIDs or atoms?

eksperimental: It does becuase `String.Chars.t/0` is defined as term.

josevalim:
Before protocol consolidation yes, but not after protocols are compiled.
And dialyzer should run after protocols are compiled.
-- 


eksperimental:
> Before protocol consolidation yes, but not after protocols are compiled. And dialyzer should run after 

I may not be getting something, but I reimplemented the function function and the specs and I do get the overloading error
https://github.com/eksperimental/test_overlapping/blob/master/lib/overlapping_io.ex

```
done in 0m1.6s
lib/overlapping_io.ex:9: Overloaded contract for 'Elixir.OverlappingIO':getn/2 has overlapping domains; such contracts are currently unsupported and are simply ignored
done (warnings were emitted)
```

josevalim:
Oh, Cool. I will take a look later then once I have some time, but let’s
remove it here for now so we can at least merge the others!
-- 


eksperimental:
Please let's hold it off for a bit, as I am adding improvements

lib/mix/lib/mix/generator.ex
```diff
@@ -115,7 +115,7 @@ defmodule Mix.Generator do
   to override it. Returns true otherwise.
   """
   @doc since: "1.9.0"
-  @spec overwrite?(Path.t(), iodata) :: boolean
+  @spec overwrite?(Path.t()) :: boolean
```
eksperimental: I was planning on reporting an issue about this. But I wan to check if it's possible to solve it first.

Is there a way the compiler can detect this kind of error and warn about it at least? the dame way when we misplace function definitions that are in the wrong order

eksperimental:
> Please let's hold it off for a bit, as I am adding improvements

Please guys have a look to the new changes and tell me what you think

lib/mix/lib/mix/generator.ex
josevalim: Not really since someone can today just put all `@spec` at the top of the module. Plus, given that the `spec` repeats the name, it would be weird to enforce "sequence" AND expect the name. We could have made somethitng like:

    @spec arg1, arg2 -> return

But it is too late for that and if we are doing it we probably need a general overhaul of how specs work (which we don't plan in doing for now).

eksperimental: Alright, I guess it could be a little job for a Credo or any other linter

fertapric:
Thanks @eksperimental! 💜

Issue #9047, opened 2019-05-16
Almost true: "heredoc start must be followed by a new line"


playfulcats:
### Environment

* Elixir & Erlang/OTP versions (elixir --version):

  Erlang/OTP 21 [erts-10.3.5] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [hipe]  
  Elixir 1.8.2 (compiled with Erlang/OTP 20)

* Operating system: 

  Ubuntu 18.10

### Current behavior

Sample code:
```elixir
'''x
'''
```
Compiler error message:
```
** (SyntaxError) test.exs:1: heredoc start must be followed by a new line after '''
    (elixir) lib/code.ex:767: Code.require_file/2
```

Replace `x` by any number of spaces before new line.
There are no any error messages, spaces just removed.

### Expected behavior

Well, perhaps, it would be more correct to not allow spaces between the heredoc start and new line just like other characters.

josevalim:
> Well, perhaps, it would be more correct to not allow spaces between the heredoc start and new line just like other characters.

Doing that right now would be a breaking change, so our best option is to say that "followed by optional whitespace and a newline". A PR would be very welcome!

josevalim:
Closing in favor of #9048.

PR #9048, opened 2019-05-17
Improve heredoc error message

kw7oe:
Address issue #9047 

I am not sure if this PR resolve the issue mentioned. If not, feel free to point me to the right direction. Thanks.

### DIFF
--- a/lib/elixir/src/elixir_tokenizer.erl
+++ b/lib/elixir/src/elixir_tokenizer.erl
@@ -879,7 +879,7 @@ extract_heredoc(Line0, Column0, Rest0, Marker, Scope) ->
           {error, {ErrorLine, ErrorColumn, Message, Token}}
       end;
     error ->
-      Message = "heredoc start must be followed by a new line after ",
+      Message = "heredoc start must be followed by optional whitespace and a new line after ",
       {error, {Line0, Column0, io_lib:format(Message, []), [Marker, Marker, Marker]}}
   end.
 
--- a/lib/elixir/src/elixir_tokenizer.erl
+++ b/lib/elixir/src/elixir_tokenizer.erl
@@ -879,7 +879,7 @@ extract_heredoc(Line0, Column0, Rest0, Marker, Scope) ->
           {error, {ErrorLine, ErrorColumn, Message, Token}}
       end;
     error ->
-      Message = "heredoc start must be followed by optional whitespace and a new line after ",
+      Message = "heredoc allows only zero or more whitespace characters followed by a new line after ",
       {error, {Line0, Column0, io_lib:format(Message, []), [Marker, Marker, Marker]}}
   end.
 
--- a/lib/elixir/test/elixir/kernel/errors_test.exs
+++ b/lib/elixir/test/elixir/kernel/errors_test.exs
@@ -159,7 +159,7 @@ defmodule Kernel.ErrorsTest do
 
   test "heredoc start" do
     assert_eval_raise SyntaxError,
-                      "nofile:1: heredoc start must be followed by optional whitespace and a new line after \"\"\"",
+                      "nofile:1: heredoc allows only zero or more whitespace characters followed by a new line after \"\"\"",
                       '"""bar\n"""'
   end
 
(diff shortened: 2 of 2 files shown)


eksperimental:
It is kind of contradictory the "MUST be followed by OPTIONAL whitespace"
here's an improved version.

> heredoc start may be followed by whitespaces after ''' but must be followed by a new line

optional
> heredoc start may be followed by whitespaces after ''' **and** must be followed by a new line

kw7oe:
> It is kind of contradictory the "MUST be followed by OPTIONAL whitespace"
> here's an improved version.
> 
> > heredoc start may be followed by whitespaces after ''' but must be immediately followed by a new line
> 
> optional
> 
> > heredoc start may be followed by whitespaces after ''' **and** must be immediately followed by a new line

Now that you mentioned it, I agree that it is kind of contradictory. Personally, I think the first improved version is better.

josevalim:
Unfortunately the """ must always come at the end. So perhaps:

> heredoc start may be followed by whitespaces and then followed by a required a new line after """

?

eksperimental:
I'm OK with that (therea a typo though, it should read "by a required new line").
Why does the message need to end with the markers?

josevalim:
> Why does the message need to end with the markers?

Yes.

> I'm OK with that (therea a typo though, it should read "by a required new line").

I am not sure where the "by" goes. Can you please post the whole suggestion? :)

kw7oe:
> I'm OK with that (there is a typo though, it should read "by a required new line").
> Why does the message need to end with the markers?

The message need to end with the markers because it will raise `TokenMissingError` instead of `SyntaxError` if the message does not end with markers _(probably due to the implementation of `parse_error` in `elixir_errors.erl`)_.

To place the marker in between in the message, I think it will require code changes in the mentioned module.

eksperimental:
there was an extra "a" in the suggestion you posted before your edit.
> heredoc start may be followed by whitespaces and then followed by a required new line after """

I would go with something like this.

> heredoc start """ may be followed by optional whitespaces and a new line is required after it


It's kind of limiting having to write the sentence ending with the `"""`, 
here's the hack to achieve it

```diff
diff --git a/lib/elixir/src/elixir_tokenizer.erl b/lib/elixir/src/elixir_tokenizer.erl
index 3b25341a8..b460e0501 100644
--- a/lib/elixir/src/elixir_tokenizer.erl
+++ b/lib/elixir/src/elixir_tokenizer.erl
@@ -879,8 +879,8 @@ extract_heredoc(Line0, Column0, Rest0, Marker, Scope) ->
           {error, {ErrorLine, ErrorColumn, Message, Token}}
       end;
     error ->
-      Message = "heredoc start must be followed by a new line after ",
-      {error, {Line0, Column0, io_lib:format(Message, []), [Marker, Marker, Marker]}}
+      Message = "heredoc start ~ts may be followed by optional whitespaces and a new line is required after ",
+      {error, {Line0, Column0, io_lib:format(Message, [[Marker, Marker, Marker]]), ["it"]}}
   end.
 
 heredoc_error_message(eof, Line, Terminator) ->
diff --git a/lib/elixir/test/elixir/kernel/errors_test.exs b/lib/elixir/test/elixir/kernel/errors_test.exs
index aa5233ef1..5967c3489 100644
--- a/lib/elixir/test/elixir/kernel/errors_test.exs
+++ b/lib/elixir/test/elixir/kernel/errors_test.exs
@@ -159,7 +159,7 @@ defmodule Kernel.ErrorsTest do
 
   test "heredoc start" do
     assert_eval_raise SyntaxError,
-                      "nofile:1: heredoc start must be followed by a new line after \"\"\"",
+                      "nofile:1: heredoc start \"\"\" may be followed by optional whitespaces and a new line is required after it",
                       '"""bar\n"""'
   end
 
```

josevalim:
The marker may be used by other parts of the tool chain, it is important
that it is actually the token and not some random word. Therefore the “””
has to come last.
-- 


josevalim:
One last take:

> heredoc allows only zero or more whitespace characters followed by a new line after

kw7oe:
I have make the changes using the following error message:

> heredoc allows only zero or more whitespace characters followed by a new line after

fertapric:
What do you think about?

> heredoc's start must be followed by a new line (whitespaces are ignored) after

or:

> heredoc's start must be followed by a new line (ignoring whitespaces) after

josevalim:
@fertapric I think the (ignoring whitespace) is not super clear. Is it before the """, after? before the newline, after?

fertapric:
Thanks @kw7oe! 💜

PR #9050, opened 2019-05-17
Resolve full umbrella path name in mix test arg

TheFirstAvenger:
When running `mix test` in an umbrella app, will remove the `app/my_app/` prefix of any passed file name, allowing autocompleted path names to be recognized.

Mentioned in: https://github.com/elixir-lang/elixir/pull/9036


lib/mix/lib/mix/tasks/test.ex
```diff
@@ -327,6 +327,16 @@ defmodule Mix.Tasks.Test do
   def run(args) do
     {opts, files} = OptionParser.parse!(args, strict: @switches)
 
+    files =
+      if Mix.Task.recursing?() do
+        case Regex.run(~r|.*\/apps\/(.*?)$|, File.cwd!()) do
```
josevalim: This looks promising! Unfortunately matching on `/apps/` could led to false positives. For example, someone may have all `apps` in an `apps` directory.

The good news is, if we are recursing, then we must be inside an umbrella, and we can get the current app with `Mix.Project.config[:app]`. So here is what I would do:

1. Check if we are recursing
2. If so, get the current app name
3. Then traverse test files. If the test file starts with "apps/", then we check if it belongs to "apps/#{current_app}", if so, we keep it without the "apps/#{current_app}" prefix, otherwise we discard it (since it belongs to another app than the current one)

Thoughts?

TheFirstAvenger: Yes, thanks, that was my initial approach, but I blanked on how to get the current app name in a task. Pushed a new version with this approach.

lib/mix/lib/mix/tasks/test.ex
```diff
@@ -327,6 +327,13 @@ defmodule Mix.Tasks.Test do
   def run(args) do
     {opts, files} = OptionParser.parse!(args, strict: @switches)
 
+    files =
+      if Mix.Task.recursing?() do
+        Enum.map(files, &String.trim_leading(&1, "apps/#{Mix.Project.config()[:app]}/"))
```
josevalim: Looks great! We also need to remove any file that starts with "apps/" but does not belong to the current app. :)

TheFirstAvenger: Since those won't match any files, should we leave them so they will appear in this line:
https://github.com/elixir-lang/elixir/blob/master/lib/mix/lib/mix/tasks/test.ex#L427
in the case that nothing matches?

josevalim:
I see. But I think that if I am running something for a given app
exclusively, we don’t want to print anything for the other apps?
-- 


TheFirstAvenger:
In my mind, the use case for that "path did not match any files" warning is to alert to a mistyped pattern, and if the mistype is in the folder name, the user may be confused as to why the pattern they typed isn't appearing in the pattern we warn them that didn't match.

Also I think if we did remove the patterns from the list, it would result in an empty list for all apps other than the one that matched the pattern, and I believe that would cause the mix test to run for *all* files on those other apps since it would think no file pattern was specified.

josevalim:
Right. Here is what I think should be a proper heuristic (note I haven't tested/compiled it):

```elixir
{app_files, all_apps_files} = Enum.split_with(files, &String.starts_with?(&1, "apps/"))
current_app_path = "apps/#{Mix.Project.config()[:app]}/"

# We have to run all of our own app files PLUS any file
# that belongs to another app but doesn't actually exist (for error reporting)
current_app_files =
  Enum.filter(app_files, fn file ->
    String.starts_with?(&1, current_app_path) or not File.exists?(Path.join("../..", file))
  end)

files = current_app_files ++ all_apps_files

# We have nothing to run but other apps do, so we do nothing
if files == [] and app_files != [] do
  :ok
else
  run(opts, files)  
end
```

So the idea is:

1. If a file looks like an app file but does not exist, we run it in all apps (for the error reporting)
2. But if we have no file to run and other apps do, we skip our suite

What do you think?

TheFirstAvenger:
So much nicer! Now when you specify a file, you only get the one apps test output and don't have to scroll up through 7 other child apps telling you they couldn't find the file 😄.

josevalim:
Beautiful. Now we only need docs and tests. Also, please make sure the
“do_run” function is private. :)
-- 


TheFirstAvenger:
👍 Tests and docs added.

lib/mix/lib/mix/tasks/test.ex
```diff
@@ -146,6 +146,11 @@ defmodule Mix.Tasks.Test do
 
       mix test test/some/particular/file_test.exs
 
+  Tests in umbrella apps can specify the full path including `apps/my_app/`
```
josevalim: In order to run all tests in an umbrella, I need to do this though:

    mix test apps/my_app/test

Correct? I.e. the "test" directory is required.

lib/mix/test/mix/tasks/test_test.exs
```diff
@@ -297,6 +297,38 @@ defmodule Mix.Tasks.TestTest do
     end)
   end
 
+  describe "umbrella with file path" do
+    test "Finds file in umbrella when root apps path specified" do
```
josevalim:
```suggestion
    test "finds file in umbrella when root apps path specified" do
```

lib/mix/test/mix/tasks/test_test.exs
```diff
@@ -297,6 +297,38 @@ defmodule Mix.Tasks.TestTest do
...
+
+        assert output =~ """
+               ==> bar
+               .
+
+               """
+      end)
+    end
+
+    test "Only runs recursive tests for apps containing specified file" do
```
josevalim:
```suggestion
    test "only runs recursive tests for apps containing specified file" do
```

josevalim:
Thanks @TheFirstAvenger!

I have added some more comments. Furthermore, we don't need to add new fixtures, we can use the existing `umbrella_cover`. We can potentially rename `umbrella_cover` to `umbrella_test`, but to me both names are OK. Your call.

Note we also need a test that if you do: "mix test apps/unknown_app/test", we get error messages.

lib/mix/lib/mix/tasks/test.ex
TheFirstAvenger: The idea with that "including" note was to highlight the difference between an umbrella and a regular app, the delta being only those two folders. `test` is required in both so isn't a difference. `test` is included in the actual example for both so it should hopefully be clear that it is required.

TheFirstAvenger:
Test names fixed and `umbrella_cover` -> `umbrella_test`

lib/mix/lib/mix/tasks/test.ex
```diff
@@ -146,6 +146,16 @@ defmodule Mix.Tasks.Test do
 
       mix test test/some/particular/file_test.exs
 
+  Tests in umbrella projects can be run from the root by specifying
+  the full suite path, including `apps/my_app/test`, in which case
+  recursive tests for other child apps will be skipped completely:
```
josevalim: @TheFirstAvenger I see, let me know what you think about this change then!

lib/mix/test/mix/tasks/test_test.exs
```diff
@@ -297,6 +297,38 @@ defmodule Mix.Tasks.TestTest do
     end)
   end
 
+  describe "umbrella with file path" do
+    test "finds file in umbrella when root apps path specified" do
```
josevalim: Btw, we can probably merge those two tests into one. Those tests are very expensive, since they need to start a separate suite, so generally try to perform multiple assertions.

You can also just add an assertion of what happen when you run `mix(["test", "apps/unknown_app/test"])`, without creating a whole new test case, for the same reasons.

TheFirstAvenger: 👍 Merged the tests into one (with the individual test reasoning added as failure messages instead) and added an assertion on the unknown app call.

josevalim:

PR #9052, opened 2019-05-18
Base erts lib dir off erts source path

mobileoverlord:
If `include_erts` is a path instead of a boolean, the erts lib dir should be based from this location instead of `:code.lib_dir()`. This is useful when trying to include cross compiled lib apps.

### DIFF
--- a/lib/mix/lib/mix/release.ex
+++ b/lib/mix/lib/mix/release.ex
@@ -221,14 +221,11 @@ defmodule Mix.Release do
   defp erts_data(erts_source) when is_binary(erts_source) do
     if File.exists?(erts_source) do
       [_, erts_version] = erts_source |> Path.basename() |> String.split("-")
-      {to_charlist(erts_source), erts_lib_dir(erts_source), to_charlist(erts_version)}
+      erts_lib_dir = erts_source |> Path.dirname() |> Path.join("lib") |> to_charlist()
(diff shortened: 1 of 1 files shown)


lib/mix/lib/mix/release.ex
```diff
@@ -83,7 +83,7 @@ defmodule Mix.Release do
     {include_erts, opts} = Keyword.pop(opts, :include_erts, true)
     {erts_source, erts_version} = erts_data(include_erts)
```
josevalim: Let's instead change `erts_data` to return the `erts_lib_dir` too, so we don't have to process `include_erts` twice.

lib/mix/lib/mix/release.ex
```diff
@@ -206,6 +206,18 @@ defmodule Mix.Release do
...
   end
 
+  defp erts_lib_dir(erts_path) when is_function(erts_path) do
+    erts_lib_dir(erts_path.())
+  end
+
+  defp erts_lib_dir(erts_path) when is_binary(erts_path) or is_list(erts_path) do
+    Path.join([erts_path, "..", "lib"])
+    |> Path.expand()
+    |> to_charlist()
```
josevalim: I think this can be written as: `erts_path |> Path.dirname |> Path.join("lib") |> to_charlist`.

mobileoverlord:
Updated to address comments from above

josevalim:

PR #9053, opened 2019-05-18
Allow release options to be a function

mobileoverlord:
Allowing the release options lets us clean up and further manage a lot of the required release parameters that need to be placed in a users mix file.

Here is an example of what this looks like for Nerves projects.

```
  def project do
    [
      app: :elixir19,
      # ..
      default_release: :nerves,
      releases: [nerves: &Nerves.release/0]
    ]
  end
```

### DIFF
--- a/lib/mix/lib/mix/release.ex
+++ b/lib/mix/lib/mix/release.ex
@@ -134,6 +134,7 @@ defmodule Mix.Release do
 
   defp find_release(name, config) do
     {name, opts} = lookup_release(name, config) || infer_release(config)
+    opts = expand_release_opts(opts)
     {apps, opts} = Keyword.pop(opts, :applications, [])
 
     if apps == [] and Mix.Project.umbrella?(config) do
@@ -185,6 +186,14 @@ defmodule Mix.Release do
     else
       {Keyword.fetch!(config, :app), []}
     end
+  end
+
+  defp expand_release_opts(opts) when is_function(opts) do
+    opts.()
+  end
+
+  defp expand_release_opts(opts) do
+    opts
   end
 
   defp bad_umbrella! do


josevalim:
@mobileoverlord my concern with this is that the user has no mechanism to customize the release and I have a hunch that it won't be enough. Is:

    releases: [my_app: [steps: [&Nerves.release/1, :assemble]]]

That much more work?

mobileoverlord:
I find that its often uncommon for users in Nerves to modify the release steps. I was thinking that since its uncommon, those uses can feel free to copy in the slug of defaults and manage it in their mix files directly. The issue is that the defaults we have are these:
```
[
      quiet: true,
      overwrite: true,
      include_executables_for: [],
      include_erts: &Nerves.Release.erts/0,
      steps: [&Nerves.Release.init/1, &Shoehorn.Release.init/1, :assemble]
]
```

Since a lot of this is meaningless to the dev, I thought it would be easiest for us to manage it.

josevalim:
At least `quiet` and `include_executables_for` can be set inside the steps. You can also set shoehorn as a step. So you would need this:

```elixir
[
      overwrite: true,
      include_erts: &Nerves.Release.erts/0,
      steps: [&Nerves.Release.init/1, :assemble]
]
```

?

josevalim:
We can do something like:

```elixir
releases: [
  nerves: [defaults: &Nerves.release/0]
]
```

But we need to be careful with painting the users into a corner, because as soon as they need to configure anything, they will need to know the defaults. For example, if they want to add a custom step, they need to know that now they need to provide the nerves step, shoehorn step, and the default one. If they try to provide a custom `env.sh/env.bat`, it won't work for Nerves either, etc. So if we hide the mechanisms from them, they may not understand why part of the releases do not work as they expect.

josevalim:
Closing per the reasons above. Let's follow up later!

PR #9054, opened 2019-05-18
Try loading app from otp_root before :code.lib_dir

mobileoverlord:
When calling load app, we should attempt to load it from the otp_root before consulting `:code.lib_dir`. This fixes issues when specifying `include_erts` as a path to a cross compiled version of erlang.

### DIFF
--- a/lib/mix/lib/mix/release.ex
+++ b/lib/mix/lib/mix/release.ex
@@ -244,7 +244,7 @@ defmodule Mix.Release do
         case :code.lib_dir(app) do
           {:error, :bad_name} -> Mix.raise("Could not find application #{inspect(app)}")
           path -> do_load_app(app, path, seen, otp_root, false)
-        end 
+        end
 
       [path] ->
         do_load_app(app, to_charlist(path), seen, otp_root, true)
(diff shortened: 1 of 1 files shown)


lib/mix/lib/mix/release.ex
```diff
@@ -237,11 +237,24 @@ defmodule Mix.Release do
   end
 
   defp load_app(app, seen, otp_root) do
+    path = Path.join([otp_root, "#{app}-*"])
+
+    case Path.wildcard(path) do
+      [] ->
     case :code.lib_dir(app) do
```
josevalim: We probably don't need this branch anymore, we can fully rely on `Path.wildcard`.

josevalim:
I have added a comment, also, please don't forget to run the formatter!

lib/mix/lib/mix/release.ex
josevalim: Actually, we do need this branch, for non OTP apps. So yeah, we just need to format and we should be good to go.

mobileoverlord:
I added the `:code.lib_dir` lookup back and formatted

josevalim:

PR #9055, opened 2019-05-18
Include erts lib dir and paths for release script

mobileoverlord:
systools requires the paths to be declared inorder to find apps that are coming from a different location than the release. This fixes issues when trying to inclide erts and erlang libs that are not part of the host system.

### DIFF
--- a/lib/mix/lib/mix/release.ex
+++ b/lib/mix/lib/mix/release.ex
@@ -432,18 +432,15 @@ defmodule Mix.Release do
   end
 
   defp build_variables(release) do
-    variables =
-      for {_, properties} <- release.applications,
-          not Keyword.fetch!(properties, :otp_app?),
-          uniq: true,
-          do: {'RELEASE_LIB', properties |> Keyword.fetch!(:path) |> :filename.dirname()}
-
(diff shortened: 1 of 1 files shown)


lib/mix/lib/mix/release.ex
```diff
@@ -425,10 +432,19 @@ defmodule Mix.Release do
   end
 
   defp build_variables(release) do
+    variables =
+      for {_, properties} <- release.applications,
+          not Keyword.fetch!(properties, :otp_app?),
+          uniq: true,
+          do: {'RELEASE_LIB', properties |> Keyword.fetch!(:path) |> :filename.dirname()}
+
+    [{'ERTS_LIB_DIR', erts_lib_dir(release.erts_source)} | variables] |> IO.inspect()
```
josevalim: By using ERTS_LIB_DIR, it means we need to set another variable as --boot-bar when assembling the release, no?

mobileoverlord:
This is failing because it depends on https://github.com/elixir-lang/elixir/pull/9052
Should we add `erts_lib_dir` to the `%Mix.Release{}` struct since its needed later?

mobileoverlord:
I tested this out and it works. Thanks!

josevalim:

PR #9056, opened 2019-05-18
Start the VM in interactive mode for rpc/eval/remote

josevalim:
This speeds up boot time and decreases memory use.

### DIFF
--- a/lib/mix/lib/mix/tasks/release.init.ex
+++ b/lib/mix/lib/mix/tasks/release.init.ex
@@ -42,10 +42,7 @@ defmodule Mix.Tasks.Release.Init do
   def vm_args_text,
     do: ~S"""
     ## Customize flags given to the VM: http://erlang.org/doc/man/erl.html
-    ## -name/-sname/-setcookie are configured via env vars, do not set them here
-
-    ## Preloads all modules instead of loading them dynamically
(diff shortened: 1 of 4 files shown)


lib/mix/lib/mix/tasks/release.init.ex
```diff
@@ -117,6 +115,7 @@ defmodule Mix.Tasks.Release.Init do
       exec "$REL_VSN_DIR/$REL_EXEC" \
            --cookie "$RELEASE_COOKIE" \
            --$RELEASE_DISTRIBUTION "$RELEASE_NODE" \
+           --erl "-mode $RELEASE_MODE" \
```
josevalim: We have been adding switches to the `elixir` executable but mostly to avoid double encoding paths, otherwise handling paths with spaces or quotes in them is very complicated. This time, I decided to pass it via the `--erl` flag directly but we could add `--mode`. I am fine with both approaches but I am dropping a comment in case someone has stronger feelings about this. Thoughts?

whatyouhide: If users are going to use this themselves frequently, I'd go with adding `--mode`, otherwise this looks fine to me :)

josevalim:

PR #9062, opened 2019-05-21
Remove Dialyzer errors in Application module

eksperimental:
We were getting the "Type specification ... is a supertype of the success typing" error

/elixir/lib/elixir/lib/application.ex:367: Type specification 'Elixir.Application':spec(app()) -> [{key(),value()}] | 'nil' is a supertype of the success typing: 'Elixir.Application':spec(atom()) -> 'nil' | [{'applications' | 'description' | 'env' | 'id' | 'included_applications' | 'maxP' | 'maxT' | 'mod' | 'modules' | 'registered' | 'start_phases' | 'vsn',_}]
/elixir/lib/elixir/lib/application.ex:382: Type specification 'Elixir.Application':spec(app(),key()) -> value() | 'nil' is a supertype of the success typing: 'Elixir.Application':spec(atom(),'applications' | 'description' | 'id' | 'included_applications' | 'maxP' | 'maxT' | 'mod' | 'modules' | 'registered' | 'start_phases' | 'vsn') -> any()

### DIFF
--- a/lib/elixir/lib/application.ex
+++ b/lib/elixir/lib/application.ex
@@ -333,13 +333,6 @@ defmodule Application do
     end
   end
 
-  @type app :: atom
-  @type key :: atom
-  @type value :: term
-  @type state :: term
-  @type start_type :: :normal | {:takeover, node} | {:failover, node}
-  @type restart_type :: :permanent | :transient | :temporary
-
   @application_keys [
     :description,
     :id,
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/application.ex
```diff
@@ -347,19 +347,11 @@ defmodule Application do
...
-  application_key_specs = fn ->
-    {:ok, ast} =
-      @application_keys
-      |> Enum.map(fn key -> ":#{key}" end)
-      |> Enum.join(" | ")
-      |> Code.string_to_quoted()
-
-    ast
-  end
+  application_key_specs = Enum.reduce(@application_keys, &{:|, [], [&1, &2]})
```
eksperimental: Awesome. I knew there was a simpler way

lib/elixir/lib/application.ex
```diff
@@ -364,7 +367,7 @@ defmodule Application do
   Note the environment is not returned as it can be accessed via
   `fetch_env/2`. Returns `nil` if the application is not loaded.
   """
-  @spec spec(app) :: [{key, value}] | nil
+  @spec spec(app) :: [{application_key, value}] | nil
```
fertapric: 👍 

Btw, the [documentation in Erlang](http://erlang.org/doc/apps/kernel/application.html#get_all_key-1) only specifies `[{atom, term}]`, but that function only returns the keys specified above ([in the source code](https://github.com/erlang/otp/blob/da13166940df546d3fb05a71948f1709d72dd354/lib/kernel/src/application_controller.erl#L394-L412)).

eksperimental: I will submit PR to OTP

whatyouhide:
Thanks @eksperimental! 💟

PR #9065, opened 2019-05-21
Support paths with white spaces in Windows releases

fertapric:


lib/mix/test/mix/tasks/release_test.exs
```diff
@@ -297,7 +298,7 @@ defmodule Mix.Tasks.ReleaseTest do
         assert String.trim_trailing(version) == "eval 0.1.0"
         refute File.exists?(Path.join(root, "RELEASE_BOOTED"))
 
-        {hello_world, 0} = System.cmd(script, ["eval", ~s[IO.puts :hello_world]])
+        {hello_world, 0} = System.cmd(script, ["eval", "IO.puts(:hello_world)"])
```
josevalim: Just to be clear, is this a style change or an actual fix? :)

lib/mix/lib/mix/tasks/release.init.ex
```diff
@@ -245,11 +245,11 @@ defmodule Mix.Tasks.Release.Init do
     popd
 
     if not defined RELEASE_NAME (set RELEASE_NAME=<%= @release.name %>)
-    if not defined RELEASE_VSN (for /f "tokens=1,2" %%K in (!RELEASE_ROOT!\releases\start_erl.data) do (set ERTS_VSN=%%K) && (set RELEASE_VSN=%%L))
+    if not defined RELEASE_VSN (for /f "tokens=1,2" %%K in ('type "!RELEASE_ROOT!\releases\start_erl.data"') do (set ERTS_VSN=%%K) && (set RELEASE_VSN=%%L))
```
fertapric: The [`FOR` command](https://ss64.com/nt/for.html) also accepts a string when the `/f` option is passed, so just quoting the path was not enough.

I got the `type` _trick_ from here: http://www.celticproductions.net/articles/12/batch-files/file-path-and-name-with-spaces-in-batch-file-for-loop.html

lib/mix/test/mix/tasks/release_test.exs
fertapric: Using parentheses (and removing the space) is an actual fix 🙂 

Replacing `~s[...]` with `"..."` is a style change 😬

josevalim: If removing the spaces is an actual thing, then we have a bug in our batch script/evaluator, because it means it is not handling spaces accordingly. `IO.puts :hello_world` should be a single command, because we are passing it as a single command. So perhaps we should leave it with space to help us find bugs?

josevalim: Which error would you get? `IO.puts/0` does not exist or similar?

fertapric: I got to the same conclusion and I was currently investigating the bug :)

fertapric: Here is the error:

Screen Shot 2019-05-21 at 20 20 57

fertapric: I think it's not related to releases:

Screen Shot 2019-05-21 at 21 10 46

josevalim: @fertapric does it change if you give it an absolute path from the beginning?

josevalim: @fertapric and are you using powershell or cmd?

josevalim: Also, which OTP version?

fertapric: > @fertapric does it change if you give it an absolute path from the beginning?

Same error apparently.

Screen Shot 2019-05-21 at 22 03 29

> @fertapric and are you using powershell or cmd?

Powershell.

> Also, which OTP version?

OTP 21.0.1

PR #9069, opened 2019-05-24
Add guards to the Stream module

whatyouhide:
Before we hadn't added guards to functions in the `Stream` module because the error message for invoking a function with the wrong number of arguments was better than the `FunctionClauseError`. Now that `FunctionClauseError`s are better and display the failed clauses and everything, having guards in `Stream` functions means that we can fail fast in many cases instead of waiting to fail until the stream is realized.


josevalim:
Just to clarify: this does not mean we want to add `is_function/1` guards to all functions. We are adding it to Stream because they are lazily evaluated, which makes debugging considerably harder.

PR #9071, opened 2019-05-24
Streamline test logs in umbrella applications

TheFirstAvenger:
Before this, each child app in an umbrella app had at least 7 newlines per child app in a test run. Because of this it didn't take many child apps for your test output to scroll way off the screen even on a fully successful `mix test`. This change streamlines the test output from a minimum of 7 down to a minimum of 2 lines when running in an umbrella.

Before:
Screen Shot 2019-05-24 at 1 34 06 PM

After:
Screen Shot 2019-05-24 at 1 30 59 PM

Note: The changes in `cli_formatter.ex` work but to remove the newline after printing the app name, I took a stab at changing `io.ex` and `test.ex`, however now `make test` fails on `ex_unit` without a useful error message and I am not sure exactly how to debug it.

### DIFF
--- a/lib/ex_unit/lib/ex_unit/cli_formatter.ex
+++ b/lib/ex_unit/lib/ex_unit/cli_formatter.ex
@@ -21,7 +21,8 @@ defmodule ExUnit.CLIFormatter do
       failure_counter: 0,
       skipped_counter: 0,
       excluded_counter: 0,
-      invalid_counter: 0
+      invalid_counter: 0,
+      condensed: Mix.Task.recursing?()
     }
 
     {:ok, config}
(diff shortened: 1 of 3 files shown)


josevalim:
Thanks @TheFirstAvenger. For those who want custom formatting, you can always implement a custom formatter, but we don't plan to go down this route on Elixir itself. I would prefer to keep everything consistent, even if verbose in an umbrella project. If you concern is missing failures, you can always pass the flag to raise as soon as any suite fails.

TheFirstAvenger:
[Done](https://github.com/TheFirstAvenger/umbrella_streamline_formatter). Thoughts on how I could drop the newline after printing the app name so the dots start on that line?

josevalim:
I don't think particular change will be possible. That's how we print app information for all Mix tasks and at this point the ExUnit formatter has not been involved at all. :(

PR #9072, opened 2019-05-24
Raise an error if the umbrella app's dir name and mix.exs app name doesn't match

### LINKED ISSUE #9038: Could not start application in umbrella if folder name has '.' or does not match
### Precheck

* Please disclose security vulnerabilities privately at elixir-security@googlegroups.com

### Environment

```
Erlang/OTP 21 [erts-10.3.4] [source] [64-bit] [smp:12:12] [ds:12:12:10] [async-threads:1] [hipe] [dtrace]

Elixir 1.8.1 (compiled with Erlang/OTP 21)
```
* Operating system: 
```
ProductName:	Mac OS X
ProductVersion:	10.14.4
BuildVersion:	18E226
```

### Current behavior
When creating an application in an umbrella application with a folder name with `.` in it, or if the name does not match; it fails to compile.

I concede that there may be a bit of configuration knowledge I'm lacking.

[Example Repo](https://github.com/andersonmcook/failed-umbrella)

Run
```sh
$ mix new ./apps/foo.bar --sup --module Bar --app bar
```

Result
```
* creating README.md
* creating .formatter.exs
* creating .gitignore
warning: redefining module FailedUmbrella.MixProject (current version defined in memory)
  /Users/my/path/failed-umbrella/mix.exs:1

Error while loading project :umbrella_check at /Users/my/path/failed-umbrella
* creating mix.exs
* creating config
* creating config/config.exs
* creating lib
* creating lib/bar.ex
* creating lib/bar/application.ex
* creating test
* creating test/test_helper.exs
* creating test/bar_test.exs

Your Mix project was created successfully.
You can use "mix" to compile it, test it, and more:

    cd ./apps/foo.bar
    mix test

Run "mix help" for more commands.
```

Run
```sh
$ iex -S mix
```

Result
```
** (Mix) Could not start application foo.bar: could not find application file: foo.bar.app
```

Renaming the application folder to `bar` allows compilation, but `foo` and `foo.bar` do not.

### Expected behavior

Folder name has less impact on whether the code compiles.

Thanks!

leandrocp:
As reported in #9038, an umbrella app may have a name and live in another dir that's different of app's name. Because of that, we can't rely on the app's directory name, and instead need to get the app name from mix.exs

Let me know if there's a better way to do that.

Fixes #9038

### DIFF
--- a/lib/mix/lib/mix/dep.ex
+++ b/lib/mix/lib/mix/dep.ex
@@ -118,7 +118,7 @@ defmodule Mix.Dep do
     end
   end
 
-  defp load_and_cache(_config, _top, _bottom, env, target) do
+  defp load_and_cache(_config, top, top, env, target) do
     converge(env: env, target: target)
   end
 
--- a/lib/mix/lib/mix/dep.ex
+++ b/lib/mix/lib/mix/dep.ex
@@ -132,17 +132,6 @@ defmodule Mix.Dep do
     seen = populate_seen(MapSet.new(), [app])
     children = get_deps(deps, tl(Enum.uniq(get_children(deps, seen, [app]))))
 
-    if bottom.project()[:apps_path] do
-      umbrella_apps =
-        deps
-        |> Enum.filter(& &1.opts[:from_umbrella])
-        |> Enum.map(& &1.app)
-
-      app in umbrella_apps ||
-        raise "cannot load #{inspect(app)} app because the defined app name in mix.exs and the app directory name are different. " <>
-                "Please make sure both are equal, otherwise the application can't be started."
-    end
-
     top_level =
       for dep <- deps,
           dep.app == app,
--- a/lib/mix/lib/mix/dep.ex
+++ b/lib/mix/lib/mix/dep.ex
@@ -118,7 +118,7 @@ defmodule Mix.Dep do
     end
   end
 
-  defp load_and_cache(_config, top, top, env, target) do
+  defp load_and_cache(_config, _top, _bottom, env, target) do
     converge(env: env, target: target)
   end
 
@@ -131,6 +131,17 @@ defmodule Mix.Dep do
     app = Keyword.fetch!(config, :app)
     seen = populate_seen(MapSet.new(), [app])
     children = get_deps(deps, tl(Enum.uniq(get_children(deps, seen, [app]))))
+
+    if bottom.project()[:apps_path] do
+      umbrella_apps =
+        deps
+        |> Enum.filter(& &1.opts[:from_umbrella])
+        |> Enum.map(& &1.app)
+
+      app in umbrella_apps ||
+        raise "cannot load #{inspect(app)} app because the defined app name in mix.exs and the app directory name are different. " <>
+                "Please make sure both are equal, otherwise the application can't be started."
+    end
 
     top_level =
       for dep <- deps,
--- a/lib/mix/lib/mix/dep/loader.ex
+++ b/lib/mix/lib/mix/dep/loader.ex
@@ -287,7 +287,7 @@ defmodule Mix.Dep.Loader do
     end
   end
 
-  defp mix_dep(%Mix.Dep{opts: opts} = dep, nil) do
+  defp mix_dep(%Mix.Dep{app: app, opts: opts} = dep, nil) do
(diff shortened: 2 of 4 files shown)


josevalim:
Actually that makes loading umbrellas much more expensive. I would rather
document that the names have to match and raise good error messages if
possible than otherwise. :)
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

leandrocp:
> Actually that makes loading umbrellas much more expensive.
That's was my concern.

> names have to match
So, I could compare the loaded name (that will be the dir name) and compare with the loaded Mix.Project right before starting the application, and if they don't match just raise an error ? That would be the best approach ?

josevalim:
Yes, but we can't check when we load the umbrella, because that puts us exactly in the place we were. We probably need to check if we are in an umbrella or not somewhere closer to where the exception originally reported happens. :)

leandrocp:
> Yes, but we can't check when we load the umbrella, because that puts us exactly in the place we were.

Exactly :D

> We probably need to check if we are in an umbrella or not somewhere closer to where the exception originally reported happens. :)

The umbrella is loaded in `Mix.Dep.Umbrella` and the exception happens in `Mix.Tasks.App.Start` when starting the application.

I'm trying to understand if there's a way to get the config of the loaded umbrella app to compare with `%Mix.Dep{}`, or something like that.

leandrocp:
@josevalim new attempt commited. Before storing in cache and after app resolution, I'm checking if the resolved app name matches the name from config, and only if we're in an umbrella project (using `bottom`, I think that's ok).

* If the message isn't good enough, please provide a new one so I can change it.
* The same logic/raise needs to happen on `mix new` ? _To avoid creating projects that doesn't match names_
* Creating a new test for this case is desirable ?
* Need to change docs to reflect the requirement of names matching ?

Loading @andersonmcook's test app displays:
Screen Shot 2019-05-24 at 19 31 25

lib/mix/lib/mix/dep.ex
```diff
@@ -132,6 +132,17 @@ defmodule Mix.Dep do
     seen = populate_seen(MapSet.new(), [app])
     children = get_deps(deps, tl(Enum.uniq(get_children(deps, seen, [app]))))
 
+    if bottom.project()[:apps_path] do
```
josevalim: So we can't call `bottom.project()` directly because users may have code in their `def project` that assumes we are running it in the project root directory. 

After taking a quick look, I think the best place to implement this check is here:

https://github.com/elixir-lang/elixir/blob/master/lib/mix/lib/mix/dep/loader.ex#L301

You are already inside the dependency, and you have both the cached project options (by calling `Mix.Project.config()`) and the dependency. Then you can compare `Mix.Project.config()[:app]` with `dep.app` and raise if they are different.

We will also need a test, you can look at `mix/umbrella_test.exs` for an example. But please don't create a new fixture, you can use one of the existing features and change the `mix.exs` of an umbrella child in the middle of the test by calling `File.write!`.

Thanks for looking into this, it is really appreciated!

lib/mix/lib/mix/dep.ex
```diff
@@ -132,6 +132,17 @@ defmodule Mix.Dep do
...
+      umbrella_apps =
+        deps
+        |> Enum.filter(& &1.opts[:from_umbrella])
+        |> Enum.map(& &1.app)
+
+      app in umbrella_apps ||
+        raise "cannot load #{inspect(app)} app because the defined app name in mix.exs and the app directory name are different. " <>
```
josevalim: Note inside Mix we use `Mix.raise` and the error messages should start in upcase.

lib/mix/lib/mix/dep.ex
leandrocp: I really appreciate your guidance. Right now I'm heading to Elixir Brasil conf, and I'll work on that on next week :)

leandrocp:
Previous commits are useless. Should I squash commits to keep only the latest one ?

fertapric:
@leandrocp we usually squash the commits before merging the PR 🙂

josevalim:
And to clarify, by "we", @fertapric means the Elixir team. So you don't have to worry about it.

josevalim:

PR #9073, opened 2019-05-25
Support unary functions in String.replace/4

### LINKED ISSUE #9023: Add String.replace(string, value, replacement_fn)

josevalim:
We also deprecate the :insert_replaced option since the
anonymous functions is strictly superior.

Closes #9023.

### DIFF
--- a/lib/elixir/lib/string.ex
+++ b/lib/elixir/lib/string.ex
@@ -1372,7 +1372,7 @@ defmodule String do
     if insert = Keyword.get(options, :insert_replaced) do
       IO.warn(
         "String.replace/4 with :insert_replaced option is deprecated. " <>
-          "Please :binary.replace/4 instead or pass an anonymous function as replacement"
+          "Please use :binary.replace/4 instead or pass an anonymous function as replacement"
       )
 
       binary_options = if Keyword.get(options, :global) != false, do: [:global], else: []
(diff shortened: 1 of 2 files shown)


lib/elixir/lib/string.ex
```diff
@@ -1347,43 +1338,89 @@ defmodule String do
...
+      IO.iodata_to_binary([replacement.("") | subject])
     end
   end
 
-  defp intersperse(subject, replacement) do
+  def replace(subject, pattern, replacement, options) when is_binary(subject) do
+    if insert = Keyword.get(options, :insert_replaced) do
+      IO.warn(
+        "String.replace/4 with :insert_replaced option is deprecated. " <>
+          "Please :binary.replace/4 instead or pass an anonymous function as replacement"
```
eksperimental:
```suggestion
          "Please use :binary.replace/4 instead or pass an anonymous function as replacement"
```

lib/elixir/lib/string.ex
```diff
@@ -1347,43 +1338,89 @@ defmodule String do
...
     case next_grapheme(subject) do
-      {current, rest} -> [current, replacement | intersperse(rest, replacement)]
+      {current, rest} -> [current, replacement.("") | intersperse_fun(rest, replacement)]
       nil -> []
     end
   end
 
-  defp translate_replace_options(options) do
-    global = if Keyword.get(options, :global) != false, do: [:global], else: []
+  defp do_replace(subject, [], _, n) do
```
eksperimental: Can we use count or counter instead of the single letter variable n?

josevalim:

PR #9078, opened 2019-05-27
Add SECURITY.md (Security Policy)

eksperimental:
GitHub is adding this new feature (still in beta), but I think it is great to have a file dedicated to this


fertapric:
Thanks @eksperimental! 💜

PR #9082, opened 2019-05-27
Fixes #9081 by improving documentation of GenServer.server() type.

### LINKED ISSUE #9081: Spec of GenServer.server() type is lacking information of `:via`-tuples
In the documentation of GenServer, the [server()](https://hexdocs.pm/elixir/GenServer.html#t:server/0) type is specified as:

```elixir
server() :: pid() | name() | {atom(), node()}
```
which is used as parameter type for many of GenServer's functions.

It might be important to mention that `{:via, registry_name, key}`-tuples are also supported, to (a) improve the documentation, and (b) make Dyalizer happy when `:via`-tuples are passed to them.

The spec probably should be:
```elixir
server() :: pid() | name() | {atom(), node()} | {:via, registry :: atom(), identifier :: term()}
```
and we might want to improve the type documentation from 'the server reference' to something slightly more informing like:

```
The server reference. This is either a plain PID or a value representing a registered name. See the 'Name registration' section of this document for more information.
```

And related: currently, the documentation of `call` mentions something like this already, (`server can be any of the values described in the "Name registration" section of the documentation for this module.`) but then it is missing from the documentation of `cast` right below.
Maybe we want to include this information here as well?

Qqwy:
Fixes #9081 by improving documentation of GenServer.server() type, as well as mentioning its possible values in the documentation of `GenServer.cast`.

### DIFF
--- a/lib/elixir/lib/gen_server.ex
+++ b/lib/elixir/lib/gen_server.ex
@@ -721,9 +721,9 @@ defmodule GenServer do
   The server reference.
 
   This is either a plain PID or a value representing a registered name.
-  See the 'Name registration' section of this document for more information.
+  See the "Name registration" section of this document for more information.
   """
-  @type server :: pid | name | {atom, node} | {:via, registry :: atom(), key :: term()}
+  @type server :: pid | name | {atom, node} | {:via, registry :: module(), key :: term()}
 
   @typedoc """
   Tuple describing the client of a call request.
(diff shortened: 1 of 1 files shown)


fertapric:
Thanks @Qqwy! 💜

fertapric:
Oh! I just realized that `name` is defined as `@type name :: atom | {:global, term} | {:via, module, term}`, so the previous code was correct. The extra docs in `cast` are good though 🙂

PR #9085, opened 2019-05-27
Remove duplicated typespec on t:GenServer.server/0

fertapric:
`{:via, module, term}` is covered by the type `name`

Related: https://github.com/elixir-lang/elixir/pull/9082


Qqwy:
Thank you for your keen eye @fertapric ! :heart:

Issue #8931, opened 2019-04-03
slack mods


natetech333:
People on Slack are running an Elixir group and the moderators/admins are bullying people if anyone asks serious questions about Slack, privacy, policies, CoC's etc... I feel that your open source group needs to approve who can open a group under the Elixir copyright. And they really need professional training and need to prove they are qualified before being approved to open a group. The moderators talk about how they enjoy watching fights. And everything is turned into a fight on the group. They are not following any kind of process or even executing the CoC in context. All the moderators need to use their real names because they hide behind avatars and screen names in order to bully people they disagree with. Unless this is addressed, I fear that this community will remain non-inclusive to other people with questions and different backgrounds. Especially people who speak other languages.

Just because I wanted to know about the security policies... I was called a conspiracy theorist, troll, told to leave and told that they enjoyed to watch fights... and that the internet is designed for people to fight. All 3 of the moderators need removed before they destroy your open source project. 3 of the mods are in your contributors list. I'd even recommend you remove them from the github.

tinyshell666:
yeap. I see them make jokes and have fun with their friends on it. As a foreigner, I do not feel welcome on it either. Try the forum. I hope Elixir enforce the new codes.

natetech333:
1 of the mods said that the internet was made for fighting. If that's the kind of mods that this project is allowing to run their group, it's turning me away from wanting to use this project. I would expect http://plataformatec.com.br/ requires some kind of qualifications and training to be an official group for Elixir. The person didn't even point to anything wrong in the code of conduct before banning people. It is just a small group of friends who say whatever they want. Then police everyone else. I am starting to understand why all of these new codes are popping up. But the moderators are the problem, half the time. It's really pushing away women and other backgrounds, and genders.

tinyshell666:
you culd file a complaint with the cyber bullying authorities in my country. and we have organizations to handle them.

adkron:
The slack channel is not the language and is not Plataformatec.

natetech333:
I understand. But is there something that can be done about making sure groups are run by qualified professionals before allowing them to use the Elixir name?

kipcole9:
I admit to being puzzled by your observations since as one of the 23,000 member of the Elixir slack i've not seen the behaviour you're describing and I scrolled back as far as I could in General and Beginner (where I typically go) and didn't seen anything.

That said, this is a software issue tracker and the request you would have seen when you opened the issue would have been to respect that this is an issue tracker.

[...]

PR #9086, opened 2019-05-28
Review of README.md

eksperimental:


josevalim:

eksperimental:
@josevalim I think it is important to mention in the README what the official communication channels are, to avoid issues like this #8931.
Personally, I thought the Elixir Forum is an official channel, but it is not.
As of now, one needs to dig deep into the Code of Conduct to find out which are these channels.

josevalim:
To be honest, given the tone of the linked discussion, I don't think adding this information to the README would have changed anything, especially since they were told this at some point but it didn't  change the conversation much.

PR #9090, opened 2019-05-28
Add 'a' to {invalid_clauses, Name} error

radar:
Fixed a small grammatical mistake in this error message.

Before:

> the function "case" cannot handle clauses with the -> operator because it is not macro. Please make sure you are invoking the proper name and that it is a macro

After:

> the function "case" cannot handle clauses with the -> operator because it is not **a** macro. Please make sure you are invoking the proper name and that it is a macro


josevalim:

PR #9093, opened 2019-05-29
Move ISSUE_TEMPLATE.md to .github/

eksperimental:
This way we don't need to keep an extra file in the root directory that is exclusively used by GitHub


josevalim:
Let's keep it as is for now, thanks!

PR #9094, opened 2019-05-29
Use https instead of http when available

eksperimental:
The servers are automatically redirecting from http:// to https://


josevalim:

PR #9096, opened 2019-05-29
Inspect.Opts.custom_options

am-kantox:
It’s easier to explain with code.

When `Inspect.inspect/2` is used with the nested structure, it sometimes makes sense to pass some custom additional options through the call stack. That might be a custom setting, meaningful for the inspected type only. For example, the level of nesting in the snippet in the test below.

I would propose the ability to pass any option to `Inspect.inspect/2` so that if it’s not a known default option, it’s being put in the newly introduced `custom_bag` struct field. It might play the role of the “accumulator,” so that all the nested objects of the same type (all objects aware of this particular option) might get it, and update in the nesting calls to `inspect/2`.

The test below should shed more light on how it might be used.

### DIFF
--- a/lib/elixir/lib/inspect/algebra.ex
+++ b/lib/elixir/lib/inspect/algebra.ex
@@ -57,6 +57,10 @@ defmodule Inspect.Opts do
     * `:inspect_fun` (since v1.9.0) - a function to build algebra documents,
       defaults to `Inspect.inspect/2`
 
+    * `:custom_bag` (since v1.9.0) - a keyword list storing custom user-defined
+      options; useful when implementing a protocol for nested structures to pass
+      the custom options through. All the non-default options passed to
+      `Kernel.inspect/2` will be put into this bag.
   """
 
   # TODO: Remove :char_lists key on v2.0
@@ -71,7 +75,8 @@ defmodule Inspect.Opts do
             pretty: false,
             safe: true,
             syntax_colors: [],
-            inspect_fun: &Inspect.inspect/2
+            inspect_fun: &Inspect.inspect/2,
+            custom_bag: []
 
   @type color_key :: atom
 
@@ -88,8 +93,33 @@ defmodule Inspect.Opts do
           pretty: boolean,
           safe: boolean,
           syntax_colors: [{color_key, IO.ANSI.ansidata()}],
-          inspect_fun: (any, t -> Inspect.Algebra.t())
+          inspect_fun: (any, t -> Inspect.Algebra.t()),
+          custom_bag: [{atom, any()}]
         }
+
+  def custom_bag(opts) do
+    Enum.reject(opts, fn
+      {k, _}
+      when k in [
+             :structs,
+             :binaries,
+             :charlists,
+             :char_lists,
+             :limit,
+             :printable_limit,
+             :width,
+             :base,
+             :pretty,
+             :safe,
+             :syntax_colors,
+             :inspect_fun
+           ] ->
+        true
+
+      _ ->
+        false
+    end)
+  end
 end
 
 defmodule Inspect.Error do
--- a/lib/elixir/lib/inspect/algebra.ex
+++ b/lib/elixir/lib/inspect/algebra.ex
@@ -101,7 +101,7 @@ defmodule Inspect.Opts do
   @doc """
   Filters out all the default options, returning the custom options only.
   """
-  @spec filter_custom_options(keyword()) :: keyword()
+  @spec filter_custom_options(opts) :: opts when opts: keyword()
   def filter_custom_options(opts) do
     Keyword.drop(opts, [
       :structs,
--- a/lib/elixir/lib/inspect/algebra.ex
+++ b/lib/elixir/lib/inspect/algebra.ex
@@ -61,7 +61,7 @@ defmodule Inspect.Opts do
       options. Useful when implementing the `Inspect` protocol for nested structs
       to pass the custom options through. All the non-default options passed to
       `Kernel.inspect/2` will be put into this list.
-      
+
   """
 
   # TODO: Remove :char_lists key on v2.0
@@ -98,6 +98,10 @@ defmodule Inspect.Opts do
           custom_options: keyword
         }
 
+  @doc """
+  Filters out all the default options, returning the custom options only.
+  """
+  @spec filter_custom_options(keyword()) :: keyword()
   def filter_custom_options(opts) do
     Keyword.drop(opts, [
       :structs,
--- a/lib/elixir/lib/inspect/algebra.ex
+++ b/lib/elixir/lib/inspect/algebra.ex
@@ -59,8 +59,7 @@ defmodule Inspect.Opts do
 
     * `:custom_options` (since v1.9.0) - a keyword list storing custom user-defined
       options. Useful when implementing the `Inspect` protocol for nested structs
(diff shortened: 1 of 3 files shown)


josevalim:
Agreed. We discussed about adding this in the past.

I just propose to rename the field name to `custom_options` or `options`. Let's see what others think.

 /cc @wojtekmach

fertapric:
`custom_options` sounds good to me.

wojtekmach:
+1 for custom_options too but don't have a strong opinion.

am-kantox:
Cool. Pushed `custom_options`.

eksperimental:
Please also don't forget to update the title of the PR accordingly

lib/elixir/lib/inspect/algebra.ex
```diff
@@ -88,8 +93,33 @@ defmodule Inspect.Opts do
           pretty: boolean,
           safe: boolean,
           syntax_colors: [{color_key, IO.ANSI.ansidata()}],
-          inspect_fun: (any, t -> Inspect.Algebra.t())
+          inspect_fun: (any, t -> Inspect.Algebra.t()),
+          custom_options: [{atom, any()}]
```
whatyouhide: This is exactly the definition of `t:keyword/0` :)

```suggestion
          custom_options: keyword
```

lib/elixir/lib/inspect/algebra.ex
```diff
@@ -88,8 +93,33 @@ defmodule Inspect.Opts do
...
         }
+
+  def custom_options(opts) do
+    Enum.reject(opts, fn
```
whatyouhide: You can replace this `Enum.reject/2` call with

```elixir
Keyword.drop(opts, [:structs, ...])
```

lib/elixir/lib/inspect/algebra.ex
```diff
@@ -88,8 +93,33 @@ defmodule Inspect.Opts do
           pretty: boolean,
           safe: boolean,
           syntax_colors: [{color_key, IO.ANSI.ansidata()}],
-          inspect_fun: (any, t -> Inspect.Algebra.t())
+          inspect_fun: (any, t -> Inspect.Algebra.t()),
+          custom_options: [{atom, any()}]
         }
+
+  def custom_options(opts) do
```
whatyouhide: I would call this function something like `filter_custom_options/1`.

lib/elixir/lib/inspect/algebra.ex
```diff
@@ -88,8 +94,33 @@ defmodule Inspect.Opts do
...
+    Enum.reject(opts, fn
+      {k, _}
```
eksperimental:
```suggestion
      {key, _}
```

lib/elixir/lib/inspect/algebra.ex
```diff
@@ -88,8 +94,33 @@ defmodule Inspect.Opts do
...
+      when k in [
```
eksperimental:
```suggestion
      when key in [
```

lib/elixir/lib/inspect/algebra.ex
```diff
@@ -88,8 +94,33 @@ defmodule Inspect.Opts do
           pretty: boolean,
           safe: boolean,
           syntax_colors: [{color_key, IO.ANSI.ansidata()}],
-          inspect_fun: (any, t -> Inspect.Algebra.t())
+          inspect_fun: (any, t -> Inspect.Algebra.t()),
+          custom_options: [{atom, any()}]
         }
+
+  def custom_options(opts) do
```
eksperimental:
```suggestion
  @spec custom_options(keyword()) :: boolean
  def custom_options(opts) do
```

lib/elixir/lib/inspect/algebra.ex
eksperimental: also Keyword.drop/2 is desirable because the function will raise if opts is not a keyword list, and with Enum.reject it will pass as is

lib/elixir/lib/inspect/algebra.ex
eksperimental: @doc """
Filters out all non-default options.
"""

am-kantox: It filters out all the _default_ options, leaving non-default only :)

lib/elixir/lib/inspect/algebra.ex
```diff
@@ -88,8 +94,30 @@ defmodule Inspect.Opts do
...
           syntax_colors: [{color_key, IO.ANSI.ansidata()}],
-          inspect_fun: (any, t -> Inspect.Algebra.t())
+          inspect_fun: (any, t -> Inspect.Algebra.t()),
+          custom_options: keyword
         }
+
+  @doc """
+  Filters out all the default options, returning the custom options only.
+  """
+  @spec filter_custom_options(keyword()) :: keyword()
```
eksperimental: sorry for the noise, but I think this one reads better
```suggestion
  @spec filter_custom_options(opts) :: opts when opts: keyword()
```

am-kantox: No worries, pushed.

lib/elixir/lib/kernel.ex
```diff
@@ -2055,7 +2055,8 @@ defmodule Kernel do
   """
   @spec inspect(Inspect.t(), keyword) :: String.t()
   def inspect(term, opts \\ []) when is_list(opts) do
-    opts = struct(Inspect.Opts, opts)
+    custom_options = Inspect.Opts.filter_custom_options(opts)
```
josevalim: I would like the custom options to be given nested:

    custom_options: [...]

Otherwise, this basically makes it impossible for Elixir to add any new key in the future, as it may now conflict with a a user key.

lib/elixir/lib/inspect/algebra.ex
```diff
@@ -88,8 +94,30 @@ defmodule Inspect.Opts do
...
-          inspect_fun: (any, t -> Inspect.Algebra.t())
+          inspect_fun: (any, t -> Inspect.Algebra.t()),
+          custom_options: keyword
         }
+
+  @doc """
+  Filters out all the default options, returning the custom options only.
+  """
+  @spec filter_custom_options(opts) :: opts when opts: keyword()
+  def filter_custom_options(opts) do
```
josevalim: Per the comment below, this function should not exist. :)

lib/elixir/lib/kernel.ex
am-kantox: I thought a lot about this and I finally came to the conclusion that user’s convenience is more valuable since Elixir API was kinda declared nearly frozen.

Anyways, fixed.

lib/elixir/lib/inspect/algebra.ex
```diff
@@ -57,6 +57,11 @@ defmodule Inspect.Opts do
     * `:inspect_fun` (since v1.9.0) - a function to build algebra documents,
       defaults to `Inspect.inspect/2`
 
+    * `:custom_options` (since v1.9.0) - a keyword list storing custom user-defined
+      options. Useful when implementing the `Inspect` protocol for nested structs
+      to pass the custom options through. All the non-default options passed to
```
fertapric: I think this last sentence should be removed now :)

josevalim:

PR #9099, opened 2019-05-30
Store relative file in module definition

### LINKED ISSUE #9095: Inconsistent filenames in stacktraces after consolidation
### Environment

* Elixir & Erlang/OTP versions (elixir --version): latest master
* Operating system: macos

### Current behavior

Inside of any mix project:
```
$ iex
iex(1)> Enum.map(1, 1)
** (Protocol.UndefinedError) protocol Enumerable not implemented for 1 of type Integer
    (elixir) lib/enum.ex:1: Enumerable.impl_for!/1
    (elixir) lib/enum.ex:141: Enumerable.reduce/3
    (elixir) lib/enum.ex:3023: Enum.map/2
```

```
$ iex -S mix
iex(1)> Enum.map(1, 1)
** (Protocol.UndefinedError) protocol Enumerable not implemented for 1 of type Integer. This protocol is implemented for the following type(s): Ecto.Adapters.SQL.Stream, Postgrex.Stream, DBConnection.Stream, DBConnection.PrepareStream, HashSet, Range, Map, Function, List, Stream, Date.Range, HashDict, GenEvent.Stream, MapSet, File.Stream, IO.Stream
    (elixir) /Users/michalmuskala/src/oss/elixir/lib/elixir/lib/enum.ex:1: Enumerable.impl_for!/1
    (elixir) /Users/michalmuskala/src/oss/elixir/lib/elixir/lib/enum.ex:141: Enumerable.reduce/3
    (elixir) lib/enum.ex:3023: Enum.map/2
```

### Expected behavior

The stacktrace entry should be the same in both cases

josevalim:
Prior to this patch, we would always compute the relative
path to the current working directory (CWD), but this meant
consolidated protocols would always get a full path since
they are always outside of their current working directory.
We address this by also storing the relative path in the
definition.

Closes #9095

### DIFF
--- a/lib/mix/test/mix/tasks/compile.protocols_test.exs
+++ b/lib/mix/test/mix/tasks/compile.protocols_test.exs
@@ -117,6 +117,34 @@ defmodule Mix.Tasks.Compile.ProtocolsTest do
     end)
   end
 
+  test "consolidated protocols keep relative path to their source" do
+    Mix.Project.push(MixTest.Case.Sample)
+
+    in_fixture("no_mixfile", fn ->
+      Mix.Tasks.Compile.Elixir.run([])
+      Mix.Tasks.Compile.Protocols.run([])
+
+      # Load consolidated
+      :code.add_patha('_build/dev/lib/sample/consolidated')
+      :code.purge(Enumerable)
+      :code.delete(Enumerable)
+
+      try do
+        Enumerable.impl_for!(:oops)
+      rescue
+        Protocol.UndefinedError ->
+          assert [{_, _, _, [file: 'lib/enum.ex'] ++ _} | _] = __STACKTRACE__
+      else
+        _ ->
+          flunk("Enum.map/2 should have failed")
+      after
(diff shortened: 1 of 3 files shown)


lib/mix/test/mix/tasks/compile.protocols_test.exs
```diff
@@ -117,6 +117,34 @@ defmodule Mix.Tasks.Compile.ProtocolsTest do
...
+      :code.delete(Enumerable)
+
+      try do
+        Enumerable.impl_for!(:oops)
+      rescue
+        Protocol.UndefinedError ->
+          assert [{_, _, _, [file: 'lib/enum.ex'] ++ _} | _] = __STACKTRACE__
+      else
+        _ ->
+          flunk("Enum.map/2 should have failed")
```
whatyouhide: Minor thing, but I think this should be `impl_for!/1`?

```suggestion
          flunk("Enumerable.impl_for!/1 should have failed")
```

josevalim:

PR #9102, opened 2019-05-31
Handle multiple versions of the same app during releases

### LINKED ISSUE #9098: Handle multiple versions of the same app in OTP during releases

fertapric:
Closes #9098

### DIFF
--- a/lib/mix/test/mix/release_test.exs
+++ b/lib/mix/test/mix/release_test.exs
@@ -110,6 +110,18 @@ defmodule Mix.ReleaseTest do
       end
     end
 
+    test "uses the latest version of an app if there are multiple versions", context do
+      in_tmp(context.test, fn ->
+        File.cp_r!(:code.root_dir(), ".", fn _, _ -> false end)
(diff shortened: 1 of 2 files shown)


lib/mix/test/mix/release_test.exs
```diff
@@ -110,6 +110,18 @@ defmodule Mix.ReleaseTest do
...
 
+    test "uses the latest version of an app if there are multiple versions", context do
+      in_tmp(context.test, fn ->
+        File.cp_r!(:code.root_dir(), ".", fn _, _ -> false end)
+        File.mkdir_p!("lib/compiler-1.0")
+        erts_source = Path.join(File.cwd!(), "erts-#{@erts_version}")
+
+        release = from_config!(nil, config(releases: [demo: [include_erts: erts_source]]), [])
+
+        assert release.applications.compiler[:vsn] != "1.0"
```
josevalim: Let's also assert `assert release.applications.compiler[:path]` points to the current working dir just in case. :)

PR #9104, opened 2019-05-31
Use SPDX full name of Apache License 2.0

chulkilee:
https://spdx.org/licenses/Apache-2.0.html

- Full name: Apache License 2.0
- Short identifier: Apache-2.0

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,6 +209,6 @@ the `doc` directory. If you are planning to contribute documentation,
 
 "Elixir" and the Elixir logo are copyright (c) 2012 Plataformatec.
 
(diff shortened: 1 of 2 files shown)


eksperimental:
Whats the benefit of linking to spdx?

josevalim:
@eksperimental it is not linking to SPDX, only standardizing on its names.

eksperimental:
Yes, I realized after I posted the question. That's why I deleted.
thanks anyway

fertapric:
Thanks @chulkilee! 💜

PR #9105, opened 2019-06-01
Mention custom scripts in mix release "One-off commands" docs

wojtekmach:
I think most projects will end up with some kind of `MyApp.ReleaseTasks` module so I think we could mention it in the docs.

### DIFF
--- a/lib/mix/lib/mix/tasks/release.ex
+++ b/lib/mix/lib/mix/tasks/release.ex
@@ -123,62 +123,34 @@ defmodule Mix.Tasks.Release do
   You can also use `remote` to connect a remote IEx session to the
   system.
 
+  #### Helper module
+
   As you operate your system, you may find yourself running some piece of code
-  as a one-off command quite often and re-typing the whole command every
-  time may become cumbersome and error prone. For example, suppose you need to
-  regularly migrate your database and occasionally build a report of currently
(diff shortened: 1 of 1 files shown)


lib/mix/lib/mix/tasks/release.ex
```diff
@@ -123,6 +123,63 @@ defmodule Mix.Tasks.Release do
...
+      end
+
+  Again, for migrating the database we'd probably use `eval` but to list
+  currently connected users we'd probably connect to a running system using
+  `rpc`:
+
+      bin/RELEASE_NAME eval "MyApp.ReleaseTasks.migrate_db()"
+      bin/RELEASE_NAME rpc "MyApp.ReleaseTasks.print_connected_users()"
+
+  Finally, you may still automate this further by creating standalone scripts
```
wojtekmach: The section below has probably way too much code so maybe it's better to show the functions but let users fill in the blanks, or skip this thing altogether?

josevalim: Yeah, we should probably keep this focused. I have pushed some updates, let me know what you think.

wojtekmach:
@josevalim looks good!

whatyouhide:
Thanks @wojtekmach! 💟

PR #9106, opened 2019-06-01
Typo

leandrocp:
Window server -> Windows server


leandrocp:
Or maybe it should be "Windows service" 🤔

lib/mix/lib/mix/tasks/release.ex
```diff
@@ -78,7 +78,7 @@ defmodule Mix.Tasks.Release do
 
     * Management scripts. Releases come with scripts to start, restart,
       connect to the running system remotely, execute RPC calls, run as
-      daemon, run as a Window server, and more.
+      daemon, run as a Windows server, and more.
```
josevalim:
```suggestion
      daemon, run as a Windows service, and more.
```

josevalim:

Issue #9084, opened 2019-05-27
Defining @Tributes starting with uppercase crashes compilation


fenollp:
### Environment

* Elixir & Erlang/OTP versions (elixir --version): 
```
Erlang/OTP 22 [erts-10.4] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1] [hipe] [dtrace]

Elixir 1.8.2 (compiled with Erlang/OTP 21)
```
* Operating system: 
```
Darwin baobao 18.6.0 Darwin Kernel Version 18.6.0: Thu Apr 25 23:16:27 PDT 2019; root:xnu-4903.261.4~2/RELEASE_X86_64 x86_64 i386 MacBookPro14,1 Darwin
```

### Current behavior

```elixir
defmodule M do
  @Bla 42

  def bla, do: @Bla
end
```
```
== Compilation error in file lib/m.ex ==
** (SyntaxError) lib/m.ex:2: syntax error before: "42"
    (elixir) lib/kernel/parallel_compiler.ex:208: anonymous fn/4 in Kernel.ParallelCompiler.spawn_workers/6
```

### Expected behaviour

Either
1. handle module attributes that have the first grapheme of their name capitalized (if they are atoms underneath, quote them?) Both compiler & formatter are affected.
1. report the error with a more specific message along the lines of `Module attributes cannot be capitalized. See @Bla in lib/m.ex:2`

josevalim:
The error is happening in the parser, which means we don't even have a notion of module attributes yet. The issue is that you are trying to write this: `@Bar 2`, which is the same as `+Bar 2`, which we can reduce to `Bar 2`, and it shows that you can't pass arguments to aliases because you can't "call" them.

I understand the error message is quite bad but all error messages coming from the parser are going to be this bad, unfortunately, and generally speaking there isn't much we can do unless we rewrite the whole parser in a tool that I am not sure exists at the moment with better error messages. Sorry. :(

fenollp:
Wait. How is `@Bar 2` the same as `+Bar 2`?

josevalim:
@ in Elixir is a unary operator, like +.

fenollp:
Alright but what makes `@` ====== `+`? As I understand it the reason `@` disappears is because it gets turned into a plus sign?

[...]

PR #9109, opened 2019-06-03
Clarify syntax for custom Module attributes

fertapric:
Related to https://github.com/elixir-lang/elixir/issues/9084

### DIFF
--- a/lib/elixir/lib/module.ex
+++ b/lib/elixir/lib/module.ex
@@ -345,8 +345,9 @@ defmodule Module do
   ### Custom attributes
 
   In addition to the built-in attributes outlined above, custom attributes may
-  also be added. A custom attribute is any valid identifier prefixed with an
-  `@` and followed by a valid Elixir value:
+  also be added. Custom attributes are expressed using the `@` operator followed
+  by a variable. The value given to the custom attribute must be a valid Elixir 
+  value:
 
       defmodule MyModule do
         @custom_attr [some: "stuff"]
--- a/lib/elixir/lib/module.ex
+++ b/lib/elixir/lib/module.ex
@@ -345,9 +345,9 @@ defmodule Module do
   ### Custom attributes
 
   In addition to the built-in attributes outlined above, custom attributes may
-  also be added. Custom attributes are expressed using the `@` operator followed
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/module.ex
```diff
@@ -345,8 +345,9 @@ defmodule Module do
   ### Custom attributes
 
   In addition to the built-in attributes outlined above, custom attributes may
-  also be added. A custom attribute is any valid identifier prefixed with an
-  `@` and followed by a valid Elixir value:
+  also be added. Custom attributes are expressed using the `@` operator followed
```
whatyouhide:
```suggestion
  also be added. Custom attributes are expressed using the `@/1` operator followed
```

This should link, right?

lib/elixir/lib/module.ex
```diff
@@ -345,8 +345,9 @@ defmodule Module do
...
+  by a variable. The value given to the custom attribute must be a valid Elixir
```
fertapric: Maybe better?

```suggestion
  by a variable expression. The value given to the custom attribute must be a valid Elixir 
```

josevalim: name perhaps?

```suggestion
  by a variable name. The value given to the custom attribute must be a valid Elixir 
```

fertapric: I used "expression" because that's the term we use in the "Syntax reference" guide. https://hexdocs.pm/elixir/syntax-reference.html#expressions

Wdyt?

> In addition to the built-in attributes outlined above, custom attributes may also be added. 
>
> Custom attributes are expressed using the `@` operator followed by a variable expression. 
> See the [Syntax Reference](syntax-reference.html) documentation for more information.
>
> The value given to the custom attribute must be a valid Elixir value:

josevalim: The module attribute syntax allows @ followed by **any expression**, not only variable ones. But the default implementation allows only variable names. So we need to fix the syntax reference too. :) I would simply say:

> Custom attributes are expressed using the @ operator followed by an expression.

fertapric: I went with "valid variable name". "an expression" might be more accurate, but I think "a valid variable name" is going to be less confusing for final users 🙂

Issue #9087, opened 2019-05-28
IEx: Autocompleting types from OTP doesn't work


wojtekmach:
### Environment

* Elixir & Erlang/OTP versions (elixir --version): `Erlang/OTP 22 [erts-10.4] [source] [64-bit] Elixir 1.9.0-dev (a8062a8) (compiled with Erlang/OTP 22)`
* Operating system: macOS

### Current behavior

Autocompleting functions works:

```
iex> :file.<tab>
advise/4                    allocate/3
(...)
```

But not types:

```
iex> t :file.<tab>
*nothing*
```

Note, listing types already works:

```
iex> t :file
@type filename() :: string()

@type filename_all() :: string() | binary()

(...)
```

### Expected behavior

Autocompleting types works.

josevalim:
@wojtekmach I believe it is because we use the Docs chunk? But I guess we can traverse the abstract code chunk instead.

josevalim:
Closing in favor of the PR!

PR #9110, opened 2019-06-03
iex: Erlang Module Type Autocomplete

maennchen:
Fix for #9087

### DIFF
--- a/lib/iex/lib/iex/autocomplete.ex
+++ b/lib/iex/lib/iex/autocomplete.ex
@@ -128,7 +128,7 @@ defmodule IEx.Autocomplete do
   defp expand_custom([?. | expr], server, fun) do
     case Code.string_to_quoted(reduce(expr)) do
       {:ok, atom} when is_atom(atom) ->
-        no()
+        expand_elixir_module_custom(atom, "", fun)
 
       {:ok, {:__aliases__, _, [h | _] = list}} when is_atom(h) ->
         case expand_alias(list, server) do
@@ -165,6 +165,9 @@ defmodule IEx.Autocomplete do
           :error ->
             no()
         end
+
+      {:ok, {{:., _, [module, type]}, _, []}} when is_atom(type) ->
+        expand_elixir_module_custom(module, Atom.to_string(type), fun)
 
       _ ->
         no()
@@ -440,15 +443,18 @@ defmodule IEx.Autocomplete do
   end
 
   defp get_module_types(mod) do
-    cond do
-      not ensure_loaded?(mod) ->
-        []
+    if ensure_loaded?(mod) do
+      case Code.Typespec.fetch_types(mod) do
+        {:ok, specs} ->
+          Enum.map(specs, fn {_kind, {name, _, args}} ->
+            {name, length(args)}
+          end)
 
-      docs = get_docs(mod, [:type]) ->
-        Enum.map(docs, &extract_name_and_arity/1)
-
-      true ->
-        exports(mod)
+        :error ->
+          []
+      end
+    else
+      []
     end
   end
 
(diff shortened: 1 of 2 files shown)


lib/iex/lib/iex/autocomplete.ex
```diff
@@ -440,15 +442,23 @@ defmodule IEx.Autocomplete do
...
+      false ->
         []
 
-      docs = get_docs(mod, [:type]) ->
-        Enum.map(docs, &extract_name_and_arity/1)
+      nil ->
+        case Typespec.fetch_types(mod) do
+          {:ok, specs} ->
+            Enum.map(specs, fn {_kind, {name, _, _}} ->
+              {name, 0}
```
wojtekmach: we shouldn't hardcode arity to zero, instead you can do:

```elixir
Enum.map(specs, fn {_kind, {name, _, args}} ->
  {name, length(args)}
```

lib/iex/test/iex/autocomplete_test.exs
```diff
@@ -65,6 +65,7 @@ defmodule IEx.AutocompleteTest do
     assert expand('t String.') == {:yes, '', ['codepoint/0', 'grapheme/0', 'pattern/0', 't/0']}
     assert expand('t String.grap') == {:yes, 'heme', []}
     assert expand('t  String.grap') == {:yes, 'heme', []}
+    assert {:yes, '', ['common_reason/0' | _]} = expand('t :ftp.')
```
wojtekmach: please add some more coverage:

```elixir
expand('t :fil') # it shows: file, filename etc - should it show: :file, :filename etc?
expand('t :ftp.comm') # doesn't work yet
```

wojtekmach:
Threre's one thing somewhat related to this PR which is how we handle `typep`s, currently we display them in the list:

```elixir
iex(2)> t :filelib
@typep filename() :: :file.name()

@typep dirname() :: filename()

@typep filename_all() :: :file.name_all()

@typep dirname_all() :: filename_all()

@typep find_file_rule() :: {objDirSuffix :: string(), srcDirSuffix :: string()}

@typep find_source_rule() ::
         {objExtension :: string(), srcExtension :: string(),
          [find_file_rule()]}
```

but if we want to show just that type, it errors given its private:

```
iex(3)> t :filelib.filename
No type information for :filelib.filename was found or :filelib.filename is private
```

for this reason, should we not list them and similarly not autocomplete them?

maennchen:
@wojtekmach This should only concern `typep`, right?

All the other types like `type` & `opaque` should be in the list I guess.

wojtekmach:
Yes, I think it's ok to list and autocomplete opaque. Btw, this concern is I think out of the scope of this PR so we don't have to change it now, but I just wanted to mention it since it's contextual.

maennchen:
@wojtekmach Another problem I see here is that there's almost no `export_type` declaration in erlang itself. Is this on purpose?

wojtekmach:
> Another problem I see here is that there's almost no export_type declaration in erlang itself. Is this on purpose?

Can you elaborate? I see plenty of export_type usage in otp source code.

lib/iex/lib/iex/autocomplete.ex
maennchen: Is this documented somewhere? I tried to guess what is what in the AST because I found no resources about the structure.

wojtekmach: I don't think it is, maybe somewhere in Erlang docs. If Code.Typespec module ever becomes public, I think that'd be a good place to document it.

lib/iex/lib/iex/autocomplete.ex
```diff
@@ -1,6 +1,8 @@
 defmodule IEx.Autocomplete do
   @moduledoc false
 
+  alias Code.Typespec
```
josevalim: We use it once, no need to alias. :)

lib/iex/lib/iex/autocomplete.ex
```diff
@@ -440,15 +445,17 @@ defmodule IEx.Autocomplete do
   end
 
   defp get_module_types(mod) do
-    cond do
-      not ensure_loaded?(mod) ->
+    with true <- ensure_loaded?(mod),
```
josevalim: Can we please convert this to two `case`s instead of a `with`? The reason is because we are matching on different results `false` and `:error`, I would rather keep it close to the caller instead of flattening everything in a `with`.

maennchen: Since `ensure_loaded` returns a boolean, i replaced it with an `if` and a `case`. Is that ok?

josevalim:

PR #9113, opened 2019-06-03
Use :timer.tc/1 to measure time diff in tests

fertapric:
The test `"timer/1"` has been failing intermittently in the CI builds https://cirrus-ci.com/task/4511441067966464. I thought using `:timer.tc/1` to measure time differences would be better than `:timer.now_diff/2`

### DIFF
--- a/lib/elixir/test/elixir/stream_test.exs
+++ b/lib/elixir/test/elixir/stream_test.exs
@@ -494,9 +494,11 @@ defmodule StreamTest do
 
   test "interval/1" do
     stream = Stream.interval(10)
-    now = :os.timestamp()
-    assert Enum.take(stream, 5) == [0, 1, 2, 3, 4]
-    assert :timer.now_diff(:os.timestamp(), now) >= 50000
+
+    {time_µs, value} = :timer.tc(fn -> Enum.take(stream, 5) end)
+
+    assert value == [0, 1, 2, 3, 4]
+    assert time_µs >= 50000
   end
 
   test "into/2 and run/1" do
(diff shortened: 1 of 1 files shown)


lib/elixir/test/elixir/stream_test.exs
```diff
@@ -494,9 +494,11 @@ defmodule StreamTest do
 
   test "interval/1" do
     stream = Stream.interval(10)
-    now = :os.timestamp()
-    assert Enum.take(stream, 5) == [0, 1, 2, 3, 4]
-    assert :timer.now_diff(:os.timestamp(), now) >= 50000
+
+    {time_µs, value} = :timer.tc(fn -> Enum.take(stream, 5) end)
```
whatyouhide: `µs` is awesome but also super yuck 😄 I am not sure we should use it in the Elixir codebase outside of testing for Unicode variables because it might make it harder to contribute or to understand why this works for newcomers. Thoughts?

devonestes:
As someone who has done a lot of this kind of testing, asserting on execution time is always something that is highly prone to false positives.

Normally I would recommend reduction counting instead, but that doesn't appear to be helpful for these tests, so I would actually recommend that you implement a retry for these tests so basically attempt this test twice and hope that it passes one of those two times. That way you don't have to rebuild the entire CI run for tests that are flaky by the very nature.

I can't say for sure, but I would expect that using `:timer.tc/1` will not actually reduce the flakiness at all.

lib/elixir/test/elixir/stream_test.exs
josevalim: Yeah, let's go with `time_us`. :)

fertapric: Changed :)

josevalim:
@devonestes those are good points. This patch is at least moving it from system_time to monotonic_time, so i would give them a try. If the issue persists, then a retry approach is a good idea.

fertapric:
@devonestes Thanks for your comments! 💜

PR #9116, opened 2019-06-05
Update Enum.find/2 documentation

JorisKok:
I found the order of the examples a bit confusing, because on other examples it starts with the 'success' examples first. The default 0 example in the middle had me looking a few times before I understood what the result should be.

### DIFF
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -939,14 +939,14 @@ defmodule Enum do
 
   ## Examples
 
-      iex> Enum.find([2, 4, 6], fn x -> rem(x, 2) == 1 end)
-      nil
-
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/enum.ex
```diff
@@ -939,15 +939,15 @@ defmodule Enum do
 
   ## Examples
 
+      iex> Enum.find([2, 3, 4], fn x -> rem(x, 2) == 1 end)
+      3
+      
       iex> Enum.find([2, 4, 6], fn x -> rem(x, 2) == 1 end)
       nil
-
+
```
josevalim: Please remove trailing whitespace.

```suggestion

```

josevalim:

PR #9118, opened 2019-06-05
Add preferred_cli_target

mobileoverlord:
This PR addresses a common issue with running mix tasks, primarily `mix test` when `MIX_TARGET` is set to a value other than `:host`. This is why the behavior mimics that of `preferred_cli_env` with the exception that this will not be able to be overridden by setting `MIX_TARGET`. I feel that this behavior is okay because `mix` is a tool that is _always_ running on your `:host`. Therefore, you should be able to specify, in the task module or project config which target the tasks should run under, no matter what the target. If this behavior needs to be overridden, it could be done by setting `preferred_cli_target` in the `Mix.Project.config`.

The common pain is that `MIX_TARGET` is typically something that you `export` which then makes running `mix test` painful because it attempts to load dependencies that may be set for `targets: @not_host`. This inevitably leads to `MIX_TARGET=host mix test` which adds a hurdle for promoting good test practices.

### DIFF
--- a/lib/mix/lib/mix/cli.ex
+++ b/lib/mix/lib/mix/cli.ex
@@ -29,7 +29,7 @@ defmodule Mix.CLI do
     load_mix_exs()
     {task, args} = get_task(args)
     ensure_hex(task)
-    change_env(task)
+    maybe_change_env_and_target(task)
     run_task(task, args)
   end
 
@@ -104,24 +104,37 @@ defmodule Mix.CLI do
     end
   end
 
-  defp change_env(task) do
-    if env = preferred_cli_env(task) do
-      Mix.env(env)
+  defp maybe_change_env_and_target(task) do
+    task = String.to_atom(task)
+    config = Mix.Project.config()
 
-      if project = Mix.Project.pop() do
-        %{name: name, file: file} = project
-        Mix.Project.push(name, file)
-      end
+    env = preferred_cli_env(task, config)
+    target = preferred_cli_target(task, config)
+    env && Mix.env(env)
+    target && Mix.target(target)
+
+    if env || target do
+      reload_project()
     end
   end
 
-  defp preferred_cli_env(task) do
+  defp reload_project() do
+    if project = Mix.Project.pop() do
+      %{name: name, file: file} = project
+      Mix.Project.push(name, file)
+    end
+  end
+
+  defp preferred_cli_env(task, config) do
     if System.get_env("MIX_ENV") do
       nil
     else
-      task = String.to_atom(task)
-      Mix.Project.config()[:preferred_cli_env][task] || Mix.Task.preferred_cli_env(task)
+      config[:preferred_cli_env][task] || Mix.Task.preferred_cli_env(task)
     end
+  end
+
+  defp preferred_cli_target(task, config) do
+    config[:preferred_cli_target][task]
   end
 
   defp load_dot_config do
--- a/lib/mix/lib/mix/project.ex
+++ b/lib/mix/lib/mix/project.ex
@@ -67,8 +67,12 @@ defmodule Mix.Project do
     * `:preferred_cli_env` - a keyword list of `{task, env}` tuples where `task`
       is the task name as an atom (for example, `:"deps.get"`) and `env` is the
       preferred environment (for example, `:test`). This option overrides what
-      specified by the tasks with the `@preferred_cli_env` attribute (see the
+      is specified by the tasks with the `@preferred_cli_env` attribute (see the
       docs for `Mix.Task`). Defaults to `[]`.
+
+    * `:preferred_cli_target` - a keyword list of `{task, target}` tuples where
+      `task` is the task name as an atom (for example, `:test`) and `target`
+      is the preferred target (for example, `:host`). Defaults to `[]`.
 
   For more options, keep an eye on the documentation for single Mix tasks; good
   examples are the `Mix.Tasks.Compile` task and all the specific compiler tasks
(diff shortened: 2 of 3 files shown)


lib/mix/lib/mix/cli.ex
```diff
@@ -115,6 +116,17 @@ defmodule Mix.CLI do
     end
   end
 
+  defp change_target(task) do
+    if target = preferred_cli_target(task) do
+      Mix.target(target)
+
+      if project = Mix.Project.pop() do
+        %{name: name, file: file} = project
+        Mix.Project.push(name, file)
```
josevalim: Because this operation busts a bunch of cache, I think we should unify change_env and change_target into a single function and pop+push the project only once.

mobileoverlord: I updated this to only reload the project once if either of the two execute.

lib/mix/lib/mix/cli.ex
```diff
@@ -29,7 +29,11 @@ defmodule Mix.CLI do
     load_mix_exs()
     {task, args} = get_task(args)
     ensure_hex(task)
-    change_env(task)
+
+    if Enum.any?([change_env(task), change_target(task)]) do
```
josevalim:
```suggestion
    if change_env(task) || change_target(task) do
```

lib/mix/lib/mix/cli.ex
```diff
@@ -104,23 +104,40 @@ defmodule Mix.CLI do
...
       nil
     else
-      task = String.to_atom(task)
-      Mix.Project.config()[:preferred_cli_env][task] || Mix.Task.preferred_cli_env(task)
+      config[:preferred_cli_env][task] || Mix.Task.preferred_cli_env(task)
+    end
+  end
+
+  defp preferred_cli_target(task, config) do
+    if System.get_env("MIX_TARGET") do
```
mobileoverlord: This no longer functions in a desirable way. Since MIX_TARGET is almost always set when cross compiling, this code almost always has no effect. I this condition needs to be eliminated and it should always check the config and task.

josevalim: Hrm, this means then there is no way for me to override a default imposed by the task, and I think we ultimately need to provide a way for the developer to override that.

Why are you setting MIX_TARGET by hand? I understand that originally we had to do it to support both old and new Elixir versions alike... but maybe you can always set the target via `Mix.target` and leave the MIX_TARGET environment variable for only when users set it?

mobileoverlord: The user almost always has `MIX_TARGET` exported when working on a project that is cross compiled since they are spending a majority of their time compiling things for the target. Otherwise they would be running `MIX_TARGET=rpi3 mix compile` `MIX_TARGET=rpi3 mix firmware` `MIX_TARGET=rpi3 mix burn` all the time. The exception comes in for `mix run` and `mix test` since these are both situations where the execution should be for the host.

If the user wants to override the behavior I suggested above, they could set it to the original target in the mix file. `preferred_cli_target: [test: Mix.target()]` But I would argue that this would be highly unlikely in the case of `mix test`

If adding `@preferred_cli_target` to mix test feels too controversial, then we could add it to the mix file as a suggested default.

josevalim:
Good idea! Let’s remove the changes in Mix.Task then and have this be
driven exclusively by the mix.exs for now.
-- 


mobileoverlord:
I've updated this with the changes
* Always change target if `preferred_cli_target` is configured.
* Removed `@preferred_cli_target` from `mix test`

mobileoverlord:
After re reading your last comment, I think what you are suggesting is that we remove the ability for a task to set `@preferred_cli_target` is that correct?

mobileoverlord:
Okay, I removed `@preferred_cli_target` from `Mix.Task`

josevalim:
@mobileoverlord thanks! PR looks great but CI is failing. :)

mobileoverlord:
I fixed the failing test and rebased this PR

lib/mix/lib/mix/project.ex
```diff
@@ -67,9 +67,15 @@ defmodule Mix.Project do
     * `:preferred_cli_env` - a keyword list of `{task, env}` tuples where `task`
       is the task name as an atom (for example, `:"deps.get"`) and `env` is the
       preferred environment (for example, `:test`). This option overrides what
-      specified by the tasks with the `@preferred_cli_env` attribute (see the
+      is specified by the tasks with the `@preferred_cli_env` attribute (see the
       docs for `Mix.Task`). Defaults to `[]`.
 
+    * `:preferred_cli_target` - a keyword list of `{task, target}` tuples where
+      `task` is the task name as an atom (for example, `:"test"`) and `target`
```
fertapric: Those quotes are unnecessary, and newer Elixir versions warn in such cases (and the formatter removes them) 🙂 

```suggestion
      `task` is the task name as an atom (for example, `:test`) and `target`
```

lib/mix/lib/mix/project.ex
```diff
@@ -67,9 +67,15 @@ defmodule Mix.Project do
...
+      is the preferred target (for example, `:host`). This option overrides what
```
fertapric: I think the sentence "This option overrides what ..." should be removed since the support for the `@preferred_cli_target` attribute was dropped for now 🙂

lib/mix/lib/mix/project.ex
mobileoverlord: AH good catch!

mobileoverlord:
@fertapric Updated with the requested documentation changes

josevalim:

PR #9120, opened 2019-06-08
--paths= in rebar3 bare compile fixes subcommand splitting on comma bug

tsloughter:
@ferd breaks down the issue here https://github.com/erlang/rebar3/issues/2102#issuecomment-500081989

But basically we need to use `--paths=` to get around an issue with how we use subscommands and getopt that can't at this time be fixed on the rebar3 side.


josevalim:

lukaszsamson:
@josevalim can you release 1.8 with this patch applied?

josevalim:
Once the .RC is out for a new version we typically stop porting bug fixes to the previous version. v1.9-rc is already out and the final should be out in a week.

PR #9121, opened 2019-06-08
iex: Autocomplete Erlang callbacks

wojtekmach:


lib/iex/lib/iex/autocomplete.ex
```diff
@@ -459,15 +459,20 @@ defmodule IEx.Autocomplete do
   end
 
   defp get_module_callbacks(mod) do
-    cond do
```
fertapric: Nitpick: you can use `with` here

```elixir
defp get_module_callbacks(mod) do
  with true <- ensure_loaded?(mod),
       {:ok, callbacks} <- Code.Typespec.fetch_callbacks(mod) do
    for {name_arity, _} <- callbacks do
      {_kind, name, arity} = IEx.Introspection.translate_callback_name_arity(name_arity)
      {name, arity}
    end
  else
    _ -> []
  end
end
```

wojtekmach: @fertapric see https://github.com/elixir-lang/elixir/pull/9110#discussion-diff-290219277R448 :D

fertapric: 😅 Ok, thanks!

fertapric:
Thanks @wojtekmach! 💜

Issue #2612, opened 2014-07-31
private and opaque types details showing up in Iex


wkhere:
Here's what we have:

``` Elixir
% cat types.ex      
defmodule Foo do
  @type   public_t :: integer
  @typep  priv_t   :: integer
  @opaque opaque_t :: integer
end
% iex               
Erlang/OTP 17 [erts-6.1] [source] [async-threads:10] [hipe] [kernel-poll:false]

Interactive Elixir (0.14.3) - press Ctrl+C to exit (type h() ENTER for help)
iex(1)> c("types.ex")
types.ex:3: warning: type priv_t() is unused
[Foo]
iex(2)> t Foo
@opaque opaque_t() :: integer()
@typep priv_t() :: integer()
@type public_t() :: integer()
```

Well opaque and private types do make a difference for dialyzer. But for clarity, shouldn't their details be silenced in IEx?

Like:

```
iex> t Foo
@opaque opaque_t()
@typep priv_t()
@type public_t() :: integer()
```

PR #9122, opened 2019-06-08
iex: Do not autocomplete private types

wojtekmach:
Ref: https://github.com/elixir-lang/elixir/pull/9110#issuecomment-498299548

> There's one thing somewhat related to this PR which is how we handle typeps, currently we display them in the list:
>
>     iex(2)> t :filelib
>     @typep filename() :: :file.name()

> but if we want to show just that type, it errors given its private:
>
>     iex(3)> t :filelib.filename
>     No type information for :filelib.filename was found or :filelib.filename is private

Btw, according to https://github.com/elixir-lang/elixir/issues/2612 we still want to list typeps, in this PR we'll just not autocomplete them.


fertapric:
Thanks @wojtekmach! 💜

PR #9125, opened 2019-06-10
Add DateTime.now!/2

wojtekmach:


lib/elixir/lib/calendar/datetime.ex
```diff
@@ -504,6 +506,39 @@ defmodule DateTime do
...
+  def now!(time_zone, time_zone_database \\ Calendar.get_time_zone_database()) do
+    case now(time_zone, time_zone_database) do
+      {:ok, datetime} ->
+        datetime
+
+      {:error, reason} ->
+        raise ArgumentError,
+              "cannot get current datetime in #{inspect(time_zone)} time zone, reason: #{
+                inspect(reason)
+              }"
```
josevalim:
```
"cannot get current datetime in #{inspect(time_zone)} time zone, reason: " <>
  inspect(reason)
```

wojtekmach: yeah, already changed but forgot to push!

whatyouhide:
Thanks @wojtekmach! 💟

PR #9126, opened 2019-06-10
Add DateTime.shift_zone!/3

wojtekmach:


lib/elixir/lib/calendar/datetime.ex
```diff
@@ -471,6 +471,32 @@ defmodule DateTime do
...
+    case shift_zone(datetime, time_zone, time_zone_database) do
+      {:ok, datetime} ->
+        datetime
+
+      {:error, reason} ->
+        raise ArgumentError,
+              "cannot shift #{inspect(datetime)} to #{inspect(time_zone)} time zone" <>
+                ", reason: #{inspect(reason)}"
+    end
+  end
```
josevalim: We need to add a doctest or a test that verifies the error case.

whatyouhide:
Thanks @wojtekmach! 💟

PR #9128, opened 2019-06-10
Add inspect/2 callback for Calendar

kipcole9:
PR based upon the conversation at https://elixirforum.com/t/how-to-support-multiple-week-calendars-in-elixir/18783/10

Date, Time, DateTime and NaiveDateTime implement the
`Inspect` protocol with a specific implementation
for `Calendar.ISO`.

This commit introduces a required callback, `inspect/2`
required in all modules implementing the `Calendar` 
behaviour.

Calendars implementing `inspect/2` must ensure
the implementation caters for `Date`, `Time`, `DateTime` and
`NaiveDateTime` structs.

### DIFF
--- a/lib/elixir/lib/calendar.ex
+++ b/lib/elixir/lib/calendar.ex
@@ -271,7 +271,7 @@ defmodule Calendar do
   The implementation must define inspect for `Date.t()`, `DateTime.t()` and
   `NaiveDateTime.t()` structs.
   """
-  @callback inspect(Time.t() | Date.t() | DateTime.t() | MaiveDateTime.t(), Inspect.Opts.t()) ::
+  @callback inspect(Time.t() | Date.t() | DateTime.t() | NaiveDateTime.t(), Inspect.Opts.t()) ::
               String.t()
 
   # General Helpers
--- a/lib/elixir/lib/calendar.ex
+++ b/lib/elixir/lib/calendar.ex
@@ -291,7 +291,7 @@ defmodule Calendar do
               microsecond,
               Inspect.Opts.t()
             ) ::
-              String.t()
+              Inspect.Algebra.t()
 
   @doc """
   Implements inspect for a datetime.
--- a/lib/elixir/lib/calendar.ex
+++ b/lib/elixir/lib/calendar.ex
@@ -267,19 +267,17 @@ defmodule Calendar do
 
   @doc """
   Implements inspect for a date.
-
   """
   @callback inspect_date(year, month, day, Inspect.Opts.t()) :: Inspect.Algebra.t()
 
   @doc """
   Implements inspect for a time.
-
   """
-  @callback inspect_time(hour, minute, second, microsecond, Inspect.Opts.t()) :: Inspect.Algebra.t()
+  @callback inspect_time(hour, minute, second, microsecond, Inspect.Opts.t()) ::
+              Inspect.Algebra.t()
 
   @doc """
   Implements inspect for a naive datetime.
-
   """
   @callback inspect_naive_datetime(
               year,
@@ -295,7 +293,6 @@ defmodule Calendar do
 
   @doc """
   Implements inspect for a datetime.
-
   """
   @callback inspect_datetime(
               year,
--- a/lib/elixir/lib/calendar.ex
+++ b/lib/elixir/lib/calendar.ex
@@ -265,6 +265,15 @@ defmodule Calendar do
   """
   @callback valid_time?(hour, minute, second, microsecond) :: boolean
 
+  @doc """
+  Implements inspect for dates, datetimes and naive_datetimes.
+
+  If implemented, must define inspect for `Date.t()`, `DateTime.t()` and
+  `NaiveDateTime.t()` structs.
+  """
+  @callback inspect(date | datetime | naive_datetime, Inspect.Opts.t()) :: String.t()
+  @optional_callbacks inspect: 2
+
   # General Helpers
 
   @doc """
--- a/lib/elixir/lib/calendar.ex
+++ b/lib/elixir/lib/calendar.ex
@@ -269,7 +269,7 @@ defmodule Calendar do
   Implements inspect for a date.
 
   """
-  @callback inspect_date(year, month, day, Inspect.Opts.t()) :: String.t()
+  @callback inspect_date(year, month, day, Inspect.Opts.t()) :: Inspect.Algebra.t()
 
   @doc """
   Implements inspect for a time.
--- a/lib/elixir/lib/calendar.ex
+++ b/lib/elixir/lib/calendar.ex
@@ -266,13 +266,51 @@ defmodule Calendar do
   @callback valid_time?(hour, minute, second, microsecond) :: boolean
 
   @doc """
-  Implements inspect for times, dates, datetimes, and "naive" datetimes.
+  Implements inspect for a date.
 
-  The implementation must define inspect for `Date.t()`, `DateTime.t()` and
-  `NaiveDateTime.t()` structs.
   """
-  @callback inspect(Time.t() | Date.t() | DateTime.t() | NaiveDateTime.t(), Inspect.Opts.t()) ::
+  @callback inspect_date(year, month, day, Inspect.Opts.t()) :: String.t()
+
+  @doc """
+  Implements inspect for a time.
+
+  """
+  @callback inspect_time(hour, minute, second, microsecond, Inspect.Opts.t()) :: String.t()
+
+  @doc """
+  Implements inspect for a naive datetime.
+
+  """
+  @callback inspect_naive_datetime(
+              year,
+              month,
+              day,
+              hour,
+              minute,
+              second,
+              microsecond,
+              Inspect.Opts.t()
+            ) ::
               String.t()
+
+  @doc """
+  Implements inspect for a datetime.
+
+  """
+  @callback inspect_datetime(
+              year,
+              month,
+              day,
+              hour,
+              minute,
+              second,
+              microsecond,
+              time_zone,
+              zone_abbr,
+              utc_offset,
+              std_offset,
+              Inspect.Opts.t()
+            ) :: String.t()
 
   # General Helpers
 
--- a/lib/elixir/lib/calendar.ex
+++ b/lib/elixir/lib/calendar.ex
@@ -310,7 +310,7 @@ defmodule Calendar do
               utc_offset,
               std_offset,
               Inspect.Opts.t()
-            ) :: String.t()
+            ) :: Inspect.Algebra.t()
 
   # General Helpers
 
--- a/lib/elixir/lib/calendar.ex
+++ b/lib/elixir/lib/calendar.ex
@@ -271,7 +271,8 @@ defmodule Calendar do
   The implementation must define inspect for `Date.t()`, `DateTime.t()` and
   `NaiveDateTime.t()` structs.
   """
-  @callback inspect(time | date | datetime | naive_datetime, Inspect.Opts.t()) :: String.t()
+  @callback inspect(Time.t() | Date.t() | DateTime.t() | MaiveDateTime.t(), Inspect.Opts.t()) ::
+              String.t()
 
   # General Helpers
 
--- a/lib/elixir/lib/calendar.ex
+++ b/lib/elixir/lib/calendar.ex
@@ -275,7 +275,7 @@ defmodule Calendar do
   Implements inspect for a time.
 
   """
-  @callback inspect_time(hour, minute, second, microsecond, Inspect.Opts.t()) :: String.t()
+  @callback inspect_time(hour, minute, second, microsecond, Inspect.Opts.t()) :: Inspect.Algebra.t()
 
   @doc """
   Implements inspect for a naive datetime.
--- a/lib/elixir/lib/calendar.ex
+++ b/lib/elixir/lib/calendar.ex
@@ -268,11 +268,10 @@ defmodule Calendar do
   @doc """
   Implements inspect for dates, datetimes and naive_datetimes.
 
-  If implemented, must define inspect for `Date.t()`, `DateTime.t()` and
+  The implementation must define inspect for `Date.t()`, `DateTime.t()` and
   `NaiveDateTime.t()` structs.
   """
   @callback inspect(date | datetime | naive_datetime, Inspect.Opts.t()) :: String.t()
-  @optional_callbacks inspect: 2
 
   # General Helpers
 
--- a/lib/elixir/lib/calendar/iso.ex
+++ b/lib/elixir/lib/calendar/iso.ex
@@ -716,36 +716,54 @@ defmodule Calendar.ISO do
   end
 
   @doc """
-  Implements the inspect protocol delegated from
-  `Date`, `Time`, `DateTime` and `NaiveDateTime`.
+  Implements the inspect protocol for a date
+  delegated from `Date`.
 
   """
   @doc since: "1.10.0"
   @impl true
-  @spec inspect(
-          Date.t() | Time.t() | DateTime.t() | NaiveDateTime.t(),
-          Inspect.Opts.t()
-        ) :: String.t()
-  def inspect(%Date{calendar: __MODULE__} = date, _) do
-    %{year: year, month: month, day: day} = date
+  @spec inspect_date(Calendar.year(), Calendar.month(), Calendar.day(), Inspect.Opts.t()) ::
+          String.t()
+  def inspect_date(year, month, day, _) do
     "~D[" <> date_to_string(year, month, day) <> "]"
   end
 
-  def inspect(%DateTime{calendar: __MODULE__} = datetime, _) do
-    %{
-      year: year,
-      month: month,
-      day: day,
-      hour: hour,
-      minute: minute,
-      second: second,
-      microsecond: microsecond,
-      time_zone: time_zone,
-      zone_abbr: zone_abbr,
-      utc_offset: utc_offset,
-      std_offset: std_offset
-    } = datetime
+  @doc """
+  Implements the inspect protocol for a datetime
+  delegated from `DateTime`.
 
+  """
+  @doc since: "1.10.0"
+  @impl true
+  @spec inspect_datetime(
+          Calendar.year(),
+          Calendar.month(),
+          Calendar.day(),
+          Calendar.hour(),
+          Calendar.minute(),
+          Calendar.second(),
+          Calendar.microsecond(),
+          Calendar.time_zone(),
+          Calendar.zone_abbr(),
+          Calendar.utc_offset(),
+          Calendar.std_offset(),
+          Inspect.Opts.t()
+        ) ::
+          String.t()
+  def inspect_datetime(
+        year,
+        month,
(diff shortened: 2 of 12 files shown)


lib/elixir/lib/calendar.ex
```diff
@@ -265,6 +265,15 @@ defmodule Calendar do
...
   @callback valid_time?(hour, minute, second, microsecond) :: boolean
 
+  @doc """
+  Implements inspect for dates, datetimes and naive_datetimes.
+
+  If implemented, must define inspect for `Date.t()`, `DateTime.t()` and
+  `NaiveDateTime.t()` structs.
+  """
+  @callback inspect(date | datetime | naive_datetime, Inspect.Opts.t()) :: String.t()
+  @optional_callbacks inspect: 2
```
josevalim: Let's make it a required callback please. :)

lib/elixir/lib/calendar.ex
```diff
@@ -265,6 +265,14 @@ defmodule Calendar do
   """
   @callback valid_time?(hour, minute, second, microsecond) :: boolean
 
+  @doc """
+  Implements inspect for dates, datetimes and naive_datetimes.
+
+  The implementation must define inspect for `Date.t()`, `DateTime.t()` and
+  `NaiveDateTime.t()` structs.
+  """
+  @callback inspect(date | datetime | naive_datetime, Inspect.Opts.t()) :: String.t()
```
josevalim: We should also pass `time` in here. :)

lib/elixir/lib/calendar.ex
kipcole9: Done.

lib/elixir/lib/calendar.ex
kipcole9: Done. Also added the implementation for `inspect/2` for time into `Calendar.ISO`. Unlike `Date`, `NaiveDateTime` and `DateTime` its not guaranteed that there is a `:calendar` field. So for time it matches on `%{calendar: calendar}` and defaults ` Inspect.Any.inspect/2` otherwise.

josevalim: Time should always have a calendar field. :) It is not guaranteed for something like CDLR, which may work with different Elixir versions, but in Elixir itself it will always be there. :)

fertapric: I think this is still unresolved:

```suggestion
  @callback inspect(time | date | datetime | naive_datetime, Inspect.Opts.t()) :: String.t()
```

lib/elixir/lib/calendar.ex
```diff
@@ -265,6 +265,14 @@ defmodule Calendar do
   """
   @callback valid_time?(hour, minute, second, microsecond) :: boolean
 
+  @doc """
+  Implements inspect for dates, datetimes and naive_datetimes.
```
fertapric:
```suggestion
  Implements inspect for times, dates, datetimes, and "naive" datetimes.
```

lib/elixir/lib/calendar.ex
```diff
@@ -265,6 +265,14 @@ defmodule Calendar do
...
+
+  The implementation must define inspect for `Date.t()`, `DateTime.t()` and
```
fertapric: According to the `@spec`, the function `inspect/2` allows any map/struct that contains the proper `time`, `date`, `naive_datetime`, or `datetime` fields. Maybe we should reword this sentence so people don't expect only `Time`, `Date`, `DateTime`, or `NaiveDateTime` structs :)

lib/elixir/lib/calendar/iso.ex
```diff
@@ -715,6 +715,89 @@ defmodule Calendar.ISO do
...
+  `Date`, `Time`, `DateTime` and `NaiveDateTime`.
+
+  """
+  @doc since: "1.10.0"
+  @impl true
+  @spec inspect(
+          Calendar.date() | Calendar.time() | Calendar.datetime() | Calendar.naive_datetime(),
+          Inspect.Opts.t()
+        ) :: String.t()
+  def inspect(%Date{} = date, _) do
```
fertapric: Should the calendar be checked in all these clauses?

```suggestion
  def inspect(%Date{calendar: __MODULE__} = date, _) do
```

fertapric: Nevermind. Inspect accepts any struct/map with the proper fields.

fertapric: 🤔 Or not...`:calendar` is part of those fields.

Btw, according to the spec, this should work:

```elixir
my_date = %{year: 2019, month: 06, day: 11, calendar: Calendar.ISO}
Calendar.ISO.inspect(my_date)
```

lib/elixir/lib/calendar.ex
kipcole9: Since this function is intended to be called by the `Inspect` protocol, I made the assumption that matching on the structs would be appropriate.  I appreciate as a callback it become public API and therefore optimistic matching may be a better strategy if thats what you recommend.

lib/elixir/lib/calendar/iso.ex
kipcole9: This is an implementation of the function `inspect/2` supporting the `Inspect` protocol.  Therefore since its being called from the protocol as `calendar.inspect` I made the assumption that matching on the calendar wasn't necessary - since it would only be dispatched here if the calendar was `Calendar.ISO`.

fertapric: @kipcole9 Yes, I understand this function was designed to be used by the `Inspect` protocol. However, it's now part of the public interface of the `Calendar.ISO` module, and I think we need to consider that users might call this function directly. 

In summary, we need to answer these questions:

1. The value given as first argument might have a different calendar, should `inspect/2` work with that value? _I think it should not work_
1. The value given as first argument might be a map or a custom struct. According to the typespec `Calendar.date() | Calendar.time() | Calendar.datetime() | Calendar.naive_datetime()`, any map or struct with the proper fields should work: 

    https://github.com/elixir-lang/elixir/blob/0c686aa124bb0a31ba4010a78836c84e4b55066e/lib/elixir/lib/calendar.ex#L72-L112

    What should be changed, the typespec or the implementation? _I think we should change the implementation to accept any map/struct with the proper fields_

kipcole9: Got it, I'll update and push another commit to this PR.

michalmuskala:
Looking at this PR I started wondering - for `to_string` the calendar has 4 callbacks - `time_to_sting`, `date_to_string`, `datetime_to_string` and `naive_datetime_to_sting`. Why not mirror this interface here? It seems weird to have separate callbacks for one protocol and just one for the other. Yes, it's more functions, on the other hand that kind of dispatching would need to be performed internally by each calendar module anyway - we can make it explicit from the get-go.

lib/elixir/lib/calendar/iso.ex
kipcole9: I pushed a commit that uses optimistic matching in line with the other calendar functions.

lib/elixir/lib/calendar.ex
josevalim: I agree. We should match on structs but then it means we need to update the specs to say so instead of using `Calendar.date`.

lib/elixir/lib/calendar/iso.ex
josevalim: @kipcole9 please revert the commit. :) Those are structs on protocol dispatch. By definition they will have all of those fields. As it was before was totally fine.

What @fertapric was saying is that we need to do:

    def inspect(%Date{calendar: __MODULE__, ...})

inside Calendar.ISO.

josevalim:
@michalmuskala there is one important difference between those functions is that the ones being added here are nominal, they work with structs and none of the other function in this module do the same. So for example, we merging `time_to_string` and `date_to_string` into one with different arities would be confusing.

I also have a concern about adding cyclic dependencies, which this patch is doing, but it can be solved in other ways.

fertapric:
@josevalim I think @michalmuskala was suggesting to split `inspect/2` into four different functions. For example: `inspect_time/2`, `inspect_date/2`, `inspect_naive_datetime/2`, and `inspect_datetime/2`.

I like the idea.

josevalim:
Yes, I understood that. I am saying we had good reasons to split it before.
We don’t have good reasons now. :)
-- 


lib/elixir/lib/calendar/iso.ex
kipcole9: Reverted the previous commit, added a match on `calendar: __MODULE__` on each of the clauses.

What would you suggest as the correct types for the `@spec`?

josevalim:
For the @spec and callback definition we should use Time.t, Date.t, and
friends.
-- 


lib/elixir/lib/calendar.ex
```diff
@@ -265,6 +265,15 @@ defmodule Calendar do
   """
   @callback valid_time?(hour, minute, second, microsecond) :: boolean
 
+  @doc """
+  Implements inspect for times, dates, datetimes, and "naive" datetimes.
+
+  The implementation must define inspect for `Date.t()`, `DateTime.t()` and
+  `NaiveDateTime.t()` structs.
+  """
+  @callback inspect(Time.t() | Date.t() | DateTime.t() | MaiveDateTime.t(), Inspect.Opts.t()) ::
```
fertapric:
```suggestion
  @callback inspect(Time.t() | Date.t() | DateTime.t() | NaiveDateTime.t(), Inspect.Opts.t()) ::
```

kipcole9: Arrrggggh. Thanks for the catch.  Getting very late this end and my eyes are beat.

michalmuskala:
@josevalim I wasn't entirely clear. I'm proposing doing exactly the same thing that the `_to_string` functions do - accept each field of the struct as separate argument, so that the calendar module doesn't deal with any of that - right now the calendar module doesn't deal with structs or maps in any of the callbacks - I don't think we should change that.

josevalim:
@michalmuskala oh, I see. I like that!

@kipcole9 can you please go with @michalmuskala's suggestion? The idea is to do:

    inspect_time(hour, minute, second, microsecond)
    inspect_date(year, month, day)
    ...

Similar to the `_to_string` functions. Thank you and apologies for the back and forth.

kipcole9:
Implemented the `inspect_*` callbacks with [this commit](https://github.com/elixir-lang/elixir/pull/9128/commits/03049874ffda8e5f24d67997c68392ae342e78d0)

Its a much cleaner interface.

lib/elixir/test/elixir/calendar/holocene.exs
```diff
@@ -78,6 +78,59 @@ defmodule Calendar.Holocene do
     )
   end
 
+  @impl true
+  @spec inspect_date(Calendar.year(), Calendar.month(), Calendar.day(), Inspect.Opts.t()) ::
```
fertapric: I think this `@spec` is not required here

lib/elixir/lib/calendar/iso.ex
```diff
@@ -715,6 +715,126 @@ defmodule Calendar.ISO do
...
+  @impl true
+  @spec inspect_time(
+          Calendar.hour(),
+          Calendar.minute(),
+          Calendar.second(),
+          Calendar.microsecond(),
+          Inspect.Opts.t()
+        ) ::
+          String.t()
+  def inspect_time(
```
fertapric: Minor nitpick: this fits into a single line like in date and naive_datetime (same in the test with the Holocene calendar)

lib/elixir/lib/calendar/iso.ex
```diff
@@ -715,6 +715,126 @@ defmodule Calendar.ISO do
...
+  def inspect_naive_datetime(year, month, day, hour, minute, second, microsecond, _opts) do
+    formatted = naive_datetime_to_string(year, month, day, hour, minute, second, microsecond)
+
+    "~N[" <> formatted <> "]"
+  end
+
+  @doc """
+  Implements the inspect protocol for a time
+  delegated from `Time`.
+
```
fertapric: Minor nitpick: this blank line at the end of the `@doc` is not common in the Elixir codebase (unless you have indentation) 🙂

lib/elixir/lib/calendar/iso.ex
```diff
@@ -715,6 +715,126 @@ defmodule Calendar.ISO do
     {0, 1}
   end
 
+  @doc """
+  Implements the inspect protocol for a date
+  delegated from `Date`.
```
fertapric: What do you mean by "delegated from `Date`"? I find that confusing 🙂

lib/elixir/lib/calendar.ex
```diff
@@ -265,6 +265,53 @@ defmodule Calendar do
   """
   @callback valid_time?(hour, minute, second, microsecond) :: boolean
 
+  @doc """
+  Implements inspect for a date.
+
+  """
+  @callback inspect_date(year, month, day, Inspect.Opts.t()) :: String.t()
```
josevalim:
```suggestion
  @callback inspect_date(year, month, day, Inspect.Opts.t()) :: Inspect.Algebra.t()
```

lib/elixir/lib/calendar.ex
```diff
@@ -265,6 +265,53 @@ defmodule Calendar do
...
+
+  @doc """
+  Implements inspect for a time.
+
+  """
+  @callback inspect_time(hour, minute, second, microsecond, Inspect.Opts.t()) :: String.t()
```
josevalim:
```suggestion
  @callback inspect_time(hour, minute, second, microsecond, Inspect.Opts.t()) :: Inspect.Algebra.t()
```

lib/elixir/lib/calendar.ex
```diff
@@ -265,6 +265,53 @@ defmodule Calendar do
...
+              year,
+              month,
+              day,
+              hour,
+              minute,
+              second,
+              microsecond,
+              Inspect.Opts.t()
+            ) ::
+              String.t()
```
josevalim:
```suggestion
              Inspect.Algebra.t()
```

lib/elixir/lib/calendar.ex
```diff
@@ -265,6 +265,53 @@ defmodule Calendar do
...
+              hour,
+              minute,
+              second,
+              microsecond,
+              time_zone,
+              zone_abbr,
+              utc_offset,
+              std_offset,
+              Inspect.Opts.t()
+            ) :: String.t()
```
josevalim:
```suggestion
            ) :: Inspect.Algebra.t()
```

lib/elixir/lib/calendar.ex
josevalim:
```suggestion
  @callback inspect_date(year, month, day, Inspect.Opts.t()) :: Inspect.Algebra.t()
```

lib/elixir/lib/calendar.ex
```diff
@@ -265,6 +265,53 @@ defmodule Calendar do
   """
   @callback valid_time?(hour, minute, second, microsecond) :: boolean
 
+  @doc """
+  Implements inspect for a date.
+
```
josevalim: Please remove empty new line here and below. :)

lib/elixir/lib/calendar/iso.ex
josevalim: Good point. I think those docs could say "Inspect implementation for `Date` in this calendar." and so on. WDYT?

lib/elixir/lib/calendar/iso.ex
```diff
@@ -715,6 +715,126 @@ defmodule Calendar.ISO do
...
+
+  """
+  @doc since: "1.10.0"
+  @impl true
+  @spec inspect_date(Calendar.year(), Calendar.month(), Calendar.day(), Inspect.Opts.t()) ::
+          String.t()
```
josevalim:
```suggestion
          Inspect.Algebra.t()
```

lib/elixir/lib/calendar/iso.ex
```diff
@@ -715,6 +715,126 @@ defmodule Calendar.ISO do
...
+          Calendar.minute(),
+          Calendar.second(),
+          Calendar.microsecond(),
+          Calendar.time_zone(),
+          Calendar.zone_abbr(),
+          Calendar.utc_offset(),
+          Calendar.std_offset(),
+          Inspect.Opts.t()
+        ) ::
+          String.t()
```
josevalim:
```suggestion
          Inspect.Algebra.t()
```

lib/elixir/lib/calendar/iso.ex
```diff
@@ -715,6 +715,126 @@ defmodule Calendar.ISO do
...
+          Calendar.year(),
+          Calendar.month(),
+          Calendar.day(),
+          Calendar.hour(),
+          Calendar.minute(),
+          Calendar.second(),
+          Calendar.microsecond(),
+          Inspect.Opts.t()
+        ) ::
+          String.t()
```
josevalim:
```suggestion
          Inspect.Algebra.t()
```

lib/elixir/lib/calendar/iso.ex
```diff
@@ -715,6 +715,126 @@ defmodule Calendar.ISO do
...
+  @doc since: "1.10.0"
+  @impl true
+  @spec inspect_time(
+          Calendar.hour(),
+          Calendar.minute(),
+          Calendar.second(),
+          Calendar.microsecond(),
+          Inspect.Opts.t()
+        ) ::
+          String.t()
```
josevalim:
```suggestion
          Inspect.Algebra.t()
```

lib/elixir/test/elixir/calendar/holocene.exs
```diff
@@ -78,6 +78,59 @@ defmodule Calendar.Holocene do
...
+    )
+  end
+
+  @impl true
+  def inspect_naive_datetime(year, month, day, hour, minute, second, microsecond, _opts) do
+    naive_datetime_to_string(year, month, day, hour, minute, second, microsecond)
+  end
+
+  @impl true
+  def inspect_time(
```
fertapric: This is not being used, a test for `Time` is missed

lib/elixir/lib/calendar/iso.ex
fertapric: Sounds good. I would link to the `Inspect` protocol using backticks though:

> `Inspect` implementation for `Date` in this calendar.

Or:

> Implements the `Inspect` protocol for `Date` in this calendar.

lib/elixir/test/elixir/calendar/holocene.exs
kipcole9: Added a test for `inspect_time/2` and fixed the docs in line with comments above.

kipcole9:
Added [this commit](https://github.com/elixir-lang/elixir/pull/9128/commits/7299eff93efe0a03515c00e72cef261a2b31d306) which I think addresses the outstanding issues:

* Fix doc formatting in `Calendar.ISO`
* Adds test for `inspect_time/5` in `time_test.exs`

fertapric:
Thanks @kipcole9! 💜

PR #9130, opened 2019-06-11
Fix redirection to null on windows

cw789:
Concerning generated `bin/xxx.bat` from `mix release`.
Otherwise I got a file called `nil` within my directory, after starting a release.
In my case with the contents of `1 file(s) copied.`

I think there aren't any tests regarding this change, but I'm not 100% sure.
Please review, as this is my very first contribution.


josevalim:

Issue #9129, opened 2019-06-11
IEx.Helpers.t displays types in reverse order of definition


wojtekmach:
### Environment

* Elixir & Erlang/OTP versions (elixir --version): master

### Current behavior

For this module:

```elixir
defmodule M do
  @type a() :: integer()

  @type b() :: integer()

  @type c() :: integer()
end
```

we get:

```elixir
iex(1)> t M
@type c() :: integer()

@type b() :: integer()

@type a() :: integer()
```

Note:

```elixir
iex(1)> Code.Typespec.fetch_types(M)
{:ok,
 [
   type: {:c, {:type, 6, :integer, []}, []},
   type: {:b, {:type, 4, :integer, []}, []},
   type: {:a, {:type, 2, :integer, []}, []}
 ]}
iex(2)> Code.fetch_docs(M)
{:docs_v1, 1, :elixir, "text/markdown", :none, %{},
 [
   {{:type, :a, 0}, 2, [], :none, %{}},
   {{:type, :b, 0}, 4, [], :none, %{}},
   {{:type, :c, 0}, 6, [], :none, %{}}
 ]}
```

Maybe we're putting them into chunks in the wrong order in the first place?

### Expected behavior

Types are sorted by name. Note, we could display them in order of definition but since we're sorting callbacks in `IEx.Helpers.b`, makes sense to sort it here too.

josevalim:
Yes, we should sort them alphabetically.
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

PR #9131, opened 2019-06-11
IEx: Sort Types in `t` helper

maennchen:
Fix for #9129

### DIFF
--- a/lib/iex/lib/iex/introspection.ex
+++ b/lib/iex/lib/iex/introspection.ex
@@ -644,7 +644,9 @@ defmodule IEx.Introspection do
         types_not_found(inspect(module))
 
       {:ok, types} ->
-        Enum.each(types, &(&1 |> format_type() |> IO.puts()))
+        types
+        |> Enum.sort_by(fn {_, {name, _, args}} -> {name, length(args)} end)
+        |> Enum.each(&(&1 |> format_type() |> IO.puts()))
     end
 
     dont_display_result()
@@ -656,15 +658,19 @@ defmodule IEx.Introspection do
         no_beam(module)
 
       {:ok, types} ->
-        printed =
-          for {kind, {^type, _, args}} = typespec <- types,
-              kind != :typep do
-            type_doc(module, type, length(args), typespec)
-            |> print_typespec()
-          end
+        types
+        |> Enum.filter(&match?({kind, {^type, _, _}} when kind in [:type, :opaque], &1))
+        |> Enum.sort_by(fn {_, {name, _, args}} -> {name, length(args)} end)
+        |> case do
+          [] ->
+            types_not_found_or_private("#{inspect(module)}.#{type}")
 
-        if printed == [] do
-          types_not_found_or_private("#{inspect(module)}.#{type}")
+          types ->
+            Enum.map(types, fn {_, {_, _, args}} = typespec ->
+              module
+              |> type_doc(type, length(args), typespec)
+              |> print_typespec()
+            end)
         end
     end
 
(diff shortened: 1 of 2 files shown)


lib/iex/lib/iex/introspection.ex
```diff
@@ -644,7 +644,16 @@ defmodule IEx.Introspection do
         types_not_found(inspect(module))
 
       {:ok, types} ->
-        Enum.each(types, &(&1 |> format_type() |> IO.puts()))
+        types
+        |> Enum.filter(&match?({kind, _} when kind in [:type, :opaque], &1))
```
maennchen: @josevalim I also filtered out `typep` here, which was not done before.

wojtekmach: According to #2612 we want to list typeps so Id be careful about making this particular change.

fertapric: I think we should add a test to prevent `typep` to be filtered out in future PRs

maennchen: @wojtekmach That's why I'm specifically pointing it out. Should I leave it in or should I remove the line?

Before this PR we've shown `typep` for `t(module)`, but not for `t({module, type})` / `t({module, type, arity})`. This didn't make sense to me, therefore I changed the behaviour of the first to reflect the others.

maennchen: @fertapric If the behavior from before is the desired one, I'll remove the line and can also add a test to make sure it is not filtered out in the future.

josevalim:
Good call @wojtekmach.
-- 


josevalim:
Let’s remove the filtering from this PR for now and handle it in another PR.
-- 


lib/iex/lib/iex/introspection.ex
maennchen: @fertapric & @wojtekmach Adressed in 841f5ab

wojtekmach: @maennchen @fertapric thanks for adding the test <3

lib/iex/lib/iex/introspection.ex
```diff
@@ -644,7 +644,15 @@ defmodule IEx.Introspection do
         types_not_found(inspect(module))
 
       {:ok, types} ->
-        Enum.each(types, &(&1 |> format_type() |> IO.puts()))
+        types
+        |> Enum.sort_by(fn {_, {name, _, args}} -> {name, length(args)} end)
+        |> case do
+          [] ->
```
fertapric: I think this clause can never be executed :)

maennchen: Without the `typep` filter, you're correct. I'll remove the case.

maennchen: Fixed in 9830d3d8089f6594a821a5fe39ccf2be0247d725

lib/iex/lib/iex/introspection.ex
```diff
@@ -677,16 +683,24 @@ defmodule IEx.Introspection do
...
-        if printed == [] do
-          types_not_found_or_private("#{inspect(module)}.#{type}")
+        types
+        |> Enum.filter(
+          &match?(
+            {kind, {^type, _, args}} when kind in [:type, :opaque] and length(args) == arity,
+            &1
+          )
+        )
+        |> Enum.sort_by(fn {_, {name, _, args}} -> {name, length(args)} end)
```
fertapric: I think you cannot have more than one result here for `{module, type, arity}`, @josevalim?

josevalim:
Yeah, you can’t have more than one result for a given module, function,
arity.
-- 


lib/iex/lib/iex/introspection.ex
maennchen: @fertapric Adressed in 4250b33c872169fc5b4d57264b22c83b1701a3ab

fertapric: The `Enum.sort_by` can also be removed

maennchen: @fertapric Adressed in 1a6c82f5685b771a9766aafca9215bb39d37c6df

josevalim:

PR #9133, opened 2019-06-12
Deduplicate relative paths in protocol consolidation

howleysv:
The issue I was having in my environment (which you can replicate using nix with `nix-shell -p elixir`)  was that `:code.get_path()` returns duplicate paths for the elixir lib in the form:
```
.../elixir/bin/../lib/elixir/ebin
.../elixir/lib/elixir/ebin
```

This caused duplicate protocol implementations to be found and a bunch of warnings were getting spit out at every compile, but I think it would be better for elixir to handle these cases gracefully.

The main addition here calls `Path.expand/1` on every path and then `Enum.uniq/1` to remove duplicates. I also noticed the args were flipped in `:lists.prefix/2` for removing the OTP paths. Passing String paths to `Protocol.extract_impls/2` then caused an issue (even though the typespec says it accepts Strings) because `:beam_lib.chunks/2` only accepts charlist paths. Incidently, I noticed that `:file.list_dir/1` DOES accept either type, so I removed the unnecessary guard.

### DIFF
--- a/lib/elixir/lib/protocol.ex
+++ b/lib/elixir/lib/protocol.ex
@@ -463,14 +463,12 @@ defmodule Protocol do
         do: mod
   end
 
-  defp list_dir(path) when is_list(path) do
+  defp list_dir(path) do
     case :file.list_dir(path) do
       {:ok, files} -> files
       _ -> []
     end
   end
-
-  defp list_dir(path), do: list_dir(to_charlist(path))
 
   defp extract_from_file(path, file, prefix, callback) do
     if :lists.prefix(prefix, file) and :filename.extension(file) == '.beam' do
@@ -478,7 +476,7 @@ defmodule Protocol do
     end
   end
 
-  defp extract_from_beam(file, callback) do
+  defp extract_from_beam(file, callback) when is_list(file) do
     case :beam_lib.chunks(file, [:attributes]) do
       {:ok, {module, [attributes: attributes]}} ->
         callback.(module, attributes)
@@ -486,6 +484,10 @@ defmodule Protocol do
       _ ->
         nil
     end
+  end
+
+  defp extract_from_beam(file, callback) do
+    extract_from_beam(to_charlist(file), callback)
   end
 
   @doc """
--- a/lib/mix/lib/mix/tasks/compile.protocols.ex
+++ b/lib/mix/lib/mix/tasks/compile.protocols.ex
@@ -109,11 +109,19 @@ defmodule Mix.Tasks.Compile.Protocols do
   end
 
   defp consolidation_paths do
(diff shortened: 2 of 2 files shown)


josevalim:
Thanks Shane!

Does this issue happen exclusively with nix? If so, is there a possibility
nix is the one duplicating the paths?

In other words, I would prefer to figure out the root cause as having
duplicate paths may cause other unforeseen issues later.
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

howleysv:
This is just where I encountered it, I didn't dig further into what might be causing it in nix but it should probably also be addressed.

I decided to take this approach since it seemed like regardless of this specific instance, if we end up with multiple paths that reference the same dir, we shouldn't treat the same beam file multiple times as separate implementations of a protocol.

josevalim:
I’d prefer to investigate the root cause and include this fix only if
really necessary. :)
-- 


howleysv:
I have narrowed this down to being related the the `ERL_LIBS` environmental variable.
If this includes the elixir library path, it results in duplicated paths in `:code.get_path/0`, which in turn results in the aforementioned Protocol issue.

I've never worked with `ERL_LIBS` before, so I'm not sure what the legitimate use-cases are for using it in the elixir ecosystem, or if duplicated paths are supposed to be tolerated or prevented.

josevalim:
@howleysv perfect. The last question is: what/who is setting/changing ERL_LIBS to point to Elixir? Do you have any idea?

howleysv:
My best guess is that it's caused by this script used by the elixir packages: https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/interpreters/elixir/setup-hook.sh

But I don't know enough about nix to debug it or to rebuild the package without this script 😞

josevalim:
Maybe you can revert the commit that adds it? https://github.com/NixOS/nixpkgs/commit/2d6d9682bb8a910f3eb5f0be8d5687c753e9732c#diff-6e905b109f4204dddbf31192e619abcc

But yeah, I don't think this is a bug in Elixir a it is being configured incorrectly. If other people are running into this, then maybe it is worth doing some detection and failing early on, but it seems you ran into an edge case.

howleysv:
Thanks for the feedback @josevalim!

I was trying to figure out if erlang considers it bad behaviour to have duplicate code paths and while it handles it without complaining, the output of `:code.clash/0` seems to suggest that it's something that isn't intended. So I'm happy to consider it bad configuration to use `ERL_LIBS` like this and I'll try get it fixed on the nix side.

If you think there's value in the bugfixes I made to `Protocol.extract_impls/2` and `Mix.Tasks.Compile.Protocols.filter_otp/2`, I can resubmit them.

PR #9134, opened 2019-06-12
Improve a comment in env.sh.eex for releases

whatyouhide:
When reading the file, it's not clear why uncommenting `RELEASE_NODE` fails and it fails with a message that is not easy to act on (`Can't set short node name!\nPlease check your configuration`).

### DIFF
--- a/lib/mix/lib/mix/tasks/release.init.ex
+++ b/lib/mix/lib/mix/tasks/release.init.ex
@@ -66,8 +66,8 @@ defmodule Mix.Tasks.Release.Init do
     #   export ELIXIR_ERL_OPTIONS="-heart"
     # fi
 
-    # Set the release to work across nodes. If using the long name format like the
-    # one below (my_app@127.0.0.1), you need to also uncomment the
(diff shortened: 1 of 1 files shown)


josevalim:
Can you please mirror this in the env.bat.eex template too?
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

lib/mix/lib/mix/tasks/release.init.ex
```diff
@@ -66,7 +66,9 @@ defmodule Mix.Tasks.Release.Init do
     #   export ELIXIR_ERL_OPTIONS="-heart"
     # fi
 
-    # Set the release to work across nodes
+    # Set the release to work across nodes. If using a long name like the
```
fertapric: "my_app@127.0.0.1" is pretty short 😛 

What do you think about?

```suggestion
    # Set the release to work across nodes. If using a long name format like the
```

whatyouhide:
> Can you please mirror this in the env.bat.eex template too?

Done.

PR #9135, opened 2019-06-12
Fix bad naming on release script for Windows

fertapric:
@josevalim this needs to be back-ported to v1.9 too


josevalim:
Approved!
-- 

PR #9136, opened 2019-06-12
Update keyword.ex docs

hassanshaikley:
Use `short hand` instead of `short-hand` to be consistent

### DIFF
--- a/lib/elixir/lib/keyword.ex
+++ b/lib/elixir/lib/keyword.ex
@@ -34,7 +34,7 @@ defmodule Keyword do
   of a keyword without quotes, Elixir will warn.
 
   Note that when keyword lists are passed as the last argument to a function,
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/keyword.ex
```diff
@@ -34,7 +34,7 @@ defmodule Keyword do
   of a keyword without quotes, Elixir will warn.
 
   Note that when keyword lists are passed as the last argument to a function,
-  if the short-hand syntax is used then the square brackets around the keyword list
+  if the short hand syntax is used then the square brackets around the keyword list
```
whatyouhide: I think that the [correct spelling is "shorthand"](https://www.merriam-webster.com/dictionary/shorthand)?

```suggestion
  if the shorthand syntax is used then the square brackets around the keyword list
```

josevalim:

PR #9139, opened 2019-06-14
Place Version.Requirement module under Basic Types in docs.exs

eksperimental:
As of now, it was floating around at the same level as Kernel and Kernel.SpecialForms


josevalim:

PR #9141, opened 2019-06-14
Fix typos

amatalai:


lib/elixir/lib/macro.ex
```diff
@@ -2,10 +2,10 @@ import Kernel, except: [to_string: 1]
 
 defmodule Macro do
   @moduledoc ~S"""
-  Macros are a compile-time construct that are invoked with Elixir's AST
+  Macros are compile-time constructs that are invoked with Elixir's AST
   as input and a superset of Elixir's AST as output.
 
-  Let's see a simple that shows the difference between functions and macros:
+  Let's see a sample that shows the difference between functions and macros:
```
eksperimental: I think this is what was meant

```suggestion
  Let's see a simple example that shows the difference between functions and macros:
```

amatalai: 👍

josevalim:

PR #9143, opened 2019-06-15
Use heredocs for top-level executables' usage docs

fxn:
This PR proposes using heredocs for the usage docs of the top-level executables.

Benefits I see:

* No need to escape double quotes within the literal.
* The rightmost column with descriptions is now vertically aligned in the source code, matching easily what the user will get printed.
* The end of the literal is obvious.
* The redirection to standard error is visible upfront.

AFAIK this is portable across different Unix shells, including `sh`.

### DIFF
--- a/bin/elixir
+++ b/bin/elixir
@@ -2,22 +2,23 @@
 set -e
 
 if [ $# -eq 0 ] || [ "$1" = "--help" ] || [ "$1" = "-h" ]; then
-  echo "Usage: $(basename "$0") [options] [.exs file] [data]
+  cat <<USAGE >&2
+Usage: $(basename "$0") [options] [.exs file] [data]
 
 ## General options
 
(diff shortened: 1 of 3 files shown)


whatyouhide:
I've read that `cat` with an heredoc is a bit slower than `echo`, is that true? Not sure we care about this but I thought I would mention it anyways.

fxn:
I guess the answer for arbitrary inputs is complicated ([example](https://unix.stackexchange.com/questions/230206/why-there-is-such-a-difference-in-execution-time-of-echo-and-cat)), but in this particular case the strings are so small that the difference, if any, is negligible.

I trigger usage of `bin/elixir` in `master` and in this branch, and cannot tell the difference.

fxn:
`/usr/bin/time` reports 0.00 (`master`) vs 0.01 (`heredocs`) consistently, so there is _some_ difference.

PR #9145, opened 2019-06-15
Tweaks to the docs of protocols

fxn:
The main point of this patch is revising the documentation of impl_for/1, because it only mentions structures.

There are small tweaks too since I was on it: some double spaces, use italics for a defined term, be a bit more precise about debug info removal after protocol consolidation, etc.

### DIFF
--- a/lib/elixir/lib/protocol.ex
+++ b/lib/elixir/lib/protocol.ex
@@ -163,15 +163,15 @@ defmodule Protocol do
 
       * `:consolidated?` - returns whether the protocol is consolidated
 
-      * `:functions` - returns keyword list of protocol functions and their arities
+      * `:functions` - returns a keyword list of protocol functions and their arities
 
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/protocol.ex
```diff
@@ -195,7 +195,7 @@ defmodule Protocol do
 
   In order to speed up dispatching in production environments, where
   all implementations are known up-front, Elixir provides a feature
-  called protocol consolidation. Consolidation directly links protocols
+  called _protocol consolidation_. Consolidation directly links protocols
```
josevalim: We don’t use this notation in Elixir docs. Could you please use asterisks for italics instead? Thanks!

fxn: Pushed!

fertapric:
Gracias @fxn! 💜

PR #9147, opened 2019-06-16
Add backwards compatibility for rel/vm.args

mobileoverlord:
A lot of existing Nerves apps have a `rel/vm.args` file. As part of our upgrade instructions, we will be advocating this file be renamed. Overlooking this step can cause ill effects when running new firmware on the target. I think it might be safer if we look for `vm.args.eex` then `vm.args`.

### DIFF
--- a/lib/mix/lib/mix/tasks/release.ex
+++ b/lib/mix/lib/mix/tasks/release.ex
@@ -1076,10 +1076,15 @@ defmodule Mix.Tasks.Release do
   end
 
   defp make_vm_args(release, path) do
-    if File.exists?("rel/vm.args.eex") do
-      copy_template("rel/vm.args.eex", path, [release: release], force: true)
-    else
-      File.write!(path, vm_args_template(release: release))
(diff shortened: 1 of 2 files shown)


josevalim:
What about the other .eex files. Is the .eex extension optional in them
too? The other thing is: are we sure those files are always equivalent?
IIRC, Distillery builtin vm.args contain either —cookie or —name which
won’t work with Elixir releases, so we may break those cases.

Maybe it is best if nerves warns in those cases with a nerves specific
update instruction that will put the proper files in place?
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

mobileoverlord:
> The other thing is: are we sure those files are always equivalent?
IIRC, Distillery builtin vm.args contain either —cookie or —name which
won’t work with Elixir releases, so we may break those cases.

I didn't consider this case, we distribute our own generated file. We will handle this during our release phase and raise.

PR #9155, opened 2019-06-24
Fix typespec of Macro.Env.t

indocomsoft:
This PR addresses the cause of dialyzer warnings because of the `macro_aliases` field of `Macro.Env` struct.


josevalim:

PR #9157, opened 2019-06-24
More details about the :native time unit, from the Erlang docs

nathanl:
See http://erlang.org/doc/man/erlang.html#type-time_unit


whatyouhide:
Thanks @nathanl! 💟

PR #9165, opened 2019-06-25
Unify code path in String.split

hrubi:
When calling `String.split` with options `parts` and/or `trim` explicitly
set to default values, different code path was taken in comparison to when
no options were given. The alternate code path has some performance
penalties as shown below, so it's better that the
"simple" case is handled always the same way.
```
iex(1)> lines = for _ <- 1..1_000_000, do: "a,b"

iex(2)> :timer.tc(fn -> Enum.each(lines, &String.split(&1, ",")) end)
{2254104, :ok}

iex(3)> :timer.tc(fn -> Enum.each(lines, &String.split(&1, ",", trim: false)) end)
{61426180, :ok}
```

### DIFF
--- a/lib/elixir/lib/string.ex
+++ b/lib/elixir/lib/string.ex
@@ -420,7 +420,7 @@ defmodule String do
     end
   end
 
-  def split(string, pattern, []) when is_tuple(pattern) or is_binary(string) do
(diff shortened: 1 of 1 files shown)


josevalim:

PR #9166, opened 2019-06-26
Add no_return to Registry.put_meta/3 spec

jdenen:
Dialyzer was blowing up one of my project's functions. Tracked it down to this spec.

### DIFF
--- a/lib/elixir/lib/registry.ex
+++ b/lib/elixir/lib/registry.ex
@@ -1019,7 +1019,7 @@ defmodule Registry do
 
   """
   @doc since: "1.4.0"
-  @spec put_meta(registry, meta_key, meta_value) :: :ok
+  @spec put_meta(registry, meta_key, meta_value) :: :ok | no_return
   def put_meta(registry, key, value) when is_atom(registry) and (is_atom(key) or is_tuple(key)) do
     try do
       :ets.insert(registry, {key, value})


josevalim:
Closing for the reasons @fishcakez said above. :) Thanks!

jdenen:
So that makes sense from the spec perspective, but then, shouldn't this function either (A) return `{:error, _}` or (B) be renamed to `Registry.put_meta!/3`?

We had a hard time tracking down our dialyzer issue because we had no indication that `put_meta/3` doesn't return in some cases. The specs don't tell us, and convention indicates that the function shouldn't raise.

Are we expected to go through every function's source in these cases? That would be... cumbersome.

Just hoping to save the next person some time ❤️

josevalim:
All functions can error, for example, if given bad arguments. Perhaps you
are passing an argument that dialyzer thinks will never succeed? What is
the error message that you are seeing?
-- 


jdenen:
This was our original, failing spec:

```elixir
@impl GenServer
def handle_continue(:start_coordinator, state) do
  {:ok, group_coordinator_pid} =
    :brod_group_coordinator.start_link(state.name, state.group, state.topics, state.config, __MODULE__, self())

  Enum.each(state.topics, fn topic ->
    :ok = :brod.start_consumer(state.name, topic, state.config)
  end)
  
  # This call is the offender. Commenting it out appeases dialyzer.
  Registry.put_meta(:"elsa_registry_#{state.name}"), :group_coordinator, group_coordinator_pid)

  {:noreply, %{state | group_coordinator_pid: group_coordinator_pid}}
end
```

It resulted in this error:
```
lib/elsa/group/manager.ex:95:no_return
Function handle_continue/2 has no local return.
```

And it was an easy enough fix:
```elixir
@impl GenServer
def handle_continue(:start_coordinator, state) do
  {:ok, group_coordinator_pid} =
    :brod_group_coordinator.start_link(state.name, state.group, state.topics, state.config, __MODULE__, self())

  Enum.each(state.topics, fn topic ->
    :ok = :brod.start_consumer(state.name, topic, state.config)
  end)
  
  Registry.put_meta(:"elsa_registry_#{state.name}"), :group_coordinator, group_coordinator_pid)

  {:noreply, %{state | group_coordinator_pid: group_coordinator_pid}}
rescue
  e -> {:stop, e, state}
end
```

Not a big deal that we had to make this fix. It's the right fix. But there's no indication that `Registry.put_meta/3` has a `raise` in it.

My initial thought (and the genesis of this PR) was to fix the spec, but I can see what you mean about `no_return`. My question is now:

> As a user of the `Registry` API, how do I know that `Registry.put_meta/3` has a `raise`?

I now know it's not the spec. Should it be the convention of function names ending with `!`?

josevalim:
> I now know it's not the spec. Should it be the convention of function names ending with !?

It depends. As I said above, most functions in Elixir can raise (for example, if given a bad argument), so if that was the rule, then almost all function in Elixir would have a `!`. So the `!` is reserved for semantic errors.

For some reason, dialyzer thinks the call to put_meta will always raise but I don't know why. I would personally avoid changing the code just to make dialyzer happy and try to have a minimal case for further discussion. In particular, note that the code you provided has a syntax error. So maybe there is something else happening that was omitted?

jdenen:
> In particular, note that the code you provided has a syntax error. So maybe there is something else happening that was omitted?

Thanks for pointing that out. It was just a copy/paste error. I cut out some private functions to reduce the lines in my example. But I fixed it above.

> try to have a minimal case for further discussion

Will follow up shortly with this.

josevalim:
@jdenen but could the private functions be the ones causing dialyzer to think there is an error?

jdenen:
It's the ambiguity of our `state` map being passed around the `GenServer` callbacks. Commenting out the `:brod_group_coordinator` line also fixed it in dialyzer's eyes, as did changing another spec from expecting `String.t` to expecting `term`.

Thanks for working this out with me! I learned quite a few things along the way.

PR #9179, opened 2019-06-30
Raise for @describetag defined before use ExUnit.Case

blatyo:
I assume it's an oversight that `@describetag` wasn't being checked here?


whatyouhide:
Thank you @blatyo! 💟

PR #9180, opened 2019-06-30
Add test for @describetag outside of describe block

whatyouhide:
Inspired by #9179.

### DIFF
--- a/lib/ex_unit/lib/ex_unit/case.ex
+++ b/lib/ex_unit/lib/ex_unit/case.ex
@@ -382,10 +382,6 @@ defmodule ExUnit.Case do
   """
   defmacro describe(message, do: block) do
     quote do
-      if Module.get_attribute(__MODULE__, :describetag) != [] do
-        raise "you can only set @describetag inside a \"describe\" block"
-      end
-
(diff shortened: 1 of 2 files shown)


lib/ex_unit/lib/ex_unit/case.ex
```diff
@@ -382,6 +382,10 @@ defmodule ExUnit.Case do
   """
   defmacro describe(message, do: block) do
     quote do
+      if Module.get_attribute(__MODULE__, :describetag) != [] do
+        raise "you can only set @describetag inside a \"describe\" block"
```
blatyo: I don't think this is correct. You can also set a describe tag before a describe block. I think this is a backwards incompatible change.

blatyo: Looks like I'm wrong. But this is already happening here:

https://github.com/elixir-lang/elixir/blob/master/lib/ex_unit/lib/ex_unit/case.ex#L415-L417

whatyouhide:
Ouch, how the hell did I miss that 😄 Updated, will merge once CI is green. Thanks @fertapric and @blatyo! 💟

PR #9181, opened 2019-06-30
Add describe and module attributes

blatyo:
This adds `ExUnit.Case.register_describe_attribute/3` and `ExUnit.Case.register_module_attribute/3`.

This was initially discussed in this thread: https://groups.google.com/forum/#!topic/elixir-lang-core/WcLghqAmBE4

### DIFF
--- a/lib/ex_unit/lib/ex_unit/case.ex
+++ b/lib/ex_unit/lib/ex_unit/case.ex
@@ -551,8 +551,11 @@ defmodule ExUnit.Case do
 
   """
   @spec register_attribute(env, name :: atom, opts :: keyword) :: :ok
-  def register_attribute(env, name, opts \\ []) when is_atom(name) and is_list(opts) do
-    register_attribute(:ex_unit_registered_test_attributes, env, name, opts)
+  def register_attribute(env, name, opts \\ [])
+  def register_attribute(%{module: mod}, name, opts), do: register_attribute(mod, name, opts)
+
+  def register_attribute(mod, name, opts) when is_atom(mod) and is_atom(name) and is_list(opts) do
+    register_attribute(:ex_unit_registered_test_attributes, mod, name, opts)
   end
 
   @doc """
@@ -590,8 +593,15 @@ defmodule ExUnit.Case do
 
   """
   @spec register_describe_attribute(env, name :: atom, opts :: keyword) :: :ok
-  def register_describe_attribute(env, name, opts \\ []) when is_atom(name) and is_list(opts) do
-    register_attribute(:ex_unit_registered_describe_attributes, env, name, opts)
+  def register_describe_attribute(env, name, opts \\ [])
+
+  def register_describe_attribute(%{module: mod}, name, opts) do
+    register_describe_attribute(mod, name, opts)
+  end
+
+  def register_describe_attribute(mod, name, opts)
+      when is_atom(mod) and is_atom(name) and is_list(opts) do
+    register_attribute(:ex_unit_registered_describe_attributes, mod, name, opts)
   end
 
   @doc """
@@ -623,8 +633,15 @@ defmodule ExUnit.Case do
 
   """
   @spec register_module_attribute(env, name :: atom, opts :: keyword) :: :ok
-  def register_module_attribute(env, name, opts \\ []) when is_atom(name) and is_list(opts) do
-    register_attribute(:ex_unit_registered_module_attributes, env, name, opts)
+  def register_module_attribute(env, name, opts \\ [])
+
+  def register_module_attribute(%{module: mod}, name, opts) do
+    register_module_attribute(mod, name, opts)
+  end
+
+  def register_module_attribute(mod, name, opts)
+      when is_atom(mod) and is_atom(name) and is_list(opts) do
+    register_attribute(:ex_unit_registered_module_attributes, mod, name, opts)
   end
 
   defp register_attribute(type, %{module: module}, name, opts) do
--- a/lib/ex_unit/lib/ex_unit/case.ex
+++ b/lib/ex_unit/lib/ex_unit/case.ex
@@ -525,11 +525,10 @@ defmodule ExUnit.Case do
   Registers a new attribute to be used during `ExUnit.Case` tests.
 
   The attribute values will be available through `context.registered`.
-  Registered values are cleared after each `ExUnit.Case.test/3` similar
+  Registered values are cleared after each `test/3` similar
   to `@tag`.
 
-  `Module.register_attribute/3` is used to register the attribute,
-  this function takes the same options.
+  This function takes the same options as `Module.register_attribute/3`.
 
   ## Examples
 
@@ -562,11 +561,10 @@ defmodule ExUnit.Case do
   Registers a new describe attribute to be used during `ExUnit.Case` tests.
 
   The attribute values will be available through `context.registered`.
-  Registered values are cleared after each `ExUnit.Case.describe/2` similar
+  Registered values are cleared after each `describe/2` similar
   to `@describetag`.
 
-  `Module.register_attribute/3` is used to register the attribute,
-  this function takes the same options.
+  This function takes the same options as `Module.register_attribute/3`.
 
   ## Examples
 
@@ -609,8 +607,7 @@ defmodule ExUnit.Case do
 
   The attribute values will be available through `context.registered`.
 
-  `Module.register_attribute/3` is used to register the attribute,
-  this function takes the same options.
+  This function takes the same options as `Module.register_attribute/3`.
 
   ## Examples
 
@@ -642,10 +639,6 @@ defmodule ExUnit.Case do
   def register_module_attribute(mod, name, opts)
       when is_atom(mod) and is_atom(name) and is_list(opts) do
     register_attribute(:ex_unit_registered_module_attributes, mod, name, opts)
-  end
-
-  defp register_attribute(type, %{module: module}, name, opts) do
-    register_attribute(type, module, name, opts)
   end
 
   defp register_attribute(type, mod, name, opts) do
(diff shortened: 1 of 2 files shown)


lib/ex_unit/lib/ex_unit/case.ex
```diff
@@ -416,6 +424,12 @@ defmodule ExUnit.Case do
       raise "@describetag must be set inside describe/2 blocks"
     end
 
+    for attribute <- Module.get_attribute(module, :ex_unit_registered_describe_attributes),
+        value = Module.get_attribute(module, attribute),
+        value not in [[], nil] do
```
blatyo: I don't have a good way to tell if the attribute is accumulated here. The problem that introduces is that technically you could do:

``` elixir
ExUnit.Case.register_describe_attribute(__MODULE__, :describe_fixture)
@describe_fixture []
describe "a thing" do
  test "has fixture without error", context do
    assert context.registered.describe_fixture == []
  end
end
```

josevalim: Yeah. For now maybe the best option is to remove the check.

lib/ex_unit/lib/ex_unit/case.ex
```diff
@@ -532,15 +550,109 @@ defmodule ExUnit.Case do
...
+          test "doesn't have attributes", context do
+            assert context.registered.describe_fixtures == []
+          end
+        end
+      end
+
+  """
+  @spec register_describe_attribute(env, name :: atom, opts :: keyword) :: :ok
+  def register_describe_attribute(env, name, opts \\ []) do
+    do_register_attribute(:ex_unit_registered_describe_attributes, env, name, opts)
```
whatyouhide: We moved away from `do_` prefixes in Elixir for private functions. Since the arity of the functions is different, you can call the private function `register_attribute/4`.

lib/ex_unit/lib/ex_unit/case.ex
```diff
@@ -532,15 +550,109 @@ defmodule ExUnit.Case do
...
 
-  def register_attribute(%{module: module}, name, opts) do
-    register_attribute(module, name, opts)
+  defp do_register_attribute(type, %{module: module}, name, opts) do
+    do_register_attribute(type, module, name, opts)
   end
 
-  def register_attribute(mod, name, opts) when is_atom(mod) and is_atom(name) and is_list(opts) do
+  defp do_register_attribute(type, mod, name, opts)
+       when is_atom(type) and is_atom(mod) and is_atom(name) and is_list(opts) do
```
whatyouhide: I think it would be better to move the guards to the public `register_module_attribute/3` so that if a `FunctionClauseError` happens it will point to the public function.

lib/ex_unit/lib/ex_unit/case.ex
```diff
@@ -532,15 +550,109 @@ defmodule ExUnit.Case do
...
+  defp validate_registered_attribute!(type, mod, name) do
+    registered_attribute_keys = [
+      :ex_unit_registered_module_attributes,
+      :ex_unit_registered_describe_attributes,
+      :ex_unit_registered_test_attributes
+    ]
+
+    for key <- registered_attribute_keys,
+        type != key && name in Module.get_attribute(mod, key) do
+      raise "cannot register attribute #{inspect(name)} multiple times"
```
whatyouhide: Personally, I think this is a problem with the argument you gave to this function, so I'd raise an `ArgumentError` here.

```suggestion
      raise ArgumentError, "cannot register attribute #{inspect(name)} multiple times"
```

lib/ex_unit/lib/ex_unit/case.ex
```diff
@@ -532,15 +550,109 @@ defmodule ExUnit.Case do
...
+
+  defp validate_registered_attribute!(type, mod, name) do
+    registered_attribute_keys = [
+      :ex_unit_registered_module_attributes,
+      :ex_unit_registered_describe_attributes,
+      :ex_unit_registered_test_attributes
+    ]
+
+    for key <- registered_attribute_keys,
+        type != key && name in Module.get_attribute(mod, key) do
```
whatyouhide: The two sides of this boolean operation are strictly boolean, so use `and/2` instead of `&&/2`.

```suggestion
        type != key and name in Module.get_attribute(mod, key) do
```

lib/ex_unit/lib/ex_unit/case.ex
blatyo: I can pull up the others, but the `is_atom(mod)` is hard to pull out here, since the public function allows an ENV or a module. Recommendations for that?

blatyo: If I leave it off, it get's cause by Module.register_attribute?/3. Maybe that's good enough?

```
** (FunctionClauseError) no function clause matching in Module.register_attribute/3

    The following arguments were given to Module.register_attribute/3:

        # 1
        1

        # 2
        :foo

        # 3
        []

        def register_attribute(module, attribute, options) when is_atom(module) and is_atom(attribute) and is_list(options)

    (elixir) lib/module.ex:1249: Module.register_attribute/3
    iex:7: (module)
```

whatyouhide: If `mod` is not an atom isn't it a very weird error? I'm fine with that if that's the case. We tend to not use defensive guards in private functions.

josevalim: Yes, we should move the guards up for nice funcction clause errors as smuch as possible.

blatyo: This is addressed

lib/ex_unit/lib/ex_unit/case.ex
```diff
@@ -532,15 +550,125 @@ defmodule ExUnit.Case do
...
+  def register_module_attribute(%{module: mod}, name, opts) do
+    register_module_attribute(mod, name, opts)
+  end
+
+  def register_module_attribute(mod, name, opts)
+      when is_atom(mod) and is_atom(name) and is_list(opts) do
+    register_attribute(:ex_unit_registered_module_attributes, mod, name, opts)
+  end
+
+  defp register_attribute(type, %{module: module}, name, opts) do
```
fertapric: I think this clause is never executed :)

lib/ex_unit/lib/ex_unit/case.ex
```diff
@@ -532,15 +550,125 @@ defmodule ExUnit.Case do
...
 
-  def register_attribute(mod, name, opts) when is_atom(mod) and is_atom(name) and is_list(opts) do
+  @doc """
+  Registers a new describe attribute to be used during `ExUnit.Case` tests.
+
+  The attribute values will be available through `context.registered`.
+  Registered values are cleared after each `ExUnit.Case.describe/2` similar
+  to `@describetag`.
+
+  `Module.register_attribute/3` is used to register the attribute,
```
fertapric: I think we should not talk about the internals. What do you think about?

> This function takes the same options as `Module.register_attribute/3`.

lib/ex_unit/lib/ex_unit/case.ex
```diff
@@ -532,15 +550,125 @@ defmodule ExUnit.Case do
...
+  def register_attribute(mod, name, opts) when is_atom(mod) and is_atom(name) and is_list(opts) do
+    register_attribute(:ex_unit_registered_test_attributes, mod, name, opts)
   end
 
-  def register_attribute(mod, name, opts) when is_atom(mod) and is_atom(name) and is_list(opts) do
+  @doc """
+  Registers a new describe attribute to be used during `ExUnit.Case` tests.
+
+  The attribute values will be available through `context.registered`.
+  Registered values are cleared after each `ExUnit.Case.describe/2` similar
```
fertapric: Nitpick:

```suggestion
  Registered values are cleared after each `describe/2` similar
```

lib/ex_unit/lib/ex_unit/case.ex
```diff
@@ -389,6 +393,10 @@ defmodule ExUnit.Case do
       after
         @ex_unit_describe nil
         Module.delete_attribute(__MODULE__, :describetag)
+
```
fertapric: I think `describe/2` should raise if the attribute is set outside the `describe/2` context, in the same way it's done with `@describetag`: `** (RuntimeError) @describetag must be set inside describe/2 blocks`

Right now this is possible:

```elixir
@describe_foo :tortilla
@describe_bar :paella
describe "with attributes outside describe" do
  test "registered subscribe attributes are in context", context do
    assert context.registered.describe_foo == :tortilla
    assert context.registered.describe_bar == [:paella]
  end
end
```

lib/ex_unit/lib/ex_unit/case.ex
```diff
@@ -532,15 +550,125 @@ defmodule ExUnit.Case do
       end
 
   """
+  @spec register_attribute(env, name :: atom, opts :: keyword) :: :ok
```
fertapric: Minor nitpick: these type annotations feel a bit redundant having the function definition right below 🙂 

```suggestion
  @spec register_attribute(env, atom, keyword) :: :ok
```

lib/ex_unit/lib/ex_unit/case.ex
josevalim: We discussed this earlier and it is not possible right now, we don't have the proper API in `Module`. That's why there was a more recent proposal on the mailing list.

fertapric: Oh! I see https://github.com/elixir-lang/elixir/pull/9181#discussion_r298852918 😓

lib/ex_unit/lib/ex_unit/case.ex
blatyo: Good catch

lib/ex_unit/lib/ex_unit/case.ex
blatyo: Maybe this was for an old version of ExDoc, but I thought if you didn't provide a name for the type, it got named `arg1`, `arg2`, etc. automatically. But, I'm not seeing that now. I'll remove them.

josevalim:

PR #9185, opened 2019-07-03
Add mix deps.unlock --check-unused flag

asummers:
Analogous to `mix format --check-formatted`, errors if there is any unused dependencies when running `mix deps.unlock --check-unused`. 

Apologies if there is supposed to be some pre discussion before opening this, but it seemed simple enough and fitting with other existing `mix` flags. 

As an aside, it seems like the `--filter` flag is undocumented.

### DIFF
--- a/lib/mix/lib/mix/tasks/deps.unlock.ex
+++ b/lib/mix/lib/mix/tasks/deps.unlock.ex
@@ -35,7 +35,7 @@ defmodule Mix.Tasks.Deps.Unlock do
         unused_apps = Mix.Dep.Lock.read() |> Map.drop(apps)
 
         unless unused_apps == %{} do
-          Mix.shell().error("warning: unused dependencies in mix.lock file")
+          Mix.raise("warning: unused dependencies in mix.lock file")
         end
 
       opts[:unused] ->
--- a/lib/mix/lib/mix/tasks/deps.unlock.ex
+++ b/lib/mix/lib/mix/tasks/deps.unlock.ex
@@ -11,12 +11,15 @@ defmodule Mix.Tasks.Deps.Unlock do
 
     * `dep1 dep2` - the name of dependencies to be unlocked
     * `--all` - unlocks all dependencies
+    * `--check-unused` - checks that the mix.lock file has no unused
+      dependencies. This is useful in pre-commit hooks and CI scripts
+      if you want to reject contributions with extra dependencies.
     * `--unused` - unlocks only unused dependencies (no longer mentioned
       in the `mix.exs` file)
 
   """
 
-  @switches [all: :boolean, unused: :boolean, filter: :string]
+  @switches [all: :boolean, check_unused: :boolean, unused: :boolean, filter: :string]
 
   @impl true
   def run(args) do
@@ -26,6 +29,14 @@ defmodule Mix.Tasks.Deps.Unlock do
     cond do
       opts[:all] ->
         Mix.Dep.Lock.write(%{})
+
+      opts[:check_unused] ->
+        apps = Mix.Dep.load_on_environment([]) |> Enum.map(& &1.app)
+        unused_apps = Mix.Dep.Lock.read() |> Map.drop(apps)
+
+        unless unused_apps == %{} do
+          Mix.shell().error("warning: unused dependencies in mix.lock file")
+        end
 
       opts[:unused] ->
         apps = Mix.Dep.load_on_environment([]) |> Enum.map(& &1.app)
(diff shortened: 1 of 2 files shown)


lib/mix/lib/mix/tasks/deps.unlock.ex
```diff
@@ -27,6 +30,14 @@ defmodule Mix.Tasks.Deps.Unlock do
       opts[:all] ->
         Mix.Dep.Lock.write(%{})
 
+      opts[:check_unused] ->
+        apps = Mix.Dep.load_on_environment([]) |> Enum.map(& &1.app)
+        unused_apps = Mix.Dep.Lock.read() |> Map.drop(apps)
+
+        unless unused_apps == %{} do
+          Mix.shell().error("warning: unused dependencies in mix.lock file")
```
josevalim: We should use `Mix.raise` so the CLI exits with status 1.

lib/mix/test/mix/tasks/deps_test.exs
```diff
@@ -308,6 +308,32 @@ defmodule Mix.Tasks.DepsTest do
...
+      Mix.Dep.Lock.write(%{whatever: "0.2.0", ok: "0.1.0"})
+      assert Mix.Dep.Lock.read() == %{whatever: "0.2.0", ok: "0.1.0"}
+      Mix.Tasks.Deps.Unlock.run(["--check-unused"])
+      assert Mix.Dep.Lock.read() == %{whatever: "0.2.0", ok: "0.1.0"}
+      error = "warning: unused dependencies in mix.lock file"
+      assert_received {:mix_shell, :error, [^error]}
+    end)
+  end
+
+  test "checks lock file has no unused deps with --check-unused", context do
```
josevalim: The tests can be merged into where you assert there is something unused, then you fix it, then nothing is unused.

asummers: Done

lib/mix/lib/mix/tasks/deps.unlock.ex
asummers: Done

lib/mix/test/mix/tasks/deps_test.exs
asummers: Unless you actually want me to call `Mix.Tasks.Deps.Unlock.run(["--unused"])` explicitly? That actually seems like what you meant now that I say it out loud.

josevalim: it looks great, thanks!

asummers: This is now taken care of.

asummers:
Do we want to adjust the message verbiage at all to suggest running `mix deps.unlock --unused` in the event it raises?

remiprev:
Glad to see this feature implemented in the language directly!

> Do we want to adjust the message verbiage at all to suggest running mix deps.unlock --unused in the event it raises?

When we implemented this feature as a Mix task in our Elixir boilerplate project, we also chose to output the list of unused locked dependencies.

https://github.com/mirego/elixir-boilerplate/blob/master/lib/mix/tasks/check.unused_locked_deps.ex

Maybe we want to do this here too 😄

lib/mix/lib/mix/tasks/deps.unlock.ex
```diff
@@ -27,6 +30,21 @@ defmodule Mix.Tasks.Deps.Unlock do
...
+
+        unused_apps =
+          Mix.Dep.Lock.read()
+          |> Map.drop(apps)
+          |> Map.keys()
+          |> Enum.sort()
+
+        unless unused_apps == [] do
+          Mix.raise(
+            "Unused dependencies in mix.lock file: #{inspect(unused_apps, limit: :infinity)}"
```
whatyouhide: I think I would prefer better formatting here for the apps. For example:

````
Unused dependencies in mix.lock file:

  * :first_dep
  * :other_dep
```

asummers: Done.

lib/mix/lib/mix/tasks/deps.unlock.ex
```diff
@@ -11,12 +11,16 @@ defmodule Mix.Tasks.Deps.Unlock do
 
     * `dep1 dep2` - the name of dependencies to be unlocked
     * `--all` - unlocks all dependencies
+    * `--filter` - unlocks only deps matching the given name
     * `--unused` - unlocks only unused dependencies (no longer mentioned
       in the `mix.exs` file)
+    * `--check-unused` - checks that the mix.lock file has no unused
```
fertapric: Nitpick:

```suggestion
    * `--check-unused` - checks that the `mix.lock` file has no unused
```

fertapric:
Thanks @asummers! 💜

PR #9189, opened 2019-07-04
MIX_BUILD_PATH specifies the build_path (and has the env added on to it)

### LINKED ISSUE #9188: Using mix clean with MIX_BUILD_PATH can erase your source code files
### Environment

```
Erlang/OTP 22 [erts-10.4.3] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [hipe]

Elixir 1.9.0 (compiled with Erlang/OTP 22)
```
* Operating system: 
Arch linux / docker

### Current behavior

In Docker I usually set MIX_BUILD_PATH to avoid permission conflicts with the host. So in my docker container I have `MIX_BUILD_PATH=/_build` and my app in `/api`.

However, `mix clean` is expecting the build path to be two directories deep, ie `/_build/dev`. So it runs `Path.dirname()`, resulting in `"/"`. Then it adds to that path, eventually resulting in the following pattern: `/*/lib/your_app_name`, and it erases everything matching that. My application is in `/api/`, so this matches.

The result: my source code gets deleted.

### Expected behavior

I would expect `mix clean` to only delete build files.

### Possible solutions

1. add the env to `MIX_ENV_PATH` in `Mix.Project.build_path/1`
2. have `mix clean` match only build files (`.beam` etc)
3. document MIX_BUILD_PATH better so that users understand that it should be eg, `/_build/dev` (https://github.com/elixir-lang/elixir/blob/master/lib/mix/lib/mix.ex#L237)
4. modify `mix clean` to recognize when it is dealing with `MIX_BUILD_PATH` so that it doesn't remove the last directory from the path.
5. filter files to be deleted to make sure they aren't actually in your application directory
6. something else?

derekkraan:
Fixes #9188.

(@josevalim I think this is what you meant?)

### DIFF
--- a/lib/mix/lib/mix/project.ex
+++ b/lib/mix/lib/mix/project.ex
@@ -472,11 +472,11 @@ defmodule Mix.Project do
   """
   @spec build_path(keyword) :: Path.t()
   def build_path(config \\ config()) do
-    System.get_env("MIX_BUILD_PATH") || config[:env_path] || env_path(config)
+    config[:env_path] || env_path(config)
   end
 
   defp env_path(config) do
-    dir = config[:build_path] || "_build"
+    dir = System.get_env("MIX_BUILD_PATH") || config[:build_path] || "_build"
     subdir = build_target() <> build_per_environment(config)
     Path.expand(dir <> "/" <> subdir)
   end
--- a/lib/mix/test/mix/project_test.exs
+++ b/lib/mix/test/mix/project_test.exs
@@ -31,12 +31,10 @@ defmodule Mix.ProjectTest do
     end
 
     test "considers MIX_BUILD_PATH" do
-      try do
-        System.put_env("MIX_BUILD_PATH", "_build")
-        assert Mix.Project.build_path() == Path.join(File.cwd!(), "_build/dev")
-      after
-        System.delete_env("MIX_BUILD_PATH")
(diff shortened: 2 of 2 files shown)


josevalim:
@derekkraan precisely. Also please add a test to `mix/project_test.exs`.

derekkraan:
:ok_hand: will do

derekkraan:
Done! the try/after ensures that this test doesn't break other tests if it breaks (which it does on `master`).

derekkraan:
I see you pushed a change. I was going to put the call to `System.delete_env/1` in an `on_exit` block. Or would you prefer I leave it as it is now?

derekkraan:
Like so:
```elixir
test "considers MIX_BUILD_PATH" do                                       
  System.put_env("MIX_BUILD_PATH", "_build")                             
                                                                         
  on_exit(fn ->                                                          
    System.delete_env("MIX_BUILD_PATH")                                  
  end)                                                                   
                                                                         
  assert Mix.Project.build_path() == Path.join(File.cwd!(), "_build/dev")
end                                                                      
```

josevalim:
@derekkraan both are fine. the current one is good enough, so i would keep it as is. :)

josevalim:

PR #9194, opened 2019-07-05
Fix doctest in Code

fertapric:


josevalim:

josevalim:
(I went ahead and merge it to fix the build asap. :))

fertapric:
@josevalim I think it should be "whose values" instead of "whose value" https://english.stackexchange.com/questions/43820/proper-way-to-handle-plurals-with-whose

> Atoms are constants whose value**s** are their own name.

PR #9200, opened 2019-07-06
Add Module.has_attribute?/2

blatyo:
Original proposal here: https://groups.google.com/d/msg/elixir-lang-core/4b9-yQ5J9KE/mAputTVBBAAJ

### DIFF
--- a/lib/elixir/lib/module.ex
+++ b/lib/elixir/lib/module.ex
@@ -1257,6 +1257,7 @@ defmodule Module do
         Module.has_attribute?(__MODULE__, :value) #=> false
       end
   """
+  @doc since: "1.10.0"
   @spec has_attribute?(module, atom) :: boolean
   def has_attribute?(module, key) when is_atom(module) and is_atom(key) do
     assert_not_compiled!(__ENV__.function, module)
--- a/lib/elixir/lib/module.ex
+++ b/lib/elixir/lib/module.ex
@@ -1256,6 +1256,7 @@ defmodule Module do
         Module.delete_attribute(__MODULE__, :value)
         Module.has_attribute?(__MODULE__, :value) #=> false
       end
+
   """
   @doc since: "1.10.0"
   @spec has_attribute?(module, atom) :: boolean
--- a/lib/elixir/lib/module.ex
+++ b/lib/elixir/lib/module.ex
@@ -1232,6 +1232,40 @@ defmodule Module do
   end
 
   @doc """
+  Checks if the given attribute has been defined.
+
+  An attribute is defined if it has been registered with `register_attribute/3`
+  or assigned a value. If an attribute has been deleted with `delete_attribute/2`
+  it is no longer considered defined.
+
+  This function can only be used on modules that have not yet been compiled.
+
+  ## Examples
+
+      defmodule MyModule do
+        @value 1
+        Module.register_attribute(__MODULE__, :other_value)
+        Module.put_attribute(__MODULE__, :another_value, 1)
+
+        Module.has_attribute?(__MODULE__, :value) #=> true
+        Module.has_attribute?(__MODULE__, :other_value) #=> true
+        Module.has_attribute?(__MODULE__, :another_value) #=> true
+
+        Module.has_attribute?(__MODULE__, :undefined) #=> false
+
+        Module.delete_attribute(__MODULE__, :value)
+        Module.has_attribute?(__MODULE__, :value) #=> false
+      end
+  """
+  @spec has_attribute?(module, atom) :: boolean
+  def has_attribute?(module, key) when is_atom(module) and is_atom(key) do
+    assert_not_compiled!(__ENV__.function, module)
+    {set, _bag} = data_tables_for(module)
+
+    :ets.member(set, key)
+  end
+
+  @doc """
   Deletes the module attribute that matches the given key.
 
   It returns the deleted attribute value (or `nil` if nothing was set).
@@ -1308,6 +1342,9 @@ defmodule Module do
     if Keyword.get(options, :accumulate) do
       :ets.insert_new(set, {attribute, [], :accumulate}) ||
         :ets.update_element(set, attribute, {3, :accumulate})
+    else
+      :ets.insert_new(bag, {:attributes, attribute})
+      :ets.insert_new(set, {attribute, nil, :unset})
     end
 
     :ok
@@ -1775,7 +1812,7 @@ defmodule Module do
       [{_, _, :accumulate}] ->
         :lists.reverse(bag_lookup_element(bag, {:accumulate, key}, 2))
 
-      [{_, val, nil}] ->
+      [{_, val, line}] when line in [:unset, nil] ->
         val
 
       [{_, val, _}] ->
@@ -1864,8 +1901,11 @@ defmodule Module do
         :ets.insert(set, {key, value, line})
         :ets.insert(bag, {:attributes, key})
     else
-      :accumulate -> :ets.insert(bag, {{:accumulate, key}, value})
-      _ -> :ets.insert(set, {key, value, line})
+      :accumulate ->
+        :ets.insert(bag, {{:accumulate, key}, value})
+
+      _ ->
+        :ets.insert(set, {key, value, line})
     end
   end
 
--- a/lib/elixir/lib/module.ex
+++ b/lib/elixir/lib/module.ex
@@ -1812,11 +1812,11 @@ defmodule Module do
       [{_, _, :accumulate}] ->
         :lists.reverse(bag_lookup_element(bag, {:accumulate, key}, 2))
 
-      [{_, val, line}] when line in [:unset, nil] ->
+      [{_, val, line}] when is_integer(line) ->
+        :ets.update_element(set, key, {3, nil})
         val
 
       [{_, val, _}] ->
-        :ets.update_element(set, key, {3, nil})
         val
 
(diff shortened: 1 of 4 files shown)


blatyo:
The build is failing because register_attribute now keeps track of non-accumulated values in the attributes set. When a persisted attribute that is not accumulated is added to the set it has a value of `nil`. Before, the value wouldn't be in the set, so it wouldn't get added to `__info__(:attributes)`.

``` elixir
defmodule MyModule do
  Module.register_attribute(__MODULE__, :foo, persisted: true)
end

# before
MyModule.__info__(:attributes) #=> []

#after
MyModule.__info__(:attributes) #=> [foo: nil]
```

Should I fix this to replicate current behavior or fix mix to not expect the attribute to not be present? On the one hand, this is a change in behavior. On the other hand, if I've registered an attribute as persisted, is it expected for it be present even if it never received a value?

lib/elixir/lib/module.ex
```diff
@@ -1231,6 +1231,43 @@ defmodule Module do
...
+        Module.delete_attribute(__MODULE__, :value)
+        Module.has_attribute?(__MODULE__, :value) #=> false
+      end
+  """
+  @spec delete_attribute(module, atom) :: boolean
+  def has_attribute?(module, key) when is_atom(module) and is_atom(key) do
+    assert_not_compiled!(__ENV__.function, module)
+    {set, _bag} = data_tables_for(module)
+
+    case :ets.lookup(set, key) do
```
josevalim: You can use ets:member.

josevalim:
It is important to have a distinction between not set and set to nil. I think the solution is to set the third element of the tuple to `:unset` (or false). Then in get_attribute, you swap those two clauses:

https://github.com/elixir-lang/elixir/blob/fb576c13ccdcf745fe7b35ff4007d2bf7b320bbb/lib/elixir/lib/module.ex#L1817-L1822

The first clause should be `[{_, val, line}] when is_integer(line) ->` and the second can be `[{_, val, _}]`.

Then when reading persisted attributes for persistence (I think in `elixir_module`) ignore anything that is `unset`.

blatyo:
Yea, I started going down that path. Finishing up a test for that now.

josevalim:
Also, this PR changes the code so `@val` doesn't warn if `@val`was previously registered but not set (today it warns if it was registered but not set). But I think this is a positive change, just make sure to add a test for it too.

lib/elixir/src/elixir_module.erl
```diff
@@ -351,7 +351,8 @@ attributes(DataSet, DataBag, PersistedAttributes) ->
 lookup_attribute(DataSet, DataBag, Key) when is_atom(Key) ->
   case ets:lookup(DataSet, Key) of
     [{Key, _, accumulate}] -> bag_lookup_element(DataBag, {accumulate, Key}, 2);
-    [{Key, Value, _}] -> [Value];
+    [{Key, nil, unset}] -> [];
+    [{_, Value, _}] -> [Value];
```
blatyo: I tried matching on `is_integer(Line)` here, but it seems something with protocols adds a value, where the third element is `nil`.

lib/elixir/test/elixir/kernel/warning_test.exs
```diff
@@ -1413,6 +1413,22 @@ defmodule Kernel.WarningTest do
     purge(Sample)
   end
 
+  test "registered attribute with no use" do
```
blatyo: I put this here, because the test that warns about this issue is here. But, this test is checking that there is no warn, so it feels a bit weird.

lib/elixir/src/elixir_module.erl
blatyo:
```
== Compilation error in file lib/calendar/date_range.ex ==
** (CaseClauseError) no case clause matching: [{:protocol_impl, [protocol: Enumerable, for: Date.Range], nil}]
    lib/calendar/date_range.ex:26: (module)
```

josevalim: The value is nil every time a module attribute is read. Check `__get_attribute__` impl.

lib/elixir/src/elixir_module.erl
```diff
@@ -351,7 +351,8 @@ attributes(DataSet, DataBag, PersistedAttributes) ->
 lookup_attribute(DataSet, DataBag, Key) when is_atom(Key) ->
   case ets:lookup(DataSet, Key) of
     [{Key, _, accumulate}] -> bag_lookup_element(DataBag, {accumulate, Key}, 2);
```
josevalim: We should either match `Key` in all of them or in none of them. None is fine, therefore:

```suggestion
    [{_, _, accumulate}] -> bag_lookup_element(DataBag, {accumulate, Key}, 2);
```

lib/elixir/src/elixir_module.erl
```diff
@@ -351,7 +351,8 @@ attributes(DataSet, DataBag, PersistedAttributes) ->
...
-    [{Key, Value, _}] -> [Value];
+    [{Key, nil, unset}] -> [];
```
josevalim:
```suggestion
    [{_, nil, unset}] -> [];
```

josevalim:
@blatyo btw, you need to cchange `Module.__get_attribute__` as I mentioned above otherwise 1. registering an attribute 2. and immediately reading it after, will cause it to be persisted, since unset will become nil. You also need to make sure registering an attribute does not warn if it not read.

lib/elixir/test/elixir/kernel/warning_test.exs
josevalim: To assert there is no warning, you just write a regular test in module_test.exs. The fact the suite won't emit a warning while running it is enough.

josevalim: In any case, I think we should warn in the case below and I don't understand why we don't. :) The case we should not warn is if the attribute is registered and nothing is done with it.

lib/elixir/lib/module.ex
```diff
@@ -1864,8 +1900,15 @@ defmodule Module do
...
         :ets.insert(bag, {:attributes, key})
     else
-      :accumulate -> :ets.insert(bag, {{:accumulate, key}, value})
-      _ -> :ets.insert(set, {key, value, line})
+      :accumulate ->
+        :ets.insert(bag, {{:accumulate, key}, value})
+
+      :unset ->
+        :ets.insert(set, {key, value, line})
+        :ets.insert(bag, {:attributes, key})
```
blatyo: This was what was causing the warning not to happen. It was only inserting the value into the set and not the bag.

blatyo:
Everything should be addressed now.

> Also, this PR changes the code so @val doesn't warn if @val was previously registered but not set (today it warns if it was registered but not set).

I tried this out and didn't get a warning in Elixir 1.9.0?

``` elixir
defmodule X do
  Module.register_attribute(X, :y, [])
  Module.register_attribute(X, :z, accumulate: true)
end
#=> no warning
```

josevalim:
> I tried this out and didn't get a warning in Elixir 1.9.0?

Hrm, so good. :D Let's just make sure it continues this way.

lib/elixir/lib/module.ex
josevalim: Good catch. I think an alternative solution would be to write to`{:attributes, key}` table on `register_attribute`, can you please check? I think I prefer to move it there because the best would be to handle `:undef` in as few places as possible.

lib/elixir/lib/module.ex
```diff
@@ -1231,6 +1231,40 @@ defmodule Module do
...
+  it is no longer considered defined.
+
+  This function can only be used on modules that have not yet been compiled.
+
+  ## Examples
+
+      defmodule MyModule do
+        @value 1
+        Module.register_attribute(__MODULE__, :other_value)
+        Module.put_attribute(__MODULE__, :another_value)
```
josevalim: Is this a get attribute? or is it missing a third argument?

blatyo: missing third value

lib/elixir/lib/module.ex
blatyo: That appears to work. One thing I don't understand is why attributes are tracked under the attributes key in the bag, when you have the attributes in the set?

josevalim: The `attributes` key would be better called `warn_attributes`, because the only purpose is to annotate them for checking for warnings later. A later PR would be appreciated to rename it if you are interested.

The reason we use two of them is because it is the most efficient way to implement this. Traversing set tables in ETS is expensive, so the key-values are stored in the set and in the bag we keep all entries in case we need to traverse.

lib/elixir/lib/module.ex
```diff
@@ -1231,6 +1231,40 @@ defmodule Module do
...
+        Module.has_attribute?(__MODULE__, :other_value) #=> true
+        Module.has_attribute?(__MODULE__, :another_value) #=> true
+
+        Module.has_attribute?(__MODULE__, :undefined) #=> false
+
+        Module.delete_attribute(__MODULE__, :value)
+        Module.has_attribute?(__MODULE__, :value) #=> false
+      end
+  """
+  @spec has_attribute?(module, atom) :: boolean
```
fertapric:
```suggestion
  @doc since: "1.10.0"
  @spec has_attribute?(module, atom) :: boolean
```

fertapric:
Thanks @blatyo! 💜

PR #9205, opened 2019-07-08
Improve the documentation for Logger

whatyouhide:


lib/logger/lib/logger.ex
```diff
@@ -278,41 +279,6 @@ defmodule Logger do
         format: "\n$time $metadata[$level] $levelpad$message\n",
         metadata: [:user_id]
 
-  ## Metadata
```
whatyouhide: Just moved this out of the "Backends" section.

PR #9210, opened 2019-07-10
Configure Cirrus CI

fertapric:
This configures Cirrus CI so we can test it before fully migrating to it.

In order for this to work, Cirrus CI must be configured first in the GitHub marketplace: https://cirrus-ci.org/guide/quick-start/

A few notes:

* Unfortunately, sending emails when the build fails requires a bit more configuration than with a regular CI. The official way is to use GitHub actions: https://github.com/cirrus-actions/email, and that requires an email account from which the emails would be sent.
* Test in Windows are not using `make test_windows` to avoid executing `make test_formatted` that causes problems with the path separators.
* The Docker images for the Windows builds are hosted under [my personal account](https://hub.docker.com/r/fertapric/elixir-ci).

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,0 +1,75 @@
+env:
+  ELIXIR_ASSERT_TIMEOUT: 2000
+  LANG: C.UTF-8
+
+test_task:
+  container:
+    image: buildpack-deps:trusty
+
+  matrix:
+    - env:
+        OTP_RELEASE: OTP-22.0
+        CHECK_REPRODUCIBLE: true
+        CHECK_POSIX_COMPLIANT: true
+    - env:
+        OTP_RELEASE: OTP-21.3.8
+    - env:
+        OTP_RELEASE: OTP-21.2
+    - env:
+        OTP_RELEASE: OTP-21.1
+    - env:
+        OTP_RELEASE: OTP-21.0
+    - env:
+        OTP_RELEASE: OTP-20.3
+    - env:
+        OTP_RELEASE: OTP-20.2
+    - env:
+        OTP_RELEASE: OTP-20.1
+    - env:
+        OTP_RELEASE: OTP-20.0
+    - env:
+        OTP_RELEASE: maint
+      allow_failures: true
+    - env:
+        OTP_RELEASE: master
+      allow_failures: true
+
+  install_script:
+    - wget -O otp.tar.gz https://repo.hex.pm/builds/otp/ubuntu-14.04/${OTP_RELEASE}.tar.gz
+    - mkdir -p otp
+    - tar zxf otp.tar.gz -C otp --strip-components=1
+    - otp/Install -minimal $(pwd)/otp
+
+  test_script:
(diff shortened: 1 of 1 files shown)


fertapric:
Here is an example of the same branch in my fork: https://github.com/fertapric/elixir/pull/1

josevalim:
@fertapric fantastic work!

We can call the formatter directly then for the WIndows build. I want to run the formatter also to help us find any specific formatter bug on Windows. It is highly unlikely... but you never know.

We can also remove `make test_windows` once we remove appveyor. It was created to work around some appveyor specific issues.

Btw, the build is failing in your fork? Is that expected?

Finally, for the github actions, is a sendgrid or a similar account enough? Also, where do we configure who is going to receive the emails? :D

fertapric:
> We can call the formatter directly then for the WIndows build. I want to run the formatter also to help us find any specific formatter bug on Windows. It is highly unlikely... but you never know.

Sounds good to me

> Btw, the build is failing in your fork? Is that expected?

It was not 😞 But I think it might be related to https://github.com/elixir-lang/elixir/issues/9167

> Finally, for the github actions, is a sendgrid or a similar account enough? 

Yeah, that should be enough.

> Also, where do we configure who is going to receive the emails? :D

Right now [it gets the information from the commit](https://github.com/cirrus-actions/email/blob/master/internal/email_generator.go#L19), but we would probably create our own actions if needed.

josevalim:
@fertapric I have configured cirrus-ci on the elixir-lang org. Next I will create a sendgrid or mailgun account so we can configure e-mails.

eksperimental:
May I ask what are the reasons behind moving from Travis to Cirrus?

josevalim:
A single place to test against Unix, FreeBSD, and Windows.

eksperimental:
Awesome!

fertapric:
> Test in Windows are not using `make test_windows` to avoid executing `make test_formatted` that causes problems with the path separators.

This has been fixed by #9227

gmile:
Now that GitHub announced their CI/CD product which supports running the test suit on Windows, Linux and macOS, would it make sense to use it instead of CirrusCI?

wojtekmach:
GitHub Actions is in beta until November 13th [1] so might make sense to wait until it's generally available but in the meantime keep it in mind of course.

[1] https://github.blog/2019-08-08-github-actions-now-supports-ci-cd/

ZizhengTai:
Not sure if this is important, but Cirrus also supports FreeBSD. GitHub Actions doesn't seem to support FreeBSD, or at least this isn't mentioned in their announcement post.

josevalim:
Yup, the lack of FreeBSD support will likely block us from using GitHub CI.

yordis:
Hey @jclem, do you know if FreeBSD will be supported?

jclem:
We want to support FreeBSD—it's on our backlog, but we don't yet have a committed time frame for it. cc @chrispat

fertapric:
Closed in favor of https://github.com/elixir-lang/elixir/pull/9534

PR #9211, opened 2019-07-10
Fix formatter removing nested parens on function with generated name

### LINKED ISSUE #9204: Formatter incorrectly remove nested parens on unquote
```elixir
quote do
  unquote(:foo)() do
    "bar"
  end
end
```

becomes

```elixir
quote do
  unquote(:foo) do
    "bar"
  end
end
```

Which has different meaning (and AST).

cybrox:
This is intended to fix #9204.

Since I haven't contributed to the formatter before and I am not completely satisfied with this solution, I marked this as WIP. The PR does solve the problem by not removing the parens before the `do` block when being called from the `remote_to_algebra` clause that matches calls as described in the issue. **However**, I am not entirely sure how long `state` is carried around and if just adding a flag to it (that is not reset) has any unintended side effects.  (Tests are passing, but still)

Is there a more idiomatic way of passing a "flag" for this special case to `call_args_to_algebra` without modifying the state?

### DIFF
--- a/lib/elixir/lib/code/formatter.ex
+++ b/lib/elixir/lib/code/formatter.ex
@@ -1079,7 +1079,7 @@ defmodule Code.Formatter do
   defp local_to_algebra(fun, meta, args, context, state) when is_atom(fun) do
     skip_parens =
       cond do
-        meta?(meta, :closing) -> :required
+        meta?(meta, :closing) -> :skip_if_only_do_end
         local_without_parens?(fun, args, state) -> :skip_unless_many_args
         true -> :skip_if_do_end
       end
@@ -1097,9 +1097,10 @@ defmodule Code.Formatter do
     {doc, state}
   end
 
-  # parens may one of:
+  # parens may be one of:
   #
   #   * :skip_unless_many_args - skips parens unless we are the argument context
+  #   * :skip_if_only_do_end - skip parens if we are do-end and the only arg
   #   * :skip_if_do_end - skip parens if we are do-end
   #   * :required - never skip parens
   #
@@ -1115,11 +1116,16 @@ defmodule Code.Formatter do
 
     if blocks = do_end_blocks(meta, last, state) do
       {call_doc, state} =
-        if rest == [] do
-          {" do", state}
-        else
-          no_parens? = parens != :required
-          call_args_to_algebra_no_blocks(meta, rest, no_parens?, list_to_keyword?, " do", state)
+        case rest do
+          [] when parens == :required ->
+            {"() do", state}
+
+          [] ->
+            {" do", state}
+
+          _ ->
+            no_parens? = parens not in [:required, :skip_if_only_do_end]
+            call_args_to_algebra_no_blocks(meta, rest, no_parens?, list_to_keyword?, " do", state)
         end
 
       {blocks_doc, state} = do_end_blocks_to_algebra(blocks, state)
--- a/lib/elixir/test/elixir/code_formatter/calls_test.exs
+++ b/lib/elixir/test/elixir/code_formatter/calls_test.exs
@@ -450,6 +450,12 @@ defmodule Code.Formatter.CallsTest do
       assert_same "unquote(call)(one, two)"
 
       assert_same """
+      unquote(call)() do
+        :ok
+      end
+      """
+
+      assert_same """
       unquote(call)(one, two) do
         :ok
       end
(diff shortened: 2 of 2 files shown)


cybrox:
Turns out this PR _breaks_ the formatter by adding parens to macros with only a do block (e.g. `try do`) but I didn't notice because there's no tests for that. I will try to figure out why and add tests for those missed cases.

lib/elixir/lib/code/formatter.ex
```diff
@@ -1016,6 +1016,7 @@ defmodule Code.Formatter do
   defp remote_to_algebra({target, meta, args}, context, state) do
     {target_doc, state} = quoted_to_algebra(target, :no_parens_arg, state)
 
+    state = Map.put_new(state, :ignore_empty_parens?, true)
```
josevalim: We avoid relying on state because it is often more complex, as we need to make sure to clean it up and so on.

For this particular issue, we already have a way to control how parens are emitted, which is exactly the `:required` atom below. You can see all available values here: https://github.com/elixir-lang/elixir/pull/9211/files#diff-2f18294c6d186e77413788b902525c57R1101

You can simply check for `:required` in the code below and emit "() do" accordingly. However, if you do so, two other tests will fail. That's because I think we need a third mode, which would be like `:skip_if_empty_do_end`, that only skips if we are in the "empty do end" part.
My suggestion is to introduce a new option, called

josevalim:
> Turns out not resetting the state does cause an issue, so I would have to set & reset the flag (ugly) or find some other solution.

It is happening exactly because the state is passed around and the flag is leaking. I posted a comment above with a better way to go about this.

lib/elixir/test/elixir/code_formatter/calls_test.exs
```diff
@@ -554,6 +554,13 @@ defmodule Code.Formatter.CallsTest do
       })
       """
     end
+
+    test "with generated function name and parens" do
+      assert_same """
+      call(1)() do
+      end
```
josevalim: We should also add a test for:

```
Remote.call(1)() do
end
```

lib/elixir/lib/code/formatter.ex
cybrox: Yeah, I figured `state` was a stupid idea but I couldn't come up with anything better. I didn't even realize `parens` was passed there. I adjusted the code to use `parens` instead.

cybrox:
Updated the code to use `parens` and added a test for remote call, looks better now. :)

lib/elixir/lib/code/formatter.ex
```diff
@@ -1017,7 +1017,7 @@ defmodule Code.Formatter do
     {target_doc, state} = quoted_to_algebra(target, :no_parens_arg, state)
 
     {{call_doc, state}, wrap_in_parens?} =
-      call_args_to_algebra(args, meta, context, :required, true, state)
+      call_args_to_algebra(args, meta, context, :skip_if_empty_do_end, true, state)
```
josevalim: I was thinking that **the other use** of `:required` in the code base would become `:skip_if_empty_do_end`. That's because what we are doing exactly in here is not skip_if_empty_do_end. We are adding if empty do end.

lib/elixir/lib/code/formatter.ex
```diff
@@ -1116,9 +1116,10 @@ defmodule Code.Formatter do
     if blocks = do_end_blocks(meta, last, state) do
       {call_doc, state} =
         if rest == [] do
-          {" do", state}
+          paren_str = if parens == :skip_if_empty_do_end, do: "()", else: ""
```
josevalim: I would prefer to reduce the amount of nested conditionals. Maybe:

```
case rest do
  [] when parens == :skip_if_empty_do_end -> 
    {"() do", state}
  [] ->
    {" do", state}
  _ ->
    ...
```

lib/elixir/lib/code/formatter.ex
cybrox: You're right. I would suggest naming it `:skip_if_only_do_end`, since the do-end is the only argument. `:skip_if_empty_do_end` to me sounds like it requires the do-end block to be empty, which would make no sense.

josevalim: Sounds good!

josevalim:

PR #9214, opened 2019-07-11
Use Module.has_attribute?/2

blatyo:
I did a review of where `Module.get_attribute/2` was used in the code base, where `Module.has_attribute?/2` makes more sense and yields more correct behavior.


blatyo:
I appear to have misunderstood how deleting an accumulated attribute works. It's more like what I'd call `Module.reset_attribute/2`. I'll need to change some of this back because of that.

blatyo:
This PR should be good for review. I'm going to do ExUnit.Case in a separate PR.

blatyo:
The test failure seems to be unrelated to my changes.

fertapric:
Thanks @blatyo! 💜

Issue #9219, opened 2019-07-12
String.split/3 with regex and including captures does not preserve :parts option


sebsel:
* Elixir & Erlang/OTP versions (elixir --version): Elixir 1.8.1 (compiled with Erlang/OTP 21)
* Operating system: macOS 10.14.5

I'm not sure if this a bug or intended behaviour, but I could not find a real answer in existing issues. Also, the tests do not cover this part, so it seems like there is either no decision made, or it is really a bug :)
https://github.com/elixir-lang/elixir/blob/master/lib/elixir/test/elixir/string_test.exs#L78

When using `String.split/3` with a regex, a special option `:include_captures` comes available. When used in conjunction with the `:parts` option, the captures do not count towards the parts. It kind of makes sense they are not counted, because the caller asks for both, but it also makes sense to count them, for the caller clearly wants two parts.

The related method `Regex.split/3` has the same behaviour, and it's tests also do not test this case.
https://github.com/elixir-lang/elixir/blob/master/lib/elixir/test/elixir/regex_test.exs#L255

### Current behaviour

```elixir
iex(1)> String.split("1,2,3", ~r/,/, include_captures: true, parts: 2)
["1", ",", "2,3"]
iex(2)> Regex.split(~r/,/, "1,2,3", include_captures: true, parts: 2)
["1", ",", "2,3"]
iex(3)> Regex.split(~r/,/, ",2,3", include_captures: true, parts: 2)
["", ",", "2,3"]
iex(4)> Regex.split(~r/,/, ",2,3", include_captures: true, parts: 2, trim: true)
[",", "2,3"]
iex(5)> Regex.split(~r/,/, "1,2,3", include_captures: true, parts: 2, trim: true)
["1", ",", "2,3"]
iex(6)> Regex.split(~r/,/, "1,2,3,4,5,6,7,8,9,0", include_captures: true, parts: 5)
["1", ",", "2", ",", "3", ",", "4", ",", "5,6,7,8,9,0"]
```

### Expected behaviour

```elixir
iex(1)> String.split("1,2,3", ~r/,/, include_captures: true, parts: 2)
["1", ",2,3"]
iex(2)> Regex.split(~r/,/, "1,2,3", include_captures: true, parts: 2)
["1", ",2,3"]
iex(3)> Regex.split(~r/,/, ",2,3", include_captures: true, parts: 2)
["", ",2,3"]
[...]

PR #9220, opened 2019-07-13
Regex.split/3 documentation and tests for :include_captures and :parts

sebsel:
As discussed in #9219, this PR keeps the current behaviour, but documents it. :)


josevalim:

PR #9222, opened 2019-07-13
Use Module.has_attribute?/2 in ExUnit.Case

blatyo:
This changes ExUnit.Case to use `Module.has_attribute?/2` in the cases where it makes sense.


fertapric:
@blatyo are you addressing https://github.com/elixir-lang/elixir/pull/9181#discussion_r298852918 in a different PR? 🙂

blatyo:
Yea, that's not part of this. `has_attribute?` isn't sufficient by itself to address that. It needs to be able to distinguish between an attribute that's accumulated or not.

fertapric:
@blatyo Perfect then. Thanks! 💜

PR #9226, opened 2019-07-16
Increase understandability of the tuple documentation

axelson:
Since Elixir data structures are immutable, emphasize that creating a new tuple is required. Also add a `result` binding to make it more clear that the last line is being used as a result for later.

This is based on a conversation on the forum: https://elixirforum.com/t/tuple-documentation-is-confusing-for-me/23993/3


whatyouhide:
Thank you @axelson! 💟

axelson:
Thanks!

PR #9234, opened 2019-07-22
Improve typespecs for Registry.start_link/1 and Registry.child_spec/1.

derekkraan:
I noticed that `child_spec/1` was missing a typespec, and "opts" seemed a little mysterious. This PR adds the missing typespec for `child_spec/1` and also cleans up the typespec for `start_link/1` (I'm not sure why that `when` was there, but it looks incorrect).

### DIFF
--- a/lib/elixir/lib/registry.ex
+++ b/lib/elixir/lib/registry.ex
@@ -209,6 +209,14 @@ defmodule Registry do
   @typedoc "A full match spec used when selecting objects in the registry"
   @type spec :: [{match_pattern, guards, body}]
 
+  @typedoc "Options used for child_spec/1 and start_link/1"
+  @type start_option ::
+          {:keys, keys}
+          | {:name, registry}
+          | {:partitions, pos_integer}
+          | {:listeners, [atom]}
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/registry.ex
```diff
@@ -209,6 +209,15 @@ defmodule Registry do
...
   @type spec :: [{match_pattern, guards, body}]
 
+  @typedoc "Options used for child_spec/1 and start_link/1"
+  @type options :: [
+          keys: keys,
+          name: registry,
+          partitions: pos_integer,
+          listeners: [atom],
+          meta: [{meta_key, meta_value}]
+        ]
```
whatyouhide: I think that the pattern that we usually use is:

```elixir
@type option :: {:keys, keys} | {:name, registry} | ...

@spec start_link([option]) :: ...
```

Wdyt?

whatyouhide: Also, I'd probably call this `start_option` instead. :)

derekkraan: @whatyouhide updated with your suggestions

josevalim:

PR #9236, opened 2019-07-23
Piping into unary and binary operator deprecation fixes

hassanshaikley:
This test should be uncommented now!

😄

### DIFF
--- a/lib/elixir/lib/macro.ex
+++ b/lib/elixir/lib/macro.ex
@@ -277,23 +277,20 @@ defmodule Macro do
   end
 
   def pipe(expr, {op, line, args} = op_args, integer) when is_list(args) do
-    if is_atom(op) and Identifier.unary_op(op) != :error do
-      raise ArgumentError,
-            "cannot pipe #{to_string(expr)} into #{to_string(op_args)}, " <>
-              "piping into a unary operator is deprecated, please use the qualified name. " <>
-              "For example, Kernel.+(5), instead of +5"
-    else
-      {op, line, List.insert_at(args, integer, expr)}
-    end
-  end
+    cond do
+      is_atom(op) and Identifier.unary_op(op) != :error ->
+        raise ArgumentError,
+              "cannot pipe #{to_string(expr)} into #{to_string(op_args)}, " <>
+                "piping into a unary operator is deprecated, please use the qualified name. " <>
+                "For example, Kernel.+(5), instead of +5"
 
-  def pipe(expr, {call, line, args} = call_args, integer) when is_list(args) do
-    if is_atom(call) and Identifier.binary_op(call) != :error do
-      raise ArgumentError,
-            "cannot pipe #{to_string(expr)} into #{to_string(call_args)}, " <>
-              "the #{to_string(call)} operator can only take two arguments"
-    else
-      {call, line, List.insert_at(args, integer, expr)}
+      is_atom(op) and Identifier.binary_op(op) != :error ->
+        raise ArgumentError,
+              "cannot pipe #{to_string(expr)} into #{to_string(op_args)}, " <>
+                "the #{to_string(op)} operator can only take two arguments"
+
+      true ->
+        {op, line, List.insert_at(args, integer, expr)}
     end
   end
 
--- a/lib/elixir/lib/macro.ex
+++ b/lib/elixir/lib/macro.ex
@@ -276,6 +276,17 @@ defmodule Macro do
     {call, line, List.insert_at([], integer, expr)}
   end
 
+  def pipe(expr, {op, line, args} = op_args, integer) when is_list(args) do
+    if is_atom(op) and Identifier.unary_op(op) != :error do
+      raise ArgumentError,
+            "cannot pipe #{to_string(expr)} into #{to_string(op_args)}, " <>
+              "piping into a unary operator is deprecated, please use the qualified name. " <>
+              "For example, Kernel.+(5), instead of +5"
+    else
+      {op, line, List.insert_at(args, integer, expr)}
+    end
+  end
+
   def pipe(expr, {call, line, args} = call_args, integer) when is_list(args) do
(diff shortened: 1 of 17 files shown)


fertapric:
I think the tests are passing because `Macro.pipe` interpret it as a binary operator. The error in `Macro.pipe` is a bit confusing because it says: 

> cannot pipe 1 into +1, the :+ operator can only take two arguments

And the `|>` macro emits a warning and an error:

```elixir
iex(1)> 1 |> + 1
warning: piping into a unary operator is deprecated, please use the qualified name. For example, Kernel.+(5), instead of +5
  iex:1

** (ArgumentError) cannot pipe 1 into +1, the :+ operator can only take two arguments
    (elixir) lib/macro.ex:155: Macro.pipe/3
    (stdlib) lists.erl:1263: :lists.foldl/3
    (elixir) expanding macro: Kernel.|>/2
    iex:1: (file)
```

I would suggest to detect unary operators in `Macro.pipe` and raise a message similar to the warning above (`piping into a unary operator is deprecated, please use the qualified name. For example, Kernel.+(5), instead of +5`) and remove the warning from `|>`. Thoughts?

josevalim:
> I would suggest to detect unary operators in Macro.pipe and raise a message similar to the warning above (piping into a unary operator is deprecated, please use the qualified name. For example, Kernel.+(5), instead of +5) and remove the warning from |>. Thoughts?

Agreed.

fertapric:
> I would suggest to detect unary operators in `Macro.pipe` and raise a message similar to the warning above (`piping into a unary operator is deprecated, please use the qualified name. For example, Kernel.+(5), instead of +5`) and remove the warning from `|>`. Thoughts?

@hassanshaikley are you going to work on this? 🙂

hassanshaikley:
@fertapric yes! I am going to work on this!

hassanshaikley:
@fertapric @josevalim how does that look?

lib/elixir/lib/macro.ex
```diff
@@ -276,6 +276,17 @@ defmodule Macro do
     {call, line, List.insert_at([], integer, expr)}
   end
 
+  def pipe(expr, {op, line, args} = op_args, integer) when is_list(args) do
```
fertapric: I believe this is going to match first and it won't emit errors for binary operators. For example, this is not going to fail and it should: `Macro.pipe(1, quote(do: [] ++ []), 0)`

I would move the `if` inside the clause below.

fertapric:
@hassanshaikley This warning needs to be removed too: https://github.com/elixir-lang/elixir/blob/master/lib/elixir/lib/kernel.ex#L3410-L3420

lib/elixir/lib/macro.ex
```diff
@@ -276,13 +276,21 @@ defmodule Macro do
...
   end
 
-  def pipe(expr, {call, line, args} = call_args, integer) when is_list(args) do
-    if is_atom(call) and Identifier.binary_op(call) != :error do
-      raise ArgumentError,
-            "cannot pipe #{to_string(expr)} into #{to_string(call_args)}, " <>
-              "the #{to_string(call)} operator can only take two arguments"
-    else
-      {call, line, List.insert_at(args, integer, expr)}
+  def pipe(expr, {op, line, args} = op_args, integer) when is_list(args) do
```
hassanshaikley: I think the variable name `op` is misleading as it could be an `op` or a `call`. Not sure what to do.

lib/elixir/test/elixir/macro_test.exs
```diff
@@ -776,7 +776,7 @@ defmodule MacroTest do
       Macro.pipe(1, {:ok}, 0)
     end
 
-    assert_raise ArgumentError, ~r"cannot pipe 1 into 1 \+ 1", fn ->
+    assert_raise ArgumentError, ~r"cannot pipe 1 into 1 \+ 1, piping into a unary", fn ->
```
fertapric: Oh! I approved too fast. This assertion should not be "piping into a unary". it should say that the `+` operator can only take two arguments. To do that you can check the length of the args.

hassanshaikley: Okay!

I just pushed a change.  Let me know what you think.

hassanshaikley:
tests seems to have failed on CI even though they pass locally.

lib/elixir/lib/macro.ex
```diff
@@ -276,13 +276,21 @@ defmodule Macro do
...
+    cond do
+      is_atom(op) and Identifier.unary_op(op) != :error and length(args) == 1 ->
+        raise ArgumentError,
+              "cannot pipe #{to_string(expr)} into #{to_string(op_args)}, " <>
+                "the #{to_string(op)} operator can only take two arguments. " <>
```
fertapric: Unary operators can only take one argument. Here is my suggestion:

```elixir
def pipe(_expr, {op, _line, [arg]}, _integer) when op == :+ or op == :- do
  raise ArgumentError,
        "piping into a unary operator is not supported, please use the qualified name: " <>
          "Kernel.#{op}(#{to_string(arg)}), instead of #{op}#{to_string(arg)}"
end

def pipe(expr, {op, line, args} = op_args, integer) when is_list(args) do
  cond do
    is_atom(op) and Identifier.unary_op(op) != :error ->
      raise ArgumentError,
            "cannot pipe #{to_string(expr)} into #{to_string(op_args)}, " <>
              "the #{to_string(op)} operator can only take one argument"

    is_atom(op) and Identifier.binary_op(op) != :error ->
      raise ArgumentError,
            "cannot pipe #{to_string(expr)} into #{to_string(op_args)}, " <>
              "the #{to_string(op)} operator can only take two arguments"

    true ->
      {op, line, List.insert_at(args, integer, expr)}
  end
end
```

hassanshaikley: Ah, that makes more sense!

josevalim:

PR #9237, opened 2019-07-23
Remove newline in Example for Agent.start

hassanshaikley:
This seems inconsistent with the rest of the docs.


lib/elixir/lib/agent.ex
```diff
@@ -260,7 +260,6 @@ defmodule Agent do
       iex> {:ok, pid} = Agent.start_link(fn -> 42 end)
       iex> Agent.get(pid, fn state -> state end)
       42
```
josevalim: The line means they are different examples and can be run in isolation. It is important to keep it. :)

hassanshaikley: Oh, woops! My bad 😄

PR #9238, opened 2019-07-23
Fix typos

aerosol:
Just some :nail_care:


whatyouhide:
Thank you @aerosol! 💟

PR #9240, opened 2019-07-24
Replace distillery reference with `mix release`

gerbal:
Update docs on `Elixir.Application` to recommend building releases with `mix release` instead of distillery.


whatyouhide:
Thanks @gerbal! 💟

PR #9241, opened 2019-07-25
Rename Mix.Tasks.Escript.Build option to `:strip_beams`

jjcarstens:
In [`Mix.Tasks.Release`](https://github.com/elixir-lang/elixir/blob/master/lib/mix/lib/mix/tasks/release.ex#L371) this option is referred to as `:strip_beams`. And even in [this modules doc](https://github.com/elixir-lang/elixir/blob/master/lib/mix/lib/mix/tasks/escript.build.ex#L34) it is referenced to look at `:strip_beams` option. However, the implementation here only uses the singular `:strip_beam` (and is the only place to do so).

This actually caused a bit of confusion in https://github.com/nerves-project/nerves/pull/443.

So this just simply changes that option and updates the doc to match how it is used elsewhere.

### DIFF
--- a/lib/mix/lib/mix/tasks/escript.build.ex
+++ b/lib/mix/lib/mix/tasks/escript.build.ex
@@ -61,7 +61,7 @@ defmodule Mix.Tasks.Escript.Build do
       Defaults to app name. Set it to `nil` if no application should
       be started.
 
-    * `:strip_beam` - if `true` strips BEAM code in the escript to remove chunks
+    * `:strip_beams` - if `true` strips BEAM code in the escript to remove chunks
       unnecessary at runtime, such as debug information and documentation.
       Defaults to `true`.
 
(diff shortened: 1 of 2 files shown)


josevalim:
Unfortunately when the option is listed, it is listed correctly though. So we need to keep backwards compatibility. We should look at both `:strip_beams` and `:strip_beam`. Please also add a todo to the code, something like this:

     # TODO: Deprecate :strip_team option

jjcarstens:
Thanks @josevalim. Commented _right_ as I put up the backwards compatibility. I just made the change to add the note as well though so this should be ready for another 👀

lib/mix/lib/mix/tasks/escript.build.ex
```diff
@@ -163,7 +163,17 @@ defmodule Mix.Tasks.Escript.Build do
     end
 
     app = Keyword.get(escript_opts, :app, project[:app])
-    strip_beam? = Keyword.get(escript_opts, :strip_beam, true)
+
+    # Need to keep :strip_beam option for backwards compatibility so
```
eksperimental:
```suggestion
    # Need to keep :strip_beam option for backward compatibility so
```
Minor nitpick. Backwards is an adverb, backward is the adjective

jjcarstens: Touché. Lol. I’ll change it

jjcarstens: Changed

josevalim:

Issue #9148, opened 2019-06-17
Improve DateTime documentation with how to use timezone databases


mguimas:
### Environment

Elixir 1.8.2

### Current behavior

The documentation of the `DateTime` module does not inform readers how to use / choose / configure timezone databases. There are no practical examples of this, only fake examples.

### Expected behavior

The documentation should provide real examples of how to use / choose / configure time zone databases, like

```
iex(8)> datetime = DateTime.utc_now()
#DateTime<2019-06-17 20:30:41.706622Z>
iex(9)> DateTime.shift_zone(datetime, "Africa/Abidjan", Tzdata.TimeZoneDatabase)
{:ok, #DateTime<2019-06-17 20:30:41.706622+00:00 GMT Africa/Abidjan>}
iex(10)> datetime = DateTime.utc_now()
#DateTime<2019-06-17 20:31:04.094127Z>
iex(11)> DateTime.shift_zone(datetime, "Africa/Maputo", Tzdata.TimeZoneDatabase)
{:ok, #DateTime<2019-06-17 22:31:04.094127+02:00 CAT Africa/Maputo>}
```

This should avoid developers loosing time figuring things out.

mguimas:
The example above is with `Tzdata` library. The documentation should refer to recognized libraries in the community.

I wonder why does not Elixir come with such a library? 🤔

wojtekmach:
> The documentation of the DateTime module does not inform readers how to use / choose / configure timezone databases. There are no practical examples of this, only fake examples.

Have you seen the section about setting it in the module doc? https://hexdocs.pm/elixir/DateTime.html?#module-time-zone-database. How would you improve it?

Setting the DB like that is preferred because you'd usually have one tz db per application anyway and so you don't need to keep passing it around.

> The example above is with Tzdata library. The documentation should refer to recognized libraries in the community.

I believe official Elixir docs seldom (if ever) mention 3rd party packages so that would be my guess why it's omitted. Since tzdata is the de facto tz db I tend to agree it could be mentioned.

> I wonder why does not Elixir come with such a library? 

A timezone db is only as good as it's up to date and keeping it fresh is tricky. Also, if something doesn't have to be in core it usually isn't. :)

whatyouhide:
> I believe official Elixir docs seldom (if ever) mention 3rd party packages so that would be my guess why it's omitted. Since tzdata is the de facto tz db I tend to agree it could be mentioned.

[...]

PR #9246, opened 2019-07-28
Add Tzdata example to DateTime docs

adamu:
Added an example of how to use `Tzdata` to the `DateTime` docs.

This was requested in #9148.

I hope replacing the `CustomTimeZoneDatabase` with `Tzdata.TimeZoneDatabase` isn't too much - I think it's still obvious what to do if you want to use an alternative database.

I think the comment about adding the dependency is necessary, because if you just configure the module name without actually adding the dependency, you won't get any errors until you try to use the calendar, and then it's not obvious what the problem is.

I also opened https://github.com/lau/tzdata/pull/85 to flesh out the Tzdata's HexDocs a bit, which I though would be helpful for people that get linked there from here.


whatyouhide:
Thanks @adamu! 💟

PR #9247, opened 2019-07-29
Add compiler tracing

josevalim:
This feature allows IDEs and other tools wanting to
perform source code analysis to do so reliably without
a need to reimplement Elixir's compiler expansion and
without relying on Elixir's private APIs.

This commit also adds `:parser_options` to compiler
options, which allows developers to combine both options
to retrieve more accurate information, such as columns.

The lexical tracker has been rewritten to use the tracing
API.

### DIFF
--- a/lib/elixir/lib/code.ex
+++ b/lib/elixir/lib/code.ex
@@ -1006,7 +1006,7 @@ defmodule Code do
       #=> :ok
 
   """
-  @doc since: "v1.10.0"
+  @doc since: "1.10.0"
   @spec put_compiler_option(atom, term) :: :ok
   def put_compiler_option(key, value) when key in @boolean_compiler_options do
     if not is_boolean(value) do
--- a/lib/elixir/lib/code.ex
+++ b/lib/elixir/lib/code.ex
@@ -56,7 +56,7 @@ defmodule Code do
       to `as`. `meta` is the alias AST metadata and `opts` are the alias options.
 
     * `{:alias_expansion, meta, as, alias}` traced whenever there is an alias
-      expension, i.e. when the user writes `as` which is expanded to `alias`.
+      expansion, i.e. when the user writes `as` which is expanded to `alias`.
       `meta` is the alias expansion AST metadata.
 
     * `{:require, meta, module, opts}` - traced whenever `module` is required.
--- a/lib/elixir/lib/code.ex
+++ b/lib/elixir/lib/code.ex
@@ -71,7 +71,7 @@ defmodule Code do
 
     * `{:remote_function, meta, module, name, arity}` and
       `{:remote_macro, meta, module, name, arity}` - traced whenever a remote
-      function or macro is defined. `meta` is the call AST metadata, `module`
+      function or macro is referenced. `meta` is the call AST metadata, `module`
       is the invoked module, followed by the `name` and `arity`.
 
     * `{:local_function, meta, module, name, arity}` and
--- a/lib/elixir/lib/code.ex
+++ b/lib/elixir/lib/code.ex
@@ -63,7 +63,7 @@ defmodule Code do
       `meta` is the require AST metadata and `opts` are the require options.
 
     * `{:struct_expansion, meta, module}` - traced whenever `module`' struct
-      is expanded. `meta` is teh struct AST metadata.
+      is expanded. `meta` is the struct AST metadata.
 
     * `{:remote_reference, meta, module}` - traced whenever there is a remote
       reference to `module`, i.e. whenever the user writes `MyModule.Foo.Bar`
--- a/lib/elixir/lib/code.ex
+++ b/lib/elixir/lib/code.ex
@@ -66,7 +66,7 @@ defmodule Code do
     * `{:require, meta, module, opts}` - traced whenever `module` is required.
       `meta` is the require AST metadata and `opts` are the require options.
 
-    * `{:struct_expansion, meta, module}` - traced whenever `module`' struct
+    * `{:struct_expansion, meta, module}` - traced whenever `module`'s struct
       is expanded. `meta` is the struct AST metadata.
 
     * `{:remote_function, meta, module, name, arity}` and
--- a/lib/elixir/lib/code.ex
+++ b/lib/elixir/lib/code.ex
@@ -989,7 +989,7 @@ defmodule Code do
       at compilation time. This can be useful when doing dynamic compilation.
       Defaults to `[]`.
 
-    * `:tracers` - a list of tracers to be used during compilation. See the
+    * `:tracers` - a list of tracers (modules) to be used during compilation. See the
       module docs for more information. Defaults to `[]`.
 
     * `:parser_options` - a keyword list of options to be given to the parser
--- a/lib/elixir/lib/code.ex
+++ b/lib/elixir/lib/code.ex
@@ -19,7 +19,8 @@ defmodule Code do
 
     * `eval_file/2` - evaluates the file contents without tracking its name. It
       returns the result of the last expression in the file, instead of the modules
-      defined in it.
+      defined in it. Evaluated files do not trigger the Compilation Tracers described
+      in the next section.
 
   In a nutshell, the first must be used when you want to keep track of the files
   handled by the system, to avoid the same file from being compiled multiple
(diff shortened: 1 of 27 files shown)


msaraiva:
So cool! ❤️ 

With this one, I'll be able to remove some hacks in ElixirSense in the near future. Thanks!

lib/elixir/lib/code.ex
```diff
@@ -28,6 +29,61 @@ defmodule Code do
...
+
+    * `{:alias_expansion, meta, as, alias}` traced whenever there is an alias
+      expension, i.e. when the user writes `as` which is expanded to `alias`.
+      `meta` is the alias expansion AST metadata.
+
+    * `{:require, meta, module, opts}` - traced whenever `module` is required.
+      `meta` is the require AST metadata and `opts` are the require options.
+
+    * `{:struct_expansion, meta, module}` - traced whenever `module`' struct
+      is expanded. `meta` is teh struct AST metadata.
```
IvanIvanoff: typo: teh -> the

josevalim: Tks!

lib/elixir/lib/code.ex
```diff
@@ -19,7 +19,8 @@ defmodule Code do
 
     * `eval_file/2` - evaluates the file contents without tracking its name. It
       returns the result of the last expression in the file, instead of the modules
-      defined in it.
+      defined in it. Evaluated files do not trigger the Compilation Tracers described
```
fertapric:
```suggestion
      defined in it. Evaluated files do not trigger the compilation tracers described
```

lib/elixir/lib/code.ex
```diff
@@ -28,6 +29,61 @@ defmodule Code do
...
+      `{:imported_macro, meta, module, name, arity}` - traced whenever an
+      imported function or macro is invoked. `meta` is the call AST metadata,
+      `module` is the module the import is from, followed by the `name` and `arity`
+      of the imported function/macro.
+
+    * `{:alias, meta, alias, as, opts}` - traced whenever `alias` is aliased
+      to `as`. `meta` is the alias AST metadata and `opts` are the alias options.
+
+    * `{:alias_expansion, meta, as, alias}` traced whenever there is an alias
+      expension, i.e. when the user writes `as` which is expanded to `alias`.
```
fertapric:
```suggestion
      expansion, i.e. when the user writes `as` which is expanded to `alias`.
```

lib/elixir/lib/code.ex
```diff
@@ -934,35 +989,76 @@ defmodule Code do
       at compilation time. This can be useful when doing dynamic compilation.
       Defaults to `[]`.
 
-  It returns the new map of compiler options.
+    * `:tracers` - a list of tracers to be used during compilation. See the
```
fertapric:
```suggestion
    * `:tracers` - a list of tracers (modules) to be used during compilation. See the
```

lib/elixir/lib/code.ex
```diff
@@ -934,35 +989,76 @@ defmodule Code do
...
 
-      :elixir_config.put(key, value)
-      {key, value}
+    if key == :parser_options and not Keyword.keyword?(value) do
+      raise "compiler option #{inspect(key)} should be a keyword list, " <>
+              "got: #{inspect(value)}"
+    end
+
+    if key == :tracers and not Enum.all?(value, &is_atom/1) do
+      raise "compiler option #{inspect(key)} should be a list of atoms, " <>
```
fertapric:
```suggestion
      raise "compiler option #{inspect(key)} should be a list of modules, " <>
```

lib/elixir/lib/code.ex
```diff
@@ -934,35 +989,76 @@ defmodule Code do
...
 
-  It returns the new map of compiler options.
+    * `:tracers` - a list of tracers to be used during compilation. See the
+      module docs for more information. Defaults to `[]`.
+
+    * `:parser_options` - a keyword list of options to be given to the parser
+      when compiling files. It accepts the same options as `string_to_quoted/2`
+      (except by the options that change the AST itself). This can be used in
+      combination with the tracer to retrieve localized information about
+      events happening duing compilation. Defaults to `[]`.
```
fertapric:
```suggestion
      events happening during compilation. Defaults to `[]`.
```

lib/elixir/lib/code.ex
feliperenan: Typo? 🤔  expension => expansion

lib/elixir/lib/code.ex
```diff
@@ -934,35 +989,76 @@ defmodule Code do
...
   """
-  @spec compiler_options(Enumerable.t()) :: %{optional(atom) => boolean}
-  def compiler_options(opts) do
-    for {key, value} <- opts, into: %{} do
-      cond do
-        key in @boolean_compiler_options ->
-          if not is_boolean(value) do
-            raise "compiler option #{inspect(key)} should be a boolean, got: #{inspect(value)}"
-          end
+  @doc since: "v1.10.0"
```
fertapric:
```suggestion
  @doc since: "1.10.0"
```

lib/elixir/lib/code.ex
josevalim:
```suggestion
      expansion, i.e. when the user writes `as` which is expanded to `alias`.
```

lib/elixir/src/elixir_expand.erl
```diff
@@ -528,7 +528,7 @@ expand_fn_capture(Meta, Arg, E) ->
   case elixir_fn:capture(Meta, Arg, E) of
     {{remote, Remote, Fun, Arity}, EE} ->
       is_atom(Remote) andalso
-        elixir_lexical:record_remote(Remote, ?key(E, function), ?key(E, lexical_tracker)),
+        elixir_env:trace({remote_function, Meta, Remote, Fun, Arity}, E),
```
fertapric: Should `Meta` be the `AttachedMeta` below?

josevalim: It doesn't need to because anyone using tracing can compute the attached meta using `E`. The reason we have AttachedMeta is for the code verification pass which doesn't have `E` anymore.

fertapric: Ah! Ok. I was mentioning it because we are passing `AttachedDotMeta` instead of `DotMeta` [a few lines below](https://github.com/elixir-lang/elixir/pull/9247/files#diff-ed0344c13ed25389035f337d94902894R804).

lib/elixir/src/elixir_expand.erl
```diff
@@ -797,10 +798,11 @@ expand_local(Meta, Name, Args, #{module := Module, function := Function} = E) ->
 
 expand_remote(Receiver, DotMeta, Right, Meta, Args, #{context := Context} = E, EL) when is_atom(Receiver) or is_tuple(Receiver) ->
   assert_no_clauses(Right, Meta, Args, E),
+  AttachedDotMeta = attach_context_module(Receiver, DotMeta, E),
 
   is_atom(Receiver) andalso
-    elixir_lexical:record_remote(Receiver, ?key(E, function), ?key(E, lexical_tracker)),
-  AttachedDotMeta = attach_context_module(Receiver, DotMeta, E),
+    elixir_env:trace({remote_function, AttachedDotMeta, Receiver, Right, length(Args)}, E),
```
```suggestion
    elixir_env:trace({remote_function, DotMeta, Receiver, Right, length(Args)}, E),
```

lib/elixir/src/elixir_expand.erl
josevalim: Yeah, I would rather not pass it. Good catch!

lib/elixir/src/elixir_lexical.erl
```diff
@@ -1,84 +1,118 @@
...
+  Only =
+    case lists:keyfind(only, 1, Opts) of
+      {only, List} when is_list(List) -> List;
+      _ -> []
+    end,
+
+  ?tracker:add_import(Pid, Module, Only, ?line(Meta), Imported and Warn),
+  ok;
+trace({alias, Meta, _Old, New, Opts}, #{lexical_tracker := Pid}) ->
+  Warn =
```
fertapric: Maybe extract this to a `should_warn(Meta, Opts)`?

lib/elixir/lib/code.ex
```diff
@@ -28,6 +29,61 @@ defmodule Code do
...
+    * `{:struct_expansion, meta, module}` - traced whenever `module`' struct
+      is expanded. `meta` is the struct AST metadata.
+
+    * `{:remote_reference, meta, module}` - traced whenever there is a remote
+      reference to `module`, i.e. whenever the user writes `MyModule.Foo.Bar`
+      in the code, regardless if it is an alias or not.
+
+    * `{:remote_function, meta, module, name, arity}` and
+      `{:remote_macro, meta, module, name, arity}` - traced whenever a remote
+      function or macro is defined. `meta` is the call AST metadata, `module`
```
fertapric:
```suggestion
      function or macro is referenced. `meta` is the call AST metadata, `module`
```

Or `is invoked`?

Same for `:local_*`

lib/mix/lib/mix/tasks/compile.elixir.ex
```diff
@@ -26,17 +27,17 @@ defmodule Mix.Tasks.Compile.Elixir do
     * `--long-compilation-threshold N` - sets the "long compilation" threshold
       (in seconds) to `N` (see the docs for `Kernel.ParallelCompiler.compile/2`)
     * `--all-warnings` - prints warnings even from files that do not need to be recompiled
+    * `--tracer` - adds a compiler tracer in addition to any specified in the `mix.exs`
```
fertapric:
```suggestion
    * `--tracer` - adds a compiler tracer in addition to any specified in the `mix.exs`file
```

lib/elixir/lib/code.ex
```diff
@@ -28,6 +29,61 @@ defmodule Code do
...
+      in the code, regardless if it is an alias or not.
+
+    * `{:remote_function, meta, module, name, arity}` and
+      `{:remote_macro, meta, module, name, arity}` - traced whenever a remote
+      function or macro is referenced. `meta` is the call AST metadata, `module`
+      is the invoked module, followed by the `name` and `arity`.
+
+    * `{:local_function, meta, module, name, arity}` and
+      `{:local_macro, meta, module, name, arity}` - traced whenever a local
+      function or macro is defined. `meta` is the call AST metadata, `module`
```
josevalim:
```suggestion
      function or macro is reference. `meta` is the call AST metadata, `module`
```

lib/elixir/src/elixir_lexical.erl
josevalim: Ah yes, good catch. I didn't notice it is reused because they were moved over from different places at different times!

lib/elixir/lib/code.ex
fertapric: You forgot the "d" at the end 😬

josevalim: I am going to blame my keyword on this one!

lib/elixir/lib/code.ex
```diff
@@ -28,6 +29,61 @@ defmodule Code do
...
+      in the code, regardless if it is an alias or not.
+
+    * `{:remote_function, meta, module, name, arity}` and
+      `{:remote_macro, meta, module, name, arity}` - traced whenever a remote
+      function or macro is referenced. `meta` is the call AST metadata, `module`
+      is the invoked module, followed by the `name` and `arity`.
+
+    * `{:local_function, meta, module, name, arity}` and
+      `{:local_macro, meta, module, name, arity}` - traced whenever a local
+      function or macro is reference. `meta` is the call AST metadata, `module`
```
josevalim:
```suggestion
      function or macro is referenced. `meta` is the call AST metadata, `module`
```

lib/elixir/lib/code.ex
```diff
@@ -934,35 +989,76 @@ defmodule Code do
       at compilation time. This can be useful when doing dynamic compilation.
       Defaults to `[]`.
 
-  It returns the new map of compiler options.
+    * `:tracers` - a list of tracers (modules) to be used during compilation. See the
```
fertapric: Should we mark `:no_warn_undefined`, `:tracers`, and `:parser_options` options with `(since v1.10.0)`?

josevalim: Yes!

fertapric:
I think `:remote_reference` is not working for Erlang modules. 

For example: 

  * `:lists.last([1,2,3])` emits only `{:remote_function, [line: 1], :lists, :last, 1}` 
  * but `List.last([1,2,3])` emits `{:remote_reference, [line: 1], List}` and `{:remote_function, [line: 1], List, :last, 1}`.

josevalim:
Good point. I think we should rename it to `alias_reference`, because it only happens or aliases.

josevalim:
Or maybe we should use `alias_expansion` for both and we will know if they are the same or not by comparing `old != new`?

josevalim:
Using expansion is trickier because the expansion emits only the expanded bit. I've renamed it to `alias_reference`!

lib/elixir/lib/code.ex
```diff
@@ -28,6 +29,61 @@ defmodule Code do
...
+      which is expanded to `alias`. `meta` is the alias expansion AST metadata.
+
+    * `{:alias_reference, meta, module}` - traced whenever there is an alias
+      in the code, i.e. whenever the user writes `MyModule.Foo.Bar` in the code,
+      regardless if it was expanded or not.
+
+    * `{:require, meta, module, opts}` - traced whenever `module` is required.
+      `meta` is the require AST metadata and `opts` are the require options.
+
+    * `{:struct_expansion, meta, module}` - traced whenever `module`' struct
```
fertapric:
```suggestion
    * `{:struct_expansion, meta, module}` - traced whenever `module`'s struct
```

josevalim:
Thank you for the very through review @fertapric! :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

josevalim:

PR #9249, opened 2019-07-29
Add Keyword.pop!/2 and Map.pop!/2

whatyouhide:
These function behave like their non-bang counterparts except that they raise when the key is not present. This can be useful for option handling in order to extract an option and raise at the same time if it's not there. Note that this is mostly useful when not having the key there is an unrecoverable error (for example, because you already validated options): if you want to return a nice error to the user, `pop_lazy/3` with a `raise` in the third-argument function might be a better idea.

### DIFF
--- a/lib/elixir/lib/keyword.ex
+++ b/lib/elixir/lib/keyword.ex
@@ -955,6 +955,32 @@ defmodule Keyword do
   end
 
   @doc """
+  Returns the first value for `key` and removes all associated antries in the keyword list,
+  raising if `key` is not present.
+
+  This function behaves like `pop/3`, but raises in cases the `key` is not present in the
+  given `keywords`.
+
+  ## Examples
+
+      iex> Keyword.pop!([a: 1], :a)
+      {1, []}
+      iex> Keyword.pop!([a: 1, a: 2], :a)
+      {1, []}
+      iex> Keyword.pop!([a: 1], :b)
+      ** (KeyError) key :b not found in: [a: 1]
+
+  """
+  @doc since: "1.10.0"
+  @spec pop!(t, key) :: {value, t}
+  def pop!(keywords, key) when is_list(keywords) and is_atom(key) do
+    case fetch(keywords, key) do
+      {:ok, value} -> {value, delete(keywords, key)}
+      :error -> raise KeyError, key: key, term: keywords
+    end
+  end
+
+  @doc """
   Returns all values for `key` and removes all associated entries in the keyword list.
 
   It returns a tuple where the first element is a list of values for `key` and the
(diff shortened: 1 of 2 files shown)


lib/elixir/lib/map.ex
```diff
@@ -636,6 +636,31 @@ defmodule Map do
...
+      iex> Map.pop!(%{a: 1}, :a)
+      {1, %{}}
+      iex> Map.pop!(%{a: 1, b: 2}, :a)
+      {1, %{b: 2}}
+      iex> Map.pop!(%{a: 1}, :b)
+      ** (KeyError) key :b not found in: %{a: 1}
+
+  """
+  @doc since: "1.10.0"
+  @spec pop(map, key) :: {value, map}
```
wojtekmach:
```suggestion
  @spec pop!(map, key) :: {value, map}
```

lib/elixir/lib/keyword.ex
```diff
@@ -954,6 +954,32 @@ defmodule Keyword do
...
+      iex> Keyword.pop!([a: 1, a: 2], :a)
+      {1, []}
+      iex> Keyword.pop!([a: 1], :b)
+      ** (KeyError) key :b not found in: [a: 1]
+
+  """
+  @doc since: "1.10.0"
+  @spec pop!(t, key) :: {value, t}
+  def pop!(keywords, key) when is_list(keywords) and is_atom(key) do
+    case fetch(keywords, key) do
```
fertapric: Is this more efficient than using the new `pop_values/2`?

lib/elixir/lib/map.ex
whatyouhide: Yuck, good catch!

lib/elixir/lib/keyword.ex
whatyouhide: Yes. We don't need to build a list of values to pop just to take the first one. I think it's better to use the same implementation as `pop/3` and just raise in case there's no key.

lib/elixir/lib/map.ex
```diff
@@ -636,6 +636,31 @@ defmodule Map do
...
+      iex> Map.pop!(%{a: 1, b: 2}, :a)
+      {1, %{b: 2}}
+      iex> Map.pop!(%{a: 1}, :b)
+      ** (KeyError) key :b not found in: %{a: 1}
+
+  """
+  @doc since: "1.10.0"
+  @spec pop!(map, key) :: {value, map}
+  def pop!(map, key) do
+    case :maps.take(key, map) do
```
michalmuskala: Should we use the `:maps.take/2` function consistently in all the `pop`-style functions?

whatyouhide: We might but that's something for another day (and maybe someone who wants to do that? 😄)

josevalim: Done in master.

PR #9250, opened 2019-07-29
Allow StringIO to set encoding

wojtekmach:
Before this patch, the encoding was always unicode. This means the
following behaviour:

    iex> {:ok, pid} = StringIO.open("")
    iex> IO.binwrite(pid, <<127, 128>>)
    iex> StringIO.contents(pid)
    {"", <<127, 194, 128>>}

One example where this may be problematic is when using
`ExUnit.CaptureIO`, e.g. we're testing that a given function outputs
binary to stdout. Currently, we have a similar issue:

    iex> ExUnit.CaptureIO.capture_io(fn -> IO.binwrite(<<127, 128>>) end)
    <<127, 194, 128>>

With this patch we can do the following:

    iex> ExUnit.CaptureIO.capture_io(fn ->
    ...>   :ok = :io.setopts(Process.group_leader(), encoding: :latin1)
    ...>   IO.binwrite(<<127, 128>>)
    ...> end)
    <<127, 128>>

If this is accepted, I'm wondering if it would be useful to set the
encoding on `StringIO.open/2` and/or in `ExUnit.CaptureIO.capture_io/2`.
As @ericmj mentioned, there could be a new `CaptureIO.capture_bin_io/2`
function too.

All that being said, an alternative approach is the following:

    iex> output = ExUnit.CaptureIO.capture_io(fn -> IO.binwrite(<<127, 128>>) end)
    iex> :unicode.characters_to_binary(output, :unicode, :latin1)
    <<127, 128>>

So it could be a matter of documentation instead.

### DIFF
--- a/lib/elixir/lib/string_io.ex
+++ b/lib/elixir/lib/string_io.ex
@@ -21,13 +21,18 @@ defmodule StringIO do
   `string` will be the initial input of the newly created
   device.
 
-  If the `:capture_prompt` option is set to `true`,
-  prompts (specified as arguments to `IO.get*` functions)
-  are captured in the output.
-
   The device will be created and sent to the function given.
   When the function returns, the device will be closed. The final
   result will be a tuple with `:ok` and the result of the function.
+
+  ## Options
+
+    * `:capture_prompt` - if set to `true`, prompts (specified as
+      arguments to `IO.get*` functions) are captured in the output.
+      Defaults to `false`.
+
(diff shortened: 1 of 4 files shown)


josevalim:
Let’s make it an option on open so it can also be used by capture_io. :D
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

PR #9251, opened 2019-07-30
Make the wildcard example equivalent with `Mix.Config` in an umbrella project

zillou:
The current method to get wildcard paths is misleading, as it only works when you start your project from the project root. If someone is in an umbrella project and start a child app, the lookup method won't get any configs, and the result maybe quite confusing.

This PR changed the example to make the example code works completely equivalent with the `Mix.Config` example.


josevalim:

PR #9253, opened 2019-07-30
Fix release pid RPC tests

wuhkuh:
Addresses #9167.

The port was closed before calling `System.cmd(script, ["pid"])`, resulting in `--rpc-eval : RPC failed with reason :nodedown` errors.

As mentioned in the issue thread, I am not completely sure why the tests pass on [certain platforms](https://cloud.drone.io/alpinelinux/aports/7875/), while they seemingly shouldn't.


josevalim:
I believe it is a race condition. The system may not have fully shutdown and can still fulfill the pid request in certain cases.

fertapric:
Thanks @wuhkuh! 💜

PR #9254, opened 2019-07-30
Access.fetch!/2 implemented

hassanshaikley:
Hey all,

I noticed there's a `key!/1` but that's the only function in this module that supports a bang. I thought maybe `fetch/2` deserved the same love but I totally understand if not!

😄

### DIFF
--- a/lib/elixir/lib/access.ex
+++ b/lib/elixir/lib/access.ex
@@ -242,7 +242,7 @@ defmodule Access do
 
   ## Examples
 
-      iex> Access.fetch(%{name: "meg", age: 26}, :name)
+      iex> Access.fetch!(%{name: "meg", age: 26}, :name)
       "meg"
 
   """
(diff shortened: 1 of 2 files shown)


lib/elixir/lib/access.ex
```diff
@@ -236,6 +236,23 @@ defmodule Access do
...
+  or raises an `ArgumentError` exception if an error is returned.
+  """
+  @spec fetch!(container, term) :: term
+  def fetch!(container, key) do
+    case fetch(container, key) do
+      {:ok, value} ->
+        value
+
+      :error ->
+        raise ArgumentError,
```
josevalim: Please see how `fetch!` is implemented in Keyword. The best is to return `raise KeyError`.

hassanshaikley: Done!

chulkilee:
I'm not a core team member.. but @hassanshaikley could you squash commits into one? That helps future readers see changes easily :heart:

josevalim:
There is no need to squash as we will do that when we merge. :)

hassanshaikley:
contrary to what Travis is saying, I ran `make clean test` with no errors!

fertapric:
Thanks @hassanshaikley! 💜

PR #9256, opened 2019-08-01
Agent docs improvements

hassanshaikley:
Hey all,

Here are some humble improvements to `Agent` docs imo.

Let me know if you have any suggestions! Or anything else in the Agent docs you would like for me to look at.

### DIFF
--- a/lib/elixir/lib/agent.ex
+++ b/lib/elixir/lib/agent.ex
@@ -135,7 +135,7 @@ defmodule Agent do
 
   ## Name registration
 
-  An agent is bound to the same name registration rules as GenServers.
+  An agent is bound to the same name registration rules as `GenServer`s.
   Read more about it in the `GenServer` documentation.
 
   ## A word on distributed agents
--- a/lib/elixir/lib/agent.ex
+++ b/lib/elixir/lib/agent.ex
@@ -53,8 +53,7 @@ defmodule Agent do
   `GenServer`s). In particular, the functions passed as arguments to the calls to
   `Agent` functions are invoked inside the agent (the server). This distinction
   is important because you may want to avoid expensive operations inside the
-  agent, as they will effectively block the agent until the request is
-  fulfilled.
+  agent, as they will block the agent until the request is fulfilled.
 
   Consider these two examples:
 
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/agent.ex
```diff
@@ -74,8 +73,8 @@ defmodule Agent do
   at least initially, or small enough to be sent to the client cheaply. Another
   factor is whether the data needs to be processed atomically: getting the
   state and calling `do_something_expensive(state)` outside of the agent means
-  that the agent's state can be updated in the meantime. This is specially
-  important in case of updates as computing the new state in the client rather
+  that the agent's state can be updated in the meantime. This is especially
```
josevalim: I am not a native speaker but I believe "specially" works best here? "specially" is about being distinctive, "especially" is about being noteworthy.

lib/elixir/lib/agent.ex
```diff
@@ -74,8 +73,8 @@ defmodule Agent do
...
+  important when updating the state as computing the new state in the client rather
```
josevalim: I am not sure if the double `-ing` reads well here: "... updating ... computing ...".

lib/elixir/lib/agent.ex
hassanshaikley: I had no idea it was an actual word! TIL.

My bad!

lib/elixir/lib/agent.ex
hassanshaikley: Hmm I find it easier to make sense of in spite of the double `-ing` personally. But I could see how others would not!

lib/elixir/lib/agent.ex
```diff
@@ -21,7 +21,7 @@ defmodule Agent do
         end
 
         def value do
-          Agent.get(__MODULE__, & &1)
+          Agent.get(__MODULE__, &(&1))
```
hassanshaikley: Also how do you feel about that. It seems inconsistent that parens aren't always used.

hassanshaikley: nevermind I think this is a pattern!

hassanshaikley:
I understand if y'all want to close this, I have no hard feelings! 

These are just some improvements I thought might be nice. 😄 

💜

lib/elixir/lib/agent.ex
```diff
@@ -136,7 +135,7 @@ defmodule Agent do
 
   ## Name registration
 
-  An agent is bound to the same name registration rules as GenServers.
+  An agent is bound to the same name registration rules as `GenServer`s.
```
josevalim: We already link to GenServer next, so the current approach is fine. :)

josevalim:
Hi @hassanshaikley, thanks for the PR. Since those are minor changes that we are not super sure about, let's keep things as is, thanks!

hassanshaikley:
Of course! 😄

Have an excellent night!

PR #9263, opened 2019-08-03
Fix container algebra on diffs

ggcampinho:
When the elements of the container are too big and rendered in different lines, the separator is missing.
This change fixes that.

References #9259


ggcampinho:
Before:
Screenshot 2019-08-03 at 17 43 02

After:
Screenshot 2019-08-03 at 17 43 44

ericmj:
💜

PR #9265, opened 2019-08-04
Document escaping in eex more clearly

axelson:
When I was searching the eex docs I wasn't able to find any info about "escaping" and didn't realize that the `<%%` form was meant for escaping an eex expression inside of eex.

### DIFF
--- a/lib/eex/lib/eex.ex
+++ b/lib/eex/lib/eex.ex
@@ -76,6 +76,12 @@ defmodule EEx do
         This will never appear
       <% end %>
 
+  To escape eex in eex use `<%% content %>`. For example:
+
+      <%%= x + 3 %>
+
+  Will be rendered as `<%= x + 3 %>`
+
   Notice that different engines may have different rules
   for each tag. Other tags may be added in future versions.
 
--- a/lib/eex/lib/eex.ex
+++ b/lib/eex/lib/eex.ex
@@ -80,7 +80,7 @@ defmodule EEx do
 
       <%%= x + 3 %>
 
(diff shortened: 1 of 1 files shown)


lib/eex/lib/eex.ex
```diff
@@ -76,6 +76,12 @@ defmodule EEx do
         This will never appear
       <% end %>
 
+  To escape eex in eex use `<%% content %>`. For example:
```
eksperimental:
```suggestion
  To escape an EEx expression in EEx use `<%% content %>`. For example:
```

lib/eex/lib/eex.ex
```diff
@@ -76,6 +76,12 @@ defmodule EEx do
...
+
+      <%%= x + 3 %>
+
+  Will be rendered as `<%= x + 3 %>`
```
eksperimental:
```suggestion
  will be rendered as `<%= x + 3 %>`
```

eksperimental: I will avoid the used of code blocks for code that would fit in a single line.

axelson: I like the separate code block because I think it will enhance the scannability of the docs. In same fashion I would consider the "Will" on this line to be starting a new sentence (and hence keep the capitalization as-is). I am willing to be overruled but I wanted to state/defend my opinion.

josevalim: I agree code blocks are better here.

fertapric:
Thanks @axelson! 💜

axelson:
Great, thanks! :smile:

PR #9266, opened 2019-08-05
WIP: prototyping Macro.to_string/1 enhancement

steven-solomon:
# AST with macro definitions yield different strings than macro invocations

I am working on a mix task that does mutations on the quoted expression returned by `Code.string_to_quoted/1`. I would like to be able to use `Macro.to_string/1` to convert the new quoted expression back into code. So I made a start on implementing this behavior, but I have hit a snag.

Currently, `Macro.to_string/1` is used by the code documentation mechanism to generate a header for macros. The current code I am working on modifies the `Macro` module to act as I expect, however I am seeing the test title "prints function/macro documentation" located at `iex/test/iex/helpers_test.exs:377` fail. 

```
actual: * defmacro def call, expr \\ nil
expected: * defmacro def(call, expr \\ nil)
```

Clearly I’m misunderstanding how the document generation works. Could you point me in the right direction?

Thanks
Steve

### DIFF
--- a/lib/elixir/lib/macro.ex
+++ b/lib/elixir/lib/macro.ex
@@ -780,6 +780,10 @@ defmodule Macro do
     fun.(ast, "defmacro " <> to_string({macro_name, context, args}, fun) <> " do\nend")
   end
 
+  def to_string({:defmodule, _, [{module_name, context, args} | _rest]} = ast, fun) do
+    fun.(ast, "defmodule " <> to_string({module_name, context, args}, fun) <> " do\nend")
+  end
+
   # Bits containers
   def to_string({:<<>>, _, parts} = ast, fun) do
     if interpolated?(ast) do
@@ -1146,15 +1150,7 @@ defmodule Macro do
   defp call_to_string_with_args(target, args, fun) do
     target = call_to_string(target, fun)
     args = args_to_string(args, fun)
-    format_call(target, args)
-  end
-
-  defp format_call(target, args) do
-    case target do
-      "defmodule" -> target <> " " <> args
-      "def" -> target <> " " <> args
-      _ -> target <> "(" <> args <> ")"
-    end
+    target <> "(" <> args <> ")"
   end
 
   defp call_to_string_for_atom(atom) do
--- a/lib/elixir/lib/macro.ex
+++ b/lib/elixir/lib/macro.ex
@@ -775,6 +775,11 @@ defmodule Macro do
     fun.(ast, "(\n  " <> block <> "\n)")
   end
 
+  # Macro definition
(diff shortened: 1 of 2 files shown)


lib/elixir/lib/macro.ex
```diff
@@ -775,6 +775,11 @@ defmodule Macro do
     fun.(ast, "(\n  " <> block <> "\n)")
   end
 
+  # Macro definition
+  def to_string({:defmacro, _, [{macro_name, context, args} | _rest]} = ast, fun) do
+    fun.(ast, "defmacro " <> to_string({macro_name, context, args}, fun) <> " do\nend")
```
steven-solomon: a hard-coded " do\nend" will not be the final implementation. =)

lib/elixir/lib/macro.ex
```diff
@@ -1141,7 +1146,15 @@ defmodule Macro do
...
     target = call_to_string(target, fun)
     args = args_to_string(args, fun)
-    target <> "(" <> args <> ")"
+    format_call(target, args)
+  end
+
+  defp format_call(target, args) do
+    case target do
+      "defmodule" -> target <> " " <> args
+      "def" -> target <> " " <> args
```
steven-solomon: I also want to remove a reference to the "def" macro.

josevalim:
The reason we don’t do so is because def and defp are regular functions in Elixir. If we special case then, what about defdelegate, what about defcall in the ExActor library?

One option has been to change Macro._to_string to use the formatter behind the scenes. @whatyouhide, do you have any news on this front?

steven-solomon:
Exactly, I’m planning to generalize this solution to work with any macro invocation, not just `def`.

josevalim:
That’s actually much harder than it sounds. :) first of all, to know if something is a macro or not, you need to execute Elixir expansion and code. For example, you need to process imports. Sometimes the import may come from a `use` or another macro, so you need to expand and execute said code. The code may have side effects, which means that Macro.to_string now may have side effects, which is a change in behavior.

Also, we cannot simply remove parens. For example, if you remove parens in `add(1, 2) + 3`, it changes the meaning of the code, so you need to do an analysis of the context. So I really don’t think this is easily attainable in Macro.to_string. Good news is that this is all done in the formatter, which is why we started exploring reusing it.

steven-solomon:
Is there anything I can do to help out on that?

josevalim:
@whatyouhide was the one working on it, so let's wait to see if we have updates.

whatyouhide:
The situation is a bit weird yes. I've opened #9291 to open up the discussion even more. FWIW, I have no progress yet on what is discussed in #9291 :)

josevalim:
Moving the disccussion there. Thanks!

PR #9267, opened 2019-08-06
Remove undocumented process flags

jayjun:
Erlang [deliberately omitted](https://github.com/erlang/otp/blob/51d33208b08a257b59a64a35a953bb44c864f115/erts/preloaded/src/erlang.erl#L2251-L2253) `:monitor_nodes` from `process_flag/2` docs. 

To avoid confusion, `Process.flag/2` docs should omit these flags too .


josevalim:

Issue #7519, opened 2018-03-30
mix deps.compile unexpectedly compiles apps in umbrella


alco:
## Observed behaviour

Create a dummy umbrella:
```
$ mix new --umbrella foo_umbrella
$ cd foo_umbrella/apps
$ mix new foo_app
```

Now in the umbrella root (`foo_umbrella/`) run the following:
```
$ mix deps.compile
==> foo_app
Compiling 1 file (.ex)
Generated foo_app app

$ mix deps.compile
<no output>

$ touch apps/foo_app/mix.exs
$ mix deps.compile
==> foo_app
Compiling 1 file (.ex)
Generated foo_app app
```

## Expected behaviour

`mix deps.compile` must not compile apps within the umbrella.

alco:
I've noticed that `Mix.Dep.loaded([])` includes `%Mix.Dep{app: :foo_app, ...}` when it is invoked in the project root. I don't know if it's expected but looks like it could be the culprit here.

josevalim:
Yes, this is by design. Each umbrella child is considered to be a dependency of the umbrella.

josevalim:
@AndrewDryga you want to compile the dependencies but not the umbrella children? or only the umbrella children?

AndrewDryga:
@josevalim dependencies without the umbrella children. So that I can cache compiled dependencies and don't spend time compiling it on each CI job execution. Alex had an exactly same use case.

josevalim:
@AndrewDryga in this case a pull request that adds a `--skip-umbrella-children` to `mix deps.compile` would be welcome.

wojtekmach:
@AndrewDryga what you can also do in the meantime is cache the entire `_build/test` (including your apps'  beam) and `mix compile --force` when building the project. It won't recompile deps, unless they've changed but make sure your code is always recompiled from scratch.

PR #9273, opened 2019-08-07
Allow mix deps.compile to ignore umbrella apps

olafura:
This addresses the problem outlined in
https://github.com/elixir-lang/elixir/issues/7519

`mix deps.compile` can be a good way of caching builds
in a CI environment. Also it's good contain the warnings of the `mix
deps.compile` from the ones of `mix compile`.


josevalim:
Looks good. We just need a test. I think we can test this by getting a basic test from mix/umbrella_test, call mix deps.compile with the flag, and assert that nothing was compiled.

olafura:
@josevalim I had to use "umbrella_dep/deps/umbrella" because "umbrella_test" doesn't have any apps depending on other umbrella apps.

fertapric:
Thanks @olafura! 💜

PR #9274, opened 2019-08-07
Take with negative number should just call drop

seanmor5:
Take when `n < 0` is just the same as calling drop with n=count(enumerable)+amount. Old take performs 6x slower than it should.

### DIFF
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -2501,6 +2501,17 @@ defmodule Enum do
     take_list(enumerable, amount)
   end
 
+  def take(enumerable, amount)
+      when is_list(enumerable) and is_integer(amount) and amount < 0 do
+    count = length(enumerable)
+
+    if amount * -1 > count do
+      enumerable
+    else
+      drop_list(enumerable, count + amount)
+    end
+  end
+
   def take(enumerable, amount) when is_integer(amount) and amount > 0 do
     {_, {res, _}} =
       Enumerable.reduce(enumerable, {:cont, {[], amount}}, fn entry, {list, n} ->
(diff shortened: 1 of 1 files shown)


josevalim:
The problem with this approach is that it traverses the collection twice, which can be really harmful for things that are expensive to traverse, such as stream or files.

josevalim:
For the reason above, we never perform two operations at once in the Enumerable functions. Can you share the benchmarks you were running and how you were running them? Maybe we can make them faster using other approaches.

seanmor5:
Here's a basic elixir implementation of `take/2` using `drop/2` when `n<0`.
```
defmodule MyEnum do
  def take(list, n) when n > 0 do
    Enum.take(list, n)
  end
  def take(list, n) when n < 0 do
    count = Enum.count(list)
    if n*-1 > count do
      Enum.take(list, n)
    else
      Enum.drop(list, count-n)
    end
  end
end
```

Run with a list of 1,000,000 elements vs. `Enum.take/2`. Inputs are `n=500_000` and `n=-500_000`. Here are the results:
benchmarks

`Enum.take/2` is 1.5x slower. (Admittedly my 6x figure was coming from me comparing `Enum.drop/2` called with `n>0` vs. `Enum.take/2` called with `n<0`, my mistake!)

josevalim:
Ok. So this happens because our current implementation says how many to drop and how many to take. In this case, because we want to take everything, that ends up being unecessary work. We could extend the protocol to say "take :infinity" as an optimization but I am not sure if we should augment our current contract for this case.

ericmj:
Could we add an `is_list/1` clause with this optimization and keep the current implementation for other enumerables?

josevalim:
@ericmj we could but solving it for everyone would be preferable, if it can be trivially done.

ericmj:
@josevalim Could we use `Enumerable.count/1` to decide which implementation to use?

josevalim:
I would like to avoid doing so too as sometimes it may be expensive (like a
disk access) but not always as expensive as traversing it.
-- 


josevalim:
One of the options I have in mind for this is to allow the second argument to the slicing fun to be `:infinity`, which means we are interested in the end of the list. The downside is that we are augmenting the contract and it would have to be implemented for all data types.

fertapric:
@josevalim I think that would be a backwards incompatible change, current implementations of the `Enumerable` protocol would not understand that `:infinity`.

One option is to pass the size as a third argument to the slicing fun. That would be similar to passing `:infinity` but we could provide a smooth transition by checking the arity of the slicing fun.

josevalim:
@fertapric wait! The slice returns a function and that function can be a closure with the size!!!! Genius!

PR #9275, opened 2019-08-08
Doc) Remove missing link and ambiguous expression

jung-hunsoo:
This PR updates documentation.

- `Kernel.struct/2` doesn't work with multiple structs.
- In the `Typespecs` page, the `t/0` of `Range` doesn't link a proper Range type; instead create `file:///T:/Range.t/0` which is unexpected. And the actual `Range.t/0` is duplicated in the same sentence.


josevalim:

PR #9277, opened 2019-08-08
Update @doc of File.open/2

### LINKED ISSUE #9276: Pid returned by File.open/2 isn't linked to caller process while docs say it should be
### Environment

* Elixir & Erlang/OTP versions (elixir --version): Erlang/OTP 21, Elixir 1.9.0
```
[gianluca@localhost ~]$ elixir --version
Erlang/OTP 21 [erts-10.3.5.4] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [hipe]

Elixir 1.9.0 (compiled with Erlang/OTP 21)
```
* Operating system: Linux x86_64, Fedora 30

### Current behavior
The process whose pid is returned by calling `File.open/2` does not seem to be linked to the caller process, since it doesn't appear in the list of processes linked to the caller, and the caller doesn't get notified if the file's process dies. See this `iex` session as example:

```
iex(1)> {:ok, f_pid} = File.open("/tmp/some_test_file", [:write])
{:ok, #PID<0.106.0>}
iex(2)> Process.info(self(), [:links])
[links: []]
iex(3)> Process.alive?(f_pid)
true
iex(4)> Process.exit(f_pid, :kill) # kill it
true
iex(5)> Process.alive?(f_pid) # it's now dead
false
iex(6)> receive do # but caller mailbox is empty
...(6)> msg -> IO.inspect(msg)
...(6)> after 1000 -> IO.puts("received nothing")
...(6)> end
received nothing
:ok
```

### Expected behavior
From the [documentation](https://hexdocs.pm/elixir/1.9.0/File.html#open/2):
>This function returns:
>
>    {:ok, io_device} - the file has been opened in the requested mode.
>
>    io_device is actually the PID of the process which handles the file. This process is linked to the process which originally opened the file. If any process to which the io_device is linked terminates, the file will be closed and the process itself will be terminated.

So I'd expect to either (I'm not sure which one was the intended behaviour, but as said in my understanding the docs are inconsistent with reality):
* see the returned pid in the list of processes linked to the caller and to have the process which called `open` get a :EXIT message if the file process dies, like it happens if I manually link it with `Process.link(f_pid)`, or
* not have that paragraph in the documentation

gianluca-nitti:
Fixes #9276

### DIFF
--- a/lib/elixir/lib/file.ex
+++ b/lib/elixir/lib/file.ex
@@ -1355,9 +1355,11 @@ defmodule File do
     * `{:ok, io_device}` - the file has been opened in the requested mode.
 
       `io_device` is actually the PID of the process which handles the file.
-      This process is linked to the process which originally opened the file.
-      If any process to which the `io_device` is linked terminates, the file
-      will be closed and the process itself will be terminated.
+      This process monitors the process that originally opened the file (the
+      owner process). If the owner process terminates, the file is closed and
+      the process itself terminates too. If any process to which the `io_device`
+      is linked terminates, the file will be closed and the process itself will
+      be terminated.
 
       An `io_device` returned from this call can be used as an argument to the
       `IO` module functions.


josevalim:

PR #9278, opened 2019-08-08
Make Module.open?/1 and family to match their docs

am-kantox:
This should be treated mostly as a proposal, but since it’s always easier to explain with code, I’ve created a PR.

Docs for `Module.open?/1` say:

```elixir
  @doc """
  Checks if a module is open.

  A module is "open" if it is currently being defined and its attributes and
  functions can be modified.
  """
  @spec open?(module) :: boolean
```

The above breaks contract if called from `__after_compile__/2` callback. What leads to the fact all the modifying functions, like `Module.put_attribute/3` are allowed and silently ignored. My concern is `Module.open?/1` should not return `true` from there, and it could have been fixed as proposed below, save for it might break the existing code (as might be seen in `lib/ex_unit/lib/ex_unit/case.ex` below.)

On the other hand, it’d always be a compilation time exception, which might be easily fixed provided it’s backed up with a good error message. OTOH, it might make external libraries to fail and I have zero ideas of how many of them.

The lightweight solution would be to raise from _modifying the module functions_, or to warn from there. I am not a big fan of breaking changes, but I am positive the deprecation and maybe raising in the future would be a proper and helpful change. After all, they did not work from there since day one.

Thoughts?

### DIFF
--- a/lib/elixir/src/elixir_code_server.erl
+++ b/lib/elixir/src/elixir_code_server.erl
@@ -9,7 +9,8 @@
   required=#{},
   mod_pool={[], [], 0},
   mod_ets=#{},
-  compilation_status=#{}
+  compilation_status=#{},
+  mod_opened=#{}
 }).
 
 call(Args) ->
@@ -26,6 +27,7 @@ start_link() ->
 init(ok) ->
   %% The table where we store module definitions
   _ = ets:new(elixir_modules, [set, protected, named_table, {read_concurrency, true}]),
+  _ = ets:new(elixir_opened_modules, [set, protected, named_table, {read_concurrency, true}]),
   {ok, #elixir_code_server{}}.
 
 handle_call({defmodule, Module, Pid, Tuple}, _From, Config) ->
@@ -39,6 +41,9 @@ handle_call({defmodule, Module, Pid, Tuple}, _From, Config) ->
 
 handle_call({undefmodule, Ref}, _From, Config) ->
   {reply, ok, undefmodule(Ref, Config)};
+
+handle_call({unopenmodule, Ref}, _From, Config) ->
+  {reply, ok, unopenmodule(Ref, Config)};
 
 handle_call({acquire, Path}, From, Config) ->
   Current = Config#elixir_code_server.required,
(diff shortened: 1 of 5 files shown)


josevalim:
Yeah, this is tricky. On the `after_callback` it basically becomes "read-only". So one option is for us to store the mode "read or write" and allow `get_attribute` to work but not `put_attribute`. But I think the simplest path is to say in `after_callback` the module is still open but any definition, module attribute change, won't be persisted.

am-kantox:
> the simplest path is to say in `after_callback` the module is still open but any definition, module attribute change, won't be persisted.

Guess how I ever discovered this :) Basically, I made a mistake, after copypaste and forget to change the name and arity, it smoothly ran and the tests were just silently failing. I think the warning from there should be at least considered as an option.

josevalim:
So I would go the read/write route. It is a bit of work but should be quite doable. Feel free to open up an issue OR update the PR. :)

am-kantox:
> So I would go the read/write route. 

Yes, explicitly taking into account that my “fix” to `lib/ex_unit/lib/ex_unit/case.ex` actually broke it. I will update the PR.

> to store the mode "read or write" and allow `get_attribute` to work but not `put_attribute`.

The following methods should raise from `after_callback`

- `delete_attribute/2`
- `eval_quoted/4`
- `make_overridable/2`
- `put_attribute/3`
- `register_attribute/3`

All the others should work as now. `open?/1` returns `true`. Correct?

josevalim:
Perfect, yes!

lib/elixir/src/elixir_code_server.erl
```diff
@@ -26,6 +27,7 @@ start_link() ->
 init(ok) ->
   %% The table where we store module definitions
   _ = ets:new(elixir_modules, [set, protected, named_table, {read_concurrency, true}]),
+  _ = ets:new(elixir_opened_modules, [set, protected, named_table, {read_concurrency, true}]),
```
josevalim: We should not use a separate table  for this. When we store the module in the `elixir_modules` table, we should have a field in there that says if it is `open` / `read_only`.

josevalim:

PR #9280, opened 2019-08-09
Allow {:from_app, app_name} as a version for releases

Gazler:
Sometimes it is desireable to lookup the version from another
application to use as the version for a release. This is true in the
case of umbrella applications, where a particular app may be targetted
for a release.

Using `{:from_app, :my_app}` will allow the version returned from
`Application.spec(:my_app, :vsn)` to be used as the version.


fertapric:
Thanks @Gazler! 💜

PR #9282, opened 2019-08-10
Improve assertions on the process mailbox

### LINKED ISSUE #6248: Improve ExUnit output for assertions on process mailbox

ggcampinho:
Uses `ExUnit.Diff` to improve `ExUnit.Assertions.assert_receive/1` and `ExUnit.Assertions.assert_received/1`

The formatting slightly changed to accommodate the diffs and it's more similar to the other assertions

Closes #6248

### DIFF
--- a/lib/ex_unit/lib/ex_unit/assertions.ex
+++ b/lib/ex_unit/lib/ex_unit/assertions.ex
@@ -544,6 +544,9 @@ defmodule ExUnit.Assertions do
     else
       {message, mailbox} = format_mailbox(messages)
 
+      # The error contains a special `context` that will be transformed
+      # into `{:match, pins}` by the formatter to execute a diff for each
+      # message in the `mailbox`
       raise ExUnit.AssertionError,
         left: pattern,
         expr: code,
--- a/lib/ex_unit/lib/ex_unit/assertions.ex
+++ b/lib/ex_unit/lib/ex_unit/assertions.ex
@@ -7,7 +7,6 @@ defmodule ExUnit.AssertionError do
 
   defexception left: @no_value,
                right: @no_value,
-               mailbox: @no_value,
                message: @no_value,
                expr: @no_value,
                args: @no_value,
@@ -545,19 +544,13 @@ defmodule ExUnit.Assertions do
     else
       {message, mailbox} = format_mailbox(messages)
 
-      mailbox =
-        case mailbox do
-          [] -> ExUnit.AssertionError.no_value()
-          _ -> {pattern, mailbox}
-        end
-
       raise ExUnit.AssertionError,
-        mailbox: mailbox,
+        left: pattern,
         expr: code,
         message:
           "Assertion failed, no matching message after #{timeout}ms" <>
             ExUnit.Assertions.__pins__(pins) <> message,
-        context: {:match, pins}
+        context: {:mailbox, pins, mailbox}
     end
   end
 
--- a/lib/ex_unit/lib/ex_unit/assertions.ex
+++ b/lib/ex_unit/lib/ex_unit/assertions.ex
@@ -546,6 +546,12 @@ defmodule ExUnit.Assertions do
     else
       {message, mailbox} = format_mailbox(messages)
 
+      mailbox =
+        case mailbox do
+          [] -> ExUnit.AssertionError.no_value()
+          _ -> {pattern, mailbox}
+        end
+
       raise ExUnit.AssertionError,
         mailbox: mailbox,
         expr: code,
--- a/lib/ex_unit/lib/ex_unit/assertions.ex
+++ b/lib/ex_unit/lib/ex_unit/assertions.ex
@@ -437,8 +437,6 @@ defmodule ExUnit.Assertions do
       send(self(), :bye)
       assert_received :hello, "Oh No!"
       ** (ExUnit.AssertionError) Oh No!
-      Process mailbox:
-        :bye
 
   You can also match against specific patterns:
 
@@ -906,8 +904,6 @@ defmodule ExUnit.Assertions do
       send(self(), :hello)
       refute_received :hello, "Oh No!"
       ** (ExUnit.AssertionError) Oh No!
-      Process mailbox:
-        :bye
 
   """
   defmacro refute_received(pattern, failure_message \\ nil) do
--- a/lib/ex_unit/lib/ex_unit/assertions.ex
+++ b/lib/ex_unit/lib/ex_unit/assertions.ex
@@ -531,11 +531,10 @@ defmodule ExUnit.Assertions do
   @doc false
   def __timeout__(pattern, code, pins, pattern_finder, timeout) do
     {:messages, messages} = Process.info(self(), :messages)
-    binary = Macro.to_string(pattern)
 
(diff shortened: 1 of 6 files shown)


ggcampinho:
Should I handle this type of case?
```
test "variable outside of the scope" do
      send(self(), {:message, 1})
      
      y = :world
      assert_received x when y == :hello
    end
```

I believe yes, but I wanted to check

ggcampinho:
Screenshot 2019-08-10 at 21 08 52
Screenshot 2019-08-10 at 21 09 01

lib/ex_unit/lib/ex_unit/diff.ex
```diff
@@ -155,6 +159,45 @@ defmodule ExUnit.Diff do
...
+    {diff_expression, post_env} = compare_quoted(expression, right, env)
+
+    vars = Map.merge(post_env.pins, post_env.current_vars)
+    bindings = for {{name, _context}, value} <- vars, do: {name, value}
+    {diff_clause, clause_equivalent?} = compare_guard_clause(clause, bindings)
+
+    diff = %__MODULE__{
+      diff_expression
+      | left: {:when, [], [diff_expression.left, diff_clause]},
+        equivalent?: diff_expression.equivalent? && clause_equivalent?
```
josevalim:
```suggestion
        equivalent?: diff_expression.equivalent? and clause_equivalent?
```

lib/ex_unit/lib/ex_unit/diff.ex
```diff
@@ -155,6 +159,45 @@ defmodule ExUnit.Diff do
...
+
+    {diff, post_env}
+  end
+
+  defp compare_guard_clause({op, _, [clause1, clause2]}, bindings)
+       when op in [:when, :or, :and] do
+    {diff_clause1, clause1_equivalent?} = compare_guard_clause(clause1, bindings)
+    {diff_clause2, clause2_equivalent?} = compare_guard_clause(clause2, bindings)
+
+    equivalent? = clause1_equivalent? || clause2_equivalent?
```
josevalim:
```suggestion
    equivalent? = clause1_equivalent? or clause2_equivalent?
```

lib/ex_unit/lib/ex_unit/assertions.ex
```diff
@@ -7,6 +7,7 @@ defmodule ExUnit.AssertionError do
 
   defexception left: @no_value,
                right: @no_value,
+               mailbox: @no_value,
```
josevalim: Instead of adding a new key, maybe we can do context: {:mailbox, pins, mailbox}?

lib/ex_unit/lib/ex_unit/assertions.ex
```diff
@@ -527,12 +529,12 @@ defmodule ExUnit.Assertions do
...
 
   @doc false
-  def __timeout__(binary, pins, pattern_finder, timeout) do
+  def __timeout__(pattern, code, pins, pattern_finder, timeout) do
     {:messages, messages} = Process.info(self(), :messages)
 
     if Enum.any?(messages, pattern_finder) do
       """
-      Found message matching #{binary} after #{timeout}ms.
+      Found message matching #{Macro.to_string(pattern)} after #{timeout}ms.
```
josevalim: We should probably raise on both branches then? Otherwise it is weird one raises and the other returns the error message.

lib/ex_unit/test/ex_unit/assertions_test.exs
```diff
@@ -331,12 +331,14 @@ defmodule ExUnit.AssertionsTest do
...
-        No message matching {:status, ^status} after 0ms.
+        Assertion failed, no matching message after 0ms
         The following variables were pinned:
           status = :valid
-        Process mailbox:
-          {:status, :invalid}\
+        Showing 1 of 1 message in the mailbox\
         """ = error.message
+
+        {pattern, [{:status, :invalid}]} = error.mailbox
```
josevalim: I am not a big fan of asserting the mailbox. The field is an implementation detail and we should probably document it as so.

lib/ex_unit/lib/ex_unit/assertions.ex
josevalim: Another option is to expose the relevant APIs in ExUnit.Diff so we can compute the diff from `assertions` instead of having to pass all information inside.

ggcampinho: The problem with exposing `ExUnit.Diff` is that `ExUnit.Formatter` has knowledge if the diffs actually should run or not, the colors that they should be in, and they also have some timeouts for the diff. I don't if it's worth to duplicate all those things or just ignore them.

I can change the context to something like `{:mailbox, pins, pattern, mailbox}` and change the `ExUnit.Diff` to return a list of diffs when doing that, but I would need to change the timeout in the formatter to be multiplied by the number of messages. Does this makes sense?

josevalim: Good points!

> I can change the context to something like {:mailbox, pins, pattern, mailbox} and change the ExUnit.Diff to return a list of diffs when doing that

You can pre-process the assertionerror struct and convert it into the structure you have today where you pass mailbox and `{:match, pins}`. My only goal is to avoid adding more fields to the future. But multiplying the diff per number of messages is also fine. :+1:

lib/ex_unit/lib/ex_unit/formatter.ex
```diff
@@ -138,20 +139,18 @@ defmodule ExUnit.Formatter do
...
-      code: unless_value(struct.expr, fn -> get_code(test, stack) || @no_value end),
-      arguments: if_value(struct.args, &format_args(&1, width)),
-      left: left,
-      right: right
+      {:note, if_value(struct.message, &format_message(&1, formatter))},
+      {:doctest, if_value(struct.doctest, &code_multiline(&1, 2 + byte_size(@counter_padding)))},
+      {:code, if_value(struct.expr, &code_multiline(&1, padding_size))},
+      {:code, unless_value(struct.expr, fn -> get_code(test, stack) || @no_value end)},
+      {:arguments, if_value(struct.args, &format_args(&1, width))}
+      | format_context(struct, formatter, inspect, padding_size, width)
```
josevalim: You can keep the same syntax as before and do right after:

```
[
  ...
]
|> Kernel.++(format_context(struct, formatter, inspect, padding_size, width))
|> ...
```

The behaviour and performance would be the same. :)

josevalim:
Absolutely fantastic, thank you! :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

PR #9283, opened 2019-08-12
Fix function arity in the DynamicSupervisor docs

jung-hunsoo:
The type `init_option()` is used in `init/1` and `start_link/1`.


josevalim:

Issue #9284, opened 2019-08-12
Path.join unexpected behavior


SofaKing18:
### Environment

* Elixir & Erlang/OTP versions (elixir --version): Elixir 1.9.1 (compiled with Erlang/OTP 20)
* Operating system: Ubuntu 18.04

### Current behavior

```
Path.join("folder1/subfolder2/", "/")           
"folder1/subfolder2/."
```

### Expected behavior
```
Path.join("folder1/subfolder2/", "/")           
"folder1/subfolder2/"
```

Dot, in the end, is weird and unexpected, is this bug? 

Didn't find this behavior in documentation

josevalim:
Yup, it is a bug. :) We should have a special case to handle when the right side is `"/"`. A PR would be welcome!

josevalim:
To be clear, I think the result should be `"folder1/subfolder2"` as join removes trailing slashes.

cybrox:
Do you want to take a stab at this @SofaKing18? Could also look into it.

cybrox:
For the sake of completion, the issue also exists with `join/1`. E.g.

```
Path.join(["/foo", "bar", "/"]) -> "/foo/bar/."
```

SofaKing18:
I would like to make PR, but if @cybrox  want  - go for it :)

cybrox:
@SofaKing18 Ah, feel free to do so then. I'm just trying to pick up issues that the reporter does not PR themselves ;)

@josevalim `Path.join/2` states _The right path will always be expanded to its relative format_, I feel a hint to this behaviour should be added to `Path.join/1` as well. I think the following can be confusing?
```
iex(13)> Path.join(["", "/", "/usr/bin"])
"./usr/bin"

iex(14)> Path.join(["/", "/", "/usr/bin"])
"/./usr/bin"
```

josevalim:
Path.join/1 is implemented on top of Path.join/2 so fixing one should fix the other.

cybrox:
I know. Fixing the reported issue would fix the second example, the first one would still persist, though. Shouldn't an empty list element be ignored instead of making the whole path relative?

PR #9285, opened 2019-08-12
Fix join/2 dot result at path end

SofaKing18:
PR for [Issue Path.join/2](https://github.com/elixir-lang/elixir/issues/9284)


josevalim:

PR #9287, opened 2019-08-13
update  capture option doc

GuSilverFlame:
this PR is a little update on the `Regex` documentation to inform that the `:all_names` capture option brings the results as a list ordered alphabetically by the `subpattern` names


whatyouhide:
Thank you @GuSilverFlame! 💟

PR #9292, opened 2019-08-14
Add Map.filter/2 based on :maps.filter

romul:
Since OTP 18 there is a pretty convenient function [:maps.filter](http://erlang.org/doc/man/maps.html#filter-2).
I think it would be nice to have `Map.filter(map, predicate)` in Elixir as well instead of doing

```elixir
map
|> Enum.filter(predicate)
|> Enum.into(%{})
```

### DIFF
--- a/lib/elixir/lib/map.ex
+++ b/lib/elixir/lib/map.ex
@@ -339,6 +339,22 @@ defmodule Map do
   end
 
   @doc """
+  Filters the `map`, i.e. returns only those key-value pairs
+  for which `fun` returns a truthy value.
+
+  ## Examples
+
+      iex> Map.filter(%{x: 1, y: 2, z: 3}, fn(k, v) -> is_atom(k) && rem(v, 2) == 0 end)
+      %{y: 2}
+
+  """
+  @doc since: "1.10.0"
+  @spec filter(map, (key, value -> as_boolean(term))) :: map
+  def filter(map, fun) when is_function(fun, 2) do
+    :maps.filter(fun, map)
+  end
+
+  @doc """
   Evaluates `fun` and puts the result under `key`
   in `map` unless `key` is already present.
 


whatyouhide:
Hey @romul, there are alternative ways to do this in Elixir. For example:

```elixir
for {key, value} <- map, some_filter(key, value), into: %{}, do: {key, value}
```

Maps also implement the `Enumerable` protocol so you have `Enum.filter/2` and `Enum.drop/2` (we'd need to add even `Map.drop/2` then but we already have that for another reason... it's a bit messy 😛).

In general, we only use the issue tracker for bugs or accepted features. For feature discussions, please use the [mailing list](https://groups.google.com/group/elixir-lang-core).

Thanks anyways! 💟

OvermindDL1:
Because I like benchmarks, I'm curious which of a few different map filtering styles is fastest.  The names are descriptive for all, however the `Enum.filter` one by itself leaves it as a list instead of converting back to a map, I added it as a baseline for the other methods and their conversion costs, and the `for` one is @whatyouhide's above (with `some_filter(key, value)` replaced with `rem(v, 2) == 0`:
```text
╰─➤  mix bench map_filter
Operating System: Linux"
CPU Information: AMD Phenom(tm) II X6 1090T Processor
Number of Available Cores: 6
Available memory: 15.67 GB
Elixir 1.8.1
Erlang 21.2.2

warmup: 2 s
time: 2 s
memory time: 2 s
parallel: 1
inputs: large-10000, medium-10, small-1
Estimated total run time: 1.50 min


Benchmarking :maps.filter with input large-10000...
Benchmarking :maps.filter with input medium-10...
Benchmarking :maps.filter with input small-1...
Benchmarking Enum.filter with input large-10000...
Benchmarking Enum.filter with input medium-10...
Benchmarking Enum.filter with input small-1...
Benchmarking Enum.filter|>Enum.into(%{}) with input large-10000...
Benchmarking Enum.filter|>Enum.into(%{}) with input medium-10...
Benchmarking Enum.filter|>Enum.into(%{}) with input small-1...
Benchmarking Enum.filter|>Map.new with input large-10000...
Benchmarking Enum.filter|>Map.new with input medium-10...
Benchmarking Enum.filter|>Map.new with input small-1...
Benchmarking for with input large-10000...
Benchmarking for with input medium-10...
Benchmarking for with input small-1...
```
```elixir
##### With input large-10000 #####
Name                                  ips        average  deviation         median         99th %
Enum.filter                        530.14        1.89 ms    ±10.43%        1.86 ms        2.21 ms
for                                511.08        1.96 ms     ±9.37%        1.92 ms        2.22 ms
:maps.filter                       505.77        1.98 ms     ±8.24%        1.96 ms        2.31 ms
Enum.filter|>Enum.into(%{})        359.96        2.78 ms    ±22.41%        2.71 ms        5.76 ms
Enum.filter|>Map.new               347.07        2.88 ms    ±23.09%        2.76 ms        6.18 ms

Comparison: 
Enum.filter                        530.14
for                                511.08 - 1.04x slower
:maps.filter                       505.77 - 1.05x slower
Enum.filter|>Enum.into(%{})        359.96 - 1.47x slower
Enum.filter|>Map.new               347.07 - 1.53x slower

Memory usage statistics:

Name                           Memory usage
Enum.filter                         1.10 MB
for                                 1.25 MB - 1.14x memory usage
:maps.filter                        0.99 MB - 0.90x memory usage
Enum.filter|>Enum.into(%{})         1.24 MB - 1.13x memory usage
Enum.filter|>Map.new                1.24 MB - 1.13x memory usage

```
```elixir
##### With input medium-10 #####
Name                                  ips        average  deviation         median         99th %
:maps.filter                     797.62 K        1.25 μs  ±2482.42%           1 μs           3 μs
for                              758.76 K        1.32 μs  ±2233.50%           1 μs           3 μs
Enum.filter|>Enum.into(%{})      452.19 K        2.21 μs   ±502.98%           2 μs           3 μs
Enum.filter|>Map.new             443.79 K        2.25 μs  ±1109.90%           2 μs           4 μs
Enum.filter                      437.99 K        2.28 μs  ±1358.04%           2 μs           4 μs

Comparison: 
:maps.filter                     797.62 K
for                              758.76 K - 1.05x slower
Enum.filter|>Enum.into(%{})      452.19 K - 1.76x slower
Enum.filter|>Map.new             443.79 K - 1.80x slower
Enum.filter                      437.99 K - 1.82x slower

Memory usage statistics:

Name                           Memory usage
:maps.filter                        1.23 KB
for                                 1.48 KB - 1.20x memory usage
Enum.filter|>Enum.into(%{})         1.54 KB - 1.25x memory usage
Enum.filter|>Map.new                1.54 KB - 1.25x memory usage
Enum.filter                         1.43 KB - 1.16x memory usage

```
```elixir
##### With input small-1 #####
Name                                  ips        average  deviation         median         99th %
Enum.filter                        7.71 M       0.130 μs ±14668.02%           0 μs           1 μs
:maps.filter                       2.91 M        0.34 μs  ±1101.90%        0.30 μs        0.70 μs
for                                2.18 M        0.46 μs    ±31.71%        0.42 μs        0.80 μs
Enum.filter|>Map.new               1.80 M        0.55 μs    ±30.30%        0.52 μs        0.95 μs
Enum.filter|>Enum.into(%{})        1.78 M        0.56 μs    ±34.62%        0.52 μs        0.96 μs

Comparison: 
Enum.filter                        7.71 M
:maps.filter                       2.91 M - 2.65x slower
for                                2.18 M - 3.54x slower
Enum.filter|>Map.new               1.80 M - 4.28x slower
Enum.filter|>Enum.into(%{})        1.78 M - 4.33x slower

Memory usage statistics:

Name                           Memory usage
Enum.filter                           440 B
:maps.filter                          336 B - 0.76x memory usage
for                                   416 B - 0.95x memory usage
Enum.filter|>Map.new                  472 B - 1.07x memory usage
Enum.filter|>Enum.into(%{})           472 B - 1.07x memory usage

```

whatyouhide:
I don't think we want to s

whatyouhide:
I don't think we want to special case based on map size. We care less about the small cases since the time is so little anyways. Anyways, please let's continue the discussion on the ML :)

whatyouhide:
Lol, mobile typing fart ahaha sorry!

OvermindDL1:
I meant an optimization opportunity for `for`, if it's not able to hit the same speed on small ones than it's probably missing an optimization that `:map.filter` has.  And I wasn't proposing anything, just I like running benchmarks so I put the output.  ^.^

romul:
Hey @whatyouhide,

`for {key, value} <- map, some_filter(key, value), into: %{}, do: {key, value}`

is an alternative, yeah. But it has 2 flaws: 
1) it could not be piped
2) it is too verbose and exposes implementation details like imperative programming does.

> for is basically faster in general, although :maps.filter is faster on small map sizes

@OvermindDL1 I couldn't say that difference in 1% on large inputs is a rationale to say that one option is faster than another, it's just within the measurement error. But your benchmarks clearly show that `for` always uses more memory.

josevalim:
The reason why `Enum.filter` is slower is because it calls `reverse`, which is useless when building a map. `for` knows better and addresses that. The difference in performance between `:maps.filter/2` and `for` is due to their implementation. The former uses the maps iterator, the second convert them to a list. That would explain the higher memory usage since `for` needs to build a large list. We need to consider using iterators instead of maps:to_lists in Elixir's Enumerable implementation but we need to do extensive benchmarks.

michalmuskala:
The iterators are noticeably slower for us. They were optimised for the erlang implementation of the maps module that calls functions with 2 arguments (for key and value). We call them with a 2-tuple. The overhead of the packing and unpacking was pretty significant when I benchmarked after iterators were released.

PR #9293, opened 2019-08-14
Allow running a single file with make test_*

Gazler:
The make test_mix command can now be run with an environment variable to
run the tests for a single file:

    make test_mix -e ELIXIR_TEST_FILE=test/mix/tasks/release_test.exs

Runing `make test_mix` will default to:

    make test_mix -e ELIXIR_TEST_FILE=test/**/*_test.exs


josevalim:
Given it is more typing than the current approach and both have the same issues with discoverability, I propose for us to continue with what we have today. thank you!

PR #9296, opened 2019-08-15
Correct guard in Enum slice optimization

fertapric:
cc @josevalim

### DIFF
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -3327,7 +3327,7 @@ defimpl Enumerable, for: List do
   @doc false
   def slice([], _start, _count, _size), do: []
   def slice(_list, _start, 0, _size), do: []
-  def slice(list, start, count, count), do: list |> drop(start)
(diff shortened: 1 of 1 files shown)


whatyouhide:
Was this a bug or is it just for clarity?

fertapric:
@whatyouhide it's a bug. The `slice` function receives the start position, the amount of elements to slice, and the size of the list. The optimization is never been used because the amount of elements to slice won't be the same as the size unless the start is `0` and I think that's already optimized somewhere else.

For example, `Enum.take([1, 2, 3], -1)` would call `slice([1, 2, 3], 1, 2, 3)`.

josevalim:
It is a fix. The clause as it was would only be true if start was 0. Good catch @fertapric!

whatyouhide:
Should we add a test then?

fertapric:
@whatyouhide it cannot be done easily because it's an optimization. If the optimization is not used, it just falls back to the slower default implementation.

whatyouhide:
Ah that's what I thought, that's why I wasn't feeling it was a bug per se :) Good catch!

PR #9299, opened 2019-08-19
Change List.Chars.to_charlist to return empty list when given nil

J3RN:
Presently, `List.Chars.to_charlist/1` returns `'nil'` when passed `nil`, which feels sort of odd to me. I believe that the canonical character list form of `nil` should be an empty list, similar to how `String.Chars.to_string/1` returns an empty string when passed `nil`.

I added a test around the new behavior, of course, and also around the handling of `true` and `false` since they are also special atoms (though they are handled well by the existing code :smile:).

### DIFF
--- a/lib/elixir/lib/list/chars.ex
+++ b/lib/elixir/lib/list/chars.ex
@@ -24,6 +24,8 @@ defprotocol List.Chars do
 end
 
 defimpl List.Chars, for: Atom do
+  def to_charlist(nil), do: ''
+
   def to_charlist(atom), do: Atom.to_charlist(atom)
 end
 
--- a/lib/elixir/test/elixir/list/chars_test.exs
+++ b/lib/elixir/test/elixir/list/chars_test.exs
@@ -6,6 +6,12 @@ defmodule List.Chars.AtomTest do
   test "basic" do
     assert to_charlist(:foo) == 'foo'
   end
+
+  test "true false nil" do
+    assert to_charlist(false) == 'false'
+    assert to_charlist(true) == 'true'
+    assert to_charlist(nil) == ''
+  end
 end
 
 defmodule List.Chars.BitStringTest do


chulkilee:
This will make `Atom.to_charlist(nil) == ''`, not `'nil'`.

I see the current behavior may surprise from other functions using `Atom.to_char_list/1` behind the scene - but I guess it should be fixed *there*, not here, as here we treat the value as atom here - not any special meaning.

(note: I'm not a core team member :wink:)

J3RN:
@chulkilee Good callout! Oddly, I discovered that this is actually in line with the current behavior with `String`; `Atom.to_string(nil) == "nil"` while `String.Chars.to_string(nil) == ""`. I can't really say whether this is right or wrong, only that this change is consistent. 🤷‍

I'd be happy to change the implementations for `Atom.to_string/1` and `Atom.to_charlist/1` if the maintainers would prefer that. :smile:

chulkilee:
Oh, first of all, `Atom.to_charlist/1` is different from `List.Chars.to_charlist` for atom - so [my previous comment](https://github.com/elixir-lang/elixir/pull/9299#issuecomment-522432092) is incorrect 🤦‍♂ 

Most implementation is just forwarding to their module - e.g. `Integer.to_charlist/1` for `Integer`. Then actual question would be

- Should we change `Atom.to_charlist/1` (which just calls `:erlang.atom_to_list(atom)` [src](https://github.com/elixir-lang/elixir/blob/v1.9.1/lib/elixir/lib/atom.ex#L37)?
- Should we treat it special so that `List.Chars.to_charlist(nil) == []` while `Atom.to_charlist(nil) == 'nil'`?

My two cents: `nil` is atom so just when we take it as an atom, it should be just plain atom, not a special thing.

josevalim:
The Aom.to_charlist function is the base function. The protocol may embed special meaning into the values, especially with the purpose of string/charlist interpolation in mind, which is where the `.Chars` protocols are mostly used.

whatyouhide:
Let's have `List.Chars` mimic the behaviour of `String.Chars` but `Atom.to_charlist/1` be like `Atom.to_string/1` :)

PR #9300, opened 2019-08-19
Use String.Chars.t in Enum.join and Enum.map_join specs

wojtekmach:


wojtekmach:
ugh, I'm dumb, I misread the arguments. What would be more like it would be: `join(t(String.Chars.t()), String.t) :: String.t`. Sorry for noise!

PR #9307, opened 2019-08-21
Add `trace` logger level

esse:
When working with Elixir logger, we actually had an issue with missing `trace` logging level.

As we can see, for example, in log4j, often used is logging level with even more verbosity than `debug` - `trace` [link](https://logging.apache.org/log4j/1.2/apidocs/org/apache/log4j/Level.html#TRACE). It's pretty useful for fine-graning logging information, which would otherwise put a lot of junk into developer console. Adding it alleviates the issue of manually adding, and removing logging statement (which is tedious). 

Together with compile-level removal of logger statements, I think it may be a nice addition to Elixir Logger.

In conclusion: This PR adds `trace` logger level, for logging info too verbose for debug level (and not always wanted even in dev mode)

### DIFF
--- a/lib/logger/lib/logger.ex
+++ b/lib/logger/lib/logger.ex
@@ -48,6 +48,7 @@ defmodule Logger do
 
   The supported levels, ordered by precedence, are:
 
+    * `:trace` - for logging any messages, too verbose for other levels
     * `:debug` - for debug-related messages
     * `:info` - for information of any kind
     * `:warn` - for warnings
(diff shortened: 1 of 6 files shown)


josevalim:
Hi @esse, thanks for the PR. We would like to ask you though to discuss this feature in the Elixir Core mailing list before hand.

In any case, note it is unlikely we will add new log levels and very unlikely that we will add a log level  that does not have an equivalent in Erlang’s logger (which is syslog based IIRC).

esse:
Wow, that was fast!

Thanks for your answer!

PR #9310, opened 2019-08-22
Fix releases not starting as a Windows Service

vais:


josevalim:
Hi @vais! Thanks for the PR. Which error are you seeing? You cannot start it or you can't stop/connect to it? The erlsrv command expects the name as a separate option, can you please try the change I pushed above? Thanks!

vais:
Hi, @josevalim, my bad, sorry for the confusion - my "fix" is totally wrong and only worked by coincidence. It looks like what's really causing the Windows service not to start is the presence of config files in the `_build\prod\rel\database\tmp` directory.

Let's take a step back, here's the problem I'm seeing when I try to start the newly installed service:
```
C:\Windows\System32>C:\Users\Administrator\Desktop\project\database\_build\prod\rel\my_project_name_here\erts-10.3\bin\erlsrv.exe start database_database
C:\Users\Administrator\Desktop\project\database\_build\prod\rel\my_project_name_here\erts-10.3\bin\erlsrv.exe: Failed to start service database_database.
Error: The process terminated unexpectedly.
```
And here's a screenshot of the same error when attempting to start that service via the Windows GUI services applet:

error

I noticed that every time I try to run the `install` command, a `_build\prod\rel\database\tmp` directory is created (if it does not already exist) and a file named something like `database-0.1.0-20190823082-1488.runtime.config` is added. If I remove or rename this file or the entire `tmp` directory, then the Windows service starts up just fine. This `tmp` directory is not recreated when the service starts. It looks like it's only created when the service is installed.

That's all I got so far...

josevalim:
Perfect. I know what is causing this issue. Meanwhile can you please try my patch and let me know if it currently works too?

josevalim:
To be clear: let me know if my patch causes triggers the same problem as above or a different one.

josevalim:
Actually, I don't know what is causing the issue. :D I thought it was because we removed the config files but we don't do that by default. Please try changing the file in this PR by adding `-debugtype reuse` before `-args`. You will now find a `.log` file in the directory of the service and hopefully that log file will tell us what is going wrong. :)

vais:
> Perfect. I know what is causing this issue.

@josevalim Great! Looking forward to seeing the fix then 😄 

> Meanwhile can you please try my patch and let me know if it currently works too?
> To be clear: let me know if my patch causes triggers the same problem as above or a different one.

I am seeing the same behavior before and after your patch: with `tmp` folder present the service does not start, without the `tmp` folder the service starts up fine.

But I think your patch is a good idea regardless - it should be `-name` or `-sname` depending on the desired distribution, i.e. having `-name` be hard-coded (as before your patch) is a bug of its own.

Thanks!

vais:
Oh, just saw your latest comment after posting mine 🤦‍♂ 
Let me see...

vais:
@josevalim here you go:
```
C:\Windows\System32>type database_database.debug
{"init terminating in do_boot",{#{'__exception__'=>true,'__struct__'=>'Elixir.ArgumentError',message=><<"could not fetch environment variable \"RELEASE_SYS_CONFIG\" because it is not set">>},[{'Elixir.System','fetch_env!',1,[{file,"lib/system.ex"},{line,518}]},{'Elixir.Config.Provider','resolve_config_path!',1,[{file,"lib/config/provider.ex"},{line,134}]},{'Elixir.Config.Provider',boot,3,[{file,"lib/config/provider.ex"},{line,157}]},{init,eval_script,2,[{file,"init.erl"},{line,941}]},{init,do_boot,3,[{file,"init.erl"},{line,817}]}]}}
init terminating in do_boot ({,[{Elixir.System,fetch_env!,1,[{_},{_}]},{Elixir.Config.Provider,resolve_config_path!,1,[{_},{_}]},{Elixir.Config.Provider,boot,3,[{_},{_}]},{init,eval_script,2,[{_},{_}]

Crash dump is being written to: erl_crash.dump...done
```

josevalim:
@vais I see! Please try adding this now:

    -args "-env RELEASE_SYS_CONFIG !RELEASE_SYS_CONFIG! ...`

And let me know if it works!

josevalim:
Adding:

    -args "-env RELEASE_SYS_CONFIG !RELEASE_SYS_CONFIG! -env RELEASE_ROOT !RELEASE_ROOT! ...

Worked here! I am pushing a fix soon!

josevalim:
*We* did it! Thank you for the report and helping sanity check everything!

PR #9314, opened 2019-08-23
Warn on trailing commas in word list

gamache:
When creating word lists (e.g., `~w(foo bar baz)`), it is a common mistake to insert trailing commas for some or all parts of the list (e.g., `~w(foo, bar, baz)`).  This PR introduces a warning when such trailing commas are encountered, in macros and regular code:

```
iex> ~w(hello, world)
warning: item "hello," in word list has a trailing comma; was this intentional?
  iex:1: (file)

["hello,", "world"]
```

### DIFF
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -5198,17 +5198,15 @@ defmodule Kernel do
       true ->
         parts = String.split(string)
 
-        :lists.foreach(
-          fn part ->
-            if :binary.last(part) == ?, do
-              IO.warn(
-                "item \"#{part}\" in word list has a trailing comma; was this intentional?",
-                stacktrace
-              )
-            end
-          end,
-          parts
-        )
+        parts_with_trailing_comma = :lists.filter(&(:binary.last(&1) == ?,), parts)
+
+        if parts_with_trailing_comma != [] do
+          IO.warn(
+            "The sigils ~w/~W do not allow trailing commas at the end of each word. " <>
+              "If the comma is necessary, define a regular list with [...], otherwise remove the comma.",
+            stacktrace
+          )
+        end
 
         case mod do
           ?s -> parts
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -5201,7 +5201,7 @@ defmodule Kernel do
           stacktrace = Macro.Env.stacktrace(caller)
 
           IO.warn(
-            "The sigils ~w/~W do not allow trailing commas at the end of each word. " <>
+            "the sigils ~w/~W do not allow trailing commas at the end of each word. " <>
               "If the comma is necessary, define a regular list with [...], otherwise remove the comma.",
             stacktrace
           )
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -5153,14 +5153,12 @@ defmodule Kernel do
   defmacro sigil_w(term, modifiers)
 
   defmacro sigil_w({:<<>>, _meta, [string]}, modifiers) when is_binary(string) do
-    stacktrace = Macro.Env.stacktrace(__CALLER__)
-    split_words(:elixir_interpolation.unescape_chars(string), modifiers, stacktrace)
+    split_words(:elixir_interpolation.unescape_chars(string), modifiers, __CALLER__)
   end
 
   defmacro sigil_w({:<<>>, meta, pieces}, modifiers) do
-    stacktrace = Macro.Env.stacktrace(__CALLER__)
     binary = {:<<>>, meta, unescape_tokens(pieces)}
-    split_words(binary, modifiers, stacktrace)
+    split_words(binary, modifiers, __CALLER__)
   end
 
   @doc ~S"""
@@ -5184,15 +5182,14 @@ defmodule Kernel do
   defmacro sigil_W(term, modifiers)
 
   defmacro sigil_W({:<<>>, _meta, [string]}, modifiers) when is_binary(string) do
-    stacktrace = Macro.Env.stacktrace(__CALLER__)
-    split_words(string, modifiers, stacktrace)
+    split_words(string, modifiers, __CALLER__)
   end
 
-  defp split_words(string, [], stacktrace) do
-    split_words(string, [?s], stacktrace)
+  defp split_words(string, [], caller) do
+    split_words(string, [?s], caller)
   end
 
-  defp split_words(string, [mod], stacktrace)
+  defp split_words(string, [mod], caller)
        when mod == ?s or mod == ?a or mod == ?c do
     case is_binary(string) do
       true ->
@@ -5201,6 +5198,8 @@ defmodule Kernel do
         parts_with_trailing_comma = :lists.filter(&(:binary.last(&1) == ?,), parts)
 
         if parts_with_trailing_comma != [] do
+          stacktrace = Macro.Env.stacktrace(caller)
+
           IO.warn(
             "The sigils ~w/~W do not allow trailing commas at the end of each word. " <>
               "If the comma is necessary, define a regular list with [...], otherwise remove the comma.",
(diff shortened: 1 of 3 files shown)


josevalim:
Thanks @gamache! I think we should probably avoid from doing this check at runtime so we don't slow down code written today (which could in itself be a regression). Since it is only a warning, it is fine that we don't always show it. What do you think?

gamache:
@josevalim I like that approach. I've pushed a commit that implements it.

whatyouhide:
@gamache but what if it *was* intentional? I'm worried about valid use cases for this. 🤔

josevalim:
@whatyouhide we need to improve the warning. In my mind, if you need commas in your word lists, then you would use the regular list syntax.

lib/elixir/lib/kernel.ex
```diff
@@ -5182,27 +5184,43 @@ defmodule Kernel do
...
        when mod == ?s or mod == ?a or mod == ?c do
     case is_binary(string) do
       true ->
         parts = String.split(string)
 
+        :lists.foreach(
+          fn part ->
+            if :binary.last(part) == ?, do
+              IO.warn(
+                "item \"#{part}\" in word list has a trailing comma; was this intentional?",
```
josevalim: Suggestion:

> The sigil w does not allow trailing commas at the end of each word. \
> If this was an accident, please remove the comma, otherwise use a regular list.

gamache:
@whatyouhide There is always plain syntax (`["foo,", "bar"]`) for those who need trailing commas and don't care for warnings. Given how common the error is -- I've personally done it at least ten times over the last twenty years -- I think this would do more good than harm.

lib/elixir/lib/kernel.ex
gamache: This sounds more like a soft error than a warning, but you're in a position to make that call, so I'm good with it. I'll update the PR when I am near a keyboard.

josevalim: Given that a warning:

1. cannot removed
2. it fails the build on --warning-as-errors

I would say all warnings are "soft errors". It is not the error message that is going to change it, as we don't expect developers to be compiling code with warnings.

gamache:
@josevalim I've made the wording change and ensured users only see it once per list.

lib/elixir/test/elixir/kernel/sigils_test.exs
```diff
@@ -121,6 +187,124 @@ bar) in ["foo\\\nbar", "foo\\\r\nbar"]
     assert ~W(Foo.Bar.Baz)a == [:"Foo.Bar.Baz"]
   end
 
+  test "sigil w/W warns on trailing comma at compile time, not runtime" do
```
fertapric: I think this could be simplified to:

```elixir
@compile {:no_warn_undefined, Kernel.SigilsTest.TrailingComma}

test "sigil w/W warns on trailing comma at compile time, not runtime" do
  for sigil <- ~w(w W),
      modifier <- ~w(a s c) do
    assert capture_err(fn ->
             Code.compile_string("""
             defmodule Kernel.SigilsTest.TrailingComma do
               def run, do: ~#{sigil}(foo, bar baz)#{modifier}
             end
             """)
           end) =~
             "The sigils ~w/~W do not allow trailing commas"

    assert capture_err(fn -> Kernel.SigilsTest.TrailingComma.run() end) == ""

    purge(Kernel.SigilsTest.TrailingComma)
  end
end

defp purge(module) do
  :code.purge(module)
  :code.delete(module)
end
```

(with the code above, all those `Kernel.SigilsTest.Runtime_*` modules would not be required)

lib/elixir/test/elixir/kernel/sigils_test.exs
```diff
@@ -121,6 +187,124 @@ bar) in ["foo\\\nbar", "foo\\\r\nbar"]
...
+               def run, do: ~w(foo, bar baz)s
+             end
+             """)
+           end) =~
+             "The sigils ~w/~W do not allow trailing commas"
+
+    assert capture_err(fn -> Kernel.SigilsTest.Runtime_Wc.run() end) == ""
+  end
+
+  test "sigil w/W warns on trailing comma inside macro" do
```
fertapric: I think there are no special conditions that would make this feature to behave differently if it's invoked inside a macro :)

I would remove this test, what do you think?

gamache: There are no special conditions _today_. :)  I can remove it, though, your call.

lib/elixir/test/elixir/kernel/sigils_test.exs
gamache: Looks good to me, I'll roll this in.

lib/elixir/test/elixir/kernel/sigils_test.exs
```diff
@@ -121,6 +187,31 @@ bar) in ["foo\\\nbar", "foo\\\r\nbar"]
...
 
+  @compile {:no_warn_undefined, Kernel.SigilsTest.TrailingComma}
+
+  test "sigil w/W warns on trailing comma at compile time, not runtime" do
+    for sigil <- ~w(w W),
+        modifier <- ~w(a s c) do
+      assert capture_err(fn ->
+               Code.compile_string("""
+               defmodule Kernel.SigilsTest.TrailingComma do
+                 def run, do: ~#{sigil}(foo, bar baz)#{modifier}
```
josevalim: We don't need a module for these... we can just call `Code.eval`.

lib/elixir/test/elixir/kernel/sigils_test.exs
```diff
@@ -1,8 +1,74 @@
 Code.require_file("../test_helper.exs", __DIR__)
 
+defmodule Kernel.SigilsTest.Macros do
```
fertapric: These modules are not needed now

lib/elixir/test/elixir/kernel/sigils_test.exs
```diff
@@ -121,6 +187,31 @@ bar) in ["foo\\\nbar", "foo\\\r\nbar"]
     assert ~W(Foo.Bar.Baz)a == [:"Foo.Bar.Baz"]
   end
 
+  @compile {:no_warn_undefined, Kernel.SigilsTest.TrailingComma}
+
+  test "sigil w/W warns on trailing comma at compile time, not runtime" do
+    for sigil <- ~w(w W),
+        modifier <- ~w(a s c) do
+      assert capture_err(fn ->
```
josevalim: If we use capture_err the test cannot be `async`. My suggestion is to move it to kernel/warnings_test. We also have conveniences there to make writing the test easy.

lib/elixir/test/elixir/kernel/sigils_test.exs
gamache: I don't understand -- how can I separately test compile time and runtime using `Code.eval`?

josevalim: The warning comes from Macro expansion and both compilation and veal have to expand macros.

gamache: Yes, but several lines down, I test whether the warning is *not* seen at runtime, and I don't know how I would do that if I used `Code.eval` above.

josevalim: You could call `Macro.expand` and pass it the sigil AST to check it is emitted during expansion. In this case, no eval necessary.

lib/elixir/lib/kernel.ex
```diff
@@ -5182,19 +5182,31 @@ defmodule Kernel do
...
       true ->
         parts = String.split(string)
 
+        parts_with_trailing_comma = :lists.filter(&(:binary.last(&1) == ?,), parts)
+
+        if parts_with_trailing_comma != [] do
+          stacktrace = Macro.Env.stacktrace(caller)
+
+          IO.warn(
+            "The sigils ~w/~W do not allow trailing commas at the end of each word. " <>
```
```suggestion
            "the sigils ~w/~W do not allow trailing commas at the end of each word. " <>
```

I believe our warning messages start in lowercase. :)

gamache: The phrasing was yours! I'll lowercase it :)

josevalim:

PR #9316, opened 2019-08-25
Add xor to list of operators in Reserved Words section

adsouza:


josevalim:
`xor` is not an operator in Elixir. :) Maybe something else gave the impression that xor is a valid operator and that should be fixed instead?

adsouza:
Indeed! Sorry about the bother.

eksperimental:
I noticed this a while ago and I found that weird. Why xor is not an operator in Elixir when it is one in Erlang?

PR #9317, opened 2019-08-26
Improve behaviour documentation

adamu:
I was doing some work with behaviours and found the documentation about how to inspect a behaviour a little lacking (resulting in me asking [this SO question](https://stackoverflow.com/questions/57653216/elixir-testing-ensure-a-module-defines-a-callback)), so I've added some information to the Typespec docs, and also linked there from the behaviour section of `Module` docs.

* Add a link from `Module` to the relevant section of the Typespec docs.
* Add an `Inspecting behviours` section to the Typespec docs that
  documents how to inspect the behaviours. This information is so far
  only available in the deprecated `Behaviour` module or the Erlang docs.

### DIFF
--- a/lib/elixir/lib/module.ex
+++ b/lib/elixir/lib/module.ex
@@ -56,6 +56,9 @@ defmodule Module do
   If the behaviour changes or `URI.HTTP` does not implement
   one of the callbacks, a warning will be raised.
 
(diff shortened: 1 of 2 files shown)


whatyouhide:
Thank you @adamu! 💟

PR #9330, opened 2019-09-06
Registration of ErlangHandler spec mismatch

hauleth:
The Dialyzer was issuing a warining on `logger:add_handler/3` call as
the key `sasl_reports?` should not be in root map but rather within
`config` key of the map.


fertapric:
Thanks @hauleth! 💜

hauleth:
Funny thing is that #9333 will remove that ;)

PR #9332, opened 2019-09-08
Ensure project's Logger config is used when running Mix tasks

fertapric:
Part of the configuration of the Logger can be changed dynamically but other part is only used at start. For example, the `:console` backend is implemented with `gen_event`, and its configuration is only read on `init` to build the state.

Logger is first started by Elixir using its default configuration, but the project's configuration is not used fully until the Logger application is restarted. This usually happens when the application is started, but not all Mix tasks do such thing.

This commit ensures that Logger is restarted before calling any Mix task.

### DIFF
--- a/lib/mix/lib/mix/cli.ex
+++ b/lib/mix/lib/mix/cli.ex
@@ -76,6 +76,8 @@ defmodule Mix.CLI do
     try do
       ensure_no_slashes(name)
       Mix.Task.run("loadconfig")
+      # Restart Logger so it uses the configuration of the project.
+      restart_logger()
       Mix.Task.run(name, args)
     rescue
       # We only rescue exceptions in the Mix namespace, all
@@ -88,6 +90,13 @@ defmodule Mix.CLI do
         else
           reraise exception, __STACKTRACE__
         end
+    end
+  end
+
+  defp restart_logger do
+    if Process.whereis(Logger) do
+      Logger.App.stop()
+      :ok = Logger.App.start()
     end
   end
 
--- a/lib/mix/lib/mix/cli.ex
+++ b/lib/mix/lib/mix/cli.ex
@@ -96,7 +96,7 @@ defmodule Mix.CLI do
   end
 
   defp logger_configured? do
-    Enum.member?(Mix.ProjectStack.config_apps(), :logger)
+    :logger in Mix.ProjectStack.config_apps()
   end
 
   defp restart_logger do
--- a/lib/mix/lib/mix/cli.ex
+++ b/lib/mix/lib/mix/cli.ex
@@ -1,6 +1,8 @@
 defmodule Mix.CLI do
   @moduledoc false
 
+  @compile {:no_warn_undefined, Logger.App}
+
   @doc """
   Runs Mix according to the command line arguments.
   """
@@ -77,7 +79,7 @@ defmodule Mix.CLI do
       ensure_no_slashes(name)
       Mix.Task.run("loadconfig")
       # Restart Logger so it uses the configuration of the project.
-      restart_logger()
+      if logger_configured?(), do: restart_logger()
       Mix.Task.run(name, args)
     rescue
       # We only rescue exceptions in the Mix namespace, all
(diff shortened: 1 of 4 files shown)


lib/mix/lib/mix/tasks/app.start.ex
```diff
@@ -69,23 +69,12 @@ defmodule Mix.Tasks.App.Start do
...
-    logger = Process.whereis(Logger)
-
-    if logger do
-      Logger.App.stop()
-    end
+    unless "--no-start" in args do
+      # Stop Logger if the application does not depend on it.
+      #
+      # Mix should not depend directly on Logger, that's why we first check if it's loaded.
+      if not logger_configured?() && Process.whereis(Logger), do: Logger.App.stop()
```
josevalim: Instead of relying on the configured check, we likely want to stop it whenever it isn't a child... I guess we would need to check if logger is started while we are traversing start_apps.

fertapric: >  I guess we would need to check if logger is started while we are traversing start_apps.

@josevalim if you are referring to `ensure_all_started` it relies on `Application.start/2` telling which dependent apps are not started:

https://github.com/elixir-lang/elixir/blob/d894408c91b7962a58b423b0f2c7084d1ab0f027/lib/mix/lib/mix/tasks/app.start.ex#L121-L131

Since `:logger` is already started the check cannot be done there. To know if `:logger` is not a child, we would need to stop it first.

What's that what you had in mind? 🙂

josevalim: We will need to bring the implementation of `start` over then so we can check if `logger` iss a dependency as we traverse.

fertapric: The implementation of `start` delegates to Erlang's `application` which involves `application_controller` and what not 😅 [The "started" check depends on the internal state of the `application_controller`](https://github.com/erlang/otp/blob/master/lib/kernel/src/application_controller.erl#L647-L649). It seems like a lot of work to detect if `:logger` is a dependency.

I think we could traverse the applications using `Application.spec` and check if `:logger` is listed on those. What do you think?

josevalim: Calling `spec` sounds good to me!

lib/mix/lib/mix/cli.ex
```diff
@@ -76,6 +78,8 @@ defmodule Mix.CLI do
     try do
       ensure_no_slashes(name)
       Mix.Task.run("loadconfig")
+      # Restart Logger so it uses the configuration of the project.
+      if logger_configured?(), do: restart_logger()
```
fertapric: I thought it would be nice to only restart the logger if it's configured by the application :)

josevalim: I like it.

lib/mix/lib/mix/tasks/app.start.ex
```diff
@@ -190,6 +180,32 @@ defmodule Mix.Tasks.App.Start do
...
+  defp logger_dependency?([], _checked_apps) do
+    false
+  end
+
+  defp logger_dependency?([app | apps], checked_apps) do
+    deps =
+      List.wrap(Application.spec(app, :applications)) ++
+        List.wrap(Application.spec(app, :included_applications))
+
+    case deps do
```
josevalim: A cond would probably work better here.

lib/mix/lib/mix/tasks/app.start.ex
```diff
@@ -190,6 +180,32 @@ defmodule Mix.Tasks.App.Start do
...
+  defp logger_dependency?(apps), do: logger_dependency?(apps, [])
+
+  defp logger_dependency?([], _checked_apps) do
+    false
+  end
+
+  defp logger_dependency?([app | apps], checked_apps) do
+    deps =
+      List.wrap(Application.spec(app, :applications)) ++
+        List.wrap(Application.spec(app, :included_applications))
```
josevalim: roflsscale: Swap the order to have the smaller list on the front. :)

lib/mix/lib/mix/cli.ex
fertapric: Unfortunately, I think this won't work if the logger is configured using the `:env` key in `mix.exs` :(

fertapric: Oh! Nevermind. `:env` is scoped per app 😬

fertapric:
Feedback applied!

fertapric:
@josevalim is this good to go? 🙂

PR #9333, opened 2019-09-08
Use Erlang's logger as main logging implementation

hauleth:
This implements Elixir's `Logger` module on top of Erlang's `logger` while providing compatibility layer.

- [x] use `logger:log/2-3` to dispatch log messages
- [x] translate Erlang messages to Elixir's logger compatible ones
- [x] use Erlang functions instead of custom Elixir ones for utilities
- [x] ~~use Erlang formatter for formatting messages instead of handler~~
- [x] use Erlang filters instead of handler


lib/logger/lib/logger.ex
```diff
@@ -459,41 +459,44 @@ defmodule Logger do
...
   Reads the current process metadata.
   """
   @spec metadata() :: metadata
   def metadata() do
-    __metadata__() |> elem(1)
+    case :logger.get_process_metadata() do
+      :undefined -> []
+      map when is_map(map) ->
+        map
+        |> Map.drop([:pid, :gl, :time, :mfa, :file, :line, :domain, :report_cb])
```
josevalim: Are you sure those are returned from the process metadata?

hauleth: No, this probably should not filter them out.

lib/logger/lib/logger.ex
```diff
@@ -711,29 +714,43 @@ defmodule Logger do
...
-          :exit, reason -> {:error, reason}
-        end
+  def __do_log__({level, _mode, _config, _pdict}, fun, metadata)
+      when is_list(metadata) and is_function(fun, 0) do
+    level =
+      case level do
+        :warn -> :warning
+        level -> level
+      end
+    metadata = Map.new(metadata)
```
josevalim: Those conversions would be expensive. I would try to keep them as maps as much as possible. Only the ones passed at runtime as keyword lists would need to be converted to maps. I would even consider first changing the backends to accept to receive map metadata... I am now sure how much it would be considered a breaking change though.

Another idea is to pass the Logger metadata as a key inside the Erlang metadata... that would avoid key conflicts and what not and allow us to keep the preivous data type.

lib/logger/lib/logger/erlang_handler.ex
```diff
@@ -65,7 +105,7 @@ defmodule Logger.ErlangHandler do
 
     metadata =
       case map do
-        %{file: file, line: line} -> [file: List.to_string(file), line: line] ++ metadata
+        %{file: file, line: line} -> [file: to_string(file), line: line] ++ metadata
```
josevalim: Is the change of the `file` type a backwards incompatible change?

hauleth: It is compatible. Just right now there are 2 possibilities:

- when log comes from the Erlang macros - the `file` will be charlist
- when log comes from the Elixir macros - the `file` will be binary

That is why there is `Kernel.to_string/1` used instead of `List.to_string/1`.

lib/logger/lib/logger.ex
hauleth: This is only temporary, as I am still working on porting rest of the API to use maps within API and change map to keyword list only in compatibility layer just before passing it further to the backends.

lib/logger/lib/logger.ex
```diff
@@ -439,16 +439,12 @@ defmodule Logger do
 
   @type backend :: :gen_event.handler()
   @type message :: IO.chardata() | String.Chars.t()
-  @type level :: :error | :info | :warn | :debug
+  @type level :: :logger.level() | :warn
   @type metadata :: keyword()
-  @levels [:error, :info, :warn, :debug]
+  @levels [:emergency, :alert, :critical, :error, :warning, :notice, :info, :debug]
```
josevalim: Just a tip: this PR will be easier to merge if we don't add new features. What I would do, if it works out, is to merge the integration and only one or two releases later add new features. Otherwise, if we add new features tied to `:logger` and then it turns out the integration doesn't work, we would be in a very hard spot.

hauleth: Ok, I haven't count that as a "new feature" as it do not add any new functions (except `Logger.warning/{1,2}`) . It only allows `Logger.log/{2,3}` and `Logger.bare_log/{2,3}` to use new logger levels. But I can roll this back or extract to separate PR.

josevalim: Well, we can call it an enhancement then but the above still applies. So we should roll this back for now. :) Tks.

hauleth:
@josevalim also I believe that the `handle_sasl_reports` option should be deprecated and always set to `true`, because according to docs:

> The new default behaviour is that the SASL application no longer affects which log events that are logged. Supervisor reports and crash reports are logged via the default logger handler which is setup by Kernel. Progress reports are by default not logged, but can be enabled by setting the primary log level to `info`, for example by using the Kernel configuration parameter `logger_level`.

josevalim:
Sounds good, but not in this PR please :) we can do it before or after!

hauleth:
I am leaving comments in all places that need deprecations and/or changes in future.

Only "new features" are:

- Support for `:warning` log level
- `Logger.warning` macro that is currently synonim of `Logger.warn`.

hauleth:
Questions needed resolving:

1. Should `handle_sasl_reports` be enabled globally for all `logger` handlers or rather it should be applied only to the `:logger` legacy handler? Currently it is later, so other handlers will receive everything.
2. Should `Logger.disable`/`Logger.enable` be local to the legacy handler or rather global? Currently it is later, so if someone opt-out of logging, then all logs will be disabled.
3. Should we name our handler `:default` or rather `Logger`? Currently it is later, as I think it should be considered temporary with target to dump legacy handler (and probably whole application) in Elixir 2.0 (if it gonna happen).
4. Should we implement support for `:logger.olp_config/0` format of options for handling overload with compatibility layer or rather we should keep old API as is?

josevalim:
> Should handle_sasl_reports be enabled globally for all logger handlers or rather it should be applied only to the :logger legacy handler? Currently it is later, so other handlers will receive everything.

Agreed.

> Should Logger.disable/Logger.enable be local to the legacy handler or rather global? Currently it is later, so if someone opt-out of logging, then all logs will be disabled.

Agreed.

> Should we name our handler :default or rather Logger? Currently it is later, as I think it should be considered temporary with target to dump legacy handler (and probably whole application) in Elixir 2.0 (if it gonna happen).

Agreed. I am not sure if it will ever be removed, it would be a major backwards incompatible change and we don't want to use major versions as an excuse to break user code. Plus, Elixir's handler is still a convenient way of having safe built-in handler with overload protection. AFAIK, Erlang doesn't expose its mechanisms for overload protection.

> Should we implement support for :logger.olp_config/0 format of options for handling overload with compatibility layer or rather we should keep old API as is?

No because of the above. AFAIK Erlang doesn't expose the building blocks for building olp config, so I don't want to be tied to an API we have no control and that we can't rely on.

hauleth:
I have removed all usage of `persistent_term` in `Logger.Config` as it is now stored as configuration options in `Logger.LegacyHandler` (which is then handled by Erlang's `logger` with hope that it uses the best store available). This greatly reduces complexity of `Logger.Config`.

Improvements to be made:

- Register `Logger.Config` handler within `adding_handler/1` callback and deregister it in `removing_handler/1` callback.
- If that happens review the need for `Logger.Watcher` server
- How to handle failures within `Logger.LegacyHandler`? If there happen any error within handler it will be deregistered, we do not want that, as this will result with no logger at all and can cause missing logs entries. Currently it is done by handling all errors within handler and `IO.inspect`ing them when they happen (very not ideal solution).

lib/logger/lib/logger.ex
```diff
@@ -691,49 +700,54 @@ defmodule Logger do
...
+    __do_log__(:warning, chardata_or_fun, metadata)
+  end
 
+  def __do_log__(level, fun, metadata) when is_function(fun, 0) and is_map(metadata) do
+    case fun.() do
       :skip ->
         :ok
+
+      {msg, meta} ->
+        :logger.log(level, msg, Map.merge(metadata, Map.new(meta)))
```
josevalim:
```suggestion
        :logger.log(level, msg, Enum.into(meta, metadata))
```

It should be the most efficient unless meta is really large.

lib/logger/lib/logger.ex
```diff
@@ -691,49 +700,54 @@ defmodule Logger do
...
+
+      msg when is_binary(msg) ->
+        :logger.log(level, msg, metadata)
     end
   end
 
+  # TODO: Remove that in Elixir 2.0
+  def __do_log__(level, atom, metadata) when is_atom(atom) and is_map(metadata) do
```
josevalim: Do we allow passing atoms?! Since when?

hauleth: At least since a8b2cb99ee

lib/logger/lib/logger/filter.ex
```diff
@@ -0,0 +1,30 @@
+defmodule Logger.Filter do
+  @moduledoc false
+
+  @doc """
+  Filter out SASL reports if `sasl_enabled?` is `true`
+  """
+  def sasl(sasl_enabled?) do
```
josevalim: We should probably only add this filter if sasl_enabled? is `false`.

lib/logger/lib/logger/filter.ex
```diff
@@ -0,0 +1,30 @@
...
+
+    {filter, sasl_enabled?}
+  end
+
+  @doc """
+  Filter out logs if current process opted out of log reports
+  """
+  def process_disabled() do
+    {fn log, _extra ->
+      if Logger.enabled?(self()) do
```
josevalim: Is there a difference between adding the filters here or simply doing this check by the time we log? Otherwise, it would be nice to NOT use filters for our own stuff (and simlpy check by the time of logging) and completely leave filters for the users to manage.

lib/logger/lib/logger.ex
```diff
@@ -515,10 +513,22 @@ defmodule Logger do
...
     :ok
   end
 
+  @doc """
+  Returns whether the logging is enabled for given process.
+
+  Currently the only accepted PID is `self()`.
+  """
+  def enabled?(pid) when pid == self() do
+    case Process.get(@metadata) do
```
josevalim: Probably `Process.get(@metadata, true)` would be enough (without the `case`).

lib/logger/lib/logger/filter.ex
hauleth: No there is not, except the fact that `process_disabled` filter is registered globally. `sasl` filter can be done within handler. The main difference is that if someone defines their own formatter then it will **not** run if any filter returns `:stop`. So there can be some problems when someone will expect formatter to not run on `sasl` entries.

lib/logger/lib/logger.ex
hauleth: Oh, I forgot that `Process.metadata` supports `default` argument.

lib/logger/lib/logger.ex
josevalim: So we can't change this. Instead of checking for atoms, call `to_string` as it was before. Sorry.

lib/logger/lib/logger.ex
```diff
@@ -813,37 +844,50 @@ defmodule Logger do
...
+
+        {Map.new(metadata), quote(do: %{unquote_splicing(quoted)})}
       else
-        {[],
+        quoted = Keyword.update!(caller, :mfa, &Macro.escape/1)
+
+        {%{},
          quote do
-           Keyword.merge(unquote(caller), unquote(metadata))
+           Map.merge(%{unquote_splicing(quoted)}, Map.new(unquote(metadata)))
```
josevalim: Use `Enum.into` here too.

lib/logger/lib/logger.ex
```diff
@@ -813,37 +844,50 @@ defmodule Logger do
...
          end}
       end
 
     compile_level = if is_atom(level), do: level, else: :error
 
-    if compile_time_purge_matching?(compile_level, compile_metadata) do
+    legacy = add_legacy_fields(compile_metadata, module, fun, arity)
```
josevalim: Instead of adding them here... could we add them on the handler? We need to extract those in the handler anyway for erlang compat, don't we?

lib/logger/lib/logger.ex
```diff
@@ -861,7 +905,8 @@ defmodule Logger do
           Logger.Config.compare_levels(level, min_level) == :lt
 
         {k, v} when is_atom(k) ->
-          Keyword.fetch(compile_metadata, k) == {:ok, v}
+          # TODO: Warn on matching on `:module` and `:function` fields
```
josevalim: My suggestion: keep module and function working but as a "metafilter" that looks inside the MFA. We could do the same in the formatter. So we only set mfa as metadata but our APIs know how to look inside the MFA to extract it out. WDYT?

lib/logger/lib/logger.ex
hauleth: No, as this is needed during compilation time for purging. Handler adds them again, but on the runtime.

lib/logger/lib/logger.ex
hauleth: Can be done. I just reintroduced the values just for this as it was simpler, but I can define it as a "metafilter" instead. It should probably be even better as it will not compute legacy fields if not needed.

lib/logger/lib/logger/legacy_handler.ex
```diff
@@ -0,0 +1,180 @@
...
+    }
+  end
+
+  @doc """
+  Hook required by `:logger`.
+  """
+  def log(%{level: erl_level, msg: msg, meta: erl_meta}, config) do
+    level = erlang_level_to_elixir_level(erl_level)
+
+    case Logger.Config.log_data(level) do
```
josevalim: So it seems we are checking this twice now: once when we call `Logger.log` and again here. We should probably have a metadata that checks if it came from `Logger` and avoid doing those checks again (if it came from Logger).

hauleth: Yes, I have fixed that locally. I need to split it into logical parts and I will commit in a second.

lib/logger/lib/logger/filter.ex
josevalim: @hauleth so let's move the sasl check inside, since that's also how it works now (sasl is always checked inside). :+1: for leaving process_disabled out, great idea.

lib/logger/lib/logger.ex
josevalim: i did a comment on how we can make compilation time purging work in this case. If we want to avoid passing module+function, we can treat them as special keys in the formatter and in the compile time purging.

josevalim: Oops, we are talking about this below. I will resoslve this one.

lib/logger/lib/logger.ex
josevalim: @hauleth exactly. I prefer the metafilter so we can start slowly unify on Erlang metadata.

lib/logger/lib/logger.ex
```diff
@@ -581,7 +591,6 @@ defmodule Logger do
   """
   @spec flush :: :ok
   def flush do
-    _ = Process.whereis(:error_logger) && :gen_event.which_handlers(:error_logger)
```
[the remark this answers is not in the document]
hauleth: How that ensures that? I can revert it, I am just curious.

lib/logger/lib/logger/filter.ex
josevalim: Sounds good! Maybe we can even add upstream!

lib/logger/lib/logger.ex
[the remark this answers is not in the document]
josevalim: Which_handlers will block until all messages are processed.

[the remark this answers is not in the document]
hauleth: Oh, ok. So I will roll it back.

lib/logger/lib/logger/filter.ex
hauleth: @josevalim about `sasl` - should it be possible to change this option in the runtime? With `logger` it is not much of the problem.

josevalim: @hauleth if the goal is to remove the feature, than I wouldn't worry about improving it. :) I would just make it a filter that is added (or removed if false) on logger start.

hauleth: @josevalim after reviewing we should not remove that feature, as the default filtering of SASL reports depends on logging level (which is `notice` by default) as SASL messages are on `debug` level. So we need filtering anyway, but I can handle it within handler.

Also I have send proposal to add per-process log level on Erlang mailing list and I had already prepared commit with that feature - https://github.com/hauleth/otp/commit/77246b9323e26a74794153d36933773de0b83a6d

hauleth:
I cannot find the reason why OTP 21.0 and OTP 21.1 fail when trying to deal with ETS tables. It seems like the table is deleted and I cannot find why.

hauleth:
@josevalim it is mostly done. I it still do not use formatter and filters to manage data, but I think that these can be done in future PRs to not prolong it too much.

lib/logger/lib/logger/filter.ex
hauleth: erlang/otp#2384 is stalled, so we need to handle it on our own for now.

lib/logger/lib/logger.ex
```diff
@@ -463,41 +441,39 @@ defmodule Logger do
...
-  end
+    case :logger.get_process_metadata() do
+      :undefined ->
+        :ok = :logger.set_process_metadata(filter_out_nils(keyword))
 
-  defp into_metadata([], metadata), do: metadata
-  defp into_metadata(keyword, metadata), do: into_metadata(keyword, [], metadata)
+      map when is_map(map) ->
+        merged = Enum.into(keyword, map)
+        metadata = filter_out_nils(merged)
```
josevalim: We can probably do the two operations above (merge and filter out) in one pass. So please either do it in one pass OR leave a TODO so we tackle it later. :)

lib/logger/lib/logger.ex
```diff
@@ -463,41 +441,39 @@ defmodule Logger do
...
 
   @doc """
   Resets the current process metadata to the given keyword list.
   """
   @spec reset_metadata(metadata) :: :ok
   def reset_metadata(keywords \\ []) do
-    {enabled?, _metadata} = __metadata__()
-    Process.put(@metadata, {enabled?, []})
+    :ok = :logger.set_process_metadata(%{})
     metadata(keywords)
```
josevalim: This function could be more efficiently implemented as:

    :ok = :logger.set_process_metadata(filter_out_nils(keyword))

lib/logger/lib/logger.ex
```diff
@@ -508,7 +484,7 @@ defmodule Logger do
   """
   @spec enable(pid) :: :ok
   def enable(pid) when pid == self() do
-    Process.put(@metadata, {true, metadata()})
+    Process.put(@metadata, true)
```
josevalim: Perhaps we could erase the pdict key now once enabled.

lib/logger/lib/logger.ex
```diff
@@ -695,49 +680,63 @@ defmodule Logger do
...
-
-        try do
-          notify(mode, {level, Process.group_leader(), tuple})
-          :ok
-        rescue
-          ArgumentError -> {:error, :noproc}
-        catch
-          :exit, reason -> {:error, reason}
-        end
+  def __do_log__(:warn, chardata_or_fun, metadata) do
```
josevalim: This clause should not be necessary because we already translate `:warn` to `:warning` on `__should_log__`?

lib/logger/lib/logger.ex
```diff
@@ -750,8 +749,25 @@ defmodule Logger do
...
+  ## Examples
+
+      Logger.warning("knob turned too far to the right")
+      Logger.warning(fn -> "dynamically calculated warning" end)
+      Logger.warning(fn -> {"dynamically calculated warning", [additional: :metadata]} end)
+
+  """
+  defmacro warning(chardata_or_fun, metadata \\ []) do
+    maybe_log(:warning, chardata_or_fun, metadata, __CALLER__)
   end
```
josevalim: Let's not add this one for now (since we are not adding `Logger.log(:warning, ...)`. Once this is merged, we will create an issue so we don't forget to add it for 1.11.

lib/logger/lib/logger.ex
```diff
@@ -817,20 +833,25 @@ defmodule Logger do
...
-        [module: module, function: form_fa(fun), line: line]
+      file
+      |> compile_time_application_and_file()
+      |> Keyword.merge(mfa: {module, fun, arity}, line: line)
 
     {compile_metadata, quoted_metadata} =
       if Keyword.keyword?(metadata) do
         metadata = Keyword.merge(caller, metadata)
-        {metadata, metadata}
+        quoted = Keyword.update!(metadata, :mfa, &Macro.escape/1)
```
josevalim: I think you can do:

    quoted = Macro.escape(Map.new(metadata))

And therefore `quote`+`unquote_splicing` below won't be necessary.

lib/logger/lib/logger.ex
```diff
@@ -817,20 +833,25 @@ defmodule Logger do
...
+
+        {Map.new(metadata), quote(do: %{unquote_splicing(quoted)})}
       else
-        {[],
+        quoted = Keyword.update!(caller, :mfa, &Macro.escape/1)
```
josevalim: Same here!

lib/logger/lib/logger.ex
```diff
@@ -178,37 +178,17 @@ defmodule Logger do
   Erlang's logging functionalities. All the configurations below must
   be set before the `:logger` application starts.
 
-    * `:handle_otp_reports` - redirects OTP reports to `Logger` so
-      they are formatted in Elixir terms. This effectively disables
-      Erlang standard logger. Defaults to `true`.
```
josevalim: This option has not been actually removed. We still need to know if we should remove Erlang handlers or not. So please add this line and the next removed paragraph back.

lib/logger/lib/logger/app.ex
```diff
@@ -77,64 +65,64 @@ defmodule Logger.App do
...
-    is_pid(Process.whereis(:logger))
+    Application.stop(:logger)
   end
 
-  defp add_elixir_handler(sasl_reports?) do
-    config = %{level: :debug, config: %{sasl_reports?: sasl_reports?}}
-    :logger.add_handler(Logger, Logger.ErlangHandler, config)
+  defp add_elixir_handler(otp_reports?, counter) do
+    level =
+      case Application.fetch_env!(:logger, :level) do
```
josevalim: Please add a TODO so we deprecate it here too.

lib/logger/lib/logger/app.ex
```diff
@@ -77,64 +65,64 @@ defmodule Logger.App do
...
-      :exit, {:noproc, _} ->
-        {:error, {:not_started, :logger}}
-    else
-      deleted_handlers ->
-        result = Application.stop(:logger)
-        add_handlers(deleted_handlers)
-        result
-    end
-  end
+    :ok = :logger.update_primary_config(%{level: :none})
```
josevalim: This is effectively disabling Erlang's logger, no? It is a bit dangerous because we can skip messages going to other loggers.

I think the steps here should be:

  1. Remove the Logger handler (so nothing is logged)
  2. Then stop the app

lib/logger/lib/logger/app.ex
```diff
@@ -77,64 +65,64 @@ defmodule Logger.App do
...
+            filter_elixir: {&Logger.Filter.elixir_domain/2, :ignore}
+          ]
+        else
+          []
+        end
+    }
+
+    primary_config = :logger.get_primary_config()
+
+    :ok = :logger.add_primary_filter(:process_disabled, {&Logger.Filter.process_disabled/2, []})
```
josevalim: Sorry, I had a slightly change of mind here. I don't want to add a filter to everything yet given Erlang/OTP may or may not add this behaviour, so let's remove the `process_disabled` filter for now and keep it only for Elixir's logger (which we already check on the `log` call).

lib/logger/lib/logger/app.ex
josevalim: Btw, the fact we need to do the above in steps is exactly why we had Logger.Config.deleted_handlers and we couldn't simply let Application.stop do them.

lib/logger/lib/logger/config.ex
```diff
@@ -1,95 +1,72 @@
...
+      truncate: Application.fetch_env!(:logger, :truncate),
+      translators: Application.fetch_env!(:logger, :translators),
+      thresholds: {sync_threshold, discard_threshold},
+      sasl: sasl_reports?
+    }
+  end
+
   def add_translator(translator) do
-    :gen_event.call(Logger, @name, {:add_translator, translator})
+    update_translators(fn t -> [translator | List.delete(t, translator)] end)
```
josevalim: This introduces races conditions if two processes are trying to add or remove translators, only one of them would in. So we have to move it back to Logger.Config to make the operation atomic.

lib/logger/lib/logger/filter.ex
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  def elixir_domain(log_event, extra) do
+    :erlang.error({:badarg, [log_event, extra]})
+  end
+
+  @doc """
+  Filter out logs if current process opted out of log reports
+  """
+  def process_disabled(_log, _extra) do
```
josevalim: As mentioned above, we can skip this filter.

lib/logger/lib/logger/filter.ex
```diff
@@ -0,0 +1,31 @@
...
+  Filter messages logged via `Logger` module when not logging OTP reports
+  """
+  def elixir_domain(%{meta: meta}, action) when action in [:stop, :ignore] do
+    case meta do
+      %{domain: [:elixir | _]} -> action
+      _ -> inverse_action(action)
+    end
+  end
+
+  def elixir_domain(log_event, extra) do
```
josevalim: Any reason why we need this clause?

hauleth: No reason, removing.

lib/logger/lib/logger/counter.ex
```diff
@@ -0,0 +1,45 @@
...
+
+  @type type() :: :ets | :counters
+  @type t() :: {type(), :ets.tid() | :counters.counters_ref()}
+
+  @spec new() :: t()
+  def new do
+    if Code.ensure_loaded?(:counters) do
+      {:counters, :counters.new(1, [:atomics])}
+    else
+      table = :ets.new(@table.Counter, [:public])
```
josevalim:
```suggestion
      table = :ets.new(@table, [:public])
```

lib/logger/lib/logger/config.ex
```diff
@@ -1,95 +1,72 @@
...
-
-  def deleted_handlers(handlers) do
-    :gen_event.call(Logger, @name, {:deleted_handlers, handlers})
-  end
-
-  def new do
-    {new_data(), new_counter()}
-  end
+  # TODO: Use counters exclusively when we require Erlang/OTP 22+.
+  defdelegate new, to: Logger.Counter
```
josevalim: Maybe we should remove this and call `Logger.Counter` directly in the appropriate places?

lib/logger/lib/logger/app.ex
hauleth: There was a problem with that filter removed, messages that are sent not via `Logger` will be delivered to the backend. For example processes spawned via `proc_lib` that raises message. Also I as a user would found that behaviour a little bit confusing as:

```
Logger.disable()

:logger.info("Foo bar")
```

Will still log messages, which can be counterintuitive, at least from my point of view.

lib/logger/lib/logger/config.ex
hauleth: We can, I did it to reduce amount of changes, but it shouldn't change much.

lib/logger/lib/logger/app.ex
hauleth: There is also another approach, add temporary filter, that will filter out `Logger` messages, and after application stops then remove it. This will ignore only messages about the `Logger` and all other messages will be still handled.

lib/logger/lib/logger/legacy_handler.ex
```diff
@@ -0,0 +1,265 @@
...
+  """
+  def log(%{meta: %{domain: [:otp, :sasl | _]}}, %{config: %{sasl: false}}) do
+    :ok
+  end
+
+  def log(%{meta: %{domain: [:supervisor_report | _]}}, %{config: %{sasl: false}}) do
+    :ok
+  end
+
+  def log(%{level: erl_level, msg: msg, meta: erl_meta}, %{id: _id, config: config}) do
```
josevalim: We should have a clause that matches on the Elixir domain. If we are in the Elixir domain, we can skip many steps below:

1. We can skip `erlang_level_to_elixir_level`
2. We can skip the `min_level` check
3. We can skip the splitting of the Erlang metadata

There are probably others.

lib/logger/lib/logger/config.ex
```diff
@@ -1,95 +1,72 @@
...
-
-  def translation_data do
-    read_data!(@data_key)
+    :logger.compare_levels(normalise(left), normalise(right))
   end
 
-  def log_data(level) do
-    %{level: min_level} = config = read_data!(@data_key)
+  defp normalise(:warn), do: :warning
+  defp normalise(level), do: level
```
josevalim: We have `String.normalize` (with z), so we should probably stick with that convention for consistency.

lib/logger/lib/logger/legacy_handler.ex
```diff
@@ -0,0 +1,265 @@
...
+    }
+  end
+
+  def adding_handler(%{config: data} = config) do
+    # TODO: When using counters exclusively then this line can be changed to use
+    # `Map.put/2` instead as we will not need to pass ETS table
+    new_data =
+      Logger.Config.default_config()
+      |> Map.merge(data || %{})
+      |> Map.put_new_lazy(:counter, &Logger.Config.new/0)
```
josevalim: So when we ask the handler to be added on Logger.App.start, we pass the counter. Why do we need to pass it here again? Is it because someone may add it *directly*? If so, I don't care about this use case. We can raise if someone tries to add it directly, especially because it won't work without the Logger process anyway.

hauleth: It is just in case when someone will use `:logger.set_handler_config/2` by the accident instead of `:logger.update_handler_config/2`.

lib/logger/lib/logger.ex
```diff
@@ -695,49 +680,63 @@ defmodule Logger do
...
 
+  def __do_log__(level, chardata, metadata)
+      when (is_binary(chardata) or is_list(chardata)) and is_map(metadata) do
+    :logger.log(level, chardata, add_elixir_domain(metadata))
+  end
+
+  # # TODO: Remove that in Elixir 2.0
+  def __do_log__(level, atom, metadata) do
+    # TODO: Uncomment this
+    # bare_log(:warning, "Passed #{inspect(atom)} which is not binary() nor iolist() to the Logger, this is deprecated")
```
josevalim: Let's enable this one, since we don't want to support it anyway and people can make it work on previous Elixir versions too:

    IO.warn "Passing #{inspect(other)} to Logger is deprecated, expected a binary or an iolist"

The line below should also be:

    :logger.log(level, to_string(other), add_elixir_domain(metadata))

As this is not only about atoms.

hauleth: You say it should be `IO.warn/1` instead of `bare_log/2`?

lib/logger/lib/logger/config.ex
josevalim: @hauleth yeah, the counter is treated especially everywhere in this new implementation, so I think it makes sense.

lib/logger/lib/logger/legacy_handler.ex
```diff
@@ -0,0 +1,265 @@
...
+    config =
+      %{
+        id: id,
+        module: module,
+        level: :all,
+        filter_default: :log,
+        formatter: {:logger_formatter, %{}},
+        filters: filters
+      }
+      |> Map.merge(new_config)
```
josevalim: Maybe we should do: `Map.merge(old_config, new_config)` where `old_config` is the second argument?

lib/logger/lib/logger.ex
```diff
@@ -695,49 +680,58 @@ defmodule Logger do
...
 
+  def __do_log__(level, chardata, metadata)
+      when (is_binary(chardata) or is_list(chardata)) and is_map(metadata) do
+    :logger.log(level, chardata, add_elixir_domain(metadata))
+  end
+
+  # # TODO: Remove that in Elixir 2.0
+  def __do_log__(level, atom, metadata) do
+    # TODO: Uncomment this
+    bare_log(:warning, "Passing #{inspect(other)} to Logger is deprecated, expected a binary or an iolist")
```
josevalim:
```suggestion
    IO.warn("Passing #{inspect(other)} to Logger is deprecated, expected a binary or an iolist")
```

We generally do not emit deprecation warnings through Logger. Also, please remove the TODO above.

josevalim: Also note you have to rename the variable atom to other, above and below. :)

lib/logger/lib/logger/app.ex
josevalim: @hauleth I like the temporary filter and it will help keep things simple. :+1:

lib/logger/lib/logger.ex
hauleth: I have tried that earlier but it failed with escaping a little bit too much:

```
:gen_event handler Logger.Backends.Console installed in Logger terminating
** (exit) an exception was raised:
    ** (Protocol.UndefinedError) protocol String.Chars not implemented for {:__aliases__, [counter: {LoggerTest, 173}, line: 124], [:CustomTest]} of type Tuple
        (elixir 1.10.0-dev) lib/string/chars.ex:3: String.Chars.impl_for!/1
        (elixir 1.10.0-dev) lib/string/chars.ex:22: String.Chars.to_string/1
        (logger 1.10.0-dev) lib/logger/formatter.ex:176: Logger.Formatter.metadata/1
        (logger 1.10.0-dev) lib/logger/formatter.ex:152: anonymous fn/6 in Logger.Formatter.format/5
        (elixir 1.10.0-dev) lib/enum.ex:2063: Enum."-reduce/3-lists^foldl/2-0-"/3
        (logger 1.10.0-dev) lib/logger/formatter.ex:151: Logger.Formatter.format/5
        (logger 1.10.0-dev) lib/logger/backends/console.ex:186: Logger.Backends.Console.format_event/5
        (logger 1.10.0-dev) lib/logger/backends/console.ex:140: Logger.Backends.Console.log_event/5
        (logger 1.10.0-dev) lib/logger/backends/console.ex:45: Logger.Backends.Console.handle_event/2
        (stdlib 3.9.2) gen_event.erl:577: :gen_event.server_update/4
        (stdlib 3.9.2) gen_event.erl:559: :gen_event.server_notify/4
        (stdlib 3.9.2) gen_event.erl:300: :gen_event.handle_msg/6
        (stdlib 3.9.2) proc_lib.erl:249: :proc_lib.init_p_do_apply/3
```

lib/logger/lib/logger/app.ex
josevalim: Good point. Given that `Logger.configure` changes the level for everyone, it makes sense if `Logger.disable` disables it for everyone too. Thanks!

lib/logger/lib/logger/legacy_handler.ex
hauleth: Translations will probably be moved one layer upward to the formatters as IMHO that makes more sense and allow user to have better integration with other backends. For example imagine that user want to use `logger_disk_log_h` but want to have Elixir-like format used in that log - right now that will require user to rewrite whole formatting on their own, with formatter as a separate layer they could use just that module as formatter.

lib/logger/lib/logger.ex
josevalim: Thanks!

lib/logger/lib/logger/app.ex
hauleth: I have tested this right now, and for some reason it works as expected without this line. I am not sure why this works though.

josevalim: Good call. It may be because Erlang no longer logs application stopped by default. Let's remove this line then and this should be good to go!

lib/logger/lib/logger.ex
```diff
@@ -695,49 +688,57 @@ defmodule Logger do
...
-      {true, pdict} ->
-        case Logger.Config.log_data(level) do
-          {:discard, _config} -> :error
-          {mode, config} -> {level, mode, config, pdict}
-        end
+    %{level: primary_level} = :logger.get_primary_config()
 
-      {false, _} ->
-        :error
+    if enabled?(self()) and compare_levels(level, primary_level) != :lt do
```
hauleth: I think that `__should_log__` should also be marked as to be removed in future as `logger` will handle all that for us. Only reason for it to stay is the compile-time log removal.

lib/logger/lib/logger/app.ex
```diff
@@ -77,64 +65,64 @@ defmodule Logger.App do
...
-    is_pid(Process.whereis(:logger))
-  end
+  defp add_elixir_handler(otp_reports?, counter) do
+    # TODO: Deprecate this is future releases, probably whole `logger.level`
+    # option and suggest users to use `kernel.logger_level` instead
+    level =
+      case Application.fetch_env!(:logger, :level) do
+        :warn -> :warning
+        other -> other
+      end
```
hauleth: @josevalim we should also think what we should do in case when user explicitly defined the `kernel.logger_level` configuration option as currently we always will lower it to the `debug` (default is `notice`). This can be also a little bit confusing for users that include Elixir libraries to their Erlang apps.

lib/logger/lib/logger/legacy_handler.ex
```diff
@@ -5,6 +5,17 @@ defmodule Logger.LegacyHandler do
...
 
+  def load_log_level do
+    # TODO: Deprecate if level is warning.
+    # TODO: Consider deprecating logger.level altogether in favor of kernel.logger_level.
+    level = case Application.fetch_env!(:logger, :level) do
+      :warn -> :warning
+      level -> level
+    end
+
+    :ok = :logger.set_primary_config(:level, level)
```
hauleth: As earlier, we need to find a way to handle situation when `kernel.logger_level` and `logger.level` are set and which one should be more important in such situation.

josevalim: Good point. Maybe we should allow logger.level to be unset and then we default to kernel.logger_level. Let's tackle this before merging.

lib/logger/lib/logger/legacy_handler.ex
josevalim: This shouldn't impact adding handler, right?

lib/logger/lib/logger.ex
josevalim: I think we will have to keep it because of anonymous functions. We don't want to evaluate them unless we think we will log. Or does Erlang logger allows us to pass anonymous function too? But I think they can't return `:skip` like ours, right?

hauleth: These functions cannot return `:skip` nor `{msg, metadata}` tuple, at least right now. Maybe it could be ported to the Erlang as well.

lib/logger/lib/logger/legacy_handler.ex
josevalim: That's a good point but I think we should go ahead and at least split them now. Otherwise performance will suffer and I want to avoid regression in behaviour (improvements later are always welcome though!).

lib/logger/lib/logger/legacy_handler.ex
hauleth: Unfortunately only way I see for doing that is to introduce breaking change:

```elixir
case Application.fetch_env(:logger, :level) do
  {:ok, :warn} -> :warning
  {:ok, level} -> level
  :error -> Application.fetch_env!(:kernel, :logger_level)
end
```

But that will set level to `:notice` by default instead of current `:debug`.

lib/logger/lib/logger/config.ex
```diff
@@ -132,6 +132,6 @@ defmodule Logger.Config do
     {:ok, %{config: data}} = :logger.get_handler_config(Logger)
     translators = fun.(data.translators)
     Application.put_env(:logger, :translators, translators)
-    :ok = :logger.set_handler_config(Logger, :config, %{data | translators: translators})
+    :ok = :logger.update_handler_config(Logger, :config, %{translators: translators})
```
hauleth: Unfortunately this will fail, as this is later addition to the API, OTP 21.0 do not have this function. This was my first approach as well.

lib/logger/lib/logger.ex
josevalim: Could we pass an anonymous function and expand it ourselves in a formatter, choosing to potentially skip it? I think we may be able to do it with a global formatter that matches on certain reports.. but that sounds like too much work. Changing OTP would be best.

lib/logger/lib/logger/config.ex
josevalim: Thanks, I will revert it.

lib/logger/lib/logger.ex
hauleth: I will try to port it to the OTP as well. For now I do not think that there is proper workaround over it as function itself is expanded before it hits later steps. In theory we could work around it by using report in form of `%{skip: true}` and proper filter, but I do not see any way how to lazily compute the metadata of the log message.

lib/logger/lib/logger/legacy_handler.ex
hauleth: No it should not.

lib/logger/lib/logger.ex
hauleth: Added issues on Erlang JIRA:

- https://bugs.erlang.org/browse/ERL-1085
- https://bugs.erlang.org/browse/ERL-1086

josevalim:
> I intentionally didn’t use logger:macro_log as it is undocumented. If we assume that we can safely use this non-public API then ok.

Yeah, I will start a conversation about this with Siri Hansen. If we want to implement our macros efficiently, then we need access to the underlying functionality used by the Erlang macros.

josevalim:
`macro_log` is to avoid duplicate work. `allow` though brings new features in (such as per module level).

hauleth:
@josevalim maybe we should always send synchronous message and delegate the overload handling to the [Logger Proxy](http://www.erlang.org/doc/man/logger_chapter.html#proxy). This would simplify the `LegacyHandler` while providing similar features.

lib/logger/lib/logger/formatter.ex
```diff
@@ -180,6 +184,18 @@ defmodule Logger.Formatter do
     []
   end
 
+  defp metadata(:domain, [head | tail]) when is_atom(head) do
+    # TODO: Use map_intersperse once rebased
```
hauleth: Maybe we should rebase right now?

lib/logger/test/logger/legacy_handler_test.exs
```diff
@@ -1,4 +1,4 @@
-defmodule Logger.LegacyHandlerTest do
```
hauleth: File need renaming as well.

hauleth:
Seems ready for me.

josevalim:
Thank you so much @hauleth for the amazing work! :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

josevalim:

PR #9336, opened 2019-09-10
fixed awkward wording in the documentation

brainlid:
- fixing up awkward wording on Collectable protocol documentation


fertapric:
Thanks @brainlid! 💜

PR #9339, opened 2019-09-11
Make sure releases boot even with read-only config files

josevalim:


goodsoft:
> @goodsoft generally speaking, do you think calling chmod +w on the file would be enough to fix it? Or are there some gotchas we should be aware of?

Not sure of any gotchas, besides maybe weird overall permissions of the copied file.
To be completely safe, though, I would `cat` one file into another instead of `cp`.

josevalim:
>  To be completely safe, though, I would cat one file into another instead of cp.

Can you please send a PR with this change? I will merge this and then you can change it and we will let CI be the judge. :D

PR #9340, opened 2019-09-11
Use cat to create runtime configuration more reliably

goodsoft:
As a follow up from #9339, this uses `cat` instead of `cp` to make sure that no peculiar permissions or symlink dereferencing happen with the resulting configuration file, and it is created with default current user permissions.


fertapric:
Thanks @goodsoft!

Do you know if the same applies for Windows' `copy`?

josevalim:
> Do you know if the same applies for Windows' copy?

it does not.

josevalim:

PR #9342, opened 2019-09-13
Simplify t:Registry.guards/0

eksperimental:
The first commit removes the redundant use of `[]` in the `guards/0` typespec.
~The second commit removes the unnecessary typespec `guard/0`, since we can do away as we do with the `body/0` and the others.~


eksperimental:
> I don't think we should remove public types like that. I'd leave `[guard]` and just remove the `| []`.

agreed. second commit has been removed.

fertapric:
Thanks @eksperimental! 💜

PR #9347, opened 2019-09-14
Optimize elixir_tokenizer

bjorng:
When comparing characters in `elixir_tokenizer`, replace `==` and `/=`
with the more efficient `=:=` and `=/=` operators.

The `=:=` and `=/=` operators are more efficient because they can
only return `true` if the operands have the same type:

    1> $* =:= 42.
    true
    2> $* =:= 42.0.
    false

In contrast, the `==` operator can return `true` when comparing
an integer and a float:

    3> $* == 42.0.
    true

Using `=:=` and `=/=` also gives the Erlang compiler more opportunites
for optimization. For example, when compiling the `is_sigil` macro,
the compiler can combine all eight comparisons into a single `select_val`
instruction instead of emitting eight separate instructions.

### DIFF
--- a/lib/elixir/src/elixir_tokenizer.erl
+++ b/lib/elixir/src/elixir_tokenizer.erl
@@ -13,103 +13,103 @@
 -define(is_downcase(S), (S >= $a andalso S =< $z)).
 
 %% Others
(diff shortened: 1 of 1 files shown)


fertapric:
Thanks @bjorng! 💜

PR #9348, opened 2019-09-14
Always adding :debug_info on erlc_options

### LINKED ISSUE #9184: erlc_options should behave more like elixirc_options
## Environment

- Elixir & Erlang/OTP versions (elixir --version):
```
Erlang/OTP 22 [erts-10.4.3] [source] [64-bit] [smp:12:12] [ds:12:12:10] [async-threads:1] [hipe]

Elixir 1.8.2 (compiled with Erlang/OTP 22)
```
- Operating system: macOS Mojave

## Current behavior

There difference in behaviour of `:elixirc_options` and `:elrc_options`. This is the inclusion of `:debug_info`. In `:elixirc_options` it is added always (or at least seems so when running Dialyzer) while in `:erlc_options` it need to be explicitly stated, otherwise Dialyzer will fail with information that it cannot find Erlang Core informations.

Example repo: https://github.com/hauleth/example_error

Way to reproduce error:

```
mix do compile --force, dialyzer
```

Way to resolve error:

```
env DEBUG_INFO=1 mix do compile --force, dialyzer
```

## Expected behavior

Mix should always add `:debug_info` option or should provide different way to ensure, that this is handled properly. Otherwise it has limited capabilities as the Erlang projects build tool.

kelvinst:
That way, if someone sets erlc_options to `[warnings_as_errors: true]`, `:debug_info` would still be on the options. To configure it to not save debug info, one should only add `debug_info: false` to erlc_options

Closes #9184

### DIFF
--- a/lib/mix/lib/mix/project.ex
+++ b/lib/mix/lib/mix/project.ex
@@ -729,7 +729,7 @@ defmodule Mix.Project do
       elixirc_paths: ["lib"],
       erlc_paths: ["src"],
       erlc_include_path: "include",
-      erlc_options: [:debug_info],
+      erlc_options: [],
       lockfile: "mix.lock",
       preferred_cli_env: [],
       start_permanent: false
--- a/lib/mix/lib/mix/tasks/compile.erlang.ex
+++ b/lib/mix/lib/mix/tasks/compile.erlang.ex
@@ -74,7 +74,7 @@ defmodule Mix.Tasks.Compile.Erlang do
       Mix.raise(":erlc_options should be a list of options, got: #{inspect(erlc_options)}")
     end
 
-    erlc_options = erlc_options ++ [:return, :report, outdir: compile_path, i: include_path]
+    erlc_options = erlc_options ++ [:debug_info, :return, :report, outdir: compile_path, i: include_path]
 
     erlc_options =
       Enum.map(erlc_options, fn
--- a/lib/mix/lib/mix/tasks/compile.erlang.ex
+++ b/lib/mix/lib/mix/tasks/compile.erlang.ex
@@ -74,7 +74,8 @@ defmodule Mix.Tasks.Compile.Erlang do
       Mix.raise(":erlc_options should be a list of options, got: #{inspect(erlc_options)}")
     end
 
-    erlc_options = erlc_options ++ [:debug_info, :return, :report, outdir: compile_path, i: include_path]
+    erlc_options =
+      erlc_options ++ [:debug_info, :return, :report, outdir: compile_path, i: include_path]
 
     erlc_options =
       Enum.map(erlc_options, fn
--- a/lib/mix/lib/mix/tasks/compile.erlang.ex
+++ b/lib/mix/lib/mix/tasks/compile.erlang.ex
@@ -41,14 +41,19 @@ defmodule Mix.Tasks.Compile.Erlang do
       Defaults to `"include"`.
 
     * `:erlc_options` - compilation options that apply to Erlang's
-      compiler. Defaults to `[:debug_info]`.
+      compiler. Defaults to `[]`.
 
       For a complete list of options, see `:compile.file/2`.
 
-  For example, to configure the `erlc_options` for your Erlang project you
-  may run:
+  ### Important notice
 
-      erlc_options: [:debug_info, {:i, 'path/to/include'}]
+  The following options are always added to `:erlc_options` when running the compiler:
+
+      [:debug_info, :return, :report, outdir: compile_path, i: include_path]
+
+  You can set any of those options like this to override them:
+
+      erlc_options: [debug_info: false, return: false, i: my_includes]
 
   """
 
--- a/lib/mix/lib/mix/tasks/compile.erlang.ex
+++ b/lib/mix/lib/mix/tasks/compile.erlang.ex
@@ -45,15 +45,10 @@ defmodule Mix.Tasks.Compile.Erlang do
 
       For a complete list of options, see `:compile.file/2`.
 
-  ### Important notice
+      The option `:debug_info` is always added to the end of it. You can
+      disable that using:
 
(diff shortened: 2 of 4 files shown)


kelvinst:
I also fixed a doc that was pointing to the wrong default value for `:debug_info` option on `Code.put_compiler_option`

kelvinst:
I'm checking how to add tests for this now.

Also, there should be a changelog notice or something like that, since the change breaks compatibility. e.g. if anyone was just dismissing `:debug_info` on `:erlc_options` to disable it, that would not work anymore, they would have to change their options to `debug_info: false`

kelvinst:
> I'm checking how to add tests for this now.

So as it is done now, I'm not sure there is any good way of testing it, since it is an inner logic inside `Mix.Tasks.Compile.Erlang.do_run`. Should I extract the erlc_options build to a public function and unit test it? Maybe we can extract it to `Mix.Compilers.Erlang.erlc_options` and even document it better there. WDYT?

josevalim:
You should be able to test it like this: https://github.com/elixir-lang/elixir/blob/master/lib/mix/test/mix/tasks/compile.erlang_test.exs#L14

kelvinst:
Yeah, I saw that, but what I couldn't figure out yet is how to check if the `:debug_info` option was actually applied to compilation.

michalmuskala:
@kelvinst it should be possible to check the generated files with `:beam_lib(path, [:debug_info])`

kelvinst:
Cool! Thanks @michalmuskala! Really helpful tip (I suppose you mean `:beam_lib.chunks`, right?)

kelvinst:
PS.: I just realized that I added the default options to the wrong place I guess, that way it will apply it only for the `mix compile.erlang` task, not when using `mix compile`. Will fix later.

kelvinst:
Actually, nevermind, just realized `mix compile` does call `mix compile.erlang` 🤦‍♂ I'm investigating why @hauleth's example app still don't work with my solution.

kelvinst:
Right, so the problem was that I was just running `../elixir/bin/mix do compile, dialyzer` from inside @hauleth's project,  and that was not using `../elixir/bin/mix` to run the subtasks I suppose. Did test setting `PATH=../elixir/bin:$PATH mix do compile, dialyzer` and it did work that way, so that really worked.

lib/mix/test/mix/tasks/compile.erlang_test.exs
```diff
@@ -174,4 +174,18 @@ defmodule Mix.Tasks.Compile.ErlangTest do
...
+  test "adds :debug_info to erlc_options by default" do
+    in_fixture("compile_erlang", fn ->
+      Mix.Tasks.Compile.Erlang.run([])
+
+      binary = File.read!("_build/dev/lib/sample/ebin/b.beam")
+
+      {:ok, {module, [debug_info: {:debug_info_v1, backend, data}]}} =
+        :beam_lib.chunks(binary, [:debug_info])
+
+      assert backend.debug_info(:elixir_v1, module, data, []) != {:error, :missing}
```
michalmuskala: I think this should assert with `:erlang_v1` - after all it's compiling erlang modules, so it's kind of expected it won't be able to return elixir debug info.
```suggestion
      assert backend.debug_info(:erlang_v1, module, data, []) != {:error, :missing}
```

kelvinst: Alrighty!

josevalim: I wouldn't even assert this. Asserting it returns a debug info chunk is probably good enough!

kelvinst: You mean something like this?

```elixir 
{:ok, {_, [debug_info: {:debug_info_v1, _, {info, _}}]}} = :beam_lib.chunks(binary, [:debug_info])
assert info != :none
```

kelvinst: I've checked and the only difference is that it returns `:none` for that part when debug info is disabled, instead of the list of attributes.

kelvinst: I have started with that actually, just changed to use `backend.debug_info` because I was afraid that maybe the format could change in the future and then the test could fail. Very unlikely, but who knows 🤷‍♂

josevalim:
The point is that if they change the format, this test should not fail. It
is ok for them to do that. As long as there is something, we are good.
-- 


lib/mix/lib/mix/project.ex
```diff
@@ -729,7 +729,7 @@ defmodule Mix.Project do
       elixirc_paths: ["lib"],
       erlc_paths: ["src"],
       erlc_include_path: "include",
-      erlc_options: [:debug_info],
+      erlc_options: [],
```
kelvinst: This can be empty since we already add it everytime when compiling erlang.

lib/mix/lib/mix/tasks/compile.erlang.ex
```diff
@@ -41,14 +41,19 @@ defmodule Mix.Tasks.Compile.Erlang do
...
       For a complete list of options, see `:compile.file/2`.
 
-  For example, to configure the `erlc_options` for your Erlang project you
-  may run:
+  ### Important notice
 
-      erlc_options: [:debug_info, {:i, 'path/to/include'}]
+  The following options are always added to `:erlc_options` when running the compiler:
+
+      [:debug_info, :return, :report, outdir: compile_path, i: include_path]
```
josevalim: Return and report are implementation details. If the user changes them, then the compiler will fail. I would remove this section (or talk about only debug_info).

kelvinst: Cool, thanks!! I will comment only about debug_info.

fertapric:
Thanks @kelvinst! 💜

PR #9352, opened 2019-09-15
Update on_definition with Module.exists?/2 docs & test

### LINKED ISSUE #9351: on_definition and Module.defines?/2
### Environment

* Elixir & Erlang/OTP versions: Elixir 1.10.0-dev (9213f61) (compiled with Erlang/OTP 22)

### Current behavior

Per [`@on_definition`](https://hexdocs.pm/elixir/Module.html?#module-on_definition-1) docs:

> Note the hook receives the quoted arguments and it is invoked before the function is stored in the module. So Module.defines?/2 will return false for the first clause of every function.

However that is not the behaviour I'm observing here:

```elixir
defmodule Hooks do
  def __on_definition__(env, _kind, name, args, _guards, _body) do
    IO.write "is #{inspect(env.module)}.#{name}/#{length(args)} defined? "
    IO.inspect Module.defines?(env.module, {name, length(args)})
  end
end

defmodule Bar do
  @on_definition Hooks

  def foo(x) when is_integer(x), do: :int
  def foo(x) when is_binary(x), do: :binary
end
```

```
Compiling 1 file (.ex)
is Bar.foo/1 defined? true
is Bar.foo/1 defined? true
```

wojtekmach:
Closes #9351.

I wasn't sure how to update the docs and since we mention elsewhere we
receive quotes arguments I just remove the whole section.

Alternatively, here's my attempt at updating the docs:

> The hook is invoked after the function was stored in the module,
> this means that, for example, `Module.defines?/2` would always
> return `true` when invoked in this hook.

### DIFF
--- a/lib/elixir/lib/module.ex
+++ b/lib/elixir/lib/module.ex
@@ -439,10 +439,6 @@ defmodule Module do
     * the list of quoted guards
     * the quoted function body
 
-  Note the hook receives the quoted arguments and it is invoked before
-  the function is stored in the module. So `Module.defines?/2` will return
-  `false` for the first clause of every function.
-
   If the function/macro being defined has multiple clauses, the hook will
   be called for each clause.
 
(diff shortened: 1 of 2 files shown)


whatyouhide:
Thank you @wojtekmach! 💟

PR #9354, opened 2019-09-16
IEx: Link to callback docs instead of copying them

wojtekmach:
Similar to https://github.com/elixir-lang/ex_doc/pull/1094


josevalim:
In this case, I would prefer to keep it inline as today because linking is convenient for browsers, but not very convenient from the CLI. :) Thoughts?

wojtekmach:
Fair enough, we still have a small issue that we show `@callback` and docs for potentially private types and such but again, because we don’t actually link them it’s less of an issue. 👍

PR #9359, opened 2019-09-21
Adding tee/2 function to Kernel

kelvinst:
This is a possible implementation to the `tee` function mentioned on
[elixir-core:7754] Proposal: Allow IO.inspect to accept a function as argument

https://groups.google.com/d/msgid/elixir-lang-core/CAGnRm4Ji7kAjv5i4_04AVp-hae4ghj1DgZx1zr%2B4LyGBYpj5_A%40mail.gmail.com

Maybe another `Kernel` function is not that good. But I think is a pretty useful
function to have it auto-imported.

PS: I didn't "eat my own dog food" yet because this might not be the implementation
you want. Once reviewed and approved we can think about the functions we can use
`tee/2` (like `IO.inspect/2`)

### DIFF
--- a/lib/ex_unit/lib/ex_unit/case.ex
+++ b/lib/ex_unit/lib/ex_unit/case.ex
@@ -223,7 +223,26 @@ defmodule ExUnit.Case do
     end
 
     quote do
-      async = !!unquote(opts)[:async]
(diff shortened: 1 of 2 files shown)


kelvinst:
So sorry, old branch just showed up here, closing.

PR #9360, opened 2019-09-21
Allow to reset async for ExUnit.Case by using it again

### LINKED ISSUE #9355: Warn when ExUnit.Case is `use`d multiple times
### Environment

```
Erlang/OTP 20 [erts-9.3.3.11] [source] [64-bit] [smp:6:6] [ds:6:6:10] [async-threads:10] [hipe] [kernel-poll:false]

Elixir 1.6.6 (compiled with OTP 20)
```
* Operating system: 
```
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Debian
Description:	Debian GNU/Linux 9.9 (stretch)
Release:	9.9
Codename:	stretch
```
(in docker)

### Current behavior

I don't know if this is a bug at all. I just discovered this and found it confusing. Given a test module with:
```
  use MyApp.CoreModel.ModelCase
  use ExUnit.Case, async: true
```
Where `ModelCase` contains `use ExUnit.CaseTemplate`, ExUnit does not throw a warning and does not set `async: true` for tests in this module.

This is confusing because this will work and the tests will have `:async` set to `true`
```
  use ExUnit.Case, async: true
  use MyApp.CoreModel.ModelCase
```

The correct approach is passing the options to `ModelCase`:
```
  use MyApp.CoreMosel.ModelCase, async: true
```

### Expected behavior

When ExUnit.Case is `use`d multiple times with different options, I get a warning.

kelvinst:
Closes #9355

### DIFF
--- a/lib/ex_unit/lib/ex_unit/case.ex
+++ b/lib/ex_unit/lib/ex_unit/case.ex
@@ -223,10 +223,8 @@ defmodule ExUnit.Case do
     end
 
     quote do
-      @ex_unit_default_opts [async: true]
-
       old_opts = Module.get_attribute(__MODULE__, :ex_unit_opts)
-      new_opts = Keyword.merge(@ex_unit_default_opts, unquote(opts) || [])
+      new_opts = Keyword.merge([async: true], unquote(opts) || [])
 
       if old_opts && old_opts != new_opts do
         message = """
--- a/lib/ex_unit/lib/ex_unit/case.ex
+++ b/lib/ex_unit/lib/ex_unit/case.ex
@@ -223,25 +223,6 @@ defmodule ExUnit.Case do
     end
 
     quote do
-      old_opts = Module.get_attribute(__MODULE__, :ex_unit_opts)
-      new_opts = Keyword.merge([async: true], unquote(opts) || [])
-
-      if old_opts && old_opts != new_opts do
-        message = """
-        ExUnit.Case was already used on this module #{inspect(__MODULE__)} \
-        with the options `#{inspect(@ex_unit_opts)}`.
-
-        Please make sure to set the right options on the first `use ExUnit.Case`, \
-        as they are not overriden by any subsequent calls.
-        """
-
-        IO.warn(message, Macro.Env.stacktrace(__ENV__))
-      else
-        Module.put_attribute(__MODULE__, :ex_unit_opts, new_opts)
-      end
-
-      async = @ex_unit_opts[:async]
-
       unless Module.has_attribute?(__MODULE__, :ex_unit_tests) do
         tag_check =
           [:moduletag, :describetag, :tag]
@@ -266,9 +247,15 @@ defmodule ExUnit.Case do
 
         @before_compile ExUnit.Case
         @after_compile ExUnit.Case
-        @ex_unit_async async
+        @ex_unit_async false
         @ex_unit_describe nil
         use ExUnit.Callbacks
+      end
+
+      async = unquote(opts)[:async]
+
+      if is_boolean(async) do
+        @ex_unit_async async
       end
 
       import ExUnit.Callbacks
--- a/lib/ex_unit/lib/ex_unit/case.ex
+++ b/lib/ex_unit/lib/ex_unit/case.ex
@@ -223,7 +223,26 @@ defmodule ExUnit.Case do
     end
 
     quote do
-      async = !!unquote(opts)[:async]
+      @ex_unit_default_opts [async: true]
+
+      old_opts = Module.get_attribute(__MODULE__, :ex_unit_opts)
+      new_opts = Keyword.merge(@ex_unit_default_opts, unquote(opts) || [])
+
+      if old_opts && old_opts != new_opts do
+        message = """
+        ExUnit.Case was already used on this module #{inspect(__MODULE__)} \
+        with the options `#{inspect(@ex_unit_opts)}`.
+
+        Please make sure to set the right options on the first `use ExUnit.Case`, \
+        as they are not overriden by any subsequent calls.
+        """
+
+        IO.warn(message, Macro.Env.stacktrace(__ENV__))
+      else
+        Module.put_attribute(__MODULE__, :ex_unit_opts, new_opts)
+      end
+
+      async = @ex_unit_opts[:async]
 
       unless Module.has_attribute?(__MODULE__, :ex_unit_tests) do
         tag_check =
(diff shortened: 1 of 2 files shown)


lib/ex_unit/test/ex_unit/case_test.exs
```diff
@@ -112,4 +114,29 @@ defmodule ExUnit.CaseTest do
...
     end
   end
+
+  test "warns for using it twice with different options" do
+    assert capture_io(:stderr, fn ->
+        defmodule WarnsUsedTwice do
+          use ExUnit.Case
+          use ExUnit.Case, async: true
+        end
+      end) == ""
```
kelvinst: Not sure if this is the expected behaviour, had to change the `__using__` macro a little bit to make this work.

lib/ex_unit/lib/ex_unit/case.ex
```diff
@@ -223,7 +223,26 @@ defmodule ExUnit.Case do
     end
 
     quote do
-      async = !!unquote(opts)[:async]
+      @ex_unit_default_opts [async: true]
```
whatyouhide: You should be able to do this with a normal variable without the need for a module attribute.

lib/ex_unit/test/ex_unit/case_test.exs
```diff
@@ -112,4 +114,29 @@ defmodule ExUnit.CaseTest do
...
+        end
+      end)
+
+    assert stderr =~ """
+           ExUnit.Case was already used on this module ExUnit.CaseTest.WarnsUsedTwice \
+           with the options `[async: true]`.
+
+           Please make sure to set the right options on the first `use ExUnit.Case`, \
+           as they are not overriden by any subsequent calls.
+           """
```
whatyouhide: There's no need to match on the whole exact message. Let's just match on

```elixir
assert stderr =~ "ExUnit.Case was already used"
```

lib/ex_unit/lib/ex_unit/case.ex
```diff
@@ -223,7 +223,26 @@ defmodule ExUnit.Case do
...
+
+      old_opts = Module.get_attribute(__MODULE__, :ex_unit_opts)
+      new_opts = Keyword.merge(@ex_unit_default_opts, unquote(opts) || [])
+
+      if old_opts && old_opts != new_opts do
```
whatyouhide: I think we should warn regardless of different options, and just warn if `use ExUnit.Case` was called before.

lib/ex_unit/lib/ex_unit/case.ex
kelvinst: True, thanks for the heads up!

lib/ex_unit/test/ex_unit/case_test.exs
kelvinst: Sure 👍

lib/ex_unit/lib/ex_unit/case.ex
kelvinst: I was afraid that there was situations that it makes sense to call it twice because of this condition: https://github.com/elixir-lang/elixir/pull/9360/files/4aaea84c6dc9ba6db0bfecbc61888efea7c1d2f5#diff-da9b91caf1cd02e59ee50c01929a2058R247

I will change to the way you suggested and check the whole elixir codebase + some personal phoenix projects to make sure that's not the case. Thanks.

kelvinst: Changed it and it warned for this test: https://github.com/elixir-lang/elixir/blob/master/lib/ex_unit/test/ex_unit/case_template_test.exs#L40

Investigating a bit its origin I also discovered the origin of the condition I mentioned on the previous comment: https://github.com/elixir-lang/elixir/commit/acf912c2c87758434d76f57276345f3daa699a78

Looks like the idea was to allow composition of `ExUnit.CaseTemplate`s, should we keep my change the way it is then? So it only warns when the use tries to change the options? Or maybe we should not warn at all and come up with another fix for the issue?

josevalim: Yes, perhaps the best solution is to simply allow async to work through multiple calls. Something like this:

    given_async = unquote(opts)[:async]

and then in the less you do:

    unless ... do
      @ex_unit_async false
    end

And after the unless you do:

    if unquote(opts)[:async] do
      @ex_unit_async true
    end

kelvinst: Yeah, I thought about doing that, but didn't do because I thought it was that way for a reason, since it is that way from the day it was introduced: https://github.com/elixir-lang/elixir/commit/acf912c2c87758434d76f57276345f3daa699a78#diff-da9b91caf1cd02e59ee50c01929a2058R35-R45

I will investigate a bit, if I don't find anything, I will change to your suggestion and check it out, maybe some tests would fail if it really breaks stuff 🤷‍♂

lib/ex_unit/lib/ex_unit/case.ex
kelvinst: Just realized the default value for async is actually wrong here, fixing with the new commits.

lib/ex_unit/lib/ex_unit/case.ex
kelvinst: About your suggestion: should we set async to false inside the `unless`, the following would set it to `false`:

```elixir
defmodule Test do
  use ExUnit.Case, async: true
  use ExUnit.Case
end
```

Maybe the behavior most people would expect here would be not to reset `async` to `false` if the option is not given, IDK.

kelvinst: Oh, right, nevermind what I said, the unless would filter out. 🤦‍♂

kelvinst: Just changed the condition to `if unquote(opts)[:async] do` that way this would reset it to false:

```elixir
defmodule Test do
  use ExUnit.Case, async: true
  use ExUnit.Case, async: false
end
```

lib/ex_unit/test/ex_unit/case_test.exs
```diff
@@ -113,3 +113,39 @@ defmodule ExUnit.CaseTest do
...
+end
+
+defmodule ExUnit.DoubleCaseTest4 do
+  use ExUnit.Case
+  use ExUnit.Case, async: true
+
+  test "async must be true", context do
+    assert context.async
+  end
+end
```
kelvinst: Couldn't find a better way to check the value saved to `async` attribute.

whatyouhide:
So I'm confused: the original issue #9355 is about warning if `ExUnit.Case` is used multiple times, but now we're merging the options and not emitting warnings. 🤔

josevalim:
We want to allow it to be called multiples times because of case template. So we are fixing it so it works as expected if called multiple times (i.e. latest option wins).

whatyouhide:
@josevalim got it!

Thanks @kelvinst! 💟

PR #9364, opened 2019-09-23
Don't allow cycles when loading .iex.exs

### LINKED ISSUE #9363: import_file/1 and import_file_if_available/1 can result in an infinite loop
* Elixir & Erlang/OTP versions (elixir --version): OTP 21, Elixir 1.8.2
* Operating system: linux, Heroku (stack-16)

### Current behavior

I have a project-local `.iex.exs` file that contains:

```elixir
import_file_if_available "~/.iex.exs"
```

When running `iex -S mix` (and probably just `iex`) from the project locally, IEx just works. On Heroku, the interactive prompt never appears because `Path.expand("~/.iex.exs") == Path.join(File.cwd!, ".iex.exs")`. Because `import_file` and `import_file_if_available` are macros, it is not possible to _conditionally_ import such a file as this results in a recursively infinite macro expansion.

### Expected behavior

The IEx prompt _should_ appear. I’m not sure if, before expansion, `import_file` should see if it’s already read the named file and refuse to do so if it has, but that would fix the issue if it is possible. I am currently on vacation, so I’m a few weeks away from being able to look at this as a possible fix.

wojtekmach:
Closes #9363

### DIFF
--- a/lib/iex/lib/iex/evaluator.ex
+++ b/lib/iex/lib/iex/evaluator.ex
@@ -256,6 +256,8 @@ defmodule IEx.Evaluator do
     {result, binding, env, scope} =
       :elixir.eval_forms(forms, state.binding, state.env, state.scope)
 
+    Process.delete(:iex_imported_paths)
+
     unless result == IEx.dont_display_result() do
       io_inspect(result)
     end
--- a/lib/iex/lib/iex/evaluator.ex
+++ b/lib/iex/lib/iex/evaluator.ex
@@ -188,7 +188,6 @@ defmodule IEx.Evaluator do
       env = :elixir.env_for_eval(state.env, file: path, line: 1)
       Process.put(:iex_imported_paths, MapSet.new([path]))
       {_result, binding, env, _scope} = :elixir.eval_forms(quoted, state.binding, env)
-      # Process.delete(:iex_imported_paths)
       %{state | binding: binding, env: :elixir.env_for_eval(env, file: "iex", line: 1)}
     catch
       kind, error ->
--- a/lib/iex/lib/iex/evaluator.ex
+++ b/lib/iex/lib/iex/evaluator.ex
@@ -186,7 +186,9 @@ defmodule IEx.Evaluator do
 
       # Evaluate the contents in the same environment server_loop will run in
       env = :elixir.env_for_eval(state.env, file: path, line: 1)
+      Process.put(:iex_imported_paths, MapSet.new([path]))
       {_result, binding, env, _scope} = :elixir.eval_forms(quoted, state.binding, env)
+      # Process.delete(:iex_imported_paths)
       %{state | binding: binding, env: :elixir.env_for_eval(env, file: "iex", line: 1)}
     catch
       kind, error ->
@@ -255,8 +257,6 @@ defmodule IEx.Evaluator do
   defp handle_eval({:ok, forms}, code, line, iex_state, state) do
     {result, binding, env, scope} =
       :elixir.eval_forms(forms, state.binding, state.env, state.scope)
-
-    Process.delete(:iex_imported_paths)
 
     unless result == IEx.dont_display_result() do
       io_inspect(result)
--- a/lib/iex/lib/iex/evaluator.ex
+++ b/lib/iex/lib/iex/evaluator.ex
@@ -194,6 +194,8 @@ defmodule IEx.Evaluator do
         io_result("Error while evaluating: #{path}")
         print_error(kind, error, __STACKTRACE__)
         state
+    after
+      Process.delete(:iex_imported_paths)
     end
   end
 
--- a/lib/iex/lib/iex/helpers.ex
+++ b/lib/iex/lib/iex/helpers.ex
@@ -1068,14 +1068,15 @@ defmodule IEx.Helpers do
     path = Path.expand(path)
 
     if not optional? or File.exists?(path) do
-      imported_paths = Process.get({__MODULE__, :imported_paths}, MapSet.new())
+      imported_paths = Process.get(:iex_imported_paths, MapSet.new())
 
       if path in imported_paths do
+        Process.delete(:iex_imported_paths)
         message = "path #{path} was already imported, circular file imports are not allowed"
         raise ArgumentError, message
       end
 
-      Process.put({__MODULE__, :imported_paths}, MapSet.put(imported_paths, path))
+      Process.put(:iex_imported_paths, MapSet.put(imported_paths, path))
       path |> File.read!() |> Code.string_to_quoted!(file: path)
     end
   end
--- a/lib/iex/lib/iex/helpers.ex
+++ b/lib/iex/lib/iex/helpers.ex
@@ -1071,13 +1071,14 @@ defmodule IEx.Helpers do
       if imported_paths = Process.get(:iex_imported_paths) do
         if path in imported_paths do
           message = "path #{path} was already imported, circular file imports are not allowed"
-          raise ArgumentError, message
+          IO.warn(message, [])
+        else
+          Process.put(:iex_imported_paths, MapSet.put(imported_paths, path))
+          path |> File.read!() |> Code.string_to_quoted!(file: path)
         end
-
-        Process.put(:iex_imported_paths, MapSet.put(imported_paths, path))
+      else
+        path |> File.read!() |> Code.string_to_quoted!(file: path)
       end
-
-      path |> File.read!() |> Code.string_to_quoted!(file: path)
     end
   end
 
--- a/lib/iex/lib/iex/helpers.ex
+++ b/lib/iex/lib/iex/helpers.ex
@@ -1068,15 +1068,16 @@ defmodule IEx.Helpers do
     path = Path.expand(path)
 
     if not optional? or File.exists?(path) do
-      imported_paths = Process.get(:iex_imported_paths, MapSet.new())
+      if imported_paths = Process.get(:iex_imported_paths) do
+        if path in imported_paths do
+          Process.delete(:iex_imported_paths)
+          message = "path #{path} was already imported, circular file imports are not allowed"
+          raise ArgumentError, message
+        end
 
-      if path in imported_paths do
-        Process.delete(:iex_imported_paths)
-        message = "path #{path} was already imported, circular file imports are not allowed"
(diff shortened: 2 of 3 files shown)


lib/iex/test/iex/helpers_test.exs
```diff
@@ -1000,6 +1000,18 @@ defmodule IEx.HelpersTest do
     test "does not raise if file is missing and using import_file_if_available" do
       assert "nil" == capture_iex("import_file_if_available \"nonexistent\"")
     end
+
+    test "circular imports" do
+      dot_1 = "import_file \"dot-iex-2\""
+      dot_2 = "import_file \"dot-iex-1\""
+
+      with_file(["dot-iex-1", "dot-iex-2"], [dot_1, dot_2], fn ->
+        capture_io(:stderr, fn ->
```
fertapric: What is this `capture_io(:stderr,` capturing? I think it might not be needed.

wojtekmach: good call, I copied it from a similar test but it turns out it wasn't needed there either :D

lib/iex/lib/iex/helpers.ex
```diff
@@ -1068,6 +1068,14 @@ defmodule IEx.Helpers do
     path = Path.expand(path)
 
     if not optional? or File.exists?(path) do
+      imported_paths = Process.get({__MODULE__, :imported_paths}, MapSet.new())
```
josevalim: I don't think we need to detect loops themselves, only if we are expanding to yourself. This could be done by passing `__CALLER__` to this function and checking if `Path.expand(path) == __CALLER__.file`. This also guarantees we won't pollute the process dictionary (which we don't clean right now and I don't think we could). :)

wojtekmach: I guess with your approach we would still have the issue that the test shows, but I agree that's probably way less common than the issue that user reported (that the path resolves to itself). Will work on a patch!

wojtekmach: @josevalim I tried that but the `caller.file` is always `iex`. Is there a way to update calling environment?

Here's a diff against master of what I tried:

```diff
diff --git a/lib/iex/lib/iex/helpers.ex b/lib/iex/lib/iex/helpers.ex
index 773ea0376..dbb66aeca 100644
--- a/lib/iex/lib/iex/helpers.ex
+++ b/lib/iex/lib/iex/helpers.ex
@@ -1057,18 +1057,20 @@ defmodule IEx.Helpers do
   in behalf of the user.
   """
   defmacro import_file_if_available(path) when is_binary(path) do
-    import_file_if_available(path, true)
+    import_file_if_available(path, __CALLER__, true)
   end
 
   defmacro import_file_if_available(_) do
     raise ArgumentError, "import_file_if_available/1 expects a literal binary as its argument"
   end
 
-  defp import_file_if_available(path, optional?) when is_binary(path) do
-    path = Path.expand(path)
+  defp import_file_if_available(path, caller, optional?) when is_binary(path) do
+    if Path.expand(path) == caller.file do
+      raise ArgumentError, "attempting to import #{inspect(path)} from itself"
+    end
 
     if not optional? or File.exists?(path) do
-      path |> File.read!() |> Code.string_to_quoted!(file: path)
+      path |> File.read!() |> Code.string_to_quoted!(file: path)
     end
   end
 
@@ -1100,7 +1102,7 @@ defmodule IEx.Helpers do
   """
   @doc since: "1.4.0"
   defmacro import_file(path) when is_binary(path) do
-    import_file_if_available(path, false)
+    import_file_if_available(path, __CALLER__, false)
   end
 
   defmacro import_file(_) do
@@ -1110,7 +1112,7 @@ defmodule IEx.Helpers do
   @doc false
   @deprecated "Use import_file_if_available/1 instead"
   defmacro import_file(path, opts) when is_binary(path) and is_list(opts) do
-    import_file_if_available(path, Keyword.get(opts, :optional, false))
+    import_file_if_available(path, __CALLER__, Keyword.get(opts, :optional, false))
   end
 
   @doc """
```

lib/iex/lib/iex/evaluator.ex
```diff
@@ -256,6 +256,8 @@ defmodule IEx.Evaluator do
     {result, binding, env, scope} =
       :elixir.eval_forms(forms, state.binding, state.env, state.scope)
 
+    Process.delete(:iex_imported_paths)
```
josevalim: Unfortuntately, we can also invoke `import_file` from inside IEx, which means the pdict will never be cleaned. My suggesstion:

1. set the process dictionary before we import in the evaluator
2. in the import_file code, only do the check and change the pdict if the pdict was already set

This means we check unfortunately only the default import but it also means we don't leak the pdict.

wojtekmach:
You mean from a “regular” iex session, as opposed to load .iex.exs? I did test both scenarios and pdict was properly cleaned.

wojtekmach:
Oh, did you mean that some others parts of IEx itself would call it and not the user per se, as in typing it to evaluate?

lib/iex/lib/iex/evaluator.ex
josevalim: Oh, you always clean the pdict after any eval. Yeah, I don't like this, it feels we are leaking concerns. :'( It would be ok to do it exclusively in the import code (because that code is about importing ater all).

wojtekmach:
Updated! Btw, the error message doesn't sound quite right to me, but can't think of a better one so let me know if you have any ideas!

lib/iex/lib/iex/evaluator.ex
```diff
@@ -186,6 +186,7 @@ defmodule IEx.Evaluator do
 
       # Evaluate the contents in the same environment server_loop will run in
       env = :elixir.env_for_eval(state.env, file: path, line: 1)
+      Process.put(:iex_imported_paths, MapSet.new([path]))
       {_result, binding, env, _scope} = :elixir.eval_forms(quoted, state.binding, env)
```
josevalim: And we clear it next, wdyt?

josevalim: We can probably clear it in an `after` clause after catch, which means we don't need to clean it when we raise in the helpers.

wojtekmach: good call, let me try that

halostatue:
If I understand this change correctly, would the Heroku case of `import_file_if_available "~/.iex.exs"` result in an exception instead of silently ignoring the circle cause by the fact that `/app` and `~/` are the same path on Heroku instances?

I’m not entirely sure that would be an improvement over the infinite loop I’m seeing in #9363.

wojtekmach:
Yes, there would be an exception raised by that's ok, your IEx session will carry on:

```
~/src/elixir[wm-circular-imports]% iex
Erlang/OTP 22 [erts-10.4.4] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1] [hipe]

Interactive Elixir (1.10.0-dev) - press Ctrl+C to exit (type h() ENTER for help)
Error while evaluating: /Users/wojtek/src/elixir/.iex.exs
** (ArgumentError) path /Users/wojtek/src/elixir/.iex.exs was already imported, circular file imports are not allowed
    (iex 1.10.0-dev) lib/iex/helpers.ex:1074: IEx.Helpers.import_file_if_available/2
    (iex 1.10.0-dev) expanding macro: IEx.Helpers.import_file/1
    .iex.exs:1: (file)
iex(1)> 42
42
iex(2)>
```

this means however that on Heroku you'd get that error every single time which is not ideal and you'd probably end up adjusting your projects `./.iex.exs` to avoid them. Silently allowing infinite imports doesn't seem great.

halostatue:
Is there a difference between the following?

```elixir
import_file_if_available "~/.iex.exs"

alias MyApp.{Repo, User}
import_if_available Ecto.Query
```

and

```elixir
alias MyApp.{Repo, User}
import_if_available Ecto.Query

import_file_if_available "~/.iex.exs"
```

If there is, then it may be worth documenting this difference or seeing if there’s a way to _warn_ rather than _raise_ on this.

wojtekmach:
There's no difference, because an error is raised you won't get that Ecto.Query import into your scope.

Yes, turning error into a warning would solve this but I'm not sure about this. Personally, I'd like to find a reliable solution for this particular situation that neither warns nor errors, and if there isn't one, while it's not ideal, I'd say that you shouldn't be doing `import_file_if_available "~/.iex.exs"` then and that's that. (I had the same problem on Heroku couple years ago, I didn't submit an issue and ended up removing the import.)

josevalim:
Since the error is because we have already imported the file, I think moving to a warning would be ok and help us ensure the functionality still works. So I would be ok with going in this direction.

wojtekmach:
Updated! This is how it works now:

```
elixir[wm-circular-imports]% cat .iex.exs
IO.inspect :before
import_file ".iex.exs"
IO.inspect :after
```

```
elixir[wm-circular-imports]% iex
Erlang/OTP 22 [erts-10.4.4] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [hipe]

Interactive Elixir (1.10.0-dev) - press Ctrl+C to exit (type h() ENTER for help)
warning: path /Users/wojtek/src/elixir/.iex.exs was already imported, circular file imports are not allowed

:before
:after
iex(1)>
```

lib/iex/lib/iex/helpers.ex
```diff
@@ -1068,7 +1068,17 @@ defmodule IEx.Helpers do
     path = Path.expand(path)
 
     if not optional? or File.exists?(path) do
-      path |> File.read!() |> Code.string_to_quoted!(file: path)
+      if imported_paths = Process.get(:iex_imported_paths) do
+        if path in imported_paths do
+          message = "path #{path} was already imported, circular file imports are not allowed"
```
wojtekmach: I think the message should say something like "skipping path #{path} because it was already imported" but not sure that'd be clear enough. Any ideas?

josevalim: Maybe?

> "path #{path} was already imported, skipping circular file imports"

What you wrote also sounds good.

fertapric:
Thanks @wojtekmach! 💜 

Btw, more info about that "sticky dir" error message: https://github.com/elixir-lang/elixir/pull/8186 (it's on my TODO list)

PR #9365, opened 2019-09-23
Fix truncate docs for millisecond precision

RBangel:
Single character fix marking `:millisecond` as an atom to match `:second` and `:microsecond`.


fertapric:
Thanks @RBangel! 💜

PR #9366, opened 2019-09-23
Adds calendar.parse_date! to support calendar independent sigil_D

kipcole9:
This is an initial draft of an implementation to have sigils ~D, ~T, ~N and ~U be calendar independent. The intent is to capture comments and suggestions on the implementation approach since it will need to be replicated for `~T`, `~U` and `~N`.

### Open item

Based upon the discussions at [Elixir Forum](https://elixirforum.com/t/proposal-allow-calendar-to-be-specified-in-d-u-n-and-t-sigils/23156/32) the current implementation splits the sigil string on `\s` and if the last entry starts with a latin capital letter in the range `?A..?Z` it treats it as a calendar module name.

On reviewing tests I believe this may not be the right approach. From the `datetime` test suite we see:
```
assert inspect(datetime) == "2000-02-29 23:00:07 BRM Calendar.Holocene"
```
In attempting to parse this in the current implementation of `sigil_D/2` it would be ok. But if instead there was an attempt to parse `~U[2000-02-29 23:00:07 BRM]"` this would fail since `BRM` would be interpreted as calendar module.

Suggestions on how to format so that the calendar can always be unambiguously identified are welcome.

### Expanding a calendar module alias

There is also one failing test (commented out for now) that tests expanding a calendar alias in the sigil string. Calling `Macro.escape/2` isn't returning the results I expected so further investigation is required.

### Commit summary

1. Adjusts Kernel.sigil_D/2 to call calendar.parse_date!/1
2. Adds the callback parse_date!/1 to the Calendar behaviour
3. Implements parse_date!/1 in Calendar.ISO
4. Implements parse_date!/1 in Calendar.Holocene
5. Adds test for sigil_D/2 with Calendar.Holocene
6. Makes minor adjustment to Calendar.Holocene tests to
better align inspect and parse outputs/inputs consistent
with Calendar.ISO date padding

### DIFF
--- a/lib/elixir/lib/calendar.ex
+++ b/lib/elixir/lib/calendar.ex
@@ -309,6 +309,11 @@ defmodule Calendar do
               Inspect.Opts.t()
             ) :: Inspect.Algebra.t()
 
+  @doc """
+  Implements date string parsing in support of `Kernel.sigil_D/2~
+  """
+  @callback parse_date!(String.t()) :: {year(), month(), day()} | no_return()
+
   # General Helpers
 
   @doc """
@@ -360,5 +365,29 @@ defmodule Calendar do
   @spec get_time_zone_database() :: time_zone_database()
   def get_time_zone_database() do
     Application.get_env(:elixir, :time_zone_database, Calendar.UTCOnlyTimeZoneDatabase)
+  end
+
+  @doc false
+  def split_date_time_string(string, caller) do
+    parts = String.split(string, " ", trim: true)
+
+    case List.last(parts) do
+      <<c::utf8, _rest::binary>> = calendar when c in ?A..?Z ->
+        calendar =
+          Module.concat(
+            [calendar]
+            |> Macro.expand(caller)
+          )
+
+        string =
+          parts
+          |> Enum.slice(0..-2)
+          |> Enum.join(" ")
+
+        {string, calendar}
+
+      _other ->
+        {string, Calendar.ISO}
+    end
   end
 end
--- a/lib/elixir/lib/calendar/iso.ex
+++ b/lib/elixir/lib/calendar/iso.ex
@@ -820,6 +820,35 @@ defmodule Calendar.ISO do
     "~T[" <> Calendar.ISO.time_to_string(hour, minute, second, microsecond) <> "]"
   end
 
+  @doc """
+  Implements date parsing in support of `Kernel.sigil_D/2` for
+  this calendar
+  """
+  @doc since: "1.10.0"
+  @impl true
+  @spec parse_date!(String.t()) ::
+          {
+            Calendar.year(),
+            Calendar.month(),
+            Calendar.day()
+          }
+          | no_return()
+
+  def parse_date!(<<?-, rest::binary>>) do
+    {year, month, day} = parse_date!(rest)
+    {-year, month, day}
+  end
+
+  def parse_date!(<<rest::binary>>) do
+    case Date.raw_from_iso8601(rest) do
+      {_year, _month, _day} = raw_date ->
+        raw_date
+
+      {:error, reason} ->
+        raise ArgumentError, Date.parse_error(rest, reason)
+    end
+  end
+
   defp offset_to_string(0, 0, "Etc/UTC", _format), do: "Z"
 
   defp offset_to_string(utc, std, _zone, format) do
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -5035,7 +5035,10 @@ defmodule Kernel do
   defmacro sigil_D(date_string, modifiers)
 
   defmacro sigil_D({:<<>>, _, [string]}, []) do
-    Macro.escape(Date.from_iso8601!(string))
+    {string, calendar} = Calendar.split_date_time_string(string, __CALLER__)
+    {year, month, day} = calendar.parse_date!(string)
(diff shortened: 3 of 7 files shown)


lib/elixir/test/elixir/calendar/holocene.exs
```diff
@@ -24,12 +24,12 @@ defmodule Calendar.Holocene do
 
   @impl true
   def date_to_string(year, month, day) do
-    "#{year}-#{month}-#{day} (HE)"
+    "#{year}-#{pad(month)}-#{pad(day)} #{inspect(__MODULE__)}"
```
wojtekmach: Here and in other places you can use the following. Even though Date.new etc doesn't support year > 9999, the following function happens to:

```suggestion
    Calendar.ISO.date_to_string(year, month, day) <> " " <> inspect(__MODULE__)
```

lib/elixir/lib/calendar/iso.ex
```diff
@@ -820,6 +820,35 @@ defmodule Calendar.ISO do
...
+  """
+  @doc since: "1.10.0"
+  @impl true
+  @spec parse_date!(String.t()) ::
+          {
+            Calendar.year(),
+            Calendar.month(),
+            Calendar.day()
+          }
+          | no_return()
```
wojtekmach: we should use `no_return` only if the function _never_ returns.

Also, since it's a callback I think we tend to just leave it with `@impl true` (and with neither `@doc` nor `@spec`) but not sure if that's a strict rule.

lib/elixir/lib/calendar.ex
```diff
@@ -309,6 +309,11 @@ defmodule Calendar do
               Inspect.Opts.t()
             ) :: Inspect.Algebra.t()
 
+  @doc """
+  Implements date string parsing in support of `Kernel.sigil_D/2~
```
wojtekmach:
```suggestion
  Implements date string parsing in support of `Kernel.sigil_D/2`.
```

lib/elixir/lib/kernel.ex
```diff
@@ -5035,7 +5035,10 @@ defmodule Kernel do
   defmacro sigil_D(date_string, modifiers)
 
   defmacro sigil_D({:<<>>, _, [string]}, []) do
-    Macro.escape(Date.from_iso8601!(string))
+    {string, calendar} = Calendar.split_date_time_string(string, __CALLER__)
+    {year, month, day} = calendar.parse_date!(string)
+    {:ok, date} = Date.new(year, month, day, calendar)
```
wojtekmach: this results in:

```elixir
iex(1)> ~D[2019-01-0 Calendar.Holocene]
** (MatchError) no match of right hand side value: {:error, :invalid_date}
```

it would be good to have a similar error like here:

```elixir
iex(2)> ~D[2019-01-0]
** (ArgumentError) cannot parse "2019-01-0" as date, reason: :invalid_format
```

lib/elixir/test/elixir/calendar/date_test.exs
```diff
@@ -103,4 +105,14 @@ defmodule DateTest do
     assert Date.diff(date1, date2) == -13
     assert Date.diff(date2, date1) == 13
   end
+
+  test "sigil_D for a calendar other than Calendar.ISO" do
+    assert %Date{year: 10001, month: 1, day: 1, calendar: Calendar.Holocene} ==
+             ~D[10001-01-01 Calendar.Holocene]
```
wojtekmach: nitpick: I believe the convention is to either do: `assert expected == actual` or `assert actual = expected`, does that make sense? In this case it would be:

```elixir
assert ~D[10001-01-01 Calendar.Holocene] ==
         %Date{year: 10001, month: 1, day: 1, calendar: Calendar.Holocene}
```

probably not a strict rule either, just something I tend to see elsewhere.

wojtekmach:
> But if instead there was an attempt to parse ~U[2000-02-29 23:00:07 BRM]" this would fail since BRM would be interpreted as calendar module.

it should raise the same error as this:

```
iex> ~U[2019-01-01 09:00:00]
** (ArgumentError) cannot parse "2019-01-01 09:00:00" as UTC datetime, reason: :missing_offset
```

for ~U, the time part must be followed with  either `Z` or `+00:00`. So there's no ambiguity!

lib/elixir/lib/calendar.ex
```diff
@@ -361,4 +366,28 @@ defmodule Calendar do
...
+  def split_date_time_string(string, caller) do
+    parts = String.split(string, " ", trim: true)
+
+    case List.last(parts) do
+      <<c::utf8, _rest::binary>> = calendar when c in ?A..?Z ->
+        calendar =
+          Module.concat(
+            [calendar]
+            |> Macro.expand(caller)
+          )
```
wojtekmach: I believe we can figure out the alias like this:

```elixir
module = Module.concat([calendar])

case Keyword.fetch(caller.aliases, module) do
  {:ok, module} -> module
  :error -> module
end
```

that seems to work when I tested it, but not sure if that's totally safe!

josevalim:
Hi @kipcole9! This looks great and I got some ideas but I think the best way to try them out is by putting them on "paper". So please hold this a bit and I will send a PR we can use as template.

josevalim:
@kipcole9 nah, it is mostly that I am not precisely sure about the best way to go without trying some of those ways myself. :)

josevalim:
Closing this in favor of #9367. :heart:

PR #9367, opened 2019-09-24
Add parse_date/parse_naive_datetime/parse_utc_datetime to Calendar behaviour

josevalim:
This is the first step in the extended sigils.

It is similar to what @kipcole9 had but I moved the whole `from_iso8601` code to the `Calendar.ISO` module. Also, the idea is that the parsing functions will be counterparts to the `_to_string` functions, so we don't need to have both `_to_string` and `_inspect`. Once we inspect each date, we will automatically inspect them as:

    "#{calendar.type_to_string(...)} #{calendar}"

Without special handling for Calendar.ISO (except by hiding the Calendar.ISO part).

I still need to do a couple things to finish this:

  * [x] Add parse_time to the contract too
  * [x] Change the sigils to use the calendar callback
  * [x] Support calendars in the sigils (both when building and when parsing)

Initial feedback is welcome before we move forward, especially on the function names.

/cc @kipcole9 @wojtekmach

### DIFF
--- a/lib/elixir/lib/calendar.ex
+++ b/lib/elixir/lib/calendar.ex
@@ -284,6 +284,8 @@ defmodule Calendar do
   @doc """
   Parses the string representation for a naive datetime returned by
   `c:naive_datetime_to_string/7` into a naive-datetime-tuple.
+
+  The given string may contain a timezone offset but it is ignored.
   """
   @callback parse_naive_datetime(String.t()) ::
               {:ok, {year, month, day, hour, minute, second, microsecond}}
--- a/lib/elixir/lib/calendar.ex
+++ b/lib/elixir/lib/calendar.ex
@@ -265,7 +265,6 @@ defmodule Calendar do
   """
   @callback valid_time?(hour, minute, second, microsecond) :: boolean
 
-
   @doc """
   Parses the string representation for a time returned by `c:time_to_string/4`
   into a time-tuple.
--- a/lib/elixir/lib/calendar.ex
+++ b/lib/elixir/lib/calendar.ex
@@ -265,6 +265,15 @@ defmodule Calendar do
   """
   @callback valid_time?(hour, minute, second, microsecond) :: boolean
 
+
+  @doc """
+  Parses the string representation for a time returned by `c:time_to_string/4`
+  into a time-tuple.
+  """
+  @callback parse_time(String.t()) ::
+              {:ok, {hour, minute, second, microsecond}}
+              | {:error, atom}
+
   @doc """
   Parses the string representation for a date returned by `c:date_to_string/3`
   into a date-tuple.
--- a/lib/elixir/lib/calendar.ex
+++ b/lib/elixir/lib/calendar.ex
@@ -265,6 +265,33 @@ defmodule Calendar do
   """
   @callback valid_time?(hour, minute, second, microsecond) :: boolean
 
+  @doc """
+  Parses the string representation for a date returned by `c:date_to_string/3`
+  into a date-tuple.
+  """
+  @callback parse_date(String.t()) ::
+              {:ok, {year, month, day}}
+              | {:error, atom}
+
+  @doc """
+  Parses the string representation for a naive datetime returned by
+  `c:naive_datetime_to_string/7` into a naive-datetime-tuple.
+  """
+  @callback parse_naive_datetime(String.t()) ::
+              {:ok, {year, month, day, hour, minute, second, microsecond}}
+              | {:error, atom}
+
+  @doc """
+  Parses the string representation for a UTC datetime returned by
+  `c:datetime_to_string/11` into a datetime-tuple.
+
+  The effective `Datetime` created by the result of this function
+  will be in UTC with the given `utc_offset`.
+  """
+  @callback parse_utc_datetime(String.t()) ::
+              {:ok, {year, month, day, hour, minute, second, microsecond}, utc_offset}
+              | {:error, atom}
+
   # General Helpers
 
   @doc """
--- a/lib/elixir/lib/calendar.ex
+++ b/lib/elixir/lib/calendar.ex
@@ -291,11 +291,11 @@ defmodule Calendar do
               | {:error, atom}
 
   @doc """
-  Parses the string representation for a UTC datetime returned by
+  Parses the string representation for a datetime returned by
   `c:datetime_to_string/11` into a datetime-tuple.
 
-  The effective `Datetime` created by the result of this function
-  will be in UTC with the given `utc_offset`.
+  The returned datetime must be in UTC. The original `utc_offset`
+  it was written in must be returned in the result.
   """
   @callback parse_utc_datetime(String.t()) ::
               {:ok, {year, month, day, hour, minute, second, microsecond}, utc_offset}
--- a/lib/elixir/lib/calendar/date.ex
+++ b/lib/elixir/lib/calendar/date.ex
@@ -748,12 +748,11 @@ defmodule Date do
   end
 
   defimpl Inspect do
-    def inspect(%{calendar: Calendar.ISO, year: year, month: month, day: day}, _) do
-      "~D[" <> Calendar.ISO.date_to_string(year, month, day) <> "]"
+    def inspect(%{calendar: calendar, year: year, month: month, day: day}, _) do
+      "~D[" <> calendar.date_to_string(year, month, day) <> suffix(calendar) <> "]"
     end
 
-    def inspect(date, opts) do
-      Inspect.Any.inspect(date, opts)
-    end
+    defp suffix(Calendar.ISO), do: ""
+    defp suffix(calendar), do: " " <> inspect(calendar)
   end
 end
--- a/lib/elixir/lib/calendar/date.ex
+++ b/lib/elixir/lib/calendar/date.ex
@@ -263,30 +263,9 @@ defmodule Date do
 
   """
   @spec from_iso8601(String.t(), Calendar.calendar()) :: {:ok, t} | {:error, atom}
-  def from_iso8601(string, calendar \\ Calendar.ISO)
-
-  def from_iso8601(<<?-, rest::binary>>, calendar) do
-    with {:ok, %{year: year} = date} <- raw_from_iso8601(rest, calendar) do
-      {:ok, %{date | year: -year}}
-    end
-  end
-
-  def from_iso8601(<<rest::binary>>, calendar) do
-    raw_from_iso8601(rest, calendar)
-  end
-
-  [match_date, guard_date, read_date] = Calendar.ISO.__match_date__()
-
-  defp raw_from_iso8601(string, calendar) do
-    with unquote(match_date) <- string,
-         true <- unquote(guard_date) do
-      {year, month, day} = unquote(read_date)
-
-      with {:ok, date} <- new(year, month, day, Calendar.ISO) do
-        convert(date, calendar)
-      end
-    else
-      _ -> {:error, :invalid_format}
+  def from_iso8601(string, calendar \\ Calendar.ISO) do
+    with {:ok, {year, month, day}} <- Calendar.ISO.parse_date(string) do
+      convert(%Date{year: year, month: month, day: day}, calendar)
     end
   end
 
--- a/lib/elixir/lib/calendar/iso.ex
+++ b/lib/elixir/lib/calendar/iso.ex
@@ -251,16 +251,16 @@ defmodule Calendar.ISO do
       {:ok, {2015, 1, 23, 23, 50, 7, {0, 0}}, 0}
 
       iex> Calendar.ISO.parse_utc_datetime("2015-01-23T23:50:07.123+02:30")
-      {:ok, {2015, 1, 23, 23, 50, 7, {123000, 3}}, 9000}
+      {:ok, {2015, 1, 23, 21, 20, 7, {123000, 3}}, 9000}
 
       iex> Calendar.ISO.parse_utc_datetime("2015-01-23T23:50:07,123+02:30")
-      {:ok, {2015, 1, 23, 23, 50, 7, {123000, 3}}, 9000}
+      {:ok, {2015, 1, 23, 21, 20, 7, {123000, 3}}, 9000}
 
       iex> Calendar.ISO.parse_utc_datetime("-2015-01-23T23:50:07Z")
       {:ok, {-2015, 1, 23, 23, 50, 7, {0, 0}}, 0}
 
       iex> Calendar.ISO.parse_utc_datetime("-2015-01-23T23:50:07,123+02:30")
-      {:ok, {-2015, 1, 23, 23, 50, 7, {123000, 3}}, 9000}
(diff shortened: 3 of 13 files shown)


lib/elixir/lib/calendar.ex
```diff
@@ -265,6 +265,42 @@ defmodule Calendar do
...
+  into a date-tuple.
+  """
+  @callback parse_date(String.t()) ::
+              {:ok, {year, month, day}}
+              | {:error, atom}
+
+  @doc """
+  Parses the string representation for a naive datetime returned by
+  `c:naive_datetime_to_string/7` into a naive-datetime-tuple.
+  """
```
wojtekmach:
```suggestion

  The given string may contain a timezone offset but it is ignored.
  """
```

lib/elixir/lib/calendar.ex
```diff
@@ -265,6 +265,42 @@ defmodule Calendar do
...
+  @callback parse_naive_datetime(String.t()) ::
+              {:ok, {year, month, day, hour, minute, second, microsecond}}
+              | {:error, atom}
+
+  @doc """
+  Parses the string representation for a UTC datetime returned by
+  `c:datetime_to_string/11` into a datetime-tuple.
+
+  The effective `Datetime` created by the result of this function
```
wojtekmach:
```suggestion
  The effective `DateTime` created by the result of this function
```

lib/elixir/test/elixir/calendar/holocene.exs
```diff
@@ -109,9 +109,19 @@ defmodule Calendar.Holocene do
   defdelegate day_of_era(year, month, day), to: Calendar.ISO
 
   @impl true
-  def valid_date?(year, month, day) do
-    :calendar.valid_date(year, month, day)
-  end
+  defdelegate parse_time(string), to: Calendar.ISO
+
+  @impl true
+  defdelegate parse_date(string), to: Calendar.ISO
```
wojtekmach: it's probably not important _yet_, but:

```elixir
iex(4)> Calendar.ISO.parse_date("12019-01-01")
{:error, :invalid_format}
```

kipcole9:
José, I like this approach a lot and really wanted to do this but was concerned I would be chopping up the code too much. Happy to dive in an do any grunt work when you are comfortable for me to do so.

lib/elixir/lib/calendar.ex
```diff
@@ -265,6 +265,43 @@ defmodule Calendar do
...
+              | {:error, atom}
+
+  @doc """
+  Parses the string representation for a datetime returned by
+  `c:datetime_to_string/11` into a datetime-tuple.
+
+  The returned datetime must be in UTC. The original `utc_offset`
+  it was written in must be returned in the result.
+  """
+  @callback parse_utc_datetime(String.t()) ::
```
wojtekmach: this name makes me think that we can only "parse UTC date times" as in succeed exclusively for UTC timezone, and return error otherwise, e.g. `:non_utc_timezone` reason, same that `~U` returns. We of course handle non-UTC datetime strings here and we need them for `DateTime.from_iso8601`. Maybe the callback should be named just `parse_datetime` then? I think the current name is fine as is, especially because we do return the utc offset, but I thought I'd mention this anyway.

(Of course for all the other callbacks we have this nice property that, say, for all valid dates we basically have : `date == parse_date(date_to_string(date))` (I'm simplifying the actual interface here). This is true for all types other than non-UTC date times. Not that we can do anything about it, but perhaps worth repeating this point.)

josevalim:
Hi @kipcole9, I had a long flight, so used the time to work on the functionality, it should all be here. The only things missing are docs and tests for inspect on DateTime. Please review. :)
 
One thing I also noticed is that `~U[2010-10-10 12:34:56+01:00]` raises today because it has a non UTC offset but we could take the offset and convert it to UTC, like `DateTime.from_iso8601` does. But that's a discussion for another day. :)

lib/elixir/lib/kernel.ex
```diff
@@ -5105,21 +5127,63 @@ defmodule Kernel do
...
+  end
 
-      {:error, reason} ->
-        raise ArgumentError,
-              "cannot parse #{inspect(string)} as UTC datetime, reason: #{inspect(reason)}"
+  defp maybe_atomize_calendar(<<alias, _::binary>> = last_part, string)
+       when alias >= ?A and alias <= ?Z do
+    string = binary_part(string, 0, byte_size(string) - byte_size(last_part) - 1)
+    {String.to_atom("Elixir." <> last_part), string}
+  end
```
kipcole9: If I'm following this correctly, an alias wouldn't be expanded here. Meaning that:
```
defmodule MyModule do
  alias Calendar.Holocene
  ~D[12345-01-01 Holocene]
end
```
Would raise because `Holocene` wouldn't be expanded to `Calendar.Holocene`.  I think it would be good if an attempt to expand an alias was possible.

lib/elixir/lib/calendar.ex
```diff
@@ -265,6 +265,15 @@ defmodule Calendar do
   """
   @callback valid_time?(hour, minute, second, microsecond) :: boolean
 
+
+  @doc """
+  Parses the string representation for a time returned by `c:time_to_string/4`
+  into a time-tuple.
+  """
+  @callback parse_time(String.t()) ::
```
kipcole9: I'm very happy to see this since it makes this part of the public API for the first time (I think).  Date parts vary a lot amongst calendars, time parsing is commonly more uniform.

lib/elixir/lib/calendar/iso.ex
```diff
@@ -32,18 +32,17 @@ defmodule Calendar.ISO do
...
   @months_in_year 12
 
   # The ISO epoch starts, in this implementation,
   # with ~D[0000-01-01]. Era "1" starts
   # on ~D[0001-01-01] which is 366 days later.
   @iso_epoch 366
 
-  @doc false
-  def __match_date__ do
+  [match_date, guard_date, read_date] =
```
kipcole9: Would you consider making these date and time match/guard clauses part of the public API? I appreciate they are intended to support ISO8601 formats only but for a large class of business calendars that are also based upon the Gregorian calendar the date/time/datetime formats are the same - only the calendar is different from a parsing perspective. It would be great to leverage a consistent way to decompose the date/time/datetime strings.

lib/elixir/lib/calendar.ex
kipcole9: I like the symmetry of `parse_date`, `parse_time` and `parse_datetime`. And since an offset can be given this would still be a reasonable choice?

kipcole9:
José, I alway take joy from watching you write software. I like this a lot. It simplifies the behaviour as you say.  As a result calendar writers can also leverage parts of the `Calendar.ISO` implementation since they are public API.

If the date and time match clauses could become a public part of the API I'd be even happier :-)

I don't feel like I've contributed anything but noise to the calendar conversation over the last year but I'm super pleased with how this looks now. 

How about I update the `Holocene` calendar so that it parses 5 digit years instead of delegating to `Calendar.ISO`?  It's not much but I can do that tonight.

lib/elixir/lib/calendar.ex
josevalim: So the `utc` is not about the input, but the fact the output must be a datetime in UTC.

lib/elixir/lib/kernel.ex
josevalim: I don't want to support aliases because it means if we copy and paste the code, then the inspect representation won't work unless the alias is the same and that literally never happens with any other inspect representation. :)

lib/elixir/lib/calendar/iso.ex
josevalim: For other calendars, I would perhaps consider using `nimble_parsec`? That's literally from where I got this code from, except we can't depend on it from Elixir itself. :(

kipcole9: Definitely doable. My thinking was more that its such a common pattern for derivative Gregorian calendars that it would be DRY to reuse it. But I can understand the counter perspective as well.

josevalim:
Thank you @kipcole9!

> I don't feel like I've contributed anything but noise to the calendar conversation over the last year but I'm super pleased with how this looks now.

This couldn't be further from the truth! You did contribute a lot of ideas (such as this one), insights and feedback to the overall calendar representations. You have been one of driving forces behind many of the calendar improvements and while I can contribute with code, I cannot contribute on these other areas on the same level as you.

> How about I update the Holocene calendar so that it parses 5 digit years instead of delegating to Calendar.ISO? It's not much but I can do that tonight.

Sounds good. I have restricted some of my tests to using only 4-digit dates, but it would be nice to support this and also change the tests to use 5-digit dates.

lib/elixir/lib/kernel.ex
kipcole9: Hmmm, my understanding was that the inspect format is always the full module name and therefore would happily parse on roundtrip because it wouldn't expand. But since its a string being parsed it could also be an unexpected side effect if expanded.

josevalim: Right, if you copy from the inspected result it would be the full name, but I am saying if you copy it from the source code. Basically, I expected `MyAlias` to be generally dependent on its module, but I wouldn't expect it to happen from inside a sigil (which are often seen as strings).

josevalim:
Docs pushed, this should be fully complete to go.

lib/elixir/lib/kernel.ex
kipcole9: Yes, I agree thats a better priority: Sigil's should be copyable without concern for context or side affects.

kipcole9:
Awesome stuff. I will update the date parser for Holocene to accept more than 5 digits in the next few hours but I don't think thats a reason to hold up merging this PR.

lib/elixir/lib/calendar/date.ex
```diff
@@ -748,12 +748,11 @@ defmodule Date do
   end
 
   defimpl Inspect do
```
kipcole9: On revisiting `Calendar.Holocene` I realised that with this change to the `Inspect/2` implementation there is now no API to emit a formatted string for a calendar that identifies itself - if we want to keep the ability to roundtrip which I believe we do.

For example: the new `calendar.to_string(date)`  contract emits an un-annotated string. Therefore is `"2000-01-01"` in `Calendar.ISO`, or `Calendar.Holocene`, or some other calendar?  Previously `to_string/1` returned "2000-01-01 (HE)" which I updated to "2000-01-01 Calendar.Holocene". The contract was that `Calendar.ISO` dates had no annotation but other calendars did.  This won't now work because `inspect/2` will add a duplicate calendar annotation.

I don't know what the right answer to this is. I really like the cleanliness of how `inspect/2` now works. But we lose the ability to `to_string/1` a date/time/date_time and be able to use that to directly disambiguate a date in a calendar.

wojtekmach: > For example: the new calendar.to_string(date) contract emits an un-annotated string. Therefore is "2000-01-01" in Calendar.ISO, or Calendar.Holocene, or some other calendar? Previously to_string/1 returned "2000-01-01 (HE)" which I updated to "2000-01-01 Calendar.Holocene". The contract was that Calendar.ISO dates had no annotation but other calendars did.

I don't think we lost anything with the new contract. If it makes sense for Holocene to keep the ` (HE)` suffix, that's still totally cool.

If we do `~D[2019-01-01 (HE) Calendar.Holocene]` then it will result in `Calendar.Holocene.parse_date("2019-01-01 (HE)")` which can look like this:

```elixir
defmodule Calendar.Holocene do
  def parse_date(string) do
    string = String.trim_trailing(string, " (HE)")
    Calendar.ISO.parse_date(string) # forgetting about 5 digit years for a sec!
  end

  def date_to_string(year, month, day) do
    Calendar.ISO.date_to_string(year, month, day) <> " (HE)"
  end
end
```

josevalim: I would also wonder if Holocene should really emit an annotated string. For example, would people using a Japanese calendar to see dates as "xyz Japanese Calendar"? `to_string` is always about how a non-programmer would interpret it.

wojtekmach: Yup, exactly. I think it's totally up to calendar implementer to add annotations, but I think that'd be rather uncommon.

kipcole9: José, yes, I agree with that having had a few coffees now.   And if localised formatting is required (as is a more likely scenario) then I have a lib for that.  Unannotated makes more sense.

Just noting that calendars aren't always about culture, often its about business and fiscal calendars which, whilst still Gregorian, are based upon different start dates. But the ambiguity in presentation should be resolved separately, not by annotating `Kernel.to_string/1` (ie I agree with. your comment).

Properly armed I'm now going to update and greatly simplify my implementations. Thanks to you and @wojtekmach for being patient through this.

Now if I can just work our a compelling proposal for weeks and quarters I think I'll move on from calendars as a happy person :-)

kipcole9:
Anything else required for this PR before merging José?

josevalim:
I am just waiting for a review from someone on the Elixir team. :)

lib/elixir/lib/kernel.ex
```diff
@@ -5057,16 +5090,44 @@ defmodule Kernel do
...
+      second: second,
+      microsecond: microsecond
+    )
   end
 
   @doc ~S"""
   Handles the sigil `~N` for naive date times.
 
+  By default, this sigil uses the built-in `Calendar.ISO`, which
+  requires naive datetimes to be written in the ISO8601 format:
```
fertapric: Nitpick:
```suggestion
  requires naive date times to be written in the ISO8601 format:
```

lib/elixir/lib/kernel.ex
```diff
@@ -5079,18 +5140,48 @@ defmodule Kernel do
...
+      second: second,
+      microsecond: microsecond
+    )
   end
 
   @doc ~S"""
   Handles the sigil `~U` to create a UTC `DateTime`.
 
+  By default, this sigil uses the built-in `Calendar.ISO`, which
+  requires naive datetimes to be written in the ISO8601 format:
```
fertapric: ?
```suggestion
  requires UTC date times to be written in the ISO8601 format:
```

lib/elixir/lib/calendar/iso.ex
```diff
@@ -73,6 +70,259 @@ defmodule Calendar.ISO do
...
+      iex> Calendar.ISO.parse_time("23:50:07A")
+      {:error, :invalid_format}
+      iex> Calendar.ISO.parse_time("23:50:07.")
+      {:error, :invalid_format}
+      iex> Calendar.ISO.parse_time("23:50:61")
+      {:error, :invalid_time}
+
+  """
+  @doc since: "1.10.0"
+  @impl true
```
fertapric: The possible error atoms are more constrained than `{:error, atom}`  in this specific implementation, should that be communicated via `@spec`?

josevalim: I would still keep it as a atom so we can add new formats in the future. From a typing perspective, adding new return types is a breaking change, so sticking with the shape is better than each individual entity.

josevalim:

PR #9368, opened 2019-09-25
Integrate Erlang's metadata and Elixir's metadata

hauleth:
Close erlef/eef-observability-wg#6

### DIFF
--- a/lib/logger/lib/logger.ex
+++ b/lib/logger/lib/logger.ex
@@ -447,36 +447,31 @@ defmodule Logger do
   @compile {:inline, __metadata__: 0}
 
   defp __metadata__ do
-    Process.get(@metadata) || {true, []}
+    Process.get(@metadata, true)
   end
 
   @doc """
   Alters the current process metadata according the given keyword list.
-
   This function will merge the given keyword list into the existing metadata,
   with the exception of setting a key to `nil`, which will remove that key
   from the metadata.
   """
   @spec metadata(metadata) :: :ok
   def metadata(keyword) do
-    {enabled?, metadata} = __metadata__()
-    Process.put(@metadata, {enabled?, into_metadata(keyword, metadata)})
-    :ok
+    case :logger.get_process_metadata() do
+      :undefined ->
+        :ok = :logger.set_process_metadata(filter_out_nils(keyword))
+
+      map when is_map(map) ->
+        merged = Enum.into(keyword, map)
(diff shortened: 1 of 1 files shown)


fertapric:
Elixir's Logger metadata allows to have duplicated keys, so this would be a backwards incompatible change. Said that, I'm not sure how common that is 🙂 

Thoughts?

josevalim:
Hi @hauleth, thanks for the PR! This also requires Erlang/OTP 21, which means we can’t merge it any sooner than the other PR. So maybe we should get all of the work done at once in that PR?

josevalim:
And to clarify, I don’t want to do things conditionally, because it may lead to subtle difference of behavior as mentioned by @fertapric. If there is a need to rush, I would rather merge the complete PR and drop support for 20 sooner.

lib/logger/lib/logger.ex
```diff
@@ -447,53 +447,50 @@ defmodule Logger do
...
 
-  defp into_metadata([], metadata), do: metadata
-  defp into_metadata(keyword, metadata), do: into_metadata(keyword, [], metadata)
-
-  defp into_metadata([{key, nil} | keyword], prepend, metadata) do
-    into_metadata(keyword, prepend, :lists.keydelete(key, 1, metadata))
-  end
+      map when is_map(map) ->
+        merged = Enum.into(keyword, map)
+        metadata = filter_out_nils(merged)
```
michalmuskala: Is the intention to also filter the existing metadata or just the new ones? If only the new one, we could possibly merge & filter in one go
```elixir
for {_k, v} = elem <- keyword, v != nil, into: map, do: elem
```

hauleth:
@fertapric while in theory having duplicated keys in metadata is possible, it is highly impropable, as almost any action on these will remove duplicates anyway.

lib/logger/lib/logger.ex
hauleth: @michalmuskala when sending data to the backend all nils will be removed anyway, so there is not much difference.

hauleth:
@josevalim I proposed it as a result of discussion during WG meeting where there was proposal to provide metadata integration sooner, before whole #9333 will be ready to merge. But in the end it is up to Elixir team how to handle that.

josevalim:
We have three months if we bring the release date early and nine months otherwise, so let's try to make it go the whole way in. :)

josevalim:
Thanks for the PR!

PR #9369, opened 2019-09-25
Make clear Stream.transform is *only* about shape transformation

bjornrud:
Was clear that it modifed stream shape.  *However*, we thought changing element return was also possible.

Took a little bit of tinkering, discussion with @tonyvanriet, and finally looking at implementation before we realized the accumulator was discarded.

Am happy to have the PR discarded or language changed.  Just want to prevent the same misunderstanding for future readers of the docs.

### DIFF
--- a/lib/elixir/lib/stream.ex
+++ b/lib/elixir/lib/stream.ex
@@ -809,13 +809,13 @@ defmodule Stream do
   many of the functions defined in this module. For example, we can implement
   `Stream.take(enum, n)` as follows:
 
-      iex> enum = 1..100
+      iex> enum = 1001..9999
       iex> n = 3
       iex> stream = Stream.transform(enum, 0, fn i, acc ->
       ...>   if acc < n, do: {[i], acc + 1}, else: {:halt, acc}
       ...> end)
       iex> Enum.to_list(stream)
-      [1, 2, 3]
+      [1001, 1002, 1003]
 
   """
   @spec transform(Enumerable.t(), acc, fun) :: Enumerable.t()
--- a/lib/elixir/lib/stream.ex
+++ b/lib/elixir/lib/stream.ex
@@ -793,7 +793,7 @@ defmodule Stream do
   end
 
   @doc """
-  Transforms an existing stream.
+  Transforms shape an existing stream.
 
   It expects an accumulator and a function that receives each stream element
   and an accumulator, and must return a tuple containing a new stream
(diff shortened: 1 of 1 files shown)


josevalim:
Hi @bjornrud!

I am not sure those clarifications would help generally. I don't believe we define what "shape" means anywhere, so introducing this word here, can cause confusion (what is "shape"?) or misunderstandings (we have different interpretations of it).

Regarding the other paragraph, everything in Elixir is immutable, so any kind of transformation indeed won't affect the source.

If is the confusion is regarding the accumulator, perhaps we can say at the end of the second paragraph that the accumulator is used during transformation and then discarded? Or is the confusion elsewhere?

Thank you!

bjornrud:
Hi @josevalim! 

Agreed adding in a new symbol is problematic and retract "shape".   

Saying the accumulator is discarded suffices.   Thinking on this more, I believe I also drew this conclusion based on the example.  As-is the example for `Stream.transform/4` accumulation coincidentally corresponds to the original `enum`.   

Would you be amenable to changing the example to not be indexed off of `1`, just so the final return in the example may not possibly be confused with the values of the accumulator?

```elixir
iex> enum = 5..105
iex> n = 3
iex> stream = Stream.transform(enum, 0, fn i, acc ->
...>   if acc < n, do: {[i], acc + 1}, else: {:halt, acc}
...> end)
iex> Enum.to_list(stream)
[5, 6, 7]
```

bjornrud:
I find myself struggling to figure out how to jump between talking about the accumulator in the anonymous function in the second paragraph, then back to the function return. 

The `Note:` itself seems similar to what I'm trying to do but has the additional problem of understanding context of `Enum.flat_map_reduce/3`.  Which is truly where I went wrong.  🤔 

So, if the example change works for you, will commit that, otherwise I withdraw and chalk this up to a reading comprehension failure of mine to follow multiple subject-changes.

josevalim:
> Would you be amenable to changing the example to not be indexed off of 1, just so the final return in the example may not possibly be confused with the values of the accumulator?

Yes, great idea! Maybe start from 11 or make it go from 101..999.

I have pushed some changes rewording the explanation. Let me know what you think and feel free to edit or expand on them.

bjornrud:
@josevalim Here ya go.  I took the liberty of moving to `1001` to avoid complicating the example with piping into `|> inspect(charlists: :as_lists)`.

josevalim:

PR #9370, opened 2019-09-25
Update Calendar.Holocene to_string/1 and parse_date/1

kipcole9:
This commit updates the test Calendar.Holocene module:
1. to_string/1 removes the "Calendar.Holocene" annotation
in the output string. This is somewhat
problematic because calling `to_string/1` on a date now no
longer emits anything which identifies what calendar it is in.

2. Implements a simple `parse_date/1` function to allow
basic testing of years >= 10_000 and <= -10_000

### DIFF
--- a/lib/elixir/test/elixir/calendar/date_test.exs
+++ b/lib/elixir/test/elixir/calendar/date_test.exs
@@ -13,6 +13,9 @@ defmodule DateTest do
     assert ~D[2000-01-01 Calendar.Holocene] ==
              %Date{calendar: Calendar.Holocene, year: 2000, month: 1, day: 1}
 
+    assert ~D[10001-01-01 Calendar.Holocene] ==
+             %Date{calendar: Calendar.Holocene, year: 10001, month: 1, day: 1}
+
     assert_raise ArgumentError,
                  ~s/cannot parse "2000-50-50" as Date for Calendar.ISO, reason: :invalid_date/,
                  fn -> Code.eval_string("~D[2000-50-50]") end
@@ -46,6 +49,12 @@ defmodule DateTest do
 
     date = %{~D[2000-01-01] | calendar: FakeCalendar}
     assert inspect(date) == "~D[1/1/2000 FakeCalendar]"
+  end
+
+  test "inspect/1 roundtrip to to_string/1" do
+    test_string = "~D[2000-01-01 Calendar.Holocene]"
+    {date, []} = Code.eval_string(test_string)
+    assert inspect(date) == test_string
   end
 
   test "compare/2" do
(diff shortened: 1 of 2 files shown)


lib/elixir/test/elixir/calendar/holocene.exs
```diff
@@ -78,50 +79,81 @@ defmodule Calendar.Holocene do
     )
   end
 
+  defp zero_pad(val, count) when val >= 0 do
+    num = Integer.to_string(val)
```
fieldinrain: I think you can use `String.pad_leading/3` to pad.

lib/elixir/test/elixir/calendar/holocene.exs
```diff
@@ -78,50 +79,81 @@ defmodule Calendar.Holocene do
...
+      {:ok, {year, month, day}}
+    else
+      {:error, :invalid_date}
+    end
+  end
 
   @impl true
-  defdelegate time_to_day_fraction(hour, minute, second, microsecond), to: Calendar.ISO
+  def valid_date?(year, month, day) when year >= 10_000 do
+    Calendar.ISO.valid_date?(year - 10_000, month, day)
```
josevalim: We can use `:calendar.valid_date`, which is not restricted to ISO dates. :)

lib/elixir/test/elixir/calendar/date_test.exs
```diff
@@ -48,6 +51,12 @@ defmodule DateTest do
     assert inspect(date) == "~D[1/1/2000 FakeCalendar]"
   end
 
+  test "inspect/1 roundtrip to to_string/1" do
+    test_string = "~D[2000-01-01 Calendar.Holocene]"
+    {date, []} = Code.eval_string(test_string)
+    assert inspect(date) == test_string
+  end
```
josevalim: Instead of adding new examples to the suite, let's just change all of the Holocene examples to include 5 digits. What do you think?

kipcole9:
Will change to `:calendar.valid_date?/1` change the examples to 5 digits.

The new case I added was actually just about round tripping - there didn't seem to be another case.  I know its not a contract to round trip, but it will be a common expectation. I can just delete it if you prefer.

lib/elixir/test/elixir/calendar/date_test.exs
kipcole9: Let me know if this should still come out? I adjusted all the `Calendar.Holocene` tests to use `years > 9999` but left this since I think its the only deliberately round trip example.

lib/elixir/test/elixir/calendar/holocene.exs
kipcole9: Done

lib/elixir/test/elixir/calendar/holocene.exs
kipcole9: Done, thanks for the comment.

lib/elixir/test/elixir/calendar/date_test.exs
josevalim: The round trip is a property of the calendar but  we are not interested in testing the calendar in this example, so we are fine without it. :)

josevalim:

PR #9378, opened 2019-10-02
Update docs for Kernel.SpecialForms.case/2

brettbeatty:
In the docs for `case/2`, there's a [section about variable scope](https://hexdocs.pm/elixir/Kernel.SpecialForms.html#case/2-variable-handling) that doesn't seem accurate to me:

> However, variables explicitly bound in the clause "body" are accessible from the outer context:
> ```elixir
> value = 7
>
> case lucky? do
>   false -> value = 13
>   true -> true
> end
>
> value
> #=> 7 or 13
> ```

When attempting the `lucky? = false` case in IEx, `value` in the outer context is `7`, not `13`:

```elixir
Erlang/OTP 22 [erts-10.4] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [hipe]

Interactive Elixir (1.9.1) - press Ctrl+C to exit (type h() ENTER for help)
iex(1)> lucky? = false
false
iex(2)> value = 7
7
iex(3)> case lucky? do
...(3)>   false -> value = 13
...(3)>   true -> true
...(3)> end
warning: variable "value" is unused (if the variable is not meant to be used, prefix it with an underscore)
  iex:4

13
iex(4)> value
7
```

I drafted some pretty minimal corrections. Would it be helpful to also include an example of rebinding `value` to the result?

### DIFF
--- a/lib/elixir/lib/kernel/special_forms.ex
+++ b/lib/elixir/lib/kernel/special_forms.ex
@@ -1752,13 +1752,17 @@ defmodule Kernel.SpecialForms do
 
       value = 7
 
-      case result do
-        :ok -> value = value + 1
-        :error -> value = value - 1
+      case lucky? do
+        false -> value = 13
+        true -> true
       end
 
       value
       #=> 7
+
+  In the example above, `value` is going to be `7` regardless of the value of
+  `lucky?`. The variable `value` bound in the clause and the variable `value`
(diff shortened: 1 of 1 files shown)


fertapric:
Thanks @brettbeatty! Nice catch! 💜

PR #9379, opened 2019-10-03
Improve :ets argument error messages

TheFirstAvenger:
First pass at improving the error messages surrounding failures in `:ets`.

Added blame logic to `ArgumentErrror` for `:ets` functions:

- `insert`
- `insert_new`
- `lookup`
- `lookup_element`.

Recognizes the following error states:
- Table missing
- Record too small (e.g. inserting a length 2 tuple into a table with `keypos` 3)
- Inserting into a `:private` or :`protected` table from a process other than the owner
- Reading from a `:private` table from a process other than the owner
- `lookup_element` when the record doesn't exist
- `lookup_element` when the record is smaller than the position specified

Logic taken from https://github.com/TheFirstAvenger/ets

### DIFF
--- a/lib/elixir/lib/exception.ex
+++ b/lib/elixir/lib/exception.ex
@@ -679,7 +679,7 @@ defmodule RuntimeError do
 end
 
 defmodule ArgumentError do
-  defexception message: "argument error"
+  defexception message: "argument error", source_pid: nil
 
   @impl true
   def blame(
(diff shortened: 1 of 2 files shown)


michalmuskala:
Overall, I think this is a nice addition, I had a similar idea some time ago, but never went though with it.

Unfortunately, there's one big issue, that makes this a risky thing in my mind - information being stale. It's entirely reasonable that a table might be restarted with different options between the time where the error is raise and when this additional information is collected. This could result in very misleading error messages. Additionally the blame callback will probably be frequently called from a different process than the one calling the `ets` functions - most likely with the shell, which again will lead to false reports. I'm not really sure there's a way around those issues.

josevalim:
We thought about the “timing” issue, which could be clarified with the
error message, but I didn’t consider the process one.

Maybe we could store the current pid in ArgumentError at the time we
assemble the exception? It is a cost but I am not very worried at the cost
of translating badarg to a struct, it should still be plenty fast. WDYT?
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

TheFirstAvenger:
@josevalim isn't the process that calls blame always the same process in which the error is raised? Or are you saying the argument error is built in a different process?

josevalim:
There are three steps:

1. Your raise an exception
2. The exception is normalized (badarg becomes ArgumentError)
3. The exception is blamed

The blame step is not guaranteed to be in the same process.
-- 


TheFirstAvenger:
I added `source_pid` to `ArgumentError`, populated it in `normalize`, and referenced it instead of `self()` in `blame_ets`:

https://github.com/elixir-lang/elixir/pull/9379/commits/034bf4e788c338bba2b267de83a18afe2af91604

michalmuskala:
The thing is that if the blame happens from another process, it's going to become much less useful, since most of it can be done only for public tables. For private or protected tables there's only very little things you could check from another process.

Additionally, I don't think the check for value can be vary misleading since the value can change in the meantime.

I'm also afraid of doing additional work in normalization because the code like below is pretty frequent and it would go through normalization, right? If so, anything we put there could suddenly be placed on a hot path in many applications slowing them down considerably.
```elixir
try
  :ets.lookup_element(:foo_table, :foo, 2)
rescue
  ArgumentError -> :default
end
```

To sum up, I really like this feature and I would love to see it happen, I just can't come up with a way that wouldn't be harmful in many cases. The only part that is somewhat safe would be to check for the protection level & the owner in the blame itself, though that would still be subject to a race condition with a table restart.

josevalim:
Those are very good points @michalmuskala. Indeed it seems we can't move forward with this.

@TheFirstAvenger I am so sorry that I prompted you for this PR, only to turn out we can't really do it in practice. I appreciate a lot that you took the time to explore its feasibility, thanks! :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

TheFirstAvenger:
One thought I had this morning was that this could be enabled by a compiler flag. If you got an ArgumentError from :ets without the flag enabled, we could include the message noting that adding the compiler flag will give more details. If an ArgumentError happens with the flag enabled, we would proceed as indicated in this PR. This would allow us to add documentation indicating these possible false-positives and the increased work that comes with the flag in the case of exceptions.

josevalim:
@TheFirstAvenger that's indeed an option, although we don't have anything like that in the language, so it would definitely be a first. From the top of my head, the biggest concern is that we would effectively change an error in `:ets.foo` from `:badarg` into something else with no indication in the source code. It is a global option and we usually stay away from those. In this case I prefer the approach in your ETS lib, as it is more explicit.

PR #9380, opened 2019-10-03
Allow for multiple paths for apps via apps_path config option

mononym:
First off, this definitely needs more documentation, but neither the `:apps_path` or `:apps` configuration options seem to be detailed anywhere but in `Mix.Project.apps_paths/1`.

Now, onto the thing. At work we want to organise our umbrella into a folder hierarchy that lets us more clearly separate different domains, such a service and library code. So, for example, we might want to do either of the following:

```
awesome_app1/ # umbrella app
  apps/services
  apps/libraries

awesome_app2/ # umbrella app
 services
 libraries
```

Currently mix doesn't really seem to play nice with this. You can only specify a single folder, and you can't nest things as there won't be any recursion. This code changes that.

Now you can specify any number of folders that can contain apps, with the only caveat being that you can *only* place Elixir applications in leaf folders. So you can do:

```
awesome_app1/ # umbrella app
  apps/services/foo # app
  apps/libraries/bar # app
```

But you cannot do:
```
awesome_app2/ # umbrella app
 apps/services/foo # app in a folder which contains a folder that is not an app
 apps/services/super_apps/bar # app nested in the same folder as an actual app
```

This code is backwards compatible.

### DIFF
--- a/lib/mix/lib/mix/project.ex
+++ b/lib/mix/lib/mix/project.ex
@@ -267,50 +267,96 @@ defmodule Mix.Project do
       if cache = Mix.ProjectStack.read_cache(key) do
         cache
       else
-        cache = config[:apps] |> umbrella_apps(apps_path) |> to_apps_paths(apps_path)
+        apps_paths = List.wrap(apps_path)
+        cache = config[:apps] |> umbrella_apps(apps_paths) |> to_apps_paths(apps_paths)
+
         Mix.ProjectStack.write_cache(key, cache)
       end
     end
   end
 
-  defp umbrella_apps(nil, apps_path) do
-    case File.ls(apps_path) do
-      {:ok, apps} -> Enum.map(apps, &String.to_atom/1)
-      {:error, _} -> []
+  defp umbrella_apps(nil, apps_paths) do
+    potential_app_paths = Enum.map(apps_paths, fn path -> File.ls(path) end)
+
+    Enum.all?(potential_app_paths, fn
+      {:ok, _} -> true
+      {:error, _} -> false
+    end)
+    |> if do
+      Enum.flat_map(potential_app_paths, fn {:ok, app_names} ->
+        Enum.map(app_names, &String.to_atom/1)
+      end)
+    else
+      []
     end
   end
 
-  defp umbrella_apps(apps, _apps_path) when is_list(apps) do
+  defp umbrella_apps(apps, _apps_paths) do
     apps
   end
 
-  defp to_apps_paths(apps, apps_path) do
+  defp to_apps_paths(apps, apps_paths) do
     for app <- apps,
-        path = path_with_mix_exs_otherwise_warn(app, apps_path),
+        path = path_with_mix_exs_otherwise_warn(app, apps_paths),
         do: {app, path},
         into: %{}
   end
 
-  defp path_with_mix_exs_otherwise_warn(app, apps_path) do
-    path = Path.join(apps_path, Atom.to_string(app))
+  defp path_with_mix_exs_otherwise_warn(app, apps_paths) do
+    paths_to_check = for path <- apps_paths, do: Path.join(path, Atom.to_string(app))
 
-    cond do
-      File.regular?(Path.join(path, "mix.exs")) ->
-        path
+    valid_paths =
+      Enum.filter(paths_to_check, fn path ->
+        # Mix file found
+        if File.regular?(Path.join(path, "mix.exs")) do
+          true
+          # Could be a directory or a stray file, doesn't matter at this point
+        else
+          false
+        end
+      end)
 
-      File.dir?(path) ->
-        Mix.shell().error(
-          "warning: path #{inspect(Path.relative_to_cwd(path))} is a directory but " <>
-            "it has no mix.exs. Mix won't consider this directory as part of your " <>
-            "umbrella application. Please add a \"mix.exs\" or set the \":apps\" key " <>
-            "in your umbrella configuration with all relevant apps names as atoms"
+    case length(valid_paths) do
+      0 ->
+        Enum.each(paths_to_check, fn path ->
+          if File.dir?(path) do
+            Mix.shell().error(
+              "warning: path #{inspect(Path.relative_to_cwd(path))} is a directory but " <>
+                "it has no mix.exs. Mix won't consider this directory as part of your " <>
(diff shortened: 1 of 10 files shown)


josevalim:
Unfortunately the option is named apps_path, so it would be awkward to suddenly start accepting a list. Plus you will have issues with the :in_umbrella option for dependencies, which won’t know how to traverse across different folders, so I am a bit skeptical about supporting this behavior, sorry. :(

josevalim:
One option is for you to do this:

    /
      libraries/
      services/

And have an umbrella app inside each of them, using regular path dependencies between apps on different umbrellas. This is the best approach as deps won’t work Across different umbrellas folders anyway.

mononym:
Didn't know about the problem with `:in_umbrella`.

Is the code at that point not able to call `Mix.Project.apps_path/1` to get the actual paths to traverse and extract the right one in `Mix.SCM.Path/accepts_options/2`?

And as far as the naming goes, I suppose it would be too much to ask for another option like `apps_paths` and behind the scenes you could make them behave the same.

josevalim:
An app inside the umbrella doesn’t know it is inside the umbrella, this is
by design because we want it to run in isolation. So I don’t see how this
feature could possibly work. :(
-- 


mononym:
If your disagreement is technical and I can't find a way to make it work reasonably, sanely, and without lowering the quality of the code base I'm fully prepared to agree with you. But if your argument against the feature is that it breaks the philosophy of being able to run anywhere I'm going to have to respectfully disagree.

This is the basic, out of the box, mix generated `mix.exs` file for an umbrella app:
```
  def project do
    [
      app: :bar,
      version: "0.1.0",
      build_path: "../../_build",
      config_path: "../../config/config.exs",
      deps_path: "../../deps",
      lockfile: "../../mix.lock",
      elixir: "~> 1.9",
      start_permanent: Mix.env() == :prod,
      deps: deps()
    ]
  end
```

You cannot run that anywhere except inside of an umbrella application. If you want to pull it out, you change configuration.

This configuration option I'm talking about only exists in the context *of an umbrella app*. You do not use `apps_path` or `:in_umbrella` *except* in the case of an umbrella app. And this configuration goes into the `mix.exs` file that is at the root of the umbrella app, not an application inside of the umbrella.

What I'm talking about is a transparent, to the applications inside of an umbrella, change. Heck, it's effectively non-existent to anyone using the default mix generated umbrella applications and their respective apps.

If you tell me that from `Mix.SCM.Path.accepts_options/2` around line 23 that I cannot tease the proper relative/absolute path out of the code then that's absolutely a killer.

Otherwise, I would love to try to find a middle ground that works. Because while you're correct that doing two umbrella apps in the same repo is a possibility, it would require changes to build scripts on our end and would bloat the dependency configuration by forcing an explicit path rather than a simple atom for configuration for anyone wanting to do this.

I agree about the naming potentially being confusing. It would be a simple thing to allow for an `:apps_paths` option which overrules `:apps_path` if it is present. I would be happy to add all the documentation needed around these options and also for the `:apps` configuration options if you just point me to where they belong.

I understand it may seem like I'm being pushy for such a simple seeming thing, but it would take a lot more developer time at work to make the suggested changes if we want to do anything but have a flat list of applications in a single repo. And I don't think I can swing that. For everyone else, frankly including me for a personal project I'm starting, who are choosing to explicitly work in an umbrella I would just like to see a little more flexibility...if technically possible.

josevalim:
> This is the basic, out of the box, mix generated mix.exs file for an umbrella app:

But you can delete the umbrella mix.exs and everything still works. That’s my point, all of the dependencies in the child are explicitly laid out. There is nothing that automatically changes behavior because we are inside an umbrella. Therefore, there is no way for in_umbrella to magically know it now needs to look at two different places.

> If you tell me that from Mix.SCM.Path.accepts_options/2 around line 23 that I cannot tease the proper relative/absolute path out of the code then that's absolutely a killer.

Exactly, there isn’t a way (that I know of that wouldn’t change today semantics).

mononym:
> Exactly, there isn’t a way (that I know of that wouldn’t change today semantics).

I hear you and agree on this point. I can't see a way to make that work without changing the whole process around how/when SCMs are run and get their data. And that's a non-starter.

> But you can delete the umbrella mix.exs and everything still works...

Well, yes, after you change the `:in_umbrella` reference to something explicit, because once you're out of the umbrella app the relative dependency isn't likely to be there anymore. And you have to change the `mix.exs` file in the nested umbrella app itself to not reference relative folders a couple levels up.

But I absolutely agree that the relative `:in_umbrella` does not work for this.

So, would you be willing to hear me out about one last possibility? And before you read, know that I haven't tried to test this out yet. I will as soon as I get a chance, but wanted to run it by you to see what you thought.

What if everything remains as it is with respect to `:in_umbrella` and `apps_path`. In this scenario even if I wanted to have an umbrella app, with nested apps in the custom folder layout I'm describing, I could not reference the dependencies even with an absolute path, ie: `path: "../services/foo" if that other app lives in the umbrella too. In fact it wouldn't even work as Mix would complain about folders not being applications since they have no `mix.exs` file.

Mix, in the context of the Umbrella, doesn't know where to even *look* for all of the apps.

What if we add a new option that is `apps_paths` that does basically what I've done already (and maybe even warns/raises if you try to use both apps_path and apps_paths options, or could transparently concat the values together...or whatever), this would let the umbrella application know *where* all of the apps are it needs to look for. No other changes.

If you want to reference *anything* that isn't an immediate sibling you need to specify a path via the `path` option. This is functionally the same as today, since `:in_umbrella` is just `path: "../#{app}"`.

`:in_umbrella` would work just the same as always. `:path` would work the same as always. And the only change would be that Mix, in the context of an Umbrella app, can look in multiple directories to build up the list of apps.

As far as I can see, the only place where `:in_umbrella` is used is in one SCM callback which means that everything else has to go off of the already compiled map of `app => path`, so this would just...work. I think.

Do you think this could work if the assumption is, and docs say, that if you're going to do anything other than having immediate siblings you have to reference them with an absolute path?

josevalim:
Sorry but I am really not inclined to add an umbrella feature that is incompatible with in_umbrella because it will surely look inconsistent and cause confusion down the road. I still believe my approach with two umbrellas at the root is the best, it will work exactly as you propose. The only downside is that you have to run “mix test” in two directories in CI and occasionally during dev. But it can also provide some welcome segregation between code too.

PR #9381, opened 2019-10-04
Improve MapSet comparison performance

sorentwo:
This changes the implementation of none_in?/2 (and a few other functions) from `match?` or a pin match to the :erlang.is_map_set/2 BIF. For `disjoint?/2` the result is 1.08x faster, according to benchmarks. Due to the use of `is_map_key` this change only works in Erlang 21.0+.

Benchmark using MapSets with 100 random keys, checking for a worst case match of the 100th key:

```
Operating System: macOS
CPU Information: Intel(R) Core(TM) i5-7360U CPU @ 2.30GHz
Number of Available Cores: 4
Available memory: 8 GB
Elixir 1.9.1
Erlang 22.0

warmup: 1 s
time: 4 s
memory time: 0 ns
parallel: 1
inputs: none specified
Estimated total run time: 10 s

Benchmarking CustomMapSet...
Benchmarking MapSet...

CustomMapSet        3.20 M      312.87 ns  ±9070.45%           0 ns        1000 ns
MapSet              2.95 M      338.53 ns  ±8711.10%           0 ns        1000 ns

Comparison:
CustomMapSet        3.20 M
MapSet              2.95 M - 1.08x slower +25.66 ns
```

I also tested this with a `:maps.iterator/1` and `:maps.next/3` based version, but the results were a fair bit slower with a medium sized set:

```
Name                        ips        average  deviation         median         99th %
CustomMapSet             3.08 M      324.89 ns  ±8159.84%           0 ns        1000 ns
MapSet                   2.17 M      461.64 ns  ±4836.36%           0 ns        1000 ns
```

For a larger set using `iterator` may be worth it due to the constant memory usage.

Edit: I made some modifications to the benchmarks and checked with a larger set of 1000 items instead of 100. There is a much bigger improvement with an iterator:

```
CustomMapSet        1.92 M        0.52 μs  ±4863.44%           0 μs           1 μs
MapSet             0.127 M        7.87 μs   ±107.59%           7 μs          27 μs

Comparison:
CustomMapSet        1.92 M
MapSet             0.127 M - 15.11x slower +7.35 μs
```

### DIFF
--- a/lib/elixir/lib/map_set.ex
+++ b/lib/elixir/lib/map_set.ex
@@ -200,16 +200,10 @@ defmodule MapSet do
     |> none_in?(map2)
   end
 
-  defp none_in?([], _) do
-    true
-  end
+  @compile {:inline, [none_in?: 2]}
 
-  defp none_in?([key | rest], map2) do
-    case map2 do
-      %{^key => _} -> false
-      _ -> none_in?(rest, map2)
-    end
-  end
+  defp none_in?([], _), do: true
+  defp none_in?([key | rest], map2), do: not :erlang.is_map_key(key, map2) and none_in?(rest, map2)
 
   @doc """
   Checks if two sets are equal.
--- a/lib/elixir/lib/map_set.ex
+++ b/lib/elixir/lib/map_set.ex
@@ -158,7 +158,8 @@ defmodule MapSet do
       when map_size(map1) < map_size(map2) * 2 do
     map =
       map1
-      |> Map.keys()
+      |> :maps.iterator()
+      |> :maps.next()
       |> filter_not_in(map2, [])
 
     %MapSet{map: map}
@@ -171,13 +172,13 @@ defmodule MapSet do
     %{map_set | map: Map.drop(map1, Map.keys(map2))}
   end
 
-  defp filter_not_in([], _map2, acc), do: :maps.from_list(acc)
+  defp filter_not_in(:none, _map2, acc), do: :maps.from_list(acc)
 
-  defp filter_not_in([key | rest], map2, acc) do
+  defp filter_not_in({key, _val, iter}, map2, acc) do
     if :erlang.is_map_key(key, map2) do
-      filter_not_in(rest, map2, acc)
+      filter_not_in(:maps.next(iter), map2, acc)
     else
-      filter_not_in(rest, map2, [{key, @dummy_value} | acc])
+      filter_not_in(:maps.next(iter), map2, [{key, @dummy_value} | acc])
     end
   end
 
@@ -197,14 +198,18 @@ defmodule MapSet do
     {map1, map2} = order_by_size(map1, map2)
 
     map1
-    |> Map.keys()
+    |> :maps.iterator()
+    |> :maps.next()
     |> none_in?(map2)
   end
 
   @compile {:inline, [none_in?: 2]}
 
-  defp none_in?([], _), do: true
-  defp none_in?([key | rest], map2), do: not :erlang.is_map_key(key, map2) and none_in?(rest, map2)
+  defp none_in?(:none, _), do: true
+
+  defp none_in?({key, _val, iter}, map2) do
+    not :erlang.is_map_key(key, map2) and none_in?(:maps.next(iter), map2)
+  end
 
   @doc """
   Checks if two sets are equal.
@@ -227,7 +232,7 @@ defmodule MapSet do
   # Elixir v1.5 change the map representation, so on
   # version mismatch we need to compare the keys directly.
   def equal?(%MapSet{map: map1}, %MapSet{map: map2}) do
-    map_size(map1) == map_size(map2) and map_subset?(Map.keys(map1), map2)
+    map_size(map1) == map_size(map2) and all_in?(map1, map2)
   end
 
   @doc """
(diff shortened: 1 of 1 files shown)


josevalim:
The difference for large sets happen because of the upfront cost in getting all keys. For a fair comparison, you also need to measure the cost of these functions when both sets are equal, that’s our worst case scenario. Can you please do so?

In any case, I believe these changes are beneficial. I have recently submitted a patch to Erlang mapsets with similar changes, and we should probably do the same here. However, we will need to wait until we drop Erlang 20. We are theoretically waiting for v1.10 to be out but we can bring it earlier depending on @hauleth’s work on the logger integration.

sorentwo:
Totally understood that we need to drop Erlang 20 first. I thought (hoped?) it was time for that based on the logger integration work and a few other issues recently.

As requested, I ran a more detailed benchmark with worst case inputs of 10 through 1 million

```
Operating System: macOS
CPU Information: Intel(R) Core(TM) i5-7360U CPU @ 2.30GHz
Number of Available Cores: 4
Available memory: 8 GB
Elixir 1.9.0
Erlang 22.0

warmup: 2 s
time: 5 s
memory time: 0 ns
parallel: 1
inputs: 10, 100, 1000, 10000, 100000, 1000000
Estimated total run time: 1.40 min

Benchmarking 1.9.0 with input 10...
Benchmarking 1.9.0 with input 100...
Benchmarking 1.9.0 with input 1000...
Benchmarking 1.9.0 with input 10000...
Benchmarking 1.9.0 with input 100000...
Benchmarking 1.9.0 with input 1000000...
Benchmarking Custom with input 10...
Benchmarking Custom with input 100...
Benchmarking Custom with input 1000...
Benchmarking Custom with input 10000...
Benchmarking Custom with input 100000...
Benchmarking Custom with input 1000000...

##### With input 10 #####
Name             ips        average  deviation         median         99th %
1.9.0         1.39 M        0.72 μs ±31532.56%           0 μs           1 μs
Custom        0.95 M        1.05 μs ±25070.78%           0 μs           1 μs

Comparison:
1.9.0         1.39 M
Custom        0.95 M - 1.46x slower +0.33 μs

##### With input 100 #####
Name             ips        average  deviation         median         99th %
Custom      364.37 K        2.74 μs ±15511.18%           0 μs           1 μs
1.9.0       266.38 K        3.75 μs ±12572.16%           0 μs           2 μs

Comparison:
Custom      364.37 K
1.9.0       266.38 K - 1.37x slower +1.01 μs

##### With input 1000 #####
Name             ips        average  deviation         median         99th %
Custom      357.60 K        2.80 μs ±15320.90%           0 μs           1 μs
1.9.0        26.68 K       37.48 μs  ±3339.23%           6 μs          19 μs

Comparison:
Custom      357.60 K
1.9.0        26.68 K - 13.40x slower +34.68 μs

##### With input 10000 #####
Name             ips        average  deviation         median         99th %
Custom      355.01 K        2.82 μs ±15286.94%           0 μs           1 μs
1.9.0         2.49 K      402.25 μs   ±944.93%         100 μs         184 μs

Comparison:
Custom      355.01 K
1.9.0         2.49 K - 142.80x slower +399.44 μs

##### With input 100000 #####
Name             ips        average  deviation         median         99th %
Custom      398.34 K     0.00251 ms ±16339.24%           0 ms     0.00100 ms
1.9.0         0.20 K        4.96 ms   ±282.56%        1.20 ms       60.25 ms

Comparison:
Custom      398.34 K
1.9.0         0.20 K - 1975.07x slower +4.96 ms

##### With input 1000000 #####
Name             ips        average  deviation         median         99th %
Custom      351.08 K     0.00285 ms ±20803.20%           0 ms     0.00100 ms
1.9.0       0.0221 K       45.33 ms   ±122.39%       15.71 ms      203.02 ms

Comparison:
Custom      351.08 K
1.9.0       0.0221 K - 15912.62x slower +45.32 ms
```

Here is the benchmark itself:

```elixir
defmodule Custom do
  def disjoint?(%MapSet{map: map1}, %MapSet{map: map2}) do
    {map1, map2} = order_by_size(map1, map2)

    map1
    |> :maps.iterator()
    |> :maps.next()
    |> none_in?(map2)
  end

  @compile {:inline, [none_in?: 2]}

  defp none_in?(:none, _), do: true

  defp none_in?({key, _val, iter}, map2) do
    not :erlang.is_map_key(key, map2) and none_in?(:maps.next(iter), map2)
  end

  @compile {:inline, [order_by_size: 2]}
  defp order_by_size(map1, map2) when map_size(map1) > map_size(map2), do: {map2, map1}
  defp order_by_size(map1, map2), do: {map1, map2}
end

make_set = fn n ->
  for index <- 0..n, into: MapSet.new() do
    :md5 |> :crypto.hash(to_string(index)) |> Base.encode64()
  end
end

Benchee.run(%{
  "1.9.0" => fn set -> MapSet.disjoint?(set, set) end,
  "Custom" => fn set -> Custom.disjoint?(set, set) end
}, inputs: %{
  "10" => make_set.(10),
  "100" => make_set.(100),
  "1000" => make_set.(1000),
  "10000" => make_set.(10000),
  "100000" => make_set.(100000),
  "1000000" => make_set.(1000000)
})
```

josevalim:
Oh wow, very interesting. So even for same inputs it is much faster. I guess it makes sense: we don’t have to allocate such a big list. So your initial benchmark was on worst case scenario all along then?

I wonder if we should also change the Enumerable implementation to use iterators. But that’s definitely a separate discussion.

josevalim:
And thank you for running further benchmarks!

sorentwo:
The original benchmark wasn’t using the exact same set, there was only partial overlap toward the “end” of the list.

lib/elixir/lib/map_set.ex
```diff
@@ -196,19 +198,17 @@ defmodule MapSet do
...
-    |> Map.keys()
+    |> :maps.iterator()
+    |> :maps.next()
     |> none_in?(map2)
   end
 
-  defp none_in?([], _) do
-    true
-  end
+  @compile {:inline, [none_in?: 2]}
```
michalmuskala: We probably shouldn't inline recursive functions, the in-liner is sometimes pretty weird with them.

michalmuskala:
Iterators make sense here since we don't need to do the unpack-repack dance with tuples we'd need to do in `Enumerable`.

lib/elixir/lib/map_set.ex
sorentwo: I ran the benchmarks again with the inlining removed, it made a negligible difference. I've removed the inline declaration for `none_in?/2` and `all_in?/2`.

josevalim:
Better late than never! :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

PR #9373, opened 2019-09-30
Add Enum.tally/2

QWYNG:
Hi! Thank you for a great programing language.
This is a suggestion and a PR in one.

Add tally function like Rubys tally.
https://bugs.ruby-lang.org/issues/11076

Let me know if there's anything to be clarified/updated.
Hope it makes sense!

PR #9383, opened 2019-10-05
add Enum.tally

QWYNG:
Hi! Thank you for great programming language.

This is a tally method like Ruby.
PHP has a similar method too.
https://www.php.net/manual/en/function.array-count-values.php

This is a discussion about this method.
https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/elixir-lang-core/hBRKIIy8QKE/2JlhpblqBgAJ

I have been slightly changed doc from #9373
I hope it makes sense!

### DIFF
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -1148,6 +1148,28 @@ defmodule Enum do
   end
 
   @doc """
+  creates a new map with keys as unique elements from enumerable,
+  and set values as the count of every element.
+  This may perform poorly on really large sets. It requires to retain every element in the collection.
+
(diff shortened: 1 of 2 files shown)


josevalim:
Hi! I don’t think we reached conclusion on the mailing list yet, as there
are some reservations on the name, so if possible, please wait until a
conclusion is made. :) also, you don’t need to send new PRs, we can reopen
the current one. Thank you!
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

QWYNG:
Srry @josevalim 
OK Ill wailt and reopen

josevalim:
No problem at all, thanks for the contributions. :)

PR #9395, opened 2019-10-11
Close unclosed backquotes (backticks)

eksperimental:
ExDoc was giving the following errors:
> lib/elixir/lib/string.ex:250: warning: Closing unclosed backquotes ` at end of input
> lib/elixir/lib/enum.ex:2319: warning: Closing unclosed backquotes ` at end of input


eksperimental:
Should we add `make docs` to the CI file, so we avoid this kind of documentation errors?
Alternatively we can add a note to RELEASE to check the output for errors when running `make zips`.
Alternatively alternatively we can modify the `zips` to fail if anything is output to stderr (if it is possible to do this).

whatyouhide:
Thank you @eksperimental! 💟

PR #9396, opened 2019-10-11
earliest and latest functions for calendar types

whossname:
As discussed on the mailing list [here](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!topic/elixir-lang-core/PCOm9ZBrym0).

The only tests here are the doc tests. Do we need more?

### DIFF
--- a/lib/elixir/lib/calendar/date.ex
+++ b/lib/elixir/lib/calendar/date.ex
@@ -452,6 +452,55 @@ defmodule Date do
   end
 
   @doc """
+  Sorts the mapped results of the `enumerable` from the earliest date to the latest date.
+
+  This function maps each element of the `enumerable` using the `mapper` function. The
+  returned value from the mapper function should be a `Date` struct. If no mapper
+  function is given, the `enumerable` elements are sorted from the earliest
+  date to the latest date.
+
+  ## Examples
+      iex> date_list = [~N[2000-02-29 09:00:00], ~D[2000-03-29], ~D[2000-02-29]]
+      iex> Date.earliest(date_list)
+      [~N[2000-02-29 09:00:00], ~D[2000-02-29], ~D[2000-03-29]]
+
+      iex> map_list = [%{d: ~D[2000-03-29]}, %{d: ~D[2000-02-29]}]
+      iex> Date.earliest(map_list, &(&1.d))
+      [%{d: ~D[2000-02-29]}, %{d: ~D[2000-03-29]}]
+
+  """
+  @spec earliest(Enum.t(), (any() -> Calendar.date())) :: list()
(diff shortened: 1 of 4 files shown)


devonestes:
This looks great! What do you think about adding nice error messages in the event users pass in mixed types by accident? For example:

```
iex(1)> Date.compare(~D[2016-04-16], ~T[16:04:28])  
** (MatchError) no match of right hand side value: ~T[16:04:28]
    (elixir) lib/calendar/date.ex:449: Date.compare/2
```

If folks are calling these new functions with a list that includes `Date`s and `Time`s and then see this type of error, it won't be immediately clear that this error is coming from this function, and it would be a nicer experience for users if they get really nice error messages.

whossname:
@devonestes I might have jumped the gun a bit on making this pull request. It is still being discussed in the mailing list.

This is the current error for the example you gave:

```
iex(1)> Date.earliest([~D[2016-04-16], ~T[16:04:28]])
 ** (MatchError) no match of right hand side value: ~T[16:04:28]
       (elixir 1.10.0-dev) lib/elixir/lib/calendar/date.ex:428: Date.compare/2
       (elixir 1.10.0-dev) lib/elixir/lib/calendar/date.ex:479: anonymous fn/2 in Date.earliest/2
       (stdlib 3.10) lists.erl:969: :lists.sort/2
       (elixir 1.10.0-dev) lib/enum.ex:2383: Enum.sort_by/3
```

I take it you think it needs to be more explicit? So throw from Date.earliest?

josevalim:
@whossname IMO it is ok to raise at the moment it is being raised right now. Most functions in the Calendar modules are structural instead of nominal, so we only compare fields, which means it will always raise on pattern matching rather than unknown type.

whossname:
closing this. This isn't the functionality requested, and it looks like a completely different approach is going to be used anyway.

josevalim:
Hi @whossname!

We will definitely accept a PR though that adds Date.earliest/2 and Date.latest/2 that returns the earliest and the latest of two given dates. The same for Time, Naive and DateTime. Would you like to submit a PR for that? Thanks!

whossname:
@josevalim OK, cool. I'll make a new PR some time tomorrow. Thanks for letting me do it.

josevalim:
@whossname please hold on a bit. I found some limitations in the proposed approach. Let me do some double checking and I will follow up. Hopefully it will be quick. :)

josevalim:
Ok, I am back. I believe we won't need those functions for now. I was able to find an alternative that is good enough with compare: #9427. Sorry for the false positive.

whossname:
No problems

PR #9399, opened 2019-10-12
ExUnit.Callbacks: Be explicit about the shape of a return value

eksperimental:
Be explicit and also used a valid typespec (`keywords` is not valid,
but `keyword` is).


fertapric:
Thanks @eksperimental! 💜

PR #9401, opened 2019-10-13
feat: Add Kernel.trunc/2 to decimal places

MariliaMJ:
Disclaimer: This is the first time I am contributing to the elixir language so if there is anything wrong with my approach or intention, please warn me, I am still learning.

I am suggesting the addition of  arity 2 Kernel.trunc function that, given an integer on the second argument, it will truncate that decimal place, as such:
```
Kernel.trunc(2.345656, 2)
> 2.34
```

In some contexts, this is very relevant. I was very surprised I did not find it nor in elixir Kernel nor in Erlang. Hopefully, it will help more people.


maennchen:
@MariliaMJ This function exists in the Float module: https://hexdocs.pm/elixir/Float.html#floor/2

kipcole9:
Its really great you want to contribute. However the issue template says:
### Precheck

* Do not use the issue tracker for help or support (try Elixir Forum, Stack Overflow, IRC, etc.)

So the recommended approach is to propose a new feature first on the core mailing list.  Proposals can also be discussed on the [Elixir Forum](https://elixirforum.com)

MariliaMJ:
@maennchen Thanks for your quick reply!! That is amazing!
I see that `Float.floor/2` really does that, but it can have a behavior like this:
```
Float.floor(12.52, 2)
12.51
```
Which is not always desired 🤔 

@kipcole9 Also many thanks for guiding me in this new experience! If this is not suitable, then it can be closed and wait for further care <3

josevalim:
Hi @MariliaMJ, thanks for the PR.

The fact that 12.52 returns 12.51 is actually correct. That’s because a float 12.52 is not represented as 12.52 in memory, it actually contain up to 13 decimal digits but 12.52 is the shortest representation for that float. Doing string based truncation on a float is  actually discouraged per the IEEE spec. If you need said precision, then you should use Decimal.

Therefore, we won’t accept this PR. Thank you!

josevalim:
To further add, truncation returning 12.51 is the same property that often causes 3.0 + 3.0 in floats sometimes return 6.00000000001. It is a property of floats. Making it always return “6.0” would be “wrong”. :)

PR #9402, opened 2019-10-13
Raise when passing an empty list instead of clauses

fertapric:
This fixes an edge case in which the compiler will process an
empty list value (`[]`) as a block with no clauses, which is not
possible.

Here is an example:

    with {:ok, value} <- Map.fetch(%{a: 1}, :a) do
      value
    else
      []
    end

With this change, the code above will not compile.

### DIFF
--- a/lib/elixir/src/elixir_clauses.erl
+++ b/lib/elixir/src/elixir_clauses.erl
@@ -308,7 +308,7 @@ expand_clauses(Meta, Kind, Fun, Clauses, E) ->
   NewKind = origin(Meta, Kind),
   expand_clauses_origin(Meta, NewKind, Fun, Clauses, E).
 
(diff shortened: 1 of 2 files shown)


lib/elixir/src/elixir_clauses.erl
```diff
@@ -308,7 +308,7 @@ expand_clauses(Meta, Kind, Fun, Clauses, E) ->
   NewKind = origin(Meta, Kind),
   expand_clauses_origin(Meta, NewKind, Fun, Clauses, E).
 
-expand_clauses_origin(Meta, Kind, Fun, {Key, Clauses}, E) when is_list(Clauses) ->
+expand_clauses_origin(Meta, Kind, Fun, {Key, [_ | _] = Clauses}, E) when is_list(Clauses) ->
```
michalmuskala: The guard could be skipped here

fertapric: Right!

PR #9403, opened 2019-10-14
Standardize use of verbs in bulleted lists

eksperimental:
* Use imperative verbs in bulleted lists
* Use "does not" instead of "do not" in bulleted lists

Related: 690c2c40c948564f7db5fccfd66246ed78c6fe8d

[ci skip]


fertapric:
Thanks @eksperimental! 📖 ✅

PR #9412, opened 2019-10-15
check for underscored pattern in Kernel.match?/2

### LINKED ISSUE #9411: Kernel.match?/2 throws a compiler warning when matching underscore
### Environment

* Elixir 1.8.1 (compiled with Erlang/OTP 21), Erlang 21.0

### Current behavior

The `Kernel.match?/2` convenience macro will throw a complier warning when matching against underscores.

```elixir
# foo.ex
defmodule Foo do
  def foo(value) do
    match?(_foo, value)
  end
end
```

```console
$ elixirc foo.ex
warning: this clause cannot match because a previous clause at line 4 always matches
  foo.ex:4
```

This warning isn't a problem with normal Elixir use, but is a bit annoying when using `match?/2` in macros. I have a DSL that uses `match?/2` to match against a pattern that may be an underscore. Under certain conditions, the macro may compile to something like

```elixir
cond do
  ..

  true and match?(_all, binding_from_outside_the_cond) and true -> :ok
end
```

which will throw the warning.

### Expected behavior

I would like `match?/2` to compile without warning when an underscore is provided as the pattern.

It also seems wise to me to optimize `match?/2` so that when

```elixir
match?(_, some_very_expensive_func())
```

is written, `some_very_expensive_func/0` is never called.

the-mikedavis:
fixes #9411

### DIFF
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -2823,16 +2823,28 @@ defmodule Kernel do
 
   """
   defmacro match?(pattern, expr) do
-    quote do
-      case unquote(expr) do
-        unquote(pattern) ->
-          true
+    case underscored_binding?(pattern) do
+      true ->
+        true
 
-        _ ->
-          false
-      end
+      false ->
+        quote do
+          case unquote(expr) do
+            unquote(pattern) ->
+              true
+
+            _ ->
+              false
+          end
+        end
     end
   end
+
+  defp underscored_binding?({binding, _line_info, nil}) when is_atom(binding) do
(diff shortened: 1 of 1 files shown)


the-mikedavis:
This won't fix the problem for more complicated expressions like `_any = _all`, which also throws the compiler warning. Is there a more robust way to detect "wildcard" patterns?

josevalim:
Hi @the-mikedavis! Thanks for the PR! However, a better fix would be to actually mark the second clause as generated. I will push a fix soon. <3

the-mikedavis:
Thanks @josevalim! I was hoping there was something cool like that 😄

PR #9414, opened 2019-10-16
Fix typespec for Calendar.ISO.year_of_era/1

eksperimental:
The return `year` is always a positive number

Dialyzer was giving this error:

> lib/calendar/iso.ex:672: Type specification 'Elixir.Calendar.ISO':year_of_era(year()) -> {year(),era::0..1} is a supertype of the success typing: 'Elixir.Calendar.ISO':year_of_era
          (integer()) -> {pos_integer(), 0 | 1}


josevalim:

PR #9415, opened 2019-10-16
Fix return value of Calendar.Time.convert/2 typespec

eksperimental:
It will never return `{:error, atom}`

Dialyzer's error:
 lib/calendar/time.ex:484: The specification for 'Elixir.Time':convert/2 states that the function might also return
          {'error', atom()} but the inferred return is
          {'ok',
           #{'__struct__' := 'Elixir.Time',
             'calendar' := _,
             'hour' := _,
             'microsecond' := _,
             'minute' := _,
             'second' := _}}


eksperimental:
I'm submitting a new PR

PR #9417, opened 2019-10-16
WIP: Fix return value of Calendar.Time.convert/2

eksperimental:
It would never return `{:error, atom}`,

Dialyzer's errors:
```
 lib/calendar/time.ex:484: The specification for 'Elixir.Time':convert/2 states that the function might also return
          {'error', atom()} but the inferred return is
          {'ok',
           #{'__struct__' := 'Elixir.Time',
             'calendar' := _,
             'hour' := _,
             'microsecond' := _,
             'minute' := _,
             'second' := _}}

lib/calendar/time.ex:546: The pattern
          {'error', _reason@1} can never match the type
          {'ok',
           #{'__struct__' := 'Elixir.Time',
             'calendar' := atom(),
             'hour' := non_neg_integer(),
             'microsecond' := {char(), 0 | 1 | 2 | 3 | 4 | 5 | 6},
             'minute' := non_neg_integer(),
             'second' := non_neg_integer()}}
```

so it also fixes convert!/2

### DIFF
--- a/lib/elixir/lib/calendar/time.ex
+++ b/lib/elixir/lib/calendar/time.ex
@@ -505,7 +505,10 @@ defmodule Time do
     {:ok, time}
   end
 
-  def convert(%{microsecond: {_, precision}} = time, calendar) do
+  def convert(
+        %{hour: _, minute: _, second: _, microsecond: {_, precision}} = time,
+        calendar
+      ) do
     {hour, minute, second, {microsecond, _}} =
       time
       |> to_day_fraction()
@@ -520,6 +523,18 @@ defmodule Time do
     }
 
     {:ok, time}
+  end
+
+  def convert(
+        %{
+          hour: _,
+          minute: _,
+          second: _,
+          microsecond: _
+        },
+        _calendar
+      ) do
+    {:error, :unrecognized_time_format}
   end
 
   @doc """


lib/elixir/lib/calendar/time.ex
```diff
@@ -522,6 +525,18 @@ defmodule Time do
...
+  def convert(
+        %{
+          hour: _,
+          minute: _,
+          second: _,
+          microsecond: _
+        },
+        _calendar
+      ) do
+    {:error, :unrecognized_time_format}
```
eksperimental: What would be a proper reason for this error?

wojtekmach: I don't think we should have this clause, if the 1st argument isn't a valid time map the function should crash.

eksperimental: I think the idea is the `convert!/2` crashes if that's the case, but as of now, `convert!/2` never raises because convert/2 never returns `{:error, _}`

wojtekmach:
I think the analysis is correct, I don't see how with current code we'd return `{:error, ...}`. Worth mentioning that, on the other hand, if we change the spec now but later do return `{:error, _}`, dialyzer won't tell us our typespec is incomplete. :(

lib/elixir/lib/calendar/time.ex
wojtekmach: Sure. But the problem is in this clause we are basically handling a situation that someone has a map with incorrect shape (incorrect microseconds) and instead of gracefully handling that, we should crash. If you wanna look at it from the types angle, we're handling a situation that shouldn't be possible given the 1st argument should be of `Calendar.time()` type.

josevalim:
Yes, the spec is more broad than the actual implementation because we are signalling users it may return `{:error, _}` in the future.  Otherwise, if don't signal that to users and then we add an `{:error, _}`, it can be considered a breaking change. If you want, I believe you can run dialyzer so it doesn't warn on "overspecs", but for contract purposes, our spec is correct. thanks!

eksperimental:
ok. thank you guys for your input

PR #9418, opened 2019-10-16
Several fixes in Code.format_{string,file}!/2

eksperimental:
Dialyzer's errors:
lib/code.ex:616: The specification for 'Elixir.Code':'format_string!'/2 states that the function might also return
          binary() but the inferred return is
          [binary()]
lib/code.ex:630: The specification for 'Elixir.Code':'format_file!'/2 states that the function might also return
          binary() but the inferred return is
          [[binary()] | 10, ...]

lead make the changes

### DIFF
--- a/lib/elixir/lib/code.ex
+++ b/lib/elixir/lib/code.ex
@@ -613,7 +613,7 @@ defmodule Code do
   the latter.
   """
   @doc since: "1.6.0"
-  @spec format_string!(binary, keyword) :: iodata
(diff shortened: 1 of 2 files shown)


josevalim:
The current specification is correct. What we guarantee is iodata. If it ends up being exclusively a list of binaries, that's an implementation detail, but not what we guarantee in terms of contract. As per the other issue, you are probably running dialyzer with too specific settings. :)

PR #9421, opened 2019-10-16
Remove typespecs from private functions

eksperimental:
The command used to find the offending lines is:
`$ ag "@spec [^\n]+\n\s+def[a-z]*?p "`


josevalim:
I believe the first one was added because otherwise dialyzer complains. The second one I would keep because the function has complex signature and that helps the developers reading the code. :) But in general you are correct, we don't spec private functions.

Thanks for the PR!

PR #9424, opened 2019-10-17
Introduce Code.binding/0 type

eksperimental:


josevalim:
Thanks @eksperimental! However, I believe the correct term is singular (or at least that's what we use in `Kernel.binding`). In other words, a binding is a group of variables and their values. So can you please move everything to singular accordingly? Thank you.

eksperimental:
@josevalim please have a look

fertapric:
Thanks @eksperimental! 💜

PR #9425, opened 2019-10-18
add  Enum.frequencies and Enum.frequencies_by

QWYNG:
Hi! Thank you for great programming language.
add frequencies and frequencies_by.

This is a discussion about this method.
https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/elixir-lang-core/hBRKIIy8QKE/2JlhpblqBgAJ

If any points to fix, don't hesitate to say.

### DIFF
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -1111,6 +1111,7 @@ defmodule Enum do
       %{"ant" => 3, "buffalo" => 2, "dingo" => 1}
       
   """
+  @doc since: "1.10.0"
   @spec frequencies(t) :: map
   def frequencies(enumerable) do
     reduce(enumerable, %{}, fn key, acc ->
@@ -1131,6 +1132,7 @@ defmodule Enum do
       %{3 => 2, 2 => 2, 1 => 1}
 
   """
+  @doc since: "1.10.0"
   @spec frequencies_by(t, (element -> any)) :: map
   def frequencies_by(enumerable, key_fun) when is_function(key_fun) do
     reduce(enumerable, %{}, fn entry, acc ->
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -1102,13 +1102,27 @@ defmodule Enum do
   end
 
   @doc """
-  Returns a map with keys as unique elements given by `key_fun` and values
+  Returns a map with keys as unique elements of `enumerable` and values
   as the count of every element.
 
   ## Examples
 
-      iex> Enum.frequencies_by(~w{ant buffalo ant ant buffalo dingo})
+      iex> Enum.frequencies(~w{ant buffalo ant ant buffalo dingo})
       %{"ant" => 3, "buffalo" => 2, "dingo" => 1}
+      
+  """
+  @spec frequencies(t) :: map
+  def frequencies(enumerable) do
+    reduce(enumerable, %{}, fn key, acc ->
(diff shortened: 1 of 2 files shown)


Eiji7:
@QWYNG How about using reduce and implement only `frequencies_by/1` and `frequencies_by/2` (see `Enum.group_by/2` and `Enum.group_by/3`)?

Here is my proposition:
```elixir
defmodule Example do
  @spec frequencies_by(Enum.t(), (Enum.element() -> any)) :: map
  def frequencies_by(enumerable, key_fun \\ fn x -> x end) when is_function(key_fun) do
    Enum.reduce(enumerable, %{}, fn entry, acc ->
      key = key_fun.(entry)
      Map.update(acc, key, 1, &(&1 + 1))
    end)
  end
end
```

josevalim:
Yes, we should go with @Eiji7 proposal, as it is quite more efficient. :)

QWYNG:
sorry, I accidentally closed

josevalim:
Sorry @QWYNG. I think there was some confusion. We should have two functions: `frequencies/1` and `frequencies_by/2`. Their implementation is very close to each other but given no `_by` functions today have a default argument, I don't think we should introduce one here. Can you please amend accordingly? Sorry for the confusion and thanks!

QWYNG:
@josevalim 
OK! That means we need two functions like `min` and` min_by`.
That way I think that `frequencies_by/2` don't need a default argument.
I'll amend.

whatyouhide:
Thank you @QWYNG! 💟

QWYNG:
@whatyouhide 
You are Welcome!
@Eiji7 @josevalim @whatyouhide 
Thank you all! Elixir is a great language!

lib/elixir/lib/enum.ex
```diff
@@ -1101,6 +1101,46 @@ defmodule Enum do
...
+      iex> Enum.frequencies_by(~w{aaa aA bbb cc c}, &String.length/1)
+      %{3 => 2, 2 => 2, 1 => 1}
+
+  """
+  @doc since: "1.10.0"
+  @spec frequencies_by(t, (element -> any)) :: map
+  def frequencies_by(enumerable, key_fun) when is_function(key_fun) do
+    reduce(enumerable, %{}, fn entry, acc ->
+      key = key_fun.(entry)
+      Map.update(acc, key, 1, &(&1 + 1))
```
michalmuskala: Given it's a library function and it should be as efficient as possible, could we inline the call to `Map.update` to avoid the fun overhead? It's tiny, but things like that add up, especially in library functions.

josevalim:
Yeah, a PR would be welcome.

QWYNG:
@michalmuskala @josevalim 
Like this?

```elixir
  def frequencies(enumerable) do
    reduce(enumerable, %{}, fn key, acc ->
      case acc do
        %{^key => value} ->
          Map.put(acc, key, value + 1)
    
        %{} ->
          Map.put(acc, key, 1)
      end
    end)
  end
```

If it's alright with @michalmuskala, I will create PR. :)

josevalim:
I would even do this:

```elixir
  def frequencies(enumerable) do
    reduce(enumerable, %{}, fn key, acc ->
      case acc do
        %{^key => value} -> %{acc | key => value + 1}
        %{} -> Map.put(acc, key, 1)
      end
    end)
  end
```

A PR is welcome!

michalmuskala:
Please go ahead @QWYNG. Thank you!

PR #9428, opened 2019-10-18
DynamicSupervisor Documentation polish

hassanshaikley:
I was reading through the docs and thought this bit read weird, so here's a PR that polishes the docs IMHO.

If y'all don't think this is helpful that's okay!

💚


josevalim:

PR #9430, opened 2019-10-18
Add example compilation tracer to docs

wojtekmach:


lib/elixir/lib/code.ex
```diff
@@ -85,6 +85,19 @@ defmodule Code do
 
   New events may be added at any time in the future, therefore it is advised
   for the `trace/2` function to have a "catch-all" clause.
+
+  Below is example tracer that prints all remote function invocations:
+
+      defmodule MyTracer do
+        def trace({:remote_function, _meta, module, name, arity}, env) do
+          IO.puts "#{env.file}:#{env.line} #{inspect(module)}.#{name}/#{arity}"
```
wojtekmach: When testing this out on a sample project, I noticed `meta` contained `:line` but not `:file`, hence I ended up using macro env to grab both. Is this the way to go? Are there cases where `meta` contains data that is not in the env?

josevalim: > Are there cases where meta contains data that is not in the env?

Yes, there is documentation about available metadata in the `Macro` module. The metadata may include more parsers information too, if you enable them in the compiler.

josevalim:

lib/elixir/lib/code.ex
wojtekmach: Got it, thank you!

PR #9433, opened 2019-10-20
Fix specs in IE break specs

eksperimental:
- function is an atom
- breaks/stops are non-negative-integers

### DIFF
--- a/lib/iex/lib/iex.ex
+++ b/lib/iex/lib/iex.ex
@@ -607,9 +607,10 @@ defmodule IEx do
     end
   end
 
-  def __break__!({:/, _, [call, arity]} = ast, stops, env) when is_integer(arity) do
+  def __break__!({:/, _, [call, arity]} = ast, stops, env)
+      when arity in 0..255 and is_integer(stops) and stops >= 0 do
     with {module, fun, []} <- Macro.decompose_call(call),
-         module when is_atom(module) <- Macro.expand(module, env) do
+         module when is_atom(module) and is_atom(fun) <- Macro.expand(module, env) do
       IEx.Pry.break!(module, fun, arity, stops)
     else
       _ ->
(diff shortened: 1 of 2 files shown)


lib/iex/lib/iex.ex
```diff
@@ -607,9 +607,10 @@ defmodule IEx do
     end
   end
 
-  def __break__!({:/, _, [call, arity]} = ast, stops, env) when is_integer(arity) do
+  def __break__!({:/, _, [call, arity]} = ast, stops, env)
+      when arity in 0..255 and is_integer(stops) and stops >= 0 do
     with {module, fun, []} <- Macro.decompose_call(call),
-         module when is_atom(module) <- Macro.expand(module, env) do
+         module when is_atom(module) and is_atom(fun) <- Macro.expand(module, env) do
```
josevalim: `is_atom(fun)` is already guaranteed by `decompose_call`, so I wouldn't check it again.

lib/iex/lib/iex.ex
```diff
@@ -766,9 +767,8 @@ defmodule IEx do
 
   """
   @doc since: "1.5.0"
-  def break!(module, function, arity, stops \\ 1) when is_integer(arity) do
-    IEx.Pry.break!(module, function, arity, stops)
-  end
+  @spec break!(module, atom, arity, non_neg_integer) :: IEx.Pry.id()
+  defdelegate break!(module, function, arity, stops), to: IEx.Pry
```
josevalim: This is changing the public API because it is dropping the default `stops` argument.

josevalim: I would drop the changes in this file, the other changes are good to go! Good catch!

eksperimental: True. I missed the default argument

eksperimental: I reduced the guard changes to a minimum. Please have a look. I'm using default arg with defdelegate, and adding specs since they are not getting imported by ExDoc. Please let me know if that's fine or everything should be dropped

whatyouhide:
Thank you @elsperimental! ❤️

PR #9434, opened 2019-10-21
Updates on DynamicSupervisor.start_child/2

eksperimental:
- Changes to its specs to reflect deprecated child spec.
- Changes to the code so not it fails with FunctionClauseError if child_spec
  doesn't match the spec.
- Documentation fixes.

### DIFF
--- a/lib/elixir/lib/dynamic_supervisor.ex
+++ b/lib/elixir/lib/dynamic_supervisor.ex
@@ -286,7 +286,7 @@ defmodule DynamicSupervisor do
   Dynamically adds a child specification to `supervisor` and starts that child.
 
   `child_spec` should be a valid child specification as detailed in the
-  "child_spec/1" section of the documentation for `Supervisor`. The child
+  "Child specification" section of the documentation for `Supervisor`. The child
   process will be started as defined in the child specification.
 
   If the child process start function returns `{:ok, child}` or `{:ok, child,
@@ -306,7 +306,13 @@ defmodule DynamicSupervisor do
   this function returns `{:error, :max_children}`.
   """
   @doc since: "1.6.0"
-  @spec start_child(Supervisor.supervisor(), Supervisor.child_spec() | {module, term} | module) ::
+  @spec start_child(
+          Supervisor.supervisor(),
+          Supervisor.child_spec()
+          | {module, term}
+          | module
+          | (old_erlang_child_spec :: :supervisor.child_spec())
+        ) ::
           on_start_child()
   def start_child(supervisor, {_, _, _, _, _, _} = child_spec) do
     validate_and_start_child(supervisor, child_spec)
--- a/lib/elixir/lib/supervisor.ex
+++ b/lib/elixir/lib/supervisor.ex
@@ -108,8 +108,8 @@ defmodule Supervisor do
   The child specification describes how the supervisor starts, shuts down,
   and restarts child processes.
 
-  The child specification contains 6 keys. The first two are required,
-  and the remaining ones are optional:
+  The child specification is a map which contains 6 elements. The first two keys
+  in the following list are required, and the remaining ones are optional:
 
     * `:id` - any term used to identify the child specification
       internally by the supervisor; defaults to the given module.
@@ -131,8 +131,8 @@ defmodule Supervisor do
     * `:type` - specifies that the child process is a `:worker` or a
       `:supervisor`. This key is optional and defaults to `:worker`.
 
-  There is a sixth key, `:modules`, that is rarely changed. It is set
-  automatically based on the value in `:start`.
+  There is a sixth key, `:modules`, which is optional and is rarely changed.
+  It is set automatically based on the `:start` value.
 
   Let's understand what the `:shutdown` and `:restart` options control.
 


lib/elixir/lib/supervisor.ex
```diff
@@ -131,8 +131,9 @@ defmodule Supervisor do
     * `:type` - specifies that the child process is a `:worker` or a
       `:supervisor`. This key is optional and defaults to `:worker`.
 
-  There is a sixth key, `:modules`, that is rarely changed. It is set
-  automatically based on the value in `:start`.
+
+    * `:modules` - this is rarely changed. This key is optional and it
+      is set automatically based on the `:start` value.
```
eksperimental: We are still not explaining what this key does.

lib/elixir/lib/dynamic_supervisor.ex
```diff
@@ -311,14 +311,26 @@ defmodule DynamicSupervisor do
...
+  def start_child(supervisor, {module, _} = child_spec) when is_atom(module) do
     validate_and_start_child(supervisor, Supervisor.child_spec(child_spec, []))
   end
 
+  def start_child(supervisor, module) when is_atom(module) do
+    validate_and_start_child(supervisor, Supervisor.child_spec(module, []))
+  end
+
+  # Supports the deprecated Supervisor.Spec.spec/0 type
+  def start_child(supervisor, {_, _, _, _, _, _} = child_spec) do
```
josevalim: They are not really deprecated. We will continue to support them for legacy reasons, but they are discouraged.

lib/elixir/lib/supervisor.ex
josevalim: Please keep it as before. The reason why we didn't set it on its own paragraph is to not highlight it. We can document it behaviour in another commit.

lib/elixir/lib/dynamic_supervisor.ex
```diff
@@ -311,14 +311,26 @@ defmodule DynamicSupervisor do
   @doc since: "1.6.0"
   @spec start_child(Supervisor.supervisor(), Supervisor.child_spec() | {module, term} | module) ::
           on_start_child()
-  def start_child(supervisor, {_, _, _, _, _, _} = child_spec) do
-    validate_and_start_child(supervisor, child_spec)
+  @spec start_child(Supervisor.supervisor(), deprecated_child_spec :: Supervisor.Spec.spec()) ::
+          on_start_child()
```
josevalim: I honestly don't see the benefit of this change. We are duplicating all of the clauses in `Supervisor.child_spec/2` here. What is the gain?

eksperimental: Well, they are not deprecated, because their usage in this module hasn't been formally depracted, but `Supervisor.Spec` has. The gain is of having two specs is that you can show in the second one that its usage is discourage (something like `discouraged_child_spec :: Supervisor.Spec.spec()` can be used instead), and user only has to look at the first one.

lib/elixir/lib/dynamic_supervisor.ex
eksperimental: I will update the comment.

lib/elixir/lib/supervisor.ex
eksperimental: please have a look at it again

lib/elixir/lib/dynamic_supervisor.ex
josevalim: They are not deprecated and they will never be, unless Erlang deprecates them, which is unlikely. Otherwise we can’t use DynamicSupervisor with an Erlang lib that returns old type specs.

josevalim: Can you please send a PR only with the docs changes? The code changes are incorrect per the reason above, thank you.

eksperimental: Sorry, but this PR does three things. 1. fixes documentation. 2. fixes the specs which still needs to be fixed regardless of whether they are split into two, or labeled as "discouraged". 3. fixes the code to emit a FunctionClauseError when a spec that is not covered by the specs.
The changes to the code do not change at all the way the code deals when provided a spec in the old format (6-tuple one)

josevalim: The code change is undesired, there is no reason to duplicate the work done by Supervisor.child_spec.

josevalim: Also we should not be referring to Supervisor.Spec. If you need to, mention “old Erlang child specs”.

josevalim: For the spec bit, you can use :supervisor.child_spec instead of Supervisor.child_spec.

josevalim: Ping! Can you please update the PR to only do 1 and 2 of what you described above? Thank you.

eksperimental:
updated!

lib/elixir/lib/dynamic_supervisor.ex
eksperimental: done

josevalim:

PR #9438, opened 2019-10-21
Allow start_options to be configured on Logger

josevalim:


lib/logger/lib/logger.ex
```diff
@@ -97,6 +97,10 @@ defmodule Logger do
       if you want to purge log calls from a dependency, the dependency must be
       recompiled.
 
+    * `:start_options` - passes start options to Logger's main process, such
+      as `:spawn_opt` and `:hibernate_after`. See `t:GenServer.option` for
```
fertapric: Should the documentation clarify that `:name` is not supported? The link to [`GenServer.option/0`](https://hexdocs.pm/elixir/GenServer.html#t:option/0) list `:name` as option.

lib/logger/lib/logger.ex
```diff
@@ -97,6 +97,10 @@ defmodule Logger do
       if you want to purge log calls from a dependency, the dependency must be
       recompiled.
 
+    * `:start_options` - passes start options to Logger's main process, such
+      as `:spawn_opt` and `:hibernate_after`. All options in `t:GenServer.option`
+      are accepted, except by `:name`.
```
fertapric:
```suggestion
      are accepted, except `:name`.
```

PR #9439, opened 2019-10-22
Consistently raise ArgumentError while merging not keywords

am-kantox:
Currently merging `Keyword` and not `Keyword` raises an exception unless the counterpart is empty.

```elixir
Keyword.merge([1, 2], [])
#⇒ [1, 2]
```

The consistent behaviour should probably be to raise if one of the arguments is not a keyword list, despite whether the other one is empty or not.

### DIFF
--- a/lib/elixir/lib/keyword.ex
+++ b/lib/elixir/lib/keyword.ex
@@ -692,24 +692,31 @@ defmodule Keyword do
   @spec merge(t, t) :: t
   def merge(keywords1, keywords2)
 
-  def merge(keywords1, []) when is_list(keywords1), do: keywords1
-  def merge([], keywords2) when is_list(keywords2), do: keywords2
+  def merge(keywords1, []) when is_list(keywords1), do: keyword_or_raise(keywords1, "first")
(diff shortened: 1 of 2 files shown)


josevalim:
I am not convinced on this change. If we were to check the validity of lists every time we process them, then everything would get more expensive. It is easy to look at the validity of the feature from the perspective of invalid keywords, but from the perspective of passing valid keywords - which is the most common scenario, it is just not worth it.

Elixir is a dynamic language, it means sometimes you will pass data around that does not conform, but it should fail at some point later.

am-kantox:
I am not convinced too. And I had the very same doubts.

But actually the change does not have any _real_ impact on processing keywords. It brings impact _only_ on merging empty keywords. And that’s why I think here we’d better be consistent.

josevalim:
It still impacts because merging empty keyword lists now becomes O(n) and I would say this is more common than you think. Imagine every function that receives opts (by default an empty list) and merges the default arguments.

am-kantox:
> function that receives opts

Makes total sense. I believe it should be closed.

josevalim:
Yeah, agreed. In any case, thanks for the PR!

PR #9442, opened 2019-10-23
Diff strings with escaping with empty strings as well

devonestes:
String.bag_distance/2 always returns 0.0 if one of the strings is empty,
but that's probably not the behavior that we want in ExUnit, so I
enabled full diffing on strings if one of the two is empty.

This fixes the reported bug, but I'm not sure it's the right way to fix
this, so if folks have any other input I'd love to hear it.


josevalim:
Just to clarify (because I made a poor report), the error is that the strings are not escaped.

Here is the current result (even without empty strings): 

```
     code:  assert "{\"foo\":1,\"bar/baz\":[1,2,3]}" == "4"
     left:  "{"foo":1,"bar/baz":[1,2,3]}"
     right: "4"
```

Here is what I expected:

```
     code:  assert "{\"foo\":1,\"bar/baz\":[1,2,3]}" == "4"
     left:  "{\"foo\":1,\"bar/baz\":[1,2,3]}"
     right: "4"
```

When you add the special case for empty strings, it solves the issue for empty strings because you then force it to go through this branch:

https://github.com/elixir-lang/elixir/blob/f1cfa523ff2171ae54fc5b09c393289306bfbea4/lib/ex_unit/lib/ex_unit/diff.ex#L709-L713

But the issue is exactly that we are not escaping (by calling Code.Identifier) in the other branches. So the correct fix is (perhaps) to move the Code.Identifier call to outside of the cond, making sure it runs in all cases.

devonestes:
I had actually tried something similar to that, but it caused some other tests to fail, so I sort of assumed that was intended behavior in some cases. I’ll take another swing at it today and see if I can get it working without those other cases failing. It was a little tricky to fix because these functions weren’t tested directly. Would you mind if I wrote tests for ExUnit.Diff directly to make this sort of thing a little more clear?

devonestes:
@josevalim Yeah, I meant asserting on the return structure. The tests in `diff_test.exs` don't use the CLI formatter, so those tests in `diff_test.exs` aren't really representative of what we see in the console output, which was part of the problem for me in just tracking down where the bug was. I eventually found that it was in `ExUnit.Diff`, but it took me a little while to get there.

PR #9443, opened 2019-10-23
List.to_integer docs include the max value for base

hassanshaikley:
I think this is an important detail to include in the documentation because the error when the base it greater than 36 is cryptic.

```
iex(3)> List.to_integer('3FF', 37)
    :erlang.list_to_integer('3FF', 37)
```

💚


whatyouhide:
Thanks @hassanshaikley! 💟

PR #9444, opened 2019-10-24
Update frequencies and frequency_by to avoid the fun overhead

QWYNG:
Thank you for Elixir.
I updated `frequencies` and `frequency_by` to avoid the function overhead with advice https://github.com/elixir-lang/elixir/pull/9425#discussion_r336789305
Thank you @michalmuskala

### DIFF
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -1115,7 +1115,10 @@ defmodule Enum do
   @spec frequencies(t) :: map
   def frequencies(enumerable) do
     reduce(enumerable, %{}, fn key, acc ->
-      Map.update(acc, key, 1, &(&1 + 1))
+      case acc do
+        %{^key => value} -> %{acc | key => value + 1}
+        %{} -> Map.put(acc, key, 1)
+      end
     end)
   end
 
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/enum.ex
```diff
@@ -1115,7 +1115,10 @@ defmodule Enum do
   @spec frequencies(t) :: map
   def frequencies(enumerable) do
     reduce(enumerable, %{}, fn key, acc ->
-      Map.update(acc, key, 1, &(&1 + 1))
+      case acc do
+        %{^key => value} -> %{acc | key => value + 1}
+        %{} -> Map.put(acc, key, 1)
```
eksperimental: I think we can avoid pattern matching on `${}` since `acc` will always be a map., we can replace it with `_`

lib/elixir/lib/enum.ex
```diff
@@ -1137,7 +1140,11 @@ defmodule Enum do
   def frequencies_by(enumerable, key_fun) when is_function(key_fun) do
     reduce(enumerable, %{}, fn entry, acc ->
       key = key_fun.(entry)
-      Map.update(acc, key, 1, &(&1 + 1))
+
+      case acc do
+        %{^key => value} -> %{acc | key => value + 1}
+        %{} -> Map.put(acc, key, 1)
```
eksperimental: and same here

lib/elixir/lib/enum.ex
QWYNG: OK I see

josevalim: It is fine as is. I prefer to write code in the style above, as it is clearer in intent.

josevalim:

QWYNG:
@eksperimental @josevalim 
Thank you!

PR #9447, opened 2019-10-25
Execute callbacks on :elixir_compiler after all warnings are emitted

### LINKED ISSUE #9440: Warnings due to recompilation are escaping to the IEx console process
### Environment

```
$ elixir --version
Erlang/OTP 22 [erts-10.5] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1] [hipe] [dtrace]

Elixir 1.9.1 (compiled with Erlang/OTP 20)
```

### Current behavior

Create a project like `mix new proj` and add the following two files under `proj/lib/

```
defmodule A do
  # delete this comment
  import B
end
```

```
defmodule B do
  def foo, do: :ok
end
```

Now run `IEx` like this

```
$ iex -S mix
Erlang/OTP 22 [erts-10.5] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1] [hipe] [dtrace]

Interactive Elixir (1.9.1) - press Ctrl+C to exit (type h() ENTER for help)
iex(1)> 
```
Then without exiting `IEx`, open `A.ex` and delete the line with the comment. After this, go back to `IEx` and recompile

```
iex(1)> recompile
Compiling 1 file (.ex)
warning: unused import B
  lib/A.ex:2

:ok
iex(2)>
```
Now in `IEx` call the `flush` function

```
iex(2)> flush
{:warning, "/Users/user/proj/lib/A.ex", 2,
 [117, 110, 117, 115, 101, 100, 32, 105, 109, 112, 111, 114, 116, 32, 'B']}
:ok
iex(3)>
```
and you will see there is a message lost from the recompilation process.

### Expected behavior

There should be no message when calling `flush` after `recompile`. Somehow the recompilation process is letting warnings to escape as messages to the console process, and this should not happen.

ggcampinho:
Currently the callbacks on :elixir_compiler are called before the warnings of unused aliases and imports are emitted, causing the messages to not consumed during the compilation. This commit changes `:elixir_lexical.run/2` to `:elixir_lexical.run/3` splitting the callback in two, one during the execution to evaluate the forms and another to be called after all the warnings are emitted.

Fixes #9440

### DIFF
--- a/lib/elixir/src/elixir_lexical.erl
+++ b/lib/elixir/src/elixir_lexical.erl
@@ -13,8 +13,8 @@ run(#{tracers := Tracers} = E, ExecutionCallback, AfterExecutionCallback) ->
 
       try ExecutionCallback(LexEnv) of
         Res ->
-          warn_unused_aliases(Pid, E),
-          warn_unused_imports(Pid, E),
+          warn_unused_aliases(Pid, LexEnv),
+          warn_unused_imports(Pid, LexEnv),
           AfterExecutionCallback(LexEnv),
           Res
       after
--- a/lib/elixir/src/elixir_lexical.erl
+++ b/lib/elixir/src/elixir_lexical.erl
@@ -23,7 +23,8 @@ run(#{tracers := Tracers} = E, ExecutionCallback, AfterExecutionCallback) ->
       end;
 
     true ->
-      ExecutionCallback(E)
+      ExecutionCallback(E),
+      AfterExecutionCallback(E)
   end.
 
 trace({import, Meta, Module, Opts}, #{lexical_tracker := Pid}) ->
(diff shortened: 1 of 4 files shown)


lib/elixir/src/elixir_lexical.erl
```diff
@@ -1,27 +1,29 @@
...
+          AfterExecutionCallback(LexEnv),
           Res
       after
         unlink(Pid),
         ?tracker:stop(Pid)
       end;
 
     true ->
-      Callback(E)
+      ExecutionCallback(E)
```
josevalim: Shouldn't the AfterExecutionCallback be invoked here too?

ggcampinho: Good catch, I missed that!

josevalim:

PR #9448, opened 2019-10-26
Fix reading whole file opened as :charlist

mszczygiel:
Hello, this code:
```elixir
{:ok, file} = File.open("file.txt"), [:charlist])
IO.read(file, :all)
```
would return only first line of  the file. While I was preparing fix for this case, I've encountered another problem. Namely `StringIO` does not comply with `:io.getopts/1` well, as it returns `{:ok. opts}` instead of just `opts`, so I've fixed it too.

I would say these are bugs, but fixing them risks breaking existing code that depends on this behaviour, and I am not sure how you usually handle such changes.

I think this PR fixes both issues I've mentioned above.


josevalim:

PR #9452, opened 2019-10-28
Warn if duplicate key found in struct

Gazler:
Warning is the preferable solution for this, if we raise instead then it
is likely that there are some existing libraries that use duplicate
struct keys that will break.

### DIFF
--- a/lib/elixir/lib/kernel/utils.ex
+++ b/lib/elixir/lib/kernel/utils.ex
@@ -96,7 +96,7 @@ defmodule Kernel.Utils do
 
   # TODO: Make it raise on v2.0
     :lists.keysort(1, fields)
-    |> warn_on_duplicate_struct_key(nil)
+    |> warn_on_duplicate_struct_key()
 
     foreach = fn
       key when is_atom(key) ->
@@ -112,17 +112,17 @@ defmodule Kernel.Utils do
     {struct, enforce_keys, Module.get_attribute(module, :derive)}
   end
 
-  defp warn_on_duplicate_struct_key([], _) do
-    nil
+  defp warn_on_duplicate_struct_key([]) do
+    :ok
   end
 
-  defp warn_on_duplicate_struct_key([{key, _} | rest], key) do
-    IO.warn(~s(duplicate key #{inspect(key)} found in struct))
-    warn_on_duplicate_struct_key(rest, key)
+  defp warn_on_duplicate_struct_key([{key, _} | [{key, _} | _] = rest]) do
+    IO.warn("duplicate key #{inspect(key)} found in struct")
+    warn_on_duplicate_struct_key(rest)
   end
 
-  defp warn_on_duplicate_struct_key([{key, _} | rest], _) do
-    warn_on_duplicate_struct_key(rest, key)
+  defp warn_on_duplicate_struct_key([_ | rest]) do
+    warn_on_duplicate_struct_key(rest)
   end
 
   @doc """
--- a/lib/elixir/lib/kernel/utils.ex
+++ b/lib/elixir/lib/kernel/utils.ex
@@ -94,6 +94,10 @@ defmodule Kernel.Utils do
     fields = :lists.map(mapper, fields)
     enforce_keys = List.wrap(Module.get_attribute(module, :enforce_keys))
 
+  # TODO: Make it raise on v2.0
+    :lists.keysort(1, fields)
+    |> warn_on_duplicate_struct_key(nil)
+
     foreach = fn
       key when is_atom(key) ->
         :ok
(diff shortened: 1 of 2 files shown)


lib/elixir/lib/kernel/utils.ex
```diff
@@ -108,6 +111,19 @@ defmodule Kernel.Utils do
     {struct, enforce_keys, Module.get_attribute(module, :derive)}
   end
 
+  defp warn_on_duplicate_struct_key([], _) do
+    nil
+  end
+
+  defp warn_on_duplicate_struct_key([{key, _} | rest], key) do
+    IO.warn(~s(duplicate key "#{key}" found in struct))
```
fertapric: Nitpick: I would use `inspect(key)` here instead of `"#{key}"`

lib/elixir/test/elixir/map_test.exs
```diff
@@ -388,6 +388,18 @@ defmodule MapTest do
     end
   end
 
+  test "defstruct warns with duplicate keys" do
+    import ExUnit.CaptureIO
+
+    assert capture_io(:stderr, fn ->
```
fertapric: If `:stderr` is used, tests cannot be `async`. I would move this test to [the `warnings_test.exs` file](https://github.com/elixir-lang/elixir/blob/master/lib/elixir/test/elixir/kernel/warning_test.exs)

lib/elixir/test/elixir/map_test.exs
```diff
@@ -388,6 +388,18 @@ defmodule MapTest do
...
+             Code.eval_string("""
+             defmodule TestMod do
+             defstruct [:foo, :foo, :bar]
```
fertapric: Minor nitpick:

```suggestion
               defstruct [:foo, :foo, :bar]
```

lib/elixir/test/elixir/map_test.exs
Gazler: I did wonder why there was a separate file for warnings, that explains it!

lib/elixir/test/elixir/kernel/warning_test.exs
```diff
@@ -1770,6 +1770,18 @@ defmodule Kernel.WarningTest do
...
+    import ExUnit.CaptureIO
+
+    assert capture_err(fn ->
+             Code.eval_string("""
+             defmodule TestMod do
+               defstruct [:foo, :foo, :bar]
+             end
+             """)
+           end) =~ "duplicate key :foo found in struct"
+  end
```
fertapric: I think it would be good to purge the module at the end of the test, just in case:

```elixir
test "defstruct warns with duplicate keys" do
  ...
after
  purge(TestMod)
end
```

lib/elixir/lib/kernel/utils.ex
```diff
@@ -108,6 +111,19 @@ defmodule Kernel.Utils do
...
+  end
+
+  defp warn_on_duplicate_struct_key([{key, _} | rest], key) do
+    IO.warn(~s(duplicate key #{inspect(key)} found in struct))
+    warn_on_duplicate_struct_key(rest, key)
+  end
+
+  defp warn_on_duplicate_struct_key([{key, _} | rest], _) do
+    warn_on_duplicate_struct_key(rest, key)
+  end
```
whatyouhide: What about this implementation:

```elixir
defp warn_on_duplicate_struct_key([{key, _} | [{key, _} | _] = rest]) do
  IO.warn(~s(duplicate key #{inspect(key)} found in struct))
  warn_on_duplicate_struct_key(rest)
end

defp warn_on_duplicate_struct_key(_other) do
  :ok
end
```

?

lib/elixir/lib/kernel/utils.ex
```diff
@@ -108,6 +111,19 @@ defmodule Kernel.Utils do
     {struct, enforce_keys, Module.get_attribute(module, :derive)}
   end
 
+  defp warn_on_duplicate_struct_key([], _) do
+    nil
+  end
+
+  defp warn_on_duplicate_struct_key([{key, _} | rest], key) do
+    IO.warn(~s(duplicate key #{inspect(key)} found in struct))
```
whatyouhide:
```suggestion
    IO.warn("duplicate key #{inspect(key)} found in struct")
```

jherdman:
Perhaps it'd be wise to consider deprecating allowing such structs instead of merely warning against them?

lib/elixir/lib/kernel/utils.ex
Gazler:
```

  defp warn_on_duplicate_struct_key([]) do
    :ok
  end

  defp warn_on_duplicate_struct_key([{key, _} | [{key, _} | _] = rest]) do
    IO.warn("duplicate key #{inspect(key)} found in struct")
    warn_on_duplicate_struct_key(rest)
  end

  defp warn_on_duplicate_struct_key([_ | rest]) do
    warn_on_duplicate_struct_key(rest)
  end
```

wojtekmach:
Given we have this:

```
iex(1)> %{a: 1, a: 1}
** (CompileError) iex:1: key :a will be overridden in map
    (stdlib 3.10) lists.erl:1263: :lists.foldl/3
```

it makes sense to me too to deprecate now and raise in v2.0.

josevalim:
I think the warning is already enough to portray the deprecation? Or is the keyword "deprecated" relevant?

In any case, we should probably add a TODO:

    # TODO: Make it raise on v2.0

lib/elixir/lib/kernel/utils.ex
whatyouhide: Oh yes, good catch, but yeah so that we can only have one argument and don't have to start with `nil` as the first key :P)

whatyouhide:
Thanks @Gazler! 💟

PR #9455, opened 2019-10-29
Prevent calling non-async functions in async tests

fertapric:
Capturing a named device other than `:stdio`, such as `:stderr`, happens globally and requires `async: false`.

### DIFF
--- a/lib/eex/test/eex_test.exs
+++ b/lib/eex/test/eex_test.exs
@@ -44,7 +44,7 @@ defmodule Clause do
 end
 
 defmodule EExTest do
-  use ExUnit.Case, async: true
+  use ExUnit.Case
 
   doctest EEx
   doctest EEx.Engine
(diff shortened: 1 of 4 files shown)


lib/elixir/test/elixir/protocol/consolidation_test.exs
```diff
@@ -7,7 +7,7 @@ files = Path.wildcard(PathHelpers.fixture_path("consolidation/*"))
 Kernel.ParallelCompiler.compile_to_path(files, path)
 
 defmodule Protocol.ConsolidationTest do
-  use ExUnit.Case, async: true
```
fertapric: The test "consolidation prevents new implementations" uses `ExUnit.CaptureIO.capture_io(:stderr`. 

I thought about moving that test to `warning_test.exs` too, but it required some of the setup done at the beginning of this module, and I thought moving all that setup was not a good idea.

lib/eex/test/eex_test.exs
```diff
@@ -44,7 +44,7 @@ defmodule Clause do
 end
 
 defmodule EExTest do
-  use ExUnit.Case, async: true
+  use ExUnit.Case
```
fertapric: A few lines below:

https://github.com/elixir-lang/elixir/blob/6a8fa2c7ef100db4e115a4e9fcfa0c028cc1d147/lib/eex/test/eex_test.exs#L226-L237

josevalim:
Just to align what we discussed, I believe it would be better to make stderr to work with concurrent watchers, similar to capture_log.

fertapric:
This is merged temporarily to avoid blocking the Cirrus CI migration. Some changes on this PR will be undone once #9527 is merged.

PR #9456, opened 2019-10-29
Add additional documentation for ExUnit.CaptureIO

devonestes:
The information about `:stderr` and `:stdio` being aliases for
`:standard_error` and `:standard_io` wasn't included in this
documentation, so I added that here. I also added some typespecs for
documentation on how users can use these functions since the function
stubs in ExDoc didn't include all of the possible ways to call them.

### DIFF
--- a/lib/ex_unit/lib/ex_unit/capture_io.ex
+++ b/lib/ex_unit/lib/ex_unit/capture_io.ex
@@ -40,6 +40,18 @@ defmodule ExUnit.CaptureIO do
   A developer can set a string as an input. The default input
   is an empty string (which is equivalent to `:eof`).
 
+  ## IO devices
+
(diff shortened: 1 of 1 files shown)


josevalim:
Thanks @devonestes for the PR. Unfortunately CI is failing because of the formatter. Can you please run `bin/elixir bin/mix format` and push again?

devonestes:
Formatted and pushed!

whatyouhide:
Do we have specs on different clauses (put before each clause) anywhere else in Elixir? I believe we usually go for a bodiless clause preceded by all specs.

josevalim:
@whatyouhide FWIW, the current approach is fine (IMO).

whatyouhide:
Thanks @devonestes! 💟

PR #9457, opened 2019-10-29
Warn if duplicate key is found in record

wojtekmach:
Similar to #9452 :D

Worth noting erlc errors on this too:

```
-record(r, {a, a}).
```

src/m.erl:2: field a already defined in record r

### DIFF
--- a/lib/elixir/test/elixir/kernel/warning_test.exs
+++ b/lib/elixir/test/elixir/kernel/warning_test.exs
@@ -1774,7 +1774,7 @@ defmodule Kernel.WarningTest do
     assert capture_err(fn ->
              Code.eval_string("""
              defmodule TestMod do
-               defstruct [:foo, :foo, :bar]
+               defstruct [:foo, :bar, foo: 1]
              end
              """)
            end) =~ "duplicate key :foo found in struct"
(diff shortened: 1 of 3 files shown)


lib/elixir/test/elixir/record_test.exs
```diff
@@ -262,6 +263,15 @@ defmodule RecordTest do
     end
   end
 
+  test "duplicate keys" do
+    assert capture_io(:stderr, fn ->
```
fertapric: Capturing a named device other than `:stdio`, such as `:stderr`, happens globally and requires `async: false`.

In this case, I would move the test to the `warning_test.exs` file.

wojtekmach: Ah, right. Fixed!

lib/elixir/test/elixir/kernel/warning_test.exs
```diff
@@ -1774,14 +1774,27 @@ defmodule Kernel.WarningTest do
...
-               defstruct [:foo, :foo, :bar]
+               defstruct [:foo, :bar, foo: 1]
              end
              """)
            end) =~ "duplicate key :foo found in struct"
   after
     purge(TestMod)
   end
 
+  test "duplicate keys" do
```
fertapric: Nitpick:

```suggestion
  test "defrecord warns with duplicate keys" do
```

josevalim:

PR #9462, opened 2019-10-31
Add .elixir_ls to generated gitignore

samphilipd:


josevalim:
Thanks @samphilipd. However, code editor specific .gitignore belong to your global gitignore. Elixir's built-in .gitignore would become quite large if we were to consider tmp files used by all editors.

whatyouhide:
For context: https://www.google.com/search?client=safari&rls=en&q=global+gitignore&ie=UTF-8&oe=UTF-8

PR #9463, opened 2019-10-31
Avoid second calculation of arity in expand_local/4

ggcampinho:


josevalim:

Issue #9464, opened 2019-11-01
Do not use error_logger throughout Elixir


josevalim:
We should use `logger` instead. This should be a straight-forward port, except for:

  * DynamicSupervisor - we need to copy how the callback is invoked in OTP's Supervisor
  * Task - we need to emit a report (as in other OTP behaviours) and update the Logger.Translator 

/cc @hauleth

josevalim:
Closing in favor of #9469.

PR #9469, opened 2019-11-02
Remove calls to error_logger

hauleth:
Related to #9464

### DIFF
--- a/lib/elixir/lib/gen_server.ex
+++ b/lib/elixir/lib/gen_server.ex
@@ -788,14 +788,16 @@ defmodule GenServer do
         :logger.error(
           %{
             label: {GenServer, :no_handle_info},
-            module: __MODULE__,
-            message: msg,
-            name: proc
+            report: %{
+              module: __MODULE__,
+              message: msg,
+              name: proc
+            }
           },
           %{
             domain: [:otp, :elixir],
             error_logger: %{tag: :error_msg},
-            report_cb: &GenServer.format_log/1
+            report_cb: &GenServer.format_report/1
           }
         )
 
@@ -1221,7 +1223,10 @@ defmodule GenServer do
   @doc false
   # TODO: support 2nd argument with maximum length and depth of the displayed
   # message
-  def format_log(%{label: {GenServer, :no_handle_info}, module: mod, message: msg, name: proc}) do
+  def format_report(%{
+        label: {GenServer, :no_handle_info},
+        report: %{module: mod, message: msg, name: proc}
+      }) do
     {'~p ~p received unexpected message in handle_info/2: ~p~n', [mod, proc, msg]}
   end
 end
--- a/lib/elixir/lib/gen_server.ex
+++ b/lib/elixir/lib/gen_server.ex
@@ -785,8 +785,20 @@ defmodule GenServer do
             {_, name} -> name
           end
 
-        pattern = '~p ~p received unexpected message in handle_info/2: ~p~n'
-        :error_logger.error_msg(pattern, [__MODULE__, proc, msg])
+        :logger.error(
+          %{
+            label: {GenServer, :no_handle_info},
+            module: __MODULE__,
+            message: msg,
+            name: proc
+          },
+          %{
+            domain: [:otp, :elixir],
+            error_logger: %{tag: :error_msg},
+            report_cb: &GenServer.format_log/1
+          }
+        )
+
         {:noreply, state}
       end
 
@@ -1204,5 +1216,12 @@ defmodule GenServer do
 
   def whereis({name, node} = server) when is_atom(name) and is_atom(node) do
     server
+  end
+
+  @doc false
+  # TODO: support 2nd argument with maximum length and depth of the displayed
+  # message
+  def format_log(%{label: {GenServer, :no_handle_info}, module: mod, message: msg, name: proc}) do
+    {'~p ~p received unexpected message in handle_info/2: ~p~n', [mod, proc, msg]}
   end
 end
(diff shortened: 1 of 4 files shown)


lib/elixir/lib/gen_server.ex
```diff
@@ -1205,4 +1217,11 @@ defmodule GenServer do
   def whereis({name, node} = server) when is_atom(name) and is_atom(node) do
     server
   end
+
+  @doc false
+  # TODO: support 2nd argument with maximum length and depth of the displayed
+  # message
```
josevalim: Shall we tackle this TODO now? I would like to avoid addnig new TODOs to the code. :)

lib/elixir/lib/task/supervised.ex
```diff
@@ -96,24 +96,51 @@ defmodule Task.Supervised do
...
+              args: args,
+              reason: {log_value(kind, value), __STACKTRACE__}
+            }
+          },
+          %{
+            domain: [:otp, :elixir],
+            error_logger: %{tag: :error_msg},
+            report_cb: &__MODULE__.format_report/2
+          }
+        )
```
josevalim: Fantastic work!

lib/elixir/lib/gen_server.ex
```diff
@@ -785,8 +785,20 @@ defmodule GenServer do
...
+          %{
+            label: {GenServer, :no_handle_info},
+            module: __MODULE__,
+            message: msg,
+            name: proc
+          },
+          %{
+            domain: [:otp, :elixir],
+            error_logger: %{tag: :error_msg},
+            report_cb: &GenServer.format_log/1
```
josevalim: Is there any reason why we don't pass this as `&format_log/1` and keep `format_log/1` private?

lib/elixir/lib/task/supervised.ex
```diff
@@ -96,24 +96,51 @@ defmodule Task.Supervised do
...
+              starter: self(),
+              function: fun,
+              args: args,
+              reason: {log_value(kind, value), __STACKTRACE__}
+            }
+          },
+          %{
+            domain: [:otp, :elixir],
+            error_logger: %{tag: :error_msg},
+            report_cb: &__MODULE__.format_report/2
```
josevalim: We should keep this one private too and make the name consistent between files (format_log or format_report).

lib/elixir/lib/gen_server.ex
hauleth: Not really. I do not know why I did it that way.

lib/elixir/lib/gen_server.ex
hauleth: I will try to tackle it before merging.

lib/elixir/lib/task/supervised.ex
hauleth: From what I see, almost all code in OTP uses remote calls to the `report_cb`, probably the reasoning is the same as in `telemetry`, as this is clearer for the VM.

lib/elixir/lib/gen_server.ex
hauleth: I decided to ignore it as OTP in general ignores it as well.

lib/elixir/lib/task/supervised.ex
josevalim: We need to have a reasoning though as doing it just in case is not good enough. Especially because it exports a public API that we would need to `@doc false`. So I would:

1. Use local functions for now
2. Verify with the OTP team why they have used remote ones

Can you please check those? Thanks.

michalmuskala: Using an external function means the whole second argument will be a compile-time literal. A local function can't be optimised like that.

lib/logger/lib/logger/translator.ex
```diff
@@ -79,22 +79,34 @@ defmodule Logger.Translator do
...
-            [formatted, "\nFunction: #{inspect(function, opts)}"] ++
-            ["\n    Args: #{inspect(args, opts)}"]
+    msg =
+      ["Task #{inspect(name)} started from #{inspect(starter)} terminating"] ++
+        [formatted, "\nFunction: #{inspect(function, opts)}"] ++
+        ["\n    Args: #{inspect(args, opts)}"]
 
-        {:ok, msg, metadata}
+    {:ok, msg, metadata}
+  end
```
hauleth: I have ported it as is, but is there any reason why do that in translator instead of using `report_cb` directly? I think it would make more sense to have uniform error reports.

josevalim: that Is a separate discussion. Unifying reports would be a good option but we would definitely have to push some things uostream before. For example, we include more information on debug level.

josevalim:
Ok, let’s go with remote calls then, thanks Michał. Btw, any reason why we
can’t convert local ones to literals?
-- 


michalmuskala:
The main technical issue is that it's not possible to represent "static" local functions in the external term format that is used for encoding the literal chunk in the BEAM files. 

Local funs that don't capture the environment (are not closures) are convered to literals when loaded, but this still means they can't be embedded within other literals.

lib/logger/lib/logger/translator.ex
hauleth: @josevalim you mean in `inspect`? This is what 2nd argument in `report_cb` is for, to pass additional informations to formatters.

josevalim: I mean we show more information in general. Compare both reports in debug mode for a crashed GenServer and you will see the differences. :)

PR #9470, opened 2019-11-02
Add is_struct/1 to Kernel

hauleth:
Rel #6611


hauleth:
I have left out the `GenEvent` module out as I didn't want to change deprecated modules.

lib/elixir/lib/kernel.ex
```diff
@@ -2198,6 +2198,54 @@ defmodule Kernel do
     :erlang.error(ArgumentError.exception(error_message))
   end
 
+  @doc """
+  Returns true if `term` is a structure; otherwise returns `false`.
```
eksperimental:
```suggestion
  Returns true if `term` is a struct; otherwise returns `false`.
```
We never use the word "structure".  The only time we use is it in `Kernel.defstructure/2` and it is for something completely different.

lib/elixir/lib/kernel.ex
```diff
@@ -2198,6 +2198,54 @@ defmodule Kernel do
...
+      :match ->
+        invalid_match!(:is_struct)
+
+      :guard ->
+        quote(do: is_atom(:erlang.map_get(unquote(term), :__struct__)))
+    end
+  end
+
+  @doc """
+  Returns true if `term` is a structure; otherwise returns `false`.
```
eksperimental:
```suggestion
  Returns true if `term` is a struct; otherwise returns `false`.
```

lib/elixir/lib/kernel.ex
```diff
@@ -2198,6 +2198,54 @@ defmodule Kernel do
...
 
+  @doc """
+  Returns true if `term` is a structure; otherwise returns `false`.
+
+  Allowed in guard tests.
+  """
+  @doc guard: true
+  defmacro is_struct(term) do
+    case __CALLER__.context do
+      nil ->
```
josevalim: This clause can be the same as the `:guard` one. We just need to check we are not inside a match.

josevalim: Actually, ignore this comment for now.

lib/elixir/lib/kernel.ex
```diff
@@ -2198,6 +2198,54 @@ defmodule Kernel do
...
+            %_{} -> true
+            _ -> false
+          end
+        end
+
+      :match ->
+        invalid_match!(:is_struct)
+
+      :guard ->
+        quote(do: is_atom(:erlang.map_get(unquote(term), :__struct__)))
```
josevalim: Sinec this function starts with the `is_prefix`, it should not fail. You should write a check like this:

```elixir
is_struct(map) or is_map(map)
```

And then pass it a map. Because you will call map_get on a map without the `__struct__`, the whole clause will fail but it shouldn't. The proper guard is:

    is_map(unquote(term)) and :erlang.is_map_key(:__struct__, unquote(term)) and is_atom(:erlang.map_get(:__struct__, unquote(term)))

lib/elixir/lib/kernel.ex
hauleth: It would need to be `is_map(term) and :erlang.map_get(key, term)` to prevent errors on non-map `term`

lib/elixir/lib/kernel.ex
```diff
@@ -2198,6 +2198,54 @@ defmodule Kernel do
...
+            %^struct{} when is_atom(struct) -> true
+            _ -> false
+          end
+        end
+
+      :match ->
+        invalid_match!(:is_struct)
+
+      :guard ->
+        quote(do: :erlang.map_get(unquote(term), :__struct__) == unquote(struct))
```
josevalim: Please apply the comments from above here. :)

lib/elixir/lib/module/types/infer.ex
```diff
@@ -246,6 +246,7 @@ defmodule Module.Types.Infer do
     {:is_port, 1} => {[:port], :boolean},
     {:is_reference, 1} => {[:reference], :boolean},
     {:is_tuple, 1} => {[:tuple], :boolean},
+    {:map_get, 2} => {[{:map, []}, :dynamic], :dynamic},
```
josevalim: Per my comments above, is_map_key should be added too once the fix is pushed. :)

josevalim:
Thank you so much for looking into this, I have dropped some comments.

hauleth:
I will update all of these later, probably tomorrow morning.

lib/elixir/lib/kernel.ex
```diff
@@ -2198,6 +2198,54 @@ defmodule Kernel do
...
+    case __CALLER__.context do
+      nil ->
+        quote do
+          case unquote(term) do
+            %_{} -> true
+            _ -> false
+          end
+        end
+
+      :match ->
```
eksperimental: nit picking here. but I would leave the `:match` clause for the end as it is the only that is the one that it will be the least used

lib/elixir/lib/kernel.ex
eksperimental: > This clause can be the same as the `:guard` one. We just need to check we are not inside a match.

if that's the case, wouldn't it be better off to inline this function?

josevalim:
Also, if the CI is failing, please rebase as I accidentaly broke master a couple minutes ago.

lib/elixir/lib/kernel.ex
josevalim: > It would need to be is_map(term) and :erlang.map_get(key, term) to prevent errors on non-map term

Yeah, we also need this in the other one. But this code is simpler to emit and verify for the non-guard case, so let's please keep this one as is.

josevalim: > @josevalim if that's the case, wouldn't it be better off to inline this function?

We only inline when there is a direct equivalent in Erlang. There isn't one here.

eksperimental: > > @josevalim if that's the case, wouldn't it be better off to inline this function?
> 
> We only inline when there is a direct equivalent in Erlang. There isn't one here.

Oh! good to know

lib/elixir/test/elixir/kernel_test.exs
```diff
@@ -247,6 +247,30 @@ defmodule KernelTest do
...
+    assert check_struct(%Macro.Env{}) == :struct
+    assert check_struct([]) == :unknown
+    assert check_struct(%{}) == :unknown
+  end
+
+  test "is_struct/2" do
+    assert is_struct(%{}, Macro.Env) == false
+    assert is_struct(%Macro.Env{}, Macro.Env) == true
+    assert is_struct(%Macro.Env{}, DateTime) == false
+    assert is_struct([], Macro.Env) == false
```
josevalim: We need to check what happens when you don't pass an atom as second argument. Ideally we would want to raise badarg, but that means we would need to make the guard fail instead of returning false and that is not easy to emulate.

Any ideas on how we could emulate it? If we can't, then we will need to postpone is_struct/2 for now. :/

eksperimental: > We need to check what happens when you don't pass an atom as second argument. Ideally we would want to raise badarg, but that means we would need to make the guard fail instead of returning false and that is not easy to emulate.
> 
> Any ideas on how we could emulate it? If we can't, then we will need to postpone is_struct/2 for now. :/

Why can't we just check for `is_atom(struct)` in https://github.com/elixir-lang/elixir/pull/9470/files#diff-e06d5ebc2907d951188e88476f85af51R2235 and if not raise `BadArgument`?
Should we care in the `:guard` for this in the guard context, since any function that raises is considered `false`?

lib/elixir/lib/kernel.ex
```diff
@@ -2198,6 +2198,54 @@ defmodule Kernel do
...
+        invalid_match!(:is_struct)
+
+      :guard ->
+        quote(do: is_atom(:erlang.map_get(:__struct__, unquote(term))))
+    end
+  end
+
+  @doc """
+  Returns true if `term` is a struct; otherwise returns `false`.
```
eksperimental: the docs are missing what the second argument does.

lib/elixir/lib/kernel.ex
```diff
@@ -2198,6 +2198,54 @@ defmodule Kernel do
...
+
+  Allowed in guard tests.
+  """
+  @doc guard: true
+  defmacro is_struct(term, struct) do
```
eksperimental: I think we should rename the second argument to `struct_type` or `struct_name`, because it is confusion, and it could be interpreted as an actual struct is needed to be passed. Actually, `Kernel.defstruct/2` works this very same way, it accepts its second argument `struct` either an atom or a struct.

lib/elixir/test/elixir/kernel_test.exs
josevalim: Because nothing but an atom is a valid struct name. So giving anything other, should be a bad argument. For example, is_function/2 raises when the second argument is not an integer.

eksperimental: > Because nothing but an atom is a valid struct name. So giving anything other, should be a bad argument. For example, is_function/2 raises when the second argument is not an integer.

I know that, and in fact I run `is_function/2` before writing my comment, but what prevents us from running `is_atom(struct)` and raising a `BarArgument` error if struct is not an atom. Or am I missing something?

josevalim:
We can’t raise from a guard (or emulate the conditional logic of: do this
or fail easily).
-- 


eksperimental:
@josevalim could you please have a look at my code?
https://github.com/eksperimental/elixir/tree/hauleth_feat/is_struct
all tests are passing

lib/elixir/lib/kernel.ex
hauleth: Failures in guards aren't errors, but just match failure. You can check this out via:

```elixir
def foo(list) when hd(list) == 1, do: :ok
def foo(other), do: :error

foo(10) == :error
```

hauleth: Added test for that and fixed it.

lib/elixir/test/elixir/kernel_test.exs
hauleth: @josevalim the only way I know to emulate such behaviour would be to:

```
(is_atom(struct) or hd([]))
and :erlang.map_get(:__struct__, map) == struct
```

josevalim: @hauleth yeah, good call. I thought about `length(:foo)` but I am really uncomfortable with said approach. It can throw off code analysis tools and what not. Let's skip `is_struct/2` for now.

lib/elixir/lib/kernel.ex
```diff
@@ -2198,6 +2198,30 @@ defmodule Kernel do
...
+            %_{} -> true
+            _ -> false
+          end
+        end
+
+      :match ->
+        invalid_match!(:is_struct)
+
+      :guard ->
+        quote(do: is_map(unquote(term)) and is_atom(:erlang.map_get(:__struct__, unquote(term))))
```
josevalim: We are missing the `:erlang.is_map_key(:__struct__, unquote(term))` check here.

To validate, write this test:

    defp check_struct_or_map(arg) when is_struct(arg) or is_map(arg), do: true
    defp check_struct_or_map(_arg), do: false

It should return true for a map, but it will return false.

hauleth: Done.

lib/elixir/test/elixir/kernel_test.exs
```diff
@@ -247,6 +247,30 @@ defmodule KernelTest do
...
 
+  defp check_struct(arg) when is_struct(arg), do: true
+  defp check_struct(_arg), do: false
+
+  defp check_struct_or_map(arg) when is_struct(arg) or is_map(arg), do: true
+  defp check_struct_or_map(_arg), do: false
+
+  test "is_struct/1" do
+    assert is_struct(%{}) == false
+    assert is_struct(%{}) == false
```
eksperimental:
```suggestion
    assert is_struct([]) == false
```
You probably meant `[]`; otherwise that line is duplicated

lib/elixir/lib/module/types/infer.ex
```diff
@@ -246,6 +246,8 @@ defmodule Module.Types.Infer do
     {:is_port, 1} => {[:port], :boolean},
     {:is_reference, 1} => {[:reference], :boolean},
     {:is_tuple, 1} => {[:tuple], :boolean},
+    {:map_get, 2} => {[:dynamic, {:map, []}], :dynamic},
+    {:is_map_key, 2} => {[:dynamic, {:map, []}], :dynamic},
```
eksperimental: I would sort these entries alphabetically

eksperimental:
Before this is merged, shouldn't this function be placed after ` def is_reference(term) do
` so it is sorted alphabetically?

josevalim:
@eksperimental we don't sort functions in modules alphabetically. Plus the functions that come earlier in this module are generally the ones that we inline. :)

josevalim:

lib/elixir/lib/kernel.ex
```diff
@@ -2198,6 +2198,33 @@ defmodule Kernel do
     :erlang.error(ArgumentError.exception(error_message))
   end
 
+  @doc """
+  Returns true if `term` is a struct; otherwise returns `false`.
+
+  Allowed in guard tests.
+  """
+  @doc guard: true
+  defmacro is_struct(term) do
```
wojtekmach: out of curiosity, why is this implementation preferred over:

```elixir
  defguard is_struct(term)
           when is_map(term) and :erlang.is_map_key(:__struct__, term) and
                  is_atom(:erlang.map_get(:__struct__, term))
```

is it more efficient in some cases?

Or because:

```
# foo.exs
is_struct(42)
```

warns:

```
warning: this expression will fail with BadMapError
  foo.exs:1
```

?

Or both? Or something else entirely?

hauleth: To avoid warning, as it would provide inconsistent behaviour when compared to the rest of `is_*/1` functions.

wojtekmach: gotcha, thanks!

PR #9474, opened 2019-11-03
Avoid race condition in test

lau:
A test is failing and it looks like the test assumes that recompilation takes at least 1 second. Because the mtime has a granularity of 1 second. Some times the compared times are equal and fails the test. Example: https://travis-ci.org/elixir-lang/elixir/jobs/606593844#L285

This adds a 1 second delay in the test.

### DIFF
--- a/lib/mix/test/mix/umbrella_test.exs
+++ b/lib/mix/test/mix/umbrella_test.exs
@@ -94,7 +94,7 @@ defmodule Mix.UmbrellaTest do
     end)
   end
 
-  test "recompiles umbrella on conffig hange" do
+  test "recompiles umbrella on config change" do
     in_fixture("umbrella_dep/deps/umbrella", fn ->
       Mix.Project.in_project(:umbrella, ".", fn _ ->
         Mix.Task.run("compile", [])
--- a/lib/mix/test/mix/umbrella_test.exs
+++ b/lib/mix/test/mix/umbrella_test.exs
@@ -98,15 +98,15 @@ defmodule Mix.UmbrellaTest do
     in_fixture("umbrella_dep/deps/umbrella", fn ->
       Mix.Project.in_project(:umbrella, ".", fn _ ->
         Mix.Task.run("compile", [])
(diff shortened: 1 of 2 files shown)


josevalim:
The ensure_touched call above should wait long enough, so we don’t need to
sleep. So we would need to investigate why it isn’t. :)
-- 


*José Valim*
www.plataformatec.com.br
Founder and Director of R&D

lau:
I see. I have changed the PR and added sleep in ensure_touched instead. This text says that `mtime` cannot be relied on: https://apenwarr.ca/log/20181113

Could it be that the file system is just not that precise for this purpose?

With the recursion calling touch and stat thousands of times until there is a difference I am thinking that the timing is very close. And if there is a slight Network Time Protocol (NTP) clock adjustment after the function returns this could cause the issue.

Sleeping should add some margin of error for the combination of mtime, touch and the filesystem.

fertapric:
Btw, it might be a good idea to rename `ensure_touched` to `ensure_touched_after` or similar 🙂

josevalim:
@fertapric good call. We don't compare the mtimes of the beam files, but we do compare with the manifest. We need to pass that as initial value.

josevalim:

fertapric:
The `mix.exs` file is created first, so `ensure_touched` might only need one loop to actually touch that file. If that happens, it would mean that the two `compile` are executed almost right before the other, which might cause the test to fail. For example:

1. The `mix.exs` file is created on second 1.
2. `compile` is run and it ends on second 2. The `mtime` of the BEAM files reflect the second 2
3. The `mix.exs` file is touched getting a `mtime` on second 2 (only one `File.touch!` needed).
4. `compile` is run again and it ends again on second 2. The `mtime` of the BEAM files reflect again the second 2

I think this could be a possible fix:

```elixir
bar = File.stat!("_build/dev/lib/bar/ebin/Elixir.Bar.beam").mtime
foo = File.stat!("_build/dev/lib/foo/ebin/Elixir.Foo.beam").mtime

ensure_touched("mix.exs", max(foo, bar))
```

PR #9476, opened 2019-11-03
Add EditorConfig to provide uniform indent rules

hauleth:
This is especially needed for Erlang where most common style use 4 spaces instead of 2 like Elixir code does.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -9,7 +9,7 @@ charset = utf-8
 indent_style = space
 indent_size = 2
 
-[*.erl]
+[*.{erl,hrl}]
 indent_style = space
 indent_size = 2
 
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,17 @@
+root = true
+
+[*]
+end_of_line = true
+insert_final_newline = true
+charset = utf-8
+
+[*.{ex,exs}]
+indent_style = space
+indent_size = 2
+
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -0,0 +1,17 @@
...
+[*]
+end_of_line = true
+insert_final_newline = true
+charset = utf-8
+
+[*.{ex,exs}]
+indent_style = space
+indent_size = 2
+
+[*.erl]
```
michalmuskala: This should probably be `*.{erl,hrl}`

hauleth: Yes, it should, and now is.

josevalim:
Thanks @hauleth! Last time we had a discussion on this we agreed that it was not yet generally adopted to justify adding another file to the repo root. We will keep paying attention and adopt it once it sees major editor support

I do realize the catch-22 in that if everyone adopted it, then everyone would adopt it faster, but right now it is really not at point we think it should be. :)

hauleth:
@josevalim for me it was done out of convenience as Vim supports a lot of commands that work on indentation (`<`, `>`, `i_^d`, `i_^t`) and these fails miserably with my defaults (4 spaces as this is what almost everyone is using). My thoughts are - I am using it, maybe others will also gain something, especially when they are newcomers to the Elixir source code.

PR #9478, opened 2019-11-03
Handle case when report_cb need 2 arguments

hauleth:


josevalim:
Thanks @hauleth! Can we add a test case for this? Please look at `translator_test.exs` for examples.

hauleth:
Done

josevalim:
Thanks. Btw, for reference, can you add a link to some OTP code that uses
report_cb/2?
-- 


hauleth:
[`proc_lib`](https://github.com/erlang/otp/blob/f489df9deb37878552c7573ac4b05d392265d1fa/lib/stdlib/src/proc_lib.erl#L512) is using `proc_lib:report_cb/2` function

josevalim:

PR #9482, opened 2019-11-03
Add Examples section to Kernel.is_struct/1

eksperimental:


eksperimental:
Why the change? in any case, I think `%URI{}` will be easier to grasp

eksperimental:
`1..2` would work too

josevalim:
@eksperimental we should not be telling users to create structs like %URI{} directly.

josevalim:

PR #9483, opened 2019-11-03
Standardize of use #=> for return values in Kernel

eksperimental:


lib/elixir/lib/kernel.ex
```diff
@@ -4258,7 +4259,7 @@ defmodule Kernel do
 
       john = %User{name: "John"}
       MyProtocol.call(john)
-      ** (Protocol.UndefinedError) protocol MyProtocol not implemented for %User{...}
+      #=> ** (Protocol.UndefinedError) protocol MyProtocol not implemented for %User{...}
```
josevalim: This is not really a return value, so I would keep it as it was.

lib/elixir/lib/kernel.ex
```diff
@@ -4284,7 +4285,7 @@ defmodule Kernel do
   belongs to the struct:
 
       %User{name: "john", unknown: :key}
-      ** (KeyError) key :unknown not found in: %User{age: 21, name: nil}
+      #=> ** (KeyError) key :unknown not found in: %User{age: 21, name: nil}
```
josevalim: Same here and the exceptions below.

eksperimental: we were using this method for code that couldn't be put in doctests because otherwise the module wouldn't compile

eksperimental: somehow we need to differentiate what is input and what is output.
in doctests input is represented by `iex>`,

josevalim: I am aware. The only reason we use `#=>` is because it may not be clear to folks that `true` is a return value in something like:

    User.is_foo?(foo)
    true

So we do:

    User.is_foo?(foo)
    #=> true

But an exception is always clear that it is an exception. Adding #=> actually makes it confusing, because it is no longer clear if it is a return value or the exception. Using the same format as in doctest is actually a positive thing.

eksperimental: I agree, but I'd prefer to have the same style throughout the docs. the are other parts were we are using this, should we remove the `#=>` when an exception is thrown?

josevalim:
Yes, good call! Let’s remove #=> whenever it is used for exception
signaling.
-- 


eksperimental:
ok. I will update this PR accordingly and send a new PR removing the "#=>" from the exceptions

eksperimental:
updated!

whatyouhide:
Thank you @eksperimental!

PR #9484, opened 2019-11-03
Add Kernel.is_struct/2

eksperimental:
This is my take on how to make `is_struct/2` raise when not an atom is provided as its second argument.

The user cannot tell from the stacktrace that it is a macro working behind the scenes in case an ArgumentError is raised.

```elixir
defmodule A do
  def a(), do: {B.b()}
end

defmodule B do
  def b(), do: {C.c()}
end

defmodule C do
  def c(), do: {D.d()}
end

defmodule D do
  def d(), do: is_struct(1..2, 1)
end
```

```elixir
iex(1)> A.a()
** (ArgumentError) invalid struct_type attribute. It expected a module or an atom, got: 1
    (elixir 1.10.0-dev) /home/eksperimental/test_elixir/is_struct/lib/is_struct.ex:14: Kernel.is_struct/2
    (is_struct 0.1.0) lib/is_struct.ex:10: C.c/0
    (is_struct 0.1.0) lib/is_struct.ex:6: B.b/0
    (is_struct 0.1.0) lib/is_struct.ex:2: A.a/0
iex(1)> D.d
** (ArgumentError) invalid struct_type attribute. It expected a module or an atom, got: 1
    (elixir 1.10.0-dev) /home/eksperimental/test_elixir/is_struct/lib/is_struct.ex:14: Kernel.is_struct/2
    (stdlib 3.9.2) erl_eval.erl:680: :erl_eval.do_apply/6
    (iex 1.10.0-dev) lib/iex/evaluator.ex:260: IEx.Evaluator.handle_eval/5
    (iex 1.10.0-dev) lib/iex/evaluator.ex:240: IEx.Evaluator.do_eval/3
    (iex 1.10.0-dev) lib/iex/evaluator.ex:218: IEx.Evaluator.eval/3
```

### DIFF
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -2235,6 +2235,73 @@ defmodule Kernel do
   end
 
   @doc """
+  Returns true if `term` is a stuct of the `struct_type`; otherwise returns `false`.
+
+  `struct_type` is expected to be a module or an atom; otherwise it raises `ArgumentError`.
+
+  Allowed in guard tests.
+
+  ## Examples
+
+      is_struct(1..2, Range)
+      #=> true
+
+      is_struct(1..2, Date)
+      #=> false
+
+      is_struct(1..2, "Range")
+      #=> ** (ArgumentError) invalid struct_type attribute. It expected a module or an atom, got: "Range"
+
+  """
+  @doc since: "1.10.0", guard: true
+  defmacro is_struct(term, struct_type) do
+    case __CALLER__.context do
+      nil ->
+        quote do
+          __is_struct__(unquote(term), unquote(struct_type), unquote(Macro.escape(__CALLER__)))
+        end
+
+      :guard ->
+        quote do
+          is_map(unquote(term)) and
+            :erlang.is_map_key(:__struct__, unquote(term)) and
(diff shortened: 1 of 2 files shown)


josevalim:
@eksperimental the issue is that we want it to raise inside the guard. Basically, we want this:

    def foo(map) when is_struct(map, 1) or is_map(map), do: true

to **never** succeed.

josevalim:
As discussed in the other issue, it would only work for something like `is_atom(struct) or hd([])` but we are not comfortable with adding such bogus guards. In any case, thanks for the PR.

eksperimental:
I totally misunderstood the situation. I thought we DID NOT WANT it to raise in the guard!

hauleth:
Exactly what @josevalim said. We both were thinking about adding impossible case that would always fail, but that could be problematic and weird from the static analyser viewpoint. I think that the "simplest" solution would be to add `:erlang.fail()` allowed in guards which would always fail.

In fact we want to fail in guards, in some cases it is reasonable, just like `hd/1` or `tl/1`.

eksperimental:
if guard safe functions can raise, does anybody know why :erlang.raise/3 is not allowed in guards?

hauleth:
Because `throw` and `exit` doesn't make sense in guard, and there is nothing you can provide for `Stacktrace`.

PR #9485, opened 2019-11-03
Remove comment arrows from Exception messages

eksperimental:


fertapric:
Thanks @eksperimental! 💜

PR #9486, opened 2019-11-03
Improve module docs in Kernel and Kernel.SpecialForms

eksperimental:
Kernel:
- prefix is `Kernel.`, not `Kernel`;
- grammatical corrections and rewording.

Kernel.SpecialFroms:
- mention that the module consist of macros which are automatically imported and can be called
  without prefix.


eksperimental:
@fertapric please have a look at the second commit.

fertapric:
Re-ok’ed 🙂

PR #9488, opened 2019-11-04
Remove anonymous function examples

sigurdsa:
Since Elixir only evaluate Logger macro arguments when the message will be logged, we should remove these examples since they might confuse people reading the docs into thinking that they should still wrap their log statements.


josevalim:

PR #9489, opened 2019-11-05
Add missing doc annotations

wojtekmach:
Brought to you by https://github.com/wojtekmach/app_manifest#usage :-)


eksperimental:
Is there a reason why we only have annotated since 1.4 and not earlier?

wojtekmach:
I think we added `since` in 1.6 or so and my guess would be we only annotated functions from recent couple of versions as older ones didn’t even receive bug fixes anymore.

josevalim:

PR #9493, opened 2019-11-05
Improvement in calendar modules

eksperimental:


lib/elixir/lib/calendar/iso.ex
```diff
@@ -953,10 +1007,12 @@ defmodule Calendar.ISO do
...
-  def valid_date?(year, month, day) do
-    month in 1..12 and year in -9999..9999 and
-      (is_integer(day) and day >= 1 and day <= days_in_month(year, month))
+  @spec valid_date?(year :: integer, month :: integer, day :: integer) :: boolean
+  def valid_date?(year, month, day)
+      when is_integer(year) and is_integer(month) and is_integer(day) do
+    year >= -9999 and year <= 9999 and
+      month >= 1 and month <= 12 and
+      day >= 1 and day <= days_in_month(year, month)
   end
```
eksperimental: I bent the rules here and changed the specs, but if we want to maintain them, this function should be rewritten as :

```elixir
  @spec valid_date?(year, month, day) :: boolean
  def valid_date?(year, month, day) when is_year(year) and is_month(month) and is_day(day) do
    day <= days_in_month(year, month)
  end
```

tests should be changed, since `Calendar.ISO.valid_date?(-1, 12, 32)` would raise `FunctionClauseError`, but I still think it is the way to go since an int out of the range specified by the specs was provided. Let me know if I should change it

josevalim:
Hi @eksperimental!

Those changes look generally good, but I just want to be clear that, an argument that does not comply to the spec is not required to raise `FunctionClauseError`. It may raise `ArgumentError`, `CaseClauseError`, or nothing at all.

Therefore, I would honestly prefer to focus on code readability and performance, so I will make some comments below to this extent.

lib/elixir/lib/calendar/iso.ex
```diff
@@ -71,6 +91,24 @@ defmodule Calendar.ISO do
...
+  defguardp is_year(year) when year in -9999..9999
+  defguardp is_year_BCE(year) when year in -9999..0
+  defguardp is_year_CE(year) when year in 1..9999
+  defguardp is_month(month) when month in 1..12
+  defguardp is_day(day) when day in 1..31
+  defguardp is_hour(hour) when is_integer(hour) and hour >= 0
+  defguardp is_minute(minute) when is_integer(minute) and minute >= 0
+  defguardp is_second(second) when is_integer(second) and second >= 0
+
+  defguardp is_microsecond(microsecond)
```
josevalim: I would say a `tuple_size(microsecond) == 2` is enough as a check here.

lib/elixir/lib/calendar/iso.ex
```diff
@@ -669,13 +708,14 @@ defmodule Calendar.ISO do
 
   """
   @doc since: "1.8.0"
-  @spec year_of_era(year) :: {pos_integer(), era :: 0..1}
+  @spec year_of_era(year) :: {year, ce} when year: 1..9999
+  @spec year_of_era(year :: -9999..0) :: {1..10000, bce}
```
josevalim: Please revert. The focus of specs is documentation and this is less clear.

lib/elixir/lib/calendar/time.ex
```diff
@@ -98,37 +98,46 @@ defmodule Time do
       {:error, :invalid_time}
       iex> Time.new(23, 59, 60, 999_999)
       {:error, :invalid_time}
+
+      # Invalid microsecond
       iex> Time.new(23, 59, 59, 1_000_000)
-      {:error, :invalid_time}
+      ** (FunctionClauseError) no function clause matching in Time.new/5
```
josevalim: Please revert. It is backwards incompatible and inconsistent with the things like invalid seconds, which do not raise, but return a tuple.

lib/elixir/lib/calendar/time.ex
```diff
@@ -98,37 +98,46 @@ defmodule Time do
...
 
-      # Invalid precision
-      Time.new(23, 59, 59, {999_999, 10})
-      {:error, :invalid_time}
+      iex> Time.new(23, 59, 59, {999_999, 10})
+      ** (FunctionClauseError) no function clause matching in Time.new/5
```
josevalim: Please revert. It is backwards incompatible and inconsistent with the things like invalid seconds, which do not raise, but return a tuple.

lib/elixir/lib/calendar/time.ex
```diff
@@ -98,37 +98,46 @@ defmodule Time do
...
 
   """
   @spec new(
           Calendar.hour(),
           Calendar.minute(),
           Calendar.second(),
-          Calendar.microsecond() | integer,
+          Calendar.microsecond() | 0..999_999,
```
josevalim: Please revert. `0..999_999` applies to the Calendar.ISO, it may not apply to others.

lib/elixir/lib/calendar/time.ex
```diff
@@ -98,37 +98,46 @@ defmodule Time do
...
           Calendar.calendar()
         ) :: {:ok, t} | {:error, atom}
   def new(hour, minute, second, microsecond \\ {0, 0}, calendar \\ Calendar.ISO)
 
-  def new(hour, minute, second, microsecond, calendar) when is_integer(microsecond) do
+  def new(hour, minute, second, microsecond, calendar) when microsecond in 0..999_999 do
```
josevalim: Please revert. `0..999_999` applies to the Calendar.ISO, it may not apply to others.

lib/elixir/lib/calendar/time.ex
```diff
@@ -98,37 +98,46 @@ defmodule Time do
...
+        hour,
+        minute,
+        second,
+        {microsecond_value, microsecond_precision} = microsecond,
+        calendar
+      )
+      when is_integer(hour) and hour >= 0 and
+             is_integer(minute) and minute >= 0 and
+             is_integer(second) and second >= 0 and
+             microsecond_value in 0..999_999 and microsecond_precision in 0..6 do
```
josevalim: Please revert. All of these limitations apply to the Calendar.ISO, it may not apply to others. Probably best to not do changes to this file in this PR.

lib/elixir/lib/calendar/naive_datetime.ex
```diff
@@ -208,17 +210,33 @@ defmodule NaiveDateTime do
...
+        {microsecond_value, microsecond_precision} = microsecond,
+        calendar
+      )
+      when is_integer(year) and year >= 0 and
+             is_integer(minute) and minute >= 0 and
+             is_integer(second) and second >= 0 and
+             is_integer(hour) and hour >= 0 and
+             is_integer(minute) and minute >= 0 and
+             is_integer(second) and second >= 0 and
+             microsecond_value in 0..999_999 and microsecond_precision in 0..6 do
```
josevalim: Please revert. All of these limitations apply to the Calendar.ISO, it may not apply to others. Probably best to not do changes to this file in this PR.

lib/elixir/lib/calendar/naive_datetime.ex
```diff
@@ -194,8 +194,10 @@ defmodule NaiveDateTime do
       {:error, :invalid_time}
       iex> NaiveDateTime.new(2000, 1, 1, 23, 59, 60, 999_999)
       {:error, :invalid_time}
+
+      # Invalid microsecond (out of the valid range)
       iex> NaiveDateTime.new(2000, 1, 1, 23, 59, 59, 1_000_000)
-      {:error, :invalid_time}
+      ** (FunctionClauseError) no function clause matching in NaiveDateTime.new/8
```
josevalim: Please revert for the reasons said in `time.ex`.

lib/elixir/test/elixir/calendar/iso_test.exs
```diff
@@ -48,7 +46,7 @@ defmodule Calendar.ISOTest do
         Calendar.ISO.naive_datetime_to_iso_days(2018, 2, 30, 0, 0, 0, 0)
       end
 
-      assert_raise ArgumentError, "invalid date: 2017-11--03", fn ->
+      assert_raise FunctionClauseError, fn ->
```
josevalim: Please revert. If we are already checking the arguments inside, there is no need to do check it in the function clause. It is just more work without any benefits.

lib/elixir/test/elixir/calendar/iso_test.exs
```diff
@@ -60,7 +58,7 @@ defmodule Calendar.ISOTest do
         Calendar.ISO.day_of_week(2018, 2, 30)
       end
 
-      assert_raise ArgumentError, "invalid date: 2017-11-00", fn ->
```
josevalim: Please revert. If we are already checking the arguments inside, there is no need to do check it in the function clause. It is just more work without any benefits.

lib/elixir/test/elixir/calendar/iso_test.exs
```diff
@@ -72,7 +70,7 @@ defmodule Calendar.ISOTest do
         Calendar.ISO.day_of_era(2018, 2, 30)
       end
 
-      assert_raise ArgumentError, "invalid date: 2017-11-00", fn ->
```
josevalim: Please revert. If we are already checking the arguments inside, there is no need to do check it in the function clause. It is just more work without any benefits.

lib/elixir/test/elixir/calendar/iso_test.exs
```diff
@@ -84,12 +82,31 @@ defmodule Calendar.ISOTest do
         Calendar.ISO.day_of_year(2018, 2, 30)
       end
 
-      assert_raise ArgumentError, "invalid date: 2017-11-00", fn ->
```
josevalim: Please revert. If we are already checking the arguments inside, there is no need to do check it in the function clause. It is just more work without any benefits.

lib/elixir/lib/calendar/iso.ex
```diff
@@ -588,10 +628,10 @@ defmodule Calendar.ISO do
 
   """
   @doc since: "1.4.0"
-  @spec day_of_week(year, month, day) :: 1..7
+  @spec day_of_week(year, month, day) :: day_of_week
   @impl true
   def day_of_week(year, month, day)
-      when is_integer(year) and is_integer(month) and is_integer(day) do
+      when is_year(year) and is_month(month) and is_day(day) do
```
josevalim: Let's remove the guards here since we already check them on `date_to_iso_days` (if not, we should).

lib/elixir/lib/calendar/iso.ex
```diff
@@ -697,16 +737,16 @@ defmodule Calendar.ISO do
 
   """
   @doc since: "1.8.0"
-  @spec day_of_era(year, month, day) :: {day :: pos_integer(), era :: 0..1}
+  @spec day_of_era(year, month, day) :: {day, era}
   @impl true
   def day_of_era(year, month, day)
-      when is_integer(year) and is_integer(month) and is_integer(day) and year > 0 do
+      when is_year_CE(year) and is_month(month) and is_day(day) do
     day = date_to_iso_days(year, month, day) - @iso_epoch + 1
```
josevalim: Let's remove the guards here since we already check them on `date_to_iso_days` (if not, we should).

lib/elixir/lib/calendar/iso.ex
```diff
@@ -697,16 +737,16 @@ defmodule Calendar.ISO do
...
     {day, 1}
   end
 
   def day_of_era(year, month, day)
-      when is_integer(year) and is_integer(month) and is_integer(day) and year < 1 do
+      when is_year_BCE(year) and is_month(month) and is_day(day) do
```
josevalim: Let's remove the guards here since we already check them on `date_to_iso_days` (if not, we should).

lib/elixir/lib/calendar/iso.ex
```diff
@@ -837,7 +885,9 @@ defmodule Calendar.ISO do
         microsecond,
         format \\ :extended
       )
-      when format in [:basic, :extended] do
+      when is_year(year) and is_month(month) and is_day(day) and is_hour(hour) and
+             is_minute(minute) and is_second(second) and is_microsecond(microsecond) and
+             format in [:basic, :extended] do
```
josevalim: Remove duplicate check (date_to_string and time_to_string already does them).

lib/elixir/lib/calendar/iso.ex
```diff
@@ -906,7 +956,11 @@ defmodule Calendar.ISO do
         std_offset,
         format \\ :extended
       )
-      when format in [:basic, :extended] do
+      when is_year(year) and is_month(month) and is_day(day) and is_hour(hour) and
+             is_minute(minute) and is_second(second) and is_microsecond(microsecond) and
+             is_time_zone(time_zone) and is_zone_abbr(zone_abbr) and is_utc_offset(utc_offset) and
+             is_std_offset(std_offset) and
```
josevalim: Remove duplicate checks (date_to_string and time_to_string already does them).

lib/elixir/lib/calendar/time.ex
eksperimental: Bear in mind that `0..999_999` comes from `t:Calendar.microseconds/0` which is what  
`Calendar.ISO` uses.

lib/elixir/lib/calendar/iso.ex
josevalim: Yeah, I think this is a good change. Good call!

lib/elixir/lib/calendar/iso.ex
```diff
@@ -980,9 +1036,10 @@ defmodule Calendar.ISO do
   @impl true
   @spec valid_time?(Calendar.hour(), Calendar.minute(), Calendar.second(), Calendar.microsecond()) ::
           boolean
-  def valid_time?(hour, minute, second, {microsecond, precision}) do
-    hour in 0..23 and minute in 0..59 and second in 0..59 and microsecond in 0..999_999 and
-      precision in 0..6
+  def valid_time?(hour, minute, second, microsecond)
+      when is_integer(hour) and hour >= 0 and is_integer(minute) and minute >= 0 and
```
josevalim: Let's only keep the type check in guards. `hour >= 0` should be moved to the body. No need for `is_microsecond` here too, let's pattern match and assert the contents without raising.

lib/elixir/lib/calendar/time.ex
josevalim: That should be eventually updated then. I will tackle it later.

eksperimental: then should this spec be `Calendar.microsecond() | non_neg_integer,` ? instead of `interger()` ?

lib/elixir/lib/calendar/iso.ex
eksperimental: is it ok if `tuple_size/1` raises within the guard if no tuple is provided?

wojtekmach: Guards don't raise, otherwise we wouldn't be able to compose them with an `or` or have different function clauses that have guards.

lib/elixir/lib/calendar/iso.ex
```diff
@@ -953,10 +1007,12 @@ defmodule Calendar.ISO do
...
-  @spec valid_date?(year, month, day) :: boolean
-  def valid_date?(year, month, day) do
-    month in 1..12 and year in -9999..9999 and
-      (is_integer(day) and day >= 1 and day <= days_in_month(year, month))
+  @spec valid_date?(year :: integer, month :: integer, day :: integer) :: boolean
+  def valid_date?(year, month, day)
+      when is_integer(year) and is_integer(month) and is_integer(day) do
+    year >= -9999 and year <= 9999 and
+      month >= 1 and month <= 12 and
+      day >= 1 and day <= days_in_month(year, month)
```
wojtekmach: I think it's fine as is, but this would be slightly shorter:

```elixir
year in -9999..9999 and
  month in 1..12 and
  day in 1..days_in_month(year, month)
```

josevalim:
And we can trust the Erlang compiler to avoid doing the same check twice as
long as it happens within the same function.
-- 


lib/elixir/lib/calendar/iso.ex
eksperimental: I love the way you used it in a range. that's great. thank you @wojtekmach

lib/elixir/lib/calendar/iso.ex
eksperimental: > Yeah, I think this is a good change. Good call!

@josevalim are you refering to my initial code change, or to my later comment https://github.com/elixir-lang/elixir/pull/9493/files#r342723097 ? I was a bit confused by this as I thought you were answering to my comment

lib/elixir/lib/calendar/iso.ex
eksperimental: done. please have a look

lib/elixir/lib/calendar/naive_datetime.ex
eksperimental: done

lib/elixir/lib/calendar/naive_datetime.ex
eksperimental: We need to review the specs then, because they are refering to Calendar types

lib/elixir/lib/calendar/time.ex
eksperimental: done

lib/elixir/lib/calendar/time.ex
eksperimental: reverted

lib/elixir/test/elixir/calendar/iso_test.exs
eksperimental: done

lib/elixir/test/elixir/calendar/iso_test.exs
eksperimental: done

lib/elixir/test/elixir/calendar/iso_test.exs
eksperimental: done

lib/elixir/lib/calendar/iso.ex
eksperimental: > Guards don't raise, otherwise we wouldn't be able to compose them with an `or` or have different function clauses that have guards.

@wojtekmach Thanks. I know that. That's why i said "within", meaning that they raise internally, but wondering if that would be considered a bad practice.
Anyway reading the Guard section I found the answer in the documention
> In the code above, we used length/1 to both check that the given thing is a list and check some properties of its length (instead of using is_list(something) and length(something) > 0).

eksperimental:
> And we can trust the Erlang compiler to avoid doing the same check twice as long as it happens within the same function. -- *José Valimwww.plataformatec.com.br <http://www.plataformatec.com.br/>Founder and Director of R&D*

that's awesome. I didn't know that.
I followed this advice to improve the readability of the code. thank you @josevalim

lib/elixir/lib/calendar/iso.ex
josevalim: Your original comment! It is still good as is, thank you!

josevalim:
Awesome job @eksperimental! I have dropped just one last comment, regarding some commented out code, and this should be good to go!

lib/elixir/lib/calendar/iso.ex
```diff
@@ -788,12 +832,17 @@ defmodule Calendar.ISO do
   @spec date_to_string(year, month, day, :basic | :extended) :: String.t()
   @impl true
   def date_to_string(year, month, day, format \\ :extended)
+      # when is_year(year) and is_month(month) and is_day(day) and format in [:basic, :extended] do
```
fertapric: I think @josevalim is talking about this commented out code.

eksperimental: thank you. comment removed

eksperimental:
Comment removed.

lib/elixir/lib/calendar/iso.ex
```diff
@@ -71,6 +91,23 @@ defmodule Calendar.ISO do
...
+  defguardp is_year_BCE(year) when year in -9999..0
+  defguardp is_year_CE(year) when year in 1..9999
+  defguardp is_month(month) when month in 1..12
+  defguardp is_day(day) when day in 1..31
+  defguardp is_hour(hour) when hour in 0..23
+  defguardp is_minute(minute) when minute in 0..59
+  defguardp is_second(second) when second in 0..59
+
+  defguardp is_microsecond(microsecond, precision)
+            when microsecond in 0..999_999 and precision in 0..6
```
josevalim: I really like your solution here of passing two arguments and leaving the destructuring up to the caller. :+1:

lib/elixir/lib/calendar/iso.ex
```diff
@@ -742,14 +777,23 @@ defmodule Calendar.ISO do
...
+        {ms_value, ms_precision} = microsecond,
+        format \\ :extended
+      )
+      when is_hour(hour) and is_minute(minute) and is_second(second) and
+             is_microsecond(ms_value, ms_precision) and format in [:basic, :extended] do
+    time_to_string_guarded(hour, minute, second, microsecond, format)
+  end
 
-  def time_to_string(hour, minute, second, {_, 0}, format) when format in [:basic, :extended] do
+  def time_to_string_guarded(hour, minute, second, {_, 0}, format) do
```
Cleidiano: Make private?

lib/elixir/lib/calendar/iso.ex
```diff
@@ -742,14 +777,23 @@ defmodule Calendar.ISO do
...
     time_to_string_format(hour, minute, second, format)
   end
 
-  def time_to_string(hour, minute, second, {microsecond, precision}, format)
-      when format in [:basic, :extended] do
+  def time_to_string_guarded(hour, minute, second, {microsecond, precision}, format) do
```
Cleidiano: Make private?

lib/elixir/lib/calendar/iso.ex
```diff
@@ -788,12 +832,16 @@ defmodule Calendar.ISO do
   @spec date_to_string(year, month, day, :basic | :extended) :: String.t()
   @impl true
   def date_to_string(year, month, day, format \\ :extended)
+      when is_integer(year) and is_integer(month) and is_integer(day) and
+             format in [:basic, :extended] do
+    date_to_string_guarded(year, month, day, format)
+  end
 
-  def date_to_string(year, month, day, :extended) do
+  def date_to_string_guarded(year, month, day, :extended) do
```
Cleidiano: Make private?

lib/elixir/lib/calendar/iso.ex
```diff
@@ -788,12 +832,16 @@ defmodule Calendar.ISO do
...
     zero_pad(year, 4) <> "-" <> zero_pad(month, 2) <> "-" <> zero_pad(day, 2)
   end
 
-  def date_to_string(year, month, day, :basic) do
+  def date_to_string_guarded(year, month, day, :basic) do
```
Cleidiano: Make private?

lib/elixir/lib/calendar/iso.ex
eksperimental: thank you @clediano. All 4 clauses have been made private

josevalim:

PR #9499, opened 2019-11-06
Minor corrections to recently introduced changes in Guards page

eksperimental:


lib/elixir/pages/Guards.md
```diff
@@ -82,7 +82,7 @@ not_nil_head?(["some_value", "another_value"])
 #=> false
 ```
 
-Even though the list `head` is not `nil`, the first clause for `not_nil_head?` fails, triggering the second clause which returns `false`. To make the guard behave correctly, you must ensure that the guard evaluates to `true`, like so:
```
eksperimental: the text here made me thing `head` was a list

josevalim:

PR #9502, opened 2019-11-07
Use type-check when it's an adjective

eksperimental:
As noun or verbs it is spelled "type check"

[ci skip]


josevalim:

PR #9507, opened 2019-11-08
Fix for default leftover in Enum.chunk/3

Eiji7:
Affects: `~> 1.7`, `~> 1.8` and `~> 1.9`
Fixes: #9506

The problem is caused by this commit: 11a0b39ad4a57236c9eb6bf17e5d46cc06621b8b

This means that it has been done when working on `v1.7.0-rc.0` for deprecating `Enum.chunk/4` in favor of `Enum.chunk_every/4`.

### DIFF
--- a/lib/elixir/test/elixir/enum_test.exs
+++ b/lib/elixir/test/elixir/enum_test.exs
@@ -49,15 +49,12 @@ defmodule EnumTest do
 
   test "chunk/3" do
     enum = Enum
-    enumerable = [1, 2, 3, 4, 5]
-    assert enum.chunk(enumerable, 2, 1) == Enum.chunk_every(enumerable, 2, 1, :discard)
+    assert enum.chunk(1..5, 2, 1) == Enum.chunk_every(1..5, 2, 1, :discard)
   end
 
   test "chunk/4" do
     enum = Enum
-    enumerable = [1, 2, 3, 4, 5]
-    assert enum.chunk(enumerable, 2, 1, :discard) == [[1, 2], [2, 3], [3, 4], [4, 5]]
-    assert enum.chunk(enumerable, 2, 1, nil) == Enum.chunk_every(enumerable, 2, 1, :discard)
+    assert enum.chunk(1..5, 2, 1, nil) == Enum.chunk_every(1..5, 2, 1, :discard)
   end
 
   test "chunk_every/2" do
--- a/lib/elixir/test/elixir/enum_test.exs
+++ b/lib/elixir/test/elixir/enum_test.exs
@@ -57,7 +57,7 @@ defmodule EnumTest do
     enum = Enum
     enumerable = [1, 2, 3, 4, 5]
     assert enum.chunk(enumerable, 2, 1, :discard) == [[1, 2], [2, 3], [3, 4], [4, 5]]
-    assert enum.chunk(enumerable, 2, 1, nil) == enum.chunk(enumerable, 2, 1, :discard)
+    assert enum.chunk(enumerable, 2, 1, nil) == Enum.chunk_every(enumerable, 2, 1, :discard)
   end
 
   test "chunk_every/2" do
--- a/lib/elixir/test/elixir/enum_test.exs
+++ b/lib/elixir/test/elixir/enum_test.exs
@@ -50,7 +50,7 @@ defmodule EnumTest do
   test "chunk/3" do
     enum = Enum
     enumerable = [1, 2, 3, 4, 5]
-    assert enum.chunk(enumerable, 2, 1) == enum.chunk(enumerable, 2, 1, :discard)
+    assert enum.chunk(enumerable, 2, 1) == Enum.chunk_every(enumerable, 2, 1, :discard)
   end
 
   test "chunk/4" do
--- a/lib/elixir/test/elixir/enum_test.exs
+++ b/lib/elixir/test/elixir/enum_test.exs
@@ -56,6 +56,15 @@ defmodule EnumTest do
     end)
   end
 
+  test "chunk/4" do
+    enumerable = [1, 2, 3, 4, 5]
+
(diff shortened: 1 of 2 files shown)


Eiji7:
Added comment to issue and updated PR description as well.

ping @josevalim and @whatyouhide

lib/elixir/test/elixir/enum_test.exs
```diff
@@ -47,6 +48,14 @@ defmodule EnumTest do
     assert Enum.at([2, 4, 6], -4) == nil
   end
 
+  test "chunk/3" do
+    enumerable = [1, 2, 3, 4, 5]
+
+    capture_io(fn ->
+      assert Enum.chunk(enumerable, 2, 1) == Enum.chunk(enumerable, 2, 1, :discard)
```
josevalim: So `Enum.chunk(enumerable, 2, 1, :discard)` was never officially supported. So I would match on the actual result and also add a test for `Enum.chunk(enumerable, 2, 1, nil)`.

lib/elixir/test/elixir/enum_test.exs
```diff
@@ -47,6 +48,14 @@ defmodule EnumTest do
     assert Enum.at([2, 4, 6], -4) == nil
   end
 
+  test "chunk/3" do
+    enumerable = [1, 2, 3, 4, 5]
+
+    capture_io(fn ->
```
josevalim: So the warning is emitted at compilation time, which means `capture_io` does not really do anything here? 🤔

josevalim: So you are not seeing the warning because of a bug in your deprecation checker but here is what you need to do:

```
enum = Enum
assert enum.chunk(enumerable, 2, 1)
```

and that should be warning free. No need for `capture_io`.

Eiji7: When you said that … However in my case I do not see warnings when using `capture_io/1` (commenting that call shows warning). Probably because it's not working strictly like `mix test`, but there is `elixir` binary called with `-r` and `-pr` arguments in `Makefile`, right?

lib/elixir/test/elixir/enum_test.exs
Eiji7: sure, no problem

lib/elixir/test/elixir/enum_test.exs
josevalim: I have fixed the bug in master, so you should see the deprecation warnings now. :)

lib/elixir/test/elixir/enum_test.exs
```diff
@@ -47,6 +48,23 @@ defmodule EnumTest do
...
+      assert Enum.chunk(enumerable, 2, 1) == Enum.chunk(enumerable, 2, 1, :discard)
+    end)
+  end
+
+  test "chunk/4" do
+    enumerable = [1, 2, 3, 4, 5]
+
+    capture_io(fn ->
+      assert Enum.chunk(enumerable, 2, 1, :discard) == [[1, 2], [2, 3], [3, 4], [4, 5]]
+      assert Enum.chunk(enumerable, 2, 1, nil) == Enum.chunk(enumerable, 2, 1, :discard)
```
josevalim:
```suggestion
      assert Enum.chunk(enumerable, 2, 1, nil) == Enum.chunk_every(enumerable, 2, 1, :discard)
```

lib/elixir/test/elixir/enum_test.exs
```diff
@@ -47,6 +47,19 @@ defmodule EnumTest do
     assert Enum.at([2, 4, 6], -4) == nil
   end
 
+  test "chunk/3" do
+    enum = Enum
+    enumerable = [1, 2, 3, 4, 5]
+    assert enum.chunk(enumerable, 2, 1) == enum.chunk(enumerable, 2, 1, :discard)
```
josevalim:
```suggestion
    assert enum.chunk(enumerable, 2, 1) == Enum.chunk_every(enumerable, 2, 1, :discard)
```

lib/elixir/test/elixir/enum_test.exs
```diff
@@ -47,6 +47,19 @@ defmodule EnumTest do
...
+  end
+
+  test "chunk/4" do
+    enum = Enum
+    enumerable = [1, 2, 3, 4, 5]
+    assert enum.chunk(enumerable, 2, 1, :discard) == [[1, 2], [2, 3], [3, 4], [4, 5]]
+    assert enum.chunk(enumerable, 2, 1, nil) == enum.chunk(enumerable, 2, 1, :discard)
```
josevalim:
```suggestion
    assert enum.chunk(enumerable, 2, 1, nil) == Enum.chunk_every(enumerable, 2, 1, :discard)
```

lib/elixir/test/elixir/enum_test.exs
```diff
@@ -47,6 +47,19 @@ defmodule EnumTest do
...
+  test "chunk/3" do
+    enum = Enum
+    enumerable = [1, 2, 3, 4, 5]
+    assert enum.chunk(enumerable, 2, 1) == enum.chunk(enumerable, 2, 1, :discard)
+  end
+
+  test "chunk/4" do
+    enum = Enum
+    enumerable = [1, 2, 3, 4, 5]
+    assert enum.chunk(enumerable, 2, 1, :discard) == [[1, 2], [2, 3], [3, 4], [4, 5]]
```
josevalim: We should remove this line because `:discard` was never a valid argument for `Enum.chunk`.  It is an implementation detail that is "leaking".

Eiji7: ok, makes sense

Eiji7:
@josevalim Is it ok now?

josevalim:

PR #9509, opened 2019-11-08
Configurable protocol consolidation timeout

Olshansk:
Allow protocol consolidation to be configurable
with a default timeout value of 30 seconds.


josevalim:
Hi @Olshansk. Instead of making it configurable, let's just make it `:infinity`. It is unlikely it will run into a race condition or deadlock forever, so we can just wait how long it takes. Can you please update accordingly?

Olshansk:
Done.

josevalim:

PR #9511, opened 2019-11-09
Fix types in Calendar

eksperimental:
The premise is that if `day` and `month` are 1-based values, `day_of_year`, `day_of_week` and `quarter_of_year` should as well.

### DIFF
--- a/lib/elixir/lib/calendar.ex
+++ b/lib/elixir/lib/calendar.ex
@@ -52,7 +52,7 @@ defmodule Calendar do
   representing the microseconds to external format. If the precision is 0,
   it means microseconds must be skipped.
   """
-  @type microsecond :: {0..999_999, 0..6}
+  @type microsecond :: {non_neg_integer, non_neg_integer}
 
   @typedoc "A calendar implementation"
   @type calendar :: module
--- a/lib/elixir/lib/calendar.ex
+++ b/lib/elixir/lib/calendar.ex
@@ -20,7 +20,7 @@ defmodule Calendar do
   @type month :: pos_integer
   @type day :: pos_integer
   @type week :: pos_integer
-  @type day_of_week :: non_neg_integer
+  @type day_of_week :: pos_integer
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/calendar.ex
```diff
@@ -16,17 +16,31 @@ defmodule Calendar do
   calendars may have a different number of days per month, months per year and so on.
   """
 
+  @type era :: non_neg_integer
```
eksperimental: this one was just moved here when sorting the types

lib/elixir/lib/calendar.ex
```diff
@@ -16,17 +16,31 @@ defmodule Calendar do
...
   @type second :: non_neg_integer
 
+  @typedoc """
+  Microseconds with stored precision.
+
+  The precision represents the number of digits that must be used when
+  representing the microseconds to external format. If the precision is 0,
+  it means microseconds must be skipped.
+  """
+  @type microsecond :: {non_neg_integer, non_neg_integer}
```
eksperimental: this one was just moved here when sorting the types

josevalim:
> The premise is that if day and month are 1-based values, day_of_year, day_of_week and quarter_of_year should as well.

I am not confident with making this choice without surveying the other possible calendar implementations. So unless that work is done, it is probably safer to keep things as is, thanks.

lib/elixir/lib/calendar.ex
eksperimental: Should we introduce this change ? As it was discussed earlier.

eksperimental:
> > The premise is that if day and month are 1-based values, day_of_year, day_of_week and quarter_of_year should as well.
> 
> I am not confident with making this choice without surveying the other possible calendar implementations. So unless that work is done, it is probably safer to keep things as is, thanks.

I will write something up and post it on Elixir Forum and the Elixir Core mailing list

PR #9517, opened 2019-11-12
Allow RELEASE_DISTRIBUTION to be set to none

josevalim:


lib/mix/lib/mix/tasks/release.ex
```diff
@@ -694,7 +695,8 @@ defmodule Mix.Tasks.Release do
       start/daemon/install commands
 
     * `RELEASE_DISTRIBUTION` - how do we want to run the distribution.
-      Using `name` (long names) or `sname` (short names). Defaults to
+      Maybe be `name` (long names), `sname` (short names) or `none`
```
fertapric: Maybe?

```suggestion
      May be `name` (long names), `sname` (short names) or `none`
```

josevalim:

PR #9523, opened 2019-11-12
Removed struct() from list of Basic Types in Typespecs.md

Eiji7:
fixes: #9522
ping: @ericmj 

Initially added in this commit: f1d4d5b034bd47d6488b808766f25a3f6fd82683

> Good find. It should only be listed in built-in types.


whatyouhide:
Thank you @Eiji7! 💟

eksperimental:
5 year-old documentation "bug".
Great catch! Thank you

PR #9527, opened 2019-11-13
Revamp capture_io for named devices to be async

### LINKED ISSUE #9508: Revamp capture_io with named process
Currently, `capture_io` does not allow concurrent captures on `:stderr` and other named devices. The goal is to allow `capture_io` on named devices to run concurrently, similar to `capture_log`, AS LONG AS no input for the capture was given (as we can't read from the capture concurrently).

Note that, if multiple captures are on the same named device, then the returned capture will include output for all captures happening at the same time. However, it is guaranteed that a single message won't be garbled. For example:

    IO.puts :stderr, "hello\nworld\n"

will always appear together in the capture output. While:

    IO.puts :stderr, "hello\n"
    IO.puts :stderr, "world\n"

May be garbled with messages from other processes. However, to reduce leaking content from between devices, every time a capture starts and the device is already being captured, we should capture the current output `byte_size` and return only the StringIO.contents from the given offset.

The mock device should be closed only when all captures are disabled. This means we will need to move the StringIO device management to the CaptureServer.

Once this work is done, there are three further improvements we can do:

  * [x] We can enable async: true on all tests that rely on capture_io(:stderr) today
  * [x] Strip kernel/warnings_test a bit (it should only hold warnings from Kernel or the compiler)
  * [ ] Add `@tag :capture_io`, `@tag capture_io: true`, `@tag capture_io: :stderr` (true means `:stdio`)

devonestes:
I hope @fertapric doesn't mind that I took a swing at this one. I just now saw that he was assigned...

With this change all StringIO stuff for named devices has been moved to the CaptureServer and we start a single StringIO for a given named device. All further captures on that device share the same StringIO, until all captures on that device have finished, at which time the StringIO is closed.

Async output from the device will be mixed in with other output, but we will offset from when a capture is first started to try and avoid too much unintended output being included in the captured output.

One thing that's a little tricky is that with the single StringIO, we only set the encoding once when the StringIO is first spawned. There's definitely some additional documentation or something that needs to come with this. Maybe if you need to set the encoding for named devices to anything other than `:unicode` then you can't do async capture_io?

Resolves #9508

### DIFF
--- a/lib/ex_unit/lib/ex_unit/capture_io.ex
+++ b/lib/ex_unit/lib/ex_unit/capture_io.ex
@@ -169,21 +169,20 @@ defmodule ExUnit.CaptureIO do
 
       {:error, {:changed_encoding, current_encoding}} ->
         raise ArgumentError, """
-        Attempted to change the encoding already set for the captured named device `#{
-          inspect(device)
-        }`
+        attempted to change the encoding for a currently captured device #{inspect(device)}.
+
         Currently set as: #{inspect(current_encoding)}
         Given: #{inspect(encoding)}
 
-        If you need to use multiple encodings on a captured named device, you cannot
-        run your test asynchronously.
+        If you need to use multiple encodings on a captured device, you cannot \
+        run your test asynchronously
         """
 
       {:error, :input_on_already_captured_device} ->
         raise ArgumentError,
-              "Attempted to give an input #{inspect(input)} for a currently captured named device `#{
-                inspect(device)
-              }`"
+              "attempted to give an input #{inspect(input)} for a currently captured device " <>
+                "#{inspect(device)}. If you need to give an input to a captured device, " <>
+                "you cannot run your test asynchronously"
     end
   end
 
--- a/lib/ex_unit/lib/ex_unit/capture_io.ex
+++ b/lib/ex_unit/lib/ex_unit/capture_io.ex
@@ -159,7 +159,7 @@ defmodule ExUnit.CaptureIO do
       {:ok, ref} ->
         try do
           fun.()
-          ExUnit.CaptureServer.device_output(ref)
+          ExUnit.CaptureServer.device_output(device, ref)
         after
           ExUnit.CaptureServer.device_capture_off(ref)
         end
--- a/lib/ex_unit/lib/ex_unit/capture_io.ex
+++ b/lib/ex_unit/lib/ex_unit/capture_io.ex
@@ -166,6 +166,24 @@ defmodule ExUnit.CaptureIO do
 
       {:error, :no_device} ->
         raise "could not find IO device registered at #{inspect(device)}"
+
+      {:error, {:changed_encoding, current_encoding}} ->
+        raise ArgumentError, """
+        Attempted to change the encoding already set for the captured named device `#{
+          inspect(device)
+        }`
+        Currently set as: #{inspect(current_encoding)}
+        Given: #{inspect(encoding)}
+
+        If you need to use multiple encodings on a captured named device, you cannot
+        run your test asynchronously.
+        """
+
+      {:error, :input_on_already_captured_device} ->
+        raise ArgumentError,
+              "Attempted to give an input #{inspect(input)} for a currently captured named device `#{
+                inspect(device)
+              }`"
     end
   end
 
--- a/lib/ex_unit/lib/ex_unit/capture_io.ex
+++ b/lib/ex_unit/lib/ex_unit/capture_io.ex
@@ -154,23 +154,18 @@ defmodule ExUnit.CaptureIO do
   defp do_capture_io(device, options, fun) do
     input = Keyword.get(options, :input, "")
     encoding = Keyword.get(options, :encoding, :unicode)
-    {:ok, string_io} = StringIO.open(input, encoding: encoding)
 
-    case ExUnit.CaptureServer.device_capture_on(device, string_io) do
+    case ExUnit.CaptureServer.device_capture_on(device, encoding, input) do
       {:ok, ref} ->
         try do
-          do_capture_io(string_io, fun)
+          fun.()
+          ExUnit.CaptureServer.device_output(ref)
         after
           ExUnit.CaptureServer.device_capture_off(ref)
         end
 
       {:error, :no_device} ->
-        _ = StringIO.close(string_io)
         raise "could not find IO device registered at #{inspect(device)}"
-
-      {:error, :already_captured} ->
-        _ = StringIO.close(string_io)
-        raise "IO device registered at #{inspect(device)} is already captured"
     end
   end
 
--- a/lib/ex_unit/lib/ex_unit/capture_io.ex
+++ b/lib/ex_unit/lib/ex_unit/capture_io.ex
@@ -35,10 +35,20 @@ defmodule ExUnit.CaptureIO do
   process and therefore can be done concurrently.
 
   However, the capturing of any other named device, such as `:stderr`,
-  happens globally and requires `async: false`.
+  happens globally and persists until the function has ended. While this means
+  it is safe to run your tests with `async: true` in many cases, captured output
+  may include output from a different test and care must be taken when using
+  `capture_io` with a named process asynchronously.
 
-  A developer can set a string as an input. The default input
-  is an empty string.
+  A developer can set a string as an input. The default input is an empty
+  string. If capturing a named device asynchronously, an input can only be given
+  to the first capture. Any further capture that is given to a capture on that
+  device will raise an exception and would indicate that the test should be run
+  synchronously.
+
+  Similarly, once a capture on a named device has begun, the encoding on that
+  device cannot be changed in a subsequent concurrent capture. An error will
+  be raised in this case.
 
   ## IO devices
 
--- a/lib/ex_unit/lib/ex_unit/capture_server.ex
+++ b/lib/ex_unit/lib/ex_unit/capture_server.ex
@@ -47,10 +47,10 @@ defmodule ExUnit.CaptureServer do
   end
 
   def handle_call({:device_output, name, ref}, _from, config) do
-    device = config.devices[name]
+    device = Map.fetch!(config.devices, name)
     {_, output} = StringIO.contents(device.pid)
     total = byte_size(output)
-    offset = device.refs[ref]
+    offset = Map.fetch!(device.refs, ref)
     output_size = total - offset
     {:reply, binary_part(output, offset, output_size), config}
   end
@@ -109,6 +109,7 @@ defmodule ExUnit.CaptureServer do
 
     try do
       Process.unregister(name)
+      Process.register(pid, name)
     rescue
       ArgumentError ->
         {:reply, {:error, :no_device}, config}
@@ -117,23 +118,16 @@ defmodule ExUnit.CaptureServer do
         ref = Process.monitor(caller)
         device = %{original_pid: original_pid, pid: pid, refs: %{ref => 0}, encoding: encoding}
         {:reply, {:ok, ref}, put_in(config.devices[name], device)}
-    after
-      try do
-        Process.register(pid, name)
-      rescue
-        ArgumentError ->
-          nil
-      end
     end
   end
 
   defp release_device(ref, %{devices: devices} = config) do
-    case device_by_ref(devices, ref) do
-      {name, device_info} ->
-        case Enum.reject(device_info.refs, &refs_equal?(&1, ref)) do
+    case Enum.find(devices, fn {_, device} -> device.refs[ref] end) do
+      {name, device} ->
+        case Enum.reject(device.refs, &(elem(&1, 0) == ref)) do
           [] ->
-            revert_device_to_original_pid(name, device_info.original_pid)
-            shut_down_string_io(device_info.pid)
+            revert_device_to_original_pid(name, device.original_pid)
+            close_string_io(device.pid)
             %{config | devices: Map.delete(devices, name)}
 
           refs ->
@@ -153,7 +147,7 @@ defmodule ExUnit.CaptureServer do
     Process.register(pid, name)
   end
 
-  defp shut_down_string_io(pid) do
+  defp close_string_io(pid) do
     StringIO.close(pid)
   rescue
     ArgumentError -> nil
@@ -175,18 +169,4 @@ defmodule ExUnit.CaptureServer do
       Logger.add_backend(:console, flush: true)
     end
   end
-
-  defp device_by_ref(devices, ref) do
-    Enum.find(devices, fn {_, device} -> offset_by_ref(device, ref) end)
-  end
-
-  defp offset_by_ref(device, ref) do
-    case Enum.find(device.refs, &refs_equal?(&1, ref)) do
-      {_, offset} -> offset
-      _ -> nil
-    end
-  end
-
-  defp refs_equal?({ref, _}, ref), do: true
-  defp refs_equal?(_, _), do: false
 end
--- a/lib/ex_unit/lib/ex_unit/capture_server.ex
+++ b/lib/ex_unit/lib/ex_unit/capture_server.ex
@@ -14,8 +14,8 @@ defmodule ExUnit.CaptureServer do
     GenServer.call(@name, {:device_capture_on, name, encoding, input}, @timeout)
   end
 
-  def device_output(ref) do
-    GenServer.call(@name, {:device_output, ref}, @timeout)
+  def device_output(name, ref) do
+    GenServer.call(@name, {:device_output, name, ref}, @timeout)
   end
 
   def device_capture_off(ref) do
@@ -46,16 +46,17 @@ defmodule ExUnit.CaptureServer do
     capture_device(name, encoding, input, config, caller)
   end
 
-  def handle_call({:device_output, ref}, _from, %{devices: devices} = config) do
-    {_, device} = device_by_ref(devices, ref)
+  def handle_call({:device_output, name, ref}, _from, config) do
+    device = config.devices[name]
     {_, output} = StringIO.contents(device.pid)
     total = byte_size(output)
-    offset = offset_by_ref(device, ref)
+    offset = device.refs[ref]
     output_size = total - offset
     {:reply, binary_part(output, offset, output_size), config}
   end
 
   def handle_call({:device_capture_off, ref}, _from, config) do
+    Process.demonitor(ref, [:flush])
     config = release_device(ref, config)
     {:reply, :ok, config}
   end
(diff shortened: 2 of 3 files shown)


fertapric:
> I hope @fertapric doesn't mind that I took a swing at this one. I just now saw that he was assigned...

No worries 🙂 

I think the safest way for the next time is to "call dibs" on the issue with a comment.

> One thing that's a little tricky is that with the single StringIO, we only set the encoding once when the StringIO is first spawned. There's definitely some additional documentation or something that needs to come with this. Maybe if you need to set the encoding for named devices to anything other than `:unicode` then you can't do async capture_io?

What I was doing in my implementation was to store the initial encoding in the CaptureServer's state and raising if the next capture was passing a different encoding.

Another things to take into account are:

* if the device is captured and you pass an `input`, I would raise because that would be reading from two different stdin buffers at the same time
* if the calling process dies, I would clean the capture reference

devonestes:
@fertapric Good call on all of those. I've implemented them and added tests for those cases.

lib/ex_unit/lib/ex_unit/capture_io.ex
```diff
@@ -154,23 +154,36 @@ defmodule ExUnit.CaptureIO do
...
       {:error, :no_device} ->
-        _ = StringIO.close(string_io)
         raise "could not find IO device registered at #{inspect(device)}"
 
-      {:error, :already_captured} ->
-        _ = StringIO.close(string_io)
-        raise "IO device registered at #{inspect(device)} is already captured"
+      {:error, {:changed_encoding, current_encoding}} ->
+        raise ArgumentError, """
+        Attempted to change the encoding already set for the captured named device `#{
```
josevalim: We don't use backticks to make code in error messages in Elixir.

josevalim: Also error messages start in lowercase. :)

lib/ex_unit/lib/ex_unit/capture_server.ex
```diff
@@ -83,25 +84,71 @@ defmodule ExUnit.CaptureServer do
...
-          try do
-            Process.unregister(name)
-          after
-            Process.register(pid, name)
-          end
-        rescue
-          ArgumentError -> nil
-        end
+  defp capture_device(name, encoding, "", %{devices: devices} = config, caller)
+       when :erlang.is_map_key(name, devices) do
```
josevalim:
```suggestion
       when is_map_key(devices, name) do
```

lib/ex_unit/lib/ex_unit/capture_server.ex
```diff
@@ -83,25 +84,71 @@ defmodule ExUnit.CaptureServer do
...
+        devices = %{devices | name => %{device | refs: refs}}
+        {:reply, {:ok, ref}, %{config | devices: devices}}
+
+      %{encoding: other_encoding} ->
+        {:reply, {:error, {:changed_encoding, other_encoding}}, config}
+    end
+  end
+
+  defp capture_device(name, _, _, %{devices: devices} = config, _)
+       when :erlang.is_map_key(name, devices) do
```
josevalim:
```suggestion
       when is_map_key(devices, name) do
```

lib/ex_unit/lib/ex_unit/capture_server.ex
```diff
@@ -83,25 +84,71 @@ defmodule ExUnit.CaptureServer do
...
 
-        %{config | devices: {names, refs}}
+    Process.unregister(name)
+    Process.register(pid, name)
 
-      {nil, _refs} ->
+    _ = Process.monitor(pid)
+    ref = Process.monitor(caller)
+
+    device = %{orig_pid: orig_pid, pid: pid, refs: [{ref, 0}], encoding: encoding}
```
josevalim: No abbreviations please. :)

lib/ex_unit/lib/ex_unit/capture_server.ex
```diff
@@ -83,25 +84,71 @@ defmodule ExUnit.CaptureServer do
...
+    devices = Map.put(devices, name, device)
```
josevalim: Please use `put_in`:

    config = put_in config.devices[name], device

lib/ex_unit/lib/ex_unit/capture_server.ex
josevalim: `refs: [{ref, 0}]` should be a map too, it will make it easier to read and remove refs.

lib/ex_unit/lib/ex_unit/capture_server.ex
```diff
@@ -83,25 +84,71 @@ defmodule ExUnit.CaptureServer do
...
-          ArgumentError -> nil
-        end
+  defp capture_device(name, encoding, "", %{devices: devices} = config, caller)
+       when :erlang.is_map_key(name, devices) do
+    case Map.get(devices, name) do
+      %{encoding: ^encoding} = device ->
+        {_, output} = StringIO.contents(device.pid)
+        ref = Process.monitor(caller)
+        refs = [{ref, byte_size(output)} | device.refs]
+        devices = %{devices | name => %{device | refs: refs}}
```
josevalim: This can be written as:

    update_in config.devices[name].refs, &[{ref, byte_size(output)} | &1]

And if you make the refs a map, as proposed below, it can be written as:

    put_in config.devices[name].refs[ref], byte_size(output)

lib/ex_unit/lib/ex_unit/capture_server.ex
```diff
@@ -83,25 +84,71 @@ defmodule ExUnit.CaptureServer do
...
+                Process.unregister(name)
+              after
+                Process.register(device_info.orig_pid, name)
+                StringIO.close(device_info.pid)
+                Process.demonitor(ref)
+              end
+            rescue
+              ArgumentError ->
+                nil
+            end
```
josevalim: It is probably worth moving this to its own function.

lib/ex_unit/test/ex_unit/capture_io_test.exs
```diff
@@ -86,6 +86,105 @@ defmodule ExUnit.CaptureIOTest do
...
+        capture_io(:stderr, fn ->
+          send(parent, :ready)
+          Process.sleep(:infinity)
+        end)
+      end)
+
+    assert_receive :ready
+
+    # Kill the process and make sure the capture server receives the down	
+    :erlang.trace(Process.whereis(ExUnit.CaptureServer), true, [:receive, tracer: self()])
```
josevalim: I would like to not use trace. That's because tracing doesn't actually check that the device was released. What you can do is to start the cacpture above with input: "foo", then kill it, and then start a capture with input "bar". Because the capture should have been released, then the input handlinig should work.

lib/ex_unit/lib/ex_unit/capture_server.ex
```diff
@@ -30,28 +34,25 @@ defmodule ExUnit.CaptureServer do
...
-      orig_pid = Process.whereis(name)
-      Process.unregister(name)
-      Process.register(pid, name)
-      ref = Process.monitor(pid)
-      refs = Map.put(refs, ref, {name, orig_pid})
-      names = Map.put(names, name, true)
-      {:reply, {:ok, ref}, %{config | devices: {names, refs}}}
-    end
+  def handle_call({:device_output, ref}, _from, %{devices: devices} = config) do
+    {_, device} = device_by_ref(devices, ref)
```
josevalim: Please pass the name as argument and if you convert `refs` to a map, than this can be:

    byte_offset = config.devices[name].refs[ref]

Which should be much simpler than traversing looking for refs.

josevalim: Alternatively, you can keep another map, of refs to device name. This way you can use the reverse map lookup here, and also on `DOWN` and `device_capture_off`.

lib/ex_unit/lib/ex_unit/capture_server.ex
```diff
@@ -83,25 +84,71 @@ defmodule ExUnit.CaptureServer do
...
+      {name, device_info} ->
+        case Enum.reject(device_info.refs, &refs_equal?(&1, ref)) do
+          [] ->
+            try do
+              try do
+                Process.unregister(name)
+              after
+                Process.register(device_info.orig_pid, name)
+                StringIO.close(device_info.pid)
+                Process.demonitor(ref)
```
josevalim: It is best to call this function only on `device_capture_off`, since in the case of the DOWN message it will already be demonitored. You also want to pass the flush argument:

So the :DOWN message is discarded.

lib/ex_unit/lib/ex_unit/capture_server.ex
```diff
@@ -83,25 +84,71 @@ defmodule ExUnit.CaptureServer do
...
+  defp capture_device(name, encoding, input, %{devices: devices} = config, caller) do
+    {:ok, pid} = StringIO.open(input, encoding: encoding)
+    orig_pid = Process.whereis(name)
 
-        %{config | devices: {names, refs}}
+    Process.unregister(name)
+    Process.register(pid, name)
 
-      {nil, _refs} ->
+    _ = Process.monitor(pid)
```
josevalim: Is this monitor or its DOWN message ever used?

lib/ex_unit/lib/ex_unit/capture_server.ex
```diff
@@ -83,25 +84,71 @@ defmodule ExUnit.CaptureServer do
...
+    case device_by_ref(devices, ref) do
+      {name, device_info} ->
+        case Enum.reject(device_info.refs, &refs_equal?(&1, ref)) do
+          [] ->
+            try do
+              try do
+                Process.unregister(name)
+              after
+                Process.register(device_info.orig_pid, name)
+                StringIO.close(device_info.pid)
```
josevalim: We should probably close the device in a separate try that happens after this one. No need to couple them. Then we can move each try/catch to its own individual function.

josevalim:
Thanks for the PR, I have added some comments!

lib/ex_unit/lib/ex_unit/capture_server.ex
```diff
@@ -83,29 +85,67 @@ defmodule ExUnit.CaptureServer do
...
-            Process.unregister(name)
-          after
-            Process.register(pid, name)
-          end
-        rescue
-          ArgumentError -> nil
-        end
+  defp capture_device(name, encoding, "", %{devices: devices} = config, caller)
+       when is_map_key(devices, name) do
+    case Map.get(devices, name) do
```
josevalim:
```suggestion
    case Map.fetch!(devices, name) do
```

lib/ex_unit/lib/ex_unit/capture_server.ex
devonestes: No, good call, this will end up essentially doing nothing.

lib/ex_unit/lib/ex_unit/capture_server.ex
```diff
@@ -83,29 +85,67 @@ defmodule ExUnit.CaptureServer do
...
+    Process.unregister(name)
+    Process.register(pid, name)
 
-      {nil, _refs} ->
+    _ = Process.monitor(pid)
+    ref = Process.monitor(caller)
+
+    device = %{original_pid: original_pid, pid: pid, refs: %{ref => 0}, encoding: encoding}
+    {:reply, {:ok, ref}, put_in(config.devices[name], device)}
+  rescue
```
josevalim: It is probably worth keeping this try/rescue only around the unregister/register calls, to make sure we are not rescuing something we should not. You can use the `else` clause in `try` to handle the success cases.

lib/ex_unit/lib/ex_unit/capture_server.ex
```diff
@@ -122,4 +169,18 @@ defmodule ExUnit.CaptureServer do
       Logger.add_backend(:console, flush: true)
     end
   end
+
+  defp device_by_ref(devices, ref) do
+    Enum.find(devices, fn {_, device} -> offset_by_ref(device, ref) end)
```
josevalim:
```suggestion
    Enum.find(devices, fn {_, device} -> Map.has_key?(devices, ref) end)
```

offset_by_ref and refs_equals can be fully removed now. :)

lib/ex_unit/test/ex_unit/capture_io_test.exs
```diff
@@ -86,6 +86,102 @@ defmodule ExUnit.CaptureIOTest do
...
+          capture_io(:stderr, fn ->
+            :io.put_chars(:standard_error, "#{num}\n")
+            Process.sleep(10)
+            :io.put_chars(:standard_error, "#{num + 10}\n")
+          end)
+
+        send(me, captured)
+      end)
+
+      Process.sleep(1)
```
josevalim: Instead of sleeps, please use explicit messaging. Something like:

```elixir
pids =
  for num <- 1..6 do
    spawn(fn ->
      captured =
        capture_io(:stderr, fn ->
          :io.put_chars(:standard_error, "#{num}\n")
          send(me, :waiting)
          assert_receive :continue
          :io.put_chars(:standard_error, "#{num + 10}\n")
        end)

      send(me, {:captured, captured})
    end)

    assert_receive :waiting
  end

Enum.each(pid, &send(&1, :continue))
```

Similar in the other tests.

lib/ex_unit/test/ex_unit/capture_io_test.exs
```diff
@@ -86,6 +86,102 @@ defmodule ExUnit.CaptureIOTest do
...
+    end)
+
+    expected = [
+      {~r/^1\n/, ~r/11\n/},
+      {~r/^2\n/, ~r/12\n/},
+      {~r/^3\n/, ~r/13\n/},
+      {~r/^4\n/, ~r/14\n/},
+      {~r/^5\n/, ~r/15\n/},
+      {~r/^6\n/, ~r/16\n/}
+    ]
```
josevalim: Let's avoid regexes if possible. We could for example log them as "before:#{num}" and "after:#{num}" and use `String.contains?` (or continue using `=~` which also works for binaries).

lib/ex_unit/test/ex_unit/capture_io_test.exs
```diff
@@ -86,6 +86,102 @@ defmodule ExUnit.CaptureIOTest do
...
+        rescue
+          e in [ArgumentError] ->
+            send(me, e.message)
+        end
+      end)
+    end)
+
+    assert_receive "attempted to give an input \"second\" for a currently captured device :standard_error. If you need to give an input to a captured device, you cannot run your test asynchronously"
+
+    refute_receive "attempted to give an input \"\" for a currently captured device :standard_error" <>
```
josevalim: Negative assertions are expensive (they always wait) and brittle (if someone change the error message and forget to update it, then it will always be wrong. Instead, do a `send(me, {:ok, input})` after `capture_io` and `assert_receive` on that instead.

lib/ex_unit/test/ex_unit/capture_io_test.exs
```diff
@@ -86,6 +86,102 @@ defmodule ExUnit.CaptureIOTest do
...
+                     _
+  end
+
+  test "CaptureServer monitors calling processes and releases the capture on exit" do
```
josevalim:
```suggestion
  test "monitors calling processes and releases the capture on exit" do
```

lib/ex_unit/lib/ex_unit/capture_server.ex
fertapric: I think it should be:

```suggestion
    Enum.find(devices, fn {_, device} -> Map.has_key?(device.refs, ref) end)
```

lib/ex_unit/lib/ex_unit/capture_server.ex
```diff
@@ -83,29 +85,74 @@ defmodule ExUnit.CaptureServer do
...
 
+  defp revert_device_to_original_pid(name, pid) do
+    Process.unregister(name)
+  rescue
+    ArgumentError -> nil
+  after
+    Process.register(pid, name)
+  end
+
+  defp shut_down_string_io(pid) do
```
fertapric: Nitpick:

```suggestion
  defp close_string_io(pid) do
```

lib/ex_unit/lib/ex_unit/capture_server.ex
```diff
@@ -30,31 +34,29 @@ defmodule ExUnit.CaptureServer do
...
-      orig_pid = Process.whereis(name)
-      Process.unregister(name)
-      Process.register(pid, name)
-      ref = Process.monitor(pid)
-      refs = Map.put(refs, ref, {name, orig_pid})
-      names = Map.put(names, name, true)
-      {:reply, {:ok, ref}, %{config | devices: {names, refs}}}
-    end
+  def handle_call({:device_output, name, ref}, _from, config) do
+    device = config.devices[name]
```
fertapric: For better error messages:

```suggestion
    device = Map.fetch!(config.devices, name)
```

Btw, is it ok to assume the device info will be there? I guess it might be since this is an internal module 🤔

lib/ex_unit/lib/ex_unit/capture_server.ex
```diff
@@ -30,31 +34,29 @@ defmodule ExUnit.CaptureServer do
...
+    {_, output} = StringIO.contents(device.pid)
+    total = byte_size(output)
+    offset = device.refs[ref]
```
fertapric: Same as above, better error message (and I think better performance) in case something goes wrong:

```suggestion
    offset = Map.fetch!(device.refs, ref)
```

lib/ex_unit/lib/ex_unit/capture_server.ex
josevalim: It is ok to assume. If it isn't, then it is a bug, and we want it to crash anyway.

fertapric:
The documentation needs to be updated too (from other PRs I think you plan to do that once all the coding is done, but "just in case")  🙂

lib/ex_unit/lib/ex_unit/capture_server.ex
josevalim: I am honestly not super worried about this. It should never fail. It should always be there.

To be clear, any choice is fine, but I don't think the argument of better error message or better performance matter here, because it should never error.

devonestes:
Yeah, I'm hoping to get all the functionality and edge cases and stuff sorted before documenting all of that, so once everyone is cool with the behavior then I'll tack on a documentation commit as well. I think we're probably there at this point, though...

lib/ex_unit/test/ex_unit/capture_io_test.exs
```diff
@@ -1,7 +1,7 @@
 Code.require_file("../test_helper.exs", __DIR__)
 
 defmodule ExUnit.CaptureIOTest do
-  use ExUnit.Case
+  use ExUnit.Case, async: true
```
josevalim: I would not make this test in particular async because we want to be more precise on this one:

```suggestion
  use ExUnit.Case
```

lib/ex_unit/test/ex_unit/capture_io_test.exs
```diff
@@ -86,6 +86,124 @@ defmodule ExUnit.CaptureIOTest do
...
+
+    ref = Process.monitor(pid)
+
+    # Kill the process and make sure the caputre is released
+    Process.exit(pid, :shutdown)
+
+    # Make sure the process has exited before we try and start a new capture
+    assert_receive {:DOWN, ^ref, _, _, _}
+
+    assert capture_io(:stderr, [input: "b"], fn -> :ok end)
```
josevalim: Beautiful!

lib/ex_unit/test/ex_unit/capture_io_test.exs
```diff
@@ -86,6 +86,124 @@ defmodule ExUnit.CaptureIOTest do
...
+    |> Enum.reverse()
+    |> Enum.each(&send(&1, :continue))
+
+    assert_received_all(Enum.to_list(1..6))
+  end
+
+  test "raises when async capturing a named device with a different encoding than the first" do
+    me = self()
+
+    pids =
```
josevalim: Nitpick. Do `[pid0, _pid1, pid2] =` then you can remove the Enum.at below.

josevalim:
Fantastic, thank you for the PR! :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

PR #9528, opened 2019-11-14
Documentation updates in Bitwise module and functions

eksperimental:
Related 1e4e05ef78b3105065f0a313bd0e1e78b2aa973e

/cc @ericmj


fertapric:
I think we can add `@spec` in some of them :)

josevalim:
Good call on `@spec`, let's do it in a separate commit.

eksperimental:
> I think we can add `@spec` in some of them :)

I will work on it

Issue #6476, opened 2017-08-17
Add option to fail tests when test scripts contain warnings


scouten:
I would like to see ExUnit have an option to fail (exit with non-zero status and appropriate warnings) whenever any `.exs` file has a compilation warning.

The solution suggested in https://github.com/elixir-lang/elixir/issues/3223#issuecomment-87724948 does not satisfy this request because:

* It _will_ fail when compiled files (*.ex) contain compilation warnings, but …
* It _will **not**_ fail when the test scripts (*.exs) contain warnings.

I would like to be able to fail a build due to warnings in the .exs files and I see no way to do so today.

josevalim:
I believe you can set warnings_as_errors via `Code.compiler_options` at the top of your mix.exs file. If you would prefer an option to `mix test`, please do send a pull request. Thank you!

scouten:
I couldn't find an example of using `Code.compiler_options` but the technique described in https://stackoverflow.com/questions/31982577/is-it-possible-to-get-the-compiler-to-exit-early-failing-the-build-if-a-compil worked for me.

deadtrickster:
Hello

I have unused variable in my test/file.exs and I have warnings_as_errors in my mix.exs.
 When I run `mix test` first time (no _build yet) warnings printed and compile fails with
`Execution failed due to warnings while using the --warnings-as-errors option`. Just as expected.
If I run `mix test` second time however, warnings as printed but tests compile doesn't fail and tests continue.

Elixir 1.4.4 OTP 20

Thanks

harlantwood:
> I believe you can set warnings_as_errors via Code.compiler_options at the top of your mix.exs file. 

To make this explicit, add to your mix project:

    elixirc_options: [warnings_as_errors: true]

PR #9532, opened 2019-11-14
Add `--warnings-as-errors` flag to `mix test`

matiasgarciaisaia:
This flag, copied over from `mix compile`, will make the compilation fail if the app itself or the tests throw any warnings.

The scaffolding is ready to allow setting other compiler parameters from `mix test` flags.

See #6476

### DIFF
--- a/lib/mix/lib/mix/tasks/test.ex
+++ b/lib/mix/lib/mix/tasks/test.ex
@@ -196,6 +196,8 @@ defmodule Mix.Tasks.Test do
     * `--timeout` - sets the timeout for the tests
     * `--trace` - runs tests with detailed reporting. Automatically sets `--max-cases` to `1`.
       Note that in trace mode test timeouts will be ignored as timeout is set to `:infinity`
+    * `--warnings-as-errors` - treats warnings in the current project and tests as errors and
+      return a non-zero exit code
 
   ## Configuration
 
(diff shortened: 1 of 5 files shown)


josevalim:
Hi @matiasgarciaisaia,

The reason we don't support this flag in `mix test` is because folks may expect the warnings during tests to also lead to errors, so it can be misleading.

The best option in your case is to enable warnings_as_errors in your mix.exs or do an explicit compilation step with the flag.

Thank you.

matiasgarciaisaia:
Hi @josevalim, thanks for taking a look.

I think I _kind of_ understand, but maybe you can help me clarify - which kind of warnings are the ones you say folks will expect to lead to errors, but wouldn't?

My need surfaced because we had a test file called `something_tests.exs`. We got a warning saying the file wasn't being run, but the test command (`MIX_ENV=test app mix do compile --warnings-as-errors, test`) didn't fail. I want _that_ `test` to be something like `test --warnings-as-errors` and fail, too.

josevalim:
>  I want that test to be something like test --warnings-as-errors and fail, too.

Correct. If you want this, the current patch is not enough (afaik).

matiasgarciaisaia:
I'll try adding that to the patch, although I'm not sure I'll be able to make it.

If it's done, that's something that you'd want to merge into `mix`, right?

josevalim:
Yup!

PR #9534, opened 2019-11-14
Configure Cirrus CI

fertapric:
Elixir is migrating from Travis to Cirrus CI to have a single place to test against Unix, FreeBSD, and Windows.


josevalim:
Should we remove .travis and .appveyor in this same PR or a separate one?

fertapric:
@josevalim I was going to do that in a separate PR because I thought you might want to compare CIs (build times, queue times, etc)

PR #9538, opened 2019-11-16
Improve Bitwise docs

eksperimental:


lib/elixir/lib/bitwise.ex
```diff
@@ -28,9 +31,8 @@ defmodule Bitwise do
 
   All bitwise functions can be used in guards:
 
-      iex> use Bitwise
       iex> odd? = fn
-      ...>   int when band(int, 1) == 1 -> true
+      ...>   int when Bitwise.band(int, 1) == 1 -> true
```
eksperimental: We no longer need to require/import this  module

josevalim:

PR #9540, opened 2019-11-16
Add List.ends_with?/2

eksperimental:
We have `String.starts_with?/2`, `String.ends_with?/2`, and `List.starts_with?/2`, but we are missing `List.ends_with?/2`.

I'm bypassing sending this proposal to the mailing list because I think it's kind of unnecessary but I can do if needed.

### DIFF
--- a/lib/elixir/lib/list.ex
+++ b/lib/elixir/lib/list.ex
@@ -741,6 +741,41 @@ defmodule List do
   end
 
   @doc """
+  Returns `true` if `list` ends with the given `tail` list; otherwise returns `false`.
+
+  If `tail` is an empty list, it returns `true`.
+
+  ### Examples
+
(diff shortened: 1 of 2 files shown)


josevalim:
Please do send to the mailing list. Opposite to the string operations,
ends_with is quite expensive, which is why I would avoid adding it. If you
have a use case, great, otherwise I would avoid adding it for
“completeness”.
-- 


*José Valim*
www.plataformatec.com.br
Founder and Director of R&D

eksperimental:
I was working with charlists and I had to check if it ended with certain chars.
Quick fix was to use `charlist |> to_string() |> String.ends_with?/2`

josevalim:
If working with strings is an option, then that’s preferable. :)

PR #9541, opened 2019-11-16
Introduce t:Calendar.ISO.microsecond/0 and fix t:Calendar.microsecond/0

eksperimental:
The reasoning is that the `0..999_999` and `0..6` values only applied
to the Calendar.ISO calendar, and that could be restrictive for other
types of calendars. As discussed in
https://github.com/elixir-lang/elixir/pull/9493#pullrequestreview-312050446


josevalim:

PR #9542, opened 2019-11-16
Add note about package managers to release docs

oskarkook:
Elixir releases behave slightly differently in the package manager Guix, as erlang & elixir will contain references to the Guix store. These will then get copied to the release and the release won't be able to run on the target without having the needed Guix packages installed.

I think this behaviour is fine, considering there already are constraints regarding the matching of host and target, but it would be great to also have this case documented in Elixir.

As a side-note, the user could run a script that replaces the references to the Guix store in the release, but documenting this here would probably just be confusing.

### DIFF
--- a/lib/mix/lib/mix/tasks/release.ex
+++ b/lib/mix/lib/mix/tasks/release.ex
@@ -271,6 +271,17 @@ defmodule Mix.Tasks.Release do
   but if necessary, you can also bundle the compiled object files in the release,
   as long as they were compiled for the same target. If doing so, you need to
   update `LD_LIBRARY_PATH` with the paths containing the bundled objects.
+  
+  Often the package manager itself has to match between the host and the target.
+  There can be differences in the build process of Elixir between package
+  managers and you can end up with cases where Elixir built with a certain
+  package manager depends on a certain file system layout. If you then use it
+  to create a release, the release will effectively have the same dependency.
+  This is especially apparent with package managers that avoid linking to global
+  executables (such as `/bin/sh`), instead tying Elixir to specific package
+  versions, installed at deterministic unique paths. In such cases, it is
+  best to use the same package manager on the target and install the same
+  Elixir version as on the host.
 
   Currently, there is no official way to cross-compile a release from one
   target triple to another, due to the complexities involved in the process.
--- a/lib/mix/lib/mix/tasks/release.ex
+++ b/lib/mix/lib/mix/tasks/release.ex
@@ -267,21 +267,16 @@ defmodule Mix.Tasks.Release do
   containing NIFs (natively-implemented functions) which may expect to
   dynamically link to libraries they use.
 
-  These system packages are typically managed using the system package manager,
-  but if necessary, you can also bundle the compiled object files in the release,
+  Of course, some OSes and package managers can differ between versions,
+  so if your goal is to have full compatibility between host and target,
+  it is best to ensure the OS and system package manager have the same
+  versions on host and target. This may even be a requirement in some
+  systems, especially so with package managers that try to create fully
+  reproducible environments (Nix, Guix).
+
+  Alternatively, you can also bundle the compiled object files in the release,
   as long as they were compiled for the same target. If doing so, you need to
   update `LD_LIBRARY_PATH` with the paths containing the bundled objects.
-  
-  Often the package manager itself has to match between the host and the target.
-  There can be differences in the build process of Elixir between package
-  managers and you can end up with cases where Elixir built with a certain
-  package manager depends on a certain file system layout. If you then use it
-  to create a release, the release will effectively have the same dependency.
-  This is especially apparent with package managers that avoid linking to global
-  executables (such as `/bin/sh`), instead tying Elixir to specific package
-  versions, installed at deterministic unique paths. In such cases, it is
-  best to use the same package manager on the target and install the same
-  Elixir version as on the host.
 
   Currently, there is no official way to cross-compile a release from one
   target triple to another, due to the complexities involved in the process.


lib/mix/lib/mix/tasks/release.ex
```diff
@@ -271,6 +271,17 @@ defmodule Mix.Tasks.Release do
...
+  Often the package manager itself has to match between the host and the target.
+  There can be differences in the build process of Elixir between package
+  managers and you can end up with cases where Elixir built with a certain
+  package manager depends on a certain file system layout. If you then use it
+  to create a release, the release will effectively have the same dependency.
+  This is especially apparent with package managers that avoid linking to global
+  executables (such as `/bin/sh`), instead tying Elixir to specific package
+  versions, installed at deterministic unique paths. In such cases, it is
+  best to use the same package manager on the target and install the same
+  Elixir version as on the host.
```
josevalim: Thanks! Can you please describe a bit more precisely what Guix is doing here? Is it changing Elixir executables themselves? If so... maybe we should make a copy of said executables at build time.

oskarkook: Guix replaces paths like `/bin/sh` to `/gnu/store/29jhbbg1hf557x8j53f9sxd9imlmf02a-bash-minimal-5.0.7/bin/sh` during build time. This is good because you can get determenistic, reproducible builds; however when a release is cut, it means these paths will also end up in the release. If the target is not using Guix, it means it won't have that `sh` executable available.

Copying those dependencies (currently I believe only `sh` and `git` might be needed) could definitely be the most elegant solution. I wonder if security is a question here.. if erlang and elixir executables are already copied from the host, it seems like the target is already exposed to certain security issues, so copying the other dependencies shouldn't raise the bar there.

To upgrade the Guix package to 1.9, I had to add the following changes to the build:
```scheme
               (substitute* '("lib/mix/lib/mix/release.ex"
                              "lib/mix/lib/mix/tasks/release.init.ex")
                 (("#!/bin/sh")
                  (string-append "#!" (which "sh"))))
               (substitute* "bin/elixir"
                 (("ERTS_BIN=")
                  (string-append
                    "ERTS_BIN="
                    ;; Elixir Releases will prepend to ERTS_BIN the path of a copy of erl.
                    ;; We detect if a release is being generated by checking the initial ERTS_BIN
                    ;; value: if it's empty, we are not in release mode and can point to the actual
                    ;; erl binary in Guix store.
                    "\nif [ -z \"$ERTS_BIN\" ]; then ERTS_BIN="
                    (string-drop-right (which "erl") 3)
                    "; fi")))
```

It replaces the `#!/bin/sh` references in `lib/mix/lib/mix/release.ex` and `lib/mix/lib/mix/tasks/release.init.ex` with references to Guix's `sh`. It also adds a check into `bin/elixir` that will reference the Guix's `erl` executable if Elixir is not running in a release (this is detected by checking the `ERTS_BIN` variable, since Elixir will append to that variable when a release is cut).

josevalim: Thanks @oskarkook. By copying, I meant having a copy of the bin/* files when Elixir itself is compiled. Then if the package manager wants to fiddle with them, at least it won't affect releases. WDYT?

oskarkook: I am not sure I follow.

Let's take `/bin/sh` under the magnifying glass for the moment. Files under bin/* in Elixir source reference it. In the Guix build environment, there is no /bin/ (at all), instead when Guix goes to build Elixir, before it starts building, it will go and modify the source to reference the available `sh` executable in the Guix store, instead of `/bin/sh`. After this step, the source is now effectively tied to the Guix store and this is fine from Guix's point of view (if a new version of the shell were released, Guix would re-build Elixir anyway). Guix can then build Elixir as usual.

From Elixir's point of view, would creating a copy of bin/* do anything if the source has been modified beforehand?

Personally, I see two potential solutions (other than just documenting the case):
1. You copy the shell itself to the release and use it to run it.
2. You allow the user to specify where the shell is when creating a release. Elixir would then modify the bin/* files accordingly before copying them.

The second option doesn't really feel like a true fix - just documenting the situation seems more appropriate and let the user deal with it. The first one actually could fix the issue, however I'm not sure what the scope of releases is - should it only try to replicate the Elixir/Erlang environment (which it currently does by copying the needed executables etc) or should it be more ambitious? As soon as you start copying the shell etc, you kind of leave the Elixir/Erlang space.

oskarkook: Of course, those source modification steps could be added after some Elixir build step that does the copying (which would itself happen before compilation).

Effectively it's the same solution as the second option I mentioned, however you would not have control over it at release time. It would also break the tests in Guix since as said, `/bin/sh` is not available in the Guix build environment. Elixir has tests that run the releases and if those tests are producing artifacts that reference `/bin/sh`, you'd need even more steps in the Guix build process to "undo" this behaviour (or just disable the tests altogether).

josevalim: Oh, I see. So even if we preserved `/bin/sh`, by the time you assemble and run the release, that wouldn't work either, because there is no `/bin/sh`.

oskarkook: Also note that the `erl` executable will also be referencing the Guix store. I can go and grep a release to see which other files will have dependencies on it, but my point is that it won't be just Elixir.

It's an annoying thing to solve in any case. My advice is to not over-think it - if the package manager forces such an environment, the user should make sure to replicate that environment on the target. Document the case and point to it if someone is having issues.

josevalim: Agreed. We already say the package manager has to install system dependencies. And `bin/sh` is just a system dependency as well. So wouldn't it be enough for us to say it is important that the OS and package manager versions themselves match generally? Something like:

> Of course, some OSes and package managers can differ between versions,
> so if your goal is to have full compatibility between host and target, it is best
> to ensure the OS and system package manager have the same versions on
> host and target. This may even be a requirement in some systems.

Thoughts?

oskarkook: Yes, that is fine and more concise than my description. I'd change the last sentence to emphasize that it's more prominent in systems that try to achieve full reproducibility:

> This may even be a requirement in some systems, especially so with package managers that try to create fully reproducible environments (Nix, Guix).

Also, something to consider is to be able to create a release and enjoy the benefits of preloading, precompilation etc, but use the system's Elixir. With a fully reproducible environment (which Guix can provide), there really shouldn't be a need to package Elixir & Erlang with the release anyway.

Presumably this is already possible (not too familiar with the internals), but not documented?

josevalim:

lib/mix/lib/mix/tasks/release.ex
josevalim: > Presumably this is already possible (not too familiar with the internals), but not documented?

Yup, you can set `include_erts: false`. Elixir has no native deps, so it is always bundled.

PR #9543, opened 2019-11-17
Return {:error, :unavalable} for ensure_compiled in a deadlock

josevalim:
Some projects like Ecto can use this information to avoid
doing certain verifications, as they may lead to false
positives.


josevalim:

PR #9544, opened 2019-11-17
Make the number of attempted function clauses in FunctionClauseError more obvious

### LINKED ISSUE #9529: "Attempted function clauses" message only shows max 10 attempted clauses
Ran into this situation which had me pretty confused for a while.

I had this incorrect code:

    Timex.set(microsecond: 0)

Which caused this error:

    ** (FunctionClauseError) no function clause matching in Timex.normalize/2

     The following arguments were given to Timex.normalize/2:

         # 1
         :microsecond

         # 2
         0

     Attempted function clauses (showing 10 out of 12):

         def normalize(:date, {year, month, day})
         def normalize(:year, year) when year < 0
         def normalize(:year, year)
         def normalize(:month, month)
         def normalize(:time, {hour, min, sec})
         def normalize(:time, {hour, min, sec, ms})
         def normalize(:hour, hour)
         def normalize(:minute, min)
         def normalize(:second, sec)
         def normalize(:millisecond, ms)

Showing 10 clauses usually would be enough, however the clause I was trying to match was the **11th** clause of 12.

Not quite realising the behaviour of this message (only showing max 10 clauses) I assumed all clauses were shown, and headed off down a less than fruitful avenue of debugging.

It was only when I happened to notice the `(showing 10 out of 12)` that I realised that the missing microsecond clause must be missing, so found the correct match in the Timex source code.

    def normalize(:microsecond, {us, p})

So my code needed to be:

    Timex.set(microsecond: {0, 0} )

Should this be showing all clauses? The error message is already pretty long and not showing all clauses (and not having any obvious way to show all if needed) is likely to be confusing.

On the downside there is the possibility of there being many clauses listed. Might it be possible to  try and figure out which clauses are closest?

Options:
1. Show all clauses always
2. Ability to configure showing all clauses
3. Show limited number, but try and figure out which clause might be closest match

joshprice:
Ended up with @fertapric's suggestion but also appended the `...` as hinted at by @josevalim which ended up making the non visible clause section more obvious.

```
The following arguments were given to Macro.to_string/2:

    # 1
    :invalid

    # 2
    :invalid

Attempted function clauses (showing 10 out of 24):

    def to_string(-{var, _, context} = ast-, fun) when -is_atom(var)- and -is_atom(context)-
    def to_string(-{:__aliases__, _, refs} = ast-, fun)
    def to_string(-{:__block__, _, [expr]} = ast-, fun)
    def to_string(-{:__block__, _, _} = ast-, fun)
    def to_string(-{:<<>>, _, parts} = ast-, fun)
    def to_string(-{:{}, _, args} = ast-, fun)
    def to_string(-{:%{}, _, args} = ast-, fun)
    def to_string(-{:%, _, [struct_name, map]} = ast-, fun)
    def to_string(-{:fn, _, [{:->, _, [_, tuple]}] = arrow} = ast-, fun) when -not(is_tuple(tuple))- or -elem(tuple, 0) != :__block__-
    def to_string(-{:fn, _, [{:->, _, _}] = block} = ast-, fun)
    ...
    (14 clauses not shown)
```

Also refactored by extracting functions for `format_args`, `format_clauses` and `non_visible_clauses` and lifted the magic number 10 into a default argument.

This refactoring may bring us closer to being able to implement @kipcole9 's suggestion which would be great: to have a built in way to show all clauses if you need to. More work required to achieve this though.

Closes #9529

### DIFF
--- a/lib/elixir/lib/exception.ex
+++ b/lib/elixir/lib/exception.ex
@@ -1062,6 +1062,8 @@ end
 defmodule FunctionClauseError do
   defexception [:module, :function, :arity, :kind, :args, :clauses]
 
+  @clause_limit 10
+
   @impl true
   def message(exception) do
     case exception do
@@ -1114,7 +1116,7 @@ defmodule FunctionClauseError do
 
     "\n\nThe following arguments were given to #{mfa}:\n" <>
       "#{format_args(args, inspect_fun)}" <>
-      "#{format_clauses(clauses, format_clause_fun)}"
+      "#{format_clauses(clauses, format_clause_fun, @clause_limit)}"
   end
 
   defp format_args(args, inspect_fun) do
@@ -1125,7 +1127,7 @@ defmodule FunctionClauseError do
     end)
   end
 
-  defp format_clauses(clauses, format_clause_fun, limit \\ 10)
+  defp format_clauses(clauses, format_clause_fun, limit)
   defp format_clauses(nil, _, _), do: ""
   defp format_clauses([], _, _), do: ""
 
--- a/lib/elixir/lib/exception.ex
+++ b/lib/elixir/lib/exception.ex
@@ -1107,46 +1107,45 @@ defmodule FunctionClauseError do
 
     mfa = Exception.format_mfa(module, function, arity)
 
-    formatted_args =
-      args
-      |> Enum.with_index(1)
-      |> Enum.map(fn {arg, i} ->
-        ["\n    # ", Integer.to_string(i), "\n    ", pad(inspect_fun.(arg)), "\n"]
-      end)
+    format_clause_fun = fn {args, guards} ->
+      code = Enum.reduce(guards, {function, [], args}, &{:when, [], [&2, &1]})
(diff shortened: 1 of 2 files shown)


lib/elixir/lib/exception.ex
```diff
@@ -1107,37 +1107,46 @@ defmodule FunctionClauseError do
...
-      end
+  defp format_args(args, inspect_fun) do
+    args
+    |> Enum.with_index(1)
+    |> Enum.map(fn {arg, i} ->
+      [pad("\n# "), Integer.to_string(i), pad("\n"), pad(inspect_fun.(arg)), "\n"]
+    end)
+  end
+
+  defp format_clauses(clauses, format_clause_fun, limit \\ 10)
```
josevalim: I am not the biggest fun of default arguments for private functions but I understan the concern is to remove the magic number. What do you think about having a module attribute called `@clauses_limit` and passing it explicitly instead?

Everything else is perfect, thank you!

joshprice: Makes sense was also considering a module attribute. Have updated

josevalim:

PR #9546, opened 2019-11-17
Reload kernel handlers before application starts

### LINKED ISSUE #9545: Erlang logger configuration isn't loaded on startup
### Environment

  ```
  Erlang/OTP 22 [erts-10.4.3] [source] [64-bit] [smp:12:12] [ds:12:12:10] [async-threads:1] [hipe]

  Elixir 1.10.0-dev (3578be2) (compiled with Erlang/OTP 22)
  ```
  But for sure this happens on all current `master`
* Operating system: Doesn't matter

### Current behavior

Configuring Erlang's `logger` via:

```elixir
config :kernel, :logger, [
  {:handler, :example, :logger_std_h, %{config: %{file: 'foo.bar'}}}
]
```

Doesn't add handler with ID `:example` as `kernel` configuration is set in stone when the application is ran via Mix. This can be confusing for users that want to use Erlang logger backends with Elixir as this will be non obvious how to manage them.

### Expected behavior

With configuration given as above this assertion should pass

```elixir
assert :example in :logger.get_handler_ids()
```

hauleth:
However I do not know how this will work with filters and module levels.
I think that these will need to be handled separately.

Close #9545

### DIFF
--- a/lib/logger/lib/logger/app.ex
+++ b/lib/logger/lib/logger/app.ex
@@ -25,6 +25,8 @@ defmodule Logger.App do
       {:ok, sup} ->
         primary_config = add_elixir_handler(otp_reports?, config)
 
+        :logger.add_handlers(:logger)
+
         default_handlers =
           if otp_reports? do
             delete_erlang_handler()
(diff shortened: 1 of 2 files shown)


josevalim:
Thanks @hauleth but I really don't want to go down this path:

1. If we load only some kernel.logger configuration but not all of them, it would be confusing. We would need to validate this would be an option for all logger config but even if we do that...

2. if we load only some kernel configuration, it would be confusing

Let's continue the discussion in #9545.

PR #9548, opened 2019-11-18
Remove TODO to make tests async

devonestes:
I found this TODO that said we should make this `async: true` once
capture_io was updated, and now that it is I'm removing the TODO and
making it async.


josevalim:
It was already removed in master, please make sure to update! Thanks!

Issue #9195, opened 2019-07-05
release fails with "init terminating in do_boot: cannot get bootfile"


1player:
* Elixir & Erlang/OTP versions (elixir --version): 1.9.0
* Operating system: macOS and elixir:1.9.0-alpine

### Current behavior

Release created by `mix release` and executed with `_build/prod/rel/app/bin/app start` fails with error:

```
{"init terminating in do_boot",{'cannot get bootfile','/path/app/_build/prod/rel/app/bin/start.boot'}}
init terminating in do_boot ({cannot get bootfile,/path/app/_build/prod/rel/app/bin/start.boot})
```

The release expects the `start.boot` file to be present in the final bin directory, but it's actually generated in `releases/<version>`

```
$ ls bin/
app
$ ls releases/latest/
consolidated/  env.sh    iex*        start.script      start_clean.script  vm.args
elixir*        app.rel  start.boot  start_clean.boot  sys.config
```

Happens both on macOS and Dockerfile based off elixir:1.9.0-alpine

josevalim:
Can you please provide an app that reproduces the error? I can assemble
releases just fine in my macos. Also, please include your Erlang/OTP
version.
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

1player:
Probably user error: it failed because I edited `rel/env.sh.eex`, uncommenting the last two lines so it reads:

```
# Set the release to work across nodes. If using the long name format like
# the one below (my_app@127.0.0.1), you need to also uncomment the
# RELEASE_DISTRIBUTION variable below.
export RELEASE_DISTRIBUTION=app
export RELEASE_NODE=<%= @release.name %>@127.0.0.2
```

Commenting the RELEASE_DISTRIBUTION variable makes it work again. I need my app to run on 127.0.0.2, and the explanation above says I also need to uncomment RELEASE_DISTRIBUTION but I'm not sure what this actually does, so I guess I need to RTFM.

What does that variable actually do?

josevalim:
Search for the env var name here: https://hexdocs.pm/mix/Mix.Tasks.Release.html#content :)

PR #9549, opened 2019-11-18
Add comment specyfing possible value of RELEASE_DISTRIBUTION inside env.{sh,bat}.eex

esse:
Using any other name there will cause release to fail - I think it's
good to have that information there - see f.e. #9195.


josevalim:

PR #9550, opened 2019-11-18
Add support for NO_COLOR environment variable

hauleth:
This is an implementation of [`NO_COLOR`][1] initiative that proposes single environment variable to prevent using colours in terminal. While I do not use that, I think that some people can be interested in such functionality.

[1]: https://no-color.org

### DIFF
--- a/bin/elixir
+++ b/bin/elixir
@@ -206,8 +206,8 @@ SCRIPT_PATH=$(dirname "$SELF")
 if [ "$OSTYPE" = "cygwin" ]; then SCRIPT_PATH=$(cygpath -m "$SCRIPT_PATH"); fi
 if [ "$MODE" != "iex" ]; then ERL="-noshell -s elixir start_cli $ERL"; fi
 
-if [ "$OS" != "Windows_NT" ]; then
-  if test -t 1 -a -t 2; then ERL="-elixir ansi_enabled true $ERL"; fi
+if [ "$OS" != "Windows_NT" -a -t 1 -a -t 2 -a "$NO_COLOR" != "" ]; then
+  ERL="-elixir ansi_enabled true $ERL"
 fi
 
 ERTS_BIN=
--- a/bin/elixir
+++ b/bin/elixir
@@ -206,8 +206,8 @@ SCRIPT_PATH=$(dirname "$SELF")
 if [ "$OSTYPE" = "cygwin" ]; then SCRIPT_PATH=$(cygpath -m "$SCRIPT_PATH"); fi
 if [ "$MODE" != "iex" ]; then ERL="-noshell -s elixir start_cli $ERL"; fi
 
-if [ "$OS" != "Windows_NT" ] && [ -t 1 ] && [ -t 2 ] && [ ! -z "$NO_COLOR" ]; then
-  ERL="-elixir ansi_enabled true $ERL"
+if [ "$OS" != "Windows_NT" ] && [ -n "$NO_COLOR" ]; then
+  if test -t 1 -a -t 2; then ERL="-elixir ansi_enabled true $ERL"; fi
 fi
 
 ERTS_BIN=
--- a/bin/elixir
+++ b/bin/elixir
@@ -206,7 +206,7 @@ SCRIPT_PATH=$(dirname "$SELF")
 if [ "$OSTYPE" = "cygwin" ]; then SCRIPT_PATH=$(cygpath -m "$SCRIPT_PATH"); fi
 if [ "$MODE" != "iex" ]; then ERL="-noshell -s elixir start_cli $ERL"; fi
 
-if [ "$OS" != "Windows_NT" -a -t 1 -a -t 2 -a "$NO_COLOR" != "" ]; then
+if [ "$OS" != "Windows_NT" ] && [ -t 1 ] && [ -t 2 ] && [ ! -z "$NO_COLOR" ]; then
   ERL="-elixir ansi_enabled true $ERL"
 fi
 


bin/elixir
```diff
@@ -206,8 +206,8 @@ SCRIPT_PATH=$(dirname "$SELF")
 if [ "$OSTYPE" = "cygwin" ]; then SCRIPT_PATH=$(cygpath -m "$SCRIPT_PATH"); fi
 if [ "$MODE" != "iex" ]; then ERL="-noshell -s elixir start_cli $ERL"; fi
 
-if [ "$OS" != "Windows_NT" ]; then
-  if test -t 1 -a -t 2; then ERL="-elixir ansi_enabled true $ERL"; fi
+if [ "$OS" != "Windows_NT" -a -t 1 -a -t 2 -a "$NO_COLOR" != ""]; then
```
josevalim: I believe `-a -t` is not supported inside the `if` in all shell implementations. So please keep this part as it was. I also think the NO_COLOR spec says the variable should be respected regardless of value, so we should probably use `! -z` or similar.

wojtekmach: yeah, there's something wrong here because I'm getting this on macOS:

```
/Users/wojtek/src/elixir/bin/elixir: line 209: [: too many arguments
```

hauleth: These are as `test foo -a bar` and `[ foo -a bar ]` [are required by POSIX to be the same](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/test.html#top)

wojtekmach: My understanding is with this patch we disable ANSI escape codes completely. However, ANSI besides colours have things like italics, underline etc - are these supposed to be disabled by NO_COLOR too?

josevalim: > These are as test foo -a bar and [ foo -a bar ] are required by POSIX to be the same

The problem is that `-t` is not in the standard and it fails in terms of syntax in some bashes but it is just fails in case of `test`.

bin/elixir
```diff
@@ -1,7 +1,7 @@
 #!/bin/sh
 set -e
 
-if [ $# -eq 0 ] || [ "$1" = "--help" ] || [ "$1" = "-h" ]; then
+if [ $# -eq 0 -o "$1" = "--help" -o "$1" = "-h" ]; then
```
josevalim: Please revert these changes. `||` is the style we use in these files and we shouldn't change it. Please use `&&` below.

bin/elixir
```diff
@@ -206,8 +206,8 @@ SCRIPT_PATH=$(dirname "$SELF")
 if [ "$OSTYPE" = "cygwin" ]; then SCRIPT_PATH=$(cygpath -m "$SCRIPT_PATH"); fi
 if [ "$MODE" != "iex" ]; then ERL="-noshell -s elixir start_cli $ERL"; fi
 
-if [ "$OS" != "Windows_NT" ]; then
-  if test -t 1 -a -t 2; then ERL="-elixir ansi_enabled true $ERL"; fi
+if [ "$OS" != "Windows_NT" -a -t 1 -a -t 2 -a ! -z "$NO_COLOR"]; then
```
josevalim: As mentioned in the now deleted commented, `-t` is not going to work as is here. Also please use && instead.

bin/elixir
josevalim: To be clear, we shouldn't change it in this PR. We can do it in a separate one.

hauleth: This requires to spawn multiple `test` instances on each run of the application, that is why I introduced this change.

bin/elixir
hauleth: I have just checked, and it is in POSIX standard, so it should be supported in all `sh` that claim to be POSIX compliant.

@wojtekmach I missed the space before `]`.

bin/elixir
hauleth: `-t fd` is POSIX standard, so if it is ran on POSIX compliant OS then it should work, otherwise this is an OS bug.

bin/elixir
josevalim: Running shellcheck on this file after changes emit these new warnings:

```
In bin/elixir line 4:
if [ $# -eq 0 -o "$1" = "--help" -o "$1" = "-h" ]; then
              ^-- SC2166: Prefer [ p ] || [ q ] as [ p -o q ] is not well defined.
                                 ^-- SC2166: Prefer [ p ] || [ q ] as [ p -o q ] is not well defined.

In bin/elixir line 209:
if [ "$OS" != "Windows_NT" -a -t 1 -a -t 2 -a ! -z "$NO_COLOR" ]; then
                           ^-- SC2166: Prefer [ p ] && [ q ] as [ p -a q ] is not well defined.
                                   ^-- SC2166: Prefer [ p ] && [ q ] as [ p -a q ] is not well defined.
                                           ^-- SC2166: Prefer [ p ] && [ q ] as [ p -a q ] is not well defined.
                                              ^-- SC2236: Use -n instead of ! -z.


For more information:
  https://www.shellcheck.net/wiki/SC2166 -- Prefer [ p ] && [ q ] as [ p -a q...
```

Please address them.

bin/elixir
josevalim: You are correct. It seems the issue is not a matter of standard, but rather people not following the standard. Please see the commit that introduced it:

https://github.com/elixir-lang/elixir/commit/a2662a4b511e52c4dc44ad4acfafd44ac5861d05

If you have a Windows machine and you can verify if it works on MinGW and Cygwin, then we will gladly accept the change, otherwise I would prefer to leave it as is (especially we want to avoid using `-a`, so the number of test calls end up being the same).

bin/elixir
hauleth: Fixed.

bin/elixir
hauleth: With change to use `&&` it shouldn't be a problem, as `&&` lazy evaluates.

josevalim:

bin/elixir
josevalim: IIRC, the issue was that it would not even parse it. Again, I will be glad to change it if someone actually verifies it on MinGW and Cygwin.

PR #9551, opened 2019-11-18
Remove the URI example for `mix archive.install` help text

joshprice:
This makes the `mix archive.install` help text consistent with the other changes from
commit 4b6b81098628f5f539a91b56357575293ca2bf55 and the `mix escript.install` help text.

Without a PR or any real description to explain this being intentional, the diff seems to imply that this may have been overlooked.


josevalim:

PR #9554, opened 2019-11-19
[ci skip] Copyedit deprecations table 

eksperimental:
~Please keep both commits when merging~

## 1. Sort "Table of deprecations" alphabetically
Deprecations are sorted,
1st by version.
2nd by Deprecated feature, which follows the following order:
- Changes that do not belong a particular module such as a feature.
- Changes that belong to a regular module. Either functions or options related it. If functions do not explicitly mention the autoimported modules such as `Kernel` and `Kernel.SpecialForms` they are still sorted based on its module name.

Elixir apps such as "Mix", "EEx", "ExUnit", "Logger" are sorted by the name of the module,
as any other module.

## 2. Copyedit the deprecations table.


josevalim:
Can you please remove or change the 2nd ordering? It is unlikely we will preserve such ordering in the long term, so or we keep the second ordering purely alphabetical OR we keep it random as it was. Thank you.

eksperimental:
Initially I did it as you are suggesting. But since some special forms didn't explicitly mention the module, looked odd to have them in thy middle of the list (such as in the case of `not left in right`).
Should we always try to mention the module when possible?

josevalim:
I think we should write it in a way that is best to read, and not try to force some conditions because of sorting. So if adding a module improves things, we should do it, otherwise we shouldn't. I don't think adding `Kernel` to `not left in right` would help, especially because it is not a `Kernel` restriction but a syntax one.

eksperimental:
done

josevalim:

PR #9555, opened 2019-11-19
Example in defmodule shows the return value as mentioned in the list

eksperimental:


josevalim:

PR #9556, opened 2019-11-19
Explicitly mention that def and defmacro define public functions/macros

eksperimental:
The concept of "public functions" is used in other parts of the module, but it is never mentioned how these functions are created.


josevalim:

PR #9558, opened 2019-11-19
Fix recently introduced bug in NO_COLOR feature

eksperimental:
Before this fix, colors are disabled in my shell, and NO_COLOR variable is unset.

Fixes a bug introduced in 499781d5dd304964981fce093ae7d1ade56d4351 (#9550)
where the variable is checked to be set, when it should be the opposite.

Additionally it removes the use of the "-a" as suggested in the POSIX
especifications
The Open Group Base Specifications Issue 7, 2018 edition
https://pubs.opengroup.org/onlinepubs/9699919799/utilities/test.html

"-a" is becoming obsolete and it may be removed in future versions.

> ...in cases where maximal portability is of concern, replace:
> test expr1 -a expr2
> with:
> test expr1 && test expr2

/cc @hauleth @josevalim @wojtekmach


josevalim:

PR #9559, opened 2019-11-19
Use etc. instead of etc

eksperimental:
[ci skip]

### DIFF
--- a/lib/elixir/lib/io.ex
+++ b/lib/elixir/lib/io.ex
@@ -84,7 +84,7 @@ defmodule IO do
   Building IO data is cheaper than concatenating binaries. Concatenating multiple
   pieces of IO data just means putting them together inside a list since IO data
   can be arbitrarily nested, and that's a cheap and efficient operation. Most of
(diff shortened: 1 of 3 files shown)


whatyouhide:
From my experience writing, my personal suggestion is to drop contractions like this (and "i.e.", "e.g.", and so on) because they make text easier to read for non-native English speakers. I would prefer to go with "and so on" instead of "etc." and since we're changing these occurrences, it might be a good chance to do it now. Thoughts?

eksperimental:
I totally agree. In fact, the first paragraph of what you read in the Elixir documentation mentions "etc.".
Should we extend this change to all of them?

eksperimental:
and we don't necessarily need to always replace them with "and so on" as there are a handful of expressions to replace them and that some would suit better than others depending on the context.

whatyouhide:
> Should we extend this change to all of them?

Yep, IMO.

> and we don't necessarily need to always replace them with "and so on" as there are a handful of expressions to replace them and that some would suit better than others depending on the context, making the reading more natural.

As long as it's not complex English, I'm all for it :)

PR #9560, opened 2019-11-19
Refer properly to Erlang datetime tuple

eksperimental:


whatyouhide:
Thank you @eksperimental! 💟

PR #9561, opened 2019-11-19
Use argument instead of arg in error messages

eksperimental:
Avoid use of abbreviations in error messages

### DIFF
--- a/lib/elixir/test/elixir/kernel/expansion_test.exs
+++ b/lib/elixir/test/elixir/kernel/expansion_test.exs
@@ -1201,16 +1201,18 @@ defmodule Kernel.ExpansionTest do
     end
 
     test "expects one argument in clauses" do
-      assert_raise CompileError, ~r"expected one arg for :do clauses \(->\) in \"cond\"", fn ->
-        code =
-          quote do
-            cond do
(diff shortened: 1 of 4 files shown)


lib/elixir/test/elixir/kernel/expansion_test.exs
```diff
@@ -1201,16 +1201,18 @@ defmodule Kernel.ExpansionTest do
...
+                       quote do
+                         cond do
+                           _, _ -> :ok
+                         end
+                       end
 
-        expand(code)
-      end
+                     expand(code)
+                   end
```
eksperimental: this whole chunk is changed by the formatter. the only change is in the error message

lib/elixir/test/elixir/kernel/expansion_test.exs
```diff
@@ -1384,16 +1386,18 @@ defmodule Kernel.ExpansionTest do
...
+                       quote do
+                         case e do
+                           _, _ -> :ok
+                         end
+                       end
 
-        expand(code)
-      end
+                     expand(code)
+                   end
```
eksperimental: same here

josevalim:

PR #9562, opened 2019-11-19
Style mention to operating systems

eksperimental:
Use:
- macOS, instead of MacOS X;
- Windows 7, instead of Win7
- Unix, instead of UNIX
- Unix-like operating system or Unix-like system, instead of just Unix

### DIFF
--- a/lib/elixir/lib/file.ex
+++ b/lib/elixir/lib/file.ex
@@ -6,7 +6,7 @@ defmodule File do
   to interact with files or IO devices, like `open/2`,
   `copy/3` and others. This module also provides higher
   level functions that work with filenames and have their naming
-  based on UNIX variants. For example, one can copy a file
+  based on Unix variants. For example, one can copy a file
   via `cp/3` and remove files and directories recursively
   via `rm_rf/1`.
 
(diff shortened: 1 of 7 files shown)


josevalim:
I am not convinced going to Unix-like systems is an improvement. Can't we just keep it as Unix? What is the benefit?

eksperimental:
It is not technically correct IMO.

Unix (trademarked as UNIX) is the operating system that derive from the original AT&T Unix, development starting in the 1970s at the Bell Labs research center by Ken Thompson, Dennis Ritchie, and others.

From: https://en.wikipedia.org/wiki/Unix
> By decree of The Open Group, the term "UNIX" refers more to a class of operating systems than to a specific implementation of an operating system; those operating systems which meet The Open Group's Single UNIX Specification should be able to bear the UNIX 98 or UNIX 03 trademarks today, after the operating system's vendor pays a substantial certification fee and annual trademark royalties to The Open Group. Systems licensed to use the UNIX trademark include AIX,[29] HP-UX, Inspur K-UX, IRIX, Solaris, Tru64 UNIX (formerly "Digital UNIX", or OSF/1), macOS, and a part of z/OS. Notably, Inspur K-UX is a Linux distribution certified as UNIX 03 compliant.

From: https://en.wikipedia.org/wiki/Unix-like
> A Unix-like (sometimes referred to as UN*X or *nix) operating system is one that behaves in a manner similar to a Unix system.

the current version of macOS can be called a Unix, but that's not the case with Linux.
For that reason I think the appropriate term is Unix-like, and not Unix/UNIX.

OvermindDL1:
Technically Mac's are more unix-like than linux is, they actually more properly fulfill posix, where linux doesn't.

josevalim:

PR #9566, opened 2019-11-19
Make Cirrus build button green

eksperimental:
The build button in the README file is red, due to test marked
with "allow_failures: true" not passing.

Adds "skip_notifications: true"

/cc @fertapric @ggcampinho


fertapric:
Good catch @eksperimental! 💜

eksperimental:
It works but the image is cached and doesn't get updated

PR #9569, opened 2019-11-20
Replace usage of "etc."

eksperimental:
Related discussion: https://github.com/elixir-lang/elixir/pull/9559

/cc @whatyouhide

### DIFF
--- a/lib/elixir/lib/calendar.ex
+++ b/lib/elixir/lib/calendar.ex
@@ -11,7 +11,7 @@ defmodule Calendar do
   For the actual date, time and datetime structures, see `Date`,
   `Time`, `NaiveDateTime` and `DateTime`.
 
-  Note the year, month, day, etc. designations are overspecified
+  Note designations for year, month, day, and the like, are overspecified
   (i.e. an integer instead of `1..12` for months) because different
   calendars may have a different number of days per month, months per year and so on.
   """
@@ -65,7 +65,7 @@ defmodule Calendar do
   @typedoc "The time zone ID according to the IANA tz database (e.g. Europe/Zurich)"
   @type time_zone :: String.t()
 
-  @typedoc "The time zone abbreviation (e.g. CET or CEST or BST etc.)"
+  @typedoc "The time zone abbreviation (e.g. CET, CEST, BST, et cetera)."
   @type zone_abbr :: String.t()
 
   @typedoc "The time zone UTC offset in seconds"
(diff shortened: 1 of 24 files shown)


lib/elixir/lib/calendar.ex
```diff
@@ -65,7 +65,7 @@ defmodule Calendar do
   @typedoc "The time zone ID according to the IANA tz database (e.g. Europe/Zurich)"
   @type time_zone :: String.t()
 
-  @typedoc "The time zone abbreviation (e.g. CET or CEST or BST etc.)"
+  @typedoc "The time zone abbreviation (e.g. CET, CEST, BST, et cetera)."
```
whatyouhide: I don't think "et cetera" is an improvement. Please let's use simple wording, "and so on" works in every case.

lib/elixir/lib/kernel/special_forms.ex
```diff
@@ -9,8 +9,8 @@ defmodule Kernel.SpecialForms do
   been automatically imported, in the same fashion as the functions and
   macros from the `Kernel` module.
 
-  These building blocks are defined in this module. Some of these special forms are lexical (like
-  `alias/2`, `case/2`, etc.). The macros `{}/1` and `<<>>/1` are also special
+  These building blocks are defined in this module. Some of these special forms are lexical (such as
+  `alias/2` and `case/2`). The macros `{}/1` and `<<>>/1` are also special
```
whatyouhide: I believe you want "or" here? 🤔

eksperimental: I thouhgt of it. I think it is AND because both are lexical. See the following sentence, it also uses AND to convey a similar message.

lib/elixir/lib/calendar.ex
eksperimental: Just to clarify the intention of the PR. We never said that the non-native speaker doesn't know the meaning of "et cetera". What we were trying to tackle is that it wouldn't know how to read the contraction, which is not "etc." is many languages.

I will change it anyway.

whatyouhide: Sorry, I thought it was implied. We have zero gain using "et cetera" instead of "and so on", and I’m all for trying to be as inclusive as possible to people with different English backgrounds, so I'd just go for the simpler language 🙃

eksperimental: No worries. All __et cetera__ entries have been replaced.

whatyouhide:
@eksperimental rebase this and you should be good to go :)

eksperimental:
@whatyouhide done

josevalim:

Issue #9567, opened 2019-11-19
Release with :tar without :include_erts fails


lasseebert:
### Environment

* Elixir & Erlang/OTP versions (elixir --version): Elixir 1.9.4 (compiled with Erlang/OTP 21)
* Operating system: Ubuntu 19.04

### Current behavior

* I have a release with `include_erts: false`, since I'm building it for an embedded device where erlang should be cross-compiled to ARM,
* My steps in the release are `[:assemble, :tar]`

When I run `mix release`, it tries to add erts to the tar, but fails: 

```
* assembling my_app-0.4.4 on MIX_ENV=dev
* skipping runtime configuration (config/releases.exs not found)
* creating _build/dev/rel/my_app/releases/0.4.4/vm.args
* creating _build/dev/rel/my_app/releases/0.4.4/env.sh
* building /path/to/my_app/_build/dev/my_app-0.4.4.tar.gz
** (MatchError) no match of right hand side value: {:error, {'/path/to/my_app/_build/dev/rel/my_app/erts-10.1', :enoent}}
    (mix) lib/mix/tasks/release.ex:1039: Mix.Tasks.Release.make_tar/1
    (mix) lib/mix/tasks/release.ex:959: Mix.Tasks.Release.run_steps/1
    (mix) lib/mix/task.ex:331: Mix.Task.run_task/3
    (mix) lib/mix/cli.ex:79: Mix.CLI.run_task/2
```

### Expected behavior

I was hoping it would not include erts in the tar.

These lines of code seems to be the critical place: https://github.com/elixir-lang/elixir/blob/cf4272d22c2b4594c40a4c5dc971de888efd37b6/lib/mix/lib/mix/tasks/release.ex#L1044-L1046

Am I right in my assumption about the expected behaviour?

Gazler:
I think we should exclude `"erts-#{release.erts_version}"` if `include_erts` is set to false.

lasseebert:
:+1: I will make a PR as suggestion to expected behaviour.

lasseebert:
PR submitted --^
The `make_tar` function also tries to add erlang libs that does not exist when releasing with `include_erts: false`, so I have excluded those in `make_tar`, but perhaps they should be excluded earlier in the release process?

josevalim:
Those applications are part of the release, they just are not copied over. I left a comment on the PR talking about how to handle them.

josevalim:
Closing in favor of the PR!

PR #9570, opened 2019-11-20
Fix release tar when include_erts is false

lasseebert:
Ref #9567 

Now explicitly excludes `erts-x.x` dir and erlang libs when building a tar for a release that has `include_erts: false`.

Otherwise building the tar would fail with `:enoent`.

I'm not sure if this is the correct place to filter out the erlang libs. It feels like they should not be part of `release.applications` at all.

### DIFF
--- a/lib/mix/lib/mix/tasks/release.ex
+++ b/lib/mix/lib/mix/tasks/release.ex
@@ -1033,7 +1033,9 @@ defmodule Mix.Tasks.Release do
         lib_dirs ++ release_files
 
     files =
-      Enum.map(dirs, &{String.to_charlist(&1), String.to_charlist(Path.join(release.path, &1))})
+      dirs
+      |> Enum.filter(&File.exists?(Path.join(release.path, &1)))
+      |> Enum.map(&{String.to_charlist(&1), String.to_charlist(Path.join(release.path, &1))})
 
     File.rm(out_path)
     :ok = :erl_tar.create(String.to_charlist(out_path), files, [:dereference, :compressed])
--- a/lib/mix/test/mix/tasks/release_test.exs
+++ b/lib/mix/test/mix/tasks/release_test.exs
@@ -67,6 +67,8 @@ defmodule Mix.Tasks.ReleaseTest do
           assert "releases/0.1.0/vm.args" in files
           assert "releases/COOKIE" in files
           assert "releases/start_erl.data" in files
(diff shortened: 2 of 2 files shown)


lib/mix/lib/mix/tasks/release.ex
```diff
@@ -1017,8 +1017,15 @@ defmodule Mix.Tasks.Release do
     out_path = Path.join([release.path, "..", "..", tar_filename]) |> Path.expand()
     info(release, [:green, "* building ", :reset, out_path])
 
+    include_erts = not is_nil(release.erts_source)
+    erlang_libs = [:compiler, :kernel, :runtime_tools, :sasl, :stdlib]
```
josevalim: Instead of checking for specific erlang libs, we should just reject any application that does not have the according directory in "lib/#{name}-#{vsn}". If they are not there, then it means it came from erts. Same thing can be done by for erts. Check if the dir is there, if it isn't, do not add it.

lib/mix/test/mix/tasks/release_test.exs
```diff
@@ -79,6 +80,28 @@ defmodule Mix.Tasks.ReleaseTest do
...
+          tar_path = Path.expand(Path.join([root, "..", "..", "demo-0.1.0.tar.gz"]))
+          message = "* building #{tar_path}"
+          assert_received {:mix_shell, :info, [^message]}
+          assert File.exists?(tar_path)
+
+          {:ok, files} = String.to_charlist(tar_path) |> :erl_tar.table([:compressed])
+          files = Enum.map(files, &to_string/1)
+
+          assert "bin/demo" in files
+          refute Enum.any?(files, &(&1 =~ "erts"))
```
josevalim: We should also check the applications have not been added?

lib/mix/lib/mix/tasks/release.ex
lasseebert: :+1: Will fix

lib/mix/test/mix/tasks/release_test.exs
lasseebert: Yes, good point.

lasseebert:
@josevalim Much simpler code now :) Now skips files and dirs that are not present in the file system. Also added your suggested assertion that erlang application is not added.

fertapric:
Thanks @lasseebert! 💜

PR #9573, opened 2019-11-21
Improvements in .cirrus.yml and increase execution speed

eksperimental:
Improvements in .cirrus.yml and increase execution speed
    
    - Make use of $CIRRUS_WORKING_DIR instead of spawning a new process with $(pwd).
    - Only run the Elixir formatter, in the latest OTP version, in the same fashion of
      CHECK_POSIX_COMPLIANT and CHECK_REPRODUCIBLE.
    - TEST_FORMATTED environmetal variable added.
    - move fast executing scripts, such as "dialyzer" to the top of the list.
    - move "rm -rf .git" and "make compile" to the install script.
    - Put each command their own named scripts (i.e., test_formatted_script, dialyzer_script, test_erlang_script, test_elixir_script, check_posix_compliant_script, check_reproducible_script).
    - Split in the Windows task, the test_script into install_script and test_script, to mimic the scripts in the test_linux_task.
    - Messages added on success or when tests are skipped, so the user is better informed in the CI results page.
    - Added script that shows the output of `elixir --version`.
    - Left a comment in "test" task in Makefile to update accoringly .cirrus.yml if updated.
    
    Speed increase: The building times for the Linux builds have been reduced considerable, to 5 minutes. Sadly, setting the number of cpus for Windows doesn't work, and it takes 13 minutes to build.
    
    - CPU increased from the default value of 2, to 8
    - Memory decreased from the default value of 4GB, to 1500MiB (bringing it to a minimum needed, to speed up the
      scheduling in the CI).
    
    See the results in:
    https://cirrus-ci.com/build/6585376826720256

/cc @fertapric

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,7 +1,9 @@
 env:
-  ELIXIR_ASSERT_TIMEOUT: 2000
-  LANG: C.UTF-8
   CIRRUS_CLONE_DEPTH: 1
+  ELIXIR_ASSERT_TIMEOUT: 2000
+  ELIXIRC_OPTS: "--warnings-as-errors"
+  ERLC_OPTS: "+warning_as_errors"
+  LANG: C.UTF-8
 
 test_linux_task:
   name: Linux, ${OTP_RELEASE}, Ubuntu 14.04
@@ -12,8 +14,6 @@ test_linux_task:
     memory: 1500Mi
 
   env:
-    ELIXIRC_OPTS: "--warnings-as-errors"
-    ERLC_OPTS: "+warning_as_errors"
     PATH: "${CIRRUS_WORKING_DIR}/otp/bin:${PATH}"
 
   matrix:
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -3,7 +3,7 @@ env:
   LANG: C.UTF-8
 
 test_linux_task:
-  name: Test (Linux - Ubuntu 14.04) / ${OTP_RELEASE}
+  name: Linux, ${OTP_RELEASE}, Ubuntu 14.04
 
   container:
     image: buildpack-deps:trusty
@@ -83,18 +83,22 @@ test_linux_task:
     fi
 
 test_windows_task:
-  name: Test (Windows Server 2019)
+  name: Windows, OTP-${OTP_RELEASE}, Windows Server 2019
+
+  matrix:
+    - env:
+        OTP_RELEASE: 22.0
+        OS_VERSION: 2019
+
+    - env:
+        OTP_RELEASE: 21.0.1
+        OS_VERSION: 2019
 
   windows_container:
+    image: fertapric/elixir-ci:otp-win64-${OTP_RELEASE}
+    os_version: ${OS_VERSION}
     cpu: 8
     memory: 1500Mi
-
-    matrix:
-      - image: fertapric/elixir-ci:otp-win64-22.0
-        os_version: 2019
-
-      - image: fertapric/elixir-ci:otp-win64-21.0.1
-        os_version: 2019
 
   install_script:
     - rmdir /s /q .git
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -2,15 +2,25 @@ env:
   ELIXIR_ASSERT_TIMEOUT: 2000
   LANG: C.UTF-8
 
-test_task:
+test_linux_task:
+  name: Test (Linux - Ubuntu 14.04) / ${OTP_RELEASE}
+
   container:
     image: buildpack-deps:trusty
+    cpu: 8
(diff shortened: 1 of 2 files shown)


.cirrus.yml
```diff
@@ -2,15 +2,25 @@ env:
   ELIXIR_ASSERT_TIMEOUT: 2000
   LANG: C.UTF-8
 
-test_task:
+test_linux_task:
+  name: Test (Linux - Ubuntu 14.04) / ${OTP_RELEASE}
```
fertapric:
```suggestion
  name: Test ${OTP_RELEASE} on Linux (Ubuntu 14.04)
```

The main goal is to have the version of the OTP release in the sidebar:

Screen Shot 2019-11-21 at 09 46 45

Makefile
```diff
@@ -236,6 +236,7 @@ zips: Precompiled.zip Docs.zip
 
 #==> Test tasks
 
+# If you modify this task, please update .cirrus.yml accordinly
```
fertapric:
```suggestion
# If you modify this task, please update .cirrus.yml accordingly
```

.cirrus.yml
```diff
@@ -34,36 +44,64 @@ test_task:
...
+  install_script:
     - rmdir /s /q .git
-    - make
+    - make compile
+
+  info_script:
+    - bin/elixir --version
+
+  test_script:
     - make --keep-going test_erlang test_elixir
```
fertapric: I think this can also be split in "Test Erlang" and "Test Elixir"

.cirrus.yml
```diff
@@ -2,15 +2,25 @@ env:
   ELIXIR_ASSERT_TIMEOUT: 2000
   LANG: C.UTF-8
```
fertapric: I think there is no need to clone the whole repo:

```suggestion
  CIRRUS_CLONE_DEPTH: 1

```

More info at: https://cirrus-ci.org/guide/writing-tasks/#behavioral-environment-variables

.cirrus.yml
```diff
@@ -34,36 +44,64 @@ test_task:
...
+    - make compile
+
+  info_script:
+    - bin/elixir --version
+
+  test_formatted_script: |
+    if [ -n "$TEST_FORMATTED" ]; then
+      make test_formatted && echo "All files found to be properly formatted."
+    else
+      echo "No need for this test in this Erlang/OTP version."
```
fertapric: Nitpick: instead of saying "No need for this test", we can say where to find this test:

```suggestion
      echo "The format of the code is checked in the last stable Erlang/OTP version."
```

.cirrus.yml
eksperimental: Excelent comment. Please have a look now

Makefile
eksperimental: fixed

.cirrus.yml
eksperimental: So true. 
done!

.cirrus.yml
```diff
@@ -34,36 +45,67 @@ test_task:
...
+      shellcheck bin/iex && echo "bin/iex is POSIX compliant"
+    else
+      echo "The format of the shell scripts is only checked in the last stable Erlang/OTP version."
+    fi
+
+  check_reproducible_script: |
+    if [ -n "$CHECK_REPRODUCIBLE" ]; then
+      make check_reproducible
+    else
+      echo "The format reproducibility of the build is only checked in the last stable Erlang/OTP version."
```
fertapric:
```suggestion
      echo "The reproducibility of the build is only checked in the last stable Erlang/OTP version."
```

josevalim:
Let’s please check it is formatted on all runs. The reason why libraries only check against the latest Elixir is because they may be using new formatter features. We don’t have this issue here. In fact, we want to check the formatter is correct on all builds.

eksperimental:
> Let’s please check it is formatted on all runs. The reason why libraries only check against the latest Elixir is because they may be using new formatter features. We don’t have this issue here. In fact, we want to check the formatter is correct on all builds.

How about the Windows tasks? Any reason why we are not doing it?

fertapric:
> How about the Windows tasks? Any reason why we are not doing it?

No reason. I think I copied an old `.cirrus.yml` configuration 😞 

`test_formatted` is compatible with Windows since https://github.com/elixir-lang/elixir/commit/5f502d0bca5e5de4b01aec3a2617c2068b4edbd9

eksperimental:
@fertapric great, I was going to ask about this, but in a new issue as this PR had too many changes. But I will add this feature now that I know then

eksperimental:
Please have a look to the new changes.

fertapric:
Awesome job @eksperimental! 👏 💜

PR #9576, opened 2019-11-21
Standardize indentation in Makefile

eksperimental:
Use TAB to ident upto parent command level,
and use white-spaces within shell expressions,
such as in if and for.


josevalim:
Honestly, I would prefer to use tabs for everything, but in the lack of a formatter for Makefile, I don’t think any style changes are worth it, because we cannot ensure they will be kept. So I will close this one for now, thanks!

PR #9577, opened 2019-11-22
Clean residual files after check_reproducible

eksperimental:
Clean residual files after check_reproducible Make task successfully finished


josevalim:
We always leave entries inside the tmp directory in a test run, as it is properly git ignored, so it is OK to do the same here too.

eksperimental:
Yes, that's true.

PR #9580, opened 2019-11-22
Clean elixir improvement

eksperimental:
* clean_elixir  target should delete .app files
* Add clean_elixir to .PHONY targets in Makefile


josevalim:
Clean_elixir is used by core developers to rebootstrap the language. It should not delete .app files. Thank you.

PR #9581, opened 2019-11-22
Introduce clean_docs file to Makefile

eksperimental:


josevalim:
I will close as we don’t need this in our workflow and doc is also gitignored. Thanks!

eksperimental:
We can leave out `clean_docs` but make use of the CLEAN_DOCS variable. as it is less error-prone to this, than to write each `rm -rf` individually

josevalim:
It is fine as is, thanks. :)

PR #9582, opened 2019-11-22
Introduce clean_erlang Make target

eksperimental:


josevalim:
Is there any use case for this addition?
-- 


*José Valim*
www.plataformatec.com.br
Founder and Director of R&D

josevalim:
I will close this as deleting only the Erlang files does not have much purpose. We only have clean_elixir for core developers to test rebootstrapping the language. Make Erlang on the other hand is always fine without cleaning it before.

eksperimental:
> I will close this as deleting only the Erlang files does not have much purpose. We only have clean_elixir for core developers to test rebootstrapping the language. Make Erlang on the other hand is always fine without cleaning it before.

I dind't know it was only for that purpose.
Why don't we just delete `Elixir.Kernel.beam` if that's the case then? (unless there are other cases)

eksperimental:
> I will close this as deleting only the Erlang files does not have much purpose. We only have clean_elixir for core developers to test rebootstrapping the language. Make Erlang on the other hand is always fine without cleaning it before.

I thought eventually when working on the erlang files, you may need to clean everything to be recompiled, but I see what you mean.

PR #9583, opened 2019-11-22
Fix Mix.Tasks.Compile.App output message

eksperimental:
When compiling Elixir, the "Generated" message was not consistent.
"Generated elixir.app" for Elixir and "Generated mix app",
for the rest.

```
Generated elixir.app
==> mix (compile)
Generated mix app
==> ex_unit (compile)
Generated ex_unit app
```

This commits add the "."


josevalim:
In this case we should change the Elixir one, as it only happens during
Elixir’s build.
-- 


*José Valim*
www.plataformatec.com.br
Founder and Director of R&D

eksperimental:
Well, either way. Thank you.

PR #9585, opened 2019-11-22
[ci] Update config to deprioritize tasks that can fail

fkorotkov:
Attempt to fix https://github.com/cirruslabs/cirrus-ci-docs/issues/524

/cc @eksperimental

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -6,8 +6,6 @@ env:
   LANG: C.UTF-8
 
 test_linux_task:
-  name: Linux, ${OTP_RELEASE}, Ubuntu 14.04
-
   container:
     image: buildpack-deps:trusty
     cpu: 8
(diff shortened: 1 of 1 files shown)


josevalim:
Ah, nice, I think this example will work. Thank you! :heart: We just need to swap things so `depends_on` is done based on OTP version instead of Linux version.

eksperimental:
Thank you @fkorotkov. We also need to put first the OTP version and it gets cut in the sidebar inside the CHECKS tab. I will see what I can do.

.cirrus.yml
```diff
@@ -17,28 +15,33 @@ test_linux_task:
...
+        - env:
+            OTP_RELEASE: OTP-21.3.8
+        - env:
+            OTP_RELEASE: OTP-21.2
+        - env:
+            OTP_RELEASE: OTP-21.1
+        - env:
+            OTP_RELEASE: OTP-21.0
+    - name: Linux Nightly, Ubuntu 14.04
+      depends_on: Linux, Ubuntu 14.04
```
eksperimental: it should also depend one the Windows task

eksperimental:
Looking at it , it will make things go slower, because it will have to wait for all the tasks in Linux to finish to start executing the other two.
The idea was for windows to execute before `maint` and `master`, but if we depend on windows, they whole suit will take 20 minutes instead of 15 that the windows build takes.

I think it will be better to place first the windows task first, and let the rest execute as we were doing.
@fkorotkov  Is an ordered maintained based on how they appear in the .yml file?

fkorotkov:
Order can be defined only via `depends_on` otherwise everything is executed in parallel ideally. Then there is the concurrent limitations for free OSS tasks that can affect any triggered tasks for a given user depending on which task will win the lock to start itself.

eksperimental:
> The idea was for windows to execute before `maint` and `master`, but if we depend on windows, they whole suit will take 20 minutes instead of 15 that the windows build takes.

I see that this is not true, as the build is marked as finished, when tasks with `skip_notifications: true` are still running.

I took a stab at it, based on this PR, 
https://github.com/eksperimental/elixir/pull/8/ 
but what I'm not happy with is with the labels/names, you can see them here,
as there we are not able to create a custom name for each individual task and still link it from "depend_on"
https://github.com/eksperimental/elixir/pull/8/checks?check_run_id=317781720

As of now, not implementing this is not an issue, as Windows tasks are taking longer to start and more than twice the time to execute, so there's plenty of time for the developement branches to run and finish. Maybe we @fkorotkov implements the `weight` feature, we can add it, and so the labels don't get messed up.

josevalim:
Sounds good! These improvements are in no way a requirement and we have overall shorter times than Travis, so everything looks great from our side. THank you @eksperimental and @fkorotkov!

PR #9589, opened 2019-11-23
Raise if a non-absolute path is given with host or authority

### LINKED ISSUE #9588: Generating URI with relative path
### Precheck

* Do not use the issue tracker for help or support (try Elixir Forum, Stack Overflow, IRC, etc.)
* Please disclose security vulnerabilities privately at elixir-security@googlegroups.com

### Environment

* Elixir & Erlang/OTP versions (elixir --version): 1.9.4
* Operating system: OS X

### Current behavior

Hello folks. It looks like this behavior can be improved:

```elixir
iex(10)> %URI{scheme: "gid", path: "hello/123"} |> URI.to_string()
"gid:hello/123"
iex(11)> %URI{scheme: "gid", authority: "foo.com", path: "hello/123"} |> URI.to_string()
"gid://foo.comhello/123"
iex(12)> %URI{scheme: "gid", host: "foo.com", path: "hello/123"} |> URI.to_string()
"gid://foo.comhello/123"
```

So it produces a non-deterministic result in a sense that if it will be parsed, the URI structure will not be the same as the source one.

### Expected behavior

Would be nice of `URI.to_string()` either to prepend `/` to the path automatically when the path is relative and host is present

```elixir
iex(10)> %URI{scheme: "gid", path: "hello/123"} |> URI.to_string()
"gid:hello/123"
iex(11)> %URI{scheme: "gid", authority: "foo.com", path: "hello/123"} |> URI.to_string()
"gid://foo.com/hello/123"
iex(12)> %URI{scheme: "gid", host: "foo.com", path: "hello/123"} |> URI.to_string()
"gid://foo.com/hello/123"
```

or to raise an exception on relative path passed like Ruby does, for example.

```ruby
2.6.5 :001 > URI::Generic.build(scheme: 'http', path: 'foo/bar').to_s
Traceback (most recent call last):
        1: from /Users/pyromaniac/.rvm/rubies/ruby-2.6.5/lib/ruby/2.6.0/uri/generic.rb:761:in `check_path'
URI::InvalidComponentError (bad component(expected absolute path component): foo/bar)
2.6.5 :002 > URI::Generic.build(scheme: 'http', host: 'baz', path: 'foo/bar').to_s
Traceback (most recent call last):
        1: from /Users/pyromaniac/.rvm/rubies/ruby-2.6.5/lib/ruby/2.6.0/uri/generic.rb:761:in `check_path'
URI::InvalidComponentError (bad component(expected absolute path component): foo/bar)
```

I personally would prefer the first one.

I can even try to prepare a PR with the fix as soon as some decision will be made.

Thanks!

josevalim:
Closes #9588.

### DIFF
--- a/lib/elixir/lib/uri.ex
+++ b/lib/elixir/lib/uri.ex
@@ -628,7 +628,7 @@ end
 defimpl String.Chars, for: URI do
   def to_string(%{host: host, authority: authority, path: path} = uri)
       when (host != nil or authority != nil) and is_binary(path) and
-             binary_part(path, 0, 1) != "/" do
+             path != "" and binary_part(path, 0, 1) != "/" do
     raise ArgumentError,
           ":path in URI must be nil or an absolute path if :host or :authority are given, " <>
             "got: #{inspect(uri)}"
--- a/lib/elixir/lib/uri.ex
+++ b/lib/elixir/lib/uri.ex
@@ -630,7 +630,7 @@ defimpl String.Chars, for: URI do
       when (host != nil or authority != nil) and is_binary(path) and
              binary_part(path, 0, 1) != "/" do
     raise ArgumentError,
-          ":path in URI must be nil or an absolute path if host or authority are given, " <>
+          ":path in URI must be nil or an absolute path if :host or :authority are given, " <>
             "got: #{inspect(uri)}"
   end
 
--- a/lib/elixir/lib/uri.ex
+++ b/lib/elixir/lib/uri.ex
@@ -626,6 +626,14 @@ defmodule URI do
 end
 
 defimpl String.Chars, for: URI do
+  def to_string(%{host: host, authority: authority, path: path} = uri)
+      when (host != nil or authority != nil) and is_binary(path) and
+             binary_part(path, 0, 1) != "/" do
+    raise ArgumentError,
+          ":path in URI must be nil or an absolute path if host or authority are given, " <>
+            "got: #{inspect(uri)}"
+  end
+
   def to_string(%{scheme: scheme, port: port, path: path, query: query, fragment: fragment} = uri) do
     uri =
       case scheme && URI.default_port(scheme) do
--- a/lib/elixir/test/elixir/uri_test.exs
+++ b/lib/elixir/test/elixir/uri_test.exs
@@ -342,9 +342,18 @@ defmodule URITest do
     assert to_string(URI.parse("http://[2001:db8::]")) == "http://[2001:db8::]"
 
     assert URI.to_string(URI.parse("http://google.com")) == "http://google.com"
+    assert URI.to_string(URI.parse("gid:hello/123")) == "gid:hello/123"
 
     assert URI.to_string(URI.parse("//user:password@google.com/")) ==
              "//user:password@google.com/"
+
+    assert_raise ArgumentError,
+                 ~r":path in URI must be nil or an absolute path if host or authority are given",
+                 fn -> %URI{authority: "foo.com", path: "hello/123"} |> URI.to_string() end
+
+    assert_raise ArgumentError,
+                 ~r":path in URI must be nil or an absolute path if host or authority are given",
+                 fn -> %URI{host: "foo.com", path: "hello/123"} |> URI.to_string() end
   end
 
   test "merge/2" do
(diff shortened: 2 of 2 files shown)


lib/elixir/test/elixir/uri_test.exs
```diff
@@ -342,9 +342,18 @@ defmodule URITest do
     assert to_string(URI.parse("http://[2001:db8::]")) == "http://[2001:db8::]"
 
     assert URI.to_string(URI.parse("http://google.com")) == "http://google.com"
+    assert URI.to_string(URI.parse("gid:hello/123")) == "gid:hello/123"
 
     assert URI.to_string(URI.parse("//user:password@google.com/")) ==
              "//user:password@google.com/"
+
+    assert_raise ArgumentError,
+                 ~r":path in URI must be nil or an absolute path if host or authority are given",
```
eksperimental:
```suggestion
                 ~r":path in URI must be nil or an absolute path if :host or :authority are given",
```

lib/elixir/lib/uri.ex
```diff
@@ -626,6 +626,14 @@ defmodule URI do
 end
 
 defimpl String.Chars, for: URI do
+  def to_string(%{host: host, authority: authority, path: path} = uri)
+      when (host != nil or authority != nil) and is_binary(path) and
+             binary_part(path, 0, 1) != "/" do
+    raise ArgumentError,
+          ":path in URI must be nil or an absolute path if host or authority are given, " <>
```
eksperimental:
```suggestion
          ":path in URI must be nil or an absolute path if :host or :authority are given, " <>
```

lib/elixir/test/elixir/uri_test.exs
```diff
@@ -342,9 +342,18 @@ defmodule URITest do
...
+                 fn -> %URI{authority: "foo.com", path: "hello/123"} |> URI.to_string() end
+
+    assert_raise ArgumentError,
+                 ~r":path in URI must be nil or an absolute path if host or authority are given",
```
eksperimental:
```suggestion
                 ~r":path in URI must be nil or an absolute path if :host or :authority are given",
```

lib/elixir/lib/uri.ex
```diff
@@ -626,6 +626,14 @@ defmodule URI do
 end
 
 defimpl String.Chars, for: URI do
+  def to_string(%{host: host, authority: authority, path: path} = uri)
+      when (host != nil or authority != nil) and is_binary(path) and
+             binary_part(path, 0, 1) != "/" do
```
eksperimental: I think we need to check not only for nil, but for empty strings, since when we build the struct, we nillify the empty strings.

this is for another PR, which I can take care of, since I tried to tackle it and it needs a bit of revamping if we want to fully cover it. just let me know if you are interested,as I have one more question

josevalim:
The code works correctly for empty strings due to the semantics of guard
clauses. We could even remove the is_binary check.
-- 


eksperimental:
```iex
    assert URI.to_string(%URI{authority: nil, path: "hello/123"}) ==
             URI.to_string(%URI{authority: "", path: "hello/123"})
```
to_string behaves differently when authority is an empty string and  nil

lib/elixir/lib/uri.ex
```diff
@@ -628,7 +628,7 @@ end
 defimpl String.Chars, for: URI do
   def to_string(%{host: host, authority: authority, path: path} = uri)
       when (host != nil or authority != nil) and is_binary(path) and
-             binary_part(path, 0, 1) != "/" do
+             path != "" and binary_part(path, 0, 1) != "/" do
```
eksperimental: it is not only about path, but about a host, authority, and others. I have a working version, I can upload to my repo

eksperimental: Here's the code
https://github.com/eksperimental/elixir/commit/bff9b98aa164ce227ad2295aa10f2669a914b878

eksperimental: this way we avoid checking everywhere down the to_string impelmentation

josevalim: I don't think we should handle both for everything. For example, if you have an empty authority, we are going to output the leading `//`. I don't think we should generally treat empty the same as absent information, unless some specific cases like path.

josevalim: To clarify, the reason we nillify on parsing is exactly so we don't have to handle such cases downstream.

josevalim:

PR #9590, opened 2019-11-23
Use URI.parse/1 instead of building %URI{} manually whenever possible

eksperimental:
Discussion:
https://github.com/elixir-lang/elixir/issues/9588#issuecomment-557803196


eksperimental:
We are still building manually the %URI{} in the last two examples, but there's no other way I guess.

josevalim:

PR #9591, opened 2019-11-23
Add missing unreserved character to URI document

hirotnk:
According to RFC 3986, section 2.3, add missing unreserved character.
The implementation already includes it.


josevalim:

PR #9592, opened 2019-11-23
fix --no-halt option does not work  when Mix.Task.run("run", ["--no-halt"])  is called.

aJunKobayashi:
### ISSUE 

* --no-halt option does not work, when Mix.Task.run("run", ["--no-halt"])  is called.
* After 3673849e36f0182c6c7f410b0f59e6252da11942, --no-halt option does not work.

### Reproduce Procedure of the bug

* please create simple mix project
* please execute `$ iex -S mix`
* input `iex(1)> Mix.Task.run("run", ["--no-halt"])`

### Current Behavior

:ok is returned  and halt.

```
iex(1)> Mix.Task.run("run", ["--no-halt"])
:ok
```

### Expected Behavior

Mix.Task.run() should not return.

### DIFF
--- a/lib/mix/lib/mix/task.ex
+++ b/lib/mix/lib/mix/task.ex
@@ -287,18 +287,27 @@ defmodule Mix.Task do
     proj = Mix.Project.get()
     alias = Mix.Project.config()[:aliases][String.to_atom(task)]
 
-    cond do
-      alias && Mix.TasksServer.run({:alias, task, proj}) ->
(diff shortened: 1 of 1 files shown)


josevalim:
The reason Why it doesn’t work is because the task was already executed and tasks cannot be executed twice by default. You can fix by using Mix.Task.rerun. In any case, thanks for the PR!

aJunKobayashi:
I understand. Thank you.

PR #9593, opened 2019-11-24
Split Linux task in stable and development, and simplify them in Cirrus CI

eksperimental:


eksperimental:
This PR creates two grups, stable and developement,
It is clear in intention what each group does.

It removes the duplication of defining "allow_failures: true" and "skip_notifications: true"  for each developement task.
Tweaks the name of the task to include "OTP-" for `master` and `maint`

/cc @fertapric

fertapric:
Thanks @eksperimental! 💜

PR #9596, opened 2019-11-25
CI: Do not auto-cancel if branch is master or a release branch

eksperimental:


.cirrus.yml
```diff
@@ -15,6 +15,10 @@ test_linux_task:
   env:
     PATH: "${CIRRUS_WORKING_DIR}/otp/bin:${PATH}"
 
+
+  # don't cancel the task execution if it's master or a release branch
+  auto_cancellation: $CIRRUS_BRANCH != 'master' && $CIRRUS_BRANCH !=~ 'v\d+\.\d+.*'
```
josevalim: Is auto cancellation reported as a build failure? If not, I don't mind the auto-cancelling. It may make it a bit harder to detect when a build failed but at the same time, it saves resources.

eksperimental: By default, autocancellation is disabled for the main branch (master), what this PR does is trear the vX.X branches as master.

eksperimental: It is treated as aborted (RED CROSS), let me check if it's reported back to github as failure or as not run.

eksperimental: See the screenshot, they are marked with a red cross as not sucessful, inside the report they are marked with a grey "!" symbol, indicating they have been skipped.

image

fertapric:
Thanks @eksperimental! 💜 

That `!=~` was weird, but it seems to be supported by Cirrus 👍

eksperimental:
I also looked for it

PR #9598, opened 2019-11-25
remove a confusing arg for Code.eval_quoted/3 in Mix.Dep.Lock

irisTa56:
I think it is confusing and unnecessary to pass `opts` for `Code.string_to_quoted/2` to `Code.eval_quoted/3` as its second argument `binding`.
Removing the argument can save a bit of computation while keeping the intended functionality.

### DIFF
--- a/lib/mix/lib/mix/dep/lock.ex
+++ b/lib/mix/lib/mix/dep/lock.ex
@@ -17,7 +17,7 @@ defmodule Mix.Dep.Lock do
     with {:ok, contents} <- File.read(lockfile),
          assert_no_merge_conflicts_in_lockfile(lockfile, contents),
          {:ok, quoted} <- Code.string_to_quoted(contents, opts),
-         {%{} = lock, _binding} <- Code.eval_quoted(quoted, opts) do
+         {%{} = lock, _binding} <- Code.eval_quoted(quoted) do
       lock
     else
       _ -> %{}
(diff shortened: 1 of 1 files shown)


lib/mix/lib/mix/dep/lock.ex
```diff
@@ -17,7 +17,7 @@ defmodule Mix.Dep.Lock do
     with {:ok, contents} <- File.read(lockfile),
          assert_no_merge_conflicts_in_lockfile(lockfile, contents),
          {:ok, quoted} <- Code.string_to_quoted(contents, opts),
-         {%{} = lock, _binding} <- Code.eval_quoted(quoted, opts) do
+         {%{} = lock, _binding} <- Code.eval_quoted(quoted) do
```
josevalim: We need to pass the `file` to eval_quoted so it is used in error reports in the odd chance there is a syntax error in the lock file (for example, if someone changes it manually). We could have two separate `opts`, but sharing them is also fine as is. Thank you!

fertapric: Should it be then?

```suggestion
         {%{} = lock, _binding} <- Code.eval_quoted(quoted, [], opts) do
```

josevalim: Oh, I see, thank you! @irisTa56, can you please apply the suggestion above?

irisTa56: Thank you!
I made a new commit before noticing the suggestion, sorry.

josevalim:

PR #9600, opened 2019-11-26
Check reproducible builds in Cirrus CI under Windows

eksperimental:
I don't know why I left this one out.

Additionally sort all ENV variables alphabetically

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -117,10 +117,11 @@ test_windows_task:
   test_elixir_script: make --keep-going test_elixir
 
   check_reproducible_script:
-    CMD.exe /C ECHO OFF &
-    IF NOT DEFINED CHECK_REPRODUCIBLE (
-      ECHO The format reproducibility of the build is only checked in the last stable Erlang/OTP version.
-    ) ELSE (
-      make check_reproducible
-    )
+    - echo off
+    - cmd.exe /C echo off &
+      IF NOT DEFINED CHECK_REPRODUCIBLE (
+        echo "The format reproducibility of the build is only checked in the last stable Erlang/OTP version."
+      ) ELSE (
+        make check_reproducible
+      )
 
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -88,6 +88,7 @@ test_windows_task:
 
   matrix:
     - env:
+        CHECK_REPRODUCIBLE: true
         OTP_RELEASE: 22.0
         OS_VERSION: 2019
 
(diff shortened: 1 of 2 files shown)


fertapric:
It seems there is no `check_reproducible_script` under the `test_windows_task` :)

.cirrus.yml
```diff
@@ -114,3 +115,10 @@ test_windows_task:
   test_erlang_script: make --keep-going test_erlang
 
   test_elixir_script: make --keep-going test_elixir
+
+  check_reproducible_script: |
+    if [ -n "$CHECK_REPRODUCIBLE" ]; then
+      make check_reproducible
+    else
+      echo "The format reproducibility of the build is only checked in the last stable Erlang/OTP version."
```
fertapric:
```suggestion
      echo "The reproducibility of the build is only checked in the last stable Erlang/OTP version."
```

eksperimental: thank you. I'm fixing it in master as well.

eksperimental:
I'm  working on the Windows version, just bear as I need to setup a VM

eksperimental:
Hi @fertapric 
Are you using MINGW in the Windows docker image? What is the setup? I need to run the diff command and it is not availble. I know I can use `fc` but that will make things a bit harder, 
Is there a change to install diff in the image?

josevalim:
I am not sure if it is worth working on this? @fertapric is working on moving the diffing to elixir scripts so we get better error reports... which means that any work done here would be eventually discarded.

eksperimental:
I didn' t know that.
I thought that script would be called only if a mismatching is found. Anyways, this PR is almost done, I will upload it and make it will still be useful.

fertapric:
@josevalim is right. The majority of the files to check in `check_reproducible` are BEAM files, and those require a specific “BEAM-diff script” (the one I’m working on) to print human-readable information 🙂

josevalim:
I will go ahead and close this one then. If we have any extra logic, we can add it to the Elixir script too.

josevalim:
Thanks @eksperimental!

eksperimental:
that's ok.

btw @fertapric where can I have access to the windows docker image, so I can test locally.

eksperimental:
> I think they are a project under his GitHub name. :) -- *José Valimwww.plataformatec.com.br <http://www.plataformatec.com.br/>Founder and Director of R&D*

I am not able to find it. if it's there, it' s private.

fertapric:
@eksperimental I still need to document the process of building and publishing the images, but here is the Dockerfile:


```Dockerfile
FROM cirrusci/windowsservercore:2019

ARG OTP_RELEASE

RUN choco install -y erlang --version %OTP_RELEASE% \
      && choco install -y msys2 --params="'/InstallDir:C:\msys2'" \
      && choco install -y make \
      && setx path "C:\msys2\usr\bin;%path%"
```

You can also pull the image from Docker if you don’t need to build the image from scratch.

eksperimental:
@fertapric Awesomeness! I think that will be enough to get me started. thank you

PR #9602, opened 2019-11-27
Add Application.compile_env/3

josevalim:
Mix stores the compile_env in each applications' .app file.

Releases and Mix tasks were also changed to validate against
the runtime configuration against any stored `compile_env`.
In case of mismatches, an error is raised during boot.

Unit tests and docs have been added, only integration tests
are missing:

  * [ ] Track `compile_env` across recompilations
  * [ ] Integration tests for mix compile.app
  * [ ] Integration tests for mix app.start
  * [ ] Integration tests for release boot

/cc @sasa1977


josevalim:
For those interested in reviewing, the bulk of the changes are in "lib/application.ex", "lib/mix/release.ex", and "lib/mix/tasks/release.ex". The rest is mostly internals rearranging to pass the values around.

lib/mix/lib/mix/project_stack.ex
```diff
@@ -263,4 +270,15 @@ defmodule Mix.ProjectStack do
     {reply, state} = fun.(state)
     {:reply, reply, state}
   end
+
+  @impl true
+  def handle_info({:compile_env, {app, path, return}}, {stack, post_config}) do
+    with [%{config: config, compile_env: compile_env} = project | stack] <- stack,
+         ^app <- config[:app] do
```
michalmuskala: Is this intended that if you read compile env from another application it's not recorded?

josevalim: Yes, I am not sure how it could work across applications, since this is stored in the `compile_env` of each individual application. I will add a note to `compile_env` docs saying we can only track for the current application and it is extremely discouraged to read another application's environment.

lib/elixir/lib/application.ex
```diff
@@ -412,12 +439,154 @@ defmodule Application do
...
+    :elixir_config.update(:compile_env_listeners, &List.delete(&1, pid))
+    :ok
+  end
+
+  @doc """
+  Reads the application environment at compilation time.
+
+  Similar to `get_env/3`, except it must be used to read values
+  at compile time. This allows Elixir to track when configuration
+  values between compile time and runtime differ.
```
whatyouhide: Initially misread this sentence and was about to suggest a fix, then noticed it's correct but I think it reads clearer this way, feel free to ignore (only submitting this comment since you can just click "commit" if you like it 😄).

```suggestion
  values change between compile time and runtime.
```

lib/elixir/lib/application.ex
```diff
@@ -412,12 +439,154 @@ defmodule Application do
...
+
+  The first argument is the application name. The second argument
+  `key_or_path`, is either an atom key or a path to traverse in
```
whatyouhide:
```suggestion
  `key_or_path` is either an atom key or a path to traverse in
```

lib/elixir/lib/application.ex
```diff
@@ -412,12 +439,154 @@ defmodule Application do
...
+  For example, imagine the following configuration:
+
+      config :my_app, :key, [foo: [bar: :baz]]
+
+  We can access it during compile time as:
+
+      Application.compile_env(:my_app, :key)
+      #=> [foo: [bar: :baz]]
+
+      Application.compile_env(:my_app, [:key, :foo])
```
whatyouhide: This behaviour differs from `Application.get_env/3` and `Application.fetch_env(!)/2`, and I like it a lot. Will we backport it to those too since it's compatible?

lib/elixir/lib/application.ex
```diff
@@ -412,12 +439,154 @@ defmodule Application do
...
+
+      Application.compile_env(:my_app, [:key, :foo, :unknown], :default)
+      #=> :default
+
+  Giving a path is useful to let Elixir know that only certain paths
+  in a large configuration are compile time dependent.
+  """
+  # TODO: Warn if get_env/fetch_env/fetch_env! is used at compile time instead of compile_env
+  @doc since: "1.10.0"
+  @spec compile_env(app, key, value) :: value
```
whatyouhide:
```suggestion
  @spec compile_env(app, key | list, value) :: value
```

right?

lib/elixir/lib/application.ex
```diff
@@ -412,12 +439,154 @@ defmodule Application do
...
+  end
+
+  @doc """
+  Reads the application environment at compilation time or raises.
+
+  This is the same as `compile_env/3` but it raises an
+  ArgumentError if the configuration is not available.
+  """
+  @doc since: "1.10.0"
+  @spec compile_env!(app, key) :: value
```
whatyouhide: Same as above for the `key` spec here.

lib/elixir/lib/application.ex
```diff
@@ -412,12 +439,154 @@ defmodule Application do
...
+        value
+
+      :error ->
+        raise ArgumentError,
+              "could not fetch application environment #{inspect(key_or_path)} for application " <>
+                "#{inspect(app)} #{fetch_env_failed_reason(app, key_or_path)}"
+    end
+  end
+
+  defp fetch_compile_env(app, key) when is_atom(key),
```
whatyouhide: Shouldn't this be public to mirror fetch vs get for the normal env?

lib/elixir/lib/application.ex
josevalim: I am not planning to backport them for now. The rationale is that compile_env behaves differently, so it is ok for it to have "special powers". We will see.

lib/elixir/lib/application.ex
josevalim:
```suggestion
  @spec compile_env!(app, key | list) :: value
```

lib/elixir/lib/application.ex
josevalim: No plans for now to make it public. We will see. :)

lib/elixir/lib/application.ex
```diff
@@ -8,78 +8,100 @@ defmodule Application do
...
+      config :APP_NAME, :redis_host, "redis.local"
 
   The function `put_env/3` allows dynamic configuration of the application
   environment, but as a rule of thumb each application is responsible for its
   own environment. Please do not use the functions in this module for directly
   accessing or modifying the environment of other applications.
 
-  The application environment can be overridden via the `-config` option of
-  `erl`, as well as command-line options, as we are going to see below.
+  ### Compile-time environment
```
sasa1977: I found this paragraph a bit hazy. It took me a few passes + some skimming through the code to figure it out. I think it could be a bit more example driven, perhaps along the lines of:

1. Start by showing the sketch of `RedisClient` to demonstrate compile-time envs. 
2. Replace the current `config :APP_NAME ...` with the one that uses `System.get_env`.
3. Demonstrate how the code is not recompiled if the OS var is changed, and the app uses the old address to access redis.
4. Replace `Application.fetch_env!` with `Application.compile_env!`
5. Demonstrate how the system fails to boot when the OS var is changed.

lib/elixir/lib/application.ex
```diff
@@ -8,78 +8,100 @@ defmodule Application do
...
+        @host Application.fetch_env!(:APP_NAME, :redis_host)
+
+        def connect do
+          # access @host here
+        end
+      end
+
+  This approach, however, has one big limitation: if the Redis host is set
+  dynamically, for example through `System.get_env/2` or only when booting
+  the application, the code won't be recompiled. This can lead into situations
```
sasa1977: What does "or only when booting the application" mean?

lib/elixir/lib/application.ex
```diff
@@ -8,78 +8,100 @@ defmodule Application do
...
+  where the configuration the code was compiled with is different from the
+  code running, which can be cumbersome to debug.
+
+  Therefore, whenever reading the environment at compilation time, we
+  recommend using `Application.compile_env!/2` instead:
+
+      @host Application.compile_env!(:APP_NAME, :redis_host)
+
+  When `compile_env!/2` is used, Elixir is capable of notifying whenever
+  those values differ between compilation time and runtime.
```
sasa1977: This is where some expansion could be useful. How exactly is Elixir capable of notifying, i.e. what happens when those values differ?

lib/mix/lib/mix/project_stack.ex
michalmuskala: If that's the case, should it produce some sort of warning when you read compile env from another application?
Alternatively, we could only allow reading env from the application currently being compiled, so it would always be `Application.compile_env([:foo, :bar, :baz])`

josevalim: The problem is that we don't have the application being compiled in Elixir. Everything is tracked by Mix since we have historically tried to keep these concerns separated (hence the listener approach

However, we may need to do further changes here. Today I have no idea which file called `compile_env`, so if that file no longer calls `compile_env`, I am unable to stop tracking it. I will have to make it a macro (but explicitly requiring Application is going to be a detriment to its usage). We could automatically require Application, but we avoided doing that as well.

josevalim:
Thanks everyone for the review. I have applied all feedback above. Unfortunately there are some limitations in this approach and I need to do some changes throughout. So I will close this PR and open a new one soon.

jamesstonehill:
This feature ended up being implemented here: https://github.com/elixir-lang/elixir/pull/9628

PR #9607, opened 2019-11-27
Deprioritize Linux Development tasks in CI

eksperimental:
Add depends_on to Linux Development task in CI.
Initially shared by @fkorotkov
https://github.com/elixir-lang/elixir/pull/9585

Implemented now that Cirrus CI supports the `alias` field.
https://github.com/cirruslabs/cirrus-ci-docs/issues/527

Thank you @fkorotkov

Note: given the current setup, Linux Development doesn't
depend on Windows Stable, due do the latter taking way
longer than the rest. When new stable tasks are added,
they should depend on them, as in the case of the upcoming
FreeBSD task.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -20,6 +20,7 @@ test_linux_task:
 
   matrix:
     - name: Linux, ${OTP_RELEASE}, Ubuntu 14.04
+      alias: Linux Stable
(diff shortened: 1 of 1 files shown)


fertapric:
Thank you @eksperimental! 💜

PR #9608, opened 2019-11-28
Allow for match assertions in doctests

devonestes:
This is related to this discussion on the mailing list: https://groups.google.com/forum/m/#!msg/elixir-lang-core/06HZwKG8E8k/KeDneED9AwAJ

This allows devs to now use proper pattern matching assertions in doctests, complete with the nice colored diff in the event of a failure. Here are some screenshots of how these tests look now when they fail (and with color on in the ExUnit config).

diff1
diff2

This is currently missing documentation, but I'll tack that commit on once folks think this behavior looks good so I know what to document 😉

### DIFF
--- a/lib/ex_unit/lib/ex_unit/assertions.ex
+++ b/lib/ex_unit/lib/ex_unit/assertions.ex
@@ -104,8 +104,8 @@ defmodule ExUnit.Assertions do
   value. In such cases, simply use `Kernel.==/2` or
   `Kernel.match?/2`.
   """
-  defmacro assert({:=, _, [left, right]} = assertion) do
-    code = escape_quoted(:assert, assertion)
+  defmacro assert({:=, meta, [left, right]} = assertion) do
+    code = escape_quoted(:assert, meta, assertion)
 
     left = __expand_pattern__(left, __CALLER__)
     vars = collect_vars_from_pattern(left)
@@ -115,19 +115,25 @@ defmodule ExUnit.Assertions do
     # is not nil nor false. We need to rewrite the if
     # to avoid silly warnings though.
     return =
-      suppress_warning(
+      if meta[:skip_boolean_assert] do
         quote do
-          case right do
-            x when x in [nil, false] ->
-              raise ExUnit.AssertionError,
-                expr: expr,
-                message: "Expected truthy, got #{inspect(right)}"
-
-            _ ->
-              :ok
-          end
+          :ok
         end
-      )
+      else
+        suppress_warning(
+          quote do
+            case right do
+              x when x in [nil, false] ->
+                raise ExUnit.AssertionError,
+                  expr: expr,
+                  message: "Expected truthy, got #{inspect(right)}"
+
+              _ ->
+                :ok
+            end
+          end
+        )
+      end
 
     match_expr =
       suppress_warning(
@@ -159,7 +165,7 @@ defmodule ExUnit.Assertions do
   end
 
   defmacro assert({:match?, meta, [left, right]} = assertion) do
-    code = escape_quoted(:assert, assertion)
+    code = escape_quoted(:assert, meta, assertion)
     match? = {:match?, meta, [left, Macro.var(:right, __MODULE__)]}
 
     left = __expand_pattern__(left, __CALLER__)
@@ -190,7 +196,7 @@ defmodule ExUnit.Assertions do
         unless value do
           raise ExUnit.AssertionError,
             args: unquote(args),
-            expr: unquote(escape_quoted(:assert, assertion)),
+            expr: unquote(escape_quoted(:assert, [], assertion)),
             message: "Expected truthy, got #{inspect(value)}"
         end
 
@@ -221,7 +227,7 @@ defmodule ExUnit.Assertions do
 
   """
   defmacro refute({:match?, meta, [left, right]} = assertion) do
-    code = escape_quoted(:refute, assertion)
+    code = escape_quoted(:refute, meta, assertion)
     match? = {:match?, meta, [left, Macro.var(:right, __MODULE__)]}
 
     left = __expand_pattern__(left, __CALLER__)
@@ -254,7 +260,7 @@ defmodule ExUnit.Assertions do
         if value do
           raise ExUnit.AssertionError,
             args: unquote(args),
-            expr: unquote(escape_quoted(:refute, assertion)),
+            expr: unquote(escape_quoted(:refute, [], assertion)),
             message: "Expected false or nil, got #{inspect(value)}"
         end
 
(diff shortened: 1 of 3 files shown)


lib/ex_unit/lib/ex_unit/doc_test.ex
```diff
@@ -528,7 +556,11 @@ defmodule ExUnit.DocTest do
 
     case String.trim_leading(line) do
       "" ->
-        raise_incomplete_doctest(line_no, module)
+        [{previous_line, _} | _] = adjusted_lines
+
+        if not String.contains?(previous_line, " = ") do
```
josevalim: I think this check is too brittle. It will pass for == or a = inside a function call. But maybe making this check robust enough is quite complicated... So I want to try a couple things out. Can you please give me commit access to your fork in case I can figure something out? :)

devonestes: I've got the "allow edits from maintainers" thing on, so you should be able to push up commits to this.

I agree about this check. originally I was doing this check later on to check the AST of the expression, but it didn't fit very well with the error messages since at that point we didn't have the line number or module. I figured this was a decent starting point, but parsing the previous line to check the AST to make sure it's a match and not something else is definitely the safer way to go.

eksperimental: Also note that we don't necessarily need spaces around `=`, and that these tests can be broken in multiple lines

```elixir
iex(1)> tuple = {1, 2}
{1, 2}
iex(2)> {1, 2}=tuple
{1, 2}
iex(3)> {1, 2}=
...(3)> tuple
{1, 2}
```

devonestes:
Ok, I've pushed up an improvement and some improved tests. It still doesn't handle the multi-line match case, though, and frankly I'm not sure how we're going to be able to handle that at the moment, so this doctest would currently raise an error instead of matching properly:

```elixir
iex> {1, 2} =
iex> {1, 2}
```

I think to make that work reliably we might need to do this check at the point where we're parsing the doctest's AST and then taking the last expression in the block, but then we're potentially losing some information about module & line number, which was why I moved it back to this earlier part of the process.

devonestes:
Ok, I went the path of parsing the full AST for each doctest while we're checking validity, and now it can accurately handle pretty gnarly cases like this one: https://github.com/elixir-lang/elixir/pull/9608/files#diff-5bf76a905e4c4c5ef22f8dfcf379bdd1R436

I think this might be good to go now! Let me know if there are any other things y'all think might still be missing or any potential edge cases that this might not cover.

josevalim:
I pushed one commit! I did the following changes:

  1. We no longer require a value after `iex>`. The idea is that you can simply check if a code does not raise, for example, `iex> Mod.check_foo_bar!(...)` and this call raising or not is enough for as a test

  2. All `=` in a doctest are wrapped in assert. This changes the formatting a bit (we don't show all doctest, only the code) but it also means we get the diffing in all possible scenarios

Since the changes above were made to make the code more general, I don't think we need to change any docs. It should be good to go!

devonestes:
I think we still should document that it's now valid to not have a line after the doctest, right? Before this change that code was invalid, so this would be a new feature.

josevalim:
The documentation already erroneously said skipping the output was ok, so I
think we are good to go. But if we can further clarify, then that’s good
too.
-- 


eksperimental:
@devonestes  Do we have a multi-line test checking for the edge case I mentioned before. Either in this PR or previously in the doctests tests?

josevalim:
@eksperimental none of these edge cases are relevant in the current implementation.

fertapric:
> We no longer require a value after `iex>`. The idea is that you can simply check if a code does not raise, for example, `iex> Mod.check_foo_bar!(...)` and this call raising or not is enough as a test

Could it be changed to "if the last line uses pattern matching, then there is not need for a value after `iex>`". The given example could be written as `iex> _ = Mod.check_foo_bar!(...)` and we could keep warning users if they write doctests that are not testing anything.

Thoughts?

fertapric:
It would also make these two cases to behave equally:

```elixir
@doc ~S'''
   iex> 1 + 2
'''

@doc "iex> 1 + 2"
```

josevalim:
Addressed all of @fertapric concerns.

lib/ex_unit/test/ex_unit/doc_test_test.exs
```diff
@@ -417,6 +444,7 @@ defmodule ExUnit.DocTestTest do
   doctest ExUnit.DocTestTest.IndentationHeredocs
   doctest ExUnit.DocTestTest.FencedHeredocs
   doctest ExUnit.DocTestTest.Haiku
+  doctest ExUnit.DocTestTest.PatternMatching, only: [tuple: 0, map: 0], import: true
```
devonestes: These two functions that did have success cases in their doctests now don't exist, so we can probably remove this line since it does nothing. In fact, I'm surprised it doesn't raise an exception...

lib/ex_unit/test/ex_unit/doc_test_test.exs
```diff
@@ -546,84 +574,188 @@ defmodule ExUnit.DocTestTest do
...
+
+    assert output =~ """
+             8) doctest module ExUnit.DocTestTest.PatternMatching (8) (ExUnit.DocTestTest.PatternMatchingRunner)
+                test/ex_unit/doc_test_test.exs:#{doctest_line}
+                match (=) failed
+                code:  %{year: 2001, day: 1} = ~D"2000-01-01"
+                left:  %{year: 2001, day: 1}
+                right: ~D[2000-01-01]
+                stacktrace:
+                  (for doctest at) test/ex_unit/doc_test_test.exs:#{starting_line + 20}: (test)
```
devonestes: We don't have any tests ensuring that the two doctests that should pass did indeed pass. Should we add something here making sure we have 2 successful tests?

lib/ex_unit/lib/ex_unit/doc_test.ex
```diff
@@ -61,12 +61,16 @@ defmodule ExUnit.DocTest do
       2
 
   If you don't want to assert for every result in a doctest, you can omit
-  the result:
+  the result in the middle of expressions:
```
eksperimental:
```suggestion
  the result in between of expressions:
```
There's a subtle different but "in the middle" reads more like "within"

lib/ex_unit/lib/ex_unit/doc_test.ex
```diff
@@ -61,12 +61,16 @@ defmodule ExUnit.DocTest do
...
 
       iex> pid = spawn(fn -> :ok end)
       iex> is_pid(pid)
       true
 
+  As well as at the end:
+
+      iex> Mod.do_a_call_that_should_not_raise!(...)
```
eksperimental: I think if the result can be omitted in between expressions and in the last one, then they can be omitted altogether.

devonestes:
Looks great! 👍

lib/ex_unit/lib/ex_unit/assertions.ex
```diff
@@ -115,19 +115,25 @@ defmodule ExUnit.Assertions do
...
-              :ok
-          end
+          :ok
         end
-      )
+      else
+        suppress_warning(
+          quote do
+            case right do
+              x when x in [nil, false] ->
```
eksperimental:
```suggestion
              result when result in [nil, false] ->
```

josevalim:

Issue #9579, opened 2019-11-22
Randomly "make check_reproducible" fails


eksperimental:
```
Diffing...
Binary files lib/elixir/ebin/Elixir.System.beam and lib/elixir/tmp/ebin_reproducible/Elixir.System.beam differ
Makefile:130: recipe for target 'check_reproducible' failed
```

And I have seen the same error in Cirrus CI, for `Elixir.UndefinedFunctionError.beam`.
It doesn't happen very often, but it is not that uncommon either.

I investigated for a bit, to no avail.

fertapric:
@eksperimental That happened before and I had a poor man's script to perform a diff between `.beam` files (chunk diff). I'm improving that script, I'll submit a PR, and it'll run in every build.

eksperimental:
Awesome. I did an hex diff it meant nothing to me, as the differences were big enough.

eksperimental:
@fertapric it may help https://github.com/elixir-lang/elixir/issues/8689#issuecomment-457239695
https://github.com/bmwiedemann/reproducibleopensuse/blob/master/filterdiff

fertapric:
@eksperimental thanks for the comment!

I started with that one but the hex-diff was hard to read, that's why I implemented the chunk-diff.

eksperimental:
Awesome. Looking forward to it!

eksperimental:
Update: Under FreeBSD it found a diff in `Elixir.Task.beam`
https://cirrus-ci.com/task/5773269864546304

> Binary files lib/elixir/ebin/Elixir.Task.beam and lib/elixir/tmp/ebin_reproducible/Elixir.Task.beam differ

josevalim:
We are now equipped to debug this. So if it happens again, we will have more precise feedback.

PR #9609, opened 2019-11-28
Replace exsplus with exrop the OTP 20 default pseudorandom algorithm

joshprice:
Replace the deprecated `exsplus` pseudorandom algorithm with `exrop`,
which has been the default algorithm since OTP 20.

Although the `exsplus` still works it has been superceded and therefore
we should not be encouraging users to use it with examples in doctests.

It makes sense to use the new default everywhere rather than change doctests only.

`exsss` is the new default algorithm which was introduced in OTP 22 (see erlang/otp@4a08841). So once OTP 22 is the lowest Erlang version we should switch to it.

> Undocumented (old) algorithms are deprecated but still implemented
> so old code relying on them will produce the same pseudo random sequences as before.

See http://erlang.org/doc/man/rand.html for more detailed documentation.

This post further clarifies the timing of the changes:

https://groups.google.com/forum/#!topic/erlang-programming/Icw0pcIsnIY

### DIFF
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -2268,6 +2268,9 @@ defmodule Enum do
   the random value. Check its documentation for setting a
   different random algorithm or a different seed.
 
+  We use the `exrop` pseudorandom algorithm here since it's the default from OTP 20,
+  however if you are using OTP 22 or above then `exsss` is the default algorithm.
+
   ## Examples
 
       # Although not necessary, let's seed the random algorithm
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -1109,7 +1109,7 @@ defmodule Enum do
 
       iex> Enum.frequencies(~w{ant buffalo ant ant buffalo dingo})
       %{"ant" => 3, "buffalo" => 2, "dingo" => 1}
-      
+
   """
   @doc since: "1.10.0"
   @spec frequencies(t) :: map
@@ -1127,10 +1127,10 @@ defmodule Enum do
   as the count of every element.
 
   ## Examples
-    
+
       iex> Enum.frequencies_by(~w{aa aA bb cc}, &String.downcase/1)
       %{"aa" => 2, "bb" => 1, "cc" => 1}
-    
+
       iex> Enum.frequencies_by(~w{aaa aA bbb cc c}, &String.length/1)
       %{3 => 2, 2 => 2, 1 => 1}
 
@@ -1540,7 +1540,7 @@ defmodule Enum do
       iex> Enum.max([~D[2017-03-31], ~D[2017-04-01]])
       ~D[2017-03-31]
 
-  In the example above, `max/2` returned March 31st instead of April 1st 
+  In the example above, `max/2` returned March 31st instead of April 1st
   because the structural comparison compares the day before the year.
   For this reason, most structs provide a "compare" function, such as
   `Date.compare/2`, which receives two structs and returns `:lt` (less than),
@@ -1954,13 +1954,13 @@ defmodule Enum do
   ## Examples
 
       # Although not necessary, let's seed the random algorithm
-      iex> :rand.seed(:exsplus, {101, 102, 103})
+      iex> :rand.seed(:exsss, {101, 102, 103})
       iex> Enum.random([1, 2, 3])
-      1
+      3
       iex> Enum.random([1, 2, 3])
       3
       iex> Enum.random(1..1_000)
-      556
+      735
 
   """
   @spec random(t) :: element
@@ -2268,11 +2268,11 @@ defmodule Enum do
   ## Examples
 
       # Although not necessary, let's seed the random algorithm
-      iex> :rand.seed(:exsplus, {1, 2, 3})
+      iex> :rand.seed(:exsss, {1, 2, 3})
+      iex> Enum.shuffle([1, 2, 3])
+      [3, 2, 1]
       iex> Enum.shuffle([1, 2, 3])
       [2, 1, 3]
-      iex> Enum.shuffle([1, 2, 3])
-      [2, 3, 1]
 
   """
   @spec shuffle(t) :: list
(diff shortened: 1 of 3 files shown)


joshprice:
Looks like the exsss algorithm was introduced later than I had assumed in OTP 22 (see https://github.com/erlang/otp/commit/4a08841634b726fdd0c45bf8944a069813162794).

Looks like the best options in that case are either `exrop` the default prior to `exsss` or even `exsp` which directly replaced `exsplus`.

Thoughts/preferences?

joshprice:
Have pushed a new commit with using `exrop`, will update the title and description to match.

eksperimental:
I am wondering if in deprecation changes like this one, involving security, we should leave a note, mentioning that in previous versions of Elixir we used `exsplus` but it has been updated. I will help people reading the documentation to update their code. Additionally, we could also mention that if user is running Erlang/OTP 22+ , they can use `exsss`

joshprice:
@eksperimental Good idea. Would you put such notes next to doctests?

eksperimental:
> @eksperimental Good idea. Would you put such notes next to doctests?

I would place them in `random/1`, right after 
> This function uses Erlang's :rand module to calculate the random value. Check its documentation for setting a different random algorithm or a different seed.

joshprice:
@eksperimental updated with clarifying docs

eksperimental:
Has there been any security issue with the outdated algorithm? If so, I think we should explicitly  mention that in previous version of Elixir we were recommending to use that one, but now it's this new one.

joshprice:
@eksperimental no security issue that I'm aware of, only less than ideal randomness. The rand module docs state in the note section, and this is linked to from the Enum docs.

> There were a number of problems in the implementation of the now undocumented algorithms, which is why they are deprecated. The new algorithms are a bit slower but do not have these problems:

> Uniform integer ranges had a skew in the probability distribution that was not noticable for small ranges but for large ranges less than the generator's precision the probability to produce a low number could be twice the probability for a high.

> Uniform integer ranges larger than or equal to the generator's precision used a floating point fallback that only calculated with 52 bits which is smaller than the requested range and therefore were not all numbers in the requested range even possible to produce.

> Uniform floats had a non-uniform density so small values i.e less than 0.5 had got smaller intervals decreasing as the generated value approached 0.0 although still uniformly distributed for sufficiently large subranges. The new algorithms produces uniformly distributed floats on the form N * 2.0^(-53) hence equally spaced.

joshprice:
Looks like this error happened on one build https://github.com/elixir-lang/elixir/issues/9579 what's the process for rerunning?

eksperimental:
> Looks like this error happened on one build #9579 what's the process for rerunning?

go to https://cirrus-ci.com/task/4666053065441280
and hit re-run , or run failed tasks.
You need to be signed-in, you can do it with your GitHub account.

josevalim:
It is fine, it is a failure on our end, tracked in another issue. This is good to go!

eksperimental:
> @eksperimental no security issue that I'm aware of, only less than ideal randomness. The rand module docs state the following in the note section, and this is linked to from the Enum docs.
> 
> > There were a number of problems in the implementation of the now undocumented algorithms, which is why they are deprecated. The new algorithms are a bit slower but do not have these problems:
> 
> > Uniform integer ranges had a skew in the probability distribution that was not noticable for small ranges but for large ranges less than the generator's precision the probability to produce a low number could be twice the probability for a high.
> 
> > Uniform integer ranges larger than or equal to the generator's precision used a floating point fallback that only calculated with 52 bits which is smaller than the requested range and therefore were not all numbers in the requested range even possible to produce.
> 
> > Uniform floats had a non-uniform density so small values i.e less than 0.5 had got smaller intervals decreasing as the generated value approached 0.0 although still uniformly distributed for sufficiently large subranges. The new algorithms produces uniformly distributed floats on the form N * 2.0^(-53) hence equally spaced.

I would consider the outdated algorithms buggy then. It wil not be totally random, making certain values have more chances to be chosen than others.
Personally, I would leave a note, like the one in the Erlang docs that you quoted, saying that in previous versions of Elixir we recommend :exsplus, but do to not true randomness happening in certain ranges, we encourage everyone using it to replace it with : `:exrop`, `:exs1024s`, or `:exsp` and mention since which Erlang/OTP version these are supported.

lib/elixir/lib/enum.ex
```diff
@@ -2267,12 +2271,16 @@ defmodule Enum do
 
   ## Examples
 
+  The examples below use the `exrop` pseudorandom algorithm since it's
+  the default from OTP 20, however if you are using OTP 22 or above then
```
eksperimental:
```suggestion
  the default from Erlang/OTP 20, however if you are using Erlang/OTP 22 or above then
```

lib/elixir/lib/enum.ex
```diff
@@ -1953,14 +1953,18 @@ defmodule Enum do
...
```
eksperimental:
```suggestion
  the default from Erlang/OTP 20, however if you are using Erlang/OTP 22 or above then
```

lib/elixir/lib/enum.ex
```diff
@@ -1953,14 +1953,18 @@ defmodule Enum do
 
   ## Examples
 
+  The examples below use the `exrop` pseudorandom algorithm since it's
```
eksperimental:
```suggestion
  The examples below use the `:exrop` pseudorandom algorithm since it's
```

lib/elixir/lib/enum.ex
```diff
@@ -1953,14 +1953,18 @@ defmodule Enum do
...
+  the default from OTP 20, however if you are using OTP 22 or above then
+  `exsss` is the default algorithm.
```
eksperimental:
```suggestion
  `:exsss` is the default algorithm.
```

lib/elixir/lib/enum.ex
```diff
@@ -2267,12 +2271,16 @@ defmodule Enum do
 
   ## Examples
 
+  The examples below use the `exrop` pseudorandom algorithm since it's
```
eksperimental:
```suggestion
  The examples below use the `:exrop` pseudorandom algorithm since it's
```

lib/elixir/lib/enum.ex
```diff
@@ -2267,12 +2271,16 @@ defmodule Enum do
...
+  the default from OTP 20, however if you are using OTP 22 or above then
+  `exsss` is the default algorithm.
```
eksperimental:
```suggestion
  `:exsss` is the default algorithm.
```

joshprice:
@eksperimental I've pushed a commit with those suggestions included. Thanks!

whatyouhide:
Thank you @joshprice! 💟

PR #9610, opened 2019-11-28
WIP: Fix bug with formatting of maps in ExUnit output without color

devonestes:
Because maps are the only things that are displayed like literals but
aren't parsed as literals in the AST, they weren't being displayed
correctly in ExUnit diffs when they weren't being parsed for the
pattern diff (when users turn color off in ExUnit config).

I ran into this while working on #9608, and here are screenshots of
the before and after behavior. I'm betting this happened when the
pattern diffing was introduced and the `left` in an assertion error
changed from being a value to either a value or a pattern.

I'd love to be able to avoid `Code.eval_quoted/1`, but that seemed
like the best/only option here.

old_failure
after


devonestes:
Actually, I just realized this won't always work, so I need to take another swing at this...

PR #9614, opened 2019-11-30
Fix table format in Base module

eksperimental:
The tables were shown as code, also it exposes two bugs in IEx when formatting tables, which are fixed in subsequent PRs

It also shows the need to create a better style for displaying tables in ExDoc. I will submit a feature request later.


josevalim:
Good catch! I would like to fix both IEx and ExDoc first before going ahead with this change than. :)

eksperimental:
I will submit two different PRs then fixing this issue (i just need to write the tests). And the feature request to ExDoc. We mostly need graphic designer or someone with good taste (not me!) to choose colors for the alternating rows

eksperimental:
Last commit add clarity to the tables.

image

image

josevalim:

PR #9615, opened 2019-12-01
Fix bug when last two columns are empty in IO.ANSI.Docs

eksperimental:
When this happend, the last "|" because the content of the second to
last column.
This happened for example with the table in the Base module

        ## Base 64 (URL and filename safe) alphabet
        ...
        12 | M | 29 | d | 46 | u | 63    | _
        13 | N | 30 | e | 47 | v | |     |
        14 | O | 31 | f | 48 | w | (pad) | =
        15 | P | 32 | g | 49 | x | |     |
        16 | Q | 33 | h | 50 | y | |     |


josevalim:

PR #9617, opened 2019-12-01
Introduce "format" make target

eksperimental:
It runs the formatter from the current elixir code,
and not from the locally or globally installed version.


eksperimental:
Let's say user has v1.6 installed, running `mix format` as suggested by the README file won't work.

josevalim:

PR #9619, opened 2019-12-01
make Mix.Dep.diverged?/1 catch :divergedreq

irisTa56:
Hi!
Status for `:divergedreq` is a 3-tuple.
https://github.com/elixir-lang/elixir/blob/1faf6787092d2822a2f294394aab48c7ed0c3e00/lib/mix/lib/mix/dep/converger.ex#L257


josevalim:

PR #9620, opened 2019-12-01
Proposal: Allow to configure secondary IEx prompt

fertapric:
Allow further customization of the IEx prompt.

It also introduces a new placeholder `%w` that is replaced by an empty string (whitespaces) of the same length than the main prompt. This placeholder is useful to write prompts that are copy-paste-able. This is how the default prompt looks like: 

```elixir
iex(1)> try do
...(1)>   :ok
...(1)> rescue
...(1)>   _ -> :error
...(1)> end
```

And this is how it would look like with `%w`:

```elixir
iex(1)> IEx.configure(default_prompt2: "%w")
:ok
iex(2)> try do
          :ok
        rescue
          _ -> :error
        end
:ok
```

This idea was inspired by [PostgreSQL's `PROMPT2` ](https://www.postgresql.org/docs/12/app-psql.html#APP-PSQL-PROMPTING), and the `%w` placeholder by an [upcoming feature in PostgreSQL 13](https://git.postgresql.org/gitweb/?p=postgresql.git;a=commitdiff;h=7f338369ca624ca6c2e4f579623274c88d325bce).

I went with "prompt2" since this is just a draft, we should probably revisit that name 🙂

### DIFF
--- a/lib/iex/lib/iex.ex
+++ b/lib/iex/lib/iex.ex
@@ -338,7 +338,9 @@ defmodule IEx do
     * `:width`
     * `:history_size`
     * `:default_prompt`
+    * `:continuation_prompt`
     * `:alive_prompt`
+    * `:alive_continuation_prompt`
 
   They are discussed individually in the sections below.
 
@@ -420,7 +422,14 @@ defmodule IEx do
   The value is a keyword list with two possible keys representing prompt types:
 
     * `:default_prompt` - used when `Node.alive?/0` returns `false`
-    * `:alive_prompt`   - used when `Node.alive?/0` returns `true`
+
+    * `:continuation_prompt` - used when `Node.alive?/0` returns `false`
+      and more input is expected
+
+    * `:alive_prompt` - used when `Node.alive?/0` returns `true`
+
+    * `:alive_continuation_prompt` - used when `Node.alive?/0` returns
+      `true` and more input is expected
 
   The following values in the prompt string will be replaced appropriately:
 
--- a/lib/iex/lib/iex/config.ex
+++ b/lib/iex/lib/iex/config.ex
@@ -9,9 +9,9 @@ defmodule IEx.Config do
     :inspect,
     :history_size,
     :default_prompt,
-    :default_prompt2,
+    :continuation_prompt,
     :alive_prompt,
-    :alive_prompt2,
+    :alive_continuation_prompt,
     :width
   ]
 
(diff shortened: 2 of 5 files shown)


eksperimental:
maybe `default_prompt2` can be `alt_prompt`, and `alive_prompt2` can be renamed `alt_alive_prompt`

eksperimental:
LOVE IT!

eksperimental:
Fertapric, it was not clear what this PR proposed to me. could you start your example with how it behaves by default, to give a clearer context? thank you
```
iex(1)> try do
...(1)>   :ok
...(1)> rescue
...(1)>   _ -> :error
...(1)> end
```

josevalim:
@fertapric do you think it is easier to copy and paste? There is still the indentation, that has to be removed. And if you paste it into your editor, you can use column selector to trim it anyway.

Finally, if the goal is to use it for copy & pasting, would you always have this enabled? Or would you change it only when you need it? Isn't changing it at while actually worse? If the goal is copy & pasting, maybe we should have something like `last_expression()` that prints the last expression?

fertapric:
@josevalim there are two things to debate in this PR:

### 1. Allow customization of the secondary prompt

This would allow users to further customize IEx. 

Here is an example in which customizing the secondary prompt could be useful:

Screen Shot 2019-12-03 at 13 52 28

With the current config, this is as far as you can get:

Screen Shot 2019-12-03 at 13 53 33

This is the feature I value the most.

### 2. Support for the `%w` placeholder

As I mentioned above, the main use case for this is to make the IEx prompts more copy-paste-able.

> do you think it is easier to copy and paste? There is still the indentation, that has to be removed. And if you paste it into your editor, you can use column selector to trim it anyway.

Indentation is not always an issue in the console.

> Finally, if the goal is to use it for copy & pasting, would you always have this enabled? Or would you change it only when you need it? Isn't changing it at while actually worse? If the goal is copy & pasting, maybe we should have something like `last_expression()` that prints the last expression?

I would say both options are fine, enabled by default or on-demand. 

Said that, I agree with you that IEx might provide better ways to copy the last expression (or maybe multi-line edits?).

fertapric:
I think the best course of action here would be to leave 1 (if you are ok with it) and move 2 to the mailing list 🙂

josevalim:
Ok! Sounds good to me. Last call it “continuation_prompt”. WDYT?
-- 


fertapric:
@josevalim "continuation_prompt" sounds good, and there are previous uses of that terminology in other software to express that concept 👍 

This is ready to review 🙂

PR #9621, opened 2019-12-01
Do not shallow clone Git repo in CI

eksperimental:
CIRRUS_CLONE_DEPTH is set to 50 (default value in Travis),
to avoid failure when auto-cancellation is enabled,
and a task needs to clone the git repo but a new commit has been added
to that branch. It will fail.

See this for more information
https://github.com/cirruslabs/cirrus-ci-docs/issues/532

More info:
https://stackoverflow.com/a/53857834

You can see a live example here: https://cirrus-ci.com/github/eksperimental/elixir/v6.0
Where the fist commit fails after second one is added. Third and fourth is after this fix.

/cc @fertapric

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,5 +1,5 @@
 env:
-  CIRRUS_CLONE_DEPTH: 1
+  CIRRUS_CLONE_DEPTH: 50
   ELIXIR_ASSERT_TIMEOUT: 2000
   ELIXIRC_OPTS: "--warnings-as-errors"
   ERLC_OPTS: "+warnings_as_errors"


josevalim:
Can we disable passing this environment variable altogether then and just use the CI's defaults?

eksperimental:
> Can we disable passing this environment variable altogether then and just use the CI's defaults?

the default is `0`, and it will download the whole 16k commit history.
It takes 9seconds, against 1sec when depth is `50`
https://cirrus-ci.com/task/5982079765970944

eksperimental:
plus the additional time it will take to run `rm -rf .git` for all the git objects

eksperimental:
Cloning in windows goes from 18 secons (with a depth of `50`) to 31sec with depth of `0`

josevalim:

PR #9623, opened 2019-12-02
Print all undefined functions as warnings and then raise

josevalim:
Somtimes one undefined function warning can hide others.
For example, in a function call like this:

    foo(bar 1, 2)

The user may be expecting it to be called as:

    foo(bar(1), 2)

But Elixir compiles it to:

    foo(bar(1, 2))

In this case, the warning would be one of "undefined
function foo/1" or "undefined function bar/2", but they
are only "half truths". By showing all undefined function
warnings, we improve errors like above and it allows
users to see all undefined calls at once, when it would
otherwise require them to compile the code multiple times.

### DIFF
--- a/lib/elixir/src/elixir_locals.erl
+++ b/lib/elixir/src/elixir_locals.erl
@@ -105,9 +105,15 @@ warn_unused_local(File, Module, All, Private) ->
 
 ensure_no_undefined_local(File, Module, All) ->
   if_tracker(Module, [], fun(Tracker) ->
-    [elixir_errors:form_error(Meta, File, ?MODULE, {Error, Tuple})
-     || {Meta, Tuple, Error} <- ?tracker:collect_undefined_locals(Tracker, All)],
-    ok
+    case ?tracker:collect_undefined_locals(Tracker, All) of
+      [] -> ok;
+
+      List ->
+        [{FirstMeta, _, _} | _] = Sorted = lists:sort(List),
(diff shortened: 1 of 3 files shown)


lib/elixir/src/elixir_locals.erl
```diff
@@ -105,9 +105,15 @@ warn_unused_local(File, Module, All, Private) ->
...
 ensure_no_undefined_local(File, Module, All) ->
   if_tracker(Module, [], fun(Tracker) ->
-    [elixir_errors:form_error(Meta, File, ?MODULE, {Error, Tuple})
-     || {Meta, Tuple, Error} <- ?tracker:collect_undefined_locals(Tracker, All)],
-    ok
+    case ?tracker:collect_undefined_locals(Tracker, All) of
+      [] -> ok;
+
+      List ->
+        [{FirstMeta, _, _} | _] = Sorted = lists:sort(List),
```
fertapric: Why is the list sorted?

josevalim: Now that we show many, I am sorting by meta, which is the line at this point.

fertapric: 👌

There are no tests checking the order, should there be?

fertapric:
I'm just missing a test that checks that more than one warning is printed 🙂

josevalim:

PR #9626, opened 2019-12-03
Use fully qualified function calls in macros

### LINKED ISSUE #9624: require ExUnit.Assertion doesn't work properly
### Environment

* Elixir & Erlang/OTP versions:
  * Erlang/OTP 21 [erts-10.0.5] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [hipe]
  * Elixir 1.8.2 (compiled with Erlang/OTP 20): 

* Operating system: Linux under Windows (WSL)

### Current behavior

Currently, this code compiles correctly:

```elixir
defmodule BugReporting.AssertBug.Works do
  import ExUnit.Assertions

  def f(a, b), do: assert(a == b)
end
```

while this doesn't:

```elixir
defmodule BugReporting.AssertBug.DoesntWork do
  require ExUnit.Assertions

  def f(a, b), do: ExUnit.Assertions.assert(a == b)
end
```

Compilation fails with the following (quite surprising) message:

```plaintext
== Compilation error in file lib/bug_reporting/assert_bug/doesnt_work.ex ==
** (CompileError) lib/bug_reporting/assert_bug/doesnt_work.ex:4: undefined function assert/2
    (elixir) src/elixir_locals.erl:107: :elixir_locals."-ensure_no_undefined_local/3-lc$^0/1-0-"/2
    (elixir) src/elixir_locals.erl:107: anonymous fn/3 in :elixir_locals.ensure_no_undefined_local/3
```

### Expected behavior

Both example above should work

### What I think is wrong

My bet is that somewhere in the `ExUnit.Assertions` module, something like this appears:

```elixir
defmacro ... do
  quote do
    assert(value, message)
  end
end
```

When one wants to use a function in a quoted expression, one should use the fully qualified name, because the unqualified function will not be available. I'm not aware of how function scoping is handled inside quote blocks, and I don't think it's documented anywhere, but in my experience, the safest option is to always use fully qualified names.

The pattern above doesn't seem to appear in the code for `ExUnit.Assertions`, but the code does build some non-trivial AST fragments to be inserted in the calling module, so the problem should be equivalent to the above. I'll investigate what exactly the problem is, but I thought I should report it right now.

devonestes:
Currently you must import `ExUnit.Assertions` to use it, but now we'll
be able to use it just by requiring it.

Resolves #9624

### DIFF
--- a/lib/ex_unit/lib/ex_unit/assertions.ex
+++ b/lib/ex_unit/lib/ex_unit/assertions.ex
@@ -175,12 +175,13 @@ defmodule ExUnit.Assertions do
       right = unquote(right)
       left = unquote(Macro.escape(left))
 
-      assert unquote(match?),
+      ExUnit.Assertions.assert(unquote(match?),
         right: right,
         left: left,
         expr: unquote(code),
         message: "match (match?) failed" <> ExUnit.Assertions.__pins__(unquote(pins)),
         context: {:match, unquote(pins)}
+      )
     end
   end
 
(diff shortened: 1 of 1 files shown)


josevalim:

PR #9627, opened 2019-12-03
Close with correct accumulator in Stream.resource/3

hrubi:
In the optimized case with a single-element list returned from the
`next_fun`, call the stream function chain in a separate try/catch
block. If it errors out, call the `after_fun` with the new accumulator
returned from `next_fun`.

Previously the `after_fun` was called with the old accumulator in such
case.

### DIFF
--- a/lib/elixir/lib/stream.ex
+++ b/lib/elixir/lib/stream.ex
@@ -1417,22 +1417,20 @@ defmodule Stream do
 
   defp do_resource(next_acc, next_fun, {:cont, acc}, fun, after_fun) do
     try do
-      # Optimize the most common cases
-      case next_fun.(next_acc) do
-        {[], next_acc} -> {:opt, {:cont, acc}, next_acc}
-        {[v], next_acc} -> {:opt, fun.(v, acc), next_acc}
-        {_, _} = other -> other
-      end
+      next_fun.(next_acc)
     catch
       kind, reason ->
         after_fun.(next_acc)
         :erlang.raise(kind, reason, __STACKTRACE__)
     else
-      {:opt, acc, next_acc} ->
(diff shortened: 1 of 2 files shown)


lib/elixir/lib/stream.ex
```diff
@@ -1420,7 +1420,6 @@ defmodule Stream do
       # Optimize the most common cases
       case next_fun.(next_acc) do
         {[], next_acc} -> {:opt, {:cont, acc}, next_acc}
```
hrubi: I'll move this downwards to the `else` branch as well to simplify things a bit. Also I've noticed that there's probably no test for this case when empty list is returned, so I'll add one ..tomorrow.

hrubi: Done.

hrubi:
I'm wondering about the failing tests. The `Kernel.ImplTest` might be due to capturing stderr and running with `async: true`, however that was deliberately done in e1f347b as it is said to be fixed in fbe922c. Maybe there are still some problems with it?

I have no idea what's wrong with the other `DynamicSupervisorTest`.

josevalim:
Those are both race conditions unrelated to this PR! The `@impl` one makes sense, we can't do negative assertions, because we may capture the output of another test. So we need to make sure the name is unique or have those particular tests as sync. The supervisor one I thought I had fixed it, I will take another look.

josevalim:
Oh, I had fixed one of the dynamic supervisor tests but there were others. They should all be fixed now.

hrubi:
@josevalim Also how does it work with backporting bug fixes - should this go also to some stable versions?

josevalim:
In this case, we will backport it ourselves, but it is unlikely we will have new minor versions. The next release should be v1.10, happening in January.

josevalim:

PR #9628, opened 2019-12-04
Add Application.compile_env/3

josevalim:
Mix stores the compile_env in each applications' .app file.

Releases and Mix tasks were also changed to validate against
the runtime configuration against any stored `compile_env`.
In case of mismatches, an error is raised during boot.

Unit tests and docs have been added, only integration tests
are missing:

  * [x] Track `compile_env` across recompilations
  * [x] Integration tests for mix compile.app
  * [x] Integration tests for mix app.start
  * [x] Integration tests for release boot


lib/elixir/lib/application.ex
```diff
@@ -8,78 +8,122 @@ defmodule Application do
...
+        @db_host Application.get_env(:my_app, :db_host, "db.local")
+
+        def start_link() do
+          SomeLib.DBClient.start_link(host: @db_host)
+        end
+      end
+
+  This approach has one big limitation: if you change the value of the
+  application environment after the code is compiled, the value used at
+  runtime is not gouing to change! For example, if you are using `mix release`
```
maennchen:
```suggestion
  runtime is not going to change! For example, if you are using `mix release`
```

josevalim:
All pending tests added. This should be ready to go!

josevalim:

PR #9631, opened 2019-12-06
Adds more examples re setup callbacks

everettvody:


lib/ex_unit/lib/ex_unit/callbacks.ex
```diff
@@ -6,11 +6,13 @@ defmodule ExUnit.Callbacks do
   `setup_all/2` callbacks, as well as the `on_exit/2`, `start_supervised/2`
   and `stop_supervised/1` functions.
 
-  The setup callbacks are defined via macros and each one can
-  optionally receive a map with test state and metadata, usually
-  referred to as `context`. The context to be used in the tests can be
-  optionally extended by the setup callbacks by returning a properly
-  structured value (see below).
+  The setup callbacks are used to define implicit
```
fertapric: What do you mean here by "implicit"? 🙂

everettvody: Per the [linked wiki page](https://en.wikipedia.org/wiki/Test_fixture#Software), there are 3 types of test fixtures: in-line, delegated, and implicit: "Implicit setup places the test fixture in a setup method which is used to set up multiple test methods."  

I'm not married to including that adjective, but I believe that is the correct adjective to refer to the exact type of fixtures that `setup` and `setup_all` facilitate.

josevalim: I agree the adjective is correct but we should not assume the reader knows the definition before hand. So if we want to include it, then we should fully define it too, and if we can include counter examples, that’s even better. Thoughts?

everettvody: I'm happy to include more examples, e.g. maybe something like these?

```
test "demo of inline fixture" do
    user = %{username: "tester", email: "test@test.com"}
    MyContext.create_user(user)
    assert MyContext.user_exists?(user[:username]) 
end
```

```
defp get_user do
  %{username: "tester", email: "test@test.com"}
end

test "demo of delegated fixture" do
    user = get_user()
    MyContext.create_user(user)
    assert MyContext.user_exists?(user[:username]) 
end
```

josevalim:
Hi @everettvody! I did some changes to streamline the examples with more textual description. I have currently removed the implicit keyword, because we were using it early on. I guess we could add it back if we had a session which discussed testing strategy in general but we can leave it out of the scope of this PR for now. Thank you!

josevalim:

PR #9641, opened 2019-12-12
Update enum.ex docs for find_value

mpneuried:
Optimised `Enum.find_value` docs to show a case that better differentiates it from the alternatives.

### DIFF
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -995,12 +995,13 @@ defmodule Enum do
   @doc """
   Similar to `find/3`, but returns the value of the function
   invocation instead of the element itself.
-  An element is considered to be found when the result ist truthy (!= nil or != false).
+
+  An element is considered to be found when the result is truthy (neither `nil` nor `false`).
 
   ## Examples
 
       iex> Enum.find_value([2, 3, 4], fn x ->
-      ...>   if x > 2, do: x*x
+      ...>   if x > 2, do: x * x
       ...> end)
       9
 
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/enum.ex
```diff
@@ -995,9 +995,15 @@ defmodule Enum do
   @doc """
   Similar to `find/3`, but returns the value of the function
   invocation instead of the element itself.
+  An element is considered to be found when the result ist truthy (!= nil or != false).
 
   ## Examples
 
+      iex> Enum.find_value([2, 3, 4], fn x ->
+      ...>   if x > 2, do: x*x
```
whatyouhide:
```suggestion
      ...>   if x > 2, do: x * x
```

lib/elixir/lib/enum.ex
```diff
@@ -995,9 +995,15 @@ defmodule Enum do
   @doc """
   Similar to `find/3`, but returns the value of the function
   invocation instead of the element itself.
+  An element is considered to be found when the result ist truthy (!= nil or != false).
```
whatyouhide: There needs to be a newline between the first sentence and the rest of the docs since the first sentence is shown as a summary in the docs :)

```suggestion

  An element is considered to be found when the result is truthy (neither `nil` nor `false`).
```

mpneuried: 👍

lib/elixir/lib/enum.ex
mpneuried: 👍

josevalim:

PR #9643, opened 2019-12-13
Docs: Replace UTF-8 code point with Unicode code point

RumataEstor:
Unicode operates with "code points" and UTF-8 is an encoding which represents "code points" as bytes.


josevalim:

PR #9649, opened 2019-12-13
Use "code point" instead of "codepoint"

eksperimental:
That's the term it is used in the Unicode standard.
related commit 99d919e0c881c1d346dde9273c059360fdfd2f57


josevalim:

Issue #9653, opened 2019-12-16
Enum.slice/2 does not work with infinite streams


thth:
### Environment

```
Erlang/OTP 22 [erts-10.4] [64-bit] [smp:12:12] [ds:12:12:10] [async-threads:1]

Elixir 1.9.1 (compiled with Erlang/OTP 20)
```

### Current behavior

`Enum.slice/3` works, `Enum.slice/2` does not.
```elixir
[1, 2, 3] |> Stream.cycle() |> Enum.slice(0, 2)
# [1, 2]

[1, 2, 3] |> Stream.cycle() |> Enum.slice(0, 5)
# [1, 2, 3, 1, 2]

[1, 2, 3] |> Stream.cycle() |> Enum.slice(0..1)
# hangs forever

[1, 2, 3] |> Stream.cycle() |> Enum.slice(0..4)
# hangs forever
```

### Expected behavior
```elixir
[1, 2, 3] |> Stream.cycle() |> Enum.slice(0, 2)
# [1, 2]

[1, 2, 3] |> Stream.cycle() |> Enum.slice(0, 5)
# [1, 2, 3, 1, 2]

[1, 2, 3] |> Stream.cycle() |> Enum.slice(0..1)
# [1, 2]

[1, 2, 3] |> Stream.cycle() |> Enum.slice(0..4)
# [1, 2, 3, 1, 2]
```

PR #9654, opened 2019-12-17
add function Enum.slice for enumerable made by Stream.unfold

sanposhiho:
This fixes issue #9653 
This issue happened because of Enum.slice /2 count elements.

Stream.cycle() return Function in Stream.unfold/2, So, add function Enum.slice for enumerable made by Stream.unfold
I make function return ArgumentError when passed negative index_range.

### DIFF
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -2341,6 +2341,17 @@ defmodule Enum do
   @spec slice(t, Range.t()) :: list
   def slice(enumerable, index_range)
 
+  def slice(enumerable, first..last) when is_function(enumerable) do
+    amount = last - first + 1
+
+    if amount > 0 and first >= 0 and last >= 0 do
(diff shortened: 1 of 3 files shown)


michalmuskala:
This is only a partial solution - it ignores any infinite enumerables that use custom structs and protocol implementations. The fact that `Stream.unfold` uses a fun-based enumerable is only accidental.

I guess a proper solution would be to allow slice to work on enumerables without computing the entire length upfront - this means it wouldn't work with negative indices, but it's still better than current situation of not working at all.

sanposhiho:
Thanks for the review :)

I make Enum.slice/2 and Enum.slice/3 not to allow negative index.

This is a tough one because negative index meke  Enum.slice convenient.
Do you have any good ideas?（; _ ; ）

sanposhiho:
Enum.slice/3 is not directory associated to the issue, so return it to the original state

PR #9661, opened 2019-12-19
Genserver to track discarded counts of logs?

SophisticaSean:
I'm concerned there's no visibility into when logs are being discarded, would there be interest and support into adding a genserver that keeps track of discarded count and eventually logs that out?

It's scary to me that logs could be getting discarded without my knowledge currently.

I would be more than happy to implement this change.

### DIFF
--- a/lib/logger/lib/logger/handler.ex
+++ b/lib/logger/lib/logger/handler.ex
@@ -80,6 +80,7 @@ defmodule Logger.Handler do
   def log(%{level: erl_level, msg: msg, meta: metadata}, %{config: config}) do
     case threshold(config) do
       :discard ->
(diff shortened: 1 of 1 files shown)


josevalim:
If you look at the threshold implementation, you can see it already bumps a counter:

https://github.com/elixir-lang/elixir/blob/c592e7b04c11f5d273fbaabb70228d283d090e52/lib/logger/lib/logger/handler.ex#L157

Once we stop discarding, a process picks it up and logs:

https://github.com/elixir-lang/elixir/blob/c592e7b04c11f5d273fbaabb70228d283d090e52/lib/logger/lib/logger/config.ex#L106

PR #9662, opened 2019-12-19
"matched pattern" -> "match"

### LINKED ISSUE #9659: "matched pattern" misused in String module
The documentation for the String module misuses the phrase "matched pattern" (twice):

    ... a function that receives the matched patterns:

The function receives the result of the pattern matches, not the matched patterns.

    ... one can inject the whole matched pattern in the replacement string.

Again, one can replace the result of the pattern match, not the matched pattern.

evantravers:
We seem to use "pattern" to describe the thing that is doing the
matching, and "match" to describe the _found_ object. This is intended
to make this docs more consistent with the language found in the `Regex`
module.

Fixes #9659

### DIFF
--- a/lib/elixir/lib/string.ex
+++ b/lib/elixir/lib/string.ex
@@ -1344,7 +1344,7 @@ defmodule String do
       "a-b,c"
 
   The pattern may also be a list of strings and the replacement may also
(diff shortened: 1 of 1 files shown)


josevalim:

PR #9663, opened 2019-12-19
Improve error message on import/2

wojtekmach:
Before:

    iex(1)> a = [flatten: 1]; import List, only: a
    ** (CaseClauseError) no case clause matching: {:only, {:a, [version: 0, line: 1], nil}}
        (elixir 1.10.0-dev) src/elixir_import.erl:10: :elixir_import.import/4

After:

    iex(1)> a = [flatten: 1]; import List, only: a
    ** (CompileError) iex:1: invalid :only option for import, expected value to be an atom :functions, :macros, or a list literal, got: a
        (elixir 1.10.0-dev) src/elixir_import.erl:22: :elixir_import.import/4


josevalim:

PR #9670, opened 2019-12-24
Fix Macro.to_string for sigils

devonestes:
Some sigils are commonly used with certain delimiters first, but in the
current implementation of Macro.to_string/1 it defaults all sigils to
using `"` as a delimiter. This changes things so that certain sigils are
shown with their "preferred" delimiters first.

Ideally I'd love if the delimiters for the given sigil were kept in the
AST metadata when compiled so we can always show the exact code that a
user input in places like ExUnit. At the moment when we show the `code`
block in ExUnit errors, if a sigil is used there it's not guaranteed to
show the actual code that the user input, but we can't do this at the
moment since we don't have the delimiter data in the AST. At least with
this change we're more likely to show users the correct delimiters.

### DIFF
--- a/lib/elixir/lib/macro.ex
+++ b/lib/elixir/lib/macro.ex
@@ -1007,7 +1007,11 @@ defmodule Macro do
     false
   end
 
-  defp interpolate({:<<>>, _, parts}, fun) do
+  defp interpolate(ast, fun) do
+    interpolate(ast, {?", ?"}, fun)
+  end
+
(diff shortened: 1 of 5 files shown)


josevalim:
@devonestes good call. I propose then that we always add the sigil information in the metadata. The information is already available under a certain flag, so we can just change it so it is always available and leverage it on `Macro.to_string/1` (like we do for the formatter). Can you please look into it?

devonestes:
Closing in favor of #9671

PR #9674, opened 2019-12-30
ExUnit: produce nice assertion errors with the pipe operator

AlphaHydrae:
With the following code:

```elixir
value = %{n: 2}
assert %{n: 1} |> Map.equal?(value)
```

Instead of just indicating that it expected truthy and got false, the following helpful error is produced:

    Expected Map.equal?/2 to return truthy, got false
    code: assert %{n: 1} |> Map.equal?(value)
    arguments:

       # 1
       %{n: 1}

       # 2
       %{n: 2}

I tried to make this work because I thought it was a shame that `assert fun(arg1, arg2)` produces a nice assertion error, but not `assert arg1 |> fun(arg2)` (the values of the arguments are not shown in the error).

I was writing functions to compare a date from an API to an expected date, and wanted to be able to write assertions in a functional style like this, without an additional library:

```elixir
assert created_at |> just_after?(now, within: 500, unit: :millisecond)
```

I also made the error message explicitly indicate the function's name and arity. You could argue it is redundant since the code of the assertion is shown, but I thought it could be helpful.

All that said, I am a total Elixir beginner and I wrote this by trial and error, with little to no understanding of what I just did with macros. I have no idea whether this will work in all cases or whether it's a bad idea because "you just shouldn't do that". Please do not hesitate to tell me why it may be a bad idea or how to improve it.

### DIFF
--- a/lib/ex_unit/lib/ex_unit/assertions.ex
+++ b/lib/ex_unit/lib/ex_unit/assertions.ex
@@ -149,6 +149,26 @@ defmodule ExUnit.Assertions do
     end
   end
 
+  defmacro assert({:|>, meta, [left, {fun, context, [_ | _] = args} = right]} = assertion) do
+    left = __expand_pattern__(left, __CALLER__)
+    expr = escape_quoted(:assert, meta, assertion)
+    fun_name = Macro.to_string({fun, context, []}) |> String.replace_trailing("()", "")
+    fun_arity = length(args) + 1
+
+    quote do
+      match? = unquote(left) |> unquote(right)
+
+      ExUnit.Assertions.assert(match?,
+        args: unquote([left | args]),
+        expr: unquote(expr),
+        message:
+          "Expected #{unquote(fun_name)}/#{unquote(fun_arity)} to return truthy, got #{
+            inspect(match?)
+          }"
+      )
+    end
+  end
+
   defmacro assert(assertion) do
     if translated = translate_assertion(:assert, assertion, __CALLER__) do
       translated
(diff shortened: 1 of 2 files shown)


josevalim:
Hi @AlphaHydrae, welcome!

I think this is a good idea! My main concern is only what happens if you have multiple pipelines. In such cases, trying to understand what are the arguments may actually be confusing?

AlphaHydrae:
Thanks!

The third test I added gives an example of that: `assert %{} |> Map.put(:foo, 1) |> Map.equal?(value)`. The assertion error shows the arguments to the last `Map.equal?` call just like the other test. It might get confusing for complex pipelines, although my gut feeling is that if you're going to debug a test, you're still better off having the information about what was actually provided to the last function than just having the result of the expression.

Maybe a better error message for a longer pipeline could help? Something to the effect of: "Expected the Map.equal?/2 function called after the last pipe to return truthy..." Provided it's even possible to differentiate between 1 or 2+ pipes with pattern matching (or any other way). If it is, I guess an alternative would be to disable this formatting if there is more than one pipe.

josevalim:
If you have more than one pipe, then you would have to unfold all of the pipes and get the last one. But that introduces another issue in that someone may have replaced the pipe itself by their custom pipe (or a pipe with special powers) and, if we rewrite it, we will change its behaviour.

Given that most people only use pipelines when they have at least two pipes (otherwise using `|>` would just make things more verbose) and the nuances in handling multiple pipes, it is probably best to keep this code as is for now.

In any case, thank you for the PR and considerations! Have a Happy New Year!

AlphaHydrae:
It seemed so harmless at the time, too bad. Thanks for taking the time to review, and happy new year as well!

PR #9677, opened 2019-12-30
Set runtime config as false when not rebooting

ggcampinho:
This allows the release skip the creation of unnecessary runtime files during the execution of the clis


josevalim:

PR #9684, opened 2020-01-03
IO.ANSI.Docs - Properly render quote blocks

sascha-wolf:
## Issue

This PR introduces markdown quote block handling in `IO.ANSI.Docs`, for blocks such as this:

```markdown
> first line of quote
> second line of __quote__
> third line of `quote`
```

Before this a quote was simply rendered as text which lead to the above quote looking like this:

\> first line of quote > second line of __quote__ > third line of `quote`

An example of this in the Elixir codebase can [be seen in `Access` (link to code)](https://github.com/elixir-lang/elixir/blob/055526057ac9992cc1df87441357c4e56d45235d/lib/elixir/lib/access.ex#L46-L55). Which got rendered like this in `iex`:

image

## Solution

Instead of defining a specific quote style (such as `doc_quote`) - which clashes with inline styling of text - this solution instead renders the quote's content in an ASCII-art box. As such the quote from above would be rendered like this:

```
+----------------------+
| first line of quote  |
| second line of quote |
| third line of quote  |
+----------------------+
```

With this change the quote from [`Access`](https://github.com/elixir-lang/elixir/blob/055526057ac9992cc1df87441357c4e56d45235d/lib/elixir/lib/access.ex#L46-L55) now gets rendered like this:

image

### DIFF
--- a/lib/elixir/test/elixir/io/ansi/docs_test.exs
+++ b/lib/elixir/test/elixir/io/ansi/docs_test.exs
@@ -57,8 +57,15 @@ defmodule IO.ANSI.DocsTest do
     assert result == "\e[33m### wibble\e[0m\n\e[0m\ntext\n\e[0m"
   end
 
-  test "single-line quote block is converted" do
-    result = format("line\n\n> normal *italics* `code`\n\nline2\n")
+  test "short single-line quote block is converted into single-line quote" do
+    result =
+      format(
+        "line\n" <>
+          "\n" <>
+          "> normal *italics* `code`\n" <>
+          "\n" <>
+          "line2\n"
+      )
 
     assert result ==
              "line\n" <>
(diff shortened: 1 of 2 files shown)


josevalim:
Thank you @sascha-wolf!

Generally speaking, we try to render IO.ANSI as close to the original markup as possible, so I think we can render it as:

```
> first line of quote
> second line of quote
> third line of quote
```

Can you please adjust the code accordingly? Thank you.

sascha-wolf:
Sure thing!

sascha-wolf:
I've updated the code to render quotes with a leading `> `; to differentiate them from regular text the `> ` is being styled according to the `doc_quote` setting which - for now - is set to `light_black`.

The end result looks like this:

image

@josevalim the quote retains the lines as originally written in markdown. It would also be possible to wrap the lines using the same logic as `write_text` - when I look at various markdown there seems to be no "standard" (e.g. GitHub retains line breaks, StackOverflow removes them).

What do you think?

josevalim:
Let’s use write_text because the lines originally written may be too short or too long!

sascha-wolf:
@josevalim I've updated the code to wrap the lines for quotes. The result now looks like this for `Access`:

image

sascha-wolf:
I'll also extend the rendering to respect empty lines for quotes.

lib/elixir/test/elixir/io/ansi/docs_test.exs
```diff
@@ -57,6 +57,69 @@ defmodule IO.ANSI.DocsTest do
     assert result == "\e[33m### wibble\e[0m\n\e[0m\ntext\n\e[0m"
   end
 
+  test "short single-line quote block is converted into single-line quote" do
```
josevalim: Just one last point: maybe the tests would be clearer with heredocs?

sascha-wolf: Done in 497c29c

sascha-wolf:
@josevalim from my perspective everything is ready to go. 🙂

ericmj:
Thank you @sascha-wolf!

PR #9688, opened 2020-01-03
Rework docs to allow calling `import_config/1` in releases

fxn:
This is actually a question in form of documentation patch 😀.

I was surprised that the docs of the `release` task forbid usage of `import_config/1` with a "MUST NOT" (it said `import_file`, but I guess it means `import_config`).

The `Config` module is unrelated to Mix and looking at the source code of `import_config/1` I see no technical reason it cannot run in releases. So my guess is that the intention of the docs was to rule out common Mix config use cases for `import_config/1` —probably written having the transition from `Mix.Config` in mind—, but perhaps the wording ended up being unnecessarily restrictive ruling out _any_ kind of invocation.

Then, I connected that suspicion with a need to avoid duplication between `config/config.exs` and `config/releases.exs`.

In my project everything that can be configured using env vars is configured that way for all environments and releases. For example,

```elixir
config :avrora, registry_url: System.fetch_env!("AVRO_REGISTRY_URL")
```

is the same in both files and so are a dozen others. Duplicating this is not only extra work, but also error-prone since you need to remember to keep both files in sync.

So, one solution that occurred to me this afternoon is to ship `priv/common_config.exs` and import it as the documentation patch explains.

Hot or not? :)

### DIFF
--- a/lib/mix/lib/mix/tasks/release.ex
+++ b/lib/mix/lib/mix/tasks/release.ex
@@ -579,7 +579,7 @@ defmodule Mix.Tasks.Release do
   Your `config/releases.exs` file needs to follow three important rules:
 
     * It MUST `import Config` at the top instead of the deprecated `use Mix.Config`
-    * It MUST NOT import any other configuration file via `import_file`
+    * It MUST NOT import any other configuration file via `import_config`
     * It MUST NOT access `Mix` in any way, as `Mix` is a build tool and it not
       available inside releases
 


josevalim:
Correct, it should have been `import_config` instead of `import_file`.

The reason why `config/releases.exs` cannot import other files is because `mix release` will only copy `config/releases.exs` inside the release. So if you reference any other file, those would have to be absolute paths, which are generally bad practice.

Note though you can import "config/releases.exs" from other configuration files - which is what I generally do and we were planning to do in Phoenix too, but Phoenix also has to support old Elixir versions.

fxn:
> The reason why config/releases.exs cannot import other files is because mix release will only copy config/releases.exs inside the release. So if you reference any other file, those would have to be absolute paths, which are generally bad practice.

However, the proposed solution using `:code.priv_dir/1` seems portable and robust? Why is that a bad approach? Would be different if you tried to build absolute paths by hand, but Erlang encapsulates that for you and `priv` is a 1st class directory in releases, so you can leverage it.

> Note though you can import "config/releases.exs" from other configuration files - which is what I generally do and we were planning to do in Phoenix too, but Phoenix also has to support old Elixir versions.

`config/releases.exs` is a configuration file for production, in the general case, some of it won't be suitable for `config/config.exs`, no?

josevalim:
I am not sure you can generally assume that applications will be loaded by
the time the config runs for you to call “code.priv_dir”. It works right
now but it is not a guarantee, so I wouldn’t generally recommend it (i.e.
use it but beware!).

And correct. We were planning to call “config/releases.exs” from
“config/prod.exs”.

So maybe we can keep the wording from today of “must not” but clarify why,
for those willing to break the rules. :)

-- 


fxn:
> And correct. We were planning to call “config/releases.exs” from “config/prod.exs”.

But then, we are not avoiding the duplicated configuration code that is common to all environments and releases.

Let my underscore that the proposal using `priv` is technical, it is one thing that is present in releases today. It is not an elegant solution per se, but I don't see a way to include a `common.exs` in the release.

josevalim:
Just to make sure, putting them in config/config.exs is not enough because
they are not available at the time the release is assembled?
-- 


fxn:
Let me explain the use case with code.

My dev and test environments run on Docker Compose, so `docker-compose.yml` has all the environment variables defined easily without messing with your computer. With that, everything in the project is uniformely configured via system env vars no matter if it is running with Mix or as a release.

One example is

```elixir
config :avrora, registry_url: System.fetch_env!("AVRO_REGISTRY_URL")
```

The environment variable `AVRO_REGISTRY_URL` is required always, both Mix and releases need to fetch it.

OK, the problem I am trying to address is that that means you have to duplicate code, because you need that line in `config/config.exs` **and also** that same line in `config/releases.exs`. Multiply that by another dozen similar configuration settings, you need to maintain two files with a good chunk of code that is equals and has to be in sync.

With my trick, I can factor that out to `priv/common_config.exs`:

```elixir
# priv/common_config.exs
config :avrora, registry_url: System.fetch_env!("AVRO_REGISTRY_URL")
```

and _reuse_ that file in both places:

```elixir
# config/config.exs
import_config "../priv/common_config.exs"
```

and also

```elixir
# config/releases.exs
import_config Path.join(:code.priv_dir(:my_app), "common_config.exs")
```

Admitedly, not elegant, but I am solving this using the existing constraints. If there was a way to say "include this file in the release in the same please `release.exs` is", I would put that common code in a more natural place.

josevalim:
Thanks @fxn! Given the context my above, my question was the following: all configuration in config/config.exs will also apply to config/releases.exs. The only reason to have it config/releases.exs is because these configs would then only be available when the release boots and now when it is assembled. But when this is the case, you don't want to put it "config/config.exs", but rather on each environment. Does it make sense?

fxn:
When Mix runs, it creates an application resource file, and the application resource file hard-codes whatever application configuration is derived from `config/config.exs`. In my dev and test environments those ones are available at compile time.

In particular, it does so when `mix release` runs, but at that time the real values for production are not needed. When the release runs, at runtime, the real values for production are known.

See what I mean?

josevalim:
Right, that's exactly it! My point is: if you import this shared configuration file in `config/config.exs` and those values are missing, then you won't be able to assemble the release, as `config/config.exs` is also read at release time. So you either already have these values when you assemble the release, and therefore you don't need this feature, or you must not import this file in config/config.exs and rather on each environment, except prod.exs, which means you need this feature even more. :D

In any case, it is too early for us to "endorse" this pattern, so I would like to wait a bit more before removing these restrictions. If by 1.12 the boot process is still the same, then I think these assumptions are effectively true and we can ship it. :)

josevalim:

fxn:
Ah! I see what you mean.

There are two options here: One is to provide dummy values when assembling the release. This is already the case in `dev` and `test`, for example AWS credentials are fake. Another one is to import the common chunk in `dev.exs`, `test.exs`, and `releases.exs` but not in `config.exs`, so those common `fetch_env!` do not run with `MIX_ENV=prod mix release`.

Talking with someone in IRC about this yesterday, they shared the feeling that there is quite a bit of duplication in practice. I believe there is these use cases something to think about for the future.

I have revised the patch to only do `import_file` → `import_config`, thanks for the discussion José ❤️.

PR #9694, opened 2020-01-06
Document compilation tracers in Code.Tracer

wojtekmach:
I think the way it's laid out currently is totally fine but I'd like to
open up discussion if moving these docs to a separate module as well as
providing a behaviour would help discoverability and make code using it
more explicit. If not now, maybe something to keep in mind for the
future :)

### DIFF
--- a/lib/elixir/docs.exs
+++ b/lib/elixir/docs.exs
@@ -86,6 +86,7 @@
     ],
     "Code & Macros": [
       Code,
+      Code.Tracer,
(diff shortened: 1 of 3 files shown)


josevalim:
I would like to avoid giving it too much attention for now. We are not aware of the impact or adoption they will have. I am holding a similar stance to the `compare` function, which deserves a `Comparable` module. :) In any case, thanks!

PR #9699, opened 2020-01-08
Clarify which home directory is referred to by `~` in `~/.iex.exs`.

hassan:
I'm accustomed to the use of `~` to represent the home directory of the person executing a command in a shell. Since that's apparently not the case when connecting to a node via `remote_console` it seems worth clarifying.


josevalim:

Issue #9702, opened 2020-01-08
File.CopyError during mix release


amatalai:
### Environment

* Elixir & Erlang/OTP versions (elixir --version): 
```
Erlang/OTP 22 [erts-10.5.6] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1]
Elixir 1.10.0-rc.0 (105770c) (compiled with Erlang/OTP 22)
```
* Operating system: alpine (`erlang:22.1.8-alpine` used as base docker image)

### Current behavior
`MIX_ENV=prod mix release`
```
* assembling myapp-0.1.0 on MIX_ENV=prod
* using config/releases.exs to configure the release at runtime
* creating _build/prod/rel/myapp/releases/0.1.0/vm.args
* creating _build/prod/rel/myapp/releases/0.1.0/env.sh

11:48:31.589 [error] Task #PID<0.92.0> started from #PID<0.8443.0> terminating
** (File.CopyError) could not copy recursively from "/usr/local/lib/erlang/erts-10.5.6/src" to "/app/_build/prod/rel/myapp/erts-10.5.6/src". /usr/local/lib/erlang/erts-10.5.6/src: no such file or directory
    (elixir 1.10.0-rc.0) lib/file.ex:898: File.cp_r!/3
    (mix 1.10.0-rc.0) lib/mix/release.ex:680: anonymous fn/4 in Mix.Release.copy_erts/1
    (elixir 1.10.0-rc.0) lib/enum.ex:2102: Enum."-reduce/3-lists^foldl/2-0-"/3
    (mix 1.10.0-rc.0) lib/mix/release.ex:677: Mix.Release.copy_erts/1
    (mix 1.10.0-rc.0) lib/mix/tasks/release.ex:1273: Mix.Tasks.Release.copy/2
    (elixir 1.10.0-rc.0) lib/task/supervised.ex:90: Task.Supervised.invoke_mfa/2
    (elixir 1.10.0-rc.0) lib/task/supervised.ex:35: Task.Supervised.reply/5
    (stdlib 3.10) proc_lib.erl:249: :proc_lib.init_p_do_apply/3
Function: &:erlang.apply/2
    Args: [#Function<0.53545855/1 in Mix.Tasks.Release.assemble/1>, [:erts]]
```

### Expected behavior

There was no error for `1.9.4-otp-22`

josevalim:
[...]

PR #9703, opened 2020-01-08
Fix mix.release raising error when src is missing in erts_source

amatalai:
Fixes issue described in #9702 

I have no idea how to write a test for this case, any guidance?


josevalim:
I have poked at it and I could not find a obvious way to test, so I will look into it next. Thanks for the PR!

PR #9712, opened 2020-01-10
Track column information in EEx if enabled

wojtekmach:
Additionally we add :column to Code.string_to_quoted*/2

### DIFF
--- a/lib/eex/lib/eex/compiler.ex
+++ b/lib/eex/lib/eex/compiler.ex
@@ -13,16 +13,19 @@ defmodule EEx.Compiler do
   def compile(source, opts) when is_binary(source) and is_list(opts) do
     file = opts[:file] || "nofile"
     line = opts[:line] || 1
+    column = 1
     trim = opts[:trim] || false
 
-    case EEx.Tokenizer.tokenize(source, line, trim: trim) do
+    case EEx.Tokenizer.tokenize(source, line, column, trim: trim) do
       {:ok, tokens} ->
         state = %{
           engine: opts[:engine] || @default_engine,
           file: file,
           line: line,
           quoted: [],
-          start_line: nil
+          start_line: nil,
+          start_column: nil,
+          parser_options: Code.get_compiler_option(:parser_options)
         }
 
         init = state.engine.init(opts)
@@ -41,13 +44,19 @@ defmodule EEx.Compiler do
     generate_buffer(rest, buffer, scope, state)
   end
 
-  defp generate_buffer([{:expr, line, mark, chars, _} | rest], buffer, scope, state) do
-    expr = Code.string_to_quoted!(chars, line: line, file: state.file)
+  defp generate_buffer([{:expr, line, column, mark, chars, _} | rest], buffer, scope, state) do
+    options = [file: state.file, line: line, column: column(column, mark)] ++ state.parser_options
+    expr = Code.string_to_quoted!(chars, options)
     buffer = state.engine.handle_expr(buffer, IO.chardata_to_string(mark), expr)
     generate_buffer(rest, buffer, scope, state)
   end
 
-  defp generate_buffer([{:start_expr, start_line, mark, chars, _} | rest], buffer, scope, state) do
+  defp generate_buffer(
+         [{:start_expr, start_line, start_column, mark, chars, _} | rest],
+         buffer,
+         scope,
+         state
+       ) do
     {contents, line, rest} = look_ahead_middle(rest, start_line, chars)
 
     {contents, rest} =
@@ -55,7 +64,13 @@ defmodule EEx.Compiler do
         rest,
         state.engine.handle_begin(buffer),
         [contents | scope],
-        %{state | quoted: [], line: line, start_line: start_line}
+        %{
+          state
+          | quoted: [],
+            line: line,
+            start_line: start_line,
+            start_column: column(start_column, mark)
+        }
       )
 
     buffer = state.engine.handle_expr(buffer, IO.chardata_to_string(mark), contents)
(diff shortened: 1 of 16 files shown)


lib/eex/lib/eex/compiler.ex
```diff
@@ -84,27 +97,39 @@ defmodule EEx.Compiler do
...
 
     :elixir_errors.erl_warn(line, state.file, message)
-    generate_buffer([{:middle_expr, line, '', chars, trimmed?} | t], buffer, scope, state)
+    generate_buffer([{:middle_expr, line, column, '', chars, trimmed?} | t], buffer, scope, state)
     # TODO: Make this an error on Elixir v2.0 since it accidentally worked previously.
     # raise EEx.SyntaxError, message: message, file: state.file, line: line
   end
 
-  defp generate_buffer([{:middle_expr, line, _, chars, _} | _], _buffer, [], state) do
+  defp generate_buffer([{:middle_expr, line, _column, _, chars, _} | _], _buffer, [], state) do
```
wojtekmach: In the future column could be used to do even better error messages.

josevalim: Should we use the opportunity and add those now to EEx.SyntaxError? Is should be a small amount of work.

lib/eex/lib/eex/compiler.ex
```diff
@@ -13,16 +13,19 @@ defmodule EEx.Compiler do
...
       {:ok, tokens} ->
         state = %{
           engine: opts[:engine] || @default_engine,
           file: file,
           line: line,
           quoted: [],
-          start_line: nil
+          start_line: nil,
+          start_column: nil,
+          columns: Code.get_compiler_option(:parser_options)[:columns] == true
```
josevalim: Maybe we should keep all `parser_options` here and pass all of them down to the `string_to_quoted!` call.

lib/eex/lib/eex/compiler.ex
```diff
@@ -13,23 +13,26 @@ defmodule EEx.Compiler do
   def compile(source, opts) when is_binary(source) and is_list(opts) do
     file = opts[:file] || "nofile"
     line = opts[:line] || 1
+    column = 1
```
wojtekmach: I didn't have a use case for it yet, but in the future we might want to do this: `column = opts[:column] || 1` to allow user to adjust things.

josevalim:

JMurphyWeb:
@wojtekmach am I right in thinking we should also have column number in `meta` for when we use `@` macro in EEx templates?

This is me trying in iex. No column data:

```elixir
# iex session: Interactive Elixir (1.11.0-dev)
Code.put_compiler_option(:parser_options, columns: true)
:ok
EEx.compile_string("""
   <% @hi %>
""")
{:__block__, [],
 [
   {{:., [line: 1],
     [{:__aliases__, [line: 1, alias: false], [:EEx, :Engine]}, :fetch_assign!]},
    [line: 1],
    [
      {:var!, [line: 1, context: EEx.Engine, import: Kernel],
       [{:assigns, [line: 1], EEx.Engine}]},
      :hi
    ]},
   {:<<>>, [], ["  ", "\n"]}
 ]}
```

It does work for accessing assigns in using `Access.get`:
```elixir
# iex session: Interactive Elixir (1.11.0-dev)
Code.put_compiler_option(:parser_options, columns: true)
:ok
EEx.compile_string("""
  <%= assigns[:hi] %>
""")
{:__block__, [],
 [
   {:=, [],
    [
      {:arg0, [], EEx.Engine},
      {{:., [], [{:__aliases__, [alias: false], [:String, :Chars]}, :to_string]},
       [],
       [
         {{:., [line: 1, column: 12], [Access, :get]}, [line: 1, column: 12],
          [{:assigns, [line: 1, column: 5], nil}, :hi]}
       ]}
    ]},
   {:<<>>, [],
    [{:"::", [], [{:arg0, [], EEx.Engine}, {:binary, [], EEx.Engine}]}, "\n"]}
 ]}
```

I am trying to build a tool inspired by import2alias which locates unused assigns in templates :)

wojtekmach:
@JMurphyWeb good catch. This seems to do the trick:

```diff
diff --git a/lib/eex/lib/eex/engine.ex b/lib/eex/lib/eex/engine.ex
index 7c110396b..a1b18a375 100644
--- a/lib/eex/lib/eex/engine.ex
+++ b/lib/eex/lib/eex/engine.ex
@@ -119,7 +119,9 @@ defmodule EEx.Engine do
   @spec handle_assign(Macro.t()) :: Macro.t()
   def handle_assign({:@, meta, [{name, _, atom}]}) when is_atom(name) and is_atom(atom) do
     line = meta[:line] || 0
-    quote(line: line, do: EEx.Engine.fetch_assign!(var!(assigns), unquote(name)))
+    column = meta[:column]
+    {name, meta, args} = quote(line: line, do: EEx.Engine.fetch_assign!(var!(assigns), unquote(name)))
+    {name, Keyword.put(meta, :column, column), args}
   end
 
   def handle_assign(arg) do
```

but it's a bit dirty so I'll look into more reliable solution!

JMurphyWeb:
Thanks @wojtekmach !

Just to confirm, this is something we will need to wait for to be merged into elixir?

Should I open an issue or take any further action? (id love to open a PR but fear this is beyond me!)

wojtekmach:
Yes, we'd have to wait until it's merged into Elixir. I'll submit a PR with this patch and we'll see how it goes.

JMurphyWeb:
I don't know if this is a bug @wojtekmach 

I've stopped working on the tool that I needed it for.

I guess there's no open issue though.

Thanks for the attempted fix 👍

PR #9717, opened 2020-01-11
Add :. and :<- to special forms

omohokcoj:
To fix `Macro.special_form?/2` behaviour:
```elixir
iex(35)> Macro.special_form?(:., 2)
false # should be true

iex(36)> Macro.special_form?(:<-, 1)
false # should be true
```

`:.` is a special form based on - https://github.com/elixir-lang/elixir/blob/v1.9.4/lib/elixir/lib/kernel/special_forms.ex#L486
`:<-` must be a special form as well (it's a part of `for` list comprehension and `with` construction).


josevalim:
Hi @omohokcoj, FWIW `<-` is not a special form in Elixir. `<-` has no default meaning in Elixir, the special forms `for` and `with` are the ones that give it meaning. I have removed it and merged the dot. Thanks!

josevalim:

PR #9718, opened 2020-01-12
Add is_struct/2 macro

hauleth:


lib/elixir/lib/kernel.ex
```diff
@@ -2250,6 +2250,49 @@ defmodule Kernel do
...
+  """
+  @doc since: "1.10.0", guard: true
+  defmacro is_struct(term, name) do
+    case __CALLER__.context do
+      nil ->
+        quote do
+          case unquote(name) do
+            name when is_atom(name) ->
+              case unquote(term) do
+                map when is_map(map) and :erlang.map_get(:__struct__, map) == name -> true
```
```suggestion
                %{__struct__: ^name} -> true
```

hauleth:
When this will get merged then probably we should also add `MapSet.is_memeber/2` macro which will work exactly as `MapSet.memeber?/2` but will also be available in guards.

josevalim:
Let's hold on the `is_member` macro because I did a propose for a new Erlang mapset and it is not yet clear if they will add such set and, if they do, if its structure will be public. :)

sorentwo:
> Let's hold on the `is_member` macro because I did a propose for a new Erlang mapset and it is not yet clear if they will add such set and, if they do, if its structure will be public. :)

For anybody else who was curious about the proposal, it is [EEP 50](https://www.erlang.org/erlang-enhancement-proposals/eep-0050.html).

josevalim:

lib/elixir/lib/kernel.ex
```diff
@@ -2250,6 +2250,50 @@ defmodule Kernel do
...
+  ## Examples
+
+      iex> is_struct(URI.parse("/"), URI)
+      true
+
+      iex> is_struct(URI.parse("/"), Macro.Env)
+      false
+
+  """
+  @doc since: "1.10.0", guard: true
```
princemaple: this should be 1.11, shouldn't it?

hauleth:
Elixir 1.10.0 is still unreleased, so we still can add new functionality. Nothing against the rules there. 

--
Łukasz Jan Niemier

princemaple:
it's on changelog of v1.11, and Jose replied to my other, wrongly placed, comment on the commit, so I sent #9726 :)

josevalim:
We avoid adding new features once a RC is out. Otherwise if the new feature
has bugs we can stay forever in RC state. :D
-- 

PR #9719, opened 2020-01-12
Don't GPG sign commits or tags in git tests

jonnystoten:
If the machine/user git config is set to sign all commits/tags, these tests fail.

### DIFF
--- a/lib/mix/test/mix/tasks/escript_test.exs
+++ b/lib/mix/test/mix/tasks/escript_test.exs
@@ -351,7 +351,7 @@ defmodule Mix.Tasks.EscriptTest do
       """)
 
       System.cmd("git", ~w[add .])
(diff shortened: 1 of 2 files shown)


adkron:
What is the reason for not wanting GPG signed commits?

josevalim:
@adkron this is just for tests, so I assume the tests fail otherwise.

adkron:
@josevalim, thanks, I feel ignorant now. I just saw the words and thought why would you want to stop them from signing their commits. I just looked at the code now and realized that this is a test making commits. Sorry for making you take the time to respond. I should have looked closer. Thanks, for your patience as always.

josevalim:
@adkron no problem, it happens to all of us!

josevalim:

PR #9720, opened 2020-01-12
Suggest Ctrl+\ in iex startup message

jonnystoten:
The current suggestion, Ctrl+C, doesn't actually exit.


josevalim:
Hi @jonnystoten! Ctrl+C shows the break menu, which has the `(a)bort` option and other interesting options which may be useful in other situations, so for now I still prefer pointing users to the break menu. Thank you!

PR #9724, opened 2020-01-13
Add :indentation option to EEx

wojtekmach:


lib/eex/lib/eex/tokenizer.ex
```diff
@@ -101,19 +102,19 @@ defmodule EEx.Tokenizer do
...
 
-  defp expr([?%, ?> | t], line, column, buffer) do
+  defp expr([?%, ?> | t], line, column, _opts, buffer) do
     {:ok, buffer, line, column + 2, t}
   end
 
-  defp expr('\n' ++ t, line, _column, buffer) do
-    expr(t, line + 1, 1, [?\n | buffer])
+  defp expr('\n' ++ t, line, _column, opts, buffer) do
+    expr(t, line + 1, opts[:column_padding] + 1, opts, [?\n | buffer])
```
josevalim: Wrong name, which means this isn't tested at all? 🤔

lib/eex/lib/eex/tokenizer.ex
```diff
@@ -37,15 +37,16 @@ defmodule EEx.Tokenizer do
   def tokenize(list, line, column, opts)
       when is_list(list) and is_integer(line) and line >= 0 and is_integer(column) and column >= 0 and
              is_list(opts) do
-    tokenize(list, line, column, opts, [], [])
+    opts = Keyword.put_new(opts, :indent, 0)
```
josevalim: Let's use the opportunity and make it a map. Only :trim and :indent seem to be used.

lib/eex/lib/eex/tokenizer.ex
wojtekmach: good catch, it broke the build too. Fixed!

lib/eex/lib/eex/compiler.ex
```diff
@@ -14,9 +14,9 @@ defmodule EEx.Compiler do
     file = opts[:file] || "nofile"
     line = opts[:line] || 1
     column = 1
-    indent = opts[:indent] || 0
+    indentation = opts[:indentation] || 0
     trim = opts[:trim] || false
-    tokenizer_options = [trim: trim, indent: indent]
+    tokenizer_options = [trim: trim, indentation: indentation]
```
josevalim: Let's make it a map and pass it to the tokenizer. The tokenizer should always assume a proper map is given (it may require some changes in the tests but since it is private API, it should be fine).

lib/eex/lib/eex.ex
```diff
@@ -39,12 +39,15 @@ defmodule EEx do
   All functions in this module accept EEx-related options.
   They are:
 
-    * `:line` - the line to be used as the template start. Defaults to 1.
     * `:file` - the file to be used in the template. Defaults to the given
       file the template is read from or to "nofile" when compiling from a string.
+    * `:line` - the line to be used as the template start. Defaults to 1.
+    * `:indentation` - (since v1.11.0) indentation to be added to each column.
```
michallepicki: It seems off to define an option using its name. Also I may be wrong, but I while internally it probably gets applied to each column in each line, I think lines are more important than columns here. How about:
```suggestion
    * `:indentation` - (since v1.11.0) amount of spaces to be added before each line.
```

wojtekmach: yeah, I'm not happy super happy with the wording either but fwiw we define e.g. `:line` in terms of line :)

My first thought when I read your proposed wording is as if we somehow literally inject space characters before each line and thus adjust output but the option is about adjusting :column in the AST so that e.g. stacktraces would point to the exact spot where we e.g. call a function. (again, not saying that's what you meant but how I'd likely have interpreted it.)

michallepicki: Right, someone could interpret it this way. Not sure how to improve it then, it could stay as is until someone else has a better idea.

josevalim: "an integer to be added to the column after every new line"?

wojtekmach: I went with:

```
    * `:indentation` - (since v1.11.0) an integer added to the column after every
       new line. Defaults to 0.
```

👍

josevalim:

PR #9725, opened 2020-01-13
Add :indentation to sigil heredoc meta

wojtekmach:


lib/elixir/src/elixir_parser.yrl
```diff
@@ -640,6 +640,13 @@ meta_from_location({Line, Column, _}) ->
   case ?columns() of
     true -> [{line, Line}, {column, Column}];
     false -> [{line, Line}]
+  end;
+
+%% TODO: dry up
+meta_from_location({Line, Column, Indent, _}) ->
```
wojtekmach: I wasn't sure where exactly to put indent into the token, whether it should be in location or some other part? Should it be always added to the ast or only under a particular option like `token_metadata: true`?

lib/elixir/lib/macro.ex
```diff
@@ -172,6 +172,7 @@ defmodule Macro do
     * `:end_of_expression` - denotes when the end of expression effectively
       happens. Available for all expressions except the last one inside a
       `__block__` (when `:token_metadata` is true)
+    * `:indent` - The indentation of a non-interpolated sigil heredoc
```
josevalim: Would the correct name here be "indentation"? Since we are not asking it to indent but rather, returning the indentation? Maybe we should call it indentation in the EEx option as well for simplicity? Like "column".

lib/elixir/src/elixir_tokenizer.erl
```diff
@@ -191,8 +191,9 @@ tokenize([$# | String], Line, Column, Scope, Tokens) ->
 tokenize([$~, S, H, H, H | T] = Original, Line, Column, Scope, Tokens) when ?is_quote(H), ?is_upcase(S) orelse ?is_downcase(S) ->
   case extract_heredoc_with_interpolation(Line, Column, Scope, ?is_downcase(S), T, H) of
     {ok, NewLine, NewColumn, Parts, Rest} ->
+      Indent = NewColumn - 4,
       {Final, Modifiers} = collect_modifiers(Rest, []),
-      Token = {sigil, {Line, Column, nil}, S, Parts, Modifiers, <<H, H, H>>},
+      Token = {sigil, {Line, Column, Indent, nil}, S, Parts, Modifiers, <<H, H, H>>},
```
josevalim: I would always include the indentation. The reason is that we want the indentation even if columns is disabled otherwise EEx.SyntaxError would be incorrect. So I would remove it from the location tuple and move it to the `sigil` tuple, right after `Modifiers` (I think the last element has to be a binary) .

lib/elixir/src/elixir_tokenizer.erl
```diff
@@ -204,7 +205,8 @@ tokenize([$~, S, H | T] = Original, Line, Column, Scope, Tokens) when ?is_sigil(
   case elixir_interpolation:extract(Line, Column + 3, Scope, ?is_downcase(S), T, sigil_terminator(H)) of
     {NewLine, NewColumn, Parts, Rest} ->
       {Final, Modifiers} = collect_modifiers(Rest, []),
-      Token = {sigil, {Line, Column, nil}, S, tokens_to_binary(Parts), Modifiers, <<H>>},
+      Indentation = nil,
```
wojtekmach: I wasn't sure what to put into non-heredoc sigils, either this (so we don't put indentation into meta) or maybe `Indentation = Column + 3`, the start of the sigil string.

josevalim: `nil` sounds good to me.

josevalim:

PR #9726, opened 2020-01-13
Fix since for is_struct/2

princemaple:


josevalim:

PR #9731, opened 2020-01-14
Increase memory on Windows builds

josevalim:


josevalim:

eksperimental:
Sorry for the late reply. I think they can by multiple of 256Mib. Anyways, if you put an invalid number, you will get an error message that will guide you.
Where we hitting a limit with the previous memory>?I have set it to the minimum, because that will give us a better slot when queuing the tasks.
If 4Gb or 5Gb is enough, I would recommend to go with those.

PR #9733, opened 2020-01-14
Reject (even more) invalid line numbers in ExUnit filters

### LINKED ISSUE #9727: Unexpected behavior of mix test with specific invalid line numbers
### Environment

* Elixir & Erlang/OTP versions (elixir --version): Elixir 1.9.4 (compiled with Erlang/OTP 20)
* Operating system: Ubuntu 18.04.3 LTS (kernel 4.15.0-72-generic)

### Current behavior

Create a new Elixir project with `mix new foo`, enter the project path with `cd foo`, then edit  `test/foo_test.exs` to the following:
```elixir
defmodule FooTest do
  use ExUnit.Case
  test "A", do: assert(true)

  test "B", do: assert(true)
  test "C", do: assert(true)
end
```

Examples of current behavior:
```
$ mix test --trace test/foo_test.exs:-43:0:1:2
Excluding tags: [:test]
Including tags: [line: "-43", line: "0", line: "1", line: "2"]


FooTest
  * test A (excluded)
  * test C (excluded)
  * test B (excluded)


Finished in 0.01 seconds
3 tests, 0 failures, 3 excluded

Randomized with seed 642462
```
```
$ mix test --trace test/foo_test.exs:4
Excluding tags: [:test]
Including tags: [line: "4"]


FooTest
  * test B (excluded)
  * test C (excluded)
  * test A (0.00ms)


Finished in 0.01 seconds
3 tests, 0 failures, 2 excluded

Randomized with seed 258138
```
```
$ foo mix test --trace test/foo_test.exs:42
Excluding tags: [:test]
Including tags: [line: "42"]


FooTest
  * test B (excluded)
  * test A (excluded)
  * test C (0.00ms)


Finished in 0.01 seconds
3 tests, 0 failures, 2 excluded

Randomized with seed 489324
```

### Expected behavior

It seems to me that the expected behavior would be (i) to warn about the specific lines that do not exist, such as -43, 0 and 42 (first and last examples); (ii) to warn about specific lines that do not contain any test definition, such as line 1 and 4 (first and second examples); and (iii) to not run any test.

adrianomitre:
If one call `mix test foo:arg` where `arg` are not natural numbers, we show a warning and reject those bad line numbers.

Extends https://github.com/elixir-lang/elixir/pull/9729 and further resolves #9727

### DIFF
--- a/lib/ex_unit/lib/ex_unit/filters.ex
+++ b/lib/ex_unit/lib/ex_unit/filters.ex
@@ -54,9 +54,11 @@ defmodule ExUnit.Filters do
         {num, ""} -> num <= 0
         _ -> true
       end
+
     if invalid do
       IO.warn("invalid line number given as ExUnit filter: #{arg}", [])
     end
+
     invalid
   end
 
--- a/lib/ex_unit/lib/ex_unit/filters.ex
+++ b/lib/ex_unit/lib/ex_unit/filters.ex
@@ -27,23 +27,35 @@ defmodule ExUnit.Filters do
       [part] ->
         {part, []}
 
-      [path | line_numbers] ->
+      parts ->
+        {reversed_line_numbers, reversed_path_parts} =
+          parts
+          |> Enum.reverse()
+          |> Enum.split_while(&match?({_, ""}, Integer.parse(&1)))
+
         line_numbers =
-          line_numbers
-          |> Enum.flat_map(&line_number_argument/1)
+          reversed_line_numbers
+          |> Enum.reject(&invalid_line_number?/1)
+          |> Enum.reverse()
+          |> Enum.map(&{:line, &1})
+
+        path =
+          reversed_path_parts
+          |> Enum.reverse()
+          |> Enum.join(":")
 
         {path, line_numbers}
     end
   end
 
-  defp line_number_argument(arg) do
+  defp invalid_line_number?(arg) do
     case Integer.parse(arg) do
       {num, ""} when num > 0 ->
-        [line: arg]
+        false
 
       _ ->
         IO.warn("invalid line number given as ExUnit filter: #{arg}", [])
-        []
+        true
     end
   end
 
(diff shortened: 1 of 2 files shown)


lib/ex_unit/lib/ex_unit/filters.ex
```diff
@@ -48,18 +48,18 @@ defmodule ExUnit.Filters do
...
-  defp invalid_line_number?("0") do
-    IO.warn("invalid line number given as ExUnit filter: 0", [])
-    true
-  end
-
-  defp invalid_line_number?("-" <> _ = num) do
-    IO.warn("invalid line number given as ExUnit filter: #{num}", [])
-    true
+  defp invalid_line_number?(arg) do
+    invalid =
```
josevalim: Hi! I think you can rewrite it to this:

```
      case Integer.parse(arg) do
        {num, ""} when num > 0 -> false
        _ ->
          IO.warn("invalid line number given as ExUnit filter: #{arg}", [])
          true
      end
```

WDYT?

adrianomitre: Even better! Implemented in https://github.com/elixir-lang/elixir/pull/9733/commits/81b51ca45fddd215493ef4729c9438ffdceb6b61

josevalim:
@adrianomitre the test that you added is failing. It seems that we don't consider :1L to be a line location at all, which makes sense, as "C:1L" is most likely a valid part on Windows. What would happen in your example then is that `mix test` will fail because no such file exists.

josevalim:
Your changes will break paths on Windows. :) We need to assume that if `:` is not followed by a number, then it is part of a path. Thank you for sanity checking though!

adrianomitre:
What test was it breaking?

adrianomitre:
Could you please re-open the PR? I have reverted the breaking change.

PR #9734, opened 2020-01-14
Add column to EEx.SyntaxError

wojtekmach:


lib/eex/lib/eex/compiler.ex
```diff
@@ -159,7 +161,8 @@ defmodule EEx.Compiler do
     raise EEx.SyntaxError,
       message: "unexpected end of string, expected a closing '<% end %>'",
       file: state.file,
-      line: state.line
+      line: state.line,
+      column: "TODO"
```
wojtekmach: I couldn't figure out a neat way to get column for this error. I guess I could have end_column for each token and keep it around in the compiler as stat.end_column. Is there a simpler way?

josevalim: @wojtekmach make the tokenizer emit a `{:eof, line, column}` when it is done. Then you change this clause and the one above to match on `[{:eof, line, column}]` instead of the empty list. :)

wojtekmach: ❤️

josevalim:

PR #9736, opened 2020-01-15
improve speed of extract_files

michaelst:
This improves the speed of extract files by at least 50%, many times even more than that. However, it does lose the ability to pass in a wildcard to Path.wildcard, it can only filter on extension.

### DIFF
--- a/lib/mix/lib/mix/tasks/test.ex
+++ b/lib/mix/lib/mix/tasks/test.ex
@@ -468,8 +468,8 @@ defmodule Mix.Tasks.Test do
 
     # Finally parse, require and load the files
     test_files = parse_files(files, shell, test_paths)
-    test_pattern = project[:test_pattern] || "*_test.exs"
-    warn_test_pattern = project[:warn_test_pattern] || "*_test.ex"
+    test_pattern = project[:test_pattern] || "_test.exs"
+    warn_test_pattern = project[:warn_test_pattern] || "_test.ex"
 
     matched_test_files =
       test_files
--- a/lib/mix/lib/mix/utils.ex
+++ b/lib/mix/lib/mix/utils.ex
@@ -212,6 +212,11 @@ defmodule Mix.Utils do
   recursively for files with the given extensions or matching
   the given patterns.
   """
+  def extract_files(paths, [first | _] = exts) when is_atom(first),
+    do: extract_files(paths, Enum.map(exts, &"#{&1}"))
+
+  def extract_files(paths, exts) when is_atom(exts), do: extract_files(paths, "#{exts}")
+
   def extract_files(paths, exts) do
     Enum.reduce(paths, [], fn item, acc ->
       if String.ends_with?(item, exts) do
--- a/lib/mix/lib/mix/utils.ex
+++ b/lib/mix/lib/mix/utils.ex
@@ -213,9 +213,9 @@ defmodule Mix.Utils do
   the given patterns.
   """
   def extract_files(paths, [first | _] = exts) when is_atom(first),
-    do: extract_files(paths, Enum.map(exts, &"#{&1}"))
+    do: extract_files(paths, Enum.map(exts, &".#{&1}"))
 
-  def extract_files(paths, exts) when is_atom(exts), do: extract_files(paths, "#{exts}")
+  def extract_files(paths, exts) when is_atom(exts), do: extract_files(paths, ".#{exts}")
 
   def extract_files(paths, exts) do
     Enum.reduce(paths, [], fn item, acc ->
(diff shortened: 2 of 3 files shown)


josevalim:
Thanks @michaelst. As you can see, this breaks the build as we are basically reinventing `Path.wildcard`. So my understand is that Path.wildcard is the one slower in Docker? If so, can you provide a Docker container that reproduces the slowdown so I can investigate it further? I would love to look at the root cause and address it directly there. Thank you.

michaelst:
Path.wildcard is slower than my implementation on both Mac and Docker. I posted benchmarks in this thread https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!topic/elixir-lang-core/iQ0mUTngg5k

josevalim:
Alright. If I can reproduce it on my mac, then that's good enough for me. How many files do you have in `lib`?

michaelst:
We have 875 .ex files in lib

josevalim:
@michaelst all at the root or are they listed as a tree somehow?

michaelst:
They are all nested with max depth of probably 4 or 5

michaelst:
@josevalim I got it down to 13 failing tests relating to the mix test task. I can work on it more tonight to get those fixed.

michaelst:
I was made aware of Benchee today and ran my code through that on this repo. Here are the results:
```
Operating System: macOS
CPU Information: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
Number of Available Cores: 12
Available memory: 16 GB
Elixir 1.9.4
Erlang 22.1

warmup: 2 s
time: 10 s
memory time: 2 s
parallel: 1
inputs: none specified
Estimated total run time: 28 s

Benchmarking Mix.Utils.extract_files...
Benchmarking new extract_files...

Name                              ips        average  deviation         median         99th %
new extract_files                3.04      328.70 ms     ±1.82%      327.83 ms      358.34 ms
Mix.Utils.extract_files          1.79      557.63 ms     ±2.56%      554.75 ms      610.58 ms

Comparison: 
new extract_files                3.04
Mix.Utils.extract_files          1.79 - 1.70x slower +228.93 ms

Memory usage statistics:

Name                       Memory usage
new extract_files              21.52 MB
Mix.Utils.extract_files       130.49 MB - 6.06x memory usage +108.97 MB

```

Here is the script I ran on the elixir repo:
```
defmodule My.Mix.Utils do
  def extract_files(paths, [first | _] = exts) when is_atom(first),
    do: extract_files(paths, Enum.map(exts, &".#{&1}"))

  def extract_files(paths, exts) when is_atom(exts), do: extract_files(paths, ".#{exts}")

  def extract_files(paths, exts) do
    Enum.reduce(paths, [], fn item, acc ->
      if String.ends_with?(item, exts) do
        [item] ++ acc
      else
        item
        |> File.ls()
        |> case do
          {:ok, files} ->
            files
            |> Enum.map(&Path.join(item, &1))
            |> extract_files(exts)
            |> Kernel.++(acc)

          _ ->
            acc
        end
      end
    end)
  end
end

Benchee.run(
  %{
    "Mix.Utils.extract_files" => fn -> Mix.Utils.extract_files(["../elixir/lib"], "*.ex") end,
    "new extract_files" => fn -> My.Mix.Utils.extract_files(["../elixir/lib"], ".ex") end
  },
  time: 10,
  memory_time: 2
)
```

michaelst:
@josevalim all tests are passing, one Windows build failed but that seems unrelated to my changes. Can you confirm?

josevalim:
@michaelst fwiw, your script has an issue that it would treat a directory named "foo.ex" as a file. :)

josevalim:
Nevermind, the directory issue also exists in the current implementation. :)

lib/mix/lib/mix/tasks/test.ex
```diff
@@ -468,8 +468,8 @@ defmodule Mix.Tasks.Test do
 
     # Finally parse, require and load the files
     test_files = parse_files(files, shell, test_paths)
-    test_pattern = project[:test_pattern] || "*_test.exs"
-    warn_test_pattern = project[:warn_test_pattern] || "*_test.ex"
+    test_pattern = project[:test_pattern] || "_test.exs"
+    warn_test_pattern = project[:warn_test_pattern] || "_test.ex"
```
josevalim: This is, unfortunately, a breaking change as both of those are public APIs. :/

josevalim:
Hi @michaelst! You were correct all along! The wildcard implementation does more system calls than necessary. I have sent a PR to Erlang/OTP here: https://github.com/erlang/otp/pull/2506 

Thank you so much for the isolating this issue! At the end of the day, I actually think your implementation is better, but unfortunately we cannot accept it due to breaking changes.

PR #9741, opened 2020-01-15
Add Task.await_many

iangreenleaf:
Adds a new function `Task.await_many` per [the proposal](https://groups.google.com/forum/#!topic/elixir-lang-core/9QQNIZMzWVE).

I'm still learning my way around the low-level process model, so it's possible I've made mistakes with the process management.

I also ended up doing some data shuffling in the interest of returning an ordered list but keeping a constant access time while constructing it. Maybe there's a better way to handle that, but if so it has escaped me.

### DIFF
--- a/lib/elixir/lib/task.ex
+++ b/lib/elixir/lib/task.ex
@@ -657,9 +657,18 @@ defmodule Task do
         Process.send_after(self(), timeout_ref, timeout)
       end
 
-    awaiting = build_awaiting(tasks, %{}, 0)
-
     try do
+      awaiting =
+        for task <- tasks, into: %{} do
+          %Task{ref: ref, owner: owner} = task
+
+          if owner != self() do
+            raise ArgumentError, invalid_owner_error(task)
+          end
+
+          {ref, true}
+        end
+
       await_many(tasks, timeout, awaiting, %{}, timeout_ref)
     after
       timer_ref && Process.cancel_timer(timer_ref)
@@ -667,8 +676,8 @@ defmodule Task do
     end
   end
 
-  defp await_many(_tasks, _timeout, map, replies, _timeout_ref) when map_size(map) == 0 do
-    build_replies_list(replies)
+  defp await_many(tasks, _timeout, map, replies, _timeout_ref) when map_size(map) == 0 do
+    for %{ref: ref} <- tasks, do: Map.fetch!(replies, ref)
   end
 
   defp await_many(tasks, timeout, awaiting, replies, timeout_ref) do
@@ -686,33 +695,10 @@ defmodule Task do
           tasks,
           timeout,
           Map.delete(awaiting, ref),
-          Map.put(replies, awaiting[ref], reply),
+          Map.put(replies, ref, reply),
           timeout_ref
         )
     end
-  end
-
-  defp build_awaiting([%Task{ref: ref, owner: owner} = task | rest], awaiting, idx) do
-    if owner != self() do
-      raise ArgumentError, invalid_owner_error(task)
-    end
-
-    build_awaiting(rest, Map.put(awaiting, ref, idx), idx + 1)
-  end
-
-  defp build_awaiting([], awaiting, _idx) do
-    awaiting
-  end
-
-  defp build_replies_list(map) do
-    Stream.unfold({map, 0}, fn
-      {remain, _} when map_size(remain) == 0 ->
-        nil
-
-      {remain, idx} ->
-        {remain[idx], {Map.delete(remain, idx), idx + 1}}
-    end)
-    |> Enum.to_list()
   end
 
   @doc false
--- a/lib/elixir/lib/task.ex
+++ b/lib/elixir/lib/task.ex
@@ -608,6 +608,118 @@ defmodule Task do
     end
   end
 
+  @doc """
+  Awaits replies from multiple tasks and returns them.
+
+  This function receives a list of tasks and waits for their replies in the
+  given time interval. It returns a list of the results, in the same order as
+  the tasks supplied in the `tasks` input argument.
+
+  If any of the task processes dies, the current process will exit with the
+  same reason as that task.
+
+  A timeout in milliseconds or `:infinity`, can be given with a default value
+  of `5000`. If the timeout is exceeded, then the current process will exit.
+  Any task processes that are linked to the current process (which is the case
+  when a task is started with `async`) will also exit. Any task processes that
+  are trapping exits or not linked to the current process will continue to run.
+
+  This function assumes the tasks' monitors are still active or the monitors'
+  `:DOWN` message is in the message queue. If any tasks have been demonitored,
+  or the message already received, this function will wait for the duration of
+  the timeout.
+
+  This function can only be called once for any given task. If you want to be
+  able to check multiple times if a long-running task has finished its
+  computation, use `yield_many/2` instead.
+
+  ## Compatibility with OTP behaviours
+
+  It is not recommended to `await` long-running tasks inside an OTP behaviour
+  such as `GenServer`. See `await/2` for more information.
+
+  ## Examples
+
+      iex> tasks = [
+      ...>   Task.async(fn -> 1 + 1 end),
+      ...>   Task.async(fn -> 2 + 3 end)
+      ...> ]
+      iex> Task.await_many(tasks)
+      [2, 5]
+
+  """
+  @spec await_many([t], timeout) :: [term]
+  def await_many(tasks, timeout \\ 5000) when is_timeout(timeout) do
+    timeout_ref = make_ref()
+
+    if timeout != :infinity do
+      Process.send_after(self(), timeout_ref, timeout)
+    end
+
+    awaiting = build_awaiting(tasks, %{}, 0)
+
+    await_many(tasks, timeout, awaiting, %{}, timeout_ref)
+  end
+
+  defp await_many(_tasks, _timeout, map, replies, timeout_ref) when map_size(map) == 0 do
+    Process.demonitor(timeout_ref, [:flush])
+    build_replies_list(replies)
+  end
+
+  defp await_many(tasks, timeout, awaiting, replies, timeout_ref) do
+    receive do
+      ^timeout_ref ->
+        exit({:timeout, {__MODULE__, :await_many, [tasks, timeout]}})
+
+      {:DOWN, ref, _, proc, reason} ->
+        if Map.has_key?(awaiting, ref) do
+          exit({reason(reason, proc), {__MODULE__, :await_many, [tasks, timeout]}})
+        else
+          await_many(tasks, timeout, awaiting, replies, timeout_ref)
+        end
+
+      {ref, reply} ->
+        case Map.fetch(awaiting, ref) do
+          {:ok, idx} ->
+            Process.demonitor(ref, [:flush])
+
+            await_many(
+              tasks,
+              timeout,
+              Map.delete(awaiting, ref),
+              Map.put(replies, idx, reply),
+              timeout_ref
+            )
+
+          :error ->
+            await_many(tasks, timeout, awaiting, replies, timeout_ref)
+        end
+    end
+  end
+
+  defp build_awaiting([%Task{ref: ref, owner: owner} = task | rest], awaiting, idx) do
+    if owner != self() do
+      raise ArgumentError, invalid_owner_error(task)
+    end
+
+    build_awaiting(rest, Map.put(awaiting, ref, idx), idx + 1)
+  end
+
+  defp build_awaiting([], awaiting, _idx) do
+    awaiting
+  end
+
+  defp build_replies_list(map) do
+    Stream.unfold({map, 0}, fn
+      {remain, _} when map_size(remain) == 0 ->
+        nil
+
+      {remain, idx} ->
+        {remain[idx], {Map.delete(remain, idx), idx + 1}}
+    end)
+    |> Enum.to_list()
+  end
+
   @doc false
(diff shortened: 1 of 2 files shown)


lib/elixir/lib/task.ex
```diff
@@ -608,6 +608,108 @@ defmodule Task do
...
+    Process.demonitor(timeout_ref, [:flush])
+    build_replies_list(replies)
+  end
+
+  defp await_many(tasks, timeout, awaiting, replies, timeout_ref) do
+    receive do
+      ^timeout_ref ->
+        exit({:timeout, {__MODULE__, :await_many, [tasks, timeout]}})
+
+      {:DOWN, ref, _, proc, reason} when :erlang.is_map_key(ref, awaiting) ->
```
josevalim:
```suggestion
      {:DOWN, ref, _, proc, reason} when is_map_key(awaiting, ref) ->
```

lib/elixir/lib/task.ex
```diff
@@ -608,6 +608,108 @@ defmodule Task do
...
+        exit({reason(reason, proc), {__MODULE__, :await_many, [tasks, timeout]}})
+
+      {ref, reply} when :erlang.is_map_key(ref, awaiting) ->
```
josevalim:
```suggestion
      {ref, reply} when is_map_key(awaiting, ref) ->
```

josevalim:
H @iangreenleaf, I think it is best if we keep the implementation closer to `yield_many`. We can probably even share the foundation for both `yield_many` and `await_many`. Here is a diff (untested) of the approach I have in mind:

```diff
diff --git a/lib/elixir/lib/task.ex b/lib/elixir/lib/task.ex
index 6f4f9f924..6dcc62100 100644
--- a/lib/elixir/lib/task.ex
+++ b/lib/elixir/lib/task.ex
@@ -757,6 +757,24 @@ defmodule Task do
   """
   @spec yield_many([t], timeout) :: [{t, {:ok, term} | {:exit, term} | nil}]
   def yield_many(tasks, timeout \\ 5000) when is_timeout(timeout) do
+    try do
+      map_many(tasks, timeout, fn task, return -> {task, return} end)
+    catch
+      {:noconnection, reason} ->
+        exit({reason, {__MODULE__, :yield_many, [tasks, timeout]}})
+    end
+  end
+
+  def await_many(tasks, timeout \\ 5000) when is_timeout(timeout) do
+    try do
+      map_many(tasks, timeout, fn task, return -> {task, return} end)
+    catch
+      {:noconnection, reason} ->
+        exit({reason, {__MODULE__, :await_many, [tasks, timeout]}})
+    end
+  end
+
+  defp map_many(tasks, timeout, fun) do
     timeout_ref = make_ref()
 
     timer_ref =
@@ -765,17 +783,14 @@ defmodule Task do
       end
 
     try do
-      yield_many(tasks, timeout_ref, :infinity)
-    catch
-      {:noconnection, reason} ->
-        exit({reason, {__MODULE__, :yield_many, [tasks, timeout]}})
+      map_many(tasks, timeout_ref, :infinity, fun)
     after
       timer_ref && Process.cancel_timer(timer_ref)
       receive do: (^timeout_ref -> :ok), after: (0 -> :ok)
     end
   end
 
-  defp yield_many([%Task{ref: ref, owner: owner} = task | rest], timeout_ref, timeout) do
+  defp map_many([%Task{ref: ref, owner: owner} = task | rest], timeout_ref, timeout, fun) do
     if owner != self() do
       raise ArgumentError, invalid_owner_error(task)
     end
@@ -783,23 +798,23 @@ defmodule Task do
     receive do
       {^ref, reply} ->
         Process.demonitor(ref, [:flush])
-        [{task, {:ok, reply}} | yield_many(rest, timeout_ref, timeout)]
+        [fun.(task, {:ok, reply}) | map_many(rest, timeout_ref, timeout, fun)]
 
       {:DOWN, ^ref, _, proc, :noconnection} ->
         throw({:noconnection, reason(:noconnection, proc)})
 
       {:DOWN, ^ref, _, _, reason} ->
-        [{task, {:exit, reason}} | yield_many(rest, timeout_ref, timeout)]
+        [fun.(task, {:exit, reason}) | map_many(rest, timeout_ref, timeout, fun)]
 
       ^timeout_ref ->
-        [{task, nil} | yield_many(rest, timeout_ref, 0)]
+        [fun.(task, nil) | map_many(rest, timeout_ref, 0, fun)]
     after
       timeout ->
-        [{task, nil} | yield_many(rest, timeout_ref, 0)]
+        [fun.(task, nil) | map_many(rest, timeout_ref, 0, fun)]
     end
   end
 
-  defp yield_many([], _timeout_ref, _timeout) do
+  defp map_many([], _timeout_ref, _timeout, _fun) do
     []
   end
```

WDYT?

iangreenleaf:
That was my original plan. What led me away from it was realizing that by calling `receive` one at a time in order, we won't be able to exit as soon as a process crashes. I think that's a worthy goal, right? Otherwise a process could crash immediately and we might still wait 5 seconds for the others to finish before exiting.

josevalim:
> That was my original plan. What led me away from it was realizing that by calling receive one at a time in order, we won't be able to exit as soon as a process crashes. I think that's a worthy goal, right? 

Good call. Let's make sure to test this scenario with a test. You will need to trap exits and have a first task that never finishes followed by a task that fails, all when using the default timeout.

I will also finish the review.

lib/elixir/lib/task.ex
```diff
@@ -608,6 +608,108 @@ defmodule Task do
...
+  """
+  @spec await_many([t], timeout) :: [term]
+  def await_many(tasks, timeout \\ 5000) when is_timeout(timeout) do
+    timeout_ref = make_ref()
+
+    if timeout != :infinity do
+      Process.send_after(self(), timeout_ref, timeout)
+    end
+
+    awaiting = build_awaiting(tasks, %{}, 0)
```
josevalim: `Process.monitor` can fail with noconnected reason, we need to catch it and re-raise. See how we are doing it in yield_many:

```
try do
  build_awaiting(tasks, %{}, 0)
catch
  :exit, :noconnection -> ...
end
```

lib/elixir/lib/task.ex
```diff
@@ -608,6 +608,108 @@ defmodule Task do
...
+
+    await_many(tasks, timeout, awaiting, %{}, timeout_ref)
```
josevalim: We need to make sure the `timer_ref` is cancelled, even in failure cases. You will need to do thiis:

try do
  await_many(tasks, timeout, awaiting, %{}, timeout_ref)
after
  timer_ref && Process.cancel_timer(timer_ref)
end
```

where `timer_ref` is the result of the `if` call above.

File: lib/elixir/lib/task.ex [hunk 5]
```diff
@@ -608,6 +608,108 @@ defmodule Task do
...
+  end
+
+  defp await_many(_tasks, _timeout, map, replies, timeout_ref) when map_size(map) == 0 do
+    Process.demonitor(timeout_ref, [:flush])
```
josevalim: This does not cancel the timer. Please see the comment in the function above. You can fully remove this line.

lib/elixir/lib/task.ex
```diff
@@ -608,6 +608,108 @@ defmodule Task do
...
+    build_replies_list(replies)
```
josevalim: This can be written as:

    for i <- 0..map_size(replies)-1, do: Map.fetch!(replies, i)

lib/elixir/test/elixir/task_test.exs
```diff
@@ -272,6 +272,139 @@ defmodule TaskTest do
     end
   end
 
+  @tag timeout: 10
+  describe "await_many/2" do
+    test "returns replies in input order" do
+      refs = [ref_1 = make_ref(), ref_2 = make_ref(), ref_3 = make_ref()]
```
josevalim: Let's start proper tasks please. There is no reason to mimic tasks otherwise.

lib/elixir/test/elixir/task_test.exs
```diff
@@ -272,6 +272,139 @@ defmodule TaskTest do
     end
   end
 
+  @tag timeout: 10
+  describe "await_many/2" do
```
josevalim: The tests use "fake tasks". Let's avoid doing that. Please start tasks instead of defining refs by hand.

lib/elixir/test/elixir/task_test.exs
```diff
@@ -272,6 +272,139 @@ defmodule TaskTest do
     end
   end
 
+  @tag timeout: 10
```
josevalim: This will set the timeout to 10 milliseconds, which is most likely not what you want. The default is already 30 seconds, you can safely remove it.

lib/elixir/test/elixir/task_test.exs
```diff
@@ -272,6 +272,139 @@ defmodule TaskTest do
...
+    test "ignores messages from other processes" do
+      other_ref = make_ref()
+      refs = [ref_1 = make_ref(), ref_2 = make_ref()]
+      tasks = Enum.map(refs, fn ref -> %Task{ref: ref, owner: self(), pid: nil} end)
+      send(self(), {ref_2, :b})
+      send(self(), other_ref)
+      send(self(), {other_ref, :z})
+      send(self(), {:DOWN, other_ref, :process, 1, :goodbye})
+      send(self(), {ref_1, :a})
+      assert Task.await_many(tasks) == [:a, :b]
```
josevalim: You should assert the spurious messages were not consumed. So do things like `assert_received {:DOWN, ^other_ref, :process, 1, :goodbye}` here. Also, please test what happens if you do `send(self(), {ref_2, :b})` and then `send(self(), {ref_2, :c})` (i.e. the `:c` one should stay in the inbox).

lib/elixir/test/elixir/task_test.exs
```diff
@@ -272,6 +272,139 @@ defmodule TaskTest do
...
+      tasks = [
+        %Task{ref: make_ref(), owner: self(), pid: nil},
+        Task.async(fn -> raise "oops" end)
+      ]
+
+      assert {{%RuntimeError{}, _}, {Task, :await_many, [^tasks, 5000]}} =
+               catch_exit(Task.await_many(tasks))
+    end
+
+    @compile {:no_warn_undefined, :module_does_not_exist}
```
josevalim: You can remove this and the next test, they exist in the other suite for formatting reasons that are not relevant here. :)

lib/elixir/test/elixir/task_test.exs
josevalim: If you need a task that never returns, you can do this: `Task.async(fn -> Process.sleep(:infinity) end)`.

lib/elixir/lib/task.ex
josevalim: Or even better, given you keep a list of all tasks, you don't need to store indexes. Instead, you can do this:

1. Instead of `build_awaiting`, you will do:

    for task <- tasks, into: %{} do
      %Task{ref: ref, owner: owner} = task

      if owner != self() do
        raise ArgumentError, invalid_owner_error(task)
      end

      {ref, true}
    end

2. Now at the end each async_many call, you will do:

```
        await_many(
          tasks,
          timeout,
          Map.delete(awaiting, ref),
          Map.put(replies, ref, reply),
          timeout_ref
        )
```

3. And in here you will do:

```
for %{ref: ref} <- tasks, do: Map.fetch!(replies, ref)
```

josevalim:
Thanks @iangreenleaf, I have finished my review. Let me know if you have any questions!

lib/elixir/test/elixir/task_test.exs
iangreenleaf: How do you recommend I control the order in which the tasks send replies for this first test? I'd like to ensure that the return value is ordered even if the replies are sent out of order. I could give each task a tiny `sleep`, but that feels a little hacky?

iangreenleaf: Heh, and likewise switching to real tasks is giving me trouble in `test "ignores messages from other processes"` because the fake response now arrives before the real one. A `sleep 1` before sending the fake response solves it for me, but I'm not sure that's very robust.

lib/elixir/lib/task.ex
iangreenleaf: Do I also need this line? I wasn't sure what it does:
```
      receive do: (^timeout_ref -> :ok), after: (0 -> :ok)
```

lib/elixir/test/elixir/task_test.exs
josevalim: Ah, for the test that asserts the ordering and for the fake messages in particular, it is ok to use fake messages. But we also need a test that works directly on the tasks. :)

lib/elixir/lib/task.ex
josevalim: Good call, yes you do. That's just in case you receive the timeout_ref after processing all messages.

lib/elixir/lib/task.ex
```diff
@@ -608,6 +608,99 @@ defmodule Task do
...
+      awaiting =
+        for task <- tasks, into: %{} do
+          %Task{ref: ref, owner: owner} = task
+
+          if owner != self() do
+            raise ArgumentError, invalid_owner_error(task)
+          end
+
+          {ref, true}
+        end
```
josevalim: Let's build `awaiting` outside of the `try` and before we compute `timeout_ref` or `timer_ref`. This should reduce the amount of code inside `try`, which is always a bit cleaner.

josevalim:
Just one comment and we can ship it :D

PR #9743, opened 2020-01-16
Adds links and examples to docs around logging

fireproofsocks:
This PR addresses the following:
- fleshes out examples when configuring custom backends
- includes examples of whitelisting metadata fields
- fixes broken link to `Logger.Backends.Console` docs (because it had `@moduledoc false`) : a brief summary was added.
- Corrects a couple small typos

### DIFF
--- a/lib/logger/lib/logger.ex
+++ b/lib/logger/lib/logger.ex
@@ -474,9 +474,9 @@ defmodule Logger do
     * `:format` - the logging format for that backend
     * `:metadata` - the metadata to include in that backend
 
-  Check the [`Logger.Backends.Console`](https://github.com/elixir-lang/elixir/blob/master/lib/logger/lib/logger/backends/console.ex)
-  implementation for examples on how to handle the recommendations in this section and how to process the
-  existing options.
+  Check the `Logger.Backends.Console` implementation in Elixir's codebase
+  for examples on how to handle the recommendations in this section and
+  how to process the existing options.
 
   ### Erlang/OTP handlers
 
--- a/lib/logger/lib/logger.ex
+++ b/lib/logger/lib/logger.ex
@@ -99,7 +99,8 @@ defmodule Logger do
   `:crash_reason`, `:initial_call`, and `:registered_name` are available
   only inside behaviours such as GenServer, Supervisor, and others.
 
-  For example, you might wish to include a custom error code in your logs:
+  For example, you might wish to include a custom `:error_code` metadata in
+  your logs:
 
       Logger.error("We have a problem", [error_code: :pc_load_letter])
 
@@ -108,12 +109,12 @@ defmodule Logger do
   your log format template:
 
       config :logger, :console,
-       format: "$message[$level]$metadata\n",
+       format: "[$level] $message $metadata\n",
        metadata: [:error_code, :file]
 
   Your logs might then receive lines like this:
 
-      We have a problem[error]error_code=pc_load_letter file=lib/app.ex
+      [error] We have a problem error_code=pc_load_letter file=lib/app.ex
 
   ## Configuration
 
@@ -280,7 +281,7 @@ defmodule Logger do
   Backends can also be added dynamically through `add_backend/2`.
 
   For example, to add multiple backends to your application, modify your
-  `config.exs`:
+  configuration:
 
       config :logger,
         backends: [:console, MyCustomBackend]
@@ -473,8 +474,8 @@ defmodule Logger do
     * `:format` - the logging format for that backend
     * `:metadata` - the metadata to include in that backend
 
-  Check `Logger.Backends.Console`'s implementation for examples on how
-  to handle the recommendations in this section and how to process the
+  Check the [`Logger.Backends.Console`](https://github.com/elixir-lang/elixir/blob/master/lib/logger/lib/logger/backends/console.ex)
+  implementation for examples on how to handle the recommendations in this section and how to process the
   existing options.
 
   ### Erlang/OTP handlers
--- a/lib/logger/lib/logger.ex
+++ b/lib/logger/lib/logger.ex
@@ -99,6 +99,22 @@ defmodule Logger do
   `:crash_reason`, `:initial_call`, and `:registered_name` are available
   only inside behaviours such as GenServer, Supervisor, and others.
 
+  For example, you might wish to include a custom error code in your logs:
+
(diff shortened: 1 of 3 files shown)


lib/logger/lib/logger.ex
```diff
@@ -99,6 +99,22 @@ defmodule Logger do
   `:crash_reason`, `:initial_call`, and `:registered_name` are available
   only inside behaviours such as GenServer, Supervisor, and others.
 
+  For example, you might wish to include a custom error code in your logs:
```
whatyouhide: I think it's important to stress that this is **metadata**, otherwise I'm afraid the reader could think that the "error code" is a specific functionality.

```suggestion
  For example, you might wish to include a custom error code metadata in your logs:
```

lib/logger/lib/logger/backends/console.ex
```diff
@@ -1,5 +1,11 @@
 defmodule Logger.Backends.Console do
-  @moduledoc false
+  @moduledoc """
+  This module is the default event handler for `Logger` used to write
+  messages to the console. It implements the
+  [`:gen_event`](https://erlang.org/doc/man/gen_event.html) behaviour.
+  This module can be used as a reference those who need to implement
+  a custom Logger backend.
+  """
```
josevalim: Let's not make it public given the official API to use it is simply `console`. :) We should probably fix the link instead.

lib/logger/lib/logger.ex
```diff
@@ -99,6 +99,22 @@ defmodule Logger do
...
+  `:error_code` key and you would need to include `$metadata` as part of
+  your log format template:
+
+      config :logger, :console,
+       format: "$message[$level]$metadata\n",
+       metadata: [:error_code, :file]
+
+  Your logs might then receive lines like this:
+
+      We have a problem[error]error_code=pc_load_letter file=lib/app.ex
```
josevalim: While this is correct, it is a bit hard to read. What if the format is `[$level] $message $metadata`?

lib/logger/lib/logger.ex
```diff
@@ -263,6 +279,30 @@ defmodule Logger do
   which must be set before the `:logger` application is started.
   Backends can also be added dynamically through `add_backend/2`.
 
+  For example, to add multiple backends to your application, modify your
+  `config.exs`:
```
whatyouhide: I would suggest we just say to modify the "configuration", without referring to a specific file name:

```suggestion
  configuration:
```

lib/logger/lib/logger/backends/console.ex
fireproofsocks: Should it link to the raw file in Github?

lib/logger/lib/logger.ex
fireproofsocks: Sounds good -- I just wanted to show that `$metadata` has to be in the formatting string.

lib/logger/lib/logger.ex
fireproofsocks: What if it read like this:
```
  For example, you might wish to include a custom `:error_code` bit of metadata in your logs:
```
?

lib/logger/lib/logger.ex
fireproofsocks: I specifically wanted to mention the `config.exs` because some of use (cough) got confused with the other configuration functions in the `Logger` module (namely `configure/1` and `configure_backend/2`). 

How about saying :
```
For example, to add multiple backends to your application, modify your `config.exs` or its derivatives:
```

lib/logger/lib/logger/backends/console.ex
josevalim: We should not link to it at all. :) Can you please tell us where the link is?

lib/logger/lib/logger.ex
josevalim: > For example, to add multiple backends to your application, modify your `config/config.exs` or its derivatives:

lib/logger/lib/logger/backends/console.ex
fireproofsocks: I linked to the file...

fireproofsocks: The bad link is on https://hexdocs.pm/logger/Logger.html#module-custom-backends . -- search for "Logger.Backends.Console" and you'll find that ExDoc has auto-created a link to https://hexdocs.pm/logger/Logger.Backends.Console.html but it's a 404.

Coming across a 404 when trying to wrap your head around Logger event handlers is really confusing, as is the fact that this module implements callbacks from ERLANG (!).  I added a link to the Erlang page in the GenEvent module to help alleviate the confusion.

josevalim: Oh, so we can leave your comments as code comments, just not a public documentation. That should help those reading the source. WDYT?

fireproofsocks: I think it will be Ok with the link to the Github file since it's really only developers who would be looking at that.

lib/logger/lib/logger.ex
```diff
@@ -263,6 +280,30 @@ defmodule Logger do
...
+          :console,
+          {MyCustomBackend, :error_backend},
+          {MyCustomBackend, :debug_backend}
+        ]
+
+      config :logger, :error_backend,
+        level: :error
+        # other options
+
+      config :logger, #debug_backend,
```
AndrewDryga:
```suggestion
      config :logger, :debug_backend,
```

lib/logger/lib/logger.ex
fertapric: I think people is going to be more familiar with `"$metadata[$level] $message\n"` since that's what [ships by default with Phoenix](https://github.com/phoenixframework/phoenix/blob/20c80d9b3f661757ad847bf9abbf08bbbb8b8c16/installer/templates/phx_single/config/config.exs#L24).

In addition, I think it reads better because the log level kind of separates the message and the metadata:

```
error_code=pc_load_letter file=lib/app.ex [error] We have a problem 
```

fireproofsocks: I pushed an update.  Again, the emphasis here is on how the metadata argument must line-up with the whitelisted metadata keys.  I don't think the level of confusion about the specifics of the formatting string (or the order of its arguments) is as high.

fertapric:
Thanks @fireproofsocks! 💜

PR #9745, opened 2020-01-17
Move evaluation of default timeout in ExUnit to runtime

### LINKED ISSUE #9744: [ex_unit] Missleading error message when using `{assert,refute}_receive` not in `test/2,3`
### Environment

* Elixir & Erlang/OTP versions (elixir --version): tested with 1.9.3-otp-22 and 1.9.4-otp-22, probably affects earlier versions as well
* Operating system: Arch Linux

### Current behavior

Example code available at GitHub: https://github.com/kbredemeier/test_template_issue

Running `mix test` yields the following compile error:

```plain
Compiling 2 files (.ex)

== Compilation error in file test/support/template.ex ==
** (ArgumentError) could not fetch application environment :assert_receive_timeout for application :ex_unit because the application was not loaded/started. If your application depends on :ex_unit at runtime, make sure to load/start it or list it under :extra_applications in your mix.exs file
    (elixir) lib/application.ex:498: Application.fetch_env!/2
    expanding macro: ExUnit.Assertions.assert_receive/1
    test/support/template.ex:13: Template.test_something/0
```

The proposed fix of adding `:ex_unit` to `:extra_applications` does of course not work.

Instead providing a timeout manually to `assert_receive` does indeed fix the error message or calling `Application.ensure_started(:ex_unit)` during compiletime.

### Expected behavior

I see 2 alternative expactations:

#### Improved error message and docs

The error message and the documentation should suggest to provide a default argument instead when not inside of `test/2,3`.

#### Change the default argument

Change the current default value, such that it is quaranteed to be resolved at runtime instead of compiletime-ish.

devonestes:
There were some confusing errors around this evaluation of the default
timeout at compile time, so now it's been moved to runtime.

Resolves #9744

### DIFF
--- a/lib/ex_unit/lib/ex_unit/assertions.ex
+++ b/lib/ex_unit/lib/ex_unit/assertions.ex
@@ -420,11 +420,7 @@ defmodule ExUnit.Assertions do
       assert_receive {:count, ^x}
 
   """
-  defmacro assert_receive(
-             pattern,
-             timeout \\ Application.fetch_env!(:ex_unit, :assert_receive_timeout),
-             failure_message \\ nil
-           ) do
+  defmacro assert_receive(pattern, timeout \\ nil, failure_message \\ nil) do
     code = escape_quoted(:assert_receive, [], pattern)
     assert_receive(pattern, timeout, failure_message, __CALLER__, code)
   end
@@ -492,13 +488,6 @@ defmodule ExUnit.Assertions do
         end
       end
 
-    timeout =
-      if is_integer(timeout) do
-        timeout
-      else
-        quote do: ExUnit.Assertions.__timeout__(unquote(timeout))
-      end
-
     failure_message =
       failure_message ||
         quote do
@@ -512,7 +501,7 @@ defmodule ExUnit.Assertions do
         end
 
     quote do
-      timeout = unquote(timeout)
+      timeout = ExUnit.Assertions.__timeout__(unquote(timeout), :assert_receive_timeout)
 
       {received, unquote(vars)} =
         receive do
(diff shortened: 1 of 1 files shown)


lib/ex_unit/lib/ex_unit/assertions.ex
```diff
@@ -493,10 +489,10 @@ defmodule ExUnit.Assertions do
       end
 
     timeout =
-      if is_integer(timeout) do
-        timeout
-      else
-        quote do: ExUnit.Assertions.__timeout__(unquote(timeout))
+      case timeout do
```
josevalim: Unfortunately the value can be `nil` at runtime but a var at compilation time, so I think we can always call `ExUnit.Assertions.__timeout__` instead and handle `nil` and defaults in there. WDYT?

devonestes: Yeah, that works better.

lib/ex_unit/lib/ex_unit/assertions.ex
```diff
@@ -529,9 +518,11 @@ defmodule ExUnit.Assertions do
...
 
   @doc false
-  def __timeout__(timeout) when is_integer(timeout) and timeout >= 0, do: timeout
+  def __timeout__(timeout, _) when is_integer(timeout) and timeout >= 0, do: timeout
 
-  def __timeout__(timeout),
+  def __timeout__(nil, key), do: Application.fetch_env!(:ex_unit, key)
+
+  def __timeout__(timeout, _),
     do: raise(ArgumentError, "timeout must be a non-negative integer, got: #{inspect(timeout)}")
```
fertapric: Now that we have the key in here, what do you think about including it in the error message?

devonestes: In what way would we use that key? Something like "timeout must be a non-negative integer, got #{timeout} - set this default in ex_unit.#{key}"?

I'm not totally sure that would be helpful for folks, but I'm happy to include it if y'all think it would be a good thing.

josevalim: Yeah, I don't see how we can use the key here given that we use it only when a timeout is not given?

fertapric: I thought we could use it for saying something like:

> assert_receive timeout must be a non-negative integer, got: ...

But nevermind, probably not worth it

fertapric:
Thanks @devonestes! 💜

PR #9750, opened 2020-01-23
Use Github actions e-mail

josevalim:


.github/workflows/ci_email.yml
```diff
@@ -5,13 +5,5 @@ jobs:
...
-      uses: docker://fertapric/elixir-ci-email:latest
-      env:
-        APP_NAME: Cirrus CI
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        MAIL_FROM: ci@elixir-lang.org
-        MAIL_HOST: smtp.sendgrid.net
-        MAIL_PASSWORD: ${{ secrets.CI_EMAIL_PASSWORD }}
-        MAIL_USERNAME: ${{ secrets.CI_EMAIL_USERNAME }}
+    - run: false
+      if: github.event.action != 'neutral' && github.event.action != 'success'
```
whatyouhide: Missing final newline :)

```suggestion
      if: github.event.action != 'neutral' && github.event.action != 'success'

```

fertapric: Does this work? It looks like `action` can be `completed`, `requested`, or `rerequested` https://developer.github.com/v3/activity/events/types/#checksuiteevent

Maybe `github.event.check_suite[conclusion]`?

josevalim: Unfortunately the check did not run from a branch but I think you are rigjt. It is either `github.event.check_suite.conclusion` or `github.event.conclusion`. I will move to `.conclusion` and push it to master and see how it goes.

PR #9754, opened 2020-01-27
Allow users to tag doctests with options

devonestes:
Now users can pass tags as options to doctests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,10 +15,6 @@
   * [Kernel] Add `is_struct/2` guard
   * [Kernel] Support `map.field` syntax in guards
 
-#### ExUnit
-
-  * [DocTest] Accept `:tags` as an option to `doctest/2`
-
 ### 2. Bug fixes
 
 ### 3. Soft-deprecations (no warnings emitted)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@
   * [Kernel] Add `is_struct/2` guard
   * [Kernel] Support `map.field` syntax in guards
 
+#### ExUnit
+
+  * [DocTest] Accept `:tags` as an option to `doctest/2`
+
 ### 2. Bug fixes
 
 ### 3. Soft-deprecations (no warnings emitted)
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,10 @@
   * [Kernel] Add `is_struct/2` guard
   * [Kernel] Support `map.field` syntax in guards
 
+#### ExUnit
```
fertapric: We usually ask to not change the CHANGELOG because it may have conflicts when merging 🙂

lib/ex_unit/lib/ex_unit/doc_test.ex
```diff
@@ -208,6 +211,7 @@ defmodule ExUnit.DocTest do
 
         for {name, test} <- ExUnit.DocTest.__doctests__(module, opts) do
           @file file
+          @tag Keyword.get(opts, :tags, [])
```
josevalim: This is a micro optimization but I think we should do this:

```
tags =
  if tags = Keyword.get(opts, :tags, []) do
    quote do: @tag unquote(tags)
  end
```

and then in here you do:

```
unquote(tags)
```

So we don't end-up calling the module attribute with nothing most of the times.

devonestes: I tried pulling it out like that, but because of the `bind_quoted` I couldn't unquote in this `quote` block, and then of course I couldn't use `bind_quoted` with the AST for `@tag unquote(tags)` because the binding was inserting that AST in the resulting output in the wrong place, so I ended up going with this in the quote block with the already bound `opts`:

```
          if tags = Keyword.get(opts, :tags) do
            @tag tags
          end
```

That way it doesn't call the module attribute every time, but it still works.

Let me know if there's a better way to handle this sort of thing!

CHANGELOG.md
devonestes: Got it - removed!

lib/ex_unit/lib/ex_unit/doc_test.ex
```diff
@@ -187,6 +187,9 @@ defmodule ExUnit.DocTest do
       should be set to `false` and a full `Module.function` construct should be
       used.
 
+    * `:tags` - a list of tags to apply to the generated tests. All tests generated
+       will have all of the given tags applied to them.
```
wojtekmach: the 2nd sentence seems redundant.

(btw, I love this addition!)

josevalim:

dkuku:
@josevalim I'm going through your aoc solutions and you used doctests a lot at the beginning.
 I think there should be a default tag for doctests. I looked at the docs and source code and it seems that specifying it per file is the only way.

PR #9757, opened 2020-01-28
Typo fix in changelog

theodowling:
`against` instead of `agaainst`


fertapric:
Thanks @theodowling! 💜

PR #9760, opened 2020-01-29
Fix docs for compiler tracing events

devonestes:
The `local_function` event is a tuple with four elements, not five.


josevalim:

PR #9761, opened 2020-01-29
Now it's possible again to use remote or rpc with compile_env

esse:
New feature: compile_env introduced a problem. While using it, it was
impossible to use remote/rpc on a running system (due to issue with
config providers).

That commit fixes it, together with associated test (failing before)

### DIFF
--- a/lib/mix/lib/mix/release.ex
+++ b/lib/mix/lib/mix/release.ex
@@ -492,9 +492,9 @@ defmodule Mix.Release do
   The boot script uses the RELEASE_LIB environment variable, which must
   be accordingly set with `--boot-var` and point to the release lib dir.
   """
-  @spec make_boot_script(t, Path.t(), [{application(), mode()}], [String.t()]) ::
+  @spec make_boot_script(t, Path.t(), [{application(), mode()}], [String.t()], atom()) ::
           :ok | {:error, String.t()}
-  def make_boot_script(release, path, modes, prepend_paths \\ []) do
+  def make_boot_script(release, path, modes, prepend_paths \\ [], boot_name \\ :start) do
     with {:ok, rel_spec} <- build_release_spec(release, modes) do
       File.write!(path <> ".rel", consultable(rel_spec), [:utf8])
 
@@ -515,7 +515,7 @@ defmodule Mix.Release do
 
           instructions =
             instructions
-            |> post_stdlib_applies(release)
+            |> post_stdlib_applies(release, boot_name)
             |> prepend_paths_to_script(prepend_paths)
 
           script = {:script, rel_info, instructions}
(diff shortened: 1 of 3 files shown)


josevalim:
Hi @esse! Unfortunately the test does not fail if I remove the fix. Can you please tell us a bit more how exactly you are running into this error? Thank you.

esse:
@josevalim are you sure?

If I remove the fix, I'm getting an error:
```
ERROR! the application :release_test has a different value set for key :static during runtime compared to compile time. Since this application environment entry was marked as compile time, this difference can lead to different behaviour than expected:

  * Compile time value was set to: :was_set
  * Runtime value was not set

To fix this error, you might:

  * Make the runtime value match the compile time one

  * Recompile your project. If the misconfigured application is a dependency, you may need to run "mix deps.compile release_test --force"

  * Alternatively, you can disable this check. If you are using releases, you can set :validate_compile_env to false in your release configuration. If you are using Mix to start your system, you can pass the --no-validate-compile-env flag


init terminating in do_boot ({,[{Elixir.Config.Provider,boot,2,[]}]})

Crash dump is being written to: erl_crash.dump...done


  1) test allows you to connect to remote system while using compile_env (Mix.Tasks.ReleaseTest)
     test/mix/tasks/release_test.exs:513
     Assertion with == failed
     code:  assert System.cmd(script, ["rpc", "ReleaseTest.hello_world()"]) == {"hello world\n", 0}
     left:  {"{\"init terminating in do_boot\",{<<\"aborting boot\">>,[{'Elixir.Config.Provider',boot,2,[]}]}}\r\n", 1}
     right: {"hello world\n", 0}
     stacktrace:
       test/mix/tasks/release_test.exs:535: anonymous fn/1 in Mix.Tasks.ReleaseTest."test allows you to connect to remote system while using compile_env"/1
       (mix 1.11.0-dev) lib/mix/project.ex:352: Mix.Project.in_project/4
       (elixir 1.11.0-dev) lib/file.ex:1544: File.cd!/2
       test/test_helper.exs:124: MixTest.Case.in_fixture/3
       test/mix/tasks/release_test.exs:514: (test)
```

let me push for a second, version with only test, without a fix so we can see that on the CI.

josevalim:
@esse Thanks, I will try again. In any case, can you please tell us about the use case anyway? Where is your config values coming from?

esse:
@josevalim I've build a micro-example of an app that shows the problem:
https://github.com/esse/compile_env_issue

esse:
@josevalim and CI failed on that test:
https://github.com/elixir-lang/elixir/pull/9761/checks?check_run_id=416858402

josevalim:
@esse thanks for the app. I can definitely reproduce the issue there. Unfortunately the test still does not fail on my machine, but that seems to a local problem. I will continue investigating. Thank you!

esse:
@josevalim it seems to be passing on Windows: https://github.com/elixir-lang/elixir/pull/9761/checks?check_run_id=416858396
interesting.
Anyway, I reverted that commit so it should be good to go.

josevalim:
So the issue is that we don't run config providers for rpc/remote. Therefore, I believe the proper fix would be to tie the running of the compile env validation with the config providers, instead of hardcoding it at the config script level. I am trying out some ideas and I will let you know.

esse:
@josevalim if you have other idea how this should be done - let me know, I'm more than happy to help :)

josevalim:
Hi @esse ! Unfortunately it is one of those ideas that you only know for sure once you play with it a bit, so please see #9762.  :) Thanks for the different ways to reproduce the issue!

esse:
sure, that makes sense. thanks :)

btw - if you have any other issues that can be more accessible, I would be also very happy to help :)

josevalim:
Sure! All of our open issues are listed on GitHub, so feel free to explore them! :)

josevalim:
Closing in favor of the new PR!

Issue #9465, opened 2019-11-01
Further unify logger and Logger


hauleth:
- [x] Convert usage of `:error_logger` to `:logger`
- [x] Migrate `:console` backend to be defined as a formatter for `logger_std_h` - this will allow to use the same format as Elixir is using for other handlers as well (requires erlang/otp#2491 - OTP 24+ and erlang/otp#2523 - OTP 23+)
- [x] Only attach built-in `Logger.Handler` if there will be backends (depends on the above)
- [x] Expose API for module and application level verbosity control (#9793)
- [x] Expose and document all of Erlang built-in logger levels (#9791)
- [x] Add support for passing map as a "message" to the `Logger` macros to allow structured logging (needs to be fully discussed before) (#9797)
- [x] Use `time` field from metadata instead of generating new timestamp (#9772)
- [x] Make `Logger.Formatter` works also as a Erlang's logger formatter for use with Erlang's logger handlers
- [x] Make translators a filter
- [x] Document usage and configuration of Erlang's handlers with Elixir's `Logger` (such as the file handler with rotation)
- [x] Add API to set the global metadata (OTP 24+) (#12413)
- [x] Expose `:logger.add_handlers/1` (with an atom) (#12412)
- [x] Rewrite `capture_log` to use Erlang handlers (#12408)
- [x] Publish and CI [`:logger_backends`](https://github.com/elixir-lang/logger_backends) with elixir: ~> 1.15 (https://hex.pm/packages/logger_backends/1.0.0-rc.0)

josevalim:
Note: don't send PRs for those yet. We will probably wait until the Logger integration stabilizes a bit before going all in.

hauleth:
Issues to track:

- erlang/otp#2444 - accept binaries in entries using format
- erlang/otp#2442 - allow ignoring messages from within the log function (by returning `:ignore`)
- erlang/otp#2491 - allow using binaries in `:logger_formatter` template to simplify integration with Elixir

hauleth:
@josevalim I slowly start working on these, so it would be at least partially ready for 1.11.

josevalim:
Perfect, the earlier the better!

hauleth:
Another one PR to core erlang/otp#2523 - this one is to fully support `:device` option in `logger_std_h`.

vans163:
Can this get tagged on too, https://bugs.erlang.org/browse/ERL-1159.

Basically to not push up the REPL prompt on Logger outputs.

[...]

PR #9772, opened 2020-01-31
Use Erlang's logger timestamp

hauleth:
Instead of generating new timestamp on each log message use existing `time` entry from `:logger.metadata()`.

Ref #9465


josevalim:
Please look at CI. Most likely a formatter issue. :)

hauleth:
Yeah, I have different view on "readable" than formatter.

josevalim:

Issue #9746, opened 2020-01-20
Make `recompile` rebuild whenever something needs to be rebuilt in the dependency tree


mguimas:
### Environment

```
$ elixir --version
Erlang/OTP 22 [erts-10.5] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1] [hipe]

Elixir 1.9.1 (compiled with Erlang/OTP 22)
```

### Current behavior

When calling `recompile` inside IEx, after changing the source files of a project included as a dependency in the project loaded via `iex -S mix`, does not rebuild the dependency project, but instead returns `:noop`. For example, after changing one source file in a dependency project
 
```
iex(18)> recompile
:noop
```

In this case the developer needs to quit IEx, and restart it to have the dependency recompile, like this:

```
iex(19)>
BREAK: (a)bort (c)ontinue (p)roc info (i)nfo (l)oaded
       (v)ersion (k)ill (D)b-tables (d)istribution
$ iex -S mix
Erlang/OTP 22 [erts-10.5] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1] [hipe]

==> dependency
Compiling 1 file (.ex)
Generated dependency app
==> main
Compiling 3 files (.ex)
Interactive Elixir (1.9.1) - press Ctrl+C to exit (type h() ENTER for help)
iex(1)>
```

This behavior is boring, and causes the loss of all bindings and their state in the IEx session.

### Expected behavior

The `recompile` command should detect when source files (and others with influence in the build result) in the dependency tree of the project loaded via `iex -S mix` have changed and need to be rebuilt, and build them.

The dependency projects I used to test are all local and included in the `main` project as `      {:dependency, in_umbrella: true}`, but `recompile` should rebuild even when dependency projects are updated from the network, for example, after issuing a `mix deps` command or similar.

Probably `recompile` needs to do the same checks that `iex -S mix` does at start to rebuild what needs to be rebuilt.

Note that today's workaround to all this is to run `recompile force: true` but this rebuilds what does not need to be rebuilt, thus being a time waste and disruptive of development flow.

josevalim:
Unfortunately this is far from trivial to implement. The issue is that the current application obviously depends on its dependencies, which may form a tree, so we would need to shut down all of the dependencies, in the proper order, and then restart them again - while busting any cache along the way. The whole process is complex and error prone, which is why for compiling dependencies, it is simpler to just start from scratch.

[...]

PR #9774, opened 2020-02-02
Handle local deps change on recompile() in iex

esse:
This PR fixes issue #9746.

I'm not sure if this is the right approach or is there are tests for the recompile feature - let me know :)

Example behaviour after this commit:
```
$ iex -S mix
Erlang/OTP 22 [erts-10.6.1] [source] [64-bit] [smp:12:12] [ds:12:12:10] [async-threads:1] [hipe]

==> dep2
Compiling 1 file (.ex)
Interactive Elixir (1.11.0-dev) - press Ctrl+C to exit (type h() ENTER for help)
iex(1)> recompile
:noop
iex(2)> Dep2.hello
:worldXZXXZZZZ2
iex(3)> recompile
==> dep2
Compiling 1 file (.ex)
:ok
iex(4)> Dep2.hello
:w
iex(5)>
```

### DIFF
--- a/lib/mix/lib/mix/task.ex
+++ b/lib/mix/lib/mix/task.ex
@@ -428,6 +428,7 @@ defmodule Mix.Task do
     recursive = (module = get(task)) && recursive(module)
 
     Mix.TasksServer.delete_many([{:task, task, proj}, {:alias, task, proj}])
+    Mix.TasksServer.delete_task_for_all_projects({:task, task})
 
     cond do
       recursive && Mix.Project.umbrella?() ->
--- a/lib/iex/lib/iex/helpers.ex
+++ b/lib/iex/lib/iex/helpers.ex
@@ -122,6 +122,9 @@ defmodule IEx.Helpers do
     Mix.Task.reenable("compile")
     Mix.Task.reenable("compile.all")
     Mix.Task.reenable("compile.protocols")
+    Mix.Task.reenable("deps.loadpaths")
+    Mix.Task.reenable("deps.precompile")
+    Mix.Task.reenable("loadpaths")
     compilers = config[:compilers] || Mix.compilers()
     Enum.each(compilers, &Mix.Task.reenable("compile.#{&1}"))
   end
--- a/lib/mix/lib/mix/tasks_server.ex
+++ b/lib/mix/lib/mix/tasks_server.ex
@@ -25,6 +25,21 @@ defmodule Mix.TasksServer do
     Agent.update(@name, &Map.drop(&1, many), @timeout)
   end
 
+  def delete_task_for_all_projects({:task, task}) do
+    Agent.update(
+      @name,
+      fn map ->
+        map
+        |> Enum.reject(fn
+          {{:task, ^task, _}, _} -> true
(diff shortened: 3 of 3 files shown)


lib/mix/lib/mix/task.ex
```diff
@@ -428,6 +428,7 @@ defmodule Mix.Task do
     recursive = (module = get(task)) && recursive(module)
 
     Mix.TasksServer.delete_many([{:task, task, proj}, {:alias, task, proj}])
+    Mix.TasksServer.delete_task_for_all_projects({:task, task})
```
josevalim: I am afraid this change will have other consequences, as I really don't expect reenabling to work across all projects. In particular, we want to reenable only the local projects, and not all of the dependencies.

I am afraid the best way to solve this issue is to add a function like: `Mix.Project.within_local_projects(fn project_config -> ... end)` which will execute within all local projects, including the current project itself, and then we run `compile` inside of each them.

However, implementing this task is far from trivial, so I am currently inclined to mark the original issue as unresolved. :/

Thank you very much for looking into this! As we can only know how hard it is to provide it once we give it a try.

esse:
@josevalim I can propose then more straightforward workaround - which seems to be working good (I've tested it on a pretty big project, with many umbrella apps):
* we add a new helper to iex, called recompile_with_dep/1, getting in an atom (or list of atoms) as a first argument, for example, `Dep2` - representing dependency to be recompiled.

Then the code of this helper would be more or less:
```
tasks = [
  {:task, "compile.all", Dep2.MixProject},
  {:task, "compile.leex", Dep2.MixProject},
  {:task, "compile.yecc", Dep2.MixProject},
  {:task, "compile.erlang", Dep2.MixProject},
  {:task, "compile", Dep2.MixProject},
  {:task, "compile.elixir", Dep2.MixProject},
  {:task, "compile.app", Dep2.MixProject},
  {:alias, "compile.all", Dep2.MixProject},
  {:alias, "compile.leex", Dep2.MixProject},
  {:alias, "compile.yecc", Dep2.MixProject},
  {:alias, "compile.erlang", Dep2.MixProject},
  {:alias, "compile", Dep2.MixProject},
  {:alias, "compile.elixir", Dep2.MixProject},
  {:alias, "compile.app", Dep2.MixProject}
]

Mix.Task.reenable("deps.loadpaths")
Mix.Task.reenable("deps.precompile")
Mix.Task.reenable("loadpaths")

Mix.TasksServer.delete_many(tasks)

recompile()
```

Which is not perfect, for obvious reasons - but allows you to recompile external dependency without exiting iex and losing bindings etc. I've tested it also with Pheonix app - it will enable you to access recompiled deps in Phoenix without restarting it, which is pretty cool.

josevalim:
Unfortunately I am afraid there are gotchas hidden behind the scenes and, if we say we support the feature, then we will be obliged to force and handle all of these gotchas.

esse:
@josevalim sure, I understand.

I've created hex package with it, if someone is feeling adventurous enough ;)
https://github.com/esse/recompile_dep
https://hex.pm/packages/recompile_dep

PR #9775, opened 2020-02-02
Escape Unicode chars in doucmentation for String codepoints/1

gsarwate:
Hexdocs for String.codepoints/1 has 5 examples. 

Fourth and fifth are different but in documentation the look the same. 

String.codepoints("é")
String.codepoints("é")

Unicode chars need to be escaped to distinguish them as intended in the source

String.codepoints("\u00e9")
String.codepoints("\u0065\u0301")


josevalim:

PR #9776, opened 2020-02-02
Optimize pattern match of access module's func

MasashiYokota:
I've refactored a function of Access module based on [this guideline about pattern matching](https://erlang.org/doc/efficiency_guide/functions.html#pattern-matching). 

I'm glad if you review this patch. Thank you.


josevalim:

PR #9784, opened 2020-02-05
Warn on use of non-match argument to with/1 (#9764)

vanvoljg:


lib/elixir/src/elixir_clauses.erl
```diff
@@ -401,6 +405,8 @@ format_error({try_with_only_else_clause, Origin}) ->
 
 format_error(unmatchable_else_in_with) ->
   "\"else\" clauses will never match because all patterns in \"with\" will always match";
+format_error(non_match_argument_in_with) ->
+  "\"with\" arguments should be matches, not statements";
```
josevalim:
```suggestion
  "\"with\" requires match expressions to be given either as \"left <- right\" or \"left = right\", "
  "please assign your expression to underscore if you want to simply execute some code and discard its value"
```

WDYT?

vanvoljg: Looks great, thanks!

I just realized I forgot to trap the warning output of my test. I'll use this new warning text and push a new commit.

lib/elixir/test/elixir/kernel/special_forms_test.exs
```diff
@@ -68,4 +68,14 @@ defmodule Kernel.SpecialFormsTest do
       end
     end
   end
+
+  describe "with" do
```
josevalim: We have a "lib/elixir/test/elixir/kernel/warning_test.exs" file, we should most likely move it there. :)

vanvoljg: I'll move it there. :-)

vanvoljg:
I'm sorry about that last push. I'm tired and didn't watch my console output.

lib/elixir/src/elixir_clauses.erl
```diff
@@ -401,6 +405,9 @@ format_error({try_with_only_else_clause, Origin}) ->
 
 format_error(unmatchable_else_in_with) ->
   "\"else\" clauses will never match because all patterns in \"with\" will always match";
+format_error(non_match_argument_in_with) ->
+  "\"with\" requires match expressions to be given either as \"left <- right\" or \"left = right\", "
+  "please assign your expression to underscore if you want to simply execute some code and discard its value";
```
whatyouhide:
```suggestion
  "please assign your expression to _ if you want to simply execute some code and discard its value";
```

lib/elixir/test/elixir/kernel/warning_test.exs
```diff
@@ -1786,6 +1786,17 @@ defmodule Kernel.WarningTest do
              "extra parentheses on a remote function capture &System.pid()/0 have been deprecated. Please remove the parentheses: &System.pid/0"
   end
 
+  test "\"with\" warns when given an argument which is not a match" do
+    assert capture_err(fn ->
+             Code.eval_string("""
+                  with fn -> true end do
+                    :warn
+                  end
```
whatyouhide: Can we indent this at the same level as the call to `Code.eval_string/1`? Not a pressing issue but would be nicer :)

lib/mix/lib/mix/dep/lock.ex
```diff
@@ -15,7 +15,7 @@ defmodule Mix.Dep.Lock do
     opts = [file: lockfile, warn_on_unnecessary_quotes: false]
 
     with {:ok, contents} <- File.read(lockfile),
-         assert_no_merge_conflicts_in_lockfile(lockfile, contents),
+         _ <- assert_no_merge_conflicts_in_lockfile(lockfile, contents),
```
whatyouhide:
```suggestion
         _ = assert_no_merge_conflicts_in_lockfile(lockfile, contents),
```

lib/elixir/src/elixir_clauses.erl
josevalim: Let's do both?

```suggestion
  "please assign your expression to "_" (underscore) if you want to simply execute some code and discard its value";
```

whatyouhide:
Thank you @vanvoljg! 💟

PR #9790, opened 2020-02-06
Fix unused variable warnings

wojtekmach:
This needs to be reviewed carefully, in some tests the intention was to
ignore the variable but in some to match against an existing one. I
tried my best to figure out which was which.

There is one thing left which is a false positive for this test:

    test "greets the world" do
      assert <<len::8, _::size(len)-binary>> = <<3, "foo">>
    end

Produces this:

    warning: variable "len" is unused (if the variable is not meant to be used, prefix it with an underscore)
      test/foo_test.exs:6: FooTest."test greets the world"/1

I wasn't able to fix this so any help appreciated!


josevalim:

PR #9791, opened 2020-02-06
Expose all logger levels

hauleth:
Ref #9465

### DIFF
--- a/lib/logger/lib/logger.ex
+++ b/lib/logger/lib/logger.ex
@@ -843,7 +843,7 @@ defmodule Logger do
     :logger.macro_log(%{}, level, chardata, add_elixir_domain(metadata))
   end
 
-  # # TODO: Remove that in Elixir 2.0
+  # TODO: Remove that in Elixir 2.0
   def __do_log__(level, other, metadata) do
     IO.warn("passing #{inspect(other)} to Logger is deprecated, expected a binary or an iolist")
     :logger.macro_log(%{}, level, to_string(other), add_elixir_domain(metadata))
--- a/lib/logger/lib/logger.ex
+++ b/lib/logger/lib/logger.ex
@@ -525,11 +525,11 @@ defmodule Logger do
   been started, which means the configuration above would have no effect.
   """
 
-  @type level :: :error | :warn | :info | :debug
+  @type level :: :emergency | :alert | :critical | :error | :warning | :warn | :notice | :info | :debug
   @type backend :: :gen_event.handler()
   @type message :: IO.chardata() | String.Chars.t()
   @type metadata :: keyword()
-  @levels [:error, :warn, :info, :debug]
+  @levels [:emergency, :alert, :critical, :error, :warning, :notice, :info, :debug]
 
   @metadata :logger_enabled
   @compile {:inline, enabled?: 1}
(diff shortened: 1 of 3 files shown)


lib/logger/test/logger_test.exs
```diff
@@ -89,14 +89,35 @@ defmodule LoggerTest do
...
+    assert Logger.level() == :error
+
+    Logger.configure(level: :alert)
+    assert Logger.level() == :error
+
+    Logger.configure(level: :emergency)
+    assert Logger.level() == :error
+
+    Logger.configure(level: :none)
+    assert Logger.level() == :error
```
hauleth: This behaviour can be a little bit confusing, but I do not know how to handle it as otherwise it would be a breaking change.

lib/logger/lib/logger.ex
```diff
@@ -971,6 +965,8 @@ defmodule Logger do
     Enum.any?(matching, fn filter ->
       Enum.all?(filter, fn
         {:level_lower_than, min_level} ->
+          min_level = if min_level == :warn, do: :warning, else: min_level
```
josevalim: Please use `Logger.Handler.elixir_level_to_erlang_level(level)` instead. Once we deprecate it, we just need to deprecate it there.

lib/logger/test/logger_test.exs
josevalim: I think we should include `:all` and `:none` as official levels and make `Logger.level` simply return the Erlang level directly.

lib/logger/test/logger_test.exs
```diff
@@ -193,8 +214,13 @@ defmodule LoggerTest do
...
+    assert Logger.compare_levels(:debug, :notice) == :lt
     assert Logger.compare_levels(:debug, :warn) == :lt
+    assert Logger.compare_levels(:debug, :warning) == :lt
     assert Logger.compare_levels(:debug, :error) == :lt
+    assert Logger.compare_levels(:debug, :critical) == :lt
+    assert Logger.compare_levels(:debug, :alert) == :lt
+    assert Logger.compare_levels(:debug, :emergency) == :lt
 
     assert Logger.compare_levels(:info, :debug) == :gt
     assert Logger.compare_levels(:info, :info) == :eq
```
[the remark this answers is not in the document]
josevalim: Yes, I don't think we need to test all combinations.

lib/logger/lib/logger.ex
hauleth: TBH unfortunately I cannot, as this will prevent using `level_lower_than: :emergency` for user that will decide to use new logger levels and I wanted to leave that option to the user, so I will only treat `:warn` as it was earlier.

It is not breaking change, as earlier there was no option to use levels above `:error` in Elixir, so it was not possible to define message that would omit `level_lower_than: :error`.

josevalim: `Logger.Handler.elixir_level_to_erlang_level(level)` simply converts :warn to :warning, everything else is returned as is. Or am I missing something?

lib/logger/test/logger_test.exs
hauleth: What about `:warn`? It would change behaviour (however I do not know if there is anything relying on specific value being returned from `Logger.level()`).

lib/logger/test/logger_test.exs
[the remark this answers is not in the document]
hauleth: So I added just few, and I will omit the rest.

lib/logger/lib/logger.ex
hauleth: Ah, yeah. I have forgot. Changed.

hauleth:
@josevalim I have documented `:none` and `:all` "meta-levels" as well.

lib/logger/test/logger_test.exs
josevalim: I think it is fine to change `:warn` because if anyone was matching on Logger.level and expecting specific results, they need to handle new outputs now anyway.

lib/logger/lib/logger.ex
```diff
@@ -971,6 +982,8 @@ defmodule Logger do
     Enum.any?(matching, fn filter ->
       Enum.all?(filter, fn
         {:level_lower_than, min_level} ->
+          min_level = Logger.Handler.elixir_level_to_erlang_level(min_level)
```
josevalim: Wait, compare_levels already calls `elixir_level_to_erlang_level`. So this line is probably not necessary at all?

hauleth: Yeah, I think that you are right. It shouldn't be a case there

josevalim:

PR #9792, opened 2020-02-07
Improve do_expand_dot private function of path module

MasashiYokota:
I've fixed the `do_expand_dot` private function of the `path` module.
I would be glad if you review this PR. Thank you.

### DIFF
--- a/lib/elixir/lib/path.ex
+++ b/lib/elixir/lib/path.ex
@@ -710,8 +710,8 @@ defmodule Path do
   defp do_expand_dot(path), do: do_expand_dot(:binary.split(path, "/", [:global]), [])
   defp do_expand_dot([".." | t], [_, _ | acc]), do: do_expand_dot(t, acc)
   defp do_expand_dot([".." | t], []), do: do_expand_dot(t, [])
-  defp do_expand_dot(["." | t], acc), do: do_expand_dot(t, acc)
-  defp do_expand_dot([h | t], acc), do: do_expand_dot(t, ["/", h | acc])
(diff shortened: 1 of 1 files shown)


ryanwinchester:
When you say "fixed", does that mean it was broken?

Then I think if it _was_ broken and still passing tests, maybe you should add test(s) that would fail without your fix?

MasashiYokota:
No. I'm sorry to confuse you. My expression was a bit wrong. This patch is for just only refactoring the `do_expand_dot` function based on [this guide](https://erlang.org/doc/efficiency_guide/functions.html#pattern-matching). 
There are no bugs as far as I know. 

Thank you.

josevalim:
Hi @MasashiYokota, afaik the compiler cannot reorganize based on guards, so this effectively makes no difference. In any case, thanks!

MasashiYokota:
@josevalim 
Thank you for your comment. As embarrassing as this is to say, I didn't know that.

I've re-refactored the function. How about the below code?
``` elixir
defp do_expand_dot(["." | t], [_, _] = acc), do: do_expand_dot(t, acc)
defp do_expand_dot(["." | t], []), do: do_expand_dot(t, [])
defp do_expand_dot([h | t], [_, _] = acc), do: do_expand_dot(t, ["/", h | acc])
defp do_expand_dot([h | t], []), do: do_expand_dot(t, ["/", h])
```

josevalim:
I Iam not sure if it is equivalent. `[_, _]` will only match a list of two elements. But generally speaking, I don't think you should worry about these improvements in general. If they provide any gains, they will be very very very minor.

MasashiYokota:
OK. I understood. Thank you so much.

PR #9793, opened 2020-02-07
Expose functions for setting per-module levels

hauleth:
Ref #9465

### DIFF
--- a/lib/logger/lib/logger.ex
+++ b/lib/logger/lib/logger.ex
@@ -708,14 +708,14 @@ defmodule Logger do
       end
 
       Logger.configure(level: :error)
-      Logger.set_module_level(Foo, :all)
+      Logger.put_module_level(Foo, :all)
 
       Foo.log()
       # This will print the message even if global level is :error
 
   """
-  @spec set_module_level(module(), level() | :all | :none) :: :ok | {:error, term()}
-  defdelegate set_module_level(mod, level), to: :logger
+  @spec put_module_level(module(), level() | :all | :none) :: :ok | {:error, term()}
+  defdelegate put_module_level(mod, level), to: :logger, as: :set_module_level
 
   @doc """
   Get logging level for given module.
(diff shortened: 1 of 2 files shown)


josevalim:
I like this PR a lot. My only concern is with the name. We don't use the `set` prefix in Elixir. We could use `put_module_level` and `get_module_level` but I find they read a bit weird for now. Any suggestions are welcome.

lib/logger/lib/logger.ex
```diff
@@ -693,6 +693,53 @@ defmodule Logger do
...
+  """
+  @spec get_module_level(module() | [module()]) :: [{module(), level() | :all | :none}]
+  defdelegate get_module_level(mod), to: :logger
+
+  @doc """
+  Set logging level for all modules in given application.
+
+  This is equivalent of doing:
+
+      for mod <- :application.get_key(app, :modules) do
```
josevalim:
```suggestion
      for mod <- Application.spec(app, :modules) do
```

josevalim:
Another option is to not add those to Elixir and simply have a section in the module docs explaining how we can use this feature directly from the Erlang API. But at the same time, I think those are important functions to have.

hauleth:
I have mixed feeling. Some people have some aversion to using Erlang functions in Elixir, but on the other hand, these functions are pretty straightforward and I do not know if it is really needed to have them delegated. It is up to Elixir Core team to decide.

hauleth:
@josevalim alternatively use `Logger.module_level/1` for getting module level and `Logger.module_level/2` for setting it. The problem would be with the application level. Also I prefer more descriptive names like used there.

lib/logger/lib/logger.ex
```diff
@@ -693,6 +693,53 @@ defmodule Logger do
     :gen_event.sync_notify(Logger, :flush)
   end
 
+  @doc """
+  Set logging level for given module.
+
+  This will take priority over the primary level set, so it can be
+  used to increase verbosity of some parts of the project.
```
whatyouhide:
```suggestion
  used to increase or decrease verbosity of some parts of the project.
```

hauleth: I used only one, as I assumed that this was obvious that it can be used to decrease verbosity, what is less obvious is that it can also be used to increase it above the primary level.

Eiji7:
> @hauleth I have mixed feeling. Some people have some aversion to using Erlang functions in Elixir, but on the other hand, these functions are pretty straightforward and I do not know if it is really needed to have them delegated. It is up to Elixir Core team to decide.

I do not have any aversion, but I found by myself that I'm not reading `Erlang` documentation as often as `Elixir`.

When reading what you wrote firstly I think that it would be (somehow) "nice" (maybe good words for that would be documentation consistency?) to have functions delegated, but thinking about it more I do not like delegating functions just for delegating them …

I know that it's not a good place for sending proposals, but just following what you wrote I would much more like code below than standard delegation …

```elixir
defmodule Example do
  @moduledoc """
  (…) `ref=:logger.get_module_level/1` (…)
  """
end
```

Having `ref=` in documentation could cause adding `Referenced functions` which would be added to `Summary` section right after `Functions` sub-section (or specific function description if used inside `@doc`). I believe that having such extra list would increase `Erlang` documentation popularity as people would see those really similar to regular links in let's say `Wikipedia`.

What do you think about it?

josevalim:
Let’s please not get sidetracked (regards of its merits). Let’s discuss it based on the features we have today and if it doesn’t make the cut, then we can rediscuss it later when the landscape changes (for whatever reason, including improvements to the docs).

hauleth:
@Eiji7 in theory you can use `:deps` entry in the `:docs` configuration for handling that. Unfortunately it doesn't seem to support Erlang docs. But I agree with @josevalim, let's not sidetrack. You can open issue in `ex_doc` to support something like that.

fertapric:
I would also delegate the `unset_*` ones, otherwise the Elixir interface feels incomplete.

hauleth:
@fertapric done.

josevalim:
We could not reach an agreement on the function names. My proposal is:

  * `get_module_level`
  * `put_module_level`
  * `delete_module_level`

And the same for `application`. The idea is to name things like `get_env`. Does anyone feel strongly against these? If not, shall we go ahead with this proposal?

lib/logger/lib/logger.ex
```diff
@@ -693,6 +693,68 @@ defmodule Logger do
...
+
+  """
+  @spec set_application_level(atom(), level() | :all | :none) :: :ok | {:error, :not_loaded}
+  defdelegate set_application_level(app, level), to: :logger
+
+  @doc """
+  Reset logging level for given module to primary level.
+  """
+  @spec unset_module_level(module() | [module()]) :: :ok
+  defdelegate unset_module_level(module), to: :logger
```
fertapric: I would include a `delete_module_level/0` or a `delete_module_level(:all)` that delegates to [`:logger.unset_module_level/0`](https://erlang.org/doc/man/logger.html#unset_module_level-0). I think being able to reset everything in just one function call is quite handy 🙂

josevalim: `delete_module_level(:all)` has a conflict with a potential module named `:all`. I suggest either `delete_all_module_levels` or `delete_all_module_level` (since there is a `get_all_env`).

hauleth: Added this as well.

josevalim:

PR #9797, opened 2020-02-08
Add structured logging

hauleth:
This implements it only in code. Documentation is not ready yet.

Ref #9465

### DIFF
--- a/lib/logger/lib/logger.ex
+++ b/lib/logger/lib/logger.ex
@@ -848,7 +848,10 @@ defmodule Logger do
 
   # TODO: Remove that in Elixir 2.0
   def __do_log__(level, other, metadata) do
-    IO.warn("passing #{inspect(other)} to Logger is deprecated, expected a map, a keyword list, a binary, or an iolist")
+    IO.warn(
+      "passing #{inspect(other)} to Logger is deprecated, expected a map, a keyword list, a binary, or an iolist"
+    )
+
     :logger.macro_log(%{}, level, to_string(other), add_elixir_domain(metadata))
   end
 
--- a/lib/logger/lib/logger.ex
+++ b/lib/logger/lib/logger.ex
@@ -834,10 +834,10 @@ defmodule Logger do
   def __do_log__(level, fun, metadata) when is_function(fun, 0) and is_map(metadata) do
     case fun.() do
       {msg, meta} ->
-        __do_log__(level, msg, Enum.into(meta, add_elixir_domain(metadata)))
+        __do_log__(level, msg, Enum.into(meta, metadata))
 
       msg ->
-        __do_log__(level, msg, add_elixir_domain(metadata))
+        __do_log__(level, msg, metadata)
     end
   end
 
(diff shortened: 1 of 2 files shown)


lib/logger/lib/logger.ex
```diff
@@ -812,20 +813,22 @@ defmodule Logger do
...
   end
 
+  defguardp is_msg(msg) when is_binary(msg) or is_list(msg) or is_map(msg)
+
   @doc false
   def __do_log__(level, fun, metadata) when is_function(fun, 0) and is_map(metadata) do
     case fun.() do
       {msg, meta} ->
-        :logger.macro_log(%{}, level, msg, Enum.into(meta, add_elixir_domain(metadata)))
+        __do_log__(level, msg, Enum.into(meta, add_elixir_domain(metadata)))
```
josevalim: Aren't we adding the elixir domain twice? Maybe we shouldn't call `add_elixir_domain` since we already do it in `do_log`.

josevalim:
Thanks @hauleth. I have added a comment, can you please take a look at it and please rebase it?

We also should document that structured metadata already arrives formatted to Logger backends, under the section that explains how Logger backends work.

lib/logger/lib/logger.ex
hauleth: I will check that. And fix it if needed.

hauleth: Yeah, it was added twice. Fixed. I will also add test for that.

hauleth:
TBH I am enormously happy how straightforward that "fix" was. I forgot how I implemented it earlier, so I needed to "reinvent" it again, and I think that code was well prepared for it.

self five

lib/logger/lib/logger.ex
```diff
@@ -812,20 +813,22 @@ defmodule Logger do
...
 
-  def __do_log__(level, chardata, metadata)
-      when (is_binary(chardata) or is_list(chardata)) and is_map(metadata) do
-    :logger.macro_log(%{}, level, chardata, add_elixir_domain(metadata))
+  def __do_log__(level, msg, metadata)
+      when is_msg(msg) and is_map(metadata) do
+    :logger.macro_log(%{}, level, msg, add_elixir_domain(metadata))
   end
 
   # # TODO: Remove that in Elixir 2.0
```
fertapric: Nitpick: I think the warning two lines below should be updated to reflect the new supported types 🙂

fertapric:
Minor nitpick: the signature of several functions in `Logger` say `chardata_or_fun` and I think we should change it to `msg_or_fun` (or similar)

lib/logger/lib/logger.ex
hauleth: Done

lib/logger/lib/logger.ex
```diff
@@ -885,8 +891,8 @@ defmodule Logger do
         Logger.#{level}("#{message}")
```
josevalim: We need to improve the generated docs to show an example of structured logging too. :)

hauleth: Ok, so I will prepare that as well.

hauleth: Added docs for reports as well.

josevalim:

PR #9798, opened 2020-02-08
Use correct return value from report_cb/2

hauleth:
Earlier it assumed that the return type is the same as in `report_cb/1` which is not true. `report_cb/1` returns format string and data while `report_cb/2` returns already formatted string.


josevalim:

PR #9802, opened 2020-02-09
Use correct return value from report_cb/2

hauleth:
Earlier it assumed that the return type is the same as in `report_cb/1` which is not true. `report_cb/1` returns format string and data while `report_cb/2` returns already formatted string.

This is back port of #9798 for 1.10 where this bug is present as well. This will require release of 1.10.1 as this can cause reregistration of `Logger.Handler` which can result in missed log messages.


josevalim:
I have already backported it to v1.10. :) Thanks in any case!

hauleth:
I haven't seen that, so I just cherry picked the branch. In that case it is great.

PR #9817, opened 2020-02-14
Use acc type alias in spec of Enum.reduce/3

gianluca-nitti:
I was reading the docs for the `Enum` module, and noticed it defines an `acc() :: any()` type alias, which is used in the spec of `reduce/2`, but not in the one for `reduce/3`. I added it so that it's clearer which one of the reducer's arguments is the accumulator in `reduce/3` as well.


josevalim:

PR #9818, opened 2020-02-14
Add Module.to_string

schneiderderek:
This works similar to `Atom.to_string`, but trims off the `Elixir.` prefix for module names.

**Motivation**

We have similar code in our codebase for handing this and I've seen similar functionality in other libraries. So it seemed fairly reasonable to open a PR to discuss the change.

**Outstanding question:**

`MyModule |> Module.to_string() |> Module.split()` doesn't work, do we need to support this? Since `Module.split()` accepts a string it seems like it should work, but on the other hand you don't need the `Module.to_string()` call since you could just do `MyModule |> Module.split()`.

### DIFF
--- a/lib/elixir/lib/module.ex
+++ b/lib/elixir/lib/module.ex
@@ -1388,6 +1388,22 @@ defmodule Module do
     raise ArgumentError, "expected an Elixir module, got: #{inspect(original)}"
   end
 
+  @doc """
+  Converts a module name to a string.
+
+  ## Examples
+
+      iex> Module.to_string(Example.Module)
+      "Example.Module"
+  """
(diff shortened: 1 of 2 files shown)


wojtekmach:
We can achieve the same result with Kernel.inspect/2: `inspect(Foo) #=> “foo”`

josevalim:
Yup, inspect is the way to go. If you think this can be made clearer, a pull request to the docs would be welcome, although I am not quite sure the best place to put it. I suspect you have just looked for it in a couple places, so you may have a better hunch than us. :) Thanks!

schneiderderek:
🤦‍♂ forgot about `inspect` for this. Thank you both! @wojtekmach @josevalim

PR #9819, opened 2020-02-15
Fix error message for import :except option

wojtekmach:
Before this patch we had:

    iex> import Kernel, except: :macros
    ** (CompileError) iex:1: invalid :except option for import, expected value to be an atom :functions, :macros, or a list literal, got: :macros
        (elixir 1.11.0-dev) src/elixir_import.erl:84: :elixir_import.calculate/6
        (elixir 1.11.0-dev) src/elixir_import.erl:24: :elixir_import.import/4


josevalim:
Good catch! Can you please look at the build failures too? Thanks!

josevalim:

PR #9820, opened 2020-02-16
Clarify System.cmd/2 environment handling

voltone:
Unsetting variables using `nil` is different from Erlang's `false`, and this conversion does not appear to be documented anywhere


josevalim:

PR #9821, opened 2020-02-16
Do not compile erlang files with warnings_as_errors unless on CI

wojtekmach:
Similar to https://github.com/elixir-lang/elixir/pull/8716 but for Erlang files.


josevalim:

PR #9822, opened 2020-02-16
Support `import Mod, only: :sigils`

wojtekmach:
Ref https://groups.google.com/d/msg/elixir-lang-core/QGU3ARQeL7w/AJMEtsYgBQAJ

### DIFF
--- a/lib/elixir/src/elixir_import.erl
+++ b/lib/elixir/src/elixir_import.erl
@@ -40,8 +40,8 @@ take_sigils(List) ->
 
 is_sigil({Name, Arity}) ->
   Arity == 2 andalso is_sigil(atom_to_list(Name));
-is_sigil("sigil_" ++ Rest) ->
-  lists:all(fun(X) -> (X >= $a andalso X =< $z) orelse (X >= $A andalso X =< $Z) end, Rest);
+is_sigil("sigil_" ++ [S]) when (S >= $a andalso S =< $z) orelse (S >= $A andalso S =< $Z) ->
+  true;
 is_sigil(Name) when is_list(Name) ->
   false.
 
--- a/lib/elixir/src/elixir_import.erl
+++ b/lib/elixir/src/elixir_import.erl
@@ -12,8 +12,14 @@ import(Meta, Ref, Opts, E) ->
         {Added1, Funs} = import_functions(Meta, Ref, Opts, E),
         {Funs, keydelete(Ref, ?key(E, macros)), Added1};
       {only, macros} ->
-        {Added2, Macs} = import_macros(true, Meta, Ref, Opts, E),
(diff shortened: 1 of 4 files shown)


lib/elixir/src/elixir_import.erl
```diff
@@ -29,6 +35,16 @@ import(Meta, Ref, Opts, E) ->
   elixir_env:trace({import, [{imported, Added} | Meta], Ref, Opts}, E),
   {Functions, Macros}.
 
+take_sigils(List) ->
+  lists:filter(fun is_sigil/1, List).
+
+is_sigil({Name, Arity}) ->
+  Arity == 2 andalso is_sigil(atom_to_list(Name));
+is_sigil("sigil_" ++ Rest) ->
+  lists:all(fun(X) -> (X >= $a andalso X =< $z) orelse (X >= $A andalso X =< $Z) end, Rest);
```
josevalim: This definition is too loose at the moment. :)

lib/elixir/src/elixir_import.erl
```diff
@@ -12,8 +12,14 @@ import(Meta, Ref, Opts, E) ->
...
-        {Added2, Macs} = import_macros(true, Meta, Ref, Opts, E),
-        {keydelete(Ref, ?key(E, functions)), Macs, Added2};
+        {Added1, Macs} = import_macros(true, Meta, Ref, Opts, E),
+        {keydelete(Ref, ?key(E, functions)), Macs, Added1};
+      {only, sigils} ->
+        {Added1, Funs1} = import_functions(Meta, Ref, Opts, E),
+        {Added2, Macs1} = import_macros(false, Meta, Ref, Opts, E),
+        Funs2 = keyupdate(Ref, fun take_sigils/1, Funs1),
+        Macs2 = keyupdate(Ref, fun take_sigils/1, Macs1),
+        {Funs2, Macs2, Added1 or Added2};
```
fertapric: I believe the trace will return `imported: true` even if no sigils are imported, and I think the trace in that case (if no sigils are imported) should be `imported: false`. In other words, I think `Added1 or Added2` needs to be calculated after filtering the sigils.

lib/elixir/test/elixir/kernel/import_test.exs
```diff
@@ -137,6 +137,22 @@ defmodule Kernel.ImportTest do
     import String
   end
 
+  defmodule Sigils do
+    # when imported it should cause conflict
+    def is_integer(_), do: false
+
+    def sigil_X(_, _), do: :x
+
+    defmacro sigil_Y(_, _), do: :y
```
josevalim: Let's define a fake sigil too, `sigil__` and make sure it is not imported. You can check if something is imported by looking inside `__ENV__`.

fertapric:
Thanks @wojtekmach! 💜

PR #9823, opened 2020-02-16
Better Git diffs

hauleth:
In Git 2.25 there comes native support for Elixir definitions. That will
make `git diff --function-context` work properly. Additionally it makes
diff headers to contain name of enclosing function.

I hope that this will work in GitHub diffs as well.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,3 @@
 lib/elixir/test/elixir/fixtures/*.txt text eol=lf
+*.ex diff=elixir
+*.exs diff=elixir


michalmuskala:
Would it make sense to generate a `.gitattributes` file when generating a new project, similar to `.gitignore`?

hauleth:
@michalmuskala I would rather document such thing and encourage people to add it to their own global Git configuration:

```sh
$ mkdir ~/.config/git
$ printf '*.ex diff=elixir\n*.exs diff=elixir\n' >> ~/.config/git/attributes
```

michalmuskala:
Makes sense. Given I know nothing about how git attributes work - why wouldn't that be a default now that git has this support? Is there some conflict on the extensions?

hauleth:
Because Git do not support default attributes for any files at all. It would be big change in Git to provide defaults and I do not have time to go through that right now.

TL;DR there is no "default" attributes in Git, everything uses the same and user can define them on their own

michalmuskala:
Makes sense. Thank you for explaining.

fertapric:
Thanks @hauleth! 💜

PR #9825, opened 2020-02-16
Add EditorConfig to core and generated projects

hauleth:
EditorConfig is project that wants to provide simple format to configure
different editors in a way, that will provide similar output code in all
of them. [There is a lot of editors that support such configuration
file OOtB][2]. Other editors often have plugins that provides such
support.

Used options are:

Global:

- `insert_final_newline` is set to true to follow POSIX standard
- `charset` is defined to be UTF-8 for editors that use different
charset by default

For Elixir files:

- `indent_style` is set to `space` as it is enforced by `mix format`
- `indent_size` is set to `2` as it is enforced by `mix format`

Only other option worth considering is `max_line_length` that is
[supported by few official plugins][1], but as `Code.format_*` functions 
will not parse that file, I preferred to not set it by default to not 
introduce confusion.

[1]: https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties#max_line_length
[2]: https://editorconfig.org/#download

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -9,7 +9,7 @@ charset = utf-8
 indent_style = space
 indent_size = 2
 
-[*.erl]
+[*.{erl,hrl}]
 indent_style = space
 indent_size = 2
 
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,17 @@
+root = true
+
+[*]
+end_of_line = true
+insert_final_newline = true
+charset = utf-8
+
+[*.{ex,exs}]
+indent_style = space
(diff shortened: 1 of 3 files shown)


hauleth:
Also there is [big list of projects that use EditorConfig already](https://github.com/editorconfig/editorconfig/wiki/Projects-Using-EditorConfig).

josevalim:
Hi, this has been proposed a couple times but, as you said, the formatter takes care of most of it, and we don't think we should be adding new files to the generated application unless they are absolutely worth it. Thanks!

PR #9826, opened 2020-02-17
Multi-letter sigils

wojtekmach:
Ref https://groups.google.com/d/msg/elixir-lang-core/QGU3ARQeL7w/AJMEtsYgBQAJ

### DIFF
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -5514,6 +5514,22 @@ defmodule Kernel do
   end
 
   @doc """
+  Handles the `~URI` sigil.
+
+  See `URI.parse/1` for more information.
+
+  ## Examples
+
+      iex> ~URI<https://elixir-lang.org>
+      ~URI<https://elixir-lang.org>
+
+  """
+  @doc since: "1.11.0"
+  defmacro sigil_URI({:<<>>, _, [string]} = _uri_string, [] = _modifiers) do
+    Macro.escape(URI.parse(string))
+  end
+
+  @doc """
   Handles the `~Version` sigil.
 
   See `Version.parse!/1` for more information.
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -5513,6 +5513,22 @@ defmodule Kernel do
     :erlang.list_to_ref('#Ref<#{string}>')
   end
 
+  @doc """
+  Handles the `~Version` sigil.
+
+  See `Version.parse!/1` for more information.
+
+  ## Examples
+
+      iex> ~Version<2.0.1-alpha1>
+      ~Version<2.0.1-alpha1>
+
+  """
+  @doc since: "1.11.0"
+  defmacro sigil_Version({:<<>>, _, [string]} = _version_string, [] = _modifiers) do
+    Macro.escape(Version.parse!(string))
+  end
+
   ## Shared functions
 
   defp assert_module_scope(env, fun, arity) do
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -745,7 +745,7 @@ defmodule Kernel do
   ## Examples
 
       make_ref()
-      #=> #Reference<0.0.0.135>
+      #=> ~Reference<0.0.0.135>
 
   """
   @spec make_ref() :: reference
@@ -5460,6 +5460,57 @@ defmodule Kernel do
 
   defp split_words(_string, _mods, _caller) do
     raise ArgumentError, "modifier must be one of: s, a, c"
+  end
+
+  @doc """
+  Handles the `~PID` sigil.
+
+  This function allows creating arbitrary PIDs, even the ones that don't actually exist in the VM.
+  As such, it should only be used for debugging purposes.
+
+  ## Examples
+
+      iex> ~PID<0.108.0>
+      ~PID<0.108.0>
+
+  """
+  @doc since: "1.11.0"
+  def sigil_PID(string, [] = _modifiers) do
+    :erlang.list_to_pid('<#{string}>')
+  end
+
+  @doc """
+  Handles the `~Port` sigil.
+
+  This function allows creating arbitrary ports, even the ones that don't actually exist in the VM.
+  As such, it should only be used for debugging purposes.
+
+  ## Examples
+
+      iex> ~Port<0.6>
+      ~Port<0.6>
+
+  """
+  @doc since: "1.11.0"
+  def sigil_Port(string, [] = _modifiers) do
+    :erlang.list_to_port('#Port<#{string}>')
+  end
+
+  @doc """
+  Handles the `~Reference` sigil.
+
+  This function allows creating arbitrary references, even the ones that don't actually exist in the VM.
+  As such, it should only be used for debugging purposes.
+
+  ## Examples
+
+      iex> ~Reference<0.2283498464.2022703108.246828>
+      ~Reference<0.2283498464.2022703108.246828>
+
+  """
+  @doc since: "1.11.0"
+  def sigil_Reference(string, [] = _modifiers) do
+    :erlang.list_to_ref('#Ref<#{string}>')
   end
 
   ## Shared functions
--- a/lib/elixir/lib/macro.ex
+++ b/lib/elixir/lib/macro.ex
@@ -54,16 +54,22 @@ defmodule Macro do
 
   Macros are also commonly used to implement custom sigils. To create a custom
   sigil, define a function with the name `sigil_{identifier}` that takes two
-  arguments. The first argument will be the string, the second will be a charlist
-  containing any modifiers. If the sigil is lower case (such as `sigil_x`) then
-  the string argument will allow interpolation. If the sigil is upper case
-  (such as `sigil_X`) then the string will not be interpolated.
+  arguments. The identifier can be an upper case letter followed by more letters,
+  or a single downcase letter. The first argument will be the string, the second
+  will be a charlist containing any modifiers. If the sigil is lower case (such
+  as `sigil_x`) then the string argument will allow interpolation.  If the sigil
+  is upper case (such as `sigil_X`, `sigil_Foo`, etc.) then the string will not be
+  interpolated.
 
   Valid modifiers include only lower and upper case letters. Other characters
   will cause a syntax error.
 
   The module containing the custom sigil must be imported before the sigil
   syntax can be used.
+
+  Elixir originally only allowed single-letter sigils. Support for multi-letter
+  sigils have been added in v1.11.0. Developers are encouraged to use multi-letter
+  sigils as their name can be more intention revealing.
 
   ### Examples
 
@@ -1100,24 +1106,46 @@ defmodule Macro do
     delimiter = Keyword.get(meta, :delimiter, "\"")
     {left, right} = delimiter_pair(delimiter)
 
-    case Atom.to_string(sigil) do
-      <<"sigil_", name>> when name >= ?A and name <= ?Z ->
+    case parse_sigil(Atom.to_string(sigil)) do
+      {:upcase, name} ->
         args = sigil_args(args, fun)
         {:<<>>, _, [binary]} = parts
-        formatted = <<?~, name, left::binary, binary::binary, right::binary, args::binary>>
+
+        formatted =
+          <<?~, name::binary, left::binary, binary::binary, right::binary, args::binary>>
+
         {:ok, fun.(ast, formatted)}
 
-      <<"sigil_", name>> when name >= ?a and name <= ?z ->
+      {:downcase, name} ->
         args = sigil_args(args, fun)
-        formatted = "~" <> <<name>> <> interpolate(parts, left, right, fun) <> args
+        formatted = "~" <> name <> interpolate(parts, left, right, fun) <> args
         {:ok, fun.(ast, formatted)}
 
-      _ ->
+      :error ->
         :error
     end
   end
 
   defp sigil_call(_other, _fun) do
+    :error
+  end
+
+  defp parse_sigil("sigil_" <> <<letter>>) when letter >= ?a and letter <= ?z do
+    {:downcase, <<letter>>}
+  end
+
+  defp parse_sigil("sigil_" <> <<first>> <> remaining) when first >= ?A and first <= ?Z do
+    if Enum.all?(
+         String.to_charlist(remaining),
+         &((&1 >= ?a and &1 <= ?z) or (&1 >= ?A and &1 <= ?Z))
+       ) do
+      {:upcase, <<first>> <> remaining}
+    else
+      :error
+    end
+  end
+
+  defp parse_sigil(_) do
     :error
   end
 
--- a/lib/elixir/lib/uri.ex
+++ b/lib/elixir/lib/uri.ex
@@ -661,7 +661,7 @@ defimpl String.Chars, for: URI do
              path != "" and binary_part(path, 0, 1) != "/" do
     raise ArgumentError,
           ":path in URI must be nil or an absolute path if :host or :authority are given, " <>
-            "got: #{inspect(uri)}"
+            "got: #{inspect(Map.from_struct(uri))}"
   end
 
   def to_string(%{scheme: scheme, port: port, path: path, query: query, fragment: fragment} = uri) do
@@ -697,5 +697,11 @@ defimpl String.Chars, for: URI do
       if(String.contains?(host, ":"), do: ["[", host | "]"], else: host),
       if(port, do: [":" | Integer.to_string(port)], else: [])
     ]
+  end
+end
+
+defimpl Inspect, for: URI do
+  def inspect(uri, _) do
+    "~URI<" <> URI.to_string(uri) <> ">"
   end
 end
--- a/lib/elixir/lib/version.ex
+++ b/lib/elixir/lib/version.ex
@@ -345,9 +345,7 @@ defmodule Version do
 
   ## Examples
 
-      iex> {:ok, version} = Version.parse("2.0.1-alpha1")
-      iex> version
-      #Version<2.0.1-alpha1>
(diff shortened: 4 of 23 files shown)


lib/iex/lib/iex/helpers.ex
```diff
@@ -1193,7 +1193,7 @@ defmodule IEx.Helpers do
   ## Examples
 
       iex> pid("0.21.32")
-      #PID<0.21.32>
+      ~PID<0.21.32>
 
   """
   def pid(string) when is_binary(string) do
```
josevalim: Please add a TODO to pid/1 - port/1 - reference/1 as they can now be deprecated. :)

lib/elixir/src/elixir_tokenizer.erl
```diff
@@ -1067,6 +1077,63 @@ reverse_number([H | T], Number, Original) ->
 reverse_number([], Number, Original) ->
   {Number, Original}.
 
+%% Sigils
+
+tokenize_sigil([H | T], Line, Column, Scope, Tokens, false, Acc) when ?is_letter(H) ->
+  tokenize_sigil(T, Line, Column, Scope, Tokens, false, [H | Acc]);
+
+tokenize_sigil([H | _] = Original, Line, Column, _Scope, Tokens, true, Acc) when ?is_letter(H) ->
+  Original2 = "~" ++ lists:reverse(Acc) ++ Original,
```
josevalim:
```suggestion
  Original2 = [$~] ++ lists:reverse(Acc) ++ Original,
```

lib/elixir/lib/kernel.ex
```diff
@@ -5462,6 +5462,46 @@ defmodule Kernel do
     raise ArgumentError, "modifier must be one of: s, a, c"
   end
 
+  @doc """
+  Handles the `~PID` sigil.
+  """
```
josevalim: We should probably add a note about hardcoding PID values. Same for Port and Reference. The idea is that you can use these sigils for copy and pasting while debugging, but PIDs are not guaranteed to be the same between VM runs.

josevalim:
We need to update the docs for the Macro module and the Syntax Reference page.

josevalim:
In the docs for the Macro module, we should note that Elixir originally only had single-letter sigils and multi-letter ones were added on v1.11.0 and we encourage developers to use the multi-letter one.

lib/elixir/lib/kernel.ex
wojtekmach: Good call.

list_to_pid docs say [1]

> This BIF is intended for debugging and is not to be used in application programs.

we have a similar warning for Function.info/1:

> **Note**: this function must be used only for debugging purposes.

should we just use that or talk a little bit more about the VM as you mentioned?

[1] http://erlang.org/doc/man/erlang.html#list_to_pid-1
[2] https://hexdocs.pm/elixir/Function.html#info/1

josevalim: Your call. The clearer we can be, the better.

wojtekmach: Done!

lib/elixir/lib/version.ex
```diff
@@ -671,7 +669,7 @@ end
 
 defimpl Inspect, for: Version do
   def inspect(self, _opts) do
-    "#Version<" <> to_string(self) <> ">"
+    "~Version<" <> to_string(self) <> ">"
```
hauleth: Wouldn't that be a breaking change? Out of all other it is the only one that could be used in a form of (checking on PIDs or Refs would almost always fail and URI didn't had `Inspect` implemented):

```elixir
iex> Version.parse("1.2.3")
#Version<1.2.3>
```

wojtekmach: Not sure I follow, could you elaborate?

Do you mean that someone could have written:

```elixir
"#Version" <> string = inspect(version)
```

and now it'd be a breaking change? FWIW we already broke such promise when we introduced `~U`. I personally wouldn't consider it a breaking change as inspect is using for showing stuff, folks shouldn't couple to that representation, but I hear what you're saying.

hauleth: No, I am talking about [opaque types in DocTests](https://hexdocs.pm/ex_unit/ExUnit.DocTest.html#module-opaque-types).

wojtekmach: Gotcha. Yeah, these doctests would break.

josevalim: Yes, whenever we change inspected representations, tests may break. However, inspected representations are allowed to change, because the primary goal is developer readability and information. I would say this is even a pro of this PR, because we will push people towards a programmatic representation instead of a textual one.

hauleth: Ok, I am just pointing this out. If you are ok with that, then I am ok. I was pointing this just in case if that wasn't intended.

lib/elixir/lib/kernel.ex
```diff
@@ -5462,6 +5462,69 @@ defmodule Kernel do
     raise ArgumentError, "modifier must be one of: s, a, c"
   end
 
+  @doc """
+  Handles the `~PID` sigil.
+
+  This function allows creating arbitrary PIDs, even the ones that don't actually exist in the VM.
+  As such, it should only be used for debugging purposes.
+  """
```
whatyouhide: Can we add an "Examples" section to this and the other sigil functions?

lib/elixir/lib/kernel.ex
```diff
@@ -5462,6 +5462,69 @@ defmodule Kernel do
...
+  @doc since: "1.11.0"
+  def sigil_PID(string, modifiers)
+
+  def sigil_PID(string, []) do
```
whatyouhide: If you're using the bodiless clause for having a nice argument name for the `modifiers` argument, you can do:

```suggestion
  def sigil_PID(string, [] = _modifiers) do
```

Shaves off a couple of lines and achieves the exact same goal. Applies to all the new functions.

lib/elixir/lib/kernel.ex
```diff
@@ -5462,6 +5462,69 @@ defmodule Kernel do
...
+  Handles the `~Version` sigil.
+
+  See `Version.parse!/1` for more information.
+  """
+  @doc since: "1.11.0"
+  defmacro sigil_Version(string, modifiers)
+
+  defmacro sigil_Version({:<<>>, _, [string]}, []) do
+    Macro.escape(Version.parse!(string))
+  end
```
whatyouhide: Should we have `~Version` and `~URI` in Kernel? I'd argue they belong in their respective modules so that people have to import them to use them. I don't think they're nearly common enough. Thoughts?

lib/elixir/lib/macro.ex
```diff
@@ -54,17 +54,23 @@ defmodule Macro do
...
+  interpolated.
 
   Valid modifiers include only lower and upper case letters. Other characters
   will cause a syntax error.
 
   The module containing the custom sigil must be imported before the sigil
   syntax can be used.
 
+  Elixir originally only allowed single-letter sigils. Support for multi-letter
+  sigils have been added in v1.11.0. Developers are encouraged to use multi-letter
```
whatyouhide:
```suggestion
  sigils has been added in v1.11.0. Developers are encouraged to use multi-letter
```

lib/iex/lib/iex/helpers.ex
```diff
@@ -1260,10 +1262,11 @@ defmodule IEx.Helpers do
   ## Examples
 
       iex> ref("0.1.2.3")
-      #Reference<0.1.2.3>
+      ~Reference<0.1.2.3>
 
   """
   @doc since: "1.6.0"
+  # TODO: deprecate in favour of ~Reference sigil
```
whatyouhide: We should likely put the version of the deprecation. I think we can soft-deprecate in 1.11, meaning we'd make this `@doc false` and then change this to 
```suggestion
  # TODO: hard-deprecate in favour of ~Reference sigil in Elixir v1.13
```

Thoughts @josevalim?

lib/elixir/lib/kernel.ex
josevalim: This is a good point. On one hand it would be convenient to have them in Kernel but at the same time it would be fairer to have them imported, as libraries outside of Elixir would also require. Does anyone have any further thoughts?

lib/elixir/lib/kernel.ex
wojtekmach: Good call. I was skeptical at first because we're showing the same exact code but I think it's an improvement nonetheless. WDYT?

Another thing we could do is maybe make it less boring with:

```
iex> is_pid(~PID<0.108.0>)
true
```

lib/elixir/lib/kernel.ex
```diff
@@ -5462,6 +5462,89 @@ defmodule Kernel do
...
+  ## Examples
+
+      iex> ~Version<2.0.1-alpha1>
+      ~Version<2.0.1-alpha1>
+
+  """
+  @doc since: "1.11.0"
+  defmacro sigil_Version({:<<>>, _, [string]} = _version_string, [] = _modifiers) do
+    Macro.escape(Version.parse!(string))
+  end
```
josevalim: So I have thought more about having this in Kernel vs in the respective modules, at least for non built-in types. The "issue" is that we already have many built-in types with sigils in Kernel, so I think it is best to keep all of them there, otherwise the line is arbitrary.

The other issue with the current implementation is that it is not "plug and play". For example, if something returns `~Decimal<1.0>`, i cannot simply use that in my code, unless the sigil is imported. This UI can be confusing. So my suggestion would be for us to, when failing to compiling due to a non-existing local function, we could check if:

1. If the function is a sigil starting with uppercase

2. If so, if there is a module with the name of the sigil and that exports a `sigil_NAME` function or macro. If so, the error message will add: "Perhaps you meant to "import Decimal, only: :sigils" before using it?" or similar

Thoughts?

lib/iex/lib/iex/helpers.ex
```diff
@@ -1193,9 +1193,10 @@ defmodule IEx.Helpers do
   ## Examples
 
       iex> pid("0.21.32")
-      #PID<0.21.32>
+      ~PID<0.21.32>
 
   """
+  # TODO: deprecate in favour of ~PID sigil
```
josevalim:
```suggestion
  # TODO: deprecate in favour of ~PID sigil in v1.13
```

lib/iex/lib/iex/helpers.ex
```diff
@@ -1227,10 +1228,11 @@ defmodule IEx.Helpers do
   ## Examples
 
       iex> port("0.4")
-      #Port<0.4>
+      ~Port<0.4>
 
   """
   @doc since: "1.8.0"
+  # TODO: deprecate in favour of ~Port sigil
```
josevalim:
```suggestion
  # TODO: deprecate in favour of ~Port sigil in v1.13
```

lib/iex/lib/iex/helpers.ex
```diff
@@ -1260,10 +1262,11 @@ defmodule IEx.Helpers do
   ## Examples
 
       iex> ref("0.1.2.3")
-      #Reference<0.1.2.3>
+      ~Reference<0.1.2.3>
 
   """
   @doc since: "1.6.0"
+  # TODO: hard-deprecate in favour of ~Reference sigil in Elixir v1.13
```
josevalim: We should `@doc false` them too.

lib/elixir/lib/kernel.ex
whatyouhide: I think it's fine to show the same thing but in Elixir we try to have this "Example" section everywhere so even if it's a bit pointless I think it's still nice to have :)

lib/elixir/lib/kernel.ex
whatyouhide: Yeah I think handling the error better at the compiler level is the way to go. In this case, I think it makes the case for having `~Version` and `~URI` out of Kernel even stronger.

josevalim: Let's separate these discussion. My main concern is: can you explain why ~Version is not in Kernel but ~Port is? Why is ~r in Kernel and not ~URI?

There is one heuristic I could come up with to explain it but it is so long that I am lazy of even typing it. :P

whatyouhide: Ahaha, okay, let's have that as a separate discussion, good point!

wojtekmach: +1 on error message from me.

michalmuskala:
I think there's an angle to consider especially with `PID`, `Port` and `Ref` in that they aren't persistent and usually shouldn't be used in modules. They are really useful in the shell, though, and in general with interactive things. In modules, I could see how it could be confusing or error prone when people just copy-paste some things from output, or assert on concrete pid value in tests, etc. Would it make sense to consider some protections against this?

lib/elixir/lib/kernel.ex
michalmuskala: `~Port` is a built-in data type. `~Version` is not. We have `is_port` in `Kernel`, we don't have `is_version` in `Kernel`. I see a similar analogy with sigils.

josevalim: Yeah! It doesn't explain why we have sigils for Regex or Date/Time though. :/

So the rule is like "it has to be a built-in type and use a multi-letter sigil" and that relies on users knowing what is a built-in type and there are probably some gotchas still. I think it is easier to go full-Oprah on this and say "you get a Kernel sigil", "you get a Kernel sigil" and "you get a Kernel sigil".

lib/elixir/lib/macro.ex
```diff
@@ -54,17 +54,23 @@ defmodule Macro do
...
   sigil, define a function with the name `sigil_{identifier}` that takes two
-  arguments. The first argument will be the string, the second will be a charlist
-  containing any modifiers. If the sigil is lower case (such as `sigil_x`) then
-  the string argument will allow interpolation. If the sigil is upper case
-  (such as `sigil_X`) then the string will not be interpolated.
+  arguments. The identifier can be an upper case letter followed by more letters,
+  or a single downcase letter. The first argument will be the string, the second
+  will be a charlist containing any modifiers. If the sigil is lower case (such
+  as `sigil_x`) then the string argument will allow interpolation.  If the sigil
+  is upper case (such as `sigil_X`, `sigil_Foo` etc) then the string will not be
```
fertapric:
```suggestion
  is upper case (such as `sigil_X`, `sigil_Foo`, etc.) then the string will not be
```

lib/elixir/lib/macro.ex
```diff
@@ -1121,6 +1130,25 @@ defmodule Macro do
     :error
   end
 
+  defp parse_sigil("sigil_" <> <<letter>>) when letter >= ?a and letter <= ?z do
+    {:downcase, <<letter>>}
+  end
+
+  defp parse_sigil("sigil_" <> letters) do
```
fertapric: I think we could check if the first letter is an uppercase letter:

```elixir
iex> ast = {:sigil_foo, [delimiter: "<"], [{:<<>>, [], ["hello"]}, []]}
iex> Macro.to_string(ast)
"~foo<hello>"
```

whatyouhide:
@michalmuskala I am very open to doing some more aggressive checking on where you use the PID/port/ref sigils, but can we have that as a separate discussion for the sake of moving this PR forward?

michalmuskala:
To me this is a problem created entirely by this feature, so it makes sense to think through solutions before the feature is merged.
It's fine with me to delay this discussion, though.

josevalim:
@michalmuskala We have added warnings to the docs saying they need to be used carefully. Is that enough? Another option is to make them a macro and make them warn if used inside a module. So we can effectively use them only in IEx.

lib/elixir/lib/macro.ex
wojtekmach: Applied manually, thanks!

lib/elixir/lib/macro.ex
wojtekmach: Good call!

lib/elixir/lib/uri.ex
```diff
@@ -699,3 +699,9 @@ defimpl String.Chars, for: URI do
     ]
   end
 end
+
+defimpl Inspect, for: URI do
+  def inspect(uri, _) do
+    "~URI<" <> URI.to_string(uri) <> ">"
```
wojtekmach: there's a problem with this sigil that I am unsure how to solve:

```elixir
iex(1)> URI.parse("https://elixir-lang.org?v=>")
~URI<https://elixir-lang.org?v=>>

iex(2)> [ ~URI<https://elixir-lang.org?v=>> ]
** (SyntaxError) iex:2: syntax error before: ']'
```

I guess we should escape `<` as `\<` in the output but maybe it'd be weird?

wojtekmach:
Regarding `~PID`, `~Port`, `~Reference`, if there are supposed to be only used in IEx, should we then define them in `IEx.Helpers` instead of on `Kernel`? The problem is if someone tries to copy-paste e.g. `~PID<0.108.0>` into a module they'd get an error that may be confusing but I guess your point is they shouldn't be doing that in the first place!

josevalim:
Hi @wojtekmach, thanks for the PR. :heart: Your implementation revealed conceptual problems that we were not aware of when simply discussing the approach, in particular:

1. Allowing PID/Port/Reference to be built directly can lead to folks hardcoding references. Moving the sigils to IEx.Helpers can help tackle the problem but introduces its own issues too.

2. Moving some data types inspection to sigils, such as URI, leads to loss of information. Today the inspected representation show all fields but the inspected one would hide them. Of course, this issue exists to any non-opaque struct, such as the calendar types, but this PR may make this issue more apparent.

3. The sigil syntax is not general purpose enough. We already knew this was the case for MapSet and what not, but this implementation even revealed some challenges in the URI sigil implementation.

For all of the reasons above, we won't be moving forward with this. Perhaps the answer is not move towards sigils after at all. For example, a more general `Decimal("1.0")` and `Version("1.0.0")` could perhaps address all of these problems, including supporting MapSet, albeit slightly more verbose. But even that in itself would require further thoughts.

Thanks everyone for the feedback and back to the drawing board!

PR #9827, opened 2020-02-17
Use simpler example for doctests and exceptions

wojtekmach:


lib/ex_unit/lib/ex_unit/doc_test.ex
```diff
@@ -130,8 +130,8 @@ defmodule ExUnit.DocTest do
 
   You can also showcase expressions raising an exception, for example:
 
-      iex(1)> String.to_atom((fn -> 1 end).())
-      ** (ArgumentError) argument error
+      iex(1)> raise "some error"
+      ** (RuntimeErroro) some error
```
josevalim: Make sure this won't emit warnings when the suite is running. :)

wojtekmach:
Eh, of course this was sent prematurely as I was making sure everything works. Should be fine now!

wojtekmach:
Btw, we unfortunately don’t have `doctest ExUnit.DocTest` :’-( as we’re showing examples where it doesn’t work.

josevalim:

PR #9840, opened 2020-02-24
Update `DateTime.from_unix` documentation

### LINKED ISSUE #9839: DateTime.from_unix negative numbers support more than what docs say.
### Environment

* Elixir & Erlang/OTP versions (elixir --version): >= 1.7.0
* Operating system: Linux 5.6

### Current behavior

The [documentation says Elixir supports](https://github.com/elixir-lang/elixir/blob/v1.10.1/lib/elixir/lib/calendar/datetime.ex#L117-L118)  negative unix epoch times up to a limit: 
> Negative Unix times are supported, up to -62167219200 seconds,
>  which is equivalent to "0000-01-01T00:00:00Z" or 0 Gregorian seconds.

But negative numbers greater than that works in 1.10.1:

```elixir
iex> DateTime.from_unix(-99_999_999_999)
{:ok, ~U[-1199-02-15 14:13:21Z]}
```

Back in 1.6.6, this didn't work:

```elixir
$ docker run -it hexpm/elixir:1.6.6-erlang-21.3.1-alpine-3.11.3 iex
Erlang/OTP 21 [erts-10.3] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [hipe]

Interactive Elixir (1.6.6) - press Ctrl+C to exit (type h() ENTER for help)
iex(1)> DateTime.from_unix(-99_999_999_999)
{:error, :invalid_unix_time}
```

It started behaving this way in 1.7.0:

```elixir
$ docker run -it hexpm/elixir:1.7.0-erlang-21.3.1-alpine-3.11.3 iex
Erlang/OTP 21 [erts-10.3] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [hipe]

Interactive Elixir (1.7.0) - press Ctrl+C to exit (type h() ENTER for help)
iex(1)> DateTime.from_unix(-99_999_999_999)
{:ok, #DateTime<-1199-02-15 14:13:21Z>}
iex(2)> DateTime.from_unix(-377_705_116_800)
{:ok, #DateTime<-9999-01-01 00:00:00Z>}
iex(3)> DateTime.from_unix(-377_705_116_801)
{:error, :invalid_unix_time}
```

### Expected behavior

I'm thinking there are two solutions? I'm not sure would be best:

- The documentation should say they're supported up to `-377_705_116_800`, since that seems to be what's happening already

- It seems that [the original issue](https://github.com/elixir-lang/elixir/issues/5124) might have intentionally wanted to bottom-out at `-62_167_219_200` to match common database constraints. If you still want this to be true, then is this a bug?  If so I assume we'd want to guard.

I'll be happy to submit a PR either way.

dbernheisel:
The documentation stated that negative numbers down to -62167219200 are
supported, but it actually works down to -377705116800. 

Add doctests to illustrate the upper and lower boundaries.

Resolves #9839

### DIFF
--- a/lib/elixir/lib/calendar/datetime.ex
+++ b/lib/elixir/lib/calendar/datetime.ex
@@ -98,7 +98,8 @@ defmodule DateTime do
 
   The integer can be given in different unit
   according to `System.convert_time_unit/3` and it will
-  be converted to microseconds internally.
+  be converted to microseconds internally. Up to
+  253402300799 seconds is supported.
 
   Unix times are always in UTC and therefore the DateTime
   will be returned in UTC.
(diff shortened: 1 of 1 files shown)


josevalim:

PR #9845, opened 2020-02-27
Clarifies explanation of unicode and Regex character classes, includes example

fireproofsocks:


josevalim:

PR #9859, opened 2020-02-28
Add `Logger.install/1`

hauleth:
This function should be used for loading configuration for user-defined applications. This just delegates to `logger:add_handlers/1` and the question remains whether we should do just that, or rather write our own function that will allow setting all other options as well.

Ref #9465


josevalim:
Given this is about handling `:logger` configuration, I think at the moment is better to keep it as `:logger.add_handlers`. Especially because all of the documentation to write and configure the handlers will be in logger. If we decide to do more work, then we can consider adding this. Thank you!

PR #9860, opened 2020-02-28
Fix File module rename tests on PASE

### LINKED ISSUE #9858: File module rename tests don't cope with behaviour difference on PASE
### Environment

* Elixir & Erlang/OTP versions (elixir --version): `Erlang/OTP 22 [erts-10.6] [source] [64-bit] [smp:2:2] [ds:2:2:10] [async-threads:1]` `Elixir 1.11.0-dev (18bf816) (compiled with Erlang/OTP 22)` (but reproducible on v1.10.x)

* Operating system: IBM i 7.2

### Current behavior

```
  1) test rename parent dir to existing sub dir (FileTest.Rename)
     test/elixir/file_test.exs:242
     Assertion with == failed
     code:  assert File.rename(src, dest) == {:error, :einval}
     left:  {:error, :eexist}
     right: {:error, :einval}
     stacktrace:
       test/elixir/file_test.exs:248: (test)



  2) test rename file to existing non-empty dir (FileTest.Rename)
     test/elixir/file_test.exs:82
     Assertion with == failed
     code:  assert File.rename(src, dest) == {:error, :eisdir}
     left:  {:error, :eexist}
     right: {:error, :eisdir}
     stacktrace:
       test/elixir/file_test.exs:88: (test)

.......................................................................................................

Finished in 145.9 seconds (82.9s on load, 63.0s on tests)
1584 doctests, 3460 tests, 2 failures, 8 excluded
```

### Expected behavior

These tests should pass. (I don't know if this should be handled by the tests/consuming applications, the Elixir stdlib, or OTP...) My guess is to what's happening is that the AIX compatibility layer (PASE) on i (which BEAM is running on) is thinly wrapping the native environment's version of the rename call, which has [slightly different behaviour](https://www.ibm.com/support/knowledgecenter/ssw_ibm_i_74/apis/renameun.htm).

Side note: Overall, I'm actually impressed with how well Elixir works on this platform. I'm actually using it to develop a simple application against the native database on the platform (via ODBC) and its going surprisingly well. The platform tests passing has been quite pleasant for me, since I'm using used to seeing red or not even being able to think of exotic languages on minority platforms...

NattyNarwhal:
IBM i PASE implements rename in a slightly weird way due to it
wrapping the native syscall. Expect this in the File module's
rename test when expecting an error. Fixes #9858.

This should be the behaviour documented (different from
AIX, which it's emulating, and strange for POSIX):

https://www.ibm.com/support/knowledgecenter/ssw_ibm_i_74/apis/renameun.htm

### DIFF
--- a/lib/elixir/test/elixir/file_test.exs
+++ b/lib/elixir/test/elixir/file_test.exs
@@ -85,7 +85,7 @@ defmodule FileTest do
 
       try do
         File.mkdir_p(Path.join(dest, "a"))
-        assert File.rename(src, dest) == {:error, :eisdir}
+        assert File.rename(src, dest) in [{:error, :eisdir}, {:error, :eexist}]
         assert File.exists?(src)
         refute File.exists?(Path.join(dest, "file.txt"))
       after
@@ -245,7 +245,7 @@ defmodule FileTest do
 
       try do
         assert File.exists?(src)
-        assert File.rename(src, dest) == {:error, :einval}
+        assert File.rename(src, dest) in [{:error, :einval}, {:error, :eexist}]
         assert File.exists?(src)
       after
         File.rm_rf(src)


josevalim:

PR #9867, opened 2020-03-01
mix release: only include the compiling host's exe by default.

azizk:
Hi!
I will migrate to mix release from distillery at work soon, so I was trying things out. I think it would be better if mix release didn't include Windows config files and executables on Unix systems and vice versa. To me it looks cleaner and more sensible if the tool output doesn't mention a different system at all (by default.) If somebody really needs to release for both systems, they can easily do so just like the documentation describes it (updated in this PR.) I fixed two tests but only executed on Linux.

I guess mix release.init shouldn't create an env.bat.eex file on Unix either, or a *.sh file on Windows. I could correct that as well, but let's see what you think of this PR first.

### DIFF
--- a/lib/mix/lib/mix/tasks/release.ex
+++ b/lib/mix/lib/mix/tasks/release.ex
@@ -437,11 +437,12 @@ defmodule Mix.Tasks.Release do
 
     * `:include_executables_for` - a list of atoms detailing for which Operating
       Systems executable files should be generated for. By default, it is set to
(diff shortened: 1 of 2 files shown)


josevalim:
I actually prefer to generate all files for consistency. It gives developers a more complete understanding on how they may want to configure the system depending on the OS. The files are also small, so those who may get bothered by it can opt-out. In any case, thanks for the PR!

PS: I also think that on Windows we should most likely keep both executables, as it is common to use both entry points from Windows.

PR #9874, opened 2020-03-05
Add deprecation warning for Kernel.apply/2 as it is going to be remov…

hassanshaikley:
…ed from erlang

There's a warning there saying that it is marked for removal so it makes sense to me to deprecate it in Elixir!

Source:  http://erlang.org/doc/man/erlang.html#apply-2

### DIFF
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -259,7 +259,7 @@ defmodule Kernel do
       4
 
   """
-  @spec apply(fun, [any]) :: any
+  @deprecated "Use Kernel.apply/3 instead"
   def apply(fun, args) do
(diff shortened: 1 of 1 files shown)


wojtekmach:
I believe apply/2 is not being deprecated, just passing a tuple (e.g.: `apply({String, :upcase, [“foo”])`) is.

hassanshaikley:
oh my. I tried executing that in elixir and it didn’t work to make sure i wasn’t misinterpreting the warning! I must have messed up the syntax.

fertapric:
As @wojtekmach said, the warning message in the documentation refers to passing a tuple `{Module, Function}` as the first argument. 

Thanks @wojtekmach and @hassanshaikley!

PR #9876, opened 2020-03-06
update String.replace/4 and split/3 docs

Cohen-Carlisle:
just some minor docs updates, specifically:

* unify description of pattern in split/3 and replace/4
* fix function references to correct arity


whatyouhide:
Thank you @Cohen-Carlisle! 💟

PR #9879, opened 2020-03-07
Improve 'check_reproducible' script

fertapric:
The new script performs a "BEAM-diff" in case of BEAM files.

Part of #9579

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -143,7 +143,6 @@ test_freebsd_task:
     LC_ALL: en_US.UTF-8
 
   install_script:
-    - sudo pkg update
     - pkg install -y erlang git gmake
     - rm -rf .git
     - gmake compile
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -143,6 +143,7 @@ test_freebsd_task:
     LC_ALL: en_US.UTF-8
 
   install_script:
+    - sudo pkg update
     - pkg install -y erlang git gmake
     - rm -rf .git
     - gmake compile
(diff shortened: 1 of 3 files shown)


lib/elixir/build_diff.exs
```diff
@@ -0,0 +1,176 @@
...
+          file_diff(file1, file2)
+        end
+
+      {:diff, diff}
+    end
+  end
+
+  defp beam_diff(content1, content2) do
+    tmp_file1 = content1 |> humanize_beam_chunks() |> inspect() |> write_tmp()
+    tmp_file2 = content2 |> humanize_beam_chunks() |> inspect() |> write_tmp()
```
josevalim:
```suggestion
    tmp_file1 = content1 |> humanize_beam_chunks() |> inspect(pretty: true, limit: :infinity) |> write_tmp()
    tmp_file2 = content2 |> humanize_beam_chunks() |> inspect(pretty: true, limit: :infinity) |> write_tmp()
```

josevalim:
Hi @fertapric! This looks great. I have dropped one comment: it is important to set limit to infinity otherwise it won't show some output.

I also wonder if it would be better if we compared the chunks ourselves and diff-ed only the chunks that are different. However, that may be unnecessary. Let's see how the diffs look like first.

Finally, one last thing, what if we name the script `ebin_diff.exs` or just `diff.exs`? Build makes me think it is part of the Elixir building/compilation process.

fertapric:
@josevalim good catch! Fixed!

I also renamed the script to `diff.exs`.

PR #9880, opened 2020-03-08
Support coverage results on OS partitioning

josevalim:


lib/mix/lib/mix/tasks/test.coverage.ex
```diff
@@ -0,0 +1,200 @@
...
+        generate_cover_results(test_coverage)
+    end
+  end
+
+  @doc false
+  def start(compile_path, opts) do
+    Mix.shell().info("Cover compiling modules ...")
+    cover_compile(compile_path)
+
+    if name = opts[:export] do
```
josevalim: For those willing to try this on v1.10, you can copy this whole module to your application and change only this line:

```suggestion
    if name = System.get_env("MIX_TEST_PARTITION") do
```

And then in your project `mix.exs`, you can do `test_coverage: [tool: Mix.Tasks.Test.Coverage]`.

lib/mix/lib/mix/tasks/test.ex
```diff
@@ -272,10 +149,15 @@ defmodule Mix.Tasks.Test do
...
-      above the specified threshold, respectively. Defaults to `[threshold: 90]`
+      or a keyword list. When a keyword list is passed, it can specify a
+      `:threshold`, which is a boolean or numeric value that enables coloring
+      of code coverage results in red or green depending on whether the
+      percentage is below or above the specified threshold, respectively.
+      Defaults to `[threshold: 90]`
+    * `:export` - a file name to export results to instead of generating
+      the result on the fly. See `mix test.coverage` to compile a report
+      from multiple exports. This option is automatically enabled when
+      used the process partitioning feature
```
```suggestion
      using the process partitioning feature
```

mackeyja92:
I am getting a couple errors testing this out. Firstly, it looks like it is crashing on my last application in my umbrella. Always the same one no matter which partition I am running tests for.

```
==> api
Cover compiling modules ...
** (EXIT from #PID<0.92.0>) killed
```

Secondly, when converging coverage together, I get this error.

```
** (RuntimeError) trying to access Mix.Project.app_path for an umbrella project but umbrellas have no app
    (mix 1.10.2) lib/mix/project.ex:579: Mix.Project.app_path/1
    (mix 1.10.2) lib/mix/project.ex:605: Mix.Project.compile_path/1
    lib/coverage.ex:28: Mix.Tasks.Test.Coverage.run/1
    (mix 1.10.2) lib/mix/task.ex:330: Mix.Task.run_task/3
    (mix 1.10.2) lib/mix/cli.ex:82: Mix.CLI.run_task/2
```

josevalim:
@mackeyja92 just to confirm, are you running on the branch or have you ported the task? If so,you ported it, can you please post your port?

The reason I ask is because the `test.coverage` task has a `@recursive true` call, which means it should always run recursively and never run inside an umbrella (which means the error you are seeing should not happen).

About the first error, I am not sure. What is your OTP version?

mackeyja92:
@josevalim I have copied it over exactly and changed the one line as you posted `if name = System.get_env("MIX_TEST_PARTITION")`. `@recursive true` is set in my port as well since it's a direct copy.

When running `MIX_TEST_PARTITION=1 mix test --cover --partitions 1` I get the first error at the end and nothing is written to disk.

When running `mix test.coverage` I get the second error.

```
$ cat .tool-versions
erlang 22.2.8
elixir 1.10.2-otp-22
```

josevalim:
Unfortunately I cannot reproduce it:

```
$ mix new umb --umbrella
$ cd umb/apps
$ mix new a
$ mix new b
$ MIX_TEST_PARTITION=1 mix test --cover --partitions 1
$ mix test.coverage
==> a
Importing cover results: cover/1.coverdata

Percentage | Module
   100.00% | A
   100.00% | Total

Generated HTML coverage results in "cover" directory
==> b
Importing cover results: cover/1.coverdata

Percentage | Module
   100.00% | B
   100.00% | Total

Generated HTML coverage results in "cover" directory
```

Can you please isolate it into a demo application that reproduces the error? In your case, is the coverage implementation a file in the same umbrella? Or does it come from a separate dependency?

mackeyja92:
@josevalim Ah, it was because it was inside my umbrella. I moved it outside the umbrella into a shared dep and they stopped crashing during tests. :doh:

However, I am still having an issue running `mix test.coverage` in the umbrella.

```
$ mix test.coverage
** (RuntimeError) trying to access Mix.Project.app_path for an umbrella project but umbrellas have no app
    (mix 1.10.2) lib/mix/project.ex:579: Mix.Project.app_path/1
    (mix 1.10.2) lib/mix/project.ex:605: Mix.Project.compile_path/1
    lib/coverage.ex:28: Mix.Tasks.Test.Coverage.run/1
    (mix 1.10.2) lib/mix/task.ex:330: Mix.Task.run_task/3
    (mix 1.10.2) lib/mix/cli.ex:82: Mix.CLI.run_task/2
```

I added an inspect to the config on line 27 and it is outputting the config of the umbrella app, not the sub apps. 

Running `mix cmd mix test.coverage` gives me the expected results for each app, but crashes on one somewhere.

```
** (MatchError) no match of right hand side value: {:error, {:no_source_code_found, Mix.Tasks.Test.Coverage}}
    lib/coverage.ex:120: anonymous fn/3 in Mix.Tasks.Test.Coverage.html/2
    (elixir 1.10.2) lib/enum.ex:2111: Enum."-reduce/3-lists^foldl/2-0-"/3
    lib/coverage.ex:119: Mix.Tasks.Test.Coverage.html/2
    (mix 1.10.2) lib/mix/task.ex:330: Mix.Task.run_task/3
    (mix 1.10.2) lib/mix/cli.ex:82: Mix.CLI.run_task/2
** (exit) 1
    (mix 1.10.2) lib/mix/tasks/cmd.ex:45: Mix.Tasks.Cmd.run/1
    (mix 1.10.2) lib/mix/task.ex:330: Mix.Task.run_task/3
    (mix 1.10.2) lib/mix/project.ex:352: Mix.Project.in_project/4
    (elixir 1.10.2) lib/file.ex:1544: File.cd!/2
    (mix 1.10.2) lib/mix/task.ex:430: anonymous fn/4 in Mix.Task.recur/1
    (elixir 1.10.2) lib/enum.ex:2111: Enum."-reduce/3-lists^foldl/2-0-"/3
    (mix 1.10.2) lib/mix/task.ex:429: Mix.Task.recur/1
    (mix 1.10.2) lib/mix/project_stack.ex:181: Mix.ProjectStack.recur/1
```

Also, as a side note, would it be possible to get coverage of the entire umbrella rolled into a single total percentage?

josevalim:
Unfortunately I have no idea why it is not respecting the `@recursive` flag. I would have to be able to reproduce it to take a further look.

> Also, as a side note, would it be possible to get coverage of the entire umbrella rolled into a single total percentage?

BUT, given this last question, you may not want to run it recursively anyway. You can roll all coverage into one but you will need to:

1. When you call `cover_compile`, you will need to add the `compile_path` directory of all umbrella children

2. Use Path.wildcard on the cover directory of all umbrella children instead of the current one

`Mix.Project.apps_paths` can help you know where is each child app (and if you are in an umbrella or not).

mackeyja92:
@josevalim Ok, thanks. I'll see if I can change mine to do that since we are looking for entire umbrella coverage instead of per app coverage.

josevalim:
After some thought, it probably makes sense to support aggregated umbrella results, *as long as as the results are still independent*, i.e. lines of project A executed when running project B won't count towards A coverage.

This has been requested in the past as well so I just went ahead and added support for it. Please see the updated docs.

josevalim:

Issue #9885, opened 2020-03-09
String.Char.to_string fails on Version struct


sysashi:
### Environment

* Elixir & Erlang/OTP versions (elixir --version):  Elixir 1.10.2 (compiled with Erlang/OTP 22)

* Operating system:  macOS Mojave 10.14.5

### Current behavior
```
iex> %Version{}

%Inspect.Error{
  message: "got Protocol.UndefinedError with message \"protocol Enumerable not implemented for nil of type Atom. This protocol is implemented for the following type(s): Ecto.Adapters.SQL.Stream, Postgrex.Stream, DBConnection.Stream, DBConnection.PrepareStream, HashSet, Range, Map, Function, List, Stream, Date.Range, HashDict, GenEvent.Stream, MapSet, File.Stream, IO.Stream\" while inspecting %{__struct__: Version, build: nil, major: nil, minor: nil, patch: nil, pre: nil}"
}
```
 
### Expected behavior
if leaving as is (without having default value `[]` for field `pre`)
```
#Version<..>
```

or something like this (making pre default to `[]` and enforcing fields that are needed to print a valid version) 
```
iex> %Version{}
** (ArgumentError) the following keys must also be given when building struct Version: [:major, :minor, :patch]
```

I've dissuced this with _lostkobrakai_ on elixir slack, he helped to narrow down the problem to https://github.com/elixir-lang/elixir/blob/b30cbb600b7edd6a3d3301cac3c082ed9b685dfb/lib/elixir/lib/version.ex#L750-L756

(case where `pre` field is nil)

josevalim:
Good catch! A PR that sets a default for pre and make the other fields required would be great!

sysashi:
I'll gladly work on it. Unfortunately I don't have much experience of contributing / even building elixir, so it might take a day or two.

PR #9886, opened 2020-03-09
Add default and enforce keys in Version struct

sysashi:
String.Char.to_string/1 was failing due to unexpected`nil` value in `pre` field.

I did not know if tests were necessary or not, but it seems like added line will raise without a fix.

Related to #9885


josevalim:

PR #9898, opened 2020-03-17
Add Mix.Shell.cmd!/3 and Mix.Shell.IO.cmd!/3

wojtekmach:
Another attempt at https://github.com/elixir-lang/elixir/pull/7100! :-)

### DIFF
--- a/lib/mix/lib/mix/shell.ex
+++ b/lib/mix/lib/mix/shell.ex
@@ -116,11 +116,8 @@ defmodule Mix.Shell do
   @doc since: "1.11.0"
   def cmd!(command, options \\ [], callback) do
     case cmd(command, options, callback) do
-      0 ->
-        0
-
-      other ->
-        Mix.raise("Command exited with status #{other}")
+      0 -> 0
+      other -> Mix.raise("Command exited with status #{other}")
     end
   end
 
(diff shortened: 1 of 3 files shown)


josevalim:
Can you provide more info about the use case? I am not specially convinced, especially because the error message is generic and if you are invoking this from the Mix task, you often want to provide more information than just saying a command failed with non-zero status. Thoughts?

wojtekmach:
The main idea was to improve one-liners as this one: https://github.com/phoenixframework/phoenix/pull/3707/files#diff-10dbbad9da2f3f3807ff3ef0c1fafc0dR72.

Commands should report issues on stderr which Mix.Shell.cmd by default redirects to stdout. Would including the command that failed in the error message help?

josevalim:
Honestly, I would prefer to keep the API surface small. It is not like this is a frequently used area of the language and having users think a bit how to handle non-zero status code won't hurt!

whatyouhide:
Agreed with @josevalim, I haven't seen enough demand for this feature and in my personal code I've *always* wanted to use a more specific error message in `Mix.raise/1`.

wojtekmach:
Ok, thx for your feedback!

PR #9900, opened 2020-03-19
Improve app name detection by the application tracer

wojtekmach:
OTP app directories contain version numbers and they were included in
the name:

Before:

    warning: :ssl.versions/0 defined in application :ssl-9.6 is used by the current application but the current application does not directly depend on :ssl-9.6. To fix this, you must do one of:

      1. If :ssl-9.6 is part of Erlang/Elixir, you must include it under :extra_applications inside "def application" in your mix.exs
    (...)

After:

    warning: :ssl.versions/0 defined in application :ssl is used by the current application but the current application does not directly depend on :ssl. To fix this, you must do one of:

      1. If :ssl is part of Erlang/Elixir, you must include it under :extra_applications inside "def application" in your mix.exs
    (...)

### DIFF
--- a/lib/mix/lib/mix/compilers/application_tracer.ex
+++ b/lib/mix/lib/mix/compilers/application_tracer.ex
@@ -57,7 +57,7 @@ defmodule Mix.Compilers.ApplicationTracer do
           # If the module belongs to this application (from another compiler), we skip it.
           # If the module is excluded, we skip it too.
           with path when is_list(path) <- :code.which(module),
-               module_app = app(path),
+               {:ok, module_app} <- app(path),
                true <- module_app != app,
                false <- module in excludes or {module, function, arity} in excludes do
             env_mfa =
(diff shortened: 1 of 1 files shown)


lib/mix/lib/mix/compilers/application_tracer.ex
```diff
@@ -57,7 +57,7 @@ defmodule Mix.Compilers.ApplicationTracer do
           # If the module belongs to this application (from another compiler), we skip it.
           # If the module is excluded, we skip it too.
           with path when is_list(path) <- :code.which(module),
-               [module_app, "ebin", _beam] <- path |> Path.split() |> Enum.take(-3),
+               module_app = app(path),
```
wojtekmach: replacing `<-` with `=` makes this less forgiving. We'd also incorrectly handle when the app name legitimately contains a dash. I don't think either is a concern but happy to change it!

josevalim: Yeah, i don't want compilation to break because we can't emit an optional warning, so I think `<-` is a better call.

josevalim:

PR #9901, opened 2020-03-20
Allowing &(unquote(f)/1) syntax

am-kantox:
Consider the following code.
```elixir
defmodule M do
  for fun <- [:fun1] do
    def capture(unquote(fun)), do: &(__MODULE__.unquote fun)/0
  end

  def fun1(), do: 42
end

M.capture(:fun1)
#⇒ &M.fun1/0
M.capture(:fun1).()
#⇒ 42
```

Unfortunately, currently, it works with fully-qualified function names, thanks to the dedicated `do_quote_call/6` in `elixir_quote.erl`, but fails to compile with local functions. 

```elixir
defmodule M do
  for fun <- [:fun1] do
    def capture(unquote(fun)), do: &(unquote fun)/0
  end

  def fun1(), do: 42
end

** (CompileError) iex:15: invalid args for &, 
   expected an expression in the format of
   &Mod.fun/arity, &local/arity or a capture 
   containing at least one argument as &1,
   got: :fun1 / 0
```

This PR fixes the issue by explicit handling of `{'&', Meta, [{'/' ...` in the more-or-less the same manner as `do_quote_call/6` does for dots.

### DIFF
--- a/lib/elixir/lib/kernel/special_forms.ex
+++ b/lib/elixir/lib/kernel/special_forms.ex
@@ -1351,8 +1351,6 @@ defmodule Kernel.SpecialForms do
   """
   defmacro unquote(:unquote_splicing)(expr), do: error!([expr])
 
-  defmacro unquote(:unquote_var)(expr), do: error!([expr])
-
   @doc ~S"""
   Comprehensions allow you to quickly build a data structure from
   an enumerable or a bitstring.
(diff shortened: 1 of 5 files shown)


josevalim:
The reason why we rewrite it is because the only option for remote calls to be valid AST is for them to be an atom. In your case though, `&:foo/1` is valid AST, although not valid semantically, and it shouldn't be the job of quote to rule semantics out. For example, `&:foo/1` could be valid in a DSL such as Ecto.

The correct fix in your case is to emit proper AST:

    def capture(unquote(fun)), do: &unquote(Macro.var(fun, __MODULE__))/0

am-kantox:
Makes a perfect sense, thank you. I was under wrong impression that `&:foo/1` is _semantically_ incorrect.

Issue #9786, opened 2020-02-05
Allow @requirements to be listed on Mix tasks


josevalim:
Today, if you want to define and use a Mix task in your project itself, you need to make sure to call "compile" or "app.start" before. The issue with this approach is that, if you update the task itself, calling "compile" will re-compile the task but not update the version currently running.

In other words, this is not enough:

```elixir
defmodule Mix.Tasks.MyTask do
  use Mix.Task

  def run(argv) do
    Mix.Task.run("compile")
    # task code
  end
end
```

To fix it, you need to do:

```elixir
defmodule Mix.Tasks.MyTask do
  use Mix.Task

  def run(argv) do
    Mix.Task.run("compile")
    __MODULE__.compiled_run(args)
  end

  def compiled_run(args) do
    # task code
  end
end
```

To make this easier, my suggestion is to add a @requirements annotation of tasks that should be run before the current one. So the outcome is this:

```elixir
defmodule Mix.Tasks.MyTask do
  use Mix.Task

  @requirements "compile"
  # or: @requirements ["compile", "app.start"]

  def run(argv) do
    # task code
  end
end
```

It is a small amount of indirection but it will yield actual improvements to those writing their own tasks.

smaximov:
> The issue with this approach is that, if you update the task itself, calling "compile" will re-compile the task but not update the version currently running

Oh, I shot myself in the foot because of this several times before, thank you for the workaround 👍 

> So the outcome is this

What if I want to run a mix task with arguments as a requirement, how would I be supposed to specify it? I see several options:

* `"compile --warnings-as-errors"` &mdash; need to split this string to separate task name from arguments, probably need to deal with shell quoting/escaping.
* `["compile", "--warnings-as-errors"]` &mdash; this would conflict with the suggested way of specifying multiple requirements.
* `{"compile", ["--warnings-as-errors"]}` &mdash; my favourite so far.

wojtekmach:
> `"compile --warnings-as-errors"` — need to split this string to separate the task name from the arguments, probably need to deal with shell quoting/escaping while splitting too.

this syntax is the same as in aliases:

```elixir
defp aliases() do
  [
    test: ["compile --warnings-as-errors", "test"]
  ]
end
```

so that would be my first choice but option 4. sounds good to me too.

smaximov:
> this syntax is the same as in aliases

[...]

PR #9905, opened 2020-03-21
Adding @requirements to Mix Task 

jdanielnd:
This is my first attempt to contribute to elixir :) So, please, let me know if it's on the right direction.

The PR tries to address #9786. 

- I'm not sure if it's running the requirements on the right place.
- Also, I'm not sure how to test the behavior described on the issue. I added requirements that write files so I can test that. But I believe the best approach would be to test the "compile" requirement described on the issue.

### DIFF
--- a/lib/mix/lib/mix/task.ex
+++ b/lib/mix/lib/mix/task.ex
@@ -314,7 +314,7 @@ defmodule Mix.Task do
 
     cond do
       alias && Mix.TasksServer.run({:alias, task, proj}) ->
-        res = run_alias(List.wrap(alias), args, :ok)
+        res = run_alias(List.wrap(alias), args, task, :ok)
         Mix.TasksServer.put({:task, task, proj})
         res
 
@@ -384,18 +384,24 @@ defmodule Mix.Task do
     end
   end
 
-  defp run_alias([h | t], alias_args, _res) when is_binary(h) do
-    [task | args] = OptionParser.split(h)
-    res = Mix.Task.run(task, join_args(args, alias_args, t))
-    run_alias(t, alias_args, res)
+  defp run_alias([h | t], alias_args, original_task, _res) when is_binary(h) do
+    case OptionParser.split(h) do
+      [^original_task | args] ->
+        res = Mix.Task.run(original_task, args ++ alias_args)
+        run_alias(t, [], original_task, res)
+
+      [task | args] ->
+        res = Mix.Task.run(task, join_args(args, alias_args, t))
+        run_alias(t, alias_args, original_task, res)
+    end
   end
 
-  defp run_alias([h | t], alias_args, _res) when is_function(h, 1) do
+  defp run_alias([h | t], alias_args, original_task, _res) when is_function(h, 1) do
     res = h.(join_args([], alias_args, t))
-    run_alias(t, alias_args, res)
+    run_alias(t, alias_args, original_task, res)
   end
 
-  defp run_alias([], _alias_task, res) do
+  defp run_alias([], _alias_task, _original_task, res) do
     res
   end
 
--- a/lib/mix/lib/mix/task.ex
+++ b/lib/mix/lib/mix/task.ex
@@ -222,10 +222,8 @@ defmodule Mix.Task do
   """
   @spec requirements(task_module) :: []
   def requirements(module) when is_atom(module) do
-    case List.keyfind(module.__info__(:attributes), :requirements, 0) do
-      {:requirements, requirements} -> requirements
-      _ -> nil
-    end
+    {:requirements, requirements} = List.keyfind(module.__info__(:attributes), :requirements, 0, {:requirements, []})
+    requirements
   end
 
   @doc """
--- a/lib/mix/lib/mix/task.ex
+++ b/lib/mix/lib/mix/task.ex
@@ -377,10 +377,10 @@ defmodule Mix.Task do
 
   def run_requirements(module) do
     if requirements = requirements(module) do
-        Enum.each(requirements, fn requirement ->
-          [task | args] = String.split(requirement)
-          Mix.Task.run(task, args)
-        end)
+      Enum.each(requirements, fn requirement ->
+        [task | args] = String.split(requirement)
+        Mix.Task.run(task, args)
+      end)
     end
   end
 
--- a/lib/mix/lib/mix/task.ex
+++ b/lib/mix/lib/mix/task.ex
@@ -41,21 +41,33 @@ defmodule Mix.Task do
   publicly visible on `mix help`. Omit this attribute if you do
   not want your task to be listed on `mix help`.
 
-  After creating a task module, run `mix compile` and any public
-  tasks should become visible to `mix help`.
+  If a task has requirements, they can be listed using the
+  `@requirements` attribute. For example:
+  
+      @requirements ["compile"]
+  
+  Tasks typically depend on the `"compile"` task, when they need
+  to access code from the current project, or the "app.start" task,
+  which compiles and starts the current app:
+  
+      @requirements ["app.start"]
+
+  Tasks can also be run directly by using `run/2`.
 
   ## Attributes
 
   There are a few attributes available in Mix tasks to
   configure them in Mix:
 
-    * `@shortdoc`  - makes the task public with a short description that appears on `mix help`
+    * `@shortdoc`  - makes the task public with a short description that appears
+      on `mix help`
     * `@recursive` - runs the task recursively in umbrella projects
-    * `@preferred_cli_env` - recommends environment to run task. It is used in absence of
-      a Mix project recommendation, or explicit `MIX_ENV`, and it only works for tasks
-      in the current project. `@preferred_cli_env` is not loaded from dependencies as
-      we need to know the environment before dependencies are loaded.
-    * `@requirements` - list of required tasks to be run before the task.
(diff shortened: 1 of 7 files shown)


lib/mix/lib/mix/task.ex
```diff
@@ -360,6 +375,19 @@ defmodule Mix.Task do
...
 
+  def run_requirements(module) do
+    case Mix.Task.requirements(module) do
+      nil ->
+        nil
+
+      requirements ->
+        Enum.each(requirements, fn requirement ->
+          [task | args] = String.split(requirement)
+          Mix.Task.rerun(task, args)
```
josevalim:
```suggestion
          Mix.Task.run(task, args)
```

lib/mix/lib/mix/task.ex
```diff
@@ -360,6 +375,19 @@ defmodule Mix.Task do
     end
   end
 
+  def run_requirements(module) do
+    case Mix.Task.requirements(module) do
+      nil ->
+        nil
+
+      requirements ->
```
josevalim: This is probably clearer as an `if`:

```suggestion
    if requirements = requirements(module) do
```

lib/mix/test/test_helper.exs
```diff
@@ -393,6 +393,7 @@ end
 defmodule Mix.Tasks.Hello do
   use Mix.Task
   @shortdoc "This is short documentation, see"
+  @requirements ["cmd touch file1.txt", "cmd touch file2.txt"]
```
josevalim: Instead of changing an existing task, define a new task in `mix/task_test.exs`. :) This particular task is used in many places. Also, it is best to not use `cmd`, since `touch` is not available on Windows. Maybe use `["help compile", "hello"]` and use `capture_io` to make sure the contents have been printed. Or similar.

jdanielnd:
Although the requirements are running (as the test shows), I tried to replace `Mix.Task.run("compile", args)` to `@requirements ["compile"]` at `mix/tasks/app.start.ex` and it didn't seemed to work.

I'll need to dig further, but the `run_requirements/1` might not be on the right place.

jdanielnd:
Got it: the `run_requirements/1` only runs if the task is called from `Mix.Task.run('task_name')`. It doesn't run if it's called from `Mix.Tasks.TaskName.run(args)`.  Trying to work it out.

lib/mix/lib/mix/task.ex
```diff
@@ -215,6 +215,19 @@ defmodule Mix.Task do
...
+  Gets requirements for the task.
+
+  Returns list with requirements (for example, `["compile"]`, or `["compile --warning-as-errors"]`), or `["compile", "app.start"]`.
+  """
+  @spec requirements(task_module) :: []
+  def requirements(module) when is_atom(module) do
+    case List.keyfind(module.__info__(:attributes), :requirements, 0) do
+      {:requirements, requirements} -> requirements
+      _ -> nil
+    end
```
whatyouhide: Since we're running through the requirements later, what if we go with defaulting to an empty list of requirements? I think that nicely signals that this task has no requirements.

```suggestion
    {:requirements, requirements} = List.keyfind(module.__info__(:attributes), :requirements, 0, {:requirements, []})
    requirements
```

lib/mix/lib/mix/task.ex
```diff
@@ -360,6 +375,15 @@ defmodule Mix.Task do
     end
   end
 
+  def run_requirements(module) do
+    if requirements = requirements(module) do
+      Enum.each(requirements, fn requirement ->
+        [task | args] = String.split(requirement)
+        Mix.Task.run(task, args)
+      end)
+    end
```
whatyouhide: If we go with the empty list of requirements as the default, this would change and the `if` would be gone:

```suggestion
    Enum.each(requirements(module), fn requirement ->
      [task | args] = String.split(requirement)
      Mix.Task.run(task, args)
    end)
```

I personally think this is a nice improvement but it's not a blocking thing, we can ship this PR regardless :)

lib/mix/lib/mix/task.ex
josevalim: I like this idea!

josevalim:
> Got it: the run_requirements/1 only runs if the task is called from Mix.Task.run('task_name'). It doesn't run if it's called from Mix.Tasks.TaskName.run(args).

This is expected behaviour, do not worry about it. :)

lib/mix/lib/mix/task.ex
jdanielnd: Great approach! :)

lib/mix/test/test_helper.exs
```diff
@@ -412,6 +412,21 @@ defmodule Mix.Tasks.Hello do
...
+
+  @moduledoc """
+  A test task.
+  """
+
+  def run(_args) do
+    "Task with requirements"
+  end
+end
+|> write_beam.()
```
josevalim: I think you can define it within the `mix/task_test.exs`, without the `write_beam.()` code, and it should all work. Have you tried it?

lib/mix/test/mix/task_test.exs
```diff
@@ -237,4 +237,15 @@ defmodule Mix.TaskTest do
...
+
+  test "requirements/1" do
+    assert Mix.Task.requirements(Mix.Tasks.WithRequirement) == ["help compile"]
+  end
+
+  test "run_requirements/1 is run during task execution" do
+    assert String.contains?(
+             ExUnit.CaptureIO.capture_io(fn -> Mix.Task.run("with_requirement") end),
+             "compile"
+           )
```
josevalim:
```suggestion
    assert ExUnit.CaptureIO.capture_io(fn ->
      assert Mix.Task.run("with_requirement") end) == "Task with requirements"
    end) =~ "mix compile"
```

lib/mix/test/test_helper.exs
jdanielnd: It works but the code must be outside `Mix.TaskTest` module definition, otherwise the namespace gets messed up. It creates de module under `Mix.TaskTest.Mix.Tasks.WithRequirement`.

It doesn't seem natural to have this module definition under the same file of `Mix.TaskTest`, but outside its definition. Any suggestions?

josevalim: You can move it to the top of the file or name it `defmodule Elixir.Mix.Tasks.WithRequirement do`. :)

jdanielnd:
1. **Different behaviors depending on caller**

   > This is expected behaviour, do not worry about it. :)

   Is it OK for the requirements to only run when called by `Mix.Task.run`? I may not see the big picture, but it seems odd to have different behaviors. Won't that create umpredictable behavior for users?

   Also, this implementation won't allow replacing the `Mix.Task.run("compile", args)` from `app.start.ex`. All this module tests are using `Mix.Tasks.App.Start.run`. Do we want to replace that in this module?

2. **Only allow predefined args**

   Current implementation only allows predefined arguments defined on `@requirements`. It doesn't allow the use of args coming from command line, such as used on `app.start.ex`: `Mix.Task.run("compile", args)`. Is it ok?

I'm also missing a test case that reproduces exactly the use case described on #9786. I'm still not sure that running `compile` through the `@requirements` will update the code version currently running. Should I dig further on that?

lib/mix/test/mix/task_test.exs
```diff
@@ -237,4 +237,14 @@ defmodule Mix.TaskTest do
...
   end
+
+  test "requirements/1" do
+    assert Mix.Task.requirements(Mix.Tasks.WithRequirement) == ["help compile"]
+  end
+
+  test "run_requirements/1 is run during task execution" do
+    assert ExUnit.CaptureIO.capture_io(fn ->
+             assert Mix.Task.run("with_requirement") == "Task with requirements"
+           end) =~ "mix compile"
```
josevalim: We also need to test that requirements won't run again if the task has already run:

```suggestion
           end) =~ "mix compile"

    Mix.Task.reenable("help")
    
    assert ExUnit.CaptureIO.capture_io(fn ->
             assert Mix.Task.run("with_requirement") == :noop
           end) == ""
```

josevalim:
> Is it OK for the requirements to only run when called by Mix.Task.run?

Yes, because you are not supposed to invoke tasks directly. Invoking tasks directly won't dispatch aliases, it won't avoid re-running them, etc. The behaviour is already drastically different which is why you are not supposed to invoke directly.

> Only allow predefined args

If you want to allow custom args, then don't use requirements.

You don't need to worry about the original case though. Your unit tests are enough. I would just improve your current unit test to show that requirements do not run when running the task again.

jdanielnd:
> You don't need to worry about the original case though. Your unit tests are enough. I would just improve your current unit test to show that requirements do not run when running the task again.

Done :)

Just added a one-liner doc explaining the attribute on `Mix.Task` module. Is there any other docs that should be updated?

_Happy for the 1st contribution :D_

josevalim:

lib/mix/lib/mix/task.ex
```diff
@@ -360,6 +389,13 @@ defmodule Mix.Task do
     end
   end
 
+  def run_requirements(module) do
+    Enum.each(requirements(module), fn requirement ->
+      [task | args] = String.split(requirement)
```
smaximov: Shouldn't it be `OptionParser.split`?

josevalim: Great catch, fixed on master!

PR #9907, opened 2020-03-23
Allow use of MFA tuple and arguments in child_spec

hauleth:
This adds support for:

- `{Agent, {m, f, a}}`
- `{Agent, {func, opts}}`
- `{Agent, {{m, f, a}, opts}}`

In supervisor child specs

### DIFF
--- a/lib/elixir/lib/agent.ex
+++ b/lib/elixir/lib/agent.ex
@@ -191,12 +191,21 @@ defmodule Agent do
   See the "Child specification" section in the `Supervisor` module for more detailed information.
   """
   @doc since: "1.5.0"
-  def child_spec(arg) do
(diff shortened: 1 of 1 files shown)


josevalim:
I think beyond a very simple example, I prefer if folks define their own modules with their own `start_link` calls rather than making the initial `child_spec` call complex. Thanks!

hauleth:
@josevalim I had use for passing the options to the `Agent` in tests (as TBH I never used `Agent` outside tests/quick PoC) and I find it quite handy (`{Agent, {func, opts}}` to be exact, the rest I added just for uniformity).

josevalim:
Adding it because of tests is even more worrying, as people will also use it in production code. For tests, `%{id: Agent, start: {Agent, :start_link, [...]}}` is your friend!

PR #9915, opened 2020-03-28
Deprecate non-atom keys on put_env, get_env, fetch_env*, delete_env

wojtekmach:
Typespecs already required an atom key but it was never enforced by
code. We're doing this because we eventually would like to support
passing a _path_ like:

    Application.get_env(:myapp, [:key, :foo, :bar])

and this way we will be consistent with recently added
`Application.compile_env/3`.

Another way to solve this issue, not part of this PR, is deprecate
`get_env`, `fetch_env` and `fetch_env!` in favour of a new `env/3` &
`env!/2` or `runtime_env/3` & `runtime_env!/2`.  The downside is we won't
have the `fetch_` (non-raising) function but fwiw we don't have it for
`compile_env` either. (If we deprecate, it should be on a really long
cycle becasue they are so ubiquitous. But if we start printing
deprecation warnings in libraries it's probably for the better. ;-))

### DIFF
--- a/lib/elixir/lib/application.ex
+++ b/lib/elixir/lib/application.ex
@@ -618,6 +618,7 @@ defmodule Application do
   """
   @spec get_env(app, key, value) :: value
   def get_env(app, key, default \\ nil) when is_atom(app) do
+    maybe_warn_on_app_env_key(key)
     :application.get_env(app, key, default)
   end
 
@@ -628,6 +629,8 @@ defmodule Application do
   """
   @spec fetch_env(app, key) :: {:ok, value} | :error
   def fetch_env(app, key) when is_atom(app) do
+    maybe_warn_on_app_env_key(key)
+
     case :application.get_env(app, key) do
       {:ok, value} -> {:ok, value}
       :undefined -> :error
@@ -689,6 +692,7 @@ defmodule Application do
   """
   @spec put_env(app, key, value, timeout: timeout, persistent: boolean) :: :ok
   def put_env(app, key, value, opts \\ []) when is_atom(app) do
+    maybe_warn_on_app_env_key(key)
     :application.set_env(app, key, value, opts)
   end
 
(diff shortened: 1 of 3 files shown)


lib/elixir/lib/application.ex
```diff
@@ -734,9 +738,17 @@ defmodule Application do
...
+    maybe_warn_on_app_env_key(key)
     :application.unset_env(app, key, opts)
   end
 
+  # TODO: turn warning into error on Elixir v2.0
+  defp maybe_warn_on_app_env_key(key) when is_atom(key),
+    do: :ok
+
+  defp maybe_warn_on_app_env_key(key),
+    do: IO.warn("passing non-atom as application env key is deprecated (got: `#{inspect(key)}`)")
```
josevalim: Those are runtime warnings, so we probably want to emit them only once per key. See the private function `IO.warn_once` and how we use it in the codebase for more info.

lib/elixir/lib/application.ex
```diff
@@ -734,9 +738,19 @@ defmodule Application do
...
+    maybe_warn_on_app_env_key(app, key)
     :application.unset_env(app, key, opts)
   end
 
+  # TODO: turn warning into error on Elixir v2.0
+  defp maybe_warn_on_app_env_key(_app, key) when is_atom(key),
+    do: :ok
+
+  defp maybe_warn_on_app_env_key(app, key) do
+    message = "passing non-atom as application env key is deprecated (got: `#{inspect(key)}`)"
```
whatyouhide: We don't use markdown in error messages :) My suggestion:

```suggestion
    message = "passing non-atom as application env key is deprecated, got: #{inspect(key)}"
```

lib/elixir/lib/application.ex
```diff
@@ -738,16 +738,18 @@ defmodule Application do
...
 
   # TODO: turn warning into error on Elixir v2.0
-  defp maybe_warn_on_app_env_key(key) when is_atom(key),
+  defp maybe_warn_on_app_env_key(_app, key) when is_atom(key),
     do: :ok
 
-  defp maybe_warn_on_app_env_key(key),
-    do: IO.warn("passing non-atom as application env key is deprecated (got: `#{inspect(key)}`)")
+  defp maybe_warn_on_app_env_key(app, key) do
+    message = "passing non-atom as application env key is deprecated (got: `#{inspect(key)}`)"
```
josevalim: We don't use `` ` `` in Elixir error messages. AFAIK Ecto is the only project that does it. :-)

```suggestion
    message = "passing non-atom as application env key is deprecated, got: #{inspect(key)}"
```

wojtekmach: image

josevalim:

PR #9916, opened 2020-03-28
Update IO.warn_once/3

wojtekmach:


josevalim:
It can be an improvement but I think as-is is fine unless we need more args. Thanks!

PR #9920, opened 2020-04-02
Load vm-args for remote shell

jeffutter:
We're writing our own empd/distribution module, similar to https://www.erlang-solutions.com/blog/erlang-and-elixir-distribution-without-epmd.html .

In order to be able to attach to a remote console, we need to load the same `rel/vm.args` file as the running instance of the beam. I'm not sure if this would cause any unintended side effects.


josevalim:
We don't load it on purpose because you may set other configuration that you do not want to apply on remote. In your case, I recommend setting the option via the ELIXIR_ERL_OPTIONS env variable in your env script and that should apply to all types commands: https://github.com/elixir-lang/elixir/blob/master/lib/mix/lib/mix/tasks/release.init.ex#L63-L71

You can also match on the command and choose exactly which options to apply. Thanks!

PR #9922, opened 2020-04-03
Fixes various typos, improves function docs, emphasizes that tasks mu…

fireproofsocks:
I dug into the docs here after running into problems with [data persisting between tests](https://elixirforum.com/t/data-persisting-between-tests/29961) -- ultimately the problem was that I didn't realize that Mix tasks cannot be called repeatedly without re-enabling.  I had no idea that was a thing, and I didn't feel that enough emphasis was given to this functionality.

This PR:
- Fixes various typos
- Improves function docs
- Emphasizes that tasks must be re-enabled before repeated runs

### DIFF
--- a/lib/mix/lib/mix/task.ex
+++ b/lib/mix/lib/mix/task.ex
@@ -246,8 +246,10 @@ defmodule Mix.Task do
   Returns the task name for the given `module`.
 
   ## Examples
+
       iex> Mix.Task.task_name(Mix.Tasks.Test)
       "test"
+
   """
   @spec task_name(task_module) :: task_name
   def task_name(module) when is_atom(module) do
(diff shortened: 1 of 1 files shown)


lib/mix/lib/mix/task.ex
```diff
@@ -244,17 +244,24 @@ defmodule Mix.Task do
 
   @doc """
   Returns the task name for the given `module`.
+
+  ## Examples
```
josevalim, line 248:
```suggestion
  ## Examples

```

lib/mix/lib/mix/task.ex
```diff
@@ -244,17 +244,24 @@ defmodule Mix.Task do
...
+      iex> Mix.Task.task_name(Mix.Tasks.Test)
+      "test"
```
josevalim, line 250:
```suggestion
      "test"

```

lib/mix/lib/mix/task.ex
```diff
@@ -244,17 +244,24 @@ defmodule Mix.Task do
...
   @doc """
-  Checks if an alias called `task` exists.
+  Checks if the given `task` name is an alias.
+
+  Returns false if the given name is not an alias or if it is not a task.
 
-  For more information about task aliasing, take a look at the "Aliasing"
-  section in the docs for `Mix`.
+  For more information about task aliasing, take a look at the
+  ["Aliasing"]((https://hexdocs.pm/mix/Mix.html#module-aliases) section in the
```
josevalim, line 263:
```suggestion
  ["Aliasing"](https://hexdocs.pm/mix/Mix.html#module-aliases) section in the
```

PR #9924, opened 2020-04-09
Fix tokenizer emitting repeated warnings in the REPL

### LINKED ISSUE #9892: Warning checker runs on every line when pasted into REPL
OTP22 Elixir 1.9.1

When you paste code into REPL it runs the warning checker on every line.

```
defmodule Test do
    def trigger() do
        atom = :"its_quoted"
        atom = :"its_quoted2"
    end

    def keep_triggering() do
        "The warning above will trigger on every evaluated line"
    end
end
```

pablocostass:
The `elixir_tokenizer` record now has a list where the warnings are stored in. It is returned upon tokenization success after filtering out duplicated warnings.

Nonetheless, the warnings are still being emitted as before, they are just being collected for now.

This will close #9892 after the warnings are emitted somewhere else.

### DIFF
--- a/lib/elixir/src/elixir_tokenizer.erl
+++ b/lib/elixir/src/elixir_tokenizer.erl
@@ -144,8 +144,8 @@ tokenize(String, Line, Opts) ->
   tokenize(String, Line, 1, Opts).
 
 tokenize([], _Line, _Column, #elixir_tokenizer{terminators=[], warnings = Warnings}, Tokens) ->
-  UniqueWarnings = ordsets:to_list(ordsets:from_list(Warnings)),
-  {ok, {lists:reverse(Tokens), UniqueWarnings}};
+  [elixir_errors:erl_warn(Line, File, Msg) || {Line, File, Msg} <- lists:reverse(Warnings)],
+  {ok, lists:reverse(Tokens)};
 
 tokenize([], EndLine, Column, Scope, Tokens) ->
   #elixir_tokenizer{terminators=[{Start, StartLine, _} | _]} = Scope,
@@ -1460,5 +1460,4 @@ maybe_warn_for_ambiguous_bang_before_equals(_Kind, _Atom, _Rest, Scope, _Line) -
   Scope.
 
 prepend_warning({Line, File, Msg}, Scope) ->
-  elixir_errors:erl_warn(Line, File, Msg),
   Scope#elixir_tokenizer{warnings = [{Line, File, Msg} | Scope#elixir_tokenizer.warnings]}.
--- a/lib/elixir/src/elixir_tokenizer.erl
+++ b/lib/elixir/src/elixir_tokenizer.erl
@@ -144,7 +144,7 @@ tokenize(String, Line, Opts) ->
   tokenize(String, Line, 1, Opts).
 
 tokenize([], _Line, _Column, #elixir_tokenizer{terminators=[], warnings = Warnings}, Tokens) ->
-  UniqueWarnings = sets:to_list(sets:from_list(Warnings)),
+  UniqueWarnings = ordsets:to_list(ordsets:from_list(Warnings)),
   {ok, {lists:reverse(Tokens), UniqueWarnings}};
 
 tokenize([], EndLine, Column, Scope, Tokens) ->
@@ -241,10 +241,7 @@ tokenize([$?, Char | T], Line, Column, Scope, Tokens) ->
     {Escape, Name} ->
       Msg = io_lib:format("found ? followed by code point 0x~.16B (~ts), please use ?~ts instead",
                           [Char, Name, Escape]),
-      elixir_errors:erl_warn(Line, Scope#elixir_tokenizer.file, Msg),
-      Scope#elixir_tokenizer{warnings =
-        Scope#elixir_tokenizer.warnings ++ [{Line, Scope#elixir_tokenizer.file, Msg}]
-      };
+      prepend_warning({Line, Scope#elixir_tokenizer.file, Msg}, Scope);
     false ->
       Scope
   end,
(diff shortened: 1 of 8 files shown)


pablocostass:
Hey @josevalim, I have for now modified the tests so that they won't fail due to pattern matching because of the newly added warnings. Should I modified them to do some extra assertions with the list of warnings?

Also, I did as you suggested and as a first step only collected said warnings, could you guide me on moving the emitting of them elsewhere?

lib/elixir/src/elixir_tokenizer.erl
```diff
@@ -143,8 +143,9 @@ tokenize(String, Line, Column, Opts) ->
 tokenize(String, Line, Opts) ->
   tokenize(String, Line, 1, Opts).
 
-tokenize([], _Line, _Column, #elixir_tokenizer{terminators=[]}, Tokens) ->
-  {ok, lists:reverse(Tokens)};
+tokenize([], _Line, _Column, #elixir_tokenizer{terminators=[], warnings = Warnings}, Tokens) ->
+  UniqueWarnings = sets:to_list(sets:from_list(Warnings)),
+  {ok, {lists:reverse(Tokens), UniqueWarnings}};
```
josevalim, line 148: I don't think we need to make them unique, they are not meant to repeat. Let's also return a three element tuple:

```suggestion
  {ok, lists:reverse(Tokens), Warnings};
```

lib/elixir/src/elixir_tokenizer.erl
```diff
@@ -236,16 +237,19 @@ tokenize([$?, $\\, H | T], Line, Column, Scope, Tokens) ->
...
 tokenize([$?, Char | T], Line, Column, Scope, Tokens) ->
-  case handle_char(Char) of
+  NewScope = case handle_char(Char) of
     {Escape, Name} ->
       Msg = io_lib:format("found ? followed by code point 0x~.16B (~ts), please use ?~ts instead",
                           [Char, Name, Escape]),
-      elixir_errors:erl_warn(Line, Scope#elixir_tokenizer.file, Msg);
+      elixir_errors:erl_warn(Line, Scope#elixir_tokenizer.file, Msg),
+      Scope#elixir_tokenizer{warnings =
+        Scope#elixir_tokenizer.warnings ++ [{Line, Scope#elixir_tokenizer.file, Msg}]
```
josevalim, line 246: Instead of appending, consider prepending the warnings:

```elixir
         [{Line, Scope#elixir_tokenizer.file, Msg} | Scope#elixir_tokenizer.warnings]
```

Then when you return it, you do `lists:reverse(Warnings)`.

Also, it probably makes sense to have a function called `prepend_warning({Line, File, Msg}, Scope)` to reduce the duplication. For now, the `prepend_warning` function can call `elixir_errors:erl_warn(Line, Scope#elixir_tokenizer.file, Msg)` and prepend the warning. This way, once we have to stop emitting the warning, we need to do it only once.

lib/elixir/src/elixir_tokenizer.erl
josevalim: Actually, now that I think about it, we don't need to return the warnings, we can just emit them here. So you can do something like this:

```suggestion
  [elixir_errors:erl_warn(Line, File, Msg) || {Line, File, Msg} <- lists:reverse(Warnings)],
  {ok, lists:reverse(Tokens)};
```

The reverse is necessary if you implement the `prepend_warnings` function I mentioned below.

lib/elixir/src/elixir_tokenizer.erl
pablocostass: Sure. No need to reverse the warning list though, as you can get an ordered one and without the duplicated warnings by using `ordsets` instead of `sets`, as I was doing.

lib/elixir/src/elixir_tokenizer.erl
```diff
@@ -143,8 +143,9 @@ tokenize(String, Line, Column, Opts) ->
 tokenize(String, Line, Opts) ->
   tokenize(String, Line, 1, Opts).
 
-tokenize([], _Line, _Column, #elixir_tokenizer{terminators=[]}, Tokens) ->
-  {ok, lists:reverse(Tokens)};
+tokenize([], _Line, _Column, #elixir_tokenizer{terminators=[], warnings = Warnings}, Tokens) ->
+  UniqueWarnings = ordsets:to_list(ordsets:from_list(Warnings)),
```
josevalim, line 147: There is no reason to use ordsets, lists:reverse/2 is the best way to go here. The warnings should be unique by defiition too. No need to try to unify them.

Finally, I think we can do the following here:

```erl
[elixir_errors:erl_warn(Line, File, Msg) || {Line, File, Msg} <- lists:reverse(Warnings)],
```

By doing so we no longer need to return the warnings and we can remove the call from `prepend_warning`. This change should be enough to fix the original issue!

josevalim:
Looks great, now we just need a test. Perhaps something in kernel/warnings_test.exs that looks like this:

```
test "does not warn on incomplete tokenization" do
  assert {:error, _} = Code.string_to_quoted(~s[:"foobar" do])
end
```

WDYT?

pablocostass:
That would only check if it indeed returns an error, but doesn't check if it is printed. Looking at the existing tests in that file, I was thinking of something like this:

```elixir
test "does not warn on incomplete tokenization" do
    expected_error = {:error, {1, "missing terminator: end (for \"do\" starting at line 1)", ""}}
    output =
      capture_err(fn ->
        assert expected_error == Code.string_to_quoted(~s[:"foobar" do])
      end)

    assert output == ""
  end
```

josevalim:
We don’t allow warnings in the test suite, so there is no need to wrap it in capture_err. Plus capture_err is concurrent and it may capture the output of other tests.

pablocostass:
Excuse my ignorance, but as the test suite is full of tests calling `capture_err` I'm having some doubts.

Although the warnings are disabled in the test suite, `:stderr` still works (i.e., changing in my code snippet the last assertion to `assert output == "not empty"` will make it fail).

Also, as the test case isn't being run with `async: true`, there shouldn't be a problem of capturing the output of other tests (hence the test suite being full of those calls to `capture_err`). Or am I missing something?

PS: Thanks for bearing with me José :)

josevalim:
If the test is not async true than that’s on. But there is no need to use capture_err anyway. :)

pablocostass:
Gotcha, will leave it as you said then.

josevalim:

PR #9928, opened 2020-04-11
Document that File.cd/1 affects the BEAM globally

aaronrenner:
I was surprised that `File.cd/1` affects the BEAM globally, instead of just changing the directory for the current process. This adds some documentation regarding this behavior.


josevalim:

PR #9934, opened 2020-04-13
Improve map merge method

caiodsc:
I found out its faster than original implementation by 2K ips.


josevalim:
I would honestly not expect this difference. Can you please provide the benchmark scripts so we can take a better look at it? Thank you!

PR #9936, opened 2020-04-14
Make spec of `c:run/1` match the documentation

hauleth:
Earlier it failed Dialyzer tests as the example in docs returned value
different from one expected by callback definition.


josevalim:

PR #9952, opened 2020-04-19
use __MODULE__ alias

sushant12:
Since `@name` is assigned IEx.Broker, I figured using the `__MODULE__` alias would make it dynamic


josevalim:

PR #9961, opened 2020-04-25
Fix dialyzer errors caused by erl_anno:anno type being opaque

### LINKED ISSUE #9948: Unexpected type error
In some cases `Code.eval_quoted` causes unexpected and weird Dialyzer errors. 
It might be caused by bug in Erlang, Elixir or Dialyzer itself. 
But it's definitely indicates what something is wrong somewhere.
Sample project and details how to reproduce bug are there
https://github.com/tim2CF/dialyzer_bug#dialyzerbug

michallepicki:
- Use erl_anno:new instead of integers for line numbers when generating erlang AST (fixes #9948 )
- Remove unnecessary ?ann macro call

### DIFF
--- a/lib/elixir/src/elixir_compiler.erl
+++ b/lib/elixir/src/elixir_compiler.erl
@@ -85,15 +85,15 @@ code_mod(Fun, Expr, Line, File, Module, Vars) when is_binary(File), is_integer(L
   Tuple = {tuple, Line, [{var, Line, Var} || {_, Var} <- Vars]},
   Relative = elixir_utils:relative_to_cwd(File),
 
-  [{attribute, Line, file, {elixir_utils:characters_to_list(Relative), 1}},
-   {attribute, Line, module, Module},
-   {attribute, Line, compile, no_auto_import},
-   {attribute, Line, export, [{Fun, 1}, {'__RELATIVE__', 0}]},
-   {function, Line, Fun, 1, [
-     {clause, Line, [Tuple], [], [Expr]}
+  [{attribute, erl_anno:new(Line), file, {elixir_utils:characters_to_list(Relative), 1}},
+   {attribute, erl_anno:new(Line), module, Module},
+   {attribute, erl_anno:new(Line), compile, no_auto_import},
+   {attribute, erl_anno:new(Line), export, [{Fun, 1}, {'__RELATIVE__', 0}]},
+   {function, erl_anno:new(Line), Fun, 1, [
+     {clause, erl_anno:new(Line), [Tuple], [], [Expr]}
    ]},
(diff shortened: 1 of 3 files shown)


lib/elixir/src/elixir_compiler.erl
```diff
@@ -85,15 +85,15 @@ code_mod(Fun, Expr, Line, File, Module, Vars) when is_binary(File), is_integer(L
   Tuple = {tuple, Line, [{var, Line, Var} || {_, Var} <- Vars]},
```
michallepicki, line 85: I should probably change this line as well. I'm wondering if there are other spots that are not getting caught by dialyzer like this one.

josevalim:
Thank you! I will wait for a conclusion from the OTP team, to let us know if line numbers are opaque or not, and then act accordingly. But if we do so, we will need to consistently use erl_anno everywhere. :)

michallepicki:
This brings the number of dialyzer issues reported in the elixir application itself on erlang 22.3 from 182 to 100 :)

I'll look at other issues (duplicate or wrong specs) and try to send separate PRs for those.

lib/elixir/src/elixir_erl.erl
```diff
@@ -68,31 +68,31 @@ invoke_local(Module, ErlName, Args) ->
...
   %% are not changed in any fashion.
-  {bin, 0, [{bin_element, 0, {string, 0, binary_to_list(Tree)}, default, default}]};
+  {bin, erl_anno:new(0), [{bin_element, erl_anno:new(0), {string, erl_anno:new(0), binary_to_list(Tree)}, default, default}]};
 elixir_to_erl(Pid) when is_pid(Pid) ->
   ?remote(0, erlang, binary_to_term, [elixir_erl:elixir_to_erl(term_to_binary(Pid))]);
 elixir_to_erl(_Other) ->
   error(badarg).
 
-elixir_to_erl_cons([H | T]) -> {cons, 0, elixir_to_erl(H), elixir_to_erl_cons(T)};
+elixir_to_erl_cons([H | T]) -> {cons, erl_anno:new(0), elixir_to_erl(H), elixir_to_erl_cons(T)};
```
josevalim, line 95: What if we put those `erl_anno:new(0)` into a macro? `?zero_ann` or similar? Then if we need to change it, we can change it once.

michallepicki: Sure, where should I put it? Next to `?ann`?

josevalim:

lib/elixir/src/elixir_erl.erl
josevalim: I had an idea of another approach, I just pushed it to master. Thank you very much for looking into this.

PR #9963, opened 2020-04-25
Fix erroneous code snippet in documentation

linduxed:
The return value in the code example is the opposite of what the code would actually return, making the example misleading.


fertapric:
Good catch @linduxed! Thanks 💜

PR #9969, opened 2020-04-26
Fix `mix cmd` examples in its moduledoc

dsdshcym:
- `mix cmd echo pwd` would only print literally `"pwd"` for each child app
- `mix cmd pwd` would print each child app's path


josevalim:

PR #9974, opened 2020-04-26
enable overriding sys.config location via env var

dch:
I am distributing a package of our app, and to keep it free from
sensitive information, I generate an OS package from the
release tarball, without embedded tokens & passwords.

The cookie, vm.args, and sys.config are provided separately
at runtime, via dev-ops tools.

For `COOKIE` & `vm.args`, this is already possible, but not for
`sys.config`.

I'd like to be able to override `RELEASE_SYS_CONFIG` if
already set, the same as `RELEASE_VM_ARGS` works today.

This PR seems sufficient for UNIX but my windows patch
doesn't handle dynamic `RUNTIME_CONFIG=true`, I'm
not familiar anymore with windows batch scripting, if
somebody could advise that would be great.

### DIFF
--- a/lib/mix/lib/mix/tasks/release.ex
+++ b/lib/mix/lib/mix/tasks/release.ex
@@ -749,6 +749,9 @@ defmodule Mix.Tasks.Release do
       It can be set to a custom value. The name part must be made only
       of letters, digits, underscores, and hyphens
 
+    * `RELEASE_SYS_CONFIG` - the location of the sys.config file. It can
+    be set to a custom path
+
     * `RELEASE_VM_ARGS` - the location of the vm.args file. It can be set
       to a custom path
 
--- a/lib/mix/lib/mix/tasks/release.ex
+++ b/lib/mix/lib/mix/tasks/release.ex
@@ -750,7 +750,7 @@ defmodule Mix.Tasks.Release do
       of letters, digits, underscores, and hyphens
 
     * `RELEASE_SYS_CONFIG` - the location of the sys.config file. It can
-    be set to a custom path
+      be set to a custom path and it must not include the `.config` extension
 
     * `RELEASE_VM_ARGS` - the location of the vm.args file. It can be set
       to a custom path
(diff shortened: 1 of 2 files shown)


lib/mix/lib/mix/tasks/release.init.ex
```diff
@@ -284,7 +284,7 @@ defmodule Mix.Tasks.Release.Init do
     if not defined RELEASE_DISTRIBUTION (set RELEASE_DISTRIBUTION=sname)
     if not defined RELEASE_BOOT_SCRIPT (set RELEASE_BOOT_SCRIPT=start)
     if not defined RELEASE_BOOT_SCRIPT_CLEAN (set RELEASE_BOOT_SCRIPT_CLEAN=start_clean)
-    set RELEASE_SYS_CONFIG=!REL_VSN_DIR!\sys
+    if not defined RELEASE_SYS_CONFIG (set RELEASE_SYS_CONFIG=!REL_VSN_DIR!\sys)
```
dch, line 287: NB I can't test this and I don't believe it works appropriate in dynamic `RUNTIME_CONFIG=true` scenario

josevalim:
Hi @dch! You will see that we do `$REL_VSN_DIR/sys.config` in other places. So what you will need to do, in regards to those references, is to, if `RUNTIME_CONFIG=true` is set, you will have to set a `RELEASE_ORIGINAL_SYS_CONFIG` before you change `RELEASE_SYS_CONFIG`, and refer to `RELEASE_ORIGINAL_SYS_CONFIG` instead of `$REL_VSN_DIR/sys.config`. If you give me access to your fork, I can also push those changes in.

lib/mix/lib/mix/tasks/release.ex
```diff
@@ -749,6 +749,9 @@ defmodule Mix.Tasks.Release do
       It can be set to a custom value. The name part must be made only
       of letters, digits, underscores, and hyphens
 
+    * `RELEASE_SYS_CONFIG` - the location of the sys.config file. It can
+    be set to a custom path
```
josevalim, line 753:
```suggestion
      be set to a custom path and it must not include the `.config` extension
```

josevalim:
I have pushed my changes. If the CI passes, then this is good to go!

josevalim:

PR #9976, opened 2020-04-26
Inline types not exported from their erlang modules

michallepicki:
This commit fixes the following errors from dialyzer analysis:
```
Unknown types:
  beam_lib:chnk_rsn/0
  erlang:message_queue_data/0
  supervisor:modules/0
  supervisor:restart/0
  supervisor:shutdown/0
```

I'm not sure if this is the right way to fix these. In my view, using unexported types is similar to accessing private APIs. Maybe there is no reason for these types not to be exported and this should be fixed upstream by exporting them? Or maybe this is a hint that Elixir is indeed accessing "private" APIs in some of these cases? Or maybe this approach is fine?

### DIFF
--- a/lib/mix/lib/mix/release.ex
+++ b/lib/mix/lib/mix/release.ex
@@ -777,7 +777,21 @@ defmodule Mix.Release do
   The exact chunks that are kept are not documented and may change in
   future versions.
   """
-  @spec strip_beam(binary()) :: {:ok, binary} | {:error, :beam_lib, :beam_lib.chnk_rsn()}
+  @type chunkid :: nonempty_charlist()
+  @type info_rsn() ::
+          {:chunk_too_big, :file.filename(), chunkid(), chunk_size :: non_neg_integer(),
+           file_size :: non_neg_integer()}
+          | {:invalid_beam_file, :file.filename(), position :: non_neg_integer()}
+          | {:invalid_chunk, :file.filename(), chunkid()}
+          | {:missing_chunk, :file.filename(), chunkid()}
+          | {:not_a_beam_file, :file.filename()}
+          | {:file_error, :file.filename(), :file.posix()}
+  @type chnk_rsn() ::
+          {:unknown_chunk, :file.filename(), atom()}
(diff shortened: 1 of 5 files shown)


lib/mix/lib/mix/release.ex
```diff
@@ -777,7 +777,21 @@ defmodule Mix.Release do
...
+          | {:invalid_chunk, :file.filename(), chunkid()}
+          | {:missing_chunk, :file.filename(), chunkid()}
+          | {:not_a_beam_file, :file.filename()}
+          | {:file_error, :file.filename(), :file.posix()}
+  @type chnk_rsn() ::
+          {:unknown_chunk, :file.filename(), atom()}
+          | {:key_missing_or_invalid, :file.filename(), :abstract_code | :debug_info}
+          | info_rsn()
+  # inlining chnk_rsn() here because :beam_lib.chnk_rsn() is not exported
+  @spec strip_beam(binary()) :: {:ok, binary} | {:error, :beam_lib, chnk_rsn()}
```
josevalim, line 794:
```suggestion
  @spec strip_beam(binary()) :: {:ok, binary} | {:error, :beam_lib, term}
```

josevalim:
I think those are good, thanks! But the chunk_rsn is leaking a bunch of beam_lib implementation details, so I think it is just better to say `term` for now. But ideally I would say all of these should be available upstream, yeah. It is good to merge IMO. :)

michallepicki:
@josevalim I'll remove my comments

josevalim:
@michallepicki I think the comments are actually good. They are inform there may be a better solution upstream, so I would keep them. :)

michallepicki:
@josevalim do we have a system for "tags" inside comments? Like an easily searchable identifier `# %inlined-erlang-type%` and a longer explaination in one of the places?

josevalim:
We have TODO, although I don't think they apply here.

josevalim:
To be honest, I think the comments were totally fine as is. :)

josevalim:

Issue #7839, opened 2018-07-07
Compiler crashes when given `S` erlc option


seancribbs:
### Environment

* Elixir & Erlang/OTP versions (elixir --version): 
```
Erlang/OTP 20 [erts-9.3.3] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:10] [hipe] [kernel-poll:false] [dtrace]

Elixir 1.6.5 (compiled with OTP 20)
```
* Operating system: macOS High Sierra 10.13.5

### Current behavior

With the Erlang compiler, you can use `+S` at the command-line to produce a dump of the BEAM opcodes. I tried setting this in the `ERL_COMPILER_OPTIONS` environment variable and the `.S` file is created, but the Elixir compiler exits with `FunctionClauseError`.

Input file (`add.ex`):
```elixir
defmodule Add do
  def add(a,b), do: id(a) + id(b)
  def id(x), do: x
end
```

Command:
```shell
$ ERL_COMPILER_OPTIONS="'S'" elixirc add.ex

== Compilation error in file add.ex ==
** (FunctionClauseError) no function clause matching in :elixir_erl.add_beam_chunks/2

    The following arguments were given to :elixir_erl.add_beam_chunks/2:

        # 1
[...]

PR #9979, opened 2020-04-26
WIP Dialyzer fixes

michallepicki:
I'll break this down in the near future and follow up with smaller PRs. This is a work in progress - some of these changes are unnecessary or invalid.

### DIFF
--- a/lib/elixir/lib/calendar/time.ex
+++ b/lib/elixir/lib/calendar/time.ex
@@ -536,6 +536,7 @@ defmodule Time do
       %Time{calendar: Calendar.Holocene, hour: 13, minute: 30, second: 15, microsecond: {0, 0}}
 
   """
+  @dialyzer {:no_match, convert!: 2}
   @doc since: "1.5.0"
   @spec convert!(Calendar.time(), Calendar.calendar()) :: t
   def convert!(time, calendar) do
--- a/lib/elixir/lib/dynamic_supervisor.ex
+++ b/lib/elixir/lib/dynamic_supervisor.ex
@@ -329,6 +329,8 @@ defmodule DynamicSupervisor do
     end
   end
 
+  # Ignore dialyzer warning for fallback clause (invalid child spec)
+  @dialyzer {:no_match, validate_child: 1}
   defp validate_child(%{id: _, start: {mod, _, _} = start} = child) do
     restart = Map.get(child, :restart, :permanent)
     type = Map.get(child, :type, :worker)
--- a/lib/elixir/lib/exception.ex
+++ b/lib/elixir/lib/exception.ex
@@ -188,9 +188,10 @@ defmodule Exception do
   Where `definition` is `:def`, `:defp`, `:defmacro` or `:defmacrop`.
   """
   @doc since: "1.5.0"
-  @spec blame_mfa(module, function, args :: [term]) ::
+  @spec blame_mfa(module, function_name :: atom, args :: [term]) ::
           {:ok, :def | :defp | :defmacro | :defmacrop, [{args :: [term], guards :: [term]}]}
           | :error
+
   def blame_mfa(module, function, args)
       when is_atom(module) and is_atom(function) and is_list(args) do
     try do
--- a/lib/elixir/lib/gen_event.ex
+++ b/lib/elixir/lib/gen_event.ex
@@ -457,6 +457,7 @@ defmodule GenEvent do
   end
 
   @doc false
+  @spec system_terminate(term, term, term, term) :: no_return
   def system_terminate(reason, parent, _debug, [name, handlers, _hib]) do
     server_terminate(reason, parent, handlers, name)
   end
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -217,6 +217,8 @@ defmodule Kernel do
   # We need this check only for bootstrap purposes.
   # Once Kernel is loaded and we recompile, it is a no-op.
   @compile {:inline, bootstrapped?: 1}
+  @dialyzer {:nowarn_function, bootstrapped?: 1}
+  @spec bootstrapped?(module) :: boolean
   case :code.ensure_loaded(Kernel) do
     {:module, _} ->
       defp bootstrapped?(_), do: true
--- a/lib/elixir/lib/string_io.ex
+++ b/lib/elixir/lib/string_io.ex
@@ -407,7 +407,6 @@ defmodule StringIO do
     end
   end
 
-  defp binary_to_list(data, _) when is_list(data), do: data
   defp binary_to_list(data, :unicode) when is_binary(data), do: String.to_charlist(data)
   defp binary_to_list(data, :latin1) when is_binary(data), do: :erlang.binary_to_list(data)
 
--- a/lib/elixir/lib/system.ex
+++ b/lib/elixir/lib/system.ex
@@ -312,9 +312,14 @@ defmodule System do
   """
   @spec user_home() :: String.t() | nil
   def user_home do
-    {:ok, [[home] | _]} = :init.get_argument(:home)
-    encoding = :file.native_name_encoding()
-    :unicode.characters_to_binary(home, encoding, encoding)
+    case :init.get_argument(:home) do
+      {:ok, [[home] | _]} ->
+        encoding = :file.native_name_encoding()
+        :unicode.characters_to_binary(home, encoding, encoding)
+
+      _ ->
+        nil
+    end
   end
 
   @doc """
--- a/lib/elixir/src/elixir_erl_compiler.erl
+++ b/lib/elixir/src/elixir_erl_compiler.erl
@@ -42,6 +42,9 @@ erl_to_core(Forms, Opts) ->
       end
   end.
 
+%% TODO: Update compile:noenv_forms/2 return type to include {ok, Module, _NotABinary, _Warnings}
+%% see 718263f and #7839 for example
+-dialyzer({no_match, compile/3}).
 compile(Forms, File, Opts) when is_list(Forms), is_list(Opts), is_binary(File) ->
   Source = elixir_utils:characters_to_list(File),
 
@@ -56,7 +59,7 @@ compile(Forms, File, Opts) when is_list(Forms), is_list(Opts), is_binary(File) -
           format_warnings(Opts, Warnings),
           {Module, Binary};
 
-        {ok, Module, _Binary, _Warnings} ->
+        {ok, Module, _NotABinary, _Warnings} ->
           elixir_errors:form_error([], File, ?MODULE, {invalid_compilation, Module});
 
         {error, Errors, Warnings} ->
--- a/lib/ex_unit/lib/ex_unit/formatter.ex
+++ b/lib/ex_unit/lib/ex_unit/formatter.ex
@@ -386,9 +386,6 @@ defmodule ExUnit.Formatter do
 
         {left, right}
 
-      nil when formatter == :expr ->
-        {if_value(left, inspect), if_value(right, inspect)}
-
       nil ->
         {if_value(left, &code_multiline(&1, padding_size)), if_value(right, inspect)}
     end
--- a/lib/iex/lib/iex.ex
+++ b/lib/iex/lib/iex.ex
@@ -638,6 +638,7 @@ defmodule IEx do
     __break__!(ast, module, fun, args, guards, stops, env)
   end
 
+  @spec __break__!(term, term) :: no_return
   def __break__!(ast, _stops) do
     raise_unknown_break_ast!(ast)
   end
--- a/lib/iex/lib/iex/cli.ex
+++ b/lib/iex/lib/iex/cli.ex
@@ -147,6 +147,7 @@ defmodule IEx.CLI do
     [dot_iex_path: find_dot_iex(:init.get_plain_arguments()), on_eof: :halt]
   end
 
+  @dialyzer {:no_return, abort: 1}
   defp abort(msg) do
     function = fn ->
       IO.puts(:stderr, msg)
--- a/lib/iex/lib/iex/helpers.ex
+++ b/lib/iex/lib/iex/helpers.ex
@@ -619,7 +619,6 @@ defmodule IEx.Helpers do
     IO.puts(IEx.color(:eval_result, ["\n## ", msg, " \n"]))
   end
 
-  defp print_entry(_key, nil), do: :ok
   defp print_entry(key, value), do: IO.puts("#{pad_key(key)}#{value}")
 
   defp print_uptime() do
--- a/lib/logger/lib/logger.ex
+++ b/lib/logger/lib/logger.ex
@@ -876,6 +876,8 @@ defmodule Logger do
 
   defguardp is_msg(msg) when is_binary(msg) or is_list(msg) or is_map(msg)
 
+  # TODO: Update :logger.macro_log/4 first argument type to include %{}
+  @dialyzer {:nowarn_function, __do_log__: 3}
   @doc false
   def __do_log__(level, fun, metadata) when is_function(fun, 0) and is_map(metadata) do
     case fun.() do
--- a/lib/mix/lib/mix/dep/loader.ex
+++ b/lib/mix/lib/mix/dep/loader.ex
@@ -256,6 +256,7 @@ defmodule Mix.Dep.Loader do
     Enum.any?(files, &File.regular?(Path.join(dest, &1)))
   end
 
+  @spec invalid_dep_format(term) :: no_return
   defp invalid_dep_format(dep) do
     Mix.raise("""
     Dependency specified in the wrong format:
--- a/lib/mix/lib/mix/dep/lock.ex
+++ b/lib/mix/lib/mix/dep/lock.ex
@@ -27,7 +27,7 @@ defmodule Mix.Dep.Lock do
   @doc """
   Receives a map and writes it as the latest lock.
   """
-  @spec write(map) :: :ok
+  @spec write(map | keyword) :: :ok
   def write(map) do
     unless map == read() do
       lines =
--- a/lib/mix/lib/mix/project.ex
+++ b/lib/mix/lib/mix/project.ex
@@ -388,7 +388,7 @@ defmodule Mix.Project do
   depends on B, then B will listed before A.
   """
   @doc since: "1.11.0"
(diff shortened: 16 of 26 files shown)


lib/mix/lib/mix/dep/lock.ex
```diff
@@ -27,7 +27,7 @@ defmodule Mix.Dep.Lock do
   @doc """
   Receives a map and writes it as the latest lock.
   """
-  @spec write(map) :: :ok
+  @spec write(map | keyword) :: :ok
```
michallepicki, line 30: This could be fixed at the call site by converting to a map before calling `Lock.write`

lib/elixir/lib/kernel.ex
```diff
@@ -217,6 +217,8 @@ defmodule Kernel do
   # We need this check only for bootstrap purposes.
   # Once Kernel is loaded and we recompile, it is a no-op.
   @compile {:inline, bootstrapped?: 1}
+  @dialyzer {:nowarn_function, bootstrapped?: 1}
+  @spec bootstrapped?(module) :: boolean
```
michallepicki, line 221: The fact that `bootstrapped?` is getting inlined is causing multiple `match` warnings in this file. These changes didn't help. I tried ignoring the warning at each call site but I couldn't figure out how to do that for `defmacro @{a, b, c}`. I'll keep investigating.

lib/mix/lib/mix/tasks/test.ex
```diff
@@ -556,7 +557,7 @@ defmodule Mix.Tasks.Test do
 
   defp filter_to_allowed_files(matched_test_files, nil), do: matched_test_files
 
-  defp filter_to_allowed_files(matched_test_files, %MapSet{} = allowed_files) do
```
michallepicki, line 559: `MapSet.t()` is a opaque type which means that we cannot pattern match on this value

lib/logger/lib/logger.ex
```diff
@@ -876,6 +876,8 @@ defmodule Logger do
 
   defguardp is_msg(msg) when is_binary(msg) or is_list(msg) or is_map(msg)
 
+  # TODO: Update :logger.macro_log/4 first argument type to include %{}
+  @dialyzer {:nowarn_function, __do_log__: 3}
```
michallepicki, line 880: There are `logger:macro_log/4` calls with an empty map as first argument in the `logger` module. How are these not causing dialyzer issues for erlang itself?

lib/mix/lib/mix/tasks/test.ex
josevalim: I would honestly keep doing this check. This is a limitation of Dialyzer-Elixir integration and I wouldn't want to make the code less precise only to satisfy Dialyzer.

lib/logger/lib/logger.ex
josevalim: Try adding `generated: true` to the quote that emits `macro_log` and see if it addresses them.

lib/elixir/lib/dynamic_supervisor.ex
```diff
@@ -329,6 +329,8 @@ defmodule DynamicSupervisor do
     end
   end
 
+  # Ignore dialyzer warning for fallback clause (invalid child spec)
+  @dialyzer {:no_match, validate_child: 1}
```
josevalim, line 333: We probably won't accept those annotations. Dialyzer is not a priority for us and I wouldn't want to add annotations just to make Dialyzer happy. :)

michallepicki: I understand, just wanted to ignore some of these to have better visibility at actual issues:
- most importantly, dialyzer errors that may bubble up to elixir end users
- dialyzer issues in erlang itself
- dead code that can be cleaned up

josevalim: Perfect, thank you for all these contributions!

lib/logger/lib/logger.ex
michallepicki: Sorry, I meant the erlang `:logger` module. I'll check how this gets through erlang's own dialyzer check on their CI

lib/elixir/lib/system.ex
```diff
@@ -312,9 +312,14 @@ defmodule System do
...
-    encoding = :file.native_name_encoding()
-    :unicode.characters_to_binary(home, encoding, encoding)
+    case :init.get_argument(:home) do
+      {:ok, [[home] | _]} ->
+        encoding = :file.native_name_encoding()
+        :unicode.characters_to_binary(home, encoding, encoding)
+
+      _ ->
+        nil
+    end
```
michallepicki, line 322: Dialyzer suggests that `user_home` can never return `nil` and I agree. I am not sure if this is the way to fix this and I don't know how to test this :S

lib/elixir/src/elixir_erl_compiler.erl
```diff
@@ -56,7 +59,7 @@ compile(Forms, File, Opts) when is_list(Forms), is_list(Opts), is_binary(File) -
           format_warnings(Opts, Warnings),
           {Module, Binary};
 
-        {ok, Module, _Binary, _Warnings} ->
+        {ok, Module, _NotABinary, _Warnings} ->
```
michallepicki, line 62: Other than this return type not being documented in erlang specs for `compile:noenv_forms/2`, this clause is also not being tested in the elixir test suite (elixir tests are passing after removing this line and the next line). If such test would be useful, it can be added based on an example from https://github.com/elixir-lang/elixir/issues/7839

lib/ex_unit/lib/ex_unit/formatter.ex
```diff
@@ -386,9 +386,6 @@ defmodule ExUnit.Formatter do
 
         {left, right}
 
-      nil when formatter == :expr ->
-        {if_value(left, inspect), if_value(right, inspect)}
```
michallepicki, line 390: This seems to be dead code? Dialyzer suggests that `formatter` is always a function and I agree.

lib/mix/lib/mix/project_stack.ex
```diff
@@ -105,7 +105,6 @@ defmodule Mix.ProjectStack do
     end)
   end
 
-  @spec compile_env([term]) :: [term]
```
michallepicki, line 108: I am not sure what the correct spec would be here, but this function is getting called with `:unset` in `Mix.Tasks.Compile.App.load_compile_env/1`

lib/mix/lib/mix/tasks/archive.install.ex
```diff
@@ -113,7 +113,7 @@ defmodule Mix.Tasks.Archive.Install do
     remove_previous_versions(previous)
 
     File.mkdir_p!(dir_dest)
-    {:ok, _} = :zip.extract(contents, cwd: dir_dest)
+    {:ok, _} = :zip.extract(contents, cwd: to_charlist(dir_dest))
```
michallepicki, line 116: Erlang expects this to be a charlist but it seems to be working for binaries right now? Not sure if it's worth looking into.

lib/mix/lib/mix/tasks/escript.build.ex
```diff
@@ -136,14 +136,6 @@ defmodule Mix.Tasks.Escript.Build do
     filename = escript_opts[:path] || script_name
     main = escript_opts[:main_module]
 
-    unless script_name do
-      error_message =
-        "Could not generate escript, no name given, " <>
-          "set :name escript option or :app in the project settings"
-
-      Mix.raise(error_message)
-    end
```
michallepicki, line 145: Dialyzer suggests this is dead code and I agree, `script_name` can not be nil and there is no test for this error. But it could be an empty string (maybe?), so possibly this should be changed to check for that?

lib/elixir/lib/calendar/time.ex
```diff
@@ -536,6 +536,7 @@ defmodule Time do
       %Time{calendar: Calendar.Holocene, hour: 13, minute: 30, second: 15, microsecond: {0, 0}}
 
   """
+  @dialyzer {:no_match, convert!: 2}
```
michallepicki, line 539: not an issue, can be ignored

lib/elixir/lib/exception.ex
```diff
@@ -188,9 +188,10 @@ defmodule Exception do
   Where `definition` is `:def`, `:defp`, `:defmacro` or `:defmacrop`.
   """
   @doc since: "1.5.0"
-  @spec blame_mfa(module, function, args :: [term]) ::
+  @spec blame_mfa(module, function_name :: atom, args :: [term]) ::
```
michallepicki, line 191: fixed

lib/elixir/lib/gen_event.ex
```diff
@@ -457,6 +457,7 @@ defmodule GenEvent do
   end
 
   @doc false
+  @spec system_terminate(term, term, term, term) :: no_return
```
michallepicki, line 460: not an issue

lib/elixir/lib/string_io.ex
```diff
@@ -407,7 +407,6 @@ defmodule StringIO do
     end
   end
 
-  defp binary_to_list(data, _) when is_list(data), do: data
```
michallepicki, line 410: not an issue

lib/iex/lib/iex.ex
```diff
@@ -638,6 +638,7 @@ defmodule IEx do
     __break__!(ast, module, fun, args, guards, stops, env)
   end
 
+  @spec __break__!(term, term) :: no_return
```
michallepicki, line 641: not an issue

lib/iex/lib/iex/cli.ex
```diff
@@ -147,6 +147,7 @@ defmodule IEx.CLI do
     [dot_iex_path: find_dot_iex(:init.get_plain_arguments()), on_eof: :halt]
   end
 
+  @dialyzer {:no_return, abort: 1}
```
michallepicki, line 150: not an issue

lib/iex/lib/iex/helpers.ex
```diff
@@ -619,7 +619,6 @@ defmodule IEx.Helpers do
     IO.puts(IEx.color(:eval_result, ["\n## ", msg, " \n"]))
   end
 
-  defp print_entry(_key, nil), do: :ok
```
michallepicki, line 622: not an issue

lib/mix/lib/mix/dep/loader.ex
```diff
@@ -256,6 +256,7 @@ defmodule Mix.Dep.Loader do
     Enum.any?(files, &File.regular?(Path.join(dest, &1)))
   end
 
+  @spec invalid_dep_format(term) :: no_return
```
michallepicki, line 259: not an issue

lib/mix/lib/mix/dep/lock.ex
michallepicki: fixed

lib/mix/lib/mix/project.ex
```diff
@@ -388,7 +388,7 @@ defmodule Mix.Project do
   depends on B, then B will listed before A.
   """
   @doc since: "1.11.0"
-  @spec deps_apps() :: atom()
+  @spec deps_apps() :: [atom()]
```
michallepicki, line 391: fixed

lib/mix/lib/mix/release.ex
```diff
@@ -50,7 +50,7 @@ defmodule Mix.Release do
 
   @type mode :: :permanent | :transient | :temporary | :load | :none
   @type application :: atom()
-  @type t :: %{
+  @type t :: %__MODULE__{
```
michallepicki, line 53: fixed

lib/mix/lib/mix/release.ex
```diff
@@ -61,6 +61,7 @@ defmodule Mix.Release do
           erts_source: charlist() | nil,
           config_providers: [{module, term}],
           options: keyword(),
+          overlays: list(),
```
michallepicki, line 64: fixed

lib/mix/lib/mix/release.ex
```diff
@@ -222,6 +223,7 @@ defmodule Mix.Release do
     end
   end
 
+  @spec bad_umbrella! :: no_return
```
michallepicki, line 226: not an issue

lib/mix/lib/mix/tasks/iex.ex
```diff
@@ -6,6 +6,7 @@ defmodule Mix.Tasks.Iex do
   """
 
   @impl true
+  @spec run(term) :: no_return
```
michallepicki, line 9: not an issue

lib/mix/lib/mix/tasks/test.ex
```diff
@@ -401,6 +401,7 @@ defmodule Mix.Tasks.Test do
     end
   end
 
+  @spec raise_with_shell(term, term) :: no_return
```
michallepicki, line 404: not an issue

lib/mix/lib/mix/utils.ex
```diff
@@ -663,8 +663,6 @@ defmodule Mix.Utils do
 
   defp proxy_auth(%URI{scheme: "http"}, http_proxy, _https_proxy), do: proxy_auth(http_proxy)
   defp proxy_auth(%URI{scheme: "https"}, _http_proxy, https_proxy), do: proxy_auth(https_proxy)
-
-  defp proxy_auth(nil), do: []
```
michallepicki, line 667: not an issue

lib/mix/lib/mix/tasks/archive.install.ex
josevalim: I will confirm to the specs just in case, so this change is good IMO.

lib/mix/lib/mix/project_stack.ex
josevalim: Please open up an issue. :)

lib/ex_unit/lib/ex_unit/formatter.ex
josevalim: Agreed.

lib/elixir/lib/system.ex
josevalim: Let's fix the typespecs for now to remove `| nil`.

lib/elixir/src/elixir_erl_compiler.erl
josevalim: It is triggered if you do things like `ERL_COMPILER_OPTIONS=to_asm` and similar. So we should leave this one as is. :)

lib/mix/lib/mix/tasks/escript.build.ex
josevalim: We can remove this. :) Good change. :+1:

lib/elixir/lib/system.ex
michallepicki: Doesn't seem right to me, there's a bang version underneath that relies on this returning a nil.

lib/elixir/src/elixir_erl_compiler.erl
michallepicki: So this can stay untested? Alright.

lib/elixir/lib/system.ex
michallepicki: This looks like the right way to fix this, per http://erlang.org/doc/man/init.html `get_name` returns `:error` when no value is associated with the flag.

lib/elixir/src/elixir_erl_compiler.erl
josevalim: Right, we just raise after all for clear better errors for people that may be doing the wrong thing. :)

lib/logger/lib/logger.ex
michallepicki: This is Elixir calling Erlang logger through Erlang's undocumented interface that Erlang uses after expanding their macros. I was wrong before, Erlang is never calling these with an empty map as location so they don't get this dialyzer error.

I am not sure why we're not using the regular Erlang function-based logging interface or why we're not passing location information but there are probably important reasons for that.

michallepicki:
After merging #9993 `make dialyze` will be passing with `DIALYZER_OPTS` set to:
```
 --no_check_plt --fullpath -Wno_unused  -Wno_return -Wno_match -Wno_fail_call
```

producing this output:
```
==> Dialyzing Elixir...
  Proceeding with analysis...
Unknown functions:
  'Elixir.Collectable.Atom':'__impl__'/1
  'Elixir.Collectable.Float':'__impl__'/1
  'Elixir.Collectable.Function':'__impl__'/1
  'Elixir.Collectable.Integer':'__impl__'/1
  'Elixir.Collectable.PID':'__impl__'/1
  'Elixir.Collectable.Port':'__impl__'/1
  'Elixir.Collectable.Reference':'__impl__'/1
  'Elixir.Collectable.Tuple':'__impl__'/1
  'Elixir.Enumerable.Atom':'__impl__'/1
  'Elixir.Enumerable.BitString':'__impl__'/1
  'Elixir.Enumerable.Float':'__impl__'/1
  'Elixir.Enumerable.Integer':'__impl__'/1
  'Elixir.Enumerable.PID':'__impl__'/1
  'Elixir.Enumerable.Port':'__impl__'/1
  'Elixir.Enumerable.Reference':'__impl__'/1
  'Elixir.Enumerable.Tuple':'__impl__'/1
  'Elixir.Hex':start/0
  'Elixir.Hex':version/0
  'Elixir.List.Chars.Function':'__impl__'/1
  'Elixir.List.Chars.Map':'__impl__'/1
  'Elixir.List.Chars.PID':'__impl__'/1
  'Elixir.List.Chars.Port':'__impl__'/1
  'Elixir.List.Chars.Reference':'__impl__'/1
  'Elixir.List.Chars.Tuple':'__impl__'/1
  'Elixir.String.Chars.Function':'__impl__'/1
  'Elixir.String.Chars.Map':'__impl__'/1
  'Elixir.String.Chars.PID':'__impl__'/1
  'Elixir.String.Chars.Port':'__impl__'/1
  'Elixir.String.Chars.Reference':'__impl__'/1
  'Elixir.String.Chars.Tuple':'__impl__'/1
 done in 2m9.42s
done (passed successfully)
```

Additionally, merging #9994 and fixing (or silencing) the `:logger.macro_log` usage with empty location would allow dialyzer to pass without disabling failed call checks (`-Wno_fail_call` silences those), which are IMO one of the most useful ones.

Closing, I'm done here :)

PR #9982, opened 2020-04-27
Call Mix.Dep.Lock.write/1 with a map in Mix.Tasks.Deps.Unlock.run/1

michallepicki:
This could be `:maps.filter(fn ... not in ..., lock)` instead but I don't know if the micro optimization is worth it


josevalim:

PR #9988, opened 2020-04-28
Remove redundant clause in ExUnit.Formatter.format_sides/6

michallepicki:
The formatter argument is a function so never should be equal to :expr

### DIFF
--- a/lib/ex_unit/lib/ex_unit/formatter.ex
+++ b/lib/ex_unit/lib/ex_unit/formatter.ex
@@ -386,9 +386,6 @@ defmodule ExUnit.Formatter do
 
         {left, right}
 
-      nil when formatter == :expr ->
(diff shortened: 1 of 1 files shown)


fertapric:
I think we could add some guards in [`ExUnit.Formatter.format_test_failure/5`](https://github.com/elixir-lang/elixir/blob/master/lib/ex_unit/lib/ex_unit/formatter.ex#L124) and [`ExUnit.Formatter.format_test_all_failure/5`](https://github.com/elixir-lang/elixir/blob/master/lib/ex_unit/lib/ex_unit/formatter.ex#L172) to ensure a function is always passed in the `formatter` argument.

michallepicki:
@fertapric  I am not going to do that but this may be the right thing to do indeed, if anyone wants to take care of it :)

josevalim:

PR #9989, opened 2020-04-28
Call :zip.extract/2 with cwd option as charlist

michallepicki:
This function call is working when cwd is a binary right now, but it is documented that cwd should be a charlist


wojtekmach:
On the flip side, OTP will likely not break users that have been passing binaries as cwd. Consider checking with them first, I'm pretty sure they'll accept a patch that updates the spec to `filename:filename_all` that matches the behaviour.

michallepicki:
@wojtekmach I am not going to do that but this may be the right thing to do indeed, if anyone wants to take care of it :)

josevalim:

PR #9991, opened 2020-04-28
Fix handling of empty user home flag in System.user_home/0

michallepicki:
According to erlang documentation, :init.get_argument/1 returns :error when no value is associated with the flag: http://erlang.org/doc/man/init.html

### DIFF
--- a/lib/elixir/lib/system.ex
+++ b/lib/elixir/lib/system.ex
@@ -312,9 +312,14 @@ defmodule System do
   """
   @spec user_home() :: String.t() | nil
   def user_home do
-    {:ok, [[home] | _]} = :init.get_argument(:home)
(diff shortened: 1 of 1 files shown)


michalmuskala:
I think that actually can't happen because the VM won't boot without `-home` flag (it's added automatically by `erl` startup script before launching `beam.emu`).

michallepicki:
Should I update the PR to change the return type to remove `| nil` and deprecate the bang version?

michallepicki:
Or if this could be possible with future Erlang versions, or other potential future Elixir runtime target, this can be just closed without merging.

josevalim:
Given https://github.com/erlang/otp/pull/2390, I would say this is the way to go!

josevalim:

PR #9992, opened 2020-04-28
Fix spec for Mix.Local.Installer.parse_args/2

michallepicki:
The function can also return an {:error, message} tuple when parsing is not successful


fertapric:
Good catch @michallepicki! 💜

michallepicki:
good catch @dialyzer ;) <3

PR #9994, opened 2020-04-28
Fix spec for Mix.ProjectStack.compile_env/1

michallepicki:
This function is getting called with :unset in Mix.Tasks.Compile.App.load_compile_env/1


josevalim:

PR #10002, opened 2020-04-29
Recompile Kernel before bootstrapping stdlib

### LINKED ISSUE #10000: Elixir.Macro.Env.beam file differs between builds
### Environment

* Elixir & Erlang/OTP versions (elixir --version):  elixir-1.10.3 ; erlang-22.3
* Operating system: openSUSE-Tumbleweed 20200427

### Current behavior

Builds of elixir packages differ between 1-core-VM and 4-core-VM in Elixir.Macro.Env.beam
The start of the main diff has `elixir-1.10.3/lib/elixir/lib/macro/env.exj Dbgi`

Two builds on 4-core-VMs produce proper identical results.

### Expected behavior

It should be possible to produce identical build results anywhere, anytime.
See https://reproducible-builds.org/ for why this matters.

This is sort of a regression on what was all fixed back in #8689
My last reproducible build happened on 2020-04-01 with elixir-1.9.4

### How to reproduce:

```bash
osc co openSUSE:Factory/elixir && cd $_
for N in 1 4 ; do
  osc build --no-service --vm-type=kvm \
    --keep-pkg=RPMS.$N -j$N
  unrpm RPMS.$N/elixir-1.10.3-0.noarch.rpm
  hexdump -C usr/lib/elixir/lib/elixir/ebin/Elixir.Macro.Env.beam > $N.hex
done
diff -u {1,4}.hex
```

josevalim:
Before this patch, we could have slightly different
AST while compiling modules in parallel, depending
if modules were recompiled or not. Now we always
recompile Kernel before compiling the whole stdlib.

We used this opportunity to simplify bootstrap by
breaking it two steps. The first step is the minimal
necessary to compile Kernel itself. The second step
contains the whole compiler.

Closes #10000.

### DIFF
--- a/lib/elixir/lib/code.ex
+++ b/lib/elixir/lib/code.ex
@@ -1308,7 +1308,7 @@ defmodule Code do
   @doc since: "1.11.0"
   @spec can_await_module_compilation? :: boolean
   def can_await_module_compilation? do
-    Process.info(self(), :error_handler) == {:error_handler, Kernel.ErrorHandler}
+    :erlang.process_info(self(), :error_handler) == {:error_handler, Kernel.ErrorHandler}
   end
 
   @doc false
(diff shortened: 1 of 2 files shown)


josevalim:

PR #10003, opened 2020-04-30
Titlecase the given string

### LINKED ISSUE #9999: Add String.titlecase/2
We already have the implementation internally. Similar to capitalize but it doesn't downcase the rest of the string.

sushant12:
Closes #9999


sushant12:
I am unaware of other cases that might fail String.titlecase/2 . If there is then please do let me know.

PR #10004, opened 2020-04-30
titlecase support for string module

### LINKED ISSUE #9999: Add String.titlecase/2

sushant12:
closes #9999

I am unaware of other cases that might fail String.titlecase/2 . If there is then please do let me know.

### DIFF
--- a/lib/elixir/lib/string.ex
+++ b/lib/elixir/lib/string.ex
@@ -835,6 +835,29 @@ defmodule String do
   defp downcase_ascii(<<>>), do: []
 
   @doc """
+  Capitalizes the first character of all the words in a string according to `mode`.
+
+  `mode` may be `:default`, `:ascii` or `:greek`. The `:default` mode considers
+  all non-conditional transformations outlined in the Unicode standard. `:ascii`
+  lowercases only the letters A to Z. `:greek` includes the context sensitive
+  mappings found in Greek.
+
+  ## Examples
+      iex> String.titlecase("hello world")
+      "Hello World"
+  """
+  @spec titlecase(t, :default | :ascii | :greek) :: t
+  def titlecase(string, mode \\ :default)
+
+  def titlecase("", _mode), do: ""
+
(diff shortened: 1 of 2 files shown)


josevalim:
Hi @sushant12. Apologies, I was not very precise on the description. This should not behave per word, because the concept of words is actually tricky to define. The implementation of this function should be like `capitalize`, except that we don't call downcase after. Something like this:

```elixir
  @spec titlecase(t, :default | :ascii | :greek) :: t
  def titlecase(string, mode \\ :default)

  def titlecase(<<char, rest::binary>>, :ascii)
      when char >= ?a and char <= ?z,
    do: <<char - 32, rest::binary>>

  def titlecase(string, :ascii) when is_binary(string), do: string

  def titlecase(string, mode) when is_binary(string) do
    {char, rest} = String.Casing.titlecase_once(string, mode)
    char <> rest
  end
```

Can you please update the implementation and tests accordingly? You can also use the tests from capitalize, making sure nothing is downcased.

sushant12:
@josevalim please correct me if I am wrong.

I need to implement `String.titlecase/2` function whose implementation should be similar to `capitalize`

> except that we don't call downcase after

When I run `String.capitalize("hello world")` it returns `"Hello world"` .
Should `String.titlecase("hello world")` return `"HELLO WORLD"` ??If yes, then isnt that same as `upcase` ?

josevalim:
```
String.capitalize("hello world") => "Hello world"
String.capitalize("HELLO WORLD") => "Hello world"
String.titlecase("HELLO WORLD") => "HELLO WORLD"
String.titlecase("hello world => "Hello world"
```

Try the code that I posted and you will see the diff. :)

josevalim:
Hi @sushant12! I have merged #10008 which has an implementation closed to the one I had in mind. Thanks again and sorry for the lack of clarify.

PR #10006, opened 2020-04-30
Add a note about `@external_resource` not being there yet

wojtekmach:


lib/elixir/lib/module.ex
```diff
@@ -252,6 +252,9 @@ defmodule Module do
   in case any of the external resources change, see for example:
   [`mix compile.elixir`](https://hexdocs.pm/mix/Mix.Tasks.Compile.Elixir.html).
 
+  Even if the given external file does not exist, the module still
+  adds a dependency on it.
```
wojtekmach, line 256: The opposite is not currently true, if we remove the file, the module is not recompiled. Is it worth changing?

josevalim: Maybe we should fix it.

josevalim, line 256:
```suggestion
  If the external resource does not exist, the module still adds a
  dependency on it, causing it to be recompiled as soon as the
  file is added.
```

josevalim:

PR #10007, opened 2020-04-30
Load app from dependencies path when it is a project dependency

ggpasqualino:
I made a spike in my job to use the built-in mix releases instead of distillery.
The release was built fine but it broke when I tried to boot it because a different version of cowboy was packed into it.
The fix was rather easy, I just removed the system package, which was mistakenly installed, from our base Docker image.
Finding the problem was not so straightforward for me though. So, I am suggesting in this PR some way to prevent this to happen and have a message early on about the problem. What do you think?

PS. I am not sure how to test such thing.

### DIFF
--- a/lib/mix/lib/mix/release.ex
+++ b/lib/mix/lib/mix/release.ex
@@ -91,15 +91,15 @@ defmodule Mix.Release do
     {include_erts, opts} = Keyword.pop(opts, :include_erts, true)
     {erts_source, erts_lib_dir, erts_version} = erts_data(include_erts)
 
-    apps_lock = Mix.Dep.Lock.read()
-    loaded_apps = apps |> Keyword.keys() |> load_apps(apps_lock, %{}, erts_lib_dir, :maybe)
+    deps_apps = Mix.Project.deps_apps()
+    loaded_apps = apps |> Keyword.keys() |> load_apps(deps_apps, %{}, erts_lib_dir, :maybe)
 
     # Make sure IEx is either an active part of the release or add it as none.
     {loaded_apps, apps} =
       if Map.has_key?(loaded_apps, :iex) do
         {loaded_apps, apps}
       else
-        {load_apps([:iex], apps_lock, loaded_apps, erts_lib_dir, :maybe), apps ++ [iex: :none]}
+        {load_apps([:iex], deps_apps, loaded_apps, erts_lib_dir, :maybe), apps ++ [iex: :none]}
       end
 
     start_boot = build_start_boot(loaded_apps, apps)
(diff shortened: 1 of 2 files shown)


josevalim:
I like this! Can you please look at the CI and push a test as well? Thank you!

ggpasqualino:
I fixed the case that broke the tests and added one more test.
To be fair I can't say I understand the tests very well 😬but I can take a look during the next days and try to improve the PR.

josevalim:
Hi @ggpasqualino! I took a second look at this and I have been wondering: having an error message about a potential conflict that cannot be trivially addressed is going to be a limitation. Maybe we should not raise but instead give the non-OTP application higher preference?

michalmuskala:
I also think this should be a bit more generic and detect if the same app appears multiple times anywhere in code path in general and not just inside OTP and outside of it. You could have installed some applications somewhere else than OTP install directory and just export `ERL_LIBS` and face the same issue.

josevalim:
@michalmuskala so I decided to check why we don't simply trust the codepath and I found #9054. I believe this patch will also break #9054. :/

ggpasqualino:
Thank you for your input, considering those I think I can reformulate this issue in a way that makes more sense.
So, my real problem is not that the app was loaded from a different path, but that a different version of the app was loaded, which happened to be in the OTP path and not the project path.
Do you think it would make sense to leave the precedence of the app path as it is now, but take the locked version of the app (if there is any) in consideration when we look it up?

josevalim:
@ggpasqualino sounds like a better option indeed. :)

lib/mix/lib/mix/release.ex
```diff
@@ -297,30 +298,44 @@ defmodule Mix.Release do
     end
   end
 
-  defp load_app(app, seen, otp_root, included) do
-    path = Path.join(otp_root, "#{app}-*")
+  defp load_app(app, apps_lock, seen, otp_root, included) do
+    case apps_lock[app] do
+      {:hex, _, version, _, _, _, _, _} ->
```
ggpasqualino, line 303: I could not find all possibilities for the lock entry

lib/mix/lib/mix/release.ex
```diff
@@ -297,30 +298,44 @@ defmodule Mix.Release do
...
+        do_load_app(app, path, apps_lock, seen, otp_root, otp_app?, included)
+    end
+  end
+
+  defp app_path(otp_root, app, version \\ "*") do
+    path = Path.join(otp_root, "#{app}-#{version}")
 
     case Path.wildcard(path) do
       [] ->
         case :code.lib_dir(app) do
```
ggpasqualino, line 318: Can this return different versions? I didn't manage to simulate it.

ggpasqualino:
I gave a second try with this idea :)

lib/mix/lib/mix/release.ex
josevalim: Yeah, unfortunately the lock format is opaque, so you cannot match on it. :/ However, I don't think you need to know the lock entry itself, you simply need to know if it is a Mix dependency or not!

So my suggestion is to simply call `Mix.Project.deps_apps`. If the app is in `deps_apps`, then you don't look at it in the otp_root, only in lib_dir!

ggpasqualino: Ok, I gave a shot to this then. I had understood first that this would conflict with https://github.com/elixir-lang/elixir/pull/9054 🤔 
The setup for this function to work in the test is quite convoluted, maybe there is some more straightforward manner

josevalim:

PR #10008, opened 2020-04-30
Add String.titlecase/2

ukrbublik:
`String.titlecase/2` is similar to `String.capitalize/2` but it doesn't downcase the rest of the string.

Closes https://github.com/elixir-lang/elixir/issues/9999

### DIFF
--- a/lib/elixir/lib/string.ex
+++ b/lib/elixir/lib/string.ex
@@ -859,13 +859,51 @@ defmodule String do
   def capitalize(string, mode \\ :default)
 
   def capitalize(<<char, rest::binary>>, :ascii) do
-    char = if char >= ?a and char <= ?z, do: char - 32, else: char
-    <<char>> <> downcase(rest, :ascii)
+    <<capitalize_ascii_char(char)>> <> downcase(rest, :ascii)
   end
 
   def capitalize(string, mode) when is_binary(string) do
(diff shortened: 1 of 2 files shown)


josevalim:

ukrbublik:
@josevalim 
> Unicode says "HH" is not titlecased, which means we need
to downcase the rest of the string, which is the behaviour
we already have in capitalize.

According to [this](https://unicode.org/faq/casemap_charprop.html), titlecase differs from uppercase.
There are special chars, eg.
`ǆ` (lowercased)  -- `ǅ` (titlecased) -- `Ǆ` (uppercased)
So we need to have titlecase function.
Maybe it should work with graphene instead of binary?

ukrbublik:
Or  it should work with binaries but in more smart way - for each word in string.
However there are caveats, like:
"In Search of Sunrise" - we shouldn't titlecase "of" to "Of"
"McDonalds" - not "Mcdonalds"
"O'Brien" - not "O'brien"

harrisi:
Sorry for bringing this old PR up, but, did this never make it into a release?

NickNeck:
@harrisi the merged PR was reverted, see above.

kipcole9:
I have a lib that implements the Unicode case folding algorithm should you need it (it also does locale specific case folding and other stuff):  https://hexdocs.pm/unicode_string/readme.html#case-mapping

Issue #9831, opened 2020-02-20
Add linear representation functions to Calendar types


josevalim:
Our Calendar types have historically missed a feature that allows them to be converted to an integer. The reason for this is that, although Unix Epoch is the standard, the Unix Epoch is normalized to UTC, so converting a naive datetime to Unix Epoch could be misleading.

However, there are many situations where converting to an integer can be useful. For example, when sorting large lists of datetimes, working on integers can be much more efficient. Given we can't standardize on Unix Epoch, my proposal is to standardize on gregorian seconds with the following functions:

```elixir
Date.to_gregorian_days(date)
Date.from_gregorian_days(days, calendar \\ Calendar.ISO)
NaiveDateTime.to_gregorian_seconds(naive_datetime)
NaiveDateTime.from_gregorian_seconds(seconds, microseconds \\ {0, 0}, calendar \\ Calendar.ISO)
```

They will mirror the `from_erl`/`to_erl` APIs. I also would like to introduce similar functions for Time, but those are not gregorian related. We could call `Time.to_seconds_after_midnight` or `Time.from_seconds_after_midnight` but I am not convinced about those, so I am glad to punt on it.

/cc @kipcole9 @Qqwy @lau

lau:
Sounds good to me.

Qqwy:
:thinking: We currently already have `Calendar.naive_datetime_to_iso_days` and `Calendar.time_to_day_fraction` which when called together on e.g. a (Naive)DateTime would form a tuple akin `{days :: int, {parts_in_day :: int, parts_per_day :: int}}` that would be quick to sort or to use for other purposes, although some care is required since we are working with a fraction.

That might be enough. Conversions to Gregorian seconds can only be done for calendars that are compatible with these. Many are, but not all of them. 

Of course, we could also take the tuple mentioned above and calculate `days * parts_per_day + parts_in_day` (this does make the assumption that all calendar entries have the same `parts_in_day` value) or days + `parts_per_day/parts_in_day` (resulting in loss of precision by turning it into a float, but no further assumptions are made).


I think we should write a benchmark to compare the performance characteristics of:
- ISOday/fraction tuples: large lists of `{int, {int, int}}` where the latter two are considered as fraction.
- large lists of 'ISOday/fraction' represented as floats.
- 'Gregorian microseconds': large lists of (big) integers.

The results of this might tell us if it is worth it to create a specialized linear representation that is limited to only certain kinds of calendars.

[...]

PR #10009, opened 2020-04-30
Add linear representation functions to Calendar types

ukrbublik:
Closes https://github.com/elixir-lang/elixir/issues/9831

Added functions:
- Date.from_gregorian_days/2
- Date.to_gregorian_days/1
- NaiveDateTime.from_gregorian_seconds/3
- NaiveDateTime.to_gregorian_seconds/1
- Time.from_seconds_after_midnight/3
- Time.to_seconds_after_midnight/1

### DIFF
--- a/lib/elixir/lib/calendar/date.ex
+++ b/lib/elixir/lib/calendar/date.ex
@@ -399,6 +399,44 @@ defmodule Date do
   end
 
   @doc """
+  Converts a number of gregorian days to a `Date` struct.
+
+  ## Examples
+
+      iex> Date.from_gregorian_days(1)
+      ~D[0000-01-02]
+      iex> Date.from_gregorian_days(730_485)
+      ~D[2000-01-01]
+      iex> Date.from_gregorian_days(-1)
+      ~D[-0001-12-31]
+
+  """
+  @spec from_gregorian_days(integer(), Calendar.calendar()) :: t
+  def from_gregorian_days(days, calendar \\ Calendar.ISO) when is_integer(days) do
+    from_iso_days({days, 0}, calendar)
+  end
+
+  @doc """
+  Converts a `Date` struct to a number of gregorian days.
+
+  ## Examples
+
+      iex> Date.to_gregorian_days(~D[0000-01-02])
+      1
+      iex> Date.to_gregorian_days(~D[2000-01-01])
+      730_485
+      iex> Date.to_gregorian_days(~N[2000-01-01 00:00:00])
+      730_485
+
+  """
+  @spec to_gregorian_days(t) :: integer()
+
+  def to_gregorian_days(date) do
+    {days, _} = to_iso_days(date)
+    days
+  end
+
+  @doc """
   Compares two date structs.
 
   Returns `:gt` if first date is later than the second
--- a/lib/elixir/lib/calendar/date.ex
+++ b/lib/elixir/lib/calendar/date.ex
@@ -417,7 +417,7 @@ defmodule Date do
   end
 
   @doc """
-  Converts a `Date` struct to a number of gregorian days.
+  Converts a `date` struct to a number of gregorian days.
 
   ## Examples
 
@@ -429,7 +429,7 @@ defmodule Date do
       730_485
 
   """
-  @spec to_gregorian_days(t) :: integer()
+  @spec to_gregorian_days(Calendar.date()) :: integer()
   def to_gregorian_days(date) do
     {days, _} = to_iso_days(date)
     days
--- a/lib/elixir/lib/calendar/date.ex
+++ b/lib/elixir/lib/calendar/date.ex
@@ -411,6 +411,7 @@ defmodule Date do
       ~D[-0001-12-31]
 
   """
+  @doc since: "1.11.0"
   @spec from_gregorian_days(integer(), Calendar.calendar()) :: t
   def from_gregorian_days(days, calendar \\ Calendar.ISO) when is_integer(days) do
     from_iso_days({days, 0}, calendar)
@@ -429,6 +430,7 @@ defmodule Date do
       730_485
 
   """
+  @doc since: "1.11.0"
   @spec to_gregorian_days(Calendar.date()) :: integer()
   def to_gregorian_days(date) do
     {days, _} = to_iso_days(date)
--- a/lib/elixir/lib/calendar/date.ex
+++ b/lib/elixir/lib/calendar/date.ex
@@ -430,7 +430,6 @@ defmodule Date do
 
   """
   @spec to_gregorian_days(t) :: integer()
-
   def to_gregorian_days(date) do
     {days, _} = to_iso_days(date)
     days
--- a/lib/elixir/lib/calendar/naive_datetime.ex
+++ b/lib/elixir/lib/calendar/naive_datetime.ex
@@ -786,6 +786,7 @@ defmodule NaiveDateTime do
       ~N[-0001-12-31 23:59:59]
 
   """
+  @doc since: "1.11.0"
   @spec from_gregorian_seconds(integer(), Calendar.microsecond(), Calendar.calendar()) :: t
   def from_gregorian_seconds(
         seconds,
@@ -823,6 +824,7 @@ defmodule NaiveDateTime do
       63_755_511_991
 
   """
+  @doc since: "1.11.0"
   @spec to_gregorian_seconds(Calendar.naive_datetime()) :: integer()
   def to_gregorian_seconds(%{
         calendar: calendar,
--- a/lib/elixir/lib/calendar/naive_datetime.ex
+++ b/lib/elixir/lib/calendar/naive_datetime.ex
@@ -824,7 +824,7 @@ defmodule NaiveDateTime do
 
   """
   @spec to_gregorian_seconds(Calendar.naive_datetime()) :: integer()
-  def to_gregorian_seconds(%NaiveDateTime{
+  def to_gregorian_seconds(%{
         calendar: calendar,
         year: year,
         month: month,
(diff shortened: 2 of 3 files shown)


lib/elixir/lib/calendar/date.ex
```diff
@@ -398,6 +398,76 @@ defmodule Date do
...
+      iex> Date.from_gregorian_days(-1)
+      {:error, :invalid_days}
+
+  """
+  @spec from_gregorian_days(non_neg_integer(), Calendar.calendar()) :: {:ok, t} | {:error, atom}
+  def from_gregorian_days(days, calendar \\ Calendar.ISO)
+
+  def from_gregorian_days(days, calendar) when is_integer(days) and days >= 0 do
+    days
+    |> :calendar.gregorian_days_to_date()
```
kipcole9, line 419: I thought one of the original design goals was to not rely on `:calendar` module?

kipcole9: Wouldn't `from_iso_days/1` also support this without `:calendar` module required?

lib/elixir/lib/calendar/date.ex
```diff
@@ -398,6 +398,76 @@ defmodule Date do
...
+      iex> Date.from_gregorian_days(730_485)
+      {:ok, ~D[2000-01-01]}
+      iex> Date.from_gregorian_days(-1)
+      {:error, :invalid_days}
+
+  """
+  @spec from_gregorian_days(non_neg_integer(), Calendar.calendar()) :: {:ok, t} | {:error, atom}
+  def from_gregorian_days(days, calendar \\ Calendar.ISO)
+
+  def from_gregorian_days(days, calendar) when is_integer(days) and days >= 0 do
```
kipcole9, line 417: Curious why you disallow negative days given that dates extend backwards as well and are supported by the calendar behaviour?

josevalim: Correct, we should allow negative days. :)

lib/elixir/lib/calendar/date.ex
josevalim: Correct. We should implement this using the current from_iso_days function, doing a very quick conversion from iso days to gregorian days (I think we need to add or remove 365, i don't remember precisely).

josevalim:
Thanks @ukrbublik , this is a great starting point! :heart: The only restriction is that we should not use the `:calendar` module since it has restrictions that our implementation does not. Instead use our existing iso days functionality.

ukrbublik:
@josevalim Updated PR, please check

lib/elixir/lib/calendar/naive_datetime.ex
```diff
@@ -769,6 +773,84 @@ defmodule NaiveDateTime do
...
+  ## Examples
+
+      iex> NaiveDateTime.to_gregorian_seconds(~N[0000-01-01 00:00:01])
+      1
+      iex> NaiveDateTime.to_gregorian_seconds(~N[2020-05-01 00:26:31.005])
+      63_755_511_991
+
+  """
+  @spec to_gregorian_seconds(t) :: integer()
+
```
josevalim, line 828:
```suggestion
```

lib/elixir/lib/calendar/naive_datetime.ex
```diff
@@ -769,6 +773,84 @@ defmodule NaiveDateTime do
...
+      iex> NaiveDateTime.from_gregorian_seconds(1)
+      ~N[0000-01-01 00:00:01]
+      iex> NaiveDateTime.from_gregorian_seconds(63_755_511_991, {5000, 3})
+      ~N[2020-05-01 00:26:31.005]
+      iex> NaiveDateTime.from_gregorian_seconds(-1)
+      ~N[-0001-12-31 23:59:59]
+
+  """
+  @spec from_gregorian_seconds(integer(), Calendar.microsecond(), Calendar.calendar()) :: t
+
```
josevalim, line 790:
```suggestion
```

lib/elixir/lib/calendar/date.ex
```diff
@@ -398,6 +398,44 @@ defmodule Date do
...
+      iex> Date.to_gregorian_days(~D[0000-01-02])
+      1
+      iex> Date.to_gregorian_days(~D[2000-01-01])
+      730_485
+      iex> Date.to_gregorian_days(~N[2000-01-01 00:00:00])
+      730_485
+
+  """
+  @spec to_gregorian_days(t) :: integer()
+
```
josevalim, line 433:
```suggestion
```

lib/elixir/lib/calendar/time.ex
```diff
@@ -358,6 +359,61 @@ defmodule Time do
...
+      iex> Time.from_seconds_after_midnight(100_000)
+      ~T[03:46:40]
+
+  """
+  @spec from_seconds_after_midnight(
+          integer(),
+          Calendar.microsecond(),
+          Calendar.calendar()
+        ) :: t
+
```
josevalim, line 382:
```suggestion
```

lib/elixir/lib/calendar/time.ex
```diff
@@ -358,6 +359,61 @@ defmodule Time do
...
+  ## Examples
+
+      iex> Time.to_seconds_after_midnight(~T[23:30:15])
+      84615
+      iex> Time.to_seconds_after_midnight(~N[2010-04-17 23:30:15.999])
+      84615
+
+  """
+  @spec to_seconds_after_midnight(t) :: Calendar.seconds_after_midnight()
+
```
josevalim, line 411:
```suggestion
```

lib/elixir/lib/calendar/date.ex
```diff
@@ -398,6 +398,43 @@ defmodule Date do
...
+      ~D[-0001-12-31]
+
+  """
+  @spec from_gregorian_days(integer(), Calendar.calendar()) :: t
+  def from_gregorian_days(days, calendar \\ Calendar.ISO) when is_integer(days) do
+    from_iso_days({days, 0}, calendar)
+  end
+
+  @doc """
+  Converts a `Date` struct to a number of gregorian days.
```
wojtekmach, line 420: since the function accepts not just Date structs but also NaiveDateTime etc, we should subtly change this here: (to. match e.g. Date.add/2 docs)

```suggestion
  Converts a `date` struct to a number of gregorian days.
```

Same goes for other similar places.

lib/elixir/lib/calendar/date.ex
```diff
@@ -398,6 +398,43 @@ defmodule Date do
...
+
+      iex> Date.to_gregorian_days(~D[0000-01-02])
+      1
+      iex> Date.to_gregorian_days(~D[2000-01-01])
+      730_485
+      iex> Date.to_gregorian_days(~N[2000-01-01 00:00:00])
+      730_485
+
+  """
+  @spec to_gregorian_days(t) :: integer()
```
wojtekmach, line 432:
```suggestion
  @spec to_gregorian_days(Calendar.date()) :: integer()
```

lib/elixir/lib/calendar/naive_datetime.ex
```diff
@@ -769,6 +773,82 @@ defmodule NaiveDateTime do
...
+  ## Examples
+
+      iex> NaiveDateTime.to_gregorian_seconds(~N[0000-01-01 00:00:01])
+      1
+      iex> NaiveDateTime.to_gregorian_seconds(~N[2020-05-01 00:26:31.005])
+      63_755_511_991
+
+  """
+  @spec to_gregorian_seconds(t) :: integer()
+  def to_gregorian_seconds(%NaiveDateTime{
```
wojtekmach, line 827: similar to previous comment, `t()` vs `Calendar.naive_datetime()`.

lib/elixir/lib/calendar/time.ex
```diff
@@ -358,6 +359,59 @@ defmodule Time do
...
+  ## Examples
+
+      iex> Time.to_seconds_after_midnight(~T[23:30:15])
+      84615
+      iex> Time.to_seconds_after_midnight(~N[2010-04-17 23:30:15.999])
+      84615
+
+  """
+  @spec to_seconds_after_midnight(t) :: Calendar.seconds_after_midnight()
```
wojtekmach, line 409: similar to previous comment, `t()` vs `Calendar.time()`.

Also, there's no `Calendar.seconds_after_midnight()` type, should we use `integer()` since that's what we use in `from_seconds_after_midnight/3` too?

```suggestion
  @spec to_seconds_after_midnight(Calendar.time()) :: integer()
```

lib/elixir/lib/calendar/date.ex
```diff
@@ -398,6 +398,43 @@ defmodule Date do
...
+
+      iex> Date.from_gregorian_days(1)
+      ~D[0000-01-02]
+      iex> Date.from_gregorian_days(730_485)
+      ~D[2000-01-01]
+      iex> Date.from_gregorian_days(-1)
+      ~D[-0001-12-31]
+
+  """
+  @spec from_gregorian_days(integer(), Calendar.calendar()) :: t
```
fertapric, line 414: I think we could document the version in which these new functions are introduced 🙂 

```suggestion
  @doc since: "1.11.0"
  @spec from_gregorian_days(integer(), Calendar.calendar()) :: t
```

The same applies for the rest of the functions.

lib/elixir/lib/calendar/time.ex
ukrbublik: Right, thanks!

lib/elixir/lib/calendar/naive_datetime.ex
fertapric: @ukrbublik the pattern matching on the struct must also be removed:

```suggestion
  def to_gregorian_seconds(%{
```

lib/elixir/lib/calendar/date.ex
ukrbublik: Done

fertapric:
Thanks @ukrbublik! 💜

PR #10017, opened 2020-05-05
Adjust @type body for Registry to match erlang ets match_spec Result

tarree:
Per the docs for [Registry.select/2](https://hexdocs.pm/elixir/Registry.html#select/2) you can provide a map for the body part of the spec, currently body is defined as `[atom | tuple]`, I've changed this now to `[term]` to match the erlang match_spec Result used by :ets.select/2

Thinking that perhaps `@type guard` should also be changed to term to match the erlang match_spec as well though?


josevalim:

PR #10019, opened 2020-05-06
Fix link in ExUnit doc

NickNeck:


wojtekmach:
The previous code was correct but didn’t work due to ExDoc bug that has since been fixed on master. See: https://hexdocs.pm/ex_unit/master/ExUnit.html#module-integration-with-mix

PR #10020, opened 2020-05-06
Improve performance of Enum.with_index/2

andersonmcook:
This is a proposed refactor of `Enum.with_index/2` with the goal of improving performance so that it is faster and consumes less memory.

I've created a small repository that benchmarks the current implementation with the new one [here](https://github.com/andersonmcook/with_index).

I've included my own benchmarks in the README, but would love if someone could replicate or disprove my results.

## Benchmarking Synopsis
- List: The updated function is approximately twice as fast and uses half the memory.
- Map: The updated function is approximately twice as fast and uses half the memory.
- MapSet: The updated function is moderately faster and uses moderately less memory.
- Range: The updated function is moderately faster and uses moderately less memory.
- Stream: The updated function is minimally faster and uses minimally less memory.

I realize this is a PR that nobody has asked for and the gains I've seen may not be universal, so I won't get my feelings hurt if this does not get merged. :)

### DIFF
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -3084,10 +3084,18 @@ defmodule Enum do
   """
   @spec with_index(t, integer) :: [{element, index}]
   def with_index(enumerable, offset \\ 0) do
-    map_reduce(enumerable, offset, fn x, acc ->
-      {{x, acc}, acc + 1}
-    end)
-    |> elem(0)
+    enumerable
+    |> to_list()
+    |> do_with_index(offset)
+  end
+
+  @spec do_with_index(list, integer) :: [{element, index}]
+  defp do_with_index([], _) do
+    []
+  end
+
+  defp do_with_index([head | tail], index) do
+    [{head, index} | do_with_index(tail, index + 1)]
   end
 
   @doc """


josevalim:
Thanks @andersonmcook! Which Erlang version did you use? I am actually surprised by these number. I would expect lists and maps to be slightly faster in your case, but not ranges. I am afraid this may be caused by https://github.com/erlang/otp/pull/2594. Can you please try benchmarking on Erlang 21 (or 23 once it comes out)?

andersonmcook:
> Thanks @andersonmcook! Which Erlang version did you use? I am actually surprised by these number. I would expect lists and maps to be slightly faster in your case, but not ranges. I am afraid this may be caused by [erlang/otp#2594](https://github.com/erlang/otp/pull/2594). Can you please try benchmarking on Erlang 21 (or 23 once it comes out)?

I ran my benchmarks on:
```
Operating System: macOS
CPU Information: Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
Number of Available Cores: 12
Available memory: 16 GB
Elixir 1.10.3
Erlang 22.3.3
```

But I'd be happy to try this on Erlang 21 tonight!

andersonmcook:
OTP 21 Benchmarks are roughly equivalent.  I'll try one of the RCs for 23 later.
https://github.com/andersonmcook/with_index/blob/master/otp-21.md

josevalim:
Thanks! I think we will have to wait the final release for OTP 23, as that PR has been merged recently. Once we do that, we should be good for merging.

andersonmcook:
Just to complete the trifecta, here are the benchmarks using OTP 23 RC 3:
https://github.com/andersonmcook/with_index/blob/master/otp-23.md

Thanks!

josevalim:
Oh, RC3 already has the patch, so we are good to go. Thank you!

josevalim:

PR #10023, opened 2020-05-08
Add note that System.stacktrace/0 will always return an empty list

michalmuskala:
Starting with OTP 23 erlang:get_stacktrace/0 will always return an
empty list. The function will be completely removed in OTP 24 - we
should plan appropriately for that.

### DIFF
--- a/lib/elixir/lib/system.ex
+++ b/lib/elixir/lib/system.ex
@@ -587,7 +587,7 @@ defmodule System do
   `__STACKTRACE__/0` inside a rescue/catch. If you want to support
   earlier Elixir versions, move `System.stacktrace/0` inside a rescue/catch.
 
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/system.ex
```diff
@@ -587,6 +587,8 @@ defmodule System do
   `__STACKTRACE__/0` inside a rescue/catch. If you want to support
   earlier Elixir versions, move `System.stacktrace/0` inside a rescue/catch.
 
+  Since OTP 23 this always returns an empty list.
```
josevalim, line 590:
```suggestion
  From Erlang/OTP 23+ this always returns an empty list.
```

josevalim:

Issue #10024, opened 2020-05-08
Exception message reported by Task is mixing up PID of task and PID of caller


mtarnovan:
Exception message reported by `Task` is mixing up PID of task and PID of caller in 1.10:

```
Erlang/OTP 22 [erts-10.5] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [hipe]

Interactive Elixir (1.10.1) - press Ctrl+C to exit (type h() ENTER for help)
iex(1)> self()                            
#PID<0.108.0>
iex(2)> Task.start(fn -> raise "boom" end)
{:ok, #PID<0.111.0>}
iex(3)> 
18:29:09.767 [error] Task #PID<0.108.0> started from #PID<0.111.0> terminating
** (RuntimeError) boom
    (stdlib 3.10) erl_eval.erl:678: :erl_eval.do_apply/6
    (elixir 1.10.1) lib/task/supervised.ex:90: Task.Supervised.invoke_mfa/2
    (stdlib 3.10) proc_lib.erl:249: :proc_lib.init_p_do_apply/3
Function: #Function<21.126501267/0 in :erl_eval.expr/5>
    Args: []
 ```
This should read: `[error] Task #PID<0.111.0> started from #PID<0.108.0> terminating`

This works correctly on 1.9:
```Erlang/OTP 22 [erts-10.5] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [hipe]

Interactive Elixir (1.9.1) - press Ctrl+C to exit (type h() ENTER for help)
iex(1)> self()                            
#PID<0.108.0>
iex(2)> Task.start(fn -> raise "boom" end)
{:ok, #PID<0.111.0>}
iex(3)> 
18:32:09.383 [error] Task #PID<0.111.0> started from #PID<0.108.0> terminating
** (RuntimeError) boom
    (stdlib) erl_eval.erl:678: :erl_eval.do_apply/6
    (elixir) lib/task/supervised.ex:90: Task.Supervised.invoke_mfa/2
    (stdlib) proc_lib.erl:249: :proc_lib.init_p_do_apply/3
Function: #Function<21.126501267/0 in :erl_eval.expr/5>
    Args: []
```

ryanzidago:
Hey, I would gladly open up a pull-request and take care of it !

josevalim:
Please do! Also please include a test if possible.

PR #10025, opened 2020-05-08
fixes issue #10024

ryanzidago:
Fixed the issue, now I will figure out how to write the test.


josevalim:
@ryanzidago actually, i got a perfect test. Change this line:

https://github.com/elixir-lang/elixir/blob/master/lib/logger/test/logger/translator_test.exs#L315

To this:

    \[error\] Task #{inspect(pid)} started from #{inspect(self())} terminating

And it should be good!

ryanzidago:
Many thanks José! 

Voilà ! The test is updated.

josevalim:

PR #10028, opened 2020-05-12
Introduce MIX_XDG

josevalim:
Unfortunately we cannot support XDG out of the box due to backwards compatibility. Today we already allow you to opt-in but it requires setting multiple environment variables. This makes the opt-in API much simpler.

We had lengthy previous discussions on this topic and on backwards compatibility, I am not planning to rehash those points here. [See this comment for a summary](https://github.com/elixir-lang/elixir/pull/9942#discussion_r409569099). For the complete discussion, search for XDG in previous issues and PRs.

If you would like to give feedback in this issue, please let me know which option you prefer:

  1. This PR
  2. Keep things as is (no changes)
  3. Remove XDG support since we can't do it automatically

If the conversation derails, as it has happened in the past, then I will make an executive decision. :) Thank you!

### DIFF
--- a/lib/mix/lib/mix.ex
+++ b/lib/mix/lib/mix.ex
@@ -258,7 +258,7 @@ defmodule Mix do
       (default: `~/.mix/rebar3`)
     * `MIX_XDG` - asks Mix to follow the [XDG Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html)
       for its home directory and configuration files. This behaviour needs to
-      be opt-in due to backwards compatibilit.y `MIX_HOME` has higher preference
+      be opt-in due to backwards compatibility. `MIX_HOME` has higher preference
       than `MIX_XDG`. If none of the variables are set, the default directory
       `~/.mix` will be used
 
--- a/lib/mix/lib/mix.ex
+++ b/lib/mix/lib/mix.ex
@@ -256,11 +256,11 @@ defmodule Mix do
       (default: `~/.mix/rebar`)
     * `MIX_REBAR3` - path to rebar3 command that overrides the one Mix installs
       (default: `~/.mix/rebar3`)
-
-  If `MIX_HOME` is not set then the environment variables `XDG_DATA_HOME` and
-  `XDG_CONFIG_HOME` will be used as directories for storage of data and
-  configuration respectively. If none of the variables are set the default
-  directory `~/.mix` will be used.
+    * `MIX_XDG` - asks Mix to follow the [XDG Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html)
+      for its home directory and configuration files. This behaviour needs to
+      be opt-in due to backwards compatibilit.y `MIX_HOME` has higher preference
+      than `MIX_XDG`. If none of the variables are set, the default directory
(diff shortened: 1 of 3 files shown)


lib/mix/lib/mix/utils.ex
```diff
@@ -15,28 +15,33 @@ defmodule Mix.Utils do
...
   defp mix_home_xdg_lookup(xdg) do
-    case {System.get_env("MIX_HOME"), System.get_env(xdg)} do
-      {directory, _} when is_binary(directory) -> directory
-      {nil, directory} when is_binary(directory) -> Path.join(directory, "mix")
-      {nil, nil} -> Path.expand("~/.mix")
+    cond do
+      dir = System.get_env("MIX_HOME") ->
+        dir
+
+      System.get_env("MIX_XDG") ->
```
maennchen, line 40: Shouldn't this check for the value as well? Something along the lines of this:

```suggestion
      System.get_env("MIX_XDG") in ["1", "true"]->
```

maennchen:
I personally like this PR more than the alternatives that are possible without introducing a breaking change.

lib/mix/lib/mix.ex
```diff
@@ -256,11 +256,11 @@ defmodule Mix do
...
     * `MIX_REBAR3` - path to rebar3 command that overrides the one Mix installs
       (default: `~/.mix/rebar3`)
-
-  If `MIX_HOME` is not set then the environment variables `XDG_DATA_HOME` and
-  `XDG_CONFIG_HOME` will be used as directories for storage of data and
-  configuration respectively. If none of the variables are set the default
-  directory `~/.mix` will be used.
+    * `MIX_XDG` - asks Mix to follow the [XDG Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html)
+      for its home directory and configuration files. This behaviour needs to
+      be opt-in due to backwards compatibilit.y `MIX_HOME` has higher preference
```
fertapric, line 261:
```suggestion
      be opt-in due to backwards compatibility. `MIX_HOME` has higher preference
```

lib/mix/lib/mix/utils.ex
```diff
@@ -15,28 +15,33 @@ defmodule Mix.Utils do
...
   Gets possible location of global Mix configuration.
 
   Possible locations:
 
-   * `~/.mix`
-   * `MIX_HOME`
-   * `XDG_CONFIG_HOME/mix`
+     * `XDG_CONFIG_HOME/mix`
+     * `MIX_HOME`
+     * `~/.mix`
```
fertapric, line 28: Although the documentation does not talk about the priority of those locations, I would order them by it:

```suggestion
     * `MIX_HOME`
     * `XDG_CONFIG_HOME/mix` (if `MIX_XDG` is set)
     * `~/.mix`
```

In addition, I would mention that `XDG_CONFIG_HOME` only works if `MIX_XDG` is set because it might not be obvious that you need to set that environment variable (for example if you land directly in the documentation of this function). The same applies to `mix_home/0` 🙂

fertapric: Ups! Nevermind. This module is not publicly documented.

josevalim:

PR #10030, opened 2020-05-12
Add IO.ANSI.clear_scrollback/0

britto:
This control sequence clears the whole scrollback buffer in xterm and other terminal applications. It is particularly useful in conjunction with `clear/0` when we want to print something continuously without polluting the buffer.

* https://en.wikipedia.org/wiki/ANSI_escape_code#Terminal_output_sequences
* https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Functions-using-CSI-_-ordered-by-the-final-character_s_

### DIFF
--- a/lib/elixir/lib/io/ansi.ex
+++ b/lib/elixir/lib/io/ansi.ex
@@ -31,7 +31,7 @@ defmodule IO.ANSI do
       IO.puts(formatted_text)
 
   A higher level and more convenient API is also available via `IO.ANSI.format/1`,
-  where you use atoms to represent each ANSI escape sequence and by default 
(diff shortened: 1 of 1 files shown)


josevalim:
Unfortutnately, opposite to `.clear`, `.clear_scrollback` does not work on Windows, not even the new Windows Terminal, so I'd prefer if we didn't support it. Others can invoke it directly though! Thanks @britto!

britto:
Thanks, @josevalim. I'll leave a link to their docs here for further reference if that feature eventually gets implemented in Windows Terminal.

https://docs.microsoft.com/en-us/windows/console/console-virtual-terminal-sequences#text-modification

PR #10031, opened 2020-05-13
Add ANSI color support for Windows

### LINKED ISSUE #9810: Enable ANSI escape codes for IEx when it is run in new Windows Terminal (Preview)
### Environment

* Operating system: 
Erlang/OTP 21 [erts-10.3] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1]
Elixir 1.10.1 (compiled with Erlang/OTP 21)

### Current behavior

Example running `clear` in IEx doesn't clear screen in new Windows Terminal that currently in preview. You have to run `Application.put_env(:elixir, :ansi_enabled, true)` to make it work. Windows Terminal (Preview) can be downloaded from Microsoft Store.

### Expected behavior

IEx to detect that it's run in new Windows Terminal and enable ANSI escape codes by default,

How to detect https://stackoverflow.com/questions/59733731/how-do-detect-if-running-in-the-new-windows-terminal

>You can check for the WT_SESSION environmental variable which is set to a v4 UUID: https://github.com/microsoft/terminal/issues/1040
>If you're looking for a quick and dirty way to check, this should work:
`!!process.env.WT_SESSION`
>There's also a more elaborate method you can use, taking advantage of is-uuid and process.platform:
`const isUUID = require("is-uuid")
process.platform === "win32" && isUUID.v4(process.env.WT_SESSION)`

Feature Request: Clients of terminal need a way to inquire about the capabilities of the terminal https://github.com/microsoft/terminal/issues/1040

josevalim:
Closes #9810.

### DIFF
--- a/bin/elixir.bat
+++ b/bin/elixir.bat
@@ -152,6 +152,10 @@ for  /d %%d in ("!SCRIPT_PATH!..\lib\*.") do (
 )
 
 :run
+reg query HKCU\Console /v VirtualTerminalLevel 2>nul | findstr /e "0x1" >nul 2>nul
+if %errorlevel% == 0 (
+  set beforeExtra=-elixir ansi_enabled true !beforeExtra!
+)
 if not !runMode! == "iex" (
   set beforeExtra=-noshell -s elixir start_cli !beforeExtra!
 )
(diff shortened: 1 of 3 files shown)


josevalim:

PR #10033, opened 2020-05-13
Allow a sorting function on Enum.min_max_by/3,4

jeroenvisser101:
This should keep backwards compatibility as it relies on the arity of
the sorter to determine if it is a empty_fallback or a sorter.

### DIFF
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -1865,9 +1865,6 @@ defmodule Enum do
   If multiple elements are considered maximal or minimal, the first one
   that was found is returned.
 
-  Calls the provided `empty_fallback` function and returns its value if
-  `enumerable` is empty. The default `empty_fallback` raises `Enum.EmptyError`.
-
   ## Examples
 
       iex> Enum.min_max_by(["aaa", "bb", "c"], fn x -> String.length(x) end)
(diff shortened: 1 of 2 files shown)


lib/elixir/test/elixir/enum_test.exs
```diff
@@ -600,9 +600,31 @@ defmodule EnumTest do
...
+    assert Enum.min_max_by(["aaa", "a", "aa"], fn x -> String.length(x) end, &>/2) == {"aaa", "a"}
+
     assert_runs_enumeration_only_once(&Enum.min_max_by(&1, fn x -> x end, fn -> nil end))
   end
 
+  test "min_max_by/4" do
+    users = [%{id: 1, date: ~D[2019-01-01]}, %{id: 2, date: ~D[2020-01-01]}]
+
+    assert Enum.min_max_by(users, & &1.date, Date) ==
+             {%{id: 1, date: ~D[2019-01-01]}, %{id: 2, date: ~D[2020-01-01]}}
```
jeroenvisser101, line 612: Note: min/max etc. also check that it correctly handles equal values, but those functions use a different sorter, and `Enum.min_max_by` has never had this behaviour as far as I could see.

lib/elixir/lib/enum.ex
```diff
@@ -1879,11 +1876,60 @@ defmodule Enum do
       iex> Enum.min_max_by([], &String.length/1, fn -> {nil, nil} end)
       {nil, nil}
 
+  The fact this function uses Erlang's term ordering means that the
```
ryanzidago, line 1879:
```suggestion
  The fact that this function uses Erlang's term ordering means that the
```

jeroenvisser101: Good suggestion, @josevalim should I change both of these (as I copied it from min_by) in this PR?

josevalim: We can do them separately. I will also take the opportunity to run a grammar check on the whole doc.

josevalim:

PR #10034, opened 2020-05-13
Document data and fun arguments to `update_in/3`

Odaeus:
I thought it would be helpful to make it clear what the arguments to `update_in/3` should be. In particular that the given function receives the value and is expected to return the new value and not a tuple like `get_and_update_in`.


josevalim:

Odaeus:
Amazing 😀, thank you both!

PR #10035, opened 2020-05-14
Add ExUnit.fetch_test_supervisor/0

wojtekmach:
Ref https://github.com/phoenixframework/phoenix_live_view/issues/887

### DIFF
--- a/lib/ex_unit/lib/ex_unit.ex
+++ b/lib/ex_unit/lib/ex_unit.ex
@@ -360,6 +360,30 @@ defmodule ExUnit do
     configure(after_suite: [function | current_callbacks])
   end
 
+  @doc """
+  Fetches the test supervisor for the current test.
+
+  This is the same supervisor as used by `ExUnit.Callbacks.start_supervised/2`
+  and similar, see `ExUnit.Callbacks` module documentation for more information.
+  """
+  @doc since: "1.11.0"
+  @spec fetch_test_supervisor!() :: pid()
+  def fetch_test_supervisor!() do
+    case ExUnit.OnExitHandler.get_supervisor(self()) do
+      {:ok, nil} ->
+        opts = [strategy: :one_for_one, max_restarts: 1_000_000, max_seconds: 1]
+        {:ok, sup} = Supervisor.start_link([], opts)
+        ExUnit.OnExitHandler.put_supervisor(self(), sup)
+        sup
+
+      {:ok, sup} ->
+        sup
+
+      :error ->
+        raise ArgumentError, "fetch_test_supervisor!/0 can only be invoked from the test process"
+    end
+  end
+
   # Persists default values in application
   # environment before the test suite starts.
   defp persist_defaults(config) do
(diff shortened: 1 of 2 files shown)


lib/ex_unit/lib/ex_unit.ex
```diff
@@ -360,6 +360,30 @@ defmodule ExUnit do
...
 
+  @doc """
+  Fetches the test supervisor for the current test.
+
+  This is the same supervisor as used by `ExUnit.Callbacks.start_supervised/2`
+  and similar, see `ExUnit.Callbacks` module documentation for more information.
+  """
+  @doc since: "1.11.0"
+  @spec fetch_test_supervisor!() :: pid()
+  def fetch_test_supervisor!() do
```
whatyouhide, line 371: I think this should go in the `ExUnit.Callbacks` module. I mean, in the same place where `start_supervised` here, rather :P

wojtekmach: we were working on this with José, we think this will be used rather infrequently so the idea to put it here was to not pollute user tests cases.

lib/ex_unit/lib/ex_unit/callbacks.ex
```diff
@@ -397,20 +395,7 @@ defmodule ExUnit.Callbacks do
...
-        {:ok, nil} ->
-          {:ok, sup} = Supervisor.start_link([], @supervisor_opts)
-          ExUnit.OnExitHandler.put_supervisor(self(), sup)
-          sup
-
-        {:ok, sup} ->
-          sup
-
-        :error ->
-          raise ArgumentError, "start_supervised/2 can only be invoked from the test process"
```
wojtekmach, line 411: see the test failure: https://github.com/elixir-lang/elixir/pull/10035/checks?check_run_id=675089179 (I will fix it soon), I was worried that by changing the exception message this may become confusing. wdyt?

lib/ex_unit/lib/ex_unit.ex
whatyouhide: Agreed it's not going to be used often. Still, I think semantically it belongs in `ExUnit.Callbacks`. Not a strong preference so up to you folks :)

josevalim: @whatyouhide I am fine with both too. :P @fertapric @ericmj does any of you have a pref?

lib/ex_unit/lib/ex_unit/callbacks.ex
josevalim: We can refer to the functions that may call it. Something like "fetch_test_supervisor!/start_supervised/start_supervised! can only ....".

josevalim: @wojtekmach another option is to name it `fetch_test_supervisor`, return `{:ok, ...} | :error` and always let the caller raise. This may be cleaner after all. Thoughts?

josevalim:

whatyouhide:
@dvic are you trapping exits in your test process?

dvic:
> @dvic are you trapping exits in your test process?

Nope. But of course I can't reproduce it anymore 😔

PR #10038, opened 2020-05-15
Add CHANGELOG.md to docs

wojtekmach:
Rendering it gives us syntax highlighting, auto-linking, sidebar sections, and more. See how v1.10 changelog would be rendered: http://wojtekmach.pl/otp_docs/elixir/elixir/changelog.html

We could duplicate this changelog to mix/, eex/, etc but I opted not to.


whatyouhide:
Thank you so much @wojtekmach! 💟

eksperimental:
Great. I always felt CHANGELOG was missing from the docs

PR #10041, opened 2020-05-16
Pass a valid os option to filename:basedir/3

michallepicki:
unix is not one of the documented options and will get interpreted as linux


josevalim:

PR #10046, opened 2020-05-18
Add :nfkd and :nfkc forms to String.normalize/2

kipcole9:
1. Added `:nfkd` and `:nfkc` normalisation forms to `String.normalize/2`
2. Added two doctests which illustrate non-roundtrip behaviour
3. Added warning that these normalisation forms do not round-trip
4. Updated the description with the 2 additional Erlang functions referenced

I did not add any unit tests since that would now seem to be testing Erlang's implementation - let me know if you would like some additional tests added however.

### DIFF
--- a/lib/elixir/lib/string.ex
+++ b/lib/elixir/lib/string.ex
@@ -682,8 +682,9 @@ defmodule String do
 
   Invalid Unicode codepoints are skipped and the remaining of
   the string is converted. If you want the algorith to stop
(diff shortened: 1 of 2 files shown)


josevalim:
We do need some tests, especially some that make sure we handle invalid bytes properly.

kipcole9:
Add the error cases (using the existing ones for :nfc and :nfd) and added additional tests from NormalizationTest.txt

josevalim:

PR #10050, opened 2020-05-19
Correct the specs for Enum aggregate functions

michallepicki:
Followup after #10049

Most importantly, I added specs for function clauses with `(element, element -> boolean) | module()`.

I moved the compatibility clause for `Enum.min_max_by/3` to before documentation with @doc false, just like it is done for other related functions.

I think because of the `@doc false` the specs (both "old" and "new") are not showing up in documentation but I think it's fine (full arity spec is showing up).

I moved the specs a bit. I think this makes it clearer which specs are related to old (possibly deprecated in the future?) function clauses. If it is preferred to keep all specs in one place I can move them back.

### DIFF
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -1532,6 +1532,7 @@ defmodule Enum do
   end
 
   @doc false
+  @spec max(t, (() -> empty_result)) :: element | empty_result when empty_result: any
   def max(enumerable, empty_fallback) when is_function(empty_fallback, 0) do
     max(enumerable, &>=/2, empty_fallback)
   end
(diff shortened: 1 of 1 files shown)


michallepicki:
Now that I am thinking about it:
```
 Therefore, if you want
  to compare structs, most structs provide a "compare" function, such as
  `Date.compare/2`, which receives two structs and returns `:lt` (less than),
  `:eq` (equal), and `:gt` (greater than). If you pass a module as the
  sorting function, Elixir will automatically use the `compare/2` function
  of said module:
```

Shouldn't all these say `:lt | :eq | :gt` instead of `boolean()` ?

michallepicki:
> Shouldn't all these say :lt | :eq | :gt instead of boolean() ?

I checked, it looks like `:lt | :eq | :gt` should be returned only for the compare function inside module, not the regular compare function. So this is fine. I may have misread but I will think if a slight reword could clear that up and I'll possibly send a followup PR

wojtekmach:
> `(element, element -> boolean) | module()`

I believe the specs are correct, when we pass module we use compare/2, otherwise we expect a function that returns boolean.

josevalim:

PR #10052, opened 2020-05-19
Added @impl to application.ex generated by mix new

akoutmos:
Whenever I create a new project with `mix new app_name --sup`, https://github.com/akoutmos/doctor validation will fail since the generated `application.ex` file doesn't have correct documentation. Not to say that my personal library should dictate how Elixir core does things....but more so I just noticed that for correctness, the generated `application.ex` file should annotate which functions satisfy the `Application` behaviour.

Thanks in advance for any feedback :)


fertapric:
Thanks @akoutmos! 💜

PR #10053, opened 2020-05-21
Add List.new/0

jjl:
It's already there for Map, MapSet and even Keyword (which would amount to the same thing).

### DIFF
--- a/lib/elixir/lib/list.ex
+++ b/lib/elixir/lib/list.ex
@@ -446,6 +446,10 @@ defmodule List do
     end
   end
 
+  @doc "Returns the empty list."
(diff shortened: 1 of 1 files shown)


josevalim:
The only reason it exists for Map and Keyword is because in the early days we had a `Dict` behaviour and `new` was part of the contract. Given there isn't really a need (or at least one was not given) for List.new, I will go ahead and close this. :)

jjl:
I think there probably should be a behaviour for being able to create an empty version of yourself from the name of the controlling module, although that's more controversial than I thought this one would be and definitely more than I can deal with today.

PR #10058, opened 2020-05-24
Add mix format.gen command to generate .formatter.exs in current project

TheFirstAvenger:
Adds a `mix format.gen` command to generate an appropriate `.formatter.exs` file in the current project. The file generated uses the same code as the one generated when running `mix new`, detecting umbrella projects and using the appropriate config. Useful for those older projects you jump into and `mix format` fails because formatter wasn't a thing back then.

### DIFF
--- a/lib/mix/lib/mix/tasks/format.ex
+++ b/lib/mix/lib/mix/tasks/format.ex
@@ -320,7 +320,8 @@ defmodule Mix.Tasks.Format do
     if no_entries_in_formatter_opts?(formatter_opts_and_subs) do
       Mix.raise(
         "Expected one or more files/patterns to be given to mix format " <>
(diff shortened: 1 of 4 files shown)


josevalim:
Thanks @TheFirstAvenger! 

I am honestly a bit torn. This is something that will become less and less useful with time and that we will never be able to remove (given the stable nature of the language). So I am thinking it should rather be a separate archive or recomend people to simply copy it from a new app.

TheFirstAvenger:
@josevalim Yes, you are right. As I was writing this, I started to think this might be the case. The times that I run into this are usually when running the formatter as the first step in [Project Ironman](https://github.com/TheFirstAvenger/ironman), and now that I think about it, it would be easy to add this check in there prior to running `mix format` and not add the legacy baggage to Core. I will add it there.

PR #10062, opened 2020-05-25
Fix read of app file located in archive .ez file

wrachwal:
This PR fixes the read of an app file when it is located inside an archive (.ez) file.

The bug was introduced in `v1.10` line with the introduction of `compile_env` feature.

### DIFF
--- a/lib/mix/lib/mix/tasks/app.start.ex
+++ b/lib/mix/lib/mix/tasks/app.start.ex
@@ -140,8 +140,8 @@ defmodule Mix.Tasks.App.Start do
             {:error, {:file.format_error(:enoent), name}}
 
           path ->
-            case :file.consult(path) do
-              {:ok, [{:application, _, properties} = application_data]} ->
+            case consult_app_file_maybe_in_archive(path) do
+              {:ok, {:application, _, properties} = application_data} ->
                 with :ok <- :application.load(application_data) do
                   if compile_env = validate_compile_env? && properties[:compile_env] do
                     # Unfortunately we can only check the current app here,
@@ -157,6 +157,19 @@ defmodule Mix.Tasks.App.Start do
                 {:error, {:file.format_error(reason), name}}
             end
         end
+    end
+  end
+
+  defp consult_app_file_maybe_in_archive(path) do
+    # The path could be located in .ez archive
+    case :erl_prim_loader.get_file(path) do
+      {:ok, bin, _full_name} ->
+        with {:ok, tokens, _} <- :erl_scan.string(String.to_charlist(bin)) do
+          :erl_parse.parse_term(tokens)
+        end
+
+      :error ->
+        {:error, :enoent}
     end
   end
 


wrachwal:
Hmm.. doesn't look tests on Windows failed on my code. No way to check, I use Linux.

josevalim:
The tests are unrelated, so no worries.

However, I have to ask, why are you loading an application from an archive? I would say Mix not loading the application from an archive is actually a feature... as those would be system deps and not app deps.

wrachwal:
Hi @josevalim, I build and provide my users a set of apps, some of which compile enormously long, esp. those generated from 3GPP (LTE & NR aka 5G) ASN.1 specs erlang codecs and elixir records. So the idea was to ship those apps packaged as archives (available if present in `ERL_LIBS` path), let them declare in `extra_applications`, and other dependencies to have as normal dependencies to compile (`deps`).
Compilation time is not the only reason for such approach, but important factor.

lib/mix/lib/mix/tasks/app.start.ex
```diff
@@ -160,6 +160,29 @@ defmodule Mix.Tasks.App.Start do
     end
   end
 
+  defp consult_app_file_maybe_in_archive(path) do
+    case :file.consult(path) do
+      {:ok, terms} ->
+        {:ok, terms}
+
+      {:error, :enotdir} ->
```
josevalim, line 168: We can skip this and fallback to `:erl_prim_loader.get_file(path)` directly.

lib/mix/lib/mix/tasks/app.start.ex
```diff
@@ -160,6 +160,29 @@ defmodule Mix.Tasks.App.Start do
...
+            end
+
+          :error ->
+            {:error, :enotdir}
+        end
+
+      {:error, reason} ->
+        {:error, reason}
+    end
+  end
```
josevalim, line 184: Here is perhaps a simpler implementation. Note we no longer wrap the term in a list though, so you will need to update the caller accordingly. :)

```
  defp consult_app_file_maybe_in_archive(path) do
    # The path could be located in .ez archive
    case :erl_prim_loader.get_file(path) do
      {:ok, bin, _full_name} ->
        with {:ok, tokens, _} <- :erl_scan.string(String.to_charlist(bin)) do
          :erl_parse.parse_term(tokens)
        end

      :error ->
        {:error, :enoent}
    end
  end
```

josevalim:
Thanks! This is better as we always exercise the same code path. :)

josevalim:

wrachwal:
Hi @josevalim, one question: could it be possible to backport this commit onto `v1.10` branch?

josevalim:
Good call. backported!

josevalim:
@wrachwal quick question: have you considered simply unzipping the archives too? They are simply .zip files and you could unzip in the directory too.

wrachwal:
@josevalim, I know but having everything in one 'package' as a unit is convenient. I put some additional metadata into them, they themselves are subject for further distribution to second-order machines. Fewer files, and files which one cannot easily break or move to inconsistency state, makes me feel better.

disclosure: i don't use `archive.build` task, but a custom elixir program which takes a set of top-level apps, analyse dependencies, selects "best" revisions of dependencies in case of mismatch, and when final set of targets is found (often a given dependency may be built from different top level apps), the build stage begins. The build process is recursive, at each step it does either `mix deps.compile` or `mix compile` driven by actual goal until all are done. A small invention (*) applied to `mix.exs` allows to not compile a dependency if the archive for it has already been built in any of earlier steps (moving from `deps/0` to `extra_applications` list in code). Meta data stored in archives allows to decide whether to keep or invalidate final artefacts (the archives) of any earlier build during the current build process (form of 'incremental build').

(*) the solution is handy not only during such builds, but to start hybrid mix apps, i.e. typically a top level mix app is compiled from sources whereas any loadable dependencies are not. the solution boils down to one function, it'd be cool to have it somewhere in Mix ;-) , but well, perhaps it is too specific.

PR #10068, opened 2020-05-29
Remove unnecessary lists:reverse/1 call in Map.new

sashaafm:
I noticed this `:lists.reverse/1` call in the inner `Map.new` functions. AFAIK this is unnecessary since, unlike lists, maps do not have order. Therefore, reversing a list should be irrelevant to creating a map from a list. Unless there's a very obscure detail about Erlang I'm missing 🙂

Thanks


josevalim:
As the tests show, the call is necessary, as the input may have duplicated keys. :)

Issue #10063, opened 2020-05-25
Add new! to Date/Time/NaiveDateTime/DateTime


josevalim:
Currently Date/Time/NaiveDateTime have non-bang versions, so we will add the `!` ones. For DateTime, we will add new/3, which expects a `date, time, time_zone`, similar to `DateTime.from_naive` but with only date and time, as well as the bang variant.

fertapric:
I think `new/*` should also accept a `time_zone_database \\ Calendar.get_time_zone_database()` as a fourth argument to fill the offset info.

josevalim:
Closing in favor of PR.

PR #10069, opened 2020-05-31
Add new! to Calendar Types

jrusso1020:
Add `new!` to Date/Time/NaiveDateTime/DateTime. This also adds
`new` to DateTime.

Closes https://github.com/elixir-lang/elixir/issues/10063

### DIFF
--- a/lib/elixir/lib/calendar/date.ex
+++ b/lib/elixir/lib/calendar/date.ex
@@ -246,8 +246,8 @@ defmodule Date do
       {:ok, value} ->
         value
 
-      {:error, :invalid_date} ->
-        raise ArgumentError, "cannot parse as date, reason: :invalid_date"
+      {:error, reason} ->
+        raise ArgumentError, "cannot build date, reason: #{inspect(reason)}"
     end
   end
 
--- a/lib/elixir/lib/calendar/date.ex
+++ b/lib/elixir/lib/calendar/date.ex
@@ -225,6 +225,33 @@ defmodule Date do
   end
 
   @doc """
+  Builds a new ISO date.
+
+  Expects all values to be integers. Returns `date` if each
+  entry fits its appropriate range, raises if the date is invalid.
+
+  ## Examples
+
+      iex> Date.new!(2000, 1, 1)
+      ~D[2000-01-01]
+      iex> Date.new!(2000, 13, 1)
+      ** (ArgumentError) cannot parse as date, reason: :invalid_date
+      iex> Date.new!(2000, 2, 29)
+      ~D[2000-02-29]
+  """
+  @doc since: "1.11.0"
+  @spec new!(Calendar.year(), Calendar.month(), Calendar.day(), Calendar.calendar()) :: t
+  def new!(year, month, day, calendar \\ Calendar.ISO) do
+    case new(year, month, day, calendar) do
+      {:ok, value} ->
+        value
+
+      {:error, :invalid_date} ->
+        raise ArgumentError, "cannot parse as date, reason: :invalid_date"
+    end
+  end
+
+  @doc """
   Converts the given date to a string according to its calendar.
 
   ### Examples
--- a/lib/elixir/lib/calendar/date.ex
+++ b/lib/elixir/lib/calendar/date.ex
@@ -235,7 +235,7 @@ defmodule Date do
       iex> Date.new!(2000, 1, 1)
       ~D[2000-01-01]
       iex> Date.new!(2000, 13, 1)
-      ** (ArgumentError) cannot parse as date, reason: :invalid_date
+      ** (ArgumentError) cannot build date, reason: :invalid_date
       iex> Date.new!(2000, 2, 29)
       ~D[2000-02-29]
   """
--- a/lib/elixir/lib/calendar/datetime.ex
+++ b/lib/elixir/lib/calendar/datetime.ex
@@ -149,6 +149,28 @@ defmodule DateTime do
         time_zone_database \\ Calendar.get_time_zone_database()
       )
 
+  def new(%Date{calendar: calendar} = date, %Time{calendar: calendar} = time, "Etc/UTC", _db) do
+    %{year: year, month: month, day: day} = date
+    %{hour: hour, minute: minute, second: second, microsecond: microsecond} = time
+
+    datetime = %DateTime{
+      calendar: calendar,
+      year: year,
+      month: month,
+      day: day,
+      hour: hour,
+      minute: minute,
+      second: second,
+      microsecond: microsecond,
+      std_offset: 0,
+      utc_offset: 0,
+      zone_abbr: "UTC",
+      time_zone: "Etc/UTC"
+    }
+
+    {:ok, datetime}
+  end
+
   def new(date, time, time_zone, time_zone_database) do
     with {:ok, naive_datetime} <- NaiveDateTime.new(date, time) do
       from_naive(naive_datetime, time_zone, time_zone_database)
--- a/lib/elixir/lib/calendar/datetime.ex
+++ b/lib/elixir/lib/calendar/datetime.ex
@@ -187,6 +187,7 @@ defmodule DateTime do
       iex> datetime = DateTime.new!(~D[2018-07-28], ~T[12:30:00], "Europe/Copenhagen", FakeTimeZoneDatabase)
       iex> datetime
       #DateTime<2018-07-28 12:30:00+02:00 CEST Europe/Copenhagen>
+
   """
   @doc since: "1.11.0"
   @spec new!(Date.t(), Time.t(), Calendar.time_zone(), Calendar.time_zone_database()) :: t
--- a/lib/elixir/lib/calendar/datetime.ex
+++ b/lib/elixir/lib/calendar/datetime.ex
@@ -94,6 +94,131 @@ defmodule DateTime do
   end
 
   @doc """
+  Builds a datetime from date and time structs
+
+  It expects a time zone to put the `DateTime` in.
+  If the time zone is not passed it will default to "Etc/UTC",
+  which always succeeds. Otherwise, the DateTime is checked against the time zone database
+  given as `time_zone_database`. See the "Time zone database"
+  section in the module documentation.
+
+  ## Examples
+
+      iex> DateTime.new(~D[2016-05-24], ~T[13:26:08.003], "Etc/UTC")
+      {:ok, ~U[2016-05-24 13:26:08.003Z]}
+
+  When the datetime is ambiguous - for instance during changing from summer
+  to winter time - the two possible valid datetimes are returned. First the one
+  that happens first, then the one that happens after.
+
+      iex> {:ambiguous, first_dt, second_dt} = DateTime.new(~D[2018-10-28], ~T[02:30:00], "Europe/Copenhagen", FakeTimeZoneDatabase)
+      iex> first_dt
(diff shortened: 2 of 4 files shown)


lib/elixir/lib/calendar/datetime.ex
```diff
@@ -93,6 +93,131 @@ defmodule DateTime do
...
+  def new(
+        date,
+        time,
+        time_zone \\ "Etc/UTC",
+        time_zone_database \\ Calendar.get_time_zone_database()
+      )
+
+  def new(date, time, time_zone, time_zone_database) do
+    with {:ok, naive_datetime} <- NaiveDateTime.new(date, time) do
+      from_naive(naive_datetime, time_zone, time_zone_database)
```
jrusso1020, line 153: to simplify this I just decided to create a `NaiveDateTime` and use the already existing `from_naive` to accomplish this. If we would prefer not to do this let me know, I was trying to limit the repetition this could cause

josevalim: In this case, we want to avoid the double conversion. I think in this case what you can do is to:
 
1. rewrite `from_naive` as `new/4`
2. `from_naive` is the same as calling `new(naive_datetime, naive_datetime, time_zone, time_zone_database)`. 

This way we keep the code simple (no duplication) and fast (no converting to naive datetime and then datetime). WDYT?

lib/elixir/lib/calendar/date.ex
```diff
@@ -224,6 +224,33 @@ defmodule Date do
...
+  """
+  @doc since: "1.11.0"
+  @spec new!(Calendar.year(), Calendar.month(), Calendar.day(), Calendar.calendar()) :: t
+  def new!(year, month, day, calendar \\ Calendar.ISO) do
+    case new(year, month, day, calendar) do
+      {:ok, value} ->
+        value
+
+      {:error, :invalid_date} ->
+        raise ArgumentError, "cannot parse as date, reason: :invalid_date"
```
josevalim, line 250: In case we add a new reason in the future, I would go with:

```suggestion
      {:error, reason} ->
        raise ArgumentError, "cannot build date, reason: #{inspect(reason)}"
```

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -93,6 +93,131 @@ defmodule DateTime do
...
+      )
+
+  def new!(date, time, time_zone, time_zone_database) do
+    case new(date, time, time_zone, time_zone_database) do
+      {:ok, datetime} ->
+        datetime
+
+      {:ambiguous, time1, time2} ->
+        raise ArgumentError,
+              "cannot parse as datetime, reason: ambiguous datetime, datetime1: #{inspect(time1)}, datetime2: #{
```
josevalim, line 206: We need to copy the reasons from `from_naive` but using `date` and `time` in the error message.

lib/elixir/lib/calendar/naive_datetime.ex
```diff
@@ -246,6 +246,61 @@ defmodule NaiveDateTime do
...
+        calendar \\ Calendar.ISO
+      )
+
+  def new!(year, month, day, hour, minute, second, microsecond, calendar) do
+    case new(year, month, day, hour, minute, second, microsecond, calendar) do
+      {:ok, naive_datetime} ->
+        naive_datetime
+
+      {:error, reason} ->
+        raise ArgumentError, "cannot parse as naive datetime, reason: #{inspect(reason)}"
```
josevalim, line 300:
```suggestion
      {:error, reason} ->
        raise ArgumentError, "cannot build naive datetime, reason: #{inspect(reason)}"
```

lib/elixir/lib/calendar/time.ex
```diff
@@ -140,6 +140,44 @@ defmodule Time do
...
+          Calendar.microsecond() | non_neg_integer,
+          Calendar.calendar()
+        ) :: t
+  def new!(hour, minute, second, microsecond \\ {0, 0}, calendar \\ Calendar.ISO) do
+    case new(hour, minute, second, microsecond, calendar) do
+      {:ok, time} ->
+        time
+
+      {:error, :invalid_time} ->
+        raise ArgumentError, "cannot parse as time, reason: :invalid_time"
```
josevalim, line 177:
```suggestion
      {:error, reason} ->
        raise ArgumentError, "cannot build time, reason: #{inspect(reason)}"
```

josevalim:
Looks great! I have added some comments.

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -93,6 +93,131 @@ defmodule DateTime do
     System.os_time() |> from_unix!(:native, calendar)
   end
 
+  @doc """
+  Builds a datetime from date and time structs
+
+  It expects a time zone to put the `DateTime` in.
+  If the time zone is not passed it will default to "Etc/UTC",
```
whatyouhide, line 100:
```suggestion
  If the time zone is not passed it will default to `"Etc/UTC"`,
```

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -93,6 +93,131 @@ defmodule DateTime do
...
+  which always succeeds. Otherwise, the DateTime is checked against the time zone database
```
whatyouhide, line 101:
```suggestion
  which always succeeds. Otherwise, the `DateTime` is checked against the time zone database
```

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -93,6 +93,131 @@ defmodule DateTime do
...
+      #DateTime<2019-03-31 01:59:59.999999+01:00 CET Europe/Copenhagen>
+      iex> just_after
+      #DateTime<2019-03-31 03:00:00+02:00 CEST Europe/Copenhagen>
+
+  Most of the time there is one, and just one, valid datetime for a certain
+  date and time in a certain time zone.
+
+      iex> {:ok, datetime} = DateTime.new(~D[2018-07-28], ~T[12:30:00], "Europe/Copenhagen", FakeTimeZoneDatabase)
+      iex> datetime
+      #DateTime<2018-07-28 12:30:00+02:00 CEST Europe/Copenhagen>
```
whatyouhide, line 135:
```suggestion
      #DateTime<2018-07-28 12:30:00+02:00 CEST Europe/Copenhagen>

```

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -93,6 +93,131 @@ defmodule DateTime do
...
+
+      iex> DateTime.new!(~D[2019-03-31], ~T[02:30:00], "Europe/Copenhagen", FakeTimeZoneDatabase)
+      ** (ArgumentError) cannot parse as datetime, reason: gap in wall time, just before: #DateTime<2019-03-31 01:59:59.999999+01:00 CET Europe/Copenhagen>, just after: #DateTime<2019-03-31 03:00:00+02:00 CEST Europe/Copenhagen>
+
+  Most of the time there is one, and just one, valid datetime for a certain
+  date and time in a certain time zone.
+
+      iex> datetime = DateTime.new!(~D[2018-07-28], ~T[12:30:00], "Europe/Copenhagen", FakeTimeZoneDatabase)
+      iex> datetime
+      #DateTime<2018-07-28 12:30:00+02:00 CEST Europe/Copenhagen>
```
whatyouhide, line 188:
```suggestion
      #DateTime<2018-07-28 12:30:00+02:00 CEST Europe/Copenhagen>

```

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -93,6 +93,131 @@ defmodule DateTime do
...
+    with {:ok, naive_datetime} <- NaiveDateTime.new(date, time) do
+      from_naive(naive_datetime, time_zone, time_zone_database)
+    end
+  end
+
+  @doc """
+  Builds a datetime from date and time structs
+
+  It expects a time zone to put the `DateTime` in.
+  If the time zone is not passed it will default to "Etc/UTC",
```
whatyouhide, line 161:
```suggestion
  If the time zone is not passed it will default to `"Etc/UTC"`,
```

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -93,6 +93,131 @@ defmodule DateTime do
...
+      )
+
+  def new(date, time, time_zone, time_zone_database) do
+    with {:ok, naive_datetime} <- NaiveDateTime.new(date, time) do
+      from_naive(naive_datetime, time_zone, time_zone_database)
+    end
+  end
+
+  @doc """
+  Builds a datetime from date and time structs
```
whatyouhide, line 158:
```suggestion
  Builds a datetime from date and time structs, raising on errors.
```

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -93,6 +93,131 @@ defmodule DateTime do
     System.os_time() |> from_unix!(:native, calendar)
   end
 
+  @doc """
+  Builds a datetime from date and time structs
```
whatyouhide, line 97:
```suggestion
  Builds a datetime from date and time structs.
```

whatyouhide:
This is super nice. I took the liberty to make a bunch of very tiny chances through the GH interface and commit those directly :)

lib/elixir/lib/calendar/datetime.ex
jrusso1020: @josevalim Thanks for reviewing this so quick! I have a few questions/comments

 I ran into an issue where `Calendar.TimeZoneDatabase.time_zone_periods_from_wall_datetime` expects `(Calendar.naive_datetime(), Calendar.time_zone())`https://github.com/elixir-lang/elixir/blob/master/lib/elixir/lib/calendar/time_zone_database.ex#L66 
So don't I need to create a `NaiveDatetime` in order to determine if it's ambiguous or has a gap?

I understand your second point that makes sense that `from_naive(naive_datetime)` is the same as `new(naive_datetime, naive_datetime)`

So is point 1) mean you want to rewrite `from_naive` to use `new/4` internally? Or that we should rewrite `new/4` using the same functions as `from_naive`? If you could clarify this a little that would be great!

jrusso1020:
Thanks for the quick reviews @josevalim and @whatyouhide! I have made most of the changes requested. I just have one question about `DateTime.new/4`'s implementation

lib/elixir/lib/calendar/datetime.ex
josevalim: I see your point. Given a naive_datetime is required anyway, the current code (as is) is the best way to go.

Perhaps my only suggestion is to add a clause like this:

```elixir
  def new(%Date{calendar: calendar} = date, %Time{calendar: calendar} = time, "Etc/UTC", _db) do
    %{year: year, month: month, day: day} = date
    %{hour: hour, minute: minute, second: second, microsecond: microsecond} = time

    datetime = %DateTime{
      calendar: calendar,
      year: year,
      month: month,
      day: day,
      hour: hour,
      minute: minute,
      second: second,
      microsecond: microsecond,
      std_offset: 0,
      utc_offset: 0,
      zone_abbr: "UTC",
      time_zone: "Etc/UTC"
    }

    {:ok, datetime}
  end
```

So we at least optimize the most common scenario (UTC). WDYT?

jrusso1020: Okay added this clause!

josevalim:

PR #10074, opened 2020-06-01
docs corrections for Mix.Generator.copy_file, Mix.Generator.copy_template

sampriddy:
The documentation for `Mix.Generator.copy_file` and `Mix.Generator.copy_template` had examples that do not match their function specs.

Related: should these examples be run as doctests?

### DIFF
--- a/lib/mix/lib/mix/generator.ex
+++ b/lib/mix/lib/mix/generator.ex
@@ -71,7 +71,7 @@ defmodule Mix.Generator do
 
   ## Examples
 
-      iex> Mix.Generator.copy_file("source/gitignore", ".gitignore", "_build\ndeps\n")
+      iex> Mix.Generator.copy_file("source/gitignore", ".gitignore")
       * creating .gitignore
       true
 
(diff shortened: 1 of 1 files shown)


lib/mix/lib/mix/generator.ex
```diff
@@ -97,7 +97,7 @@ defmodule Mix.Generator do
 
   ## Examples
 
-      iex> Mix.Generator.copy_template("source/gitignore", ".gitignore", [project_path: path])
```
sampriddy, line 100: :project_path is not listed among the valid options, and does not seem to be used

lib/mix/lib/mix/generator.ex
```diff
@@ -71,7 +71,7 @@ defmodule Mix.Generator do
 
   ## Examples
 
-      iex> Mix.Generator.copy_file("source/gitignore", ".gitignore", "_build\ndeps\n")
```
sampriddy, line 74: third argument should be a keyword list of options, this appears to be a copy/paste error from the `create_file` docs

josevalim:

lib/mix/lib/mix/generator.ex
sampriddy: I made a mistake here. I think project_path was meant to be an assign. I'll make a quick followup PR to put this back in.

josevalim: Ah, good call. Maybe we should do this:

    iex> assigns = [project_path: ...]

for clarity?

PR #10085, opened 2020-06-05
Print which deps got unlocked

wojtekmach:
Before this patch, we only printed which deps got unlocked when
`--filter` option was used. Now we do it everywhere else, that is
when passing specific apps to unlock and when using `--unused`.


josevalim:

PR #10086, opened 2020-06-08
Add run_if macro to kernel

esse:
Following the discussion on the mailing list:
https://groups.google.com/d/msgid/elixir-lang-core/f513b901-fd93-4f28-88a2-f496a3c84a21o%40googlegroups.com?utm_medium=email&utm_source=footer

### DIFF
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -3512,6 +3512,29 @@ defmodule Kernel do
   end
 
   @doc """
+  Provides a run_if macro, that accepts three arguments:
+  * argument
+  * condition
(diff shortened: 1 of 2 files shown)


josevalim:
Thanks for the PR but let's wait until the discussion finalizes. There is usually an explicit approval asking for the PR. :) I am in particular not happy with `run_if` because of the implicit dependency on the first argument. For example, I would expect `run_if` to return `nil` if the condition is not true, which is not what is desired here.

esse:
Ah, sure! I thought there was a consensus and it was okay to submit a PR :)

W dniu pon., 8.06.2020 o 11:50 José Valim <notifications@github.com>
napisał(a):

Issue #9603, opened 2019-11-27
Support Erlang/OTP 23 new features


josevalim:
To be implemented when Erlang 23 is released:

  * [x] Show Erlang docs in the shell (see #9843)
  * [x] Tackle #9337

To be implemented when we require Erlang 23 exclusively:

  * [x] Provide `Path.safe_relative_path/1/2` on top of `filelib:safe_relative_path/2`
  * [x] Add `Node.spawn_monitor`
  * [x] Support new rules in map and binary matching ([EEP 52](https://github.com/erlang/otp/pull/2521))
  * [x] Rewrite `Atom.to_string/1` to `:erlang.atom_to_binary/1` (and others)

lukaszsamson:
Wouldn't `:code.all_available/0` make sense in iex autocomplete? https://github.com/elixir-lang/elixir/blob/660ac71d0eccbe9c7424e708e36493d6cecadc6b/lib/iex/lib/iex/autocomplete.ex#L429

josevalim:
@lukaszsamson not really because it would load from applications you don't depend on. It could be good for discovery but it can end-up bringing all of OTP, which perhaps is a bit too much.

michallepicki:
Can some of those items be worked on now? (will Elixir 1.14 require OTP 23?)

josevalim:
Yes, 1.14 will require Erlang/OTP 23! So we can work on those. I will push a commit that removes Erlang/OTP 22 in a few minutes. PRs welcome!

josevalim:
With PR #11558, this will all be tackled. Thanks everyone!

hkrutzer:
Will send_request added in Erlang/OTP 23 be added to GenServer?

josevalim:
No directly. My plan is to add async_call or task_call which is same functionality but using tasks.

PR #10091, opened 2020-06-10
Change release default mode to interactive

### LINKED ISSUE #9337: Starting releases is slow
### Environment

- elixir 1.9.1-otp-21
- OS: Linux (ubuntu and archlinux)

### Current behavior

Starting an application build with the build in `mix release` from `elixir 1.9.1` is way slower than starting the same application build with `distillery`.

Unfortunately, I don't have a public test case, but I did create some measurements. Every time I measured the time to entry in `Application.start`:

```
## Distillery (elixir 1.8.2-otp-21)
- date && _build/prod/rel/app/bin/app foreground: < 3 seconds
- date && _build/prod/rel/app/bin/app console: < 3 seconds

## Distillery (elixir 1.9.1-otp-21)
- date && _build/prod/rel/app/bin/app foreground: < 3 seconds
- date && _build/prod/rel/app/bin/app console: < 3 seconds

## Elixir releases (elixir 1.9.1-otp-21)
- date && _build/prod/rel/app/bin/app start: 10 seconds
- date && _build/prod/rel/app/bin/app start_iex: 10 seconds

## env MIX_ENV=prod mix phx.server --preload-modules (elixir 1.9.1-otp-21)
- date && env MIX_ENV=prod mix phx.server --preload-modules: < 3 seconds

## Experiment: Elixir releases without runtime configuration (elixir 1.9.1-otp-21)
- date && _build/prod/rel/app/bin/app start: < 3 seconds
```

As you can see, using regular releases with runtime configuration are way slower than distillery. I have the feeling that even without runtime configuration they are still slower, but that's hard to measure. There is a post on elixirforum where others are reporting the same issue (https://elixirforum.com/t/elixir-1-9-release-slow-boot-time/) and pointing out that not including runtime configuration did not fix this for them (although it improved a bit)

Having about 10 seconds for a fairly simple application to before it starts is a huge overhead. It would be nice if we had similar startup performance as with `distillery`.

gordalina:
Closes #9337.
Helps #9603.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -98,10 +98,6 @@ test_windows_task:
   matrix:
     - env:
         OS_VERSION: 2019
-        OTP_RELEASE: 23.0
-
-    - env:
-        OS_VERSION: 2019
         OTP_RELEASE: 22.0
 
     - env:
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -27,6 +27,8 @@ test_linux_task:
         - env:
             CHECK_POSIX_COMPLIANT: true
             CHECK_REPRODUCIBLE: true
+            OTP_RELEASE: OTP-23.0
+        - env:
             OTP_RELEASE: OTP-22.3
         - env:
             OTP_RELEASE: OTP-22.0
@@ -94,6 +96,10 @@ test_windows_task:
   alias: Windows Stable
 
   matrix:
+    - env:
+        OS_VERSION: 2019
+        OTP_RELEASE: 23.0
+
     - env:
         OS_VERSION: 2019
         OTP_RELEASE: 22.0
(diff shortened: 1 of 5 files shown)


gordalina:
Sounds good, I'll update the PR.

Is there a better way to check the OTP version other than calling ```elixir --eval "IO.puts(System.otp_release())"``` from the boot script?

josevalim:
@gordalina oh, instead of checking the version on the boot script itself, check it when generating the boot file. Something like:

```eex
-mode <%= maybe_interactive(@release, "$RELEASE_MODE") %>
```

where

```elixir
defp maybe_interactice(release, other) do
  if ...otp >= 23 and Keyword.get(release.options, :reboot_system_after_config, true) do
    "interactive"
  else
    other
  end
end
```

gordalina:
@josevalim Wouldn't that check the OTP version at release time rather than runtime? Or am I missing something here?

josevalim:
It is at release time but the release includes the OTP version inside itself anyway. :)

gordalina:
That makes sense! I'll update the PR.

josevalim:
@gordalina absolutely beautiful. Just one test failure and we should be good to go!

gordalina:
@josevalim this is ready for review. I've squashed the commits.

Issue #10095, opened 2020-06-11
warning reports a misleading message on unused default arguments


DaTrader:
### Environment

* Elixir & Erlang/OTP versions (elixir --version): 1.10.3 / OTP 22
* Operating system: Linux Mint

### Current behavior

A private function taking n arguments, say 5 with the last two set to default values, when compiled without another function in the module ever invoking it with less than 4 arguments, gives a misleading warning message:
warning: the first 4 default arguments in foo/5 are never used
ex:
```
defp foo( arg1, arg2, arg3, arg4 \\ nil, arg5 \\ nil) do
..
end
```
### Expected behavior

The warning message should read something like:
warning: the default value for argument no. 4 in foo/5 is never used

josevalim:
Closing in favor of PR.

xxdavid:
Hi! I'm following up on this issue. The messages were indeed reworded in #10097 by @kelvinst but I am still not convinced that they are correct.

Given this module
```elixir
defmodule FooBar do
  defp foo(arg1, arg2, arg3, arg4 \\ nil, arg5 \\ nil) do
    {arg1, arg2, arg3, arg4, arg5}
  end

  def bar do
    foo(:a, :b, :c, :d)
  end
end
```
the Elixir compiler (1.15.0) reports the following:
```
warning: the default values for the first 4 optional arguments in foo/5 are never used
  foo.ex:2: FooBar (module)
```

But the message doesn't make much sense to me… It talks about *first 4 optional arguments* of `foo/5` but the function has only 2 optional arguments! I think it wants to convey the idea that the default value of the 4th argument wasn't used (also the implementation of `Module.LocalsTracker` testifies this if I understand it correctly). Something like @DaTrader proposed (*the default value for argument no. 4 in foo/5 is never used*) or for example *the default value for the 4th argument in foo/5 is never used* would be more correct IMHO. What do you think?

PR #10097, opened 2020-06-11
Rewording the default arguments warnings

kelvinst:
Refs #10095 

I took the liberty of also rewording the other warnings about arguments with default values, so they all look alike.

### DIFF
--- a/lib/elixir/src/elixir_locals.erl
+++ b/lib/elixir/src/elixir_locals.erl
@@ -121,13 +121,13 @@ format_error({function_conflict, {Receiver, {Name, Arity}}}) ->
     [elixir_aliases:inspect(Receiver), Name, Arity]);
 
 format_error({unused_args, {Name, Arity}}) ->
-  io_lib:format("default arguments in ~ts/~B are never used", [Name, Arity]);
+  io_lib:format("default values for the arguments in ~ts/~B are never used", [Name, Arity]);
 
 format_error({unused_args, {Name, Arity}, 1}) ->
-  io_lib:format("the first default argument in ~ts/~B is never used", [Name, Arity]);
+  io_lib:format("the default value for the first argument in ~ts/~B is never used", [Name, Arity]);
 
 format_error({unused_args, {Name, Arity}, Count}) ->
-  io_lib:format("the first ~B default arguments in ~ts/~B are never used", [Count, Name, Arity]);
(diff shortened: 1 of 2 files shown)


lib/elixir/src/elixir_locals.erl
```diff
@@ -121,13 +121,13 @@ format_error({function_conflict, {Receiver, {Name, Arity}}}) ->
     [elixir_aliases:inspect(Receiver), Name, Arity]);
 
 format_error({unused_args, {Name, Arity}}) ->
-  io_lib:format("default arguments in ~ts/~B are never used", [Name, Arity]);
+  io_lib:format("default values for the arguments in ~ts/~B are never used", [Name, Arity]);
 
 format_error({unused_args, {Name, Arity}, 1}) ->
-  io_lib:format("the first default argument in ~ts/~B is never used", [Name, Arity]);
+  io_lib:format("the default value for the first argument in ~ts/~B is never used", [Name, Arity]);
```
josevalim, line 127:
```suggestion
  io_lib:format("the default value for the first optional argument in ~ts/~B is never used", [Name, Arity]);
```

?

lib/elixir/src/elixir_locals.erl
```diff
@@ -121,13 +121,13 @@ format_error({function_conflict, {Receiver, {Name, Arity}}}) ->
...
 
 format_error({unused_args, {Name, Arity}, Count}) ->
-  io_lib:format("the first ~B default arguments in ~ts/~B are never used", [Count, Name, Arity]);
+  io_lib:format("the default values for the first ~B arguments in ~ts/~B are never used", [Count, Name, Arity]);
```
josevalim, line 130:
```suggestion
  io_lib:format("the default values for the first ~B optional arguments in ~ts/~B are never used", [Count, Name, Arity]);
```

?

lib/elixir/src/elixir_locals.erl
kelvinst: Done

lib/elixir/src/elixir_locals.erl
kelvinst: Done

josevalim:
@kelvinst can you please run the formatter? Then this should be good to go!

kelvinst:
> @kelvinst can you please run the formatter? Then this should be good to go!

Done!

josevalim:

PR #10100, opened 2020-06-15
Update .cirrus.yml to latest Erlang/OTP versions

eksperimental:


josevalim:
Those are supposed to be patch releases, I don’t believe we need to update the latest all the time. Thanks!

josevalim:
Closing this for the reasons above. :)

PR #10101, opened 2020-06-15
Update .cirrus.yml to latest Erlang/OTP versions and changes from master

eksperimental:
The most important change is that since v1.10.3 supports Erlang/OTP 23, it is added to the CI


josevalim:

PR #10102, opened 2020-06-15
Update copyright

eksperimental:
I'm  assuming the rights are currently owned by Dashbit. Please correct me if I'm wrong.


josevalim:
The legal process is under way but it has not been concluded yet. We will update the entries accordingly once it is done. Thanks.

PR #10105, opened 2020-06-17
Use Erlang/OTP "VERSION", instead of OTP "VERSION"

eksperimental:
/cc @whatyouhide


josevalim:

PR #10111, opened 2020-06-19
include callers in translated Logger metadata for Task

mitchellhenke:
Discussed with @josevalim a bit, and this is to be able to get callers from Logger metadata

Things I'm not sure on: Whether new test should be written, or if there are other modules (GenServer?) that could also do this.

### DIFF
--- a/lib/elixir/lib/task/supervised.ex
+++ b/lib/elixir/lib/task/supervised.ex
@@ -112,7 +112,8 @@ defmodule Task.Supervised do
           %{
             domain: [:otp, :elixir],
             error_logger: %{tag: :error_msg},
-            report_cb: &__MODULE__.format_report/1
+            report_cb: &__MODULE__.format_report/1,
+            callers: Process.get(:"$callers")
           }
         )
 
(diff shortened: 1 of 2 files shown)


lib/elixir/lib/task/supervised.ex
```diff
@@ -112,7 +112,8 @@ defmodule Task.Supervised do
           %{
             domain: [:otp, :elixir],
             error_logger: %{tag: :error_msg},
-            report_cb: &__MODULE__.format_report/1
+            report_cb: &__MODULE__.format_report/1,
+            callers: Process.get(:"$callers")
```
hauleth, line 116: I am not sure if this is public API that we can use there.

josevalim:
This looks great! There are some other tests in this file that say “translates Task”, we should add a similar assertion to them too if possible. Also please rename the test that you currently changed and the one below it to say “:proc_lib+Task”, since we are addi Task based assertions to them. :) thanks!

mitchellhenke:
@josevalim updated!

josevalim:

PR #10112, opened 2020-06-19
Correct doc, add spec and update TODO comment in System.stacktrace/0

eksperimental:
the documentation fix corrects the use of "from" and the ~plug~ plus suffix.


josevalim:

PR #10121, opened 2020-06-22
Add --repo to archive.install and escript.install

### LINKED ISSUE #10120: Support --repo flag on archive.install hex / escript.install hex
/cc @wojtekmach

wojtekmach:
Closes #10120

### DIFF
--- a/lib/mix/lib/mix/local/installer.ex
+++ b/lib/mix/lib/mix/local/installer.ex
@@ -233,7 +233,7 @@ defmodule Mix.Local.Installer do
 
     dep_opts =
       opts
-      |> Keyword.take([:organization])
+      |> Keyword.take([:organization, :repo])
       |> Keyword.put(:hex, String.to_atom(package_name))
 
     {:fetcher, {String.to_atom(app_name), version, dep_opts}}
--- a/lib/mix/lib/mix/tasks/archive.install.ex
+++ b/lib/mix/lib/mix/tasks/archive.install.ex
@@ -51,7 +51,10 @@ defmodule Mix.Tasks.Archive.Install do
       from Git, GitHub, or Hex
 
     * `--organization` - specifies an organization to use if fetching the package
-      from a private Hex repository
+      from a private Hex repository. Setting `--organization acme` has the same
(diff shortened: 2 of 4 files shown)


josevalim:

wojtekmach:
There's a bug that I'll address soon and that is organization and repo can be composed together:

    $ mix archive.install hex foo --repo bar --organization baz
    Failed to fetch record for 'hexpm:baz/foo' from registry (using cache instead)

This should have tried fetching from `bar:baz/foo`.

I'll also take another stab at the docs to make it more obvious, looking back at my changes they perhaps give too much emphasis on organisations and as if they (incorrectly) overwrite the repo.

wojtekmach:
I'm sorry, I'm afraid this is wrong still :(

For this dep:

    {:hexpm_prod_foo_private, "~> 0.1.0", repo: "acme", organization: "hex"}

we end up with this lock:

    % mix deps
    * hexpm_prod_foo_private (Hex package) (mix)
      locked at 0.1.0 (hexpm:hex/hexpm_prod_foo_private) 6c5b9142

and so we ended up with the `repo: "hexpm/hex"` and our `repo: "acme"` was ignored, we've silently disregarded one of the supplied values.

In https://github.com/elixir-lang/elixir/pull/10134 we made it so you can use `repo` along with `organization`, and so we have inconsistent behaviour between `mix deps` and `mix archive.install` that we should fix.

Looking back I think allowing to compose `--repo` with `--organization` and effectively overwriting the repo that you supplied was a mistake - I think it's overloading the term repo too much. But we still need to somehow refer to organisations within external repositories, if the only way is to do `--repo acme:private` and we shouldn't use `--organization` for that, it means `--organization` would only be useful for hexpm, which would be a mistake. I think I have a clean solution for this that I'll mention soon.

wojtekmach:
Spoken with @josevalim and we'd move forward with allowing `--repo acme --organization foo` (and it meaning the same thing as `--repo acme:foo`). Thus the `mix *.install` tasks are good and we just need to change `mix deps.get` which will most likely be a change in Hex.

Issue #10122, opened 2020-06-24
Provide more granularity for the strip_beams release configuration


noozo:
Currently `mix release` provides the boolean flag `strip_beams` to control whether or not beam files are stripped of metadata like documentation, for example.

It would be good to be able to specify exactly *what* to strip and what to keep.

One example of this necessity can be seen here: https://github.com/open-api-spex/open_api_spex/issues/242
tldr> OpenApiSpex relies on code documentation for generating OpenAPI documentation at runtime, which means it would be nice to tell mix release not to strip the beam files out of the docs only, hence keeping the final compiled files a bit smaller.

For instance, in my own project, using strip_beams flag right now increases my final docker container size from 71MB to 96MB. I'm guessing if we could keep docs alone this size would be much smaller.

josevalim:
Thank you! We will definitely ship this on v1.11. A PR is welcome!

noozo:
@josevalim, working on a PR for this. My approach is to allow another option to be passed (possibly named strip_beams_exceptions), a list that can contain chunks not to be stripped when the strip_beams option is true. You would pass, for instance [:docs, :debug_info] and everything would be stripped as usual, except those 2 chunks (besides the already @significant_chunks).

Does that sound reasonable, or would another way be preferable?

Another couple of questions, while i'm at it:
- Should the same ability be added to Mix.Tasks.Escript.Build?
- When opening a PR is it better to wait until everything is working, or can WIP open PR's exist?

Thanks!

josevalim:
I believe we can do: strip_beams: [keep: “Docs”]. We should support it on Build too. Open the PR whenever you would like to get feedback. :)

noozo:
Yeah, that sounds better. Cool, will do!

PR #10125, opened 2020-06-25
Adding support for being more granular when stripping beam files

noozo:
This PR attempts to add a bit more granular control over what gets stripped from the BEAM files, when the strip_beams option is set to true in the mix release options.

For now it allows passing a couple of exceptions in a list, namely :docs and :debug_info. The exceptions passed in the options will *not* get stripped from the BEAM files.

For reference: https://github.com/elixir-lang/elixir/issues/10122

Be gentle on me, it's my first one :)

### DIFF
--- a/lib/mix/lib/mix/release.ex
+++ b/lib/mix/lib/mix/release.ex
@@ -757,10 +757,11 @@ defmodule Mix.Release do
   def copy_ebin(release, source, target) do
     with {:ok, [_ | _] = files} <- File.ls(source) do
       File.mkdir_p!(target)
-      strip_beams? = Keyword.get(release.options, :strip_beams, true)
-      chunks_to_keep = release.options
-        |> Keyword.get(:strip_beams_exceptions, [])
-        |> chunks_to_keep()
+
+      {strip_beams?, chunks_to_keep} =
+        release.options
+        |> Keyword.get(:strip_beams, true)
+        |> parse_strip_beams_option()
 
       for file <- files do
         source_file = Path.join(source, file)
@@ -780,6 +781,9 @@ defmodule Mix.Release do
     end
   end
 
+  @spec strip_beam(binary()) :: {:ok, binary} | {:error, :beam_lib, term}
+  def strip_beam(binary), do: strip_beam(binary, [])
+
   @doc """
   Strips a beam file for a release.
 
@@ -790,8 +794,8 @@ defmodule Mix.Release do
   future versions.
   """
   @spec strip_beam(binary(), List.t()) :: {:ok, binary} | {:error, :beam_lib, term}
-  def strip_beam(binary, chunks_to_keep) do
-    case :beam_lib.chunks(binary, chunks_to_keep, [:allow_missing_chunks]) do
+  def strip_beam(binary, exceptions) do
+    case :beam_lib.chunks(binary, @significant_chunks ++ exceptions, [:allow_missing_chunks]) do
       {:ok, {_, chunks}} ->
         chunks = for {name, chunk} <- chunks, is_binary(chunk), do: {name, chunk}
         {:ok, binary} = :beam_lib.build_module(chunks)
@@ -806,13 +810,15 @@ defmodule Mix.Release do
     end
   end
 
-  defp chunks_to_keep(exceptions) do
-    @significant_chunks ++ Enum.map(exceptions, fn ex ->
-      case ex do
-        :docs -> 'Docs'
-        :debug_info -> :debug_info
-        # NOTE: More chunks we might want to allow keeping?
+  @spec parse_strip_beams_option(true | false | Keyword.t()) :: {true | false, List.t() | nil}
+  def parse_strip_beams_option(option) do
+    {strip?, exceptions} =
+      case option do
+        [keep: exceptions] when is_list(exceptions) -> {true, exceptions}
+        true -> {true, []}
+        false -> {false, nil}
       end
-    end)
+
+    {strip?, @significant_chunks ++ exceptions}
   end
 end
--- a/lib/mix/lib/mix/release.ex
+++ b/lib/mix/lib/mix/release.ex
@@ -790,9 +790,11 @@ defmodule Mix.Release do
   The exact chunks that are kept are not documented and may change in
   future versions.
   """
-  @spec strip_beam(binary(), List.t()) :: {:ok, binary} | {:error, :beam_lib, term}
-  def strip_beam(binary, options \\ [keep: []]) when is_list(options) do
-    case :beam_lib.chunks(binary, Enum.uniq(@significant_chunks ++ options[:keep]), [
+  @spec strip_beam(binary(), keyword()) :: {:ok, binary()} | {:error, :beam_lib, term()}
+  def strip_beam(binary, options \\ []) when is_list(options) do
+    chunks_to_keep = Keyword.get(options, :keep, [])
+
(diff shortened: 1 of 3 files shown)


lib/mix/lib/mix/release.ex
```diff
@@ -786,9 +793,9 @@ defmodule Mix.Release do
   The exact chunks that are kept are not documented and may change in
   future versions.
   """
-  @spec strip_beam(binary()) :: {:ok, binary} | {:error, :beam_lib, term}
-  def strip_beam(binary) do
-    case :beam_lib.chunks(binary, @significant_chunks, [:allow_missing_chunks]) do
+  @spec strip_beam(binary(), List.t()) :: {:ok, binary} | {:error, :beam_lib, term}
+  def strip_beam(binary, exceptions) do
```
fertapric, line 797: I think we could use here the syntax for default arguments:

```suggestion
  def strip_beam(binary, exceptions \\ []) do
```

and maybe include some guards?

```suggestion
  def strip_beam(binary, exceptions \\ []) when is_list(exceptions) do
```

lib/mix/lib/mix/release.ex
```diff
@@ -802,4 +809,16 @@ defmodule Mix.Release do
         error
     end
   end
+
+  @spec parse_strip_beams_option(true | false | Keyword.t()) :: {true | false, List.t() | nil}
+  def parse_strip_beams_option(option) do
```
fertapric, line 814: I think this is a private function 🙂 

```suggestion
  defp parse_strip_beams_option(option) do
```

lib/mix/lib/mix/release.ex
```diff
@@ -777,6 +781,9 @@ defmodule Mix.Release do
     end
   end
 
+  @spec strip_beam(binary()) :: {:ok, binary} | {:error, :beam_lib, term}
+  def strip_beam(binary), do: strip_beam(binary, [])
```
josevalim, line 785: We can skip this clause. Instead, pass a default to strip_beam/2 below, as suggested by @fertapric!

lib/mix/lib/mix/release.ex
```diff
@@ -786,9 +793,9 @@ defmodule Mix.Release do
   The exact chunks that are kept are not documented and may change in
   future versions.
   """
-  @spec strip_beam(binary()) :: {:ok, binary} | {:error, :beam_lib, term}
-  def strip_beam(binary) do
-    case :beam_lib.chunks(binary, @significant_chunks, [:allow_missing_chunks]) do
+  @spec strip_beam(binary(), List.t()) :: {:ok, binary} | {:error, :beam_lib, term}
+  def strip_beam(binary, exceptions) do
+    case :beam_lib.chunks(binary, @significant_chunks ++ exceptions, [:allow_missing_chunks]) do
```
josevalim, line 798: We probably want to do:

```suggestion
    case :beam_lib.chunks(binary, Enum.uniq(@significant_chunks ++ exceptions), [:allow_missing_chunks]) do
```

lib/mix/lib/mix/tasks/escript.build.ex
```diff
@@ -159,10 +161,12 @@ defmodule Mix.Tasks.Escript.Build do
...
     #
     # TODO: Deprecate :strip_beam option on v1.13
-    strip_beams? =
-      Keyword.get_lazy(escript_opts, :strip_beams, fn ->
+    {strip_beams?, chunks_to_keep} =
+      escript_opts
+      |> Keyword.get_lazy(:strip_beams, fn ->
         Keyword.get(escript_opts, :strip_beam, true)
       end)
+      |> Mix.Release.parse_strip_beams_option()
```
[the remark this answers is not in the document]
noozo: Yeah i agree, it also felt a bit weird to me, tbh :)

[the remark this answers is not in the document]
noozo: @fertapric The only part i feel bad about is that, in that case,  i probably have to move the parsing of the option to inside `Mix.Release.parse_strip_beams_option`, which will effectively run it for each BEAM :(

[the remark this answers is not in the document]
fertapric: @noozo I see 🤔   Then maybe we can keep `parse_strip_beams_option` public, but mark it with `@doc false` though.

Or maybe duplicate that function in both modules since the logic is quite simple. Both sound good to me.

lib/mix/lib/mix/release.ex
```diff
@@ -802,4 +809,16 @@ defmodule Mix.Release do
...
+  @spec parse_strip_beams_option(true | false | Keyword.t()) :: {true | false, List.t() | nil}
+  def parse_strip_beams_option(option) do
+    {strip?, exceptions} =
+      case option do
+        [keep: exceptions] when is_list(exceptions) -> {true, exceptions}
+        true -> {true, []}
+        false -> {false, nil}
+      end
+
+    {strip?, @significant_chunks ++ exceptions}
```
fertapric, line 822: I think we are adding the `@significant_chunks` twice, here and in `strip_beam/1,2`

lib/mix/lib/mix/release.ex
```diff
@@ -786,9 +790,11 @@ defmodule Mix.Release do
   The exact chunks that are kept are not documented and may change in
   future versions.
   """
-  @spec strip_beam(binary()) :: {:ok, binary} | {:error, :beam_lib, term}
-  def strip_beam(binary) do
-    case :beam_lib.chunks(binary, @significant_chunks, [:allow_missing_chunks]) do
+  @spec strip_beam(binary(), List.t()) :: {:ok, binary} | {:error, :beam_lib, term}
+  def strip_beam(binary, options \\ [keep: []]) when is_list(options) do
+    case :beam_lib.chunks(binary, Enum.uniq(@significant_chunks ++ options[:keep]), [
```
fertapric, line 795: I would default to `[]` (that's a common default value when passing `options`) and get the info from `options` assuming the `:keep` key might not be there. In short words, this should work `strip_beam(beam, [])` 🙂 


Ah! And I would also change the `List.t()` in the spec to `keyword()` here and in `parse_strip_beams_option`

noozo: I thought you meant the opposite from your first comment, when you said that calling strip_beam like this felt strange: `Mix.Release.strip_beam(beam, ['Docs', 'Dbgi'])` :)

fertapric: Oh! the `[]` in `strip_beam(beam, [])` is the `options` not the list of chunks to keep. I think some code would explain it better:

```suggestion
  @spec strip_beam(binary(), keyword) :: {:ok, binary} | {:error, :beam_lib, term}
  def strip_beam(binary, options \\ []) when is_list(options) do
    chunks_to_keep = Keyword.get(options, :keep, [])
    case :beam_lib.chunks(binary, Enum.uniq(@significant_chunks ++ chunks_to_keep), [
```

Sorry for the misunderstanding 😅

noozo: Oh, you mean the list still having the keep inside, but omitting it from the signature :)

noozo: LOL i wrote the exact same code you posted :)

noozo:
@fertapric Who usually merges? The author, one of the reviewers, Jose?

josevalim:
@noozo we use two reviewers in all code-related PRs.

josevalim:

PR #10132, opened 2020-06-28
Normalize paths before expanding them

### LINKED ISSUE #9896: elixirc does not compile modules from files
I have a folder with sub-folders containing .ex/.exs files with Elixir modules. All I need to do is go through all the folders and simply compile the modules. 
I have a small PowerShell script to do that: 
```
$files = get-childitem . -Recurse -Filter *.ex*
foreach ($file in $files)
{	
    elixirc $file.Fullname          # tried $file.Name as well, doesn't work either
}
```
For some reason that does not do the job.
Passing neither `$file.Fullname` nor `$file.Name` works. 
Very confusing, because it's as simplistic of a task as it can get.
### Environment

* Elixir & Erlang/OTP versions (elixir --version): Erlang/OTP 21 [erts-10.3] [64-bit] [smp:12:12] [ds:12:12:10] [async-threads:1]
* Operating system: Windows 10, v. 10.0.18363 Build 18363 

### Current behavior
For each {filename_here}.ex/.exs file matched, I get a message: 
`No files matched pattern(s) {filename_here}.ex`

### Expected behavior
For each file, a module should be compiled.

josevalim:
Closes #9896

### DIFF
--- a/lib/elixir/lib/kernel/cli.ex
+++ b/lib/elixir/lib/kernel/cli.ex
@@ -528,6 +528,7 @@ defmodule Kernel.CLI do
 
   defp filter_patterns(pattern) do
     pattern
+    |> Path.expand()
     |> Path.wildcard()
     |> :lists.usort()
     |> Enum.filter(&File.regular?/1)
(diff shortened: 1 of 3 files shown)


josevalim:

PR #10135, opened 2020-06-29
Add :tmp_dir tag to ExUnit

wojtekmach:
https://groups.google.com/u/1/g/elixir-lang-core/c/GfwqeDGqVoA/m/YR-NLrAzBgAJ

As part of this PR, I've also started using the new feature in a few tests.
There is a lot more places where this could be used and I thought about
addressing that in future PRs, but let me know otherwise!

### DIFF
--- a/lib/elixir/test/elixir/kernel/cli_test.exs
+++ b/lib/elixir/test/elixir/kernel/cli_test.exs
@@ -107,27 +107,25 @@ end
 defmodule Kernel.CLI.CompileTest do
   use ExUnit.Case, async: true
 
+  @moduletag :tmp_dir
+
   setup context do
-    # Set up a per-test temporary directory, so we can run these with async: true.
-    # We use the test's line number as the directory name, so they won't conflict.
-    tmp_dir_path = tmp_path("beams/#{context[:line]}")
-    beam_file_path = Path.join([tmp_dir_path, "Elixir.CompileSample.beam"])
+    beam_file_path = Path.join([context.tmp_dir, "Elixir.CompileSample.beam"])
     fixture = fixture_path("compile_sample.ex")
-    File.mkdir_p!(tmp_dir_path)
-    {:ok, [tmp_dir_path: tmp_dir_path, beam_file_path: beam_file_path, fixture: fixture]}
+    {:ok, [beam_file_path: beam_file_path, fixture: fixture]}
   end
 
   test "compiles code", context do
-    assert elixirc('#{context[:fixture]} -o #{context[:tmp_dir_path]}') == ""
-    assert File.regular?(context[:beam_file_path])
+    assert elixirc('#{context.fixture} -o \'#{context.tmp_dir}\'') == ""
+    assert File.regular?(context.beam_file_path)
 
     # Assert that the module is loaded into memory with the proper destination for the BEAM file.
-    Code.append_path(context[:tmp_dir_path])
-    assert :code.which(CompileSample) |> List.to_string() == Path.expand(context[:beam_file_path])
+    Code.append_path(context.tmp_dir)
+    assert :code.which(CompileSample) |> List.to_string() == Path.expand(context.beam_file_path)
   after
     :code.purge(CompileSample)
     :code.delete(CompileSample)
-    Code.delete_path(context[:tmp_dir_path])
+    Code.delete_path(context.tmp_dir)
   end
 
   @tag :windows
@@ -143,34 +141,34 @@ defmodule Kernel.CLI.CompileTest do
   after
     :code.purge(CompileSample)
     :code.delete(CompileSample)
-    Code.delete_path(context[:tmp_dir_path])
+    Code.delete_path(context.tmp_dir)
   end
 
   test "fails on missing patterns", context do
-    output = elixirc('#{context[:fixture]} non_existing.ex -o #{context[:tmp_dir_path]}')
+    output = elixirc('#{context.fixture} non_existing.ex -o \'#{context.tmp_dir}\'')
     assert output =~ "non_existing.ex"
     refute output =~ "compile_sample.ex"
-    refute File.exists?(context[:beam_file_path])
+    refute File.exists?(context.beam_file_path)
   end
 
   test "fails on missing write access to .beam file", context do
-    compilation_args = '#{context[:fixture]} -o #{context[:tmp_dir_path]}'
+    compilation_args = '#{context.fixture} -o \'#{context.tmp_dir}\''
 
     assert elixirc(compilation_args) == ""
-    assert File.regular?(context[:beam_file_path])
+    assert File.regular?(context.beam_file_path)
 
     # Set the .beam file to read-only
-    File.chmod!(context[:beam_file_path], 4)
-    {:ok, %{access: access}} = File.stat(context[:beam_file_path])
+    File.chmod!(context.beam_file_path, 4)
+    {:ok, %{access: access}} = File.stat(context.beam_file_path)
 
     # Can only assert when read-only applies to the user
     if access != :read_write do
       output = elixirc(compilation_args)
 
       expected =
-        "(File.Error) could not write to file #{inspect(context[:beam_file_path])}: permission denied"
+        "(File.Error) could not write to file #{inspect(context.beam_file_path)}: permission denied"
 
-      assert String.contains?(output, expected)
+      assert output =~ expected
     end
   end
 end
--- a/lib/elixir/test/elixir/kernel/cli_test.exs
+++ b/lib/elixir/test/elixir/kernel/cli_test.exs
@@ -130,9 +130,9 @@ defmodule Kernel.CLI.CompileTest do
 
   @tag :windows
   test "compiles code with Windows paths", context do
-    fixture = String.replace(context[:fixture], "/", "\\")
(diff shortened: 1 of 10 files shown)


wojtekmach:
Forgot to mention, while not part of this PR, a potential future enhancement is a similar tag `:cd_tmp_dir` which would the exact same thing, plus:

1. `File.cd!` into the temp directory on setup
2. `File.cd!` back on exit
3. crash with an error message when used with `async: true` (since File.cd/1 isn't concurrency friendly)

lib/elixir/test/elixir/kernel/cli_test.exs
```diff
@@ -107,27 +107,25 @@ end
...
+    Code.append_path(context.tmp_dir)
+    assert :code.which(CompileSample) |> List.to_string() == Path.expand(context.beam_file_path)
   after
     :code.purge(CompileSample)
     :code.delete(CompileSample)
-    Code.delete_path(context[:tmp_dir_path])
+    Code.delete_path(context.tmp_dir)
   end
 
   @tag :windows
```
josevalim, line 131: Please change this test too. :) We was just added.

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -259,6 +260,39 @@ defmodule ExUnit.Runner do
...
+    File.mkdir_p!(path)
+    Map.put(context, :tmp_dir, path)
+  end
+
+  defp escape_path(path) do
+    case :os.type() do
+      {:win32, _} -> String.replace(path, ~R'[~#%&*{}\\:<>?/+|"]', "_")
+      _ -> path
+    end
+    |> String.replace("/", "_")
```
josevalim, line 293: I would always run the escape characters. I also think the escape characters can be written like this:

```
@escape Enum.map('[~#%&*{}\\:<>?/+|"]', <<&1::utf8>>)
```

lib/ex_unit/lib/ex_unit/case.ex
```diff
@@ -209,6 +219,31 @@ defmodule ExUnit.Case do
...
+  ## Tmp Dir
+
+  ExUnit automatically creates a temporary directory for tests tagged with
+  `:tmp_dir` and puts the path to that directory into the test context.
+  The directory is removed before being created to ensure we start with a blank
+  slate.
+
+  The temporary directory path is unique (includes the test module and test name)
+  and thus appropriate for running tests concurrently. You can customize it by
+  setting the tag to a string, e.g.: `tmp_dir: "my_path"`.
```
josevalim, line 231: It seems this option actually changes the path name but the name is still nested inside "module/test". Is this expected? I was thinking that, if you pass your own `tmp_dir`, then we should give it to the test as is... but I don't have a good use case for that. Maybe we should keep it simple and allow it to only be booleans for  now?

lib/ex_unit/lib/ex_unit/runner.ex
josevalim: Then you can do `String.replace(test, @escape, "_")`.

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -259,6 +260,39 @@ defmodule ExUnit.Runner do
...
+  end
+
+  defp maybe_create_tmp_dir(_, context, _) do
+    context
+  end
+
+  defp create_tmp_dir!(test, extra_path, context) do
+    module = escape_path(inspect(test.module))
+    name = escape_path(to_string(test.name))
+    path = Path.join(["tmp", module, name, extra_path])
```
josevalim, line 282: Should we include `tmp/` in the .gitignore of new files generated by `mix new`?

wojtekmach: Good call

lib/mix/test/mix/cli_test.exs
```diff
@@ -169,8 +171,9 @@ defmodule Mix.CLITest do
     System.delete_env("MIX_EXS")
   end
 
-  test "new with tests and cover" do
-    in_tmp("new_with_tests", fn ->
+  @tag tmp_dir: "new_with_tests"
```
wojtekmach, line 174: this is the test where customising the path was useful, by having the final directory as we want, we can simply do: `mix new .` below.

lib/ex_unit/lib/ex_unit/case.ex
wojtekmach: Yes, I made it like this on purpose, I think keeping the test under tmp/:module/:name keeps things more tidy but yeah, it isn't obvious. My thinking was the actual path isn't that important, just the final directory is. I pointed out an existing test where setting custom path is useful, see the other comment.

lib/ex_unit/lib/ex_unit/runner.ex
wojtekmach: Good call. I decided to still escape things individually, that is we have: `tmp/MyTest/test my test`, as opposed to: `tmp/MyTest_test my test` (`/` would have been escaped)

lib/mix/test/mix/cli_test.exs
josevalim: Oh, I see. I understand it then. Let's keep it as is!

lib/ex_unit/lib/ex_unit/case.ex
josevalim: Good call. Let's just document it then this configures the directory inside the module/test folder.

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -259,6 +260,39 @@ defmodule ExUnit.Runner do
...
 
+  defp maybe_create_tmp_dir(test, context, %{tmp_dir: true}) do
+    create_tmp_dir!(test, "", context)
+  end
+
+  defp maybe_create_tmp_dir(test, context, %{tmp_dir: path}) when is_binary(path) do
+    create_tmp_dir!(test, path, context)
+  end
+
+  defp maybe_create_tmp_dir(_, _, %{tmp_dir: other}) do
```
fertapric, line 271: Should it maybe accept `tmp_dir: false` too just in case someone wants to override a `@moduletag`?

wojtekmach: Good call, updated!

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -259,6 +260,37 @@ defmodule ExUnit.Runner do
...
+  end
+
+  defp maybe_create_tmp_dir(_, context, _) do
+    context
+  end
+
+  defp create_tmp_dir!(test, extra_path, context) do
+    module = escape_path(inspect(test.module))
+    name = escape_path(to_string(test.name))
+    path = Path.join(["tmp", module, name, extra_path])
```
fertapric, line 286: I think it could be handy to be able to configure the base directory. For example: `config :ex_unit, root_tmp_dir: System.get_env("TMPDIR", "mytmp")`. Thoughts?

lib/ex_unit/lib/ex_unit/runner.ex
```diff
@@ -259,6 +260,41 @@ defmodule ExUnit.Runner do
...
+  defp maybe_create_tmp_dir(test, context, %{tmp_dir: path}) when is_binary(path) do
+    create_tmp_dir!(test, path, context)
+  end
+
+  defp maybe_create_tmp_dir(_, context, %{tmp_dir: false}) do
+    context
+  end
+
+  defp maybe_create_tmp_dir(_, _, %{tmp_dir: other}) do
+    raise ArgumentError, "expected :tmp_dir to be atom true or a string, got: #{inspect(other)}"
```
fertapric, line 276:
```suggestion
    raise ArgumentError, "expected :tmp_dir to be a boolean or a string, got: #{inspect(other)}"
```

lib/ex_unit/lib/ex_unit/runner.ex
josevalim: I thought the same but I don't have a use case for it yet, so I am glad to sit on it until a need arrives.

fertapric: This might be one: if you run your tests inside a Docker container and you mount the root directory, having the option to write to a directory that is not synced with the local machine might give a little boost in performance 🙂

josevalim: Ah, I see! If someone wants to explore this, it would be welcome. In particular, if we can have some numbers that show it is indeed better, it will help us move it forward. :+1:

josevalim:

lib/ex_unit/lib/ex_unit/case.ex
```diff
@@ -141,20 +141,30 @@ defmodule ExUnit.Case do
...
   The following tags customize how tests behave:
 
     * `:capture_log` - see the "Log Capture" section below
+
     * `:skip` - skips the test with the given reason
+
     * `:timeout` - customizes the test timeout in milliseconds (defaults to 60000).
       Accepts `:infinity` as a timeout value.
 
+    * `:tmp_dir` - (since v1.11.0) see the "Tmp Dir" section below
```
eksperimental, line 166: I would advice not mention the `@since` for attributes, as more features are added to the language, it will become noise in the docs.

Issue #10114, opened 2020-06-20
CompileError defining a guard using a list with >1023 elements


amalbuquerque:
### Environment

* Elixir & Erlang/OTP versions (elixir --version): Elixir 1.10.3 with Erlang/OTP 22

```
Erlang/OTP 22 [erts-10.4.4] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-thread
s:1] [hipe]

Elixir 1.10.3 (compiled with Erlang/OTP 22)
```

* Operating system: Linux Mint 19.1 Tessa

### Current behavior

If I define a guard clause using a list with more than 1023 elements, it fails to compile with a `CompileError`. I think it's related with the guard clause, since using a similar list to define a function works (`def xyz(value) when value in @foo1024, do: "hello #{value}"`).

Code to reproduce the error:

```
defmodule Abc do
  @foo1023 1..1023 |> Enum.map(&"foo#{&1}")
  @foo1024 1..1024 |> Enum.map(&"foo#{&1}")

  # defguard is_foo?(value) when value in @foo1023 # works

  defguard is_foo?(value) when value in @foo1024 # CompileError
end
```

Error:

```
❯ elixirc defguard_error.ex

== Compilation error in file defguard_error.ex ==
** (CompileError) Elixir.Abc: function 'MACRO-is_foo?'/2+1047:
  An implementation limit was reached.
  Try reducing the complexity of this function.

  Instruction: {put_list,{y,0},{literal,[<<"foo1024">>]},{x,1023}}

    (stdlib 3.9.2) lists.erl:1338: :lists.foreach/2
    (elixir 1.10.3) src/elixir_erl_compiler.erl:12: anonymous fn/3 in :elixir_erl
_compiler.spawn/2
```

### Expected behavior

To be able to use an arbitrary-size list in my custom guard definitions.

If not possible in fact to have a `defguard` like this, I'd expect a more user-friendly error message.

Thanks in advance :pray: :purple_heart:

eksperimental:
@josevalim regarding the commit, 
there list can be split into two in's such as in this example.

```
defmodule TooLargeList do
  @list1 1..1024 |> Enum.to_list
  @list2 1..1000 |> Enum.to_list
  @list3 1001..2001  |> Enum.to_list

  def value(value) when value in @list2 or value in @list3,
  	do: :ok
  def value(_),
  	do: :error
end
```

[...]

PR #10138, opened 2020-06-30
Deal with lists with more than 1023 elements on the right hand side of in/2

eksperimental:
When dealing with lists with more than 1023 elements,
the VM will crash, so we split the list in chunks of 1023 elements,
and we fold them into one long expression.

Solves #10114

/cc @amalbuquerque


josevalim:
If adding an `or` fixes it, then there is something very wrong with code we are emitting, because we are emitting are just ORs too.

PR #10139, opened 2020-06-30
Do not use backticks in error messages, simplify field/fun printing

josevalim:


eksperimental:
I noticed this style in ExDoc. We should make it consistent.

josevalim:
It is fine if ExDoc used it, as long as it is consistent within itself.

PR #10147, opened 2020-07-03
Show ref information when getting or updating mix dependencies from git sources

noozo:
Adds reference information (tag, branch, commit hash) to the mix deps task when fetching dependencies from git.

Currently when getting or updating libs from git sources:

```
* Updating open_api_spex (https://github.com/open-api-spex/open_api_spex.git)
* Updating stripity_stripe (https://github.com/code-corps/stripity_stripe.git)
* Updating elixir_utils (git@github.com:DroverLtd/elixir_utils.git)
```

With this PR:

```
* Updating open_api_spex (https://github.com/open-api-spex/open_api_spex.git - origin/master)
* Updating stripity_stripe (https://github.com/code-corps/stripity_stripe.git - origin/master)
* Updating elixir_utils (git@github.com:DroverLtd/elixir_utils.git - 1.2.3)
```

The code can probably be much more optimised, suggestions welcomed :)

### DIFF
--- a/lib/mix/lib/mix/scm/git.ex
+++ b/lib/mix/lib/mix/scm/git.ex
@@ -7,12 +7,14 @@ defmodule Mix.SCM.Git do
   end
 
   def format(opts) do
-    {ref_type, ref_value} = cond do
-      opts[:tag] -> {:tag, opts[:tag]}
-      opts[:branch] -> {:branch, opts[:branch]}
-      opts[:ref] -> {:ref, opts[:ref]}
-      true -> {nil, nil}
-    end
+    {ref_type, ref_value} =
+      cond do
+        opts[:tag] -> {:tag, opts[:tag]}
+        opts[:branch] -> {:branch, opts[:branch]}
+        opts[:ref] -> {:ref, opts[:ref]}
+        true -> {nil, nil}
+      end
+
     extra_info = if ref_type, do: " - #{ref_type}: #{ref_value}", else: ""
     "#{opts[:git]}#{extra_info}"
   end
--- a/lib/mix/lib/mix/scm/git.ex
+++ b/lib/mix/lib/mix/scm/git.ex
@@ -7,7 +7,14 @@ defmodule Mix.SCM.Git do
   end
 
   def format(opts) do
-    opts[:git]
+    {ref_type, ref_value} = cond do
+      opts[:tag] -> {:tag, opts[:tag]}
+      opts[:branch] -> {:branch, opts[:branch]}
+      opts[:ref] -> {:ref, opts[:ref]}
+      true -> {nil, nil}
+    end
+    extra_info = if ref_type, do: " - #{ref_type}: #{ref_value}", else: ""
+    "#{opts[:git]}#{extra_info}"
   end
 
   def format_lock(opts) do
--- a/lib/mix/lib/mix/scm/git.ex
+++ b/lib/mix/lib/mix/scm/git.ex
@@ -7,7 +7,7 @@ defmodule Mix.SCM.Git do
   end
 
   def format(opts) do
-    "#{opts[:git]} (#{get_opts_rev(opts)})"
+    "#{opts[:git]} - #{get_opts_rev(opts)}"
   end
 
   def format_lock(opts) do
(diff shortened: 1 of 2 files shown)


lib/mix/lib/mix/scm/git.ex
```diff
@@ -7,7 +7,14 @@ defmodule Mix.SCM.Git do
...
   def format(opts) do
-    opts[:git]
+    {ref_type, ref_value} = cond do
+      opts[:tag] -> {:tag, opts[:tag]}
+      opts[:branch] -> {:branch, opts[:branch]}
+      opts[:ref] -> {:ref, opts[:ref]}
+      true -> {nil, nil}
+    end
+    extra_info = if ref_type, do: " - #{ref_type}: #{ref_value}", else: ""
+    "#{opts[:git]}#{extra_info}"
```
josevalim, line 19: There is already some code related to this here: https://github.com/elixir-lang/elixir/pull/10147/files#diff-dac65d3a4f087adb3405c3e1e64fd076R25

Maybe we can refactor and reuse it. Something like:

    "#{opts[:git]}#{extra_info(opts)}"

josevalim: Also please take a look at the tests. :)

noozo: will take a look, thx

noozo: think you posted a link to this one :)

noozo: @josevalim

josevalim: Oops, you have to click the expand button below to see it. Here is a better link: https://github.com/elixir-lang/elixir/blob/master/lib/mix/lib/mix/scm/git.ex#L18-L22

noozo:
Changed to reuse existing private function that already formatted that information.

Also added tests.

lib/mix/lib/mix/scm/git.ex
```diff
@@ -7,7 +7,7 @@ defmodule Mix.SCM.Git do
   end
 
   def format(opts) do
-    opts[:git]
+    "#{opts[:git]} (#{get_opts_rev(opts)})"
```
josevalim, line 10: Sorry, I didn't meant to say using `get_opts_rev` but rather `format_lock(opts)`. Note it may return `nil` though.

noozo: But isn't this one enough for the purposes? I initially started with extracting the common code from inside format_lock, but then found this one. What is the advantage of using format_lock (note that i don't fully understand that method yet (or the overall system architecture for that matter) :)).

josevalim: Actually, using `format_lock` would lead to duplication on the `mix deps` output, so your approach is better. However, I am not a big fan of the double parens in the output. Maybe we should use this instead?

```suggestion
    "#{opts[:git]} - #{get_opts_rev(opts)}"
```

noozo: Yeah, me neither. Agreed :)

lib/mix/test/mix/scm/git_test.exs
```diff
@@ -22,6 +22,22 @@ defmodule Mix.SCM.GitTest do
...
+             "https://github.com/elixir-lang/some_dep.git - origin/master"
+
+    assert Mix.SCM.Git.format(Keyword.put(opts, :tag, "v1")) ==
+             "https://github.com/elixir-lang/some_dep.git - v1"
+
+    assert Mix.SCM.Git.format(Keyword.put(opts, :branch, "b")) ==
+             "https://github.com/elixir-lang/some_dep.git - origin/b"
+
+    assert Mix.SCM.Git.format(Keyword.put(opts, :ref, "xx")) ==
+             "https://github.com/elixir-lang/some_dep.git - xx"
```
josevalim, line 38:
```suggestion
    assert Mix.SCM.Git.format(Keyword.put(opts, :ref, "abcdef")) ==
             "https://github.com/elixir-lang/some_dep.git - abcdef"
```

josevalim:

Issue #8262, opened 2018-10-09
mix clean --deps does not clean out .erl files generated by yecc/leex


f355:
### Current behavior

When running `mix clean --deps`, the `.erl` files generated from `.xrl`/`.yrl` are left behind.

### Expected behavior

All artefacts and intermediate files for the dependencies are deleted after running `mix clean --deps`, leaving the tree as it looked after `mix deps.get`, before the compilation.

### Steps to reproduce

I'll use Absinthe as an example dependency. I'm using a GitHub tag instead of Hex version, because the artefact pushed to Hex has the `.erl` files in question included for some reason (which might be a separate issue around `mix hex.publish` that's outside of the scope for this bug report).

- Create a new project: `mix new mix_bug && cd mix_bug`
- Edit mix.exs to add `{:absinthe, git: "https://github.com/absinthe-graphql/absinthe.git", tag: "v1.4.13"}` to the list of deps.
- Fetch the dependencies: `mix deps.get`
- Verify there are no `.erl` files:
```bash
$ ls deps/absinthe/src
absinthe_lexer.xrl  absinthe_parser.yrl
```
- Compile and clean the project: `mix compile && mix clean --deps`
- The `.erl` files should be gone after cleaning, but they are not:
```bash
$ ls deps/absinthe/src/
absinthe_lexer.erl  absinthe_lexer.xrl  absinthe_parser.erl absinthe_parser.yrl
```

### Environment

* Elixir & Erlang/OTP versions (elixir --version): 
```bash
$ elixir --version
Erlang/OTP 21 [erts-10.1] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [hipe] [dtrace]

Elixir 1.7.3 (compiled with Erlang/OTP 21)
```
* Mix version:
```bash
$ mix --version
Erlang/OTP 21 [erts-10.1] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [hipe] [dtrace]

Mix 1.7.3 (compiled with Erlang/OTP 21)
```
* Operating system: 
macOS Mojave 10.14

ggcampinho:
I was checking the `clean` task and it uses a simple approach of cleaning the `_build/deps` folder instead of invoking the `deps.clean` task [like it does for the project](https://github.com/elixir-lang/elixir/blob/master/lib/mix/lib/mix/tasks/clean.ex#L27-L30). Would make sense to call this task?

[...]

PR #10153, opened 2020-07-04
Deep clean the .erl files generated by yecc/leex.

vgsantoniazzi:
When cleaning with --deps flag, iterate over dependencies and remove generated .erl files.

This feature was discussed in #8262.

### DIFF
--- a/lib/mix/lib/mix/tasks/clean.ex
+++ b/lib/mix/lib/mix/tasks/clean.ex
@@ -36,9 +36,16 @@ defmodule Mix.Tasks.Clean do
       |> Path.join("*#{opts[:only]}")
 
     if opts[:deps] do
-      build
-      |> Path.wildcard()
-      |> Enum.each(&File.rm_rf/1)
+      deps_paths =
+        Mix.Project.deps_paths()
+        |> Map.values()
+        |> Enum.map(&Path.join(&1, "**/*.erl"))
+
+      for path <- [build] ++ deps_paths do
+        path
+        |> Path.wildcard()
+        |> Enum.each(&File.rm_rf/1)
+      end
     else
       build
       |> Path.join("lib/#{Mix.Project.config()[:app]}")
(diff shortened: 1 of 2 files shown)


vgsantoniazzi:
If the solution is good enough, I'll add the tests. I've opened the PR for discussion :bow:

lib/mix/lib/mix/tasks/clean.ex
```diff
@@ -36,9 +36,16 @@ defmodule Mix.Tasks.Clean do
       |> Path.join("*#{opts[:only]}")
 
     if opts[:deps] do
-      build
-      |> Path.wildcard()
-      |> Enum.each(&File.rm_rf/1)
+      deps_paths =
+        Mix.Project.deps_paths()
+        |> Map.values()
+        |> Enum.map(&Path.join(&1, "**/*.erl"))
```
josevalim, line 42: Unfortunately we cannot simply remove all .erl files because we will remove source files for many erlang projects, that have `.erl` files in their deps. The implementation of this functionality needs to be tied to the `Mix.Task.Compiler` behaviour and it is likely quite more complex than this. :(

michalmuskala: An alternative could be to modify the compilers that output the `.erl` files either not to do that at all, our output them inside `_build`.

vgsantoniazzi: What about removing `.erl` files that have their uncompiled version?

If the only way that you saw is going through the `Mix.Task.Compiler` I can try to do this.

josevalim: @michalmuskala I tried that back then but unfortunately I ran into major issues, I don't recall them right now. Modifying the compilers to not output the .erl files actually sounds great. It solves the problem altogether.

michalmuskala: Yeah, it will also be a fair bit of work, unfortunately. There are some limitations of the yecc and leex compilers themselves to work around to do this

vgsantoniazzi:
Ok, based on the comments, this PR should not be forward. Thanks for reviewing it.

PR #10158, opened 2020-07-06
Add 'default' parameter to `Access.at/1` (now `Access.at/2`)

henrik:
As suggested by José in https://groups.google.com/g/elixir-lang-core/c/2bq8SlGVSZs/m/-3ri7oQUBAAJ.

I tried to find a balance between doctests vs adding to the "actual" tests that I was happy with. But let me know what you prefer.

### DIFF
--- a/lib/elixir/lib/access.ex
+++ b/lib/elixir/lib/access.ex
@@ -7,7 +7,7 @@ defmodule Access do
 
   `Access` supports keyword lists (`Keyword`) and maps (`Map`) out
   of the box. Keywords supports only atoms keys, keys for maps can
-  be of any type. Both returns `nil` if the key does not exist:
+  be of any type. Both return `nil` if the key does not exist:
 
       iex> keywords = [a: 1, b: 2]
       iex> keywords[:a]
@@ -648,7 +648,7 @@ defmodule Access do
       ...> end)
       {"mary", [%{name: "john"}, %{name: "MARY"}]}
 
-  `at/1` can also be used to pop elements out of a list or
+  `at/2` can also be used to pop elements out of a list or
   a key inside of a list:
 
       iex> list = [%{name: "john"}, %{name: "mary"}]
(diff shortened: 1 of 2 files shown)


lib/elixir/lib/access.ex
```diff
@@ -670,49 +676,57 @@ defmodule Access do
...
-  defp get_and_update_at([head | rest], index, next, updates) when index > 0 do
-    get_and_update_at(rest, index - 1, next, [head | updates])
+  defp get_and_update_at([head | rest], index, default, next, updates) when index > 0 do
+    get_and_update_at(rest, index - 1, default, next, [head | updates])
   end
 
-  defp get_and_update_at([], _index, _next, updates) do
-    {nil, :lists.reverse(updates)}
+  defp get_and_update_at([], _index, default, _next, updates) do
+    {default, :lists.reverse(updates)}
```
josevalim, line 729: Hrm. In key/2, we will call “next” with the provided default. However here it doesn’t make sense, because we would need to fill in the list until the index. But this means the behavior is inconsistent within functions and within at/2 itself. Get is called with the default, get_and_update isn’t. Not sure how to proceed. Maybe we shouldn’t add default here after all? 🤔

henrik: Honestly I got a bit lost in all the steps of `get_and_update_at`, so I made a stab at it and figured that the tests or the PR would catch anything I missed. I'll try to digest this after a break, though I suspect I might not have much to add. Happy to skip this if you think it's best :)

josevalim: The patch was perfect, the issue is rather a conceptual one. Unfortunately I don't think there is an easy answer, sorry for leading you in the wrong direction. We should still be fine for `at!` tho!

henrik:
No worries at all. It was fun to do anyway, and now we know :)

PR #10159, opened 2020-07-06
Add a GitHub action to lint the Markdown.

jbampton:
Add a Markdownlint config file.

Lint Markdown for rules:
MD003 - Header style
MD014 Dollar signs used before commands without showing output

https://www.npmjs.com/package/markdownlint-cli

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,15 @@
+name: Lint Markdown
+
+on: [push, pull_request]
+
+jobs:
+  test:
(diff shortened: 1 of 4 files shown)


josevalim:
Hi @jbampton! Thanks for the PR but given we disable most rules and we are mostly consistent anyway, I would prefer to not add a new dependency.

I particularly don't require the dollar sign rule, I think it is most about showing its execution context than the output, and the header could be improved for the README, in case you want to send a separate PR for it. Thanks!

PR #10162, opened 2020-07-07
redact username/password in mix output log

gregors:
ok so let's consider I'm doing things ahem, the _wrong_ _way_ and don't have proper ssh keys set up for a private github repository. So instead of doing things the right way I slap something silly into my mix.exs file such as

```elixir
{ :my_repo, git: "https://USERNAME:#{System.get_env("GITHUB_TOKEN")}@github.com/gregors/my_repo.git }
```

Now where would I get such an idea? Maybe reading old things [like this](https://stephenbussey.com/2018/02/05/sourcing-libraries-from-private-github.html)

Now if one were going to do this, you end up seeing logs being displayed such as 
```bash
Getting apex (https://USERNAME:password@github.com/gregors/my_repo.git)
```

That's no good, we don't want our username and password ending up in log files (even if it does end up in our lock file ahem)

I know there's a quiet option for mix logging, but I think we should try to redact that info when possible.  I've  added some code & tests to highlight my goal. Additional knowledge and feedback greatly appreciated.

### DIFF
--- a/lib/mix/lib/mix/scm/git.ex
+++ b/lib/mix/lib/mix/scm/git.ex
@@ -8,11 +8,10 @@ defmodule Mix.SCM.Git do
 
   def format(opts) do
     if rev = get_opts_rev(opts) do
-      "#{opts[:git]} - #{rev}"
+      "#{redact_uri(opts[:git])} - #{rev}"
     else
-      opts[:git]
+      redact_uri(opts[:git])
     end
-    |> redact_git_url
   end
 
   def format_lock(opts) do
@@ -187,7 +186,7 @@ defmodule Mix.SCM.Git do
   defp validate_git_options(opts) do
     err =
       "You should specify only one of branch, ref or tag, and only once. " <>
-        "Error on Git dependency: #{opts[:git]}"
+        "Error on Git dependency: #{redact_uri(opts[:git])}"
 
     validate_single_uniq(opts, [:branch, :ref, :tag], err)
   end
@@ -233,13 +232,10 @@ defmodule Mix.SCM.Git do
     end
   end
 
-  defp redact_git_url(url) do
-    cond do
-      String.match?(url, ~r'https://[^:]+:[^:@]+@.+'i) ->
-        String.replace(url, ~r'https://[^:]+:[^:@]+(?=@.+)'i, "https://****:****")
-
-      true ->
-        url
+  defp redact_uri(git) do
+    case URI.parse(git) do
+      %{userinfo: nil} -> git
+      uri -> URI.to_string(%{uri | userinfo: "****:****"})
     end
   end
 
--- a/lib/mix/lib/mix/scm/git.ex
+++ b/lib/mix/lib/mix/scm/git.ex
@@ -12,6 +12,7 @@ defmodule Mix.SCM.Git do
     else
       opts[:git]
     end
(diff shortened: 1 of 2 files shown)


gregors:
why are the tests so flakey?

lib/mix/lib/mix/scm/git.ex
```diff
@@ -12,6 +12,7 @@ defmodule Mix.SCM.Git do
     else
       opts[:git]
     end
+    |> redact_git_url
```
josevalim, line 15: Instead of parsing the whole text with a regex, my suggestion is to use the functions in the URI module, something like this:

```elixir
defp redact_uri(git) do
  case URI.parse(git) do
    %{username: nil, password: nil} -> git
    uri -> URI.to_string(%{uri | username: "****", password: "****"})
  end
end
```

Then you call it with `opts[:git]` on both `format` and `format_lock`. WDYT?

josevalim:
> why are the tests so flakey?

Windows, unfortunately. The Windows CI gets stuck at many points during filesystem operations, sometimes even when cloning the github repo, installing deps, etc. :(

gregors:
@josevalim I added the alternate redact implementation - slight tweak as  userinfo is combined. Also redacted leakage in an error statement. `format_lock` doesn't actually leak the url. It's only used when reporting. I haven't tracked down where the actual url gets written to the lock file yet. But I'm wondering if doing so might break things unintentionally for some people.

lib/mix/lib/mix/scm/git.ex
```diff
@@ -232,6 +232,13 @@ defmodule Mix.SCM.Git do
     end
   end
 
+  defp redact_uri(git) do
+    case URI.parse(git) do
+      %{userinfo: nil} -> git
+      uri -> URI.to_string(%{uri | userinfo: "****:****"})
```
fertapric, line 238: Nitpick: I think the password in the `userinfo` subcomponent is optional, being `http://username@example.com`, `http://username:@example.com`, and `http://username:password@example.com` all valid. 

If that's the case, I would not use a semicolon `:` in the placeholder to avoid any confusion:

```suggestion
      uri -> URI.to_string(%{uri | userinfo: "****"})
```

Thoughts?

josevalim: @fertapric in your case there is a colon though. Plus without `:` it changes the meaning. I personally prefer the current approach. :)

josevalim:
@gregors `format_lock` is the command used by `mix deps`. But it seems we indeed don't use the git url there, so we are good. :)

lib/mix/lib/mix/scm/git.ex
fertapric: @josevalim I was mentioning that because GitHub allows to clone a repo passing a GitHub token in the _userinfo_: `git clone https://mygithubtoken@github.com/fertapric/my_repo.git` 🙂 

But I'm ok with the current placeholder 👍

josevalim: Oh, I see. The `:` is not required at all. Hrm, we could check if `:` is there or not and generate something accordingly but overall I don't think it matters? I am fine with whatever you pick. :)

fertapric:
Thanks @gregors! 💜  First commit 🙌

PR #10165, opened 2020-07-08
Fix shell script: double quote to prevent globbing and word splitting.

jbampton:
Checked with ShellCheck.
https://www.shellcheck.net/

https://github.com/koalaman/shellcheck/wiki/SC2086


josevalim:
As CI shows, The commands were correct. :) Those flags are meant to splat and not be a single value. we already use shellcheck to validate those scripts. Thank you!

PR #10169, opened 2020-07-08
Allow function captures in pipelines

### LINKED ISSUE #10154: Support literal anonymous functions in pipe
The goal is to support the following syntaxes:

```elixir
string
|> &Regex.replace(~r/foo/, &1, "bar")
|> fn x -> Regex.replace(~r/bar/, x, "baz") end
```

I have originally argued against this extension because, if we add the syntax above, the following is not true:

```elixir
replace_foo = &Regex.replace(~r/foo/, &1, "bar")
replace_bar = fn x -> Regex.replace(~r/bar/, x, "baz") end

string
|> replace_foo
|> replace_bar
```

I.e. I cannot simply replace the right-side of `|>` by a variable. However, this was a red herring because we can *never* replace the right side of `|>` by a variable in Elixir. Therefore, the only issue with this approach is the ambiguity between `|> var`, which means piping to a local function unless you write it as `|> var.()`. We could address this ambiguity by requiring parenthesis if you are trying to pipe to an existing variable.

## Tasks

Those can be submitted as distinct PRs:

  * [ ] Support `fn` in pipelines (it should raise if the function expects more than one argument)
  * [ ] Support `&local/1` and `&Mod.remote/1` in pipelines (raise if arity is not 1)
  * [ ] Support `&` in pipelines (it should raise if there is a nested `&` or `&2`)
  * [ ] Warn if piping to a `|> var` and there is such var in scope (either `var()` or `var.()` should be used)

hauleth:
- [ ] Support `fn` in pipelines (it should raise if the function expects more than one argument)
- [x] Support `&local/1` and `&Mod.remote/1` in pipelines (raise if arity is not 1)
- [ ] Support `&` in pipelines (it should raise if there is a nested `&` or `&2`)
- [ ] Warn if piping to `a |> var` and there is such `var` in scope (either `var()` or `var.()` should be used)

Close #10154

### DIFF
--- a/lib/elixir/lib/macro.ex
+++ b/lib/elixir/lib/macro.ex
@@ -232,6 +232,10 @@ defmodule Macro do
     unpipe(right, unpipe(left, acc))
   end
 
+  defp unpipe({:&, _, [{:/, _, [func, 1]}]}, acc) do
+    [{func, 0} | acc]
+  end
+
   defp unpipe(other, acc) do
     [{other, 0} | acc]
   end
--- a/lib/elixir/test/elixir/kernel_test.exs
+++ b/lib/elixir/test/elixir/kernel_test.exs
@@ -1058,6 +1058,14 @@ defmodule KernelTest do
       assert Enum.map([1, 2, 3], &(&1 |> twice |> twice)) == [4, 8, 12]
     end
 
+    test "function capture" do
+      # We need to split this as `==` bounds stronger than `&foo/1` which was
+      # causing problems
+      pipe = [1, [2], 3] |> (&List.flatten/1)
+
+      assert pipe == [1, 2, 3]
+    end
+
     test "with anonymous functions" do
       assert 1 |> (&(&1 * 2)).() == 2
       assert [1] |> (&hd(&1)).() == 1


lib/elixir/lib/macro.ex
```diff
@@ -232,6 +232,10 @@ defmodule Macro do
     unpipe(right, unpipe(left, acc))
   end
 
+  defp unpipe({:&, _, [{:/, _, [func, 1]}]}, acc) do
+    [{func, 0} | acc]
+  end
```
josevalim, line 237: Thank you for the PR. We should handle this in the `pipe` function rather than in `unpipe`. Unpipe is only about breaking it apart. :) We should also improve the error message for `{:&, _, [{:/, _, [func, n]}]} when n != 1` to say you must pipe to a function of arity 1.

hauleth: My point in having it in `unpipe` is to output `{func, n}` in case of `&foo(a, &1, c)` instead of calling lambda. However in this case I can move this to `pipe` as the mentioned case will be only in one situation which can be handled in `unpipe` and all other cases can be handled in `pipe`.

josevalim: It is really unexpected for unpipe to break anything besides pipes apart though. So we need to change it.

josevalim: The other cases shouldn’t be handled in unpipe as well. Unpipe is about breaking pipes into a list without changing the underlying AST.

josevalim: We also shouldn’t expect `&` to be in the format of a function calls it could be any expression, even &[&1], so we need to handle it in a case, as described in the original issue.

hauleth: Yes, but my point is that `&foo(a, &1, b)` translates much better to `{func, 1}` than `{func, 0}` where it is then processed further in the `pipe`. As otherwise the 3rd argument to `pipe` is almost pointless as we do not use it anywhere.

josevalim:
Closing this as per the parent issue. Thanks!

PR #10170, opened 2020-07-09
Support anonymous function in pipeline

### LINKED ISSUE #10154: Support literal anonymous functions in pipe

edisonywh:
This commit rewrites `fn` into `case` in pipeline. With this change now
we can support using anonymous function in pipeline, like so:

```
string
|> fn x -> Regex.replace(~r/bar/, x, "baz") end
```

Raise error if anonymous functions in pipeline expect more than 1-arity.

Fixes #10154.

### DIFF
--- a/lib/elixir/lib/macro.ex
+++ b/lib/elixir/lib/macro.ex
@@ -265,13 +265,15 @@ defmodule Macro do
             "since #{unquote}/1 is used to build the Elixir AST itself"
   end
 
-  # {:fn, _, _} is what we get when we pipe into an anonymous function without
-  # calling it, for example, `:foo |> (fn x -> x end)`.
-  def pipe(expr, {:fn, _, _}, _integer) do
+  def pipe(expr, {:fn, _line, [{:->, _, [args | _]}]}, _integer)
+      when is_list(args) and length(args) > 1 do
     raise ArgumentError,
-          "cannot pipe #{to_string(expr)} into an anonymous function without" <>
-            " calling the function; use something like (fn ... end).() or" <>
-            " define the anonymous function as a regular private function"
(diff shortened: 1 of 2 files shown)


edisonywh:
Apologies, my tab on the GitHub Issue was not refreshed so I did not see that it was closed already. Closing this then.

fertapric:
@edisonywh Thanks for the effort! 🙌

PR #10185, opened 2020-07-13
Add support for :hour and :minute time units

sorentwo:
This enhances time unit aware functions in the `Time` module so that they can operate on `:hour` and `:minute`, along with standard `System.time_unit/0` values. This eliminates the need to convert all values to seconds, which allows more expressive time manipulation.

Along with changes to `Time.add/3` and `Time.diff/3`, this adds a new extended `time_unit` type to `Time`.

### DIFF
--- a/lib/elixir/lib/calendar/time.ex
+++ b/lib/elixir/lib/calendar/time.ex
@@ -491,7 +491,7 @@ defmodule Time do
   @doc since: "1.6.0"
   @spec add(Calendar.time(), integer, time_unit()) :: t()
   def add(%{calendar: calendar} = time, number, unit \\ :second) when is_integer(number) do
-    total = time_to_microseconds(time) + convert_time_unit(number, unit)
+    total = time_to_microseconds(time) + time_unit_to_microseconds(number, unit)
     parts = Integer.mod(total, @parts_per_day)
     {hour, minute, second, microsecond} = calendar.time_from_day_fraction({parts, @parts_per_day})
 
@@ -779,9 +779,11 @@ defmodule Time do
 
   ## Helpers
 
-  defp convert_time_unit(number, :hour), do: convert_time_unit(number * 60 * 60, :second)
-  defp convert_time_unit(number, :minute), do: convert_time_unit(number * 60, :second)
-  defp convert_time_unit(number, unit), do: System.convert_time_unit(number, unit, :microsecond)
+  defp time_unit_to_microseconds(number, :hour), do: convert_time_unit(number * 60 * 60, :second)
+  defp time_unit_to_microseconds(number, :minute), do: convert_time_unit(number * 60, :second)
+
+  defp time_unit_to_microseconds(number, unit),
+    do: System.convert_time_unit(number, unit, :microsecond)
 
   defp to_day_fraction(%{
          hour: hour,
--- a/lib/elixir/lib/calendar/time.ex
+++ b/lib/elixir/lib/calendar/time.ex
@@ -779,8 +779,11 @@ defmodule Time do
 
   ## Helpers
 
-  defp time_unit_to_microseconds(number, :hour), do: convert_time_unit(number * 60 * 60, :second)
-  defp time_unit_to_microseconds(number, :minute), do: convert_time_unit(number * 60, :second)
+  defp time_unit_to_microseconds(number, :hour),
+    do: time_unit_to_microseconds(number * 60 * 60, :second)
+
+  defp time_unit_to_microseconds(number, :minute),
+    do: time_unit_to_microseconds(number * 60, :second)
 
   defp time_unit_to_microseconds(number, unit),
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/calendar/time.ex
```diff
@@ -656,12 +666,9 @@ defmodule Time do
...
 
-  The answer can be returned in any `unit` available from
-  `t:System.time_unit/0`. If the first time value is earlier than
-  the second, a negative number is returned.
-
-  This function returns the difference in seconds where seconds
-  are measured according to `Calendar.ISO`.
+  If the first time value is earlier than the second, a negative number is
+  returned. The answer can be returned in any `unit` available from
+  `t:time_unit/0`.
```
sorentwo, line 671: These two sections seemed contradictory so I removed the second segment. Looking at it now I realize that there isn't any mention of `Calendar.ISO` units, so I need to put that back.

sorentwo:
This is the first of four changes (Time, Date, DateTime, NaiveDateTime). I opened the PR to get feedback on the approach before proceeding on the next few changes. The changes are atomic and stand alone so I opted for separate PRs, but I'm happy to combine them instead if desired.

lib/elixir/lib/calendar/time.ex
```diff
@@ -679,14 +686,29 @@ defmodule Time do
...
-  @spec diff(Calendar.time(), Calendar.time(), System.time_unit()) :: integer
+  @spec diff(Calendar.time(), Calendar.time(), time_unit()) :: integer
   def diff(time1, time2, unit \\ :second)
 
+  def diff(%{calendar: Calendar.ISO} = time1, %{calendar: Calendar.ISO} = time2, :hour) do
+    time1.hour - time2.hour
+  end
+
+  def diff(%{calendar: Calendar.ISO} = time1, %{calendar: Calendar.ISO} = time2, :minute) do
+    (time1.hour - time2.hour) * 60 + (time1.minute - time2.minute)
```
josevalim, line 709: This will return the wrong result for this:

    Time.diff(~T[00:01:20], ~T[00:00:40])

It should return zero. I am afraid you will have to add all seconds and microseconds together, as in the clause below, and only then extracts minutes and hours.

lib/elixir/lib/calendar/time.ex
```diff
@@ -656,12 +666,9 @@ defmodule Time do
   additional information about a date or time zone is ignored when calculating
   the difference.
 
-  The answer can be returned in any `unit` available from
-  `t:System.time_unit/0`. If the first time value is earlier than
-  the second, a negative number is returned.
-
-  This function returns the difference in seconds where seconds
-  are measured according to `Calendar.ISO`.
+  If the first time value is earlier than the second, a negative number is
```
josevalim, line 669:
```suggestion
  The `unit` controls the measurement unit for the returned value.
  The `unit` conversion always truncates lower units. For example, a
  difference of 90 seconds will be returned as 1 if `:minute` is given.

  If the first time value is earlier than the second, a negative number is
```

lib/elixir/lib/calendar/time.ex
sorentwo: Excellent point. It felt like something was missing here.

sorentwo: I've modified it to use the base `diff` implementation and `div/2` the result 👍

lib/elixir/lib/calendar/time.ex
```diff
@@ -469,19 +476,22 @@ defmodule Time do
...
+      ~T[11:00:00.000000]
 
   """
   @doc since: "1.6.0"
-  @spec add(Calendar.time(), integer, System.time_unit()) :: t
+  @spec add(Calendar.time(), integer, time_unit()) :: t()
   def add(%{calendar: calendar} = time, number, unit \\ :second) when is_integer(number) do
-    number = System.convert_time_unit(number, unit, :microsecond)
-    total = time_to_microseconds(time) + number
+    total = time_to_microseconds(time) + convert_time_unit(number, unit)
```
fertapric, line 494: I would be more specific about the final unit:

```suggestion
    total = time_to_microseconds(time) + time_unit_to_microseconds(number, unit)
```

or:

```suggestion
    total = time_to_microseconds(time) + convert_time_unit(number, unit, :microsecond)
```


Thoughts? 🙂

sorentwo: I like the consistency of `time_unit_to_microseconds/2` 👍

fertapric:
Thanks @sorentwo! 💜

entertainyou:
It looks this PR got reverted, @josevalim Sorry to interrupt, why it got reverted?

wojtekmach:
see this comment and discussion around it for the reason: https://github.com/elixir-lang/elixir/pull/10199#issuecomment-659525151

PR #10193, opened 2020-07-14
Add `Access.at!`

henrik:
As discussed in https://groups.google.com/d/msgid/elixir-lang-core/3a5d1489-dfa5-4c0c-bcd1-5cb85eac2f3fn%40googlegroups.com.

This exception is a copy-paste of `Enum.OutOfBoundsError`. Makes me wonder if they should be joined into a single unprefixed `OutOfBoundsError` at some point. There is only an unprefixed version of `KeyError`, for example.

### DIFF
--- a/lib/elixir/lib/access.ex
+++ b/lib/elixir/lib/access.ex
@@ -683,36 +683,68 @@ defmodule Access do
   end
 
   defp at(:get_and_update, data, index, next) when is_list(data) do
-    get_and_update_at(data, index, next, [])
+    get_and_update_at(data, index, next, [], fn -> nil end)
   end
 
   defp at(_op, data, _index, _next) do
     raise "Access.at/1 expected a list, got: #{inspect(data)}"
   end
 
-  defp get_and_update_at([head | rest], 0, next, updates) do
+  defp get_and_update_at([head | rest], 0, next, updates, _default_fun) do
     case next.(head) do
       {get, update} -> {get, :lists.reverse([update | updates], rest)}
       :pop -> {head, :lists.reverse(updates, rest)}
     end
   end
 
-  defp get_and_update_at([_ | _] = list, index, next, updates) when index < 0 do
+  defp get_and_update_at([_ | _] = list, index, next, updates, default_fun) when index < 0 do
     list_length = length(list)
 
     if list_length + index >= 0 do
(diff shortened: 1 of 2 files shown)


henrik:
Force-pushed an off-by-one fix for `get_and_update_in`.

lib/elixir/lib/access.ex
```diff
@@ -715,6 +715,42 @@ defmodule Access do
...
+
+  defp at!(:get, data, index, next) when is_list(data) do
+    case Enum.fetch(data, index) do
+      {:ok, value} -> next.(value)
+      :error -> raise Access.OutOfBoundsError
+    end
+  end
+
+  defp at!(:get_and_update, data, index, next) when is_list(data) do
+    if index >= length(data) do
```
josevalim, line 743: Unfortunately this will traverse the data twice and it may not work for negative indexes. My suggestion is to allow an anonymous function to be given to `get_and_update_at` which will be called whenever it returns `nil`. For the regular case, the anonymous function will be `fn -> nil end` but for those case we will raise. WDYT?

lib/elixir/lib/access.ex
```diff
@@ -715,6 +715,42 @@ defmodule Access do
...
+  """
+  @spec at!(integer) :: access_fun(data :: list, get_value :: term)
+  def at!(index) when is_integer(index) do
+    fn op, data, next -> at!(op, data, index, next) end
+  end
+
+  defp at!(:get, data, index, next) when is_list(data) do
+    case Enum.fetch(data, index) do
+      {:ok, value} -> next.(value)
+      :error -> raise Access.OutOfBoundsError
```
josevalim, line 738: It is fine to use `Enum.OutOfBoundsError` instead of introducing a new one. :)

lib/elixir/lib/access.ex
henrik: Good catch, and makes sense 👍🏻 Thanks!

henrik:
Alright, I've force-pushed [those fixes](https://github.com/henrik/elixir/commit/4931a9ca8723fdbe2e6ec1251a8faa70a1e0ef57) though GitHub hasn't updated the PR with them yet at the time of writing.

josevalim:

PR #10194, opened 2020-07-14
Remove incorrect arities in test descriptions

henrik:
These tests actually use `get_and_update_in/3` and `pop_in/2`.

I removed the arities, rather than correcting them, since it seems like unnecessary detail here.

Also, the `get_in` test just above does not mention arity.


henrik:
❤️

PR #10198, opened 2020-07-15
Warn if a variable that looks like a compiler variable is unused

### LINKED ISSUE #10197: __MODULE__ with extra underscores still parses and compiles
# Environment

* Operating system: 

### Current behavior

```Elixir
iex(1)> defmodule Example do
...(1)>   defstruct []
...(1)>
...(1)>   def func(%____MODULE__{}), do: true
...(1)> end
{:module, Example,
 <<70, 79, 82, 49, 0, 0, 7, 76, 66, 69, 65, 77, 65, 116, 85, 56, 0, 0, 0, 194,
   0, 0, 0, 20, 14, 69, 108, 105, 120, 105, 114, 46, 69, 120, 97, 109, 112, 108,
   101, 8, 95, 95, 105, 110, 102, 111, 95, ...>>, {:func, 1}}
iex(2)> Example.func(%Example{})
true
```

### Expected behavior

A compilation error.

### Thoughts

Is this expected? I would generally expect the compiler to be more strict. Additional `_` after `MODULE` are also ignored.

josevalim:
Closes #10197.

### DIFF
--- a/lib/elixir/src/elixir_env.erl
+++ b/lib/elixir/src/elixir_env.erl
@@ -93,7 +93,7 @@ reset_unused_vars(#{unused_vars := {_Unused, Version}} = E) ->
 
 check_unused_vars(#{unused_vars := {Unused, _Version}} = E) ->
   [elixir_errors:form_warn([{line, Line}], E, ?MODULE, {unused_var, Name}) ||
-    {{Name, _}, Line} <- maps:to_list(Unused), Line /= false, not_underscored(Name)],
+    {{Name, _}, Line} <- maps:to_list(Unused), Line /= false, is_unused_var(Name)],
   E.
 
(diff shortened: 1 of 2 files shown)


josevalim:

michalmuskala:
Should this warning be based on some edit distnace to one of the compiler variables. I'm afraid it might casue some false positives. It wasn't documented (and still isn't) that variables of the form `__UPPER__` are reserved for the compiler.

josevalim:
They are not really reserved to the compiler as someone could indeed use `__UPPER__`. Indeed there is a chance for a false positive here, in case someone willingly chose to name a var something like `_HTTP` or similar but I would argue that, even in those cases, it is probably best to rename it to `_http` or similar as someone would be inclined to think `__HTTP` does have a special meaning. We can always collect feedback too and improve accordingly. :)

PR #10200, opened 2020-07-16
Use GitHub Actions for Windows CI

fertapric:
Here are some benefits of moving to GitHub
Actions:

  * The Docker image `fertapric/elixir-ci:*` is no longer needed.

  * Build times has been reduced ~50%: current times are ~20 mins, and with GitHub Actions ~10 mins.

  * Jobs seem to be scheduled faster: a few seconds compared to minutes.

  * Running instances seem more stable: I have not experienced `ExUnit.TimeoutError` errors nor memory allocation issues.


eksperimental:
Well. I guess you cannot compete with Microsoft on this one. Good choice

PR #10205, opened 2020-07-18
Update typespecs and enforce with and limit in Inspect.Algebra

eksperimental:
Additionally it rephrases docs.
************************

I made a typo in `:infinity` and it was giving a cryptic error not pointing me to what was wrong. ArithmeticError

### DIFF
--- a/lib/elixir/lib/inspect/algebra.ex
+++ b/lib/elixir/lib/inspect/algebra.ex
@@ -31,22 +31,24 @@ defmodule Inspect.Opts do
       options. Useful when implementing the `Inspect` protocol for nested structs
       to pass the custom options through.
 
-    * `:inspect_fun` (since v1.9.0) - a function to build algebra documents,
-      defaults to `Inspect.inspect/2`.
+    * `:inspect_fun` (since v1.9.0) - a function to build algebra documents.
+      Defaults to `Inspect.inspect/2`.
 
     * `:limit` - limits the number of items that are inspected for tuples,
-      bitstrings, maps, lists and any other collection of items. It does not
-      apply to printable strings nor printable charlists and defaults to 50.
+      bitstrings, maps, lists and any other collection of items. It neither
+      applies to printable strings nor to printable charlists.
       If you don't want to limit the number of items to a particular number,
-      use `:infinity`.
+      use `:infinity`. It accepts a positive integer or `:infinity`.
+      Defaults to `50`.
 
-    * `:pretty` - if set to `true` enables pretty printing, defaults to `false`.
+    * `:pretty` - if set to `true` enables pretty printing. Defaults to `false`.
 
     * `:printable_limit` - limits the number of characters that are inspected
       on printable strings and printable charlists. You can use `String.printable?/1`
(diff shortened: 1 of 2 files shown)


lib/elixir/lib/inspect/algebra.ex
```diff
@@ -31,30 +31,32 @@ defmodule Inspect.Opts do
...
-    * `:inspect_fun` (since v1.9.0) - a function to build algebra documents,
-      defaults to `Inspect.inspect/2`.
+    * `:inspect_fun` (since v1.9.0) - a function to build algebra documents.
+      Defaults to `Inspect.inspect/2`.
 
     * `:limit` - limits the number of items that are inspected for tuples,
-      bitstrings, maps, lists and any other collection of items. It does not
-      apply to printable strings nor printable charlists and defaults to 50.
+      bitstrings, maps, lists and any other collection of items. It neither
+      applies to printable strings nor to printable charlists.
```
josevalim, line 39: Is the previous sentence incorrect? I am not a native speaker, so I always struggle with neither. I am wondering if there is a clearer way to write this?

eksperimental: nor is usually used after neither.

This one reads better IMO.

* `:limit` - limits the number of items that are inspected for tuples, 
bitstrings, maps, lists and any other collection of items, with the exception of 
printable strings and printable charlists.

josevalim: I love it. :+1: I would even add:

> with the exception of printable strings and printable charlists which use the `:printable_limit` option

WDYT?

eksperimental: Yeah.. it looks great. pushed!

lib/elixir/lib/inspect/algebra.ex
```diff
@@ -900,7 +906,12 @@ defmodule Inspect.Algebra do
...
 
-  @spec fits?(width :: integer(), column :: integer(), break? :: boolean(), entries) :: boolean()
+  @spec fits?(
+          width :: non_neg_integer(),
+          column :: non_neg_integer(),
+          break? :: boolean(),
+          entries
+        ) :: boolean()
         when entries:
                maybe_improper_list({integer(), mode(), t()}, {:tail, boolean(), entries} | [])
```
eksperimental, line 916: I  am not sure what this interger() is but you could tell me if is should be non_neg_integer() or not.

josevalim:
Please note it broke the Logger test suite. :)

eksperimental:
> Please note it broke the Logger test suite. :)

Doesn't seem easy to solve after tinkering with it for a while, and I have never got my hands into Logger.
Do you think you could see if you can fix it?

josevalim:
@eksperimental please replace Logger.Utils.inspect_data by this:

```elixir
  defp inspect_data([data | _], opts) do
    width = if opts.width == :none, do: 80, else: opts.width

    data
    |> Inspect.Algebra.to_doc(opts)
    |> Inspect.Algebra.format(width)
  end
```

eksperimental:
done!

josevalim:

PR #10207, opened 2020-07-18
Make iex autocomplete work after pin and negation operators

lukaszsamson:


josevalim:

PR #10209, opened 2020-07-20
Properly handle GenServer message in logger translators

### LINKED ISSUE #10208: `GenServer.format_report/1` is not used for unknown `handle_info/2` messages
### Environment

  ```
  Erlang/OTP 22 [erts-10.7] [source] [64-bit] [smp:12:12] [ds:12:12:10] [async-threads:1] [hipe]

  Elixir 1.10.4 (compiled with Erlang/OTP 22)
  ```
* Operating system:
  ```
  Darwin NiunioBook-2.local 19.5.0 Darwin Kernel Version 19.5.0: Tue May 26 20:41:44 PDT 2020; root:xnu-6153.121.2~2/RELEASE_X86_64 x86_64 i386 MacBookPro15,1 Darwin
  ```

### Current behavior

When running this script:

```elixir
defmodule Foo do
  use GenServer

  def init(_), do: {:ok, []}
end

{:ok, pid} = GenServer.start_link(Foo, [])

send(pid, :foo)
```

The resulting error message is:

```
12:53:46.870 [error] [message: :foo, module: Foo, name: #PID<0.98.0>]
```

### Expected behavior

It should use `report_cb` from log message metadata or output message in the format similar to such:

```
12:53:46.870 [error] GenServer #PID<0.98.0> received unexpected message in handle_info/2: :test
```

hauleth:
Close #10208

### DIFF
--- a/lib/logger/lib/logger/handler.ex
+++ b/lib/logger/lib/logger/handler.ex
@@ -117,6 +117,7 @@ defmodule Logger.Handler do
   defp do_log(level, msg, metadata, config) do
     %{level: erl_min_level} = :logger.get_primary_config()
     min_level = erlang_level_to_elixir_level(erl_min_level)
+    %{truncate: truncate} = config
 
     try do
       case msg do
@@ -139,6 +140,9 @@ defmodule Logger.Handler do
            "Failure while translating Erlang's logger event\n",
            Exception.format(:error, e, __STACKTRACE__)
          ], metadata}
+    else
+      :none -> {translate_fallback(msg, metadata, truncate), metadata}
+      other -> other
     end
   end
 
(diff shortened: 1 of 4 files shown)


josevalim:
Thanks! I thought we were supposed to fallback to report_cb if there is no translation. What am I missing? :)

hauleth:
@josevalim it seems that if the report is in form of `%{label: term(), report: term()}` then the `report_cb` will not be called. In addition to that, I think that we would need something like `IO.format/2` that would print data in "Elixir-like" format for `~w` and `~p` modifiers. So the Elixir users would not be confused by messages like:

```
'Elixir.MyApp.Server' <0.1.0> received unexpected message in handle_info/2: {foo, #{bar => <<"baz">>}}
```

When they would expect something like:

```
MyApp.Server #PID<0.1.0> received unexpected message in handle_info/2: {:foo, %{bar: "baz"}}
```

hauleth:
I already created basis for [parsing `io:format()` string](https://github.com/hauleth/sigil_f/blob/49ccdfcb562b6444df7e8b80af6b09a4ca659868/lib/sigil_f.ex#L51) which could be used for such parser. So we would be able to directly reuse most of `report_cb` output instead of additional handling of it in the `Logger.Translators`.

josevalim:
We also have this functionality in `Logger` here: https://github.com/elixir-lang/elixir/blob/master/lib/logger/lib/logger/utils.ex#L91

We use it with `io_lib:build_text/2`.

hauleth:
@josevalim this probably should be back ported to 1.10 as it is issue there as well.

josevalim:

PR #10213, opened 2020-07-20
Correct the @since attribute for {Keyword, Map}.replace3

eksperimental:
They have been undeprecated in 284881c370a241042ad5be88914446f893383f42,
but the functions have been available in the language, beside de deprecation warning, since v1.5.0
when they made their way into the language in 3f88ab745eed2209007e9ff872b15fc5df252867


josevalim:
They could not be used before without warnings though, so I believe v1.11.0 is more accurate. thank you!

PR #10214, opened 2020-07-20
Raise ArgumentError in Keyword.keys when list is not a keyword list

### LINKED ISSUE #10010: Keyword.keys/1 accepts non-keyword lists and returns their keys
### Environment

* Elixir & Erlang/OTP versions (elixir --version): 1.9.4 and OTP 22
* Operating system: Linux dev 4.19.76-linuxkit #1 SMP Fri Mar 6 18:33:07 UTC 2020 x86_64 GNU/Linux (inside Docker)

### Current behavior

```
iex(7)> Keyword.has_key?([{"my_key", 42}], "my_key")
** (FunctionClauseError) no function clause matching in Keyword.has_key?/2

    The following arguments were given to Keyword.has_key?/2:

        # 1
        [{"my_key", 42}]

        # 2
        "my_key"

        def has_key?(keywords, key) when is_list(keywords) and is_atom(key)

    (elixir) lib/keyword.ex:810: Keyword.has_key?/2
iex(7)> Keyword.keys([{"my_key", 42}])
["my_key"]
```

as you can see in the above snippet when using `Keyword.has_key?/2` it properly results in an error saying it failed to match any clause because the key must be an `atom` (as shown in `is_atom(key)`). However, in the following function `Keyword.keys/1` it returns the `"my_key"` key which is not an `atom`. 

This behaviour feels inconsistent to me. Because, AFAIK, `Keyword.t` represents a list of pairs `{atom(), term()}`, as seen here:

```
iex(11)> t Keyword
@type key() :: atom()

@type t() :: [{key(), value()}]
```

---

Going further down this path I've noticed that the `lists.keymember/3` function from Erlang's core library also provides inconsistent behaviour:

```
iex(8)> :lists.keymember("my_key", 1, [{"my_key", 42}])
true
```

EDIT: this snippet above is actually correct because in Erlang's `lists` the key can be anything as per the [docs](https://erlang.org/doc/man/lists.html#keymember-3).

---

Finally, looking at Elixir's implementation of `Keyword.keys/2` [here](https://github.com/elixir-lang/elixir/blob/bdde9e619bc9f3565f1efd01a7ab46fd880d727c/lib/elixir/lib/keyword.ex#L439) this seems to result due to using Erlang's `lists` module which probably is the root cause of these behaviours.

### Expected behavior
The `Keyword` module only accepts actual `Keyword.t()`.

### Solution
I'm not sure what the solution here would be since Elixir is a dynamic language and we don't know if a given list respects the `Keyword.t` spec before actually evaluating it. Maybe, check if each key is an `atom` and raise an error otherwise? Or just return the keys which are `atom`?

eksperimental:
Previously it would return something like this:

** (FunctionClauseError) no function clause matching in anonymous fn/1 in Keyword.keys/1

    The following arguments were given to anonymous fn/1 in Keyword.keys/1:

        # 1
        {"fetch/2", {:fetch, 2}}

Closes #10010, and it improves 3fd68ef62a90f3b914ed2e881e5d152431f78bbc

### DIFF
--- a/lib/elixir/lib/keyword.ex
+++ b/lib/elixir/lib/keyword.ex
@@ -436,7 +436,7 @@ defmodule Keyword do
       [:a, :b, :a]
 
       iex> Keyword.keys([{:a, 1}, {"b", 2}, {:c, 3}])
-      ** (ArgumentError) expected a keyword list, but an element in the list is not a keyword; got: {"b", 2}
+      ** (ArgumentError) expected a keyword list, but an element in the list is not a two-element tuple with an atom key; got: {"b", 2}
 
   """
   @spec keys(t) :: [key]
@@ -448,7 +448,7 @@ defmodule Keyword do
 
         element ->
           raise ArgumentError,
-                "expected a keyword list, but an element in the list is not a keyword; got: #{
+                "expected a keyword list, but an element in the list is not a two-element tuple with an atom key; got: #{
                   inspect(element)
                 }"
       end,
--- a/lib/elixir/lib/keyword.ex
+++ b/lib/elixir/lib/keyword.ex
@@ -436,7 +436,7 @@ defmodule Keyword do
       [:a, :b, :a]
 
       iex> Keyword.keys([{:a, 1}, {"b", 2}, {:c, 3}])
-      ** (ArgumentError) expected a keyword list, but an element in the list is not a two-element tuple with an atom as its first element, got: {"b", 2}
+      ** (ArgumentError) expected a keyword list, but an entry in the list is not a two-element tuple with an atom as its first element, got: {"b", 2}
 
   """
   @spec keys(t) :: [key]
@@ -452,7 +452,7 @@ defmodule Keyword do
     catch
       element ->
         raise ArgumentError,
-              "expected a keyword list, but an element in the list is not a two-element tuple with an atom as its first element, " <>
+              "expected a keyword list, but an entry in the list is not a two-element tuple with an atom as its first element, " <>
                 "got: #{inspect(element)}"
     end
   end
--- a/lib/elixir/lib/keyword.ex
+++ b/lib/elixir/lib/keyword.ex
@@ -444,27 +444,16 @@ defmodule Keyword do
     try do
       :lists.map(
         fn
-          {key, _} when is_atom(key) ->
-            key
-
-          element ->
-            throw(element)
+          {key, _} when is_atom(key) -> key
+          element -> throw(element)
         end,
         keywords
       )
     catch
       element ->
-        stacktrace =
-          Enum.drop_while(__STACKTRACE__, fn
-            {Keyword, :keys, 1, _} -> false
-            {Keyword, _, _, _} -> true
-          end)
-
-        reraise ArgumentError,
-                "expected a keyword list, but an element in the list is not a two-element tuple with its first element being an atom; got: #{
-                  inspect(element)
-                }",
-                stacktrace
+        raise ArgumentError,
+              "expected a keyword list, but an element in the list is not a two-element tuple with an atom key as first element, " <>
+                "got: #{inspect(element)}"
     end
   end
 
--- a/lib/elixir/lib/keyword.ex
+++ b/lib/elixir/lib/keyword.ex
@@ -436,7 +436,7 @@ defmodule Keyword do
       [:a, :b, :a]
 
       iex> Keyword.keys([{:a, 1}, {"b", 2}, {:c, 3}])
-      ** (ArgumentError) expected a keyword list, but an element in the list is not a two-element tuple with its first element being an atom; got: {"b", 2}
+      ** (ArgumentError) expected a keyword list, but an element in the list is not a two-element tuple with an atom as its first element, got: {"b", 2}
 
   """
   @spec keys(t) :: [key]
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/keyword.ex
```diff
@@ -431,13 +431,29 @@ defmodule Keyword do
...
+    :lists.map(
+      fn
+        {key, _} when is_atom(key) ->
+          key
+
+        element ->
+          raise ArgumentError,
+                "expected a keyword list, but an element in the list is not a keyword; got: #{
+                  inspect(element)
+                }"
```
josevalim, line 453:
```suggestion
                "expected a keyword list, but an element in the list is not a keyword, " <>
                  "got: #{inspect(element)}"
```

michalmuskala:
Raising deep from within the map call might result in a  pretty poor error message with the whole printable stacktrace being only the `map` calls. I would suggest throwing from within the map, catching outside and re-raising with a stacktrace of the `Keyword.keys/1` directly

lib/elixir/lib/keyword.ex
eksperimental: Implement your suggestion with the addition of the article "an".

I am not 100% sure on the accuracy of the message though, as tuples don't have keys.
How about something like this?

> expected a keyword list, but an element in the list is not a two-element tuple with its first element being an atom; got:

lib/elixir/lib/keyword.ex
```diff
@@ -431,13 +431,41 @@ defmodule Keyword do
...
+            {Keyword, :keys, 1, _} -> false
+            {Keyword, _, _, _} -> true
+          end)
+
+        reraise ArgumentError,
+                "expected a keyword list, but an element in the list is not a two-element tuple with its first element being an atom; got: #{
+                  inspect(element)
+                }",
+                stacktrace
+    end
```
josevalim, line 468:
```suggestion
        raise ArgumentError,
              "expected a keyword list, but an element in the list is not a two-element tuple with an atom key as first element, " <>
                "got: #{inspect(element)}"
    end
```

lib/elixir/lib/keyword.ex
```diff
@@ -431,13 +431,41 @@ defmodule Keyword do
...
+      :lists.map(
+        fn
+          {key, _} when is_atom(key) ->
+            key
+
+          element ->
+            throw(element)
+        end,
+        keywords
+      )
```
josevalim, line 454:
```suggestion
      :lists.map(
        fn
          {key, _} when is_atom(key) -> key
          element -> throw(element)
        end,
        keywords
      )
```

lib/elixir/lib/keyword.ex
```diff
@@ -431,13 +431,30 @@ defmodule Keyword do
 
       iex> Keyword.keys(a: 1, b: 2)
       [:a, :b]
+
       iex> Keyword.keys(a: 1, b: 2, a: 3)
       [:a, :b, :a]
 
+      iex> Keyword.keys([{:a, 1}, {"b", 2}, {:c, 3}])
+      ** (ArgumentError) expected a keyword list, but an element in the list is not a two-element tuple with its first element being an atom; got: {"b", 2}
```
josevalim, line 439:
```suggestion
      ** (ArgumentError) expected a keyword list, but an element in the list is not a two-element tuple with an atom as its first element, got: {"b", 2}
```

lib/elixir/lib/keyword.ex
```diff
@@ -431,13 +431,30 @@ defmodule Keyword do
...
+        fn
+          {key, _} when is_atom(key) -> key
+          element -> throw(element)
+        end,
+        keywords
+      )
+    catch
+      element ->
+        raise ArgumentError,
+              "expected a keyword list, but an element in the list is not a two-element tuple with an atom key as first element, " <>
```
josevalim, line 455:
```suggestion
              "expected a keyword list, but an element in the list is not a two-element tuple with an atom as its first element, " <>
```

lib/elixir/lib/keyword.ex
```diff
@@ -431,13 +431,30 @@ defmodule Keyword do
...
+        fn
+          {key, _} when is_atom(key) -> key
+          element -> throw(element)
+        end,
+        keywords
+      )
+    catch
+      element ->
+        raise ArgumentError,
+              "expected a keyword list, but an element in the list is not a two-element tuple with an atom as its first element, " <>
```
josevalim, line 455:
```suggestion
              "expected a keyword list, but an entry in the list is not a two-element tuple with an atom as its first element, " <>
```

lib/elixir/lib/keyword.ex
```diff
@@ -431,13 +431,30 @@ defmodule Keyword do
 
       iex> Keyword.keys(a: 1, b: 2)
       [:a, :b]
+
       iex> Keyword.keys(a: 1, b: 2, a: 3)
       [:a, :b, :a]
 
+      iex> Keyword.keys([{:a, 1}, {"b", 2}, {:c, 3}])
+      ** (ArgumentError) expected a keyword list, but an element in the list is not a two-element tuple with an atom as its first element, got: {"b", 2}
```
josevalim, line 439:
```suggestion
      ** (ArgumentError) expected a keyword list, but an entry in the list is not a two-element tuple with an atom as its first element, got: {"b", 2}
```

josevalim:

PR #10215, opened 2020-07-21
Improve keyword list definition and introduce the concept of key

eksperimental:


lib/elixir/lib/keyword.ex
```diff
@@ -1,8 +1,10 @@
 defmodule Keyword do
   @moduledoc """
-  Keyword lists are lists of two-element tuples, where the first
-  element of the tuple is an atom and the second element can be any
-  value, used mostly to work with optional values.
+  A keyword list is a list that consists exclusively of two-element tuples.
+  The first element of these tuples is known as the _key_, and it must be an atom.
+  The second element, known as the _value_, can be any term.
```
josevalim, line 5:
```suggestion

  The first element of these tuples is known as the *key*, and it must be an atom.
  The second element, known as the *value*, can be any term.
```

josevalim:

eksperimental:
Awesome!

PR #9903, opened 2020-03-20
Document operators on the Operators page

glv:
Currently, the [Operators page](https://hexdocs.pm/elixir/1.10.2/operators.html) includes a precedence table, plus information about comparison operators and customization/overriding. It really should have documentation for *all* of the operators, or at least pointers to that documentation (especially since the [Syntax reference page](https://hexdocs.pm/elixir/1.10.2/syntax-reference.html#operators) points you there for a "full reference"). This change fixes that situation.

I'm an Elixir newbie, so I may be wrong about some of the details here. Please point out anything I got wrong!

(Also: I noticed this when I was trying to learn the actual semantics and allowable usage of `in` and `not in`, and it took me a long time to finally find their documentation under Kernel / Guards. So I also added short notes to `Enum.member?/2` and `Enumerable.member?/2` for other newcomers like me.)

PR #10216, opened 2020-07-21
Clarify use of Enum.member?/2 in in/2 operator

eksperimental:
Introduced in: 9a9b1f48bc5a63f364e3c5329ceccea7b67fd3b7 (#9903)
/cc @glv

### DIFF
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -162,8 +162,8 @@ defprotocol Enumerable do
   Otherwise it should return `{:error, __MODULE__}` and a default algorithm
   built on top of `reduce/3` that runs in linear time will be used.
 
-  The [`in`](`in/2`) and [`not in`](`in/2`) operators work by calling this
-  function.
+  When not called whithin guards, the [`in`](`in/2`) and [`not in`](`in/2`)
+  operators work by using this function.
   """
   @spec member?(t, term) :: {:ok, boolean} | {:error, module}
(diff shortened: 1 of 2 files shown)


glv:
Thanks for copying me on this! If the technicalities are different when used in a guard, I'm happy for that to be clarified.

However, my PR that's referenced here was actually prompted by a long struggle — as an Elixir newbie — to find clear documentation of the precise semantics of `in` and `not in` (what types they work with, etc). So I updated `operators.md` so that each operator clearly links to some other part of the documentation that provides more details about the semantics.

To that end, please don't remove the reference to `Enum.member?` in `operators.md` without replacing it with some other link to the details.

glv:
Oh, sorry … I see now that the link to `in/2` at the start of the line has all the info required. Carry on!

eksperimental:
> To that end, please don't remove the reference to `Enum.member?` in `operators.md` without replacing it with some other link to the details.

The reason why I deleted and didn't update it it is because it would be too much information for a list that aims to enumerate the operators.

eksperimental:
> Oh, sorry … I see now that the link to `in/2` at the start of the line has all the info required. Carry on!

yes. I was going to say that.

eksperimental:
@glv I have updated the descriptions. Please have a look

glv:
Love it, maintains consistency with the other elements of the list by having a short description. Thank for the change!

josevalim:

PR #10217, opened 2020-07-21
Improve Operators documentation

eksperimental:
- Use "less-than", "greater-than", and "equal to" operators. This the actual name used by Wikipedia and Unicode.
- Give a name to every opertor in Kernel. Mention it at the beggining of the first paragraph, so it appears in the summary.
- Minor improvements here and there.

### DIFF
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -1989,7 +1989,7 @@ defmodule Kernel do
   end
 
   @doc """
-  Regular expression or sub-string match operator. Matches the term on the `left`
+  Text-based match operator. Matches the term on the `left`
   against the regular expression or string on the `right`.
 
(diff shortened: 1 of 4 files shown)


lib/elixir/lib/kernel.ex
```diff
@@ -1989,11 +1989,12 @@ defmodule Kernel do
   end
 
   @doc """
-  Matches the term on the `left` against the regular expression or string on the
-  `right`.
+  Regular expression or sub-string match operator. Matches the term on the `left`
```
josevalim, line 1992: Text-based match operator.

josevalim:

lib/elixir/lib/kernel.ex
eksperimental: Thank you

PR #10219, opened 2020-07-22
Mention file:line in Already defined as ... error message

eksperimental:


lib/elixir/test/elixir/kernel/errors_test.exs
```diff
@@ -888,12 +888,14 @@ defmodule Kernel.ErrorsTest do
...
 
   test "def defmacro clause change" do
-    assert_eval_raise CompileError, "nofile:3: defmacro foo/1 already defined as def", '''
-    defmodule Kernel.ErrorsTest.DefDefmacroClauseChange do
-      def foo(1), do: 1
-      defmacro foo(x), do: x
-    end
-    '''
+    assert_eval_raise CompileError,
+                      "nofile:3: defmacro foo/1 already defined as def in nofile:3",
```
josevalim, line 892: You are using the wrong File:Line, given it is just being repeated in the error message. You would have to pick the other File:Line pairing.

eksperimental: You are right.  Fixed!

lib/elixir/src/elixir_def.erl
```diff
@@ -246,7 +246,7 @@ store_definition(Check, Kind, Meta, Name, Arity, File, Module, Defaults, Clauses
   {MaxDefaults, FirstMeta} =
     case ets:lookup(Set, {def, Tuple}) of
       [{_, StoredKind, StoredMeta, StoredFile, StoredCheck, {StoredDefaults, LastHasBody, LastDefaults}}] ->
-        check_valid_kind(Meta, File, Name, Arity, Kind, StoredKind),
+        check_valid_kind(Meta, File, Name, Arity, Kind, StoredKind, StoredMeta),
```
josevalim, line 249: You need to pass StoredFile too and use that in the error message too.

eksperimental: Sorry. Fixed

josevalim:

PR #10221, opened 2020-07-22
Update Map.update example to use full lambda syntax on first example

olivermt:
Hello.

I was training a new person today and he was very confused on the &(&...) syntax for updating in Map.update.

I have changed the first example to use the full lambda syntax so that it is more clear that the argument should be a regular lambda with an output and that the value is set to the return value of this function.

### DIFF
--- a/lib/elixir/lib/map.ex
+++ b/lib/elixir/lib/map.ex
@@ -604,7 +604,7 @@ defmodule Map do
 
   ## Examples
 
-      iex> Map.update(%{a: 1}, :a, 13, fn existing -> existing * 2 end)
+      iex> Map.update(%{a: 1}, :a, 13, fn existing_value -> existing_value * 2 end)
       %{a: 2}
       iex> Map.update(%{a: 1}, :b, 11, &(&1 * 2))
       %{a: 1, b: 11}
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/map.ex
```diff
@@ -604,7 +604,7 @@ defmodule Map do
 
   ## Examples
 
-      iex> Map.update(%{a: 1}, :a, 13, &(&1 * 2))
+      iex> Map.update(%{a: 1}, :a, 13, fn existing -> existing * 2 end)
```
eksperimental, line 607:
```suggestion
      iex> Map.update(%{a: 1}, :a, 13, fn value -> value * 2 end)
```

eksperimental: `value` reads better IMO

olivermt: How about `existing_value`? There is nothing wrong with being extra verbose imo.

eksperimental: Yes. But use `current_value`. That's what's being used in the example in `Map.get_and_update/3`

eksperimental: and I would replace the captures in the second example as well.

josevalim, line 607:
```suggestion
      iex> Map.update(%{a: 1}, :a, 13, fn existing_value -> existing_value * 2 end)
```

josevalim:

PR #10222, opened 2020-07-22
Improve Kernel documentation

eksperimental:


lib/elixir/lib/kernel.ex
```diff
@@ -1406,12 +1406,14 @@ defmodule Kernel do
   @doc guard: true
   @spec not true :: false
   @spec not false :: true
-  def not value do
-    :erlang.not(value)
+  def not term do
+    :erlang.not(term)
```
josevalim, line 1410: I am not sure I agree. I think `value` is, generally speaking, more welcoming to newcomers. `term` is more of an Erlang/Elixir thing. WDYT?

eksperimental: Personally I would leave `value`  for `key/value` and `term` is what is used in the Kernel module. `value` is only used here.

josevalim: Hrm, not sure I agree. :) For example, we say enums are collection of values and not collection of terms. The usage of value is more general than key-values. :)

eksperimental: I' m  not talking about generally speaking, but the terminology for the Enum module is collection of elements.

I am not saying value is wrong, or doesn't sound right. I was just trying to make it consistent. Same as with the other change that I replace `arg` with `term` because we never use arg in the module.

Either way is fine for me.

josevalim: > I' m not talking about generally speaking, but the terminology for the Enum module is collection of elements.

You are correct. :)

I definitely agree that both `term` and `value` are better than `arg`. But i recently heard about people being confused about what `term` means, so I thought maybe `value` could be better here.

eksperimental: reverted!

josevalim:

PR #10223, opened 2020-07-22
Make spec clearer for update functions in Map

eksperimental:
* It removes the concept of the "get" value (which should have been "gotten" IMO).
* Renames "initial" argument to "default"
* Makes use of the concept of "current value" which was already present in some functions, and uses this in the examples and the specs consistently

### DIFF
--- a/lib/elixir/lib/map.ex
+++ b/lib/elixir/lib/map.ex
@@ -599,28 +599,28 @@ defmodule Map do
 
   If `key` is present in `map` with value `value`, `fun` is invoked with
   argument `value` and its result is used as the new value of `key`. If `key` is
-  not present in `map`, `initial` is inserted as the value of `key`. The initial
+  not present in `map`, `default` is inserted as the value of `key`. The default
   value will not be passed through the update function.
 
   ## Examples
 
-      iex> Map.update(%{a: 1}, :a, 13, fn existing_value -> existing_value * 2 end)
+      iex> Map.update(%{a: 1}, :a, 13, fn current_value -> current_value * 2 end)
       %{a: 2}
-      iex> Map.update(%{a: 1}, :b, 11, &(&1 * 2))
+      iex> Map.update(%{a: 1}, :b, 11, fn current_value -> current_value * 2 end)
       %{a: 1, b: 11}
 
   """
-  @spec update(map, key, value, (value -> value)) :: map
-  def update(map, key, initial, fun) when is_function(fun, 1) do
+  @spec update(map, key, default :: value, (current_value :: value -> updated_value :: value)) :: map
+  def update(map, key, default, fun) when is_function(fun, 1) do
     case map do
       %{^key => value} ->
         put(map, key, fun.(value))
 
       %{} ->
-        put(map, key, initial)
+        put(map, key, default)
 
       other ->
-        :erlang.error({:badmap, other}, [map, key, initial, fun])
+        :erlang.error({:badmap, other}, [map, key, default, fun])
     end
   end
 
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/map.ex
```diff
@@ -599,28 +599,28 @@ defmodule Map do
 
   If `key` is present in `map` with value `value`, `fun` is invoked with
   argument `value` and its result is used as the new value of `key`. If `key` is
-  not present in `map`, `initial` is inserted as the value of `key`. The initial
+  not present in `map`, `default` is inserted as the value of `key`. The default
   value will not be passed through the update function.
```
josevalim, line 603: In my mind, `default` would be passed to the function, because that's usually what we expect from `default`? This is probably why `default` wasn't used in the first place?

lib/elixir/lib/map.ex
```diff
@@ -599,28 +599,28 @@ defmodule Map do
...
 
   ## Examples
 
-      iex> Map.update(%{a: 1}, :a, 13, fn existing_value -> existing_value * 2 end)
+      iex> Map.update(%{a: 1}, :a, 13, fn current_value -> current_value * 2 end)
       %{a: 2}
-      iex> Map.update(%{a: 1}, :b, 11, &(&1 * 2))
```
josevalim, line 609: I definitely prefer `existing_value`, as it makes it clear it won't be called if there is no value.

lib/elixir/lib/map.ex
```diff
@@ -845,7 +845,9 @@ defmodule Map do
       {nil, %{a: 1}}
 
   """
-  @spec get_and_update(map, key, (value -> {get, value} | :pop)) :: {get, map} when get: term
+  @spec get_and_update(map, key, (value -> {current_value, new_value :: value} | :pop)) ::
+          {current_value, map}
```
josevalim, line 849: :+1: for `current_value` here, as it is also called if the value is nil. I also like `new_value`!

lib/elixir/lib/map.ex
```diff
@@ -885,8 +887,9 @@ defmodule Map do
       {1, %{}}
 
   """
-  @spec get_and_update!(map, key, (value -> {get, value} | :pop)) :: {get, map}
-        when get: term
+  @spec get_and_update!(map, key, (value -> {current_value, new_value :: value} | :pop)) ::
+          {current_value, map}
+        when current_value: value
```
josevalim, line 892: Can we keep those two lines together? Or will the formatter complain?

lib/elixir/lib/map.ex
eksperimental: > In my mind, `default` would be passed to the function, because that's usually what we expect from `default`? This is probably why `default` wasn't used in the first place?

I interpret it the exact opposite way. I think of initial as a value that that will change to change,  such as in the initial value of the accumulator in Enum.reduce/2. Default I associate it with the default arg in Map.get, or even  Enum.find/3 uses default, and a function

lib/elixir/lib/map.ex
eksperimental: The formatter splits them

lib/elixir/lib/map.ex
eksperimental: True! they are really good names for stressing the distinction

eksperimental: corrected

lib/elixir/lib/map.ex
josevalim: I see. Good points! Note we may need to do similar changes in the Keyword module. :)

josevalim:

lib/elixir/lib/map.ex
eksperimental: Sure, I can take care of those changes as well.

eksperimental: > I see. Good points! Note we may need to do similar changes in the Keyword module. :)
done
https://github.com/elixir-lang/elixir/pull/10228

PR #10226, opened 2020-07-23
fix bug of only taking first app option in mix cmd

bruteforcecat:
currently there is a bug when we run `mix cmd --app foo --app bar pwd`, only the first app option will be used.

This PR fix this. This is my first PR to Elixir so please let me know if there is anything I missed. I have ran `make format` and `make test`


josevalim:

PR #10228, opened 2020-07-23
Make spec clearer for update functions in Keywords

eksperimental:
Inspired by c1a538e4a8bee7bc1c5be803534b8e4ee467c086 (#10223)

* Fixes the spec of Keyword.get_and_update!/3 that was missing the :pop return value for the update function
* It removes the concept of the "get" value and replaced it consistently with "current value"
* Renames "initial" argument to "default"
* Uses "existing value" when applicable


josevalim:

PR #10231, opened 2020-07-26
Fix ambiguity on Path.expand/1 docs

diogommartins:
I think it is appropriate to change the current example so as to avoid any misleading interpretation.


josevalim:

PR #10233, opened 2020-07-27
Rename "initial" argument to "default" in Dic and HashDict

eksperimental:
To be consistent with recently introduced changes: #10228, #10223


josevalim:

Issue #10232, opened 2020-07-27
Tests in `mix test --trace` are hard to trace


eksperimental:
The results by --trace were not clear as I couldn't immediately identify the test that was emitting a warning.

```
ExDoc.Formatter.HTMLTest
  ...
  * test generates extras without any other content (10.9ms)
  ...
```

The following points are noted:

**1.  the test file is omitted. Only the module is shown.**
I think we could show the source of the file next to the module.

```
ExDoc.Formatter.HTMLTest (test/ex_doc/formatter/html_test.exs)
```

It is hard to find the test when used inside a describe.
That's because.

**2. there is no line number**

**3. the test name is the joining of the title in describe and the title of the test, plus it is prefixed with the word test.**

I searched for "test generates extras without any other content", "generates extras without any other content",
"generates extras without any" and nothing turned up.

We could show some structured results, including line numbers 

## Tentative format

```
ExDoc.Formatter.HTMLTest  (test/ex_doc/formatter/html_test.exs)
  * generates extras [:298]
      - without api-reference [:416] (64.9ms)
```

another simpler solution would be to keep it flat as it is, and add the line number and source file.

```
ExDoc.Formatter.HTMLTest  (test/ex_doc/formatter/html_test.exs)
  * test generates extras without api-reference [:416] (64.9ms)
```

Eiji7:
@eksperimental I like "simpler solution", but there is still something missing …

I propose something like:
```elixir
ExDoc.Formatter.HTMLTest  (test/ex_doc/formatter/html_test.exs)
  * test: generates extras [line 298] without api-reference [line 416] (64.9ms)
```

or shorter version:
```elixir
ExDoc.Formatter.HTMLTest  (test/ex_doc/formatter/html_test.exs)
  * test: generates_extras:298 without_api_reference:416 (64.9ms)
```

In both cases `test` looks more like `prefix` with `:` character after it and there is line number as separator between `describe` name and `test` name. Also for me at first sight `[:416]` is not clear and may confuse beginners whereas `without_api_reference:416` looks more like `file_name:line` which I believe is well known format.

josevalim:
We no longer have the describe information but we can add line+file.

eksperimental:
> ```elixir
> ExDoc.Formatter.HTMLTest  (test/ex_doc/formatter/html_test.exs)
>   * test: generates extras [line 298] without api-reference [line 416] (64.9ms)
> ```

Loved this one.
[...]

PR #10234, opened 2020-07-27
Additional changes to ExUnit trace

eksperimental:
- Include file:line to slowest test results.
  Additionally it changes the order of the module, so it is consistent with the rest
  of the results.
- Print line number while test is being executed.
- Sort struct and spec fields alphabetically.

Related: fbfc8d2fd04c97c7fbfe7e35a482f76b67fbe3a4, #10232


josevalim:
:green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:I wanted to avoid

PR #10238, opened 2020-07-28
Fix Code.Typespec.spec_to_quoted/2 when union in  %{__struct__:} is used

eksperimental:
Originally reported in ExDoc.
https://github.com/elixir-lang/ex_doc/issues/1239

Code.Typespec.spec_to_quoted/2 wouldn't know how to deal with a union inside a
map with __struct__ key, such as in:

`@spec foo(%{__struct__: Foo | Bar}) :: any()`

/cc @ericmj @zdcthomas

### DIFF
--- a/lib/elixir/lib/code/typespec.ex
+++ b/lib/elixir/lib/code/typespec.ex
@@ -273,13 +273,13 @@ defmodule Code.Typespec do
           {{:optional, [], [typespec_to_quoted(k)]}, typespec_to_quoted(v)}
       end)
 
-    {struct, fields} = Keyword.pop(fields, :__struct__)
-    map = {:%{}, [line: line], fields}
+    case List.keytake(fields, :__struct__, 0) do
+      {{:__struct__, struct}, fields_pruned} when is_atom(struct) and struct != nil ->
+        map_pruned = {:%{}, [line: line], fields_pruned}
+        {:%, [line: line], [struct, map_pruned]}
 
-    if struct do
-      {:%, [line: line], [struct, map]}
-    else
-      map
+      _ ->
+        {:%{}, [line: line], fields}
     end
   end
 
--- a/lib/elixir/test/elixir/typespec_test.exs
+++ b/lib/elixir/test/elixir/typespec_test.exs
@@ -1183,6 +1183,69 @@ defmodule TypespecTest do
       end)
     end
 
+    test "spec_to_quoted with maps with __struct__ key" do
+      defmodule A do
+        defstruct [:key]
+      end
+
+      defmodule B do
+        defstruct [:key]
+      end
+
+      bytecode =
+        test_module do
+          @spec single_struct(%A{}) :: :ok
+          def single_struct(arg), do: {:ok, arg}
+
+          @spec single_struct_key(%{__struct__: A}) :: :ok
+          def single_struct_key(arg), do: {:ok, arg}
+
+          @spec single_struct_key_type(%{__struct__: atom()}) :: :ok
+          def single_struct_key_type(arg), do: {:ok, arg}
+
+          @spec union_struct(%A{} | %B{}) :: :ok
+          def union_struct(arg), do: {:ok, arg}
+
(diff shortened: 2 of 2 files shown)


lib/elixir/test/elixir/typespec_test.exs
```diff
@@ -1183,6 +1183,53 @@ defmodule TypespecTest do
...
+          def single_struct(arg), do: {:ok, arg}
+
+          @spec single_struct_key(%{__struct__: A}) :: {:ok}
+          def single_struct_key(arg), do: {:ok, arg}
+
+          @spec union_struct(%A{} | %B{}) :: {:ok}
+          def union_struct(arg), do: {:ok, arg}
+
+          @spec union_struct_key(%{__struct__: A | B}) :: {:ok}
+          def union_struct_key(arg), do: {:ok, arg}
```
eksperimental, line 1207: This is the only one that actually needs to be tested. But I wanted to make sure the other ones will pass so I wrote the test for the. Let me know if we don't want them all, and I will remove them
/cc @ericmj

lib/elixir/lib/code/typespec.ex
```diff
@@ -276,10 +276,20 @@ defmodule Code.Typespec do
     {struct, fields} = Keyword.pop(fields, :__struct__)
     map = {:%{}, [line: line], fields}
 
-    if struct do
-      {:%, [line: line], [struct, map]}
-    else
-      map
+    case struct do
```
josevalim, line 279: I think a better fix would be to check this:

    if is_atom(struct) and struct != nil do

After all, that's the only case we want to rewrite it to `%Foo{}`. If struct is anything else, we should probably keep its original format. For example, `%{__struct__: atom()}` should not be rewritten to `%atom(){...}` either. WDYT?

eksperimental: true.

eksperimental:
I have updated the PR. I removed the use of Keyword.pop, since `fields` is not a keyword list. Also, I guess it optimizes it for non-struct maps.

lib/elixir/lib/code/typespec.ex
```diff
@@ -273,13 +273,16 @@ defmodule Code.Typespec do
           {{:optional, [], [typespec_to_quoted(k)]}, typespec_to_quoted(v)}
       end)
 
-    {struct, fields} = Keyword.pop(fields, :__struct__)
-    map = {:%{}, [line: line], fields}
+    with {:__struct__, struct} <- List.keyfind(fields, :__struct__, 0),
```
josevalim, line 276: You can also use case+List.keytake. I think that will be cleaner:

```suggestion
    case List.keytake(fields, :__struct__, 0) do
      {{:__struct__, struct}, fields_pruned} when is_atom(struct) and struct != nil ->
        {:%, [line: line], [struct, fields_pruned]}

      _ ->
        {:%{}, [line: line], fields}
    end
```

lib/elixir/test/elixir/typespec_test.exs
```diff
@@ -1183,6 +1183,69 @@ defmodule TypespecTest do
...
+          @spec single_struct_key_type(%{__struct__: atom()}) :: {:ok}
+          def single_struct_key_type(arg), do: {:ok, arg}
+
+          @spec union_struct(%A{} | %B{}) :: {:ok}
+          def union_struct(arg), do: {:ok, arg}
+
+          @spec union_struct_key(%{__struct__: A | B}) :: {:ok}
+          def union_struct_key(arg), do: {:ok, arg}
+
+          @spec union_struct_key_type(%{__struct__: atom() | A | binary()}) :: {:ok}
```
josevalim, line 1212: Please convert these `{:ok}` into `:ok`. As far as Elixir is concerned, we never use `{:ok}`. :) Thanks!

eksperimental:
pushed!

josevalim:

eksperimental:
Thank you for your input

PR #10240, opened 2020-07-29
Add Map.update_lazy/4 and Keyword.update_lazy/4

marcelotto:
`Map.update/4` and `Keyword.update/4` seem to be the only functions in `Map` and `Keyword` which take a default value and don't have a lazy variant for cases where it is expensive to compute this value. Or is there a specific reason for not having a lazy variant in these cases?

### DIFF
--- a/lib/elixir/lib/keyword.ex
+++ b/lib/elixir/lib/keyword.ex
@@ -916,6 +916,49 @@ defmodule Keyword do
   end
 
   @doc """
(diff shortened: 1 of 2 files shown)


josevalim:
Can you please open up a discussion on the mailing list for general feedback? My initial thought is this is very specific and a combination of fetch+update will probably suit better. Thanks!

PR #10245, opened 2020-08-01
Make url parse more robust

ClarkThan:
- We can not guarantee someone use a strictly normalized url.
- Allow invalid escaped chars which prefixed by `%` like Python's urllib.parse.unquote.
```python
>>> from urllib.parse import unquote
>>> unquote('up%%2525increase%20rate')
'up%%25increase rate'
```
- So we should parse `"%{%20%7E"` to `"%{ ~"` other than raise error
```elixir
iex> URI.decode("up%%2525increase%20rate")
** (ArgumentError) malformed URI "up%%2525increase%20rate"
    (elixir 1.10.1) lib/uri.ex:344: URI.decode/1
```

### DIFF
--- a/lib/elixir/lib/uri.ex
+++ b/lib/elixir/lib/uri.ex
@@ -339,9 +339,6 @@ defmodule URI do
   @spec decode(binary) :: binary
   def decode(uri) do
     unpercent(uri, "", false)
-  catch
-    :malformed_uri ->
-      raise ArgumentError, "malformed URI #{inspect(uri)}"
   end
 
   @doc """
@@ -356,9 +353,6 @@ defmodule URI do
   @spec decode_www_form(binary) :: binary
   def decode_www_form(string) when is_binary(string) do
     unpercent(string, "", true)
-  catch
-    :malformed_uri ->
-      raise ArgumentError, "malformed URI #{inspect(string)}"
   end
 
   defp unpercent(<<?+, tail::binary>>, acc, spaces = true) do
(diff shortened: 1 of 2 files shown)


josevalim:
Hi @ClarkThan, I am not sure if we should accept this pull request.

A URI should be either already decoded or it needs to be decoded. We should never decode only certain characters in the URI. The RFC linked in the module documentation documents this very clearly and allowing something like "up%%2525increase%20rate" to become "up%%2525increase rate" most likely means we are doing double decoding.

ClarkThan:
> Hi @ClarkThan, I am not sure if we should accept this pull request.
> 
> A URI should be either already decoded or it needs to be decoded. We should never decode only certain characters in the URI. The RFC linked in the module documentation documents this very clearly and allowing something like "up%%2525increase%20rate" to become "up%%2525increase rate" most likely means we are doing double decoding.

Hi @josevalim . Yes, It does: RFC 3986 indeed mentioned that: "Implementations must not
   percent-encode or decode the same string more than once..." .  And in shortly,  my thought about this pull request is that Try to decode the percent prefixed chars as much as possible, if can't then regard it as a normal  `%` char and continue to check the next char.  Back to the specific example of `"up%%2525increase%20rate"`,  when it decoded to `"up%%2525increase rate"`,  in fact it doesn't happen double decode at all,  it's just decoded `%25` (4th - 6th chars)  and `%20` only ONE time. I hope I have explained clearly.

josevalim:
Correct. So we agree a string has only two possible states: decoded or not. Therefore, a string like "up%%2525increase%20rate" has only two possible scenarios:

1. It has already been decoded (i.e. it was originally %2520rate) and now we will decode it again, which is wrong. Therefore the best course of action is to return the string is.

2. The string is invalid. It is hard to say what is the best course of action here but I would err on keeping it as is.

So my point is that I can't think of a scenario where we should partially decode it. Or it is invalid or it is double decoding.

ClarkThan:
> Correct. So we agree a string has only two possible states: decoded or not. Therefore, a string like "up%%2525increase%20rate" has only two possible scenarios:
> 
> 1. It has already been decoded (i.e. it was originally %2520rate) and now we will decode it again, which is wrong. Therefore the best course of action is to return the string is.
> 2. The string is invalid. It is hard to say what is the best course of action here but I would err on keeping it as is.
> 
> So my point is that I can't think of a scenario where we should partially decode it. Or it is invalid or it is double decoding.

OK,  I got it.  Just like we always except: 
`URI.encode(URI.decode(query_string)) == query_string` is true

josevalim:
>  URI.encode(URI.decode(query_string)) == query_string is true

That's an even better summary. :) Closing then!

PR #10246, opened 2020-08-01
support rebar3 resource type git_subdir

tsloughter:
Related rebar3 PR which add git_subdir: https://github.com/erlang/rebar3/pull/2239

Hoping to get feedback while I figure out adding tests.

### DIFF
--- a/lib/mix/lib/mix/rebar.ex
+++ b/lib/mix/lib/mix/rebar.ex
@@ -152,18 +152,9 @@ defmodule Mix.Rebar do
     [scm, url | source] = Tuple.to_list(source)
 
     {scm, additional_opts} =
-      case scm do
-        :git_subdir ->
-          case source do
-            [_, sparse_dir | _] ->
-              {:git, [{:sparse, sparse_dir}]}
-
-            _ ->
-              {:git, []}
-          end
-
-        _ ->
-          {:git, []}
+      case {scm, source} do
+        {:git_subdir, [_, sparse_dir | _]} -> {:git, [sparse: sparse_dir]}
+        {_, _} -> {:git, []}
       end
 
     ref =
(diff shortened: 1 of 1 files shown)


lib/mix/lib/mix/rebar.ex
```diff
@@ -151,6 +151,21 @@ defmodule Mix.Rebar do
...
+            [_, sparse_dir | _] ->
+              {:git, [{:sparse, sparse_dir}]}
+
+            _ ->
+              {:git, []}
+          end
+
+        _ ->
+          {:git, []}
+      end
```
josevalim, line 167:
```suggestion
    {scm, additional_opts} =
      case {scm, source} do
        {:git_subdir, [_, sparse_dir | _]} -> {:git, [{:sparse, sparse_dir}]}
        {_, _} -> {:git, []}
      end
```

josevalim: Also, couldn't someone try to use git_subdir with branch/tag/ref? In this case, shouldn't this case also return the new source? Something like this:

```suggestion
    {scm, additional_opts, source} =
      case {scm, source} do
        {:git_subdir, [_, sparse_dir | source]} -> {:git, [sparse: sparse_dir], source}
        {_, _} -> {:git, [], source}
      end
```

?

tsloughter: It has to include a branch/tag/ref, that is why the head of the source is skipped and then parsed in the next `case`.

tsloughter: Format is: `{git_subdir, Url, BranchTagRef, Dir}`.

josevalim: ahhhhhhhhhhh, I see! So ignore my second suggestion, although I think my first suggestion is still good, right?

josevalim:

PR #10249, opened 2020-08-04
Allow passing "Any" as the exception in ExUnits "assert_raise"

PatrickSachs:
I recently had the need to only assert that a function I called crashed, but cannot pre-determine the exception type that will be raised.

As far as I have seen, ExUnit only allows hard coded exception types in `assert_raise`. For this reason I added a case to allow all types if the exception type is set to `Any`. (I chose `Any` because it is also used as the fallback in protocols, but this can easily be swapped out if preferred)

This change also has the nice side effect that it would now be possible to only test the exception message, regardless of its type for which I also added a doctest as an example.

### DIFF
--- a/lib/ex_unit/lib/ex_unit/assertions.ex
+++ b/lib/ex_unit/lib/ex_unit/assertions.ex
@@ -715,7 +715,8 @@ defmodule ExUnit.Assertions do
   @doc ~S"""
   Asserts the `exception` is raised during `function` execution with
   the expected `message`, which can be a `Regex` or an exact `String`.
-  Returns the rescued exception, fails otherwise.
+  Returns the rescued exception, fails otherwise. Use `Any` to match
+  all exceptions and only look at the message.
 
   ## Examples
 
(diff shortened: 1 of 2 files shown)


wojtekmach:
IMHO it's a very rare use case and so I'd do it explicitly in your test:

```elixir
try do
  something_that_raises()
  flunk("should not reach")
rescue
  e ->
    assert e.message == ...
end
```

If we were to support your use case, we should find a more extensible solution (and not hardcode to `Any`), perhaps:

```elixir
# additionally matches on exception fields:
assert_raise %CompileError{file: "foo.ex"}, fn -> ... end

# accepts any exception
assert_raise %{}, fn -> ... end
```

but yeah, I probably wouldn't do it after all. :-)

josevalim:
Hi @PatrickSachs! `Any` does not have such a general meeting in Elixir, besides protocol implementations, which are not really used here - so I would be worried about introducing said precedent here.

If you don't care about the exception type itself, you can always use `catch_error` + `Exception.message` to do manual assertions, or do like @wojtekmach suggested. Thanks!

PatrickSachs:
Thanks for the feedback, I'll go ahead and use that instead! Thank you for your time.

PR #10250, opened 2020-08-05
Add language to text, shell, shell-session, powershell blockcodes

eksperimental:
The language names are the ones supported by GitHub's Linguist project
https://github.com/github/linguist/blob/master/lib/linguist/languages.yml


josevalim:
Thanks @eksperimental ! Those names are not really supported by ExDoc, so I would rather keep them as is, which is at least cleaner in the source code. Especially given they are one or two lines at best.

PR #10252, opened 2020-08-06
Fix since doc for mix cmd task

jeroenvisser101:
Noticed this was already backported to 1.10.4, so if that was intentional, let's update the docs :)

(PS. Should I set the base to v1.10?)


josevalim:

PR #10254, opened 2020-08-10
Fix not being able to escape "|" inside a table (docstring)

marekzajac97:
Small issue when trying to include "|" character inside a table cell in docstring. There already is `String.replace("\\\|", "|")` inside `render_column/2` that I guess is supposed to convert escaped vertical bar back into "|" but `split_into_columns/2` does not check for escapes.

### DIFF
--- a/lib/elixir/test/elixir/io/ansi/docs_test.exs
+++ b/lib/elixir/test/elixir/io/ansi/docs_test.exs
@@ -553,12 +553,13 @@ defmodule IO.ANSI.DocsTest do
     test "table with escaped \"|\" inside cell" do
       table = "a | smth\\|smth_else | c\nd | e | f"
 
-      expected = """
-      a | smth|smth_else | c
-      d | e              | f
-      \e[0m
-      """
-      |> String.trim_trailing()
+      expected =
+        """
+        a | smth|smth_else | c
+        d | e              | f
+        \e[0m
+        """
+        |> String.trim_trailing()
 
       assert format_markdown(table) == expected
     end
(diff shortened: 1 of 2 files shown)


josevalim:
Thank you @marekzajac97! Can you please add a test of the table you are trying to render?

marekzajac97:
@josevalim Yes ofc. I added example table to the tests

lib/elixir/test/elixir/io/ansi/docs_test.exs
```diff
@@ -550,6 +550,20 @@ defmodule IO.ANSI.DocsTest do
       assert format_markdown("a | b | c\nd | e") == "a | b | c\nd | e |  \n\e[0m"
     end
 
+    test "table with escaped \"|\" inside cell" do
+      table = "a | smth\\|smth_else | c\nd | e | f"
```
eksperimental, line 554: Could please you format the table accordingly as it would be in the docs?

josevalim: The examples below use one-liners too, so this is fine. :)

josevalim:

lib/elixir/test/elixir/io/ansi/docs_test.exs
marekzajac97: Done

marekzajac97:
@josevalim May I ask you also to propagate this change to v1.10 branch as well please?

josevalim:
The next release will be 1.11-RC so it makes no difference to backport or not. :)

PR #10255, opened 2020-08-10
Add --warnings-as-errors option to mix test

wojtekmach:
Ref: https://github.com/elixir-lang/elixir/issues/6476#issuecomment-323284855

### DIFF
--- a/lib/mix/lib/mix/tasks/test.ex
+++ b/lib/mix/lib/mix/tasks/test.ex
@@ -34,44 +34,70 @@ defmodule Mix.Tasks.Test do
   ## Command line options
 
     * `--color` - enables color in the output
+
     * `--cover` - runs coverage tool. See "Coverage" section below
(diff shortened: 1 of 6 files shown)


wojtekmach:
Unfortunately we won't move forward with this as it's easy to break this behaviour with aliases. For example, when working with Ecto it's common to have this alias:

```
test: ["ecto.create --quiet", "ecto.migrate --quiet", "test"]
```

it happens so that `ecto.create` calls `compile` under the hood and so our project will be compiled _without_ `--warnings-as-errors` (even though the tests eventually _will_). This can be really confusing and unfortunately there doesn't seem to be anything we can do about it.

Fortunately there's a way to achieve the same thing today and that is run the following (e.g. on your CI):

```
- run: MIX_ENV=test mix do compile --warnings-as-errors, test
```

PR #10266, opened 2020-08-14
adds @since to the System module

supersimple:
System.os_time, System.schedulers, System.schedulers_online were added in 1.3.0, which i know because i tried to use them on an earlier system.


josevalim:

PR #10267, opened 2020-08-14
make Child Specification section of Supervisor documentation a bit more accurate

adrianomitre:
Nit picking, but why not?

Alternatives considered:
- The child specification is a map **which contains 2–6** elements.
- The child specification is a map **which contains 2 to 6** elements.
- The child specification is a map **which may contain up to** 6 elements.
- The child specification is a map **containing up to** 6 elements.

### DIFF
--- a/lib/elixir/lib/supervisor.ex
+++ b/lib/elixir/lib/supervisor.ex
@@ -108,7 +108,7 @@ defmodule Supervisor do
   The child specification describes how the supervisor starts, shuts down,
   and restarts child processes.
 
-  The child specification is a map which contains 6 elements. The first two keys
+  The child specification is a map with 2 to 6 elements. The first two keys
   in the following list are required, and the remaining ones are optional:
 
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/supervisor.ex
```diff
@@ -108,7 +108,7 @@ defmodule Supervisor do
   The child specification describes how the supervisor starts, shuts down,
   and restarts child processes.
 
-  The child specification is a map which contains 6 elements. The first two keys
+  The child specification is a map with 2 to 6 elements. The first two keys
```
josevalim, line 111:
```suggestion
  The child specification is a map containing up to 6 elements. The first two keys
```

josevalim:
Love it! I went with your last suggestion, since the next sentences break it down anyway. :green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

josevalim:

PR #10273, opened 2020-08-20
Use length instead of len in arguments and var names

eksperimental:


josevalim:

PR #10274, opened 2020-08-21
Raise CompileError when trying to define reserved types

### LINKED ISSUE #10263: Do not allow user to define `required/1` and `optional/1` types
Should be considered special cases and should error on compilation in the same way we we error when we try to define a built-in type for example.

As of now, they can be defined, but it is ambiguous.

```elixir
  @type required(arg) :: [arg]

  @type extra :: %{
          required(atom()) => term(),
        }
```

vnbrs:
This PR updates `Typespec.deftypespec/6` to raise a compilation error when trying to define reserved types such as `required/1` and `optional/1` with `@type`, `@typep` and `@opaque`.

Closes #10263

### DIFF
--- a/lib/elixir/lib/kernel/typespec.ex
+++ b/lib/elixir/lib/kernel/typespec.ex
@@ -131,11 +131,18 @@ defmodule Kernel.Typespec do
     store_typespec(bag, kind, expr, pos)
   end
 
+  @reserved_signatures [{:required, 1}, {:optional, 1}]
   def deftypespec(kind, expr, line, file, module, pos)
       when kind in [:type, :typep, :opaque] do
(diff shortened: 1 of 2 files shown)


lib/elixir/lib/kernel/typespec.ex
```diff
@@ -131,11 +131,18 @@ defmodule Kernel.Typespec do
     store_typespec(bag, kind, expr, pos)
   end
 
+  @reserved_signatures [{:required, 1}, {:optional, 1}]
```
josevalim, line 134:
```suggestion
  @reserved_signatures [required: 1, optional: 1]
```

josevalim:

PR #10283, opened 2020-08-25
Adds note: data structure in first position in Protocol

DianaOlympos:
Just got bitten by this.

I am not sure this is the best place to put it nor the best wording. I do think it deserve a note.

In a protocol, the data structure we are abstracting need to be the first argument in all the protocol functions. Could not find it written in the doc elsewhere

### DIFF
--- a/lib/elixir/lib/protocol.ex
+++ b/lib/elixir/lib/protocol.ex
@@ -47,6 +47,10 @@ defmodule Protocol do
   `size` information on lists, rather its value needs to be
   computed with `length`.
 
+  Note that the data structure you are implementing the protocol for
+  need to be in the first argument to the functions defined in the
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/protocol.ex
```diff
@@ -47,6 +47,10 @@ defmodule Protocol do
   `size` information on lists, rather its value needs to be
   computed with `length`.
 
+  Note that the data structure you are implementing the protocol for
+  need to be in the first argument to the functions defined in the
+  protocol.
```
josevalim, line 52:
```suggestion
  The data structure you are implementing the protocol for
  must be the first argument to all functions defined in the
  protocol.
```

josevalim:

PR #10285, opened 2020-08-26
Added newline to invalid_quoted_expr compilation error message

ruslanjan:
Added newline to invalid_quoted_expr compilation error message for better readability.

Spent hours because of not noticing first words of third sentence and seems like it should not be on the same line.


josevalim:
Thank you @ruslanjan! We don't add newlines to separate sentences in a paragraph because it may cause the opposite issue: which is where the text can be broken preemptively, especially on smaller screens.

So unfortunately there isn't much we can change here without also making it less than ideal for someone else. And given we don't generally do this in any other error messages, I prefer to keep it as is.

Once again, thanks!

PR #10287, opened 2020-08-27
Pass code location to :logger.macro_log/4

### LINKED ISSUE #10279: Logger usage causes dialyzer warnings
### Environment

```
$ elixir --version
Erlang/OTP 23 [erts-11.0.2] [source] [64-bit] [smp:6:6] [ds:6:6:10] [async-threads:1] [hipe]

Elixir 1.11.0-dev (82dca55) (compiled with Erlang/OTP 23)
```

### Current behavior

When using Logger with Elixir master, each Logger usage causes dialyzer warnings to be reported. For example:
```elixir
def hello do
  require Logger
  Logger.debug("here")
  :world
end
```

causes this warning to be reported on the `Logger.debug("here")` line:
```
The function call will not succeed.

Logger.__do_log__(_ :: any(), <<104, 101, 114, 101>>, %{
  :application => :dialyzer_logger_example,
  :file => [1..255, ...],
  :line => 17,
  :mfa => {DialyzerLoggerExample, :hello, 0}
})

will never return since it differs in arguments with
positions 2nd from the success typing arguments:

(any(), (() -> any()), map())
ElixirLS Dialyzer
```

### Expected behavior

Logger usage doesn't cause Dialyzer warnings to be reported.

### Notes

This is because Elixir uses Erlang's `:logger.macro_log` underlying implementation of its macro-based logger interface, in the `Logger.__do_log__` function [here](https://github.com/elixir-lang/elixir/blob/cf9ddd9194552aa7898e9f4b64717d578870fad9/lib/logger/lib/logger.ex#L896) without passing in proper location as first argument, as its is documented in `:logger.macro_log` [spec](https://github.com/erlang/otp/blob/49b65e66dce66ba88b70b2582a869ca6eb2fe00c/lib/kernel/src/logger.erl#L263).

I poked around and this `__do_log__` definition silences the warning, but I'm not sure if it's the correct fix:
```elixir
  def __do_log__(level, msg, metadata) when level in @levels and is_map(metadata) do
    location = %{mfa: metadata[:mfa], file: metadata[:file], line: metadata[:line]}
    if is_msg(msg) do
      :logger.macro_log(location, level, msg, add_elixir_domain(metadata))
```

michallepicki:
Closes #10279

Alternative to #10280

### DIFF
--- a/lib/logger/lib/logger.ex
+++ b/lib/logger/lib/logger.ex
@@ -881,7 +881,10 @@ defmodule Logger do
   defguardp is_msg(msg) when is_binary(msg) or is_list(msg) or is_map(msg)
 
   @doc false
-  def __do_log__(level, fun, metadata) when is_function(fun, 0) and is_map(metadata) do
+  def __do_log__(level, fun, metadata, location \\ %{})
+
+  def __do_log__(level, fun, metadata, location)
+      when is_function(fun, 0) and is_map(metadata) and is_map(location) do
     case fun.() do
       {msg, meta} ->
         __do_log__(level, msg, Enum.into(meta, metadata))
@@ -891,16 +894,17 @@ defmodule Logger do
     end
   end
 
-  def __do_log__(level, msg, metadata) when level in @levels and is_map(metadata) do
+  def __do_log__(level, msg, metadata, location)
+      when level in @levels and is_map(metadata) and is_map(location) do
     if is_msg(msg) do
-      :logger.macro_log(%{}, level, msg, add_elixir_domain(metadata))
+      :logger.macro_log(location, level, msg, add_elixir_domain(metadata))
     else
       # TODO: Remove this branch in Elixir v2.0
       IO.warn(
         "passing #{inspect(msg)} to Logger is deprecated, expected a map, a keyword list, a binary, or an iolist"
       )
 
-      :logger.macro_log(%{}, level, to_string(msg), add_elixir_domain(metadata))
+      :logger.macro_log(location, level, to_string(msg), add_elixir_domain(metadata))
     end
   end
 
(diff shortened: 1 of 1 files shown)


lib/logger/lib/logger.ex
```diff
@@ -990,15 +994,17 @@ defmodule Logger do
...
-
-          _ ->
-            []
-        end
+    [{:file, file} | _] = application_and_file = compile_time_application_and_file(caller)
+
+    {caller, location} =
+      case caller do
+        %{module: module, function: {fun, arity}, line: line} ->
+          {application_and_file ++ [mfa: {module, fun, arity}, line: line],
```
josevalim, line 1002: We shokd not include mfa and file and stuff in the caller if it is already on location. :)

michallepicki: Updated!

lib/logger/lib/logger.ex
```diff
@@ -990,24 +994,25 @@ defmodule Logger do
...
-            [mfa: {module, fun, arity}, line: line]
-
-          _ ->
-            []
-        end
+    [{:file, file} | maybe_application] = compile_time_application_and_file(caller)
+
+    location =
+      case caller do
+        %{module: module, function: {fun, arity}, line: line} ->
```
michallepicki, line 1001: I think the only information missing here that could make this pattern match to fail would be `function:` being `nil`. Maybe it would be better to always pass at least `file` and `line` in `location` even when `mfa` can't be constructed? Especially after `file` was removed from `quoted_metadata` in ecd2bcebc9b26459085574f7d0ce224fe36049c7

lib/logger/lib/logger.ex
```diff
@@ -881,7 +881,10 @@ defmodule Logger do
   defguardp is_msg(msg) when is_binary(msg) or is_list(msg) or is_map(msg)
 
   @doc false
-  def __do_log__(level, fun, metadata) when is_function(fun, 0) and is_map(metadata) do
+  def __do_log__(level, fun, metadata, location \\ %{})
```
josevalim, line 884: Can you make the location a required argument and pass it as first argument (or right before metadata)? It would reflect that it actually comes before metadata (i.e. lower precedence).

josevalim:

PR #10288, opened 2020-08-28
Change link in "Protocols" section in Kernel docs

schnittchen:
Previously linking to `defprotocol/2` which contains no useful information and links to `Protocol`. Save reader one click.


josevalim:

PR #10292, opened 2020-08-31
Display all values in logs

hauleth:
In pre-1.10 versions all values were inspected in default formatter (if there was no other formatting setting yet). Such change was unintended breaking change, and this should fix that to match "old" behaviour.

### DIFF
--- a/lib/logger/lib/logger/formatter.ex
+++ b/lib/logger/lib/logger/formatter.ex
@@ -182,6 +182,7 @@ defmodule Logger.Formatter do
 
   defp metadata(:time, _), do: nil
   defp metadata(:gl, _), do: nil
+  defp metadata(:report_cb, _), do: nil
 
   defp metadata(_, nil), do: nil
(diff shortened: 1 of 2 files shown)


josevalim:
This has already been fixed in master. Also note that we never simply inspected everything before so I would be skeptical about doing it now.  Thanks!

josevalim:
Ah, the exclusion of report_cb is a good call though, if you want to send a PR for that, please let me know, otherwise I can push this change too.

hauleth:
@josevalim ignoring explictly `report_cb` is pointless as these values will always fall into the `nil`-ing clause (function). So additional explicit check is IMHO unneeded.

josevalim:
@hauleth I was thinking it helps avoid going through the protocol lookup for something we know we will discard but not adding it also works!

hauleth:
@josevalim we could "filter out" all types that we know that do not have `String.Chars` implemented. Namely functions, ports, and tuples.

PR #10293, opened 2020-08-31
Correct title for Mix aliases section

eksperimental:


fertapric:
Good catch @eksperimental! 💜

PR #10294, opened 2020-09-01
Mention right associativity of the `--/2` operator in docs

tajchumber:
Just thought I would add a small mention to this as it has caused a few headaches to me ahah!

Happy to add more if needed or move it to a different place.

### DIFF
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -1384,6 +1384,11 @@ defmodule Kernel do
       iex> [1, 2, 3, 2, 1] -- [1, 2, 2]
       [3, 1]
 
+  It is important to note that `--/2` is right associative meaning:
+  
+      iex> [1, 2, 3] -- [2] -- [3]
+      [1, 3]
+
   """
   @spec list -- list :: list
   def left -- right do
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -1384,7 +1384,7 @@ defmodule Kernel do
       iex> [1, 2, 3, 2, 1] -- [1, 2, 2]
       [3, 1]
 
-  It is important to note that `--/2` is right associative meaning:
+  The `--/2` operator is right associative, meaning:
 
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/kernel.ex
```diff
@@ -1384,6 +1384,11 @@ defmodule Kernel do
       iex> [1, 2, 3, 2, 1] -- [1, 2, 2]
       [3, 1]
 
+  It is important to note that `--/2` is right associative meaning:
```
eksperimental, line 1387: I think to clarify the example it should also show how it is interpreted with parentheses.

lib/elixir/lib/kernel.ex
```diff
@@ -1384,6 +1384,11 @@ defmodule Kernel do
...
+
+      iex> [1, 2, 3] -- [2] -- [3]
+      [1, 3]
+
```
josevalim, line 1391:
```suggestion

  As it is equivalent to:

      iex> [1, 2, 3] -- ([2] -- [3])
      [1, 3]

```

eksperimental:
For completeness we should add the same to ++/2

josevalim:
It doesn't matter on `++` because the order of the operands do not matter there, at least in terms of correctness.

lib/elixir/lib/kernel.ex
```diff
@@ -1384,6 +1384,16 @@ defmodule Kernel do
       iex> [1, 2, 3, 2, 1] -- [1, 2, 2]
       [3, 1]
 
+  It is important to note that `--/2` is right associative meaning:
```
eksperimental, line 1387:
```suggestion
  It is important to note that the `--/2` operator has a *right to left* associativity:
```

josevalim: I personally find "right associative" clearer, which seems to be correct usage from the wikipedia page: https://en.wikipedia.org/wiki/Operator_associativity

josevalim:
```suggestion
The `--/2` operator is right associative, meaning:
```

PR #10295, opened 2020-09-02
Link to functions in Operators associativity table

eksperimental:


josevalim:
I don't think we should do this because most operators in Elixir have no meaning at the precedence. For example, `quote do: 1 + 2` does not mean we will use `Kernel.+`, therefore we should be able to talk about them without linking to particular implementations. Thank you!

eksperimental:
> `quote do: 1 + 2`

so what do we use there then?

PR #10296, opened 2020-09-02
Implement Map.take_in/2

samuelmanzanera:
Support of Map.take_in/2 to extract values using nested keys (recursively).

I think it may be useful when you only want a part of the map -- mostly when its deep nested map (i.e KV stores, DB engines, etc..) to reduce memory footprint with copy and message passing.

### DIFF
--- a/lib/elixir/lib/map.ex
+++ b/lib/elixir/lib/map.ex
@@ -433,6 +433,45 @@ defmodule Map do
   end
 
   @doc """
+  Returns a new map by extracting the key-value pairs in `map` where the key
+  is in `keys`. Keys can be nested to traverse recursively nested map
+
+  If `keys` contains keys that are not in `map`, they're simply ignored.
+
+  ## Examples
+
+      iex> Map.take_in(%{a: 1, b: %{c: 10, d: %{e: 5, f: 3}}}, [b: [:c, d: [:f]]])
+      %{b: %{c: 10, d: %{f: 3}}}
+
+  """
+  @spec take_in(map :: map(), keys :: [key | list]) :: map()
+  def take_in(map = %{}, []) when is_map(map), do: map
+
+  def take_in(map = %{}, keys) when is_map(map) and is_list(keys) do
+    Enum.reduce(map, %{}, fn {k, v}, acc ->
+      case v do
+        %{} ->
+          put(acc, k, take_in(v, Keyword.get(keys, k, [])))
+
+        _ ->
+          do_take_in(acc, map, k, keys)
+      end
+    end)
+  end
+
+  def take_in(non_map, _) do
+    :erlang.error({:badmap, non_map})
+  end
+
+  defp do_take_in(acc, map, key, subkeys) do
+    if key in subkeys do
+      put(acc, key, get(map, key))
+    else
+      acc
+    end
+  end
+
+  @doc """
   Gets the value for a specific `key` in `map`.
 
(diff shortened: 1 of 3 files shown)


whatyouhide:
Hey @samuelmanzanera, has this been proposed through the elixir-lang-core mailing list first? As we explain [on the website](https://elixir-lang.org/development.html), we reserve the issue tracker for bugs and we tend to not accept features that haven't been thoroughly discussed first. An immediate problem I see with this is that it would need to be a generic mechanism based on `Access` like `get_in` or `update_in` and not specific to maps. Thanks!

josevalim:
I initially like this proposal, so please send it through the mailing list. We also need to discuss details such as what happens if you ask to take from something that is not a map. The current implementation ignores them but it may be better to raise. We would also have to implement the same functionality for keywords.

My only concern is that this functionality can be quite inefficient compared to pattern matching. For example, instead of:

    iex> Map.take_in(%{a: 1, b: %{c: 10, d: %{e: 5, f: 3}}}, [b: [:c, d: [:f]]])
    %{b: %{c: 10, d: %{f: 3}}}

I could do:

    iex> %{b: %{c: c, d: %{f: f}}} = %{a: 1, b: %{c: 10, d: %{e: 5, f: 3}}}
    iex> %{b: %{c: c, d: %{f: f}}}
    
But I am not sure at the moment about how to avoid the duplication - but it is something worth keeping in mind.

josevalim:
Closing pending the ML discussion. :)

samuelmanzanera:
Thanks for you replies.

One thing the pattern matching cannot achieve (as far as I know) it's for an example like this one
```elixir
iex> %{a: a, b: %{d: d}} = %{a: "hello", b: %{c: "hi", d: "hola"}}
%{a: "hello", b: %{c: "hi", d: "hola"}}
```

Instead with Map.take_in
```elixir
iex> Map.take_in(%{a: "hello", b: %{c: "hi", d: "hola"}}, [:a, b: [:d]])
%{a: "hello", b: %{d: "hola"}}
```

The `c`is well excluded as the expected filtered keys.

Pattern matching of course can be more efficient but include some data not expected/desired.

> The current implementation ignores them but it may be better to raise
Actually the implementation raise if it's not a map but the Keyword implemention can be nice also

Integrate it inside Access may be prefered than duplicate Map and Keyword implementation. (Didn't think about it @whatyouhide)

I will include the proposition in the mailing list as asked. Thanks

josevalim:
> One thing the pattern matching cannot achieve (as far as I know) it's for an example like this one

You have to repeat the pattern after matching it:

    iex> %{a: a, b: %{d: d}} = %{a: "hello", b: %{c: "hi", d: "hola"}}
    iex> %{a: a, b: %{d: d}}
    %{a: "hello", b: %{d: "hola"}}

Which is an annoyance but something that we could try to tackle.

samuelmanzanera:
> > One thing the pattern matching cannot achieve (as far as I know) it's for an example like this one
> 
> You have to repeat the pattern after matching it:
> 
> ```
> iex> %{a: a, b: %{d: d}} = %{a: "hello", b: %{c: "hi", d: "hola"}}
> iex> %{a: a, b: %{d: d}}
> %{a: "hello", b: %{d: "hola"}}
> ```
> 
> Which is an annoyance but something that we could try to tackle.

Thanks sorry my bad!

PR #10299, opened 2020-09-02
add version scheme parameter to System.otp_release

adrianomitre:
This pull request addresses kind of the same problem addressed by https://github.com/elixir-lang/elixir/pull/8576 without compromising backward compatibility (i.e., no breaking changes).

Having this functionality on System would reduce redundancy (proliferation of copy-and-pasted solutions, including by proeminent Elixir projects such as [`hex`](https://github.com/hexpm/hex/blob/92f31922/lib/hex/utils.ex#L202) and [`dialyze`](https://github.com/fishcakez/dialyze/blob/6698ae58/lib/mix/tasks/dialyze.ex#L168)) and avoid maintainability issues in case an alteration has to be propagated (e.g., a bug is found or the specs change).

Furthermore, we have some indication that some people expect such functionality, which is somewhat analogous to `System.version/0`, to be there.

### DIFF
--- a/lib/elixir/lib/system.ex
+++ b/lib/elixir/lib/system.ex
@@ -1006,11 +1006,49 @@ defmodule System do
   end
 
   @doc """
-  Returns the Erlang/OTP release number.
+  Returns the Erlang/OTP release in the specified version scheme.
+  If none is specified, defaults to major version only.
   """
-  @spec otp_release :: String.t()
-  def otp_release do
-    :erlang.list_to_binary(:erlang.system_info(:otp_release))
+  @spec otp_release(:major_only | :otp_version_scheme | :semantic_versioning_scheme) :: String.t()
+  def otp_release(version_scheme \\ :major_only) do
+    case version_scheme do
+      :major_only ->
+        :erlang.list_to_binary(:erlang.system_info(:otp_release))
+
+      :otp_version_scheme ->
+        otp_version()
+
+      :semantic_versioning_scheme ->
+        normalize_to_semantic_versioning_scheme(otp_release(:otp_version_scheme))
+    end
+  end
+
+  # From https://github.com/hexpm/hex/blob/92f31922/lib/hex/utils.ex#L202
+  defp otp_version do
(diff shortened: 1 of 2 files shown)


wojtekmach:
> proliferation of copy-and-pasted solutions, including by proeminent Elixir projects such as hex_core

could you show the relevant part of hex_core that is problematic?

adrianomitre:
> > proliferation of copy-and-pasted solutions, including by proeminent Elixir projects such as hex_core
> 
> could you show the relevant part of hex_core that is problematic?

There is no part of hex_core that is problematic _per se_.

The following is a non-exhaustive list indicates proliferation of copy-and-pasted solutions:
1. [invention](https://github.com/fishcakez/dialyze/blob/6698ae582c77940ee10b4babe4adeff22f1b7779/lib/mix/tasks/dialyze.ex#L168)
2. [copy](https://github.com/hexpm/hex/blob/92f31922/lib/hex/utils.ex#L202)
3. [reinvention](https://github.com/mirego/elixir-boilerplate/blob/2e5170a2acaa8d0bd8d465c41bc7409e1accec95/lib/mix/tasks/erlang.check_version.ex#L23)

wojtekmach:
None of the links above points to hex_core (it’s a separate project from hex which is one of the links). I didn’t think we had such logic so was just double checking!

adrianomitre:
> None of the links above points to hex_core (it’s a separate project from hex which is one of the links). I didn’t think we had such logic so was just double checking!

Oh, I see. Description updated.

josevalim:
I understand why this is necessary but I would be much more comfortable if Erlang exposed an API to do so. Furthermore, Erlang does not follow semver nor its schema (it can have four parts for example), so we really shouldn't try to pretend it does.

But in a nutshell I would like to see this feature in OTP and then we can simply proxy it. Thanks!

adrianomitre:
For the record, I encapsulated the functionality of this pull request into a little lightweight library called [otp_version](https://hex.pm/packages/otp_version).

PR #10300, opened 2020-09-03
Include Erlang severity level to Logger metadata

henrythebuilder:
HI,
Elixir integrate new log levels (notice, critical, alert, and emergency) but this information "_is lost for 'compatibility with pre 1.10 backends'_"

In logger/handler.ex there is a 'TODO':
_# TODO: Remove this mapping once we remove old Logger Backends (v2.0)_

So in the meantime is possible to maintain the original level information almost in the metadata inserting an additional key/value `{:erl_level, real_level}` in order to be able to post-process/filter logs by the 'real severity level'.

### DIFF
--- a/lib/logger/lib/logger/handler.ex
+++ b/lib/logger/lib/logger/handler.ex
@@ -91,7 +91,11 @@ defmodule Logger.Handler do
 
           {message, %{gl: gl} = metadata} ->
             timestamp = Map.get_lazy(metadata, :time, fn -> :os.system_time(:microsecond) end)
-            metadata = erlang_metadata_to_elixir_metadata(metadata)
+
+            metadata =
+              erlang_metadata_to_elixir_metadata(metadata)
+              |> Keyword.put(:erl_level, erl_level)
+
             %{truncate: truncate, utc_log: utc_log?} = config
 
             event = {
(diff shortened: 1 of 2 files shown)


lib/logger/lib/logger/handler.ex
```diff
@@ -91,7 +91,11 @@ defmodule Logger.Handler do
 
           {message, %{gl: gl} = metadata} ->
             timestamp = Map.get_lazy(metadata, :time, fn -> :os.system_time(:microsecond) end)
-            metadata = erlang_metadata_to_elixir_metadata(metadata)
+
+            metadata =
+              erlang_metadata_to_elixir_metadata(metadata)
+              |> Keyword.put(:erl_level, erl_level)
```
josevalim, line 97:
```suggestion
              [erl_level: erl_level] ++ erlang_metadata_to_elixir_metadata(metadata)
```

henrythebuilder:
Thanks!

lib/logger/lib/logger/handler.ex
henrythebuilder: thanks for suggestion!

henrythebuilder: P.S. thanks again for suggestion but when I read it the merge was already done, thanks again and sorry for delay

josevalim: It was my bad! I usually apply the suggestion myself but I forgot! :D I will fix it locally, thanks!

PR #10302, opened 2020-09-05
Document atom comparison

axelson:
Currently how atoms are compared is not described in the Elixir docs, or at least (not where I would expect).

Sentence is copied from the erlang docs:
https://erlang.org/doc/reference_manual/expressions.html#term-comparisons


josevalim:

PR #10307, opened 2020-09-07
Enum.intersperse: Return early for empty enumerables

Sayanc93:
The codepath involving `reduce` and `reverse` seems like an overhead for empty Enumerables. This PR suggests to return early with empty list, skipping these functions.

### DIFF
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -1222,15 +1222,16 @@ defmodule Enum do
   """
   @spec intersperse(t, element) :: list
   def intersperse(enumerable, element) do
-    list =
-      enumerable
-      |> reduce([], fn x, acc -> [x, element | acc] end)
-      |> :lists.reverse()
+    if count(enumerable) > 0 do
+      # Head is a superfluous intersperser element
+      [_ | t] =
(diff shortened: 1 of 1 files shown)


eksperimental:
By using `count` you need to traverse the enumerable one more time, in the case where the size is not stored, such as in lists. Besides that reduce and reverse are optimized for empty lists.

lib/elixir/lib/enum.ex
```diff
@@ -1222,15 +1222,16 @@ defmodule Enum do
   """
   @spec intersperse(t, element) :: list
   def intersperse(enumerable, element) do
```
eksperimental, line 1224:
```suggestion
  def intersperse(enumerable, element)

  def intersperse([], _element), do: []
  def intersperse([head], _element), do: [head]

  def intersperse(enumerable, element) do

```

eksperimental (commented):
This is an optimization that could be implemented without doing additional traversing

josevalim:
As @eksperimental said, the `reduce` call is already optimized to return early, exactly so we don't need to add similar checks everywhere. We could check for one element list but it is probably not worth it. In any case, thank you!

PR #10312, opened 2020-09-09
Fix `&//2` parsing issue

### LINKED ISSUE #10311: Impossible to pass `&//2` as a parameter
### Environment

* Elixir & Erlang/OTP versions (elixir --version): `Elixir 1.10.1 (compiled with Erlang/OTP 22)`

### Current behavior

Parsing errors when `&//2` is followed by a comma:

```elixir
iex> apply(&//2, [1, 2]) 
** (CompileError) iex:3: undefined function apply/1
iex> {&//2, :op}
** (SyntaxError) iex:6: unexpected comma. Parentheses are required to solve ambiguity inside containers.

This error may happen when you forget a comma in a list or other container:

    [a, b c, d]

Or when you have ambiguous calls:

    [one, two three, four, five]

In the example above, we don't know if the parameters "four" and "five" belongs to the list or the function "two". You can solve this by explicitly adding parentheses:

    [one, two(three, four), five]

Elixir cannot compile otherwise. Syntax error before: ','
```

The following work though: 
```elixir
# with other operators
iex> apply(&*/2, [1, 2]) 
2
# when not followed by a comma
iex> {:op, &//2}  
{:op, &:erlang.//2}
iex> inspect(&//2)               
"&:erlang.//2"
# with brackets
iex> apply((&//2), [1, 2])
0.5
```

Interestingly, the formatter removes the brackets when I tried them as a workaround. I eventually managed to hack around by defining a constant: `@div &//2`.

### Expected behaviour

Ideally:
```elixir
iex> apply(&//2, [1, 2]) 
0.5
```

But if not working due to a legit reason / known limitation, it might good to at least make sure the formatter doesn't remove the needed bracket.

I think I've got a working PR, will try to open it.

sabiwara:
Closes #10311

I am not 100% sure I'm not introducing any side effect for alternative syntax, but seems to be fixing the issue without failing other tests.

Before/after the change:

```erlang
- [{identifier, {1, 1, nil}, '&'},
+ [{capture_op, {1, 1, nil}, '&'},
```

### DIFF
--- a/lib/elixir/src/elixir_tokenizer.erl
+++ b/lib/elixir/src/elixir_tokenizer.erl
@@ -715,14 +715,11 @@ handle_unary_op([$: | Rest], Line, Column, _Kind, Length, Op, Scope, Tokens) whe
   tokenize(Rest, Line, Column + Length + 1, Scope, [Token | Tokens]);
 
 handle_unary_op(Rest, Line, Column, Kind, Length, Op, Scope, Tokens) ->
-  case {strip_horizontal_space(Rest, 0), Tokens, Kind} of
-    {{[$/ | _] = Remaining, Extra}, [{capture_op, _, '&'} | _], _} ->
+  case {strip_horizontal_space(Rest, 0), Tokens} of
+    {{[$/ | _] = Remaining, Extra}, [{capture_op, _, '&'} | _]} ->
       Token = {identifier, {Line, Column, nil}, Op},
       tokenize(Remaining, Line, Column + Length + Extra, Scope, [Token | Tokens]);
-    {{[$/ | _] = Remaining, Extra}, _, dual_op} ->
-      Token = {identifier, {Line, Column, nil}, Op},
-      tokenize(Remaining, Line, Column + Length + Extra, Scope, [Token | Tokens]);
-    {{Remaining, Extra}, _, _} ->
+    {{Remaining, Extra}, _} ->
       Token = {Kind, {Line, Column, nil}, Op},
       tokenize(Remaining, Line, Column + Length + Extra, Scope, [Token | Tokens])
   end.
--- a/lib/elixir/src/elixir_tokenizer.erl
+++ b/lib/elixir/src/elixir_tokenizer.erl
@@ -715,15 +715,14 @@ handle_unary_op([$: | Rest], Line, Column, _Kind, Length, Op, Scope, Tokens) whe
   tokenize(Rest, Line, Column + Length + 1, Scope, [Token | Tokens]);
 
 handle_unary_op(Rest, Line, Column, Kind, Length, Op, Scope, Tokens) ->
-  Token = case {strip_horizontal_space(Rest, 0), Kind} of
-    {{[$/, $/ | _] = Remaining, Extra}, capture_op} ->
-      {capture_op, {Line, Column, nil}, Op};
-    {{[$/ | _] = Remaining, Extra}, _} ->
-      {identifier, {Line, Column, nil}, Op};
+  case {strip_horizontal_space(Rest, 0), Tokens} of
+    {{[$/ | _] = Remaining, Extra}, [{capture_op, _, '&'} | _]} ->
+      Token = {identifier, {Line, Column, nil}, Op},
+      tokenize(Remaining, Line, Column + Length + Extra, Scope, [Token | Tokens]);
     {{Remaining, Extra}, _} ->
-      {Kind, {Line, Column, nil}, Op}
-  end,
-  tokenize(Remaining, Line, Column + Length + Extra, Scope, [Token | Tokens]).
+      Token = {Kind, {Line, Column, nil}, Op},
+      tokenize(Remaining, Line, Column + Length + Extra, Scope, [Token | Tokens])
+  end.
 
 handle_op([$: | Rest], Line, Column, _Kind, Length, Op, Scope, Tokens) when ?is_space(hd(Rest)) ->
   Token = {kw_identifier, {Line, Column, nil}, Op},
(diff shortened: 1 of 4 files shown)


lib/elixir/src/elixir_tokenizer.erl
```diff
@@ -715,14 +715,15 @@ handle_unary_op([$: | Rest], Line, Column, _Kind, Length, Op, Scope, Tokens) whe
...
-      tokenize(Remaining, Line, Column + Length + Extra, Scope, [Token | Tokens])
-  end.
+  Token = case {strip_horizontal_space(Rest, 0), Kind} of
+    {{[$/, $/ | _] = Remaining, Extra}, capture_op} ->
+      {capture_op, {Line, Column, nil}, Op};
+    {{[$/ | _] = Remaining, Extra}, _} ->
+      {identifier, {Line, Column, nil}, Op};
+    {{Remaining, Extra}, _} ->
+      {Kind, {Line, Column, nil}, Op}
+  end,
```
josevalim, line 725:
```suggestion
  Token = case strip_horizontal_space(Rest, 0) of
    {[$/, $/ | _] = Remaining, Extra} when Kind == capture_op ->
      {capture_op, {Line, Column, nil}, Op};
    {[$/ | _] = Remaining, Extra} ->
      {identifier, {Line, Column, nil}, Op};
    {Remaining, Extra} ->
      {Kind, {Line, Column, nil}, Op}
  end,
```

lib/elixir/test/erlang/tokenizer_test.erl
```diff
@@ -209,7 +209,11 @@ capture_test() ->
    {unary_op, {1, 2, nil}, 'not'},
    {int, {1, 6, 1}, "1"},
    {',', {1, 7, 0}},
-   {int, {1, 9, 2}, "2"}] = tokenize("&not 1, 2").
+   {int, {1, 9, 2}, "2"}] = tokenize("&not 1, 2"),
+  [{capture_op, {1, 1, nil}, '&'},
+   {identifier, {1, 2, nil}, '/'},
+   {mult_op, {1, 3, nil}, '/'},
+   {int, {1, 4, 2}, "2"}] = tokenize("&//2").
```
josevalim, line 216: I think we also need to make `&/  /2` and `& / /2` work and the current patch does not cover those cases. I am not sure how to make it work though.

lib/elixir/src/elixir_tokenizer.erl
```diff
@@ -715,14 +715,15 @@ handle_unary_op([$: | Rest], Line, Column, _Kind, Length, Op, Scope, Tokens) whe
...
+  tokenize(Remaining, Line, Column + Length + Extra, Scope, [Token | Tokens]).
```
josevalim, line 726:
```suggestion
  case {strip_horizontal_space(Rest, 0), Tokens} of
    {{[$/ | _] = Remaining, Extra}, [{capture_op, _, '&'} | _]} ->
      Token = {identifier, {Line, Column, nil}, Op},
      tokenize(Remaining, Line, Column + Length + Extra, Scope, [Token | Tokens]);
    {{Remaining, Extra}, _} ->
      Token = {Kind, {Line, Column, nil}, Op},
      tokenize(Remaining, Line, Column + Length + Extra, Scope, [Token | Tokens])
  end.
```

lib/elixir/test/erlang/tokenizer_test.erl
josevalim: I think I was able to find a working implementation! Can you please try the suggestion in the comment above? We also need tests in this function for the following constructs: `&//2`, `& / /2`, `& //2`, `&/ /2`, and `& &/1`. Thank you!

lib/elixir/src/elixir_tokenizer.erl
sabiwara: Thanks a lot! :+1:

This works for all the tests you specified, however it seems to be breaking some [quote test doing `quote(do: + / 2)`](https://github.com/elixir-lang/elixir/blob/8770a61/lib/elixir/test/elixir/kernel/quote_test.exs#L267) though.
I can't figure if this is:
1. a legit case that got broken and needs to be addressed by an extra case
2. if this operation didn't make sense in the first place outside of a capture operator, and was relying on the current bug to work? (but maybe it is used by some metaprogramming somewhere?)

I went with 1. which seemed safer and tried a fix that I will push.

sabiwara:
Thanks a lot for the fix @josevalim +1

This works for all the tests you specified, however it seemed to be breaking some quote test doing quote(do: + / 2) though.
I can't figure if this:
1. is a legit case that got broken and needs to be addressed by an extra case
2. didn't make sense in the first place outside of a capture operator, and was relying on the current bug to work? (but maybe it is used by some metaprogramming somewhere?)

I went with 1 and pushed a tentative fix in https://github.com/elixir-lang/elixir/pull/10312/commits/88c2b5f155e2a1ed463b379616d4c89822a2755e#, please take another look :bow:

josevalim:
`+/2` doesn’t make sense. Probably missed the leading &, so we can fix it instead!

sabiwara:
Got it, will fix it this way :+1:

lib/elixir/test/elixir/kernel/quote_test.exs
```diff
@@ -263,11 +263,6 @@ defmodule Kernel.QuoteTest do
     assert quote(do: &1.foo) == quote(do: &1.foo)
   end
 
-  test "operators slash arity" do
```
sabiwara, line 266: Is it OK to simply remove the test?
We could just run `make format` to fix the build (with the new tokenizer it became `quote(do: &+/2)`), if it makes any sense to keep a test for this oc. I have no opinion, just asking for confirmation :slightly_smiling_face:

josevalim: Yes, it is ok. We do cover `& &/1` and `& +/2` elsewhere. That was always the syntax!

josevalim:

lib/elixir/test/elixir/kernel/quote_test.exs
sabiwara: Noted, thanks for clarifying! :heart:

PR #10315, opened 2020-09-10
Add offset to NaiveDateTime.to_iso8601/2 (making it to_iso8601/3).

JonRowe:
Hello!
As discussed over here https://groups.google.com/g/elixir-lang-core/c/JOGUFUyMm90 this adds the ability to provide an offset to `to_iso8601` to allow formatting a specified offset in the string, without having additional timezone information.

I added this to `NaiveDateTime` in the end to make it clear it has no time zone significance, adding it to `DateTime` is more complicated due to the need to deal with potentially changing timezones or raising when used with a non utc timzone, I thus came to the conclusion that `NaiveDateTime` was a better home for the extended functionality.

As a consequence it can be used to override a `DateTime`s natural timezone which I guess I could avoid with additional pattern matching (and/or an error?) additionally if desired this could be made to have parity between the two `to_iso8601` function heads but that would raise the question of how to handle the potential change in timezone in `DateTime`

### DIFF
--- a/lib/elixir/lib/calendar/datetime.ex
+++ b/lib/elixir/lib/calendar/datetime.ex
@@ -899,6 +899,7 @@ defmodule DateTime do
 
   Only supports converting datetimes which are in the ISO calendar,
   attempting to convert datetimes from other calendars will raise.
+  You can also optionally specify an offset for the formatted string.
 
   WARNING: the ISO 8601 datetime format does not contain the time zone nor
   its abbreviation, which means information is lost when converting to such
@@ -930,11 +931,31 @@ defmodule DateTime do
       iex> DateTime.to_iso8601(dt, :basic)
       "20000229T230007-0400"
 
-  """
-  @spec to_iso8601(Calendar.datetime(), :extended | :basic) :: String.t()
-  def to_iso8601(datetime, format \\ :extended)
+      iex> dt = %DateTime{year: 2000, month: 2, day: 29, zone_abbr: "AMT",
+      ...>                hour: 23, minute: 0, second: 7, microsecond: {0, 0},
+      ...>                utc_offset: -14400, std_offset: 0, time_zone: "America/Manaus"}
+      iex> DateTime.to_iso8601(dt, :extended, 3600)
+      "2000-03-01T04:00:07+01:00"
 
-  def to_iso8601(%{calendar: Calendar.ISO} = datetime, format)
+      iex> dt = %DateTime{year: 2000, month: 2, day: 29, zone_abbr: "AMT",
+      ...>                hour: 23, minute: 0, second: 7, microsecond: {0, 0},
+      ...>                utc_offset: -14400, std_offset: 0, time_zone: "America/Manaus"}
+      iex> DateTime.to_iso8601(dt, :extended, 0)
+      "2000-03-01T03:00:07+00:00"
+
+      iex> dt = %DateTime{year: 2000, month: 3, day: 01, zone_abbr: "UTC",
+      ...>                hour: 03, minute: 0, second: 7, microsecond: {0, 0},
+      ...>                utc_offset: 0, std_offset: 0, time_zone: "Etc/UTC"}
+      iex> DateTime.to_iso8601(dt, :extended, 0)
+      "2000-03-01T03:00:07Z"
+
+      iex> {:ok, dt, offset} = DateTime.from_iso8601("2000-03-01T03:00:07Z")
+      iex> "2000-03-01T03:00:07Z" = DateTime.to_iso8601(dt, :extended, offset)
+  """
+  @spec to_iso8601(Calendar.datetime(), :basic | :extended, nil | integer()) :: String.t()
+  def to_iso8601(datetime, format \\ :extended, offset \\ nil)
+
+  def to_iso8601(%{calendar: Calendar.ISO} = datetime, format, nil)
       when format in [:extended, :basic] do
     %{
       year: year,
@@ -949,16 +970,51 @@ defmodule DateTime do
       std_offset: std_offset
     } = datetime
 
-    Calendar.ISO.date_to_string(year, month, day, format) <>
-      "T" <>
-      Calendar.ISO.time_to_string(hour, minute, second, microsecond, format) <>
+    datetime_to_string(year, month, day, hour, minute, second, microsecond, format) <>
       Calendar.ISO.offset_to_string(utc_offset, std_offset, time_zone, format)
   end
 
-  def to_iso8601(%{calendar: _} = datetime, format) when format in [:extended, :basic] do
+  def to_iso8601(
+        %{calendar: Calendar.ISO, microsecond: {_, precision}, time_zone: "Etc/UTC"} = datetime,
+        format,
+        0
+      )
+      when format in [:extended, :basic] do
+    {year, month, day, hour, minute, second, {microsecond, _}} = shift_by_offset(datetime, 0)
+
+    datetime_to_string(year, month, day, hour, minute, second, {microsecond, precision}, format) <>
+      "Z"
+  end
+
+  def to_iso8601(%{calendar: Calendar.ISO} = datetime, format, offset)
+      when format in [:extended, :basic] do
+    {_, precision} = datetime.microsecond
+    {year, month, day, hour, minute, second, {microsecond, _}} = shift_by_offset(datetime, offset)
+
+    datetime_to_string(year, month, day, hour, minute, second, {microsecond, precision}, format) <>
+      Calendar.ISO.offset_to_string(offset, 0, nil, format)
+  end
+
+  def to_iso8601(%{calendar: _} = datetime, format, offset) when format in [:extended, :basic] do
     datetime
     |> convert!(Calendar.ISO)
-    |> to_iso8601(format)
+    |> to_iso8601(format, offset)
+  end
+
+  defp shift_by_offset(%{calendar: calendar} = datetime, offset) do
+    total_offset = datetime.utc_offset + datetime.std_offset
+
+    datetime
+    |> to_iso_days()
+    # Subtract total original offset in order to get UTC and add the new offset
+    |> Calendar.ISO.add_day_fraction_to_iso_days(offset - total_offset, 86400)
+    |> calendar.naive_datetime_from_iso_days()
+  end
+
+  defp datetime_to_string(year, month, day, hour, minute, second, microsecond, format) do
+    Calendar.ISO.date_to_string(year, month, day, format) <>
+      "T" <>
+      Calendar.ISO.time_to_string(hour, minute, second, microsecond, format)
   end
 
   @doc """


josevalim:
Sorry @JonRowe, I think there was some misunderstanding. I think we should add the extra argument to `DateTime.to_iso8691` since that's the one that returns the offset. The positive news is that it should be a simpler change there as it already handles offsets. Please let me know what you think!

josevalim:
Errr, please ignore my comment. I reviewed the code before reading your description. :) I will follow up shortly.

lib/elixir/lib/calendar/naive_datetime.ex
```diff
@@ -720,23 +721,41 @@ defmodule NaiveDateTime do
...
+      ...>                hour: 23, minute: 0, second: 7, microsecond: {0, 0},
+      ...>                utc_offset: 3600, std_offset: 0, time_zone: "Europe/Warsaw"}
+      iex> NaiveDateTime.to_iso8601(dt, :extended, 7200)
+      "2000-02-29T23:00:07+02:00"
+
   """
-  @spec to_iso8601(Calendar.naive_datetime(), :basic | :extended) :: String.t()
-  def to_iso8601(naive_datetime, format \\ :extended)
+  @spec to_iso8601(Calendar.naive_datetime(), :basic | :extended, nil | integer()) :: String.t()
+  def to_iso8601(naive_datetime, format \\ :extended, offset \\ nil)
```
josevalim, line 756: I think the default offset can be zero.

JonRowe: I considered that, but I didn't want to change the default behaviour of printing nothing by default, a `0` offset should print as `Z`.

lib/elixir/lib/calendar/naive_datetime.ex
```diff
@@ -720,23 +721,41 @@ defmodule NaiveDateTime do
       iex> NaiveDateTime.to_iso8601(~N[2000-02-28 23:00:13.001])
       "2000-02-28T23:00:13.001"
 
+      iex> NaiveDateTime.to_iso8601(~N[2000-02-28 23:00:13.001], :extended, 0)
+      "2000-02-28T23:00:13.001Z"
```
josevalim, line 725: It cannot return "Z" because Z implies Etc/UTC and we have no idea if this naive datetime is actually in UTC. I think it should simply show nothing (i.e. same as above).

josevalim:
@JonRowe ok, I have fully read your description now, apologies. In my mind, if we were to add this to DateTime, we would add the given offset to the DateTime's offset, but I can see this being confusing as adding to the offset effectively change its meaning. So I like your direction! Can you please do the changes above? Thanks!

lib/elixir/lib/calendar/naive_datetime.ex
josevalim: We should not print zero as Z as that requires Etc/UTC, which we can't assume here.

josevalim: Unless we want to show zero as "+00:00" and not offset as nothing. But I don't know if that's a thing. Thoughts?

josevalim: Apparently it is a thing. For example, a -01:00 timezone on summer time. So I think we indeed need to treat `nil` and `0` separately. We just need to make sure to show zero as "+00:00" instead of Z. Perhaps I would even make it two distinct function signatures: `to_iso8601/2` and `to_iso8601/3`.

JonRowe: Ok I can definitely seperate it to two seperate functions, and I'm sort of ok with `+00:00` but this reintroduces a disparity, a time parsed from `Z` will be treated as UTC but theres no "one way" of parsing it back to `Z`, e.g. the end user has to care about timezones when they don't have any timezone information.

My aim is to make a reversible set of function calls such that given a string you can get back to that string without involving a timezone lookup... Not being able to set `Z` from input causes a problem for that...

lib/elixir/lib/calendar/naive_datetime.ex
JonRowe: It needs to be able to return `Z` for reversibility, plus in ISO8601 `Z` is the same as `+00:00` (as it has no timezone information).

JonRowe:
I've replied to your feedback @josevalim, there might be some re-interpretation as a compromise between our goals :joy: but it'll be next week before I get back to this, thanks for your time so far!

josevalim:
@kipcole9 can you please confirm if Z implies UTC or just +00:00?

josevalim:
So I have verified and indeed Z is from Zulu which implies UTC. Given your counter-arguments, I suggest for us to go with my original proposal of adding this to DateTime.to_iso8601 and say the given offset is added to the DateTime offset.

josevalim:
Ping. :) Just making sure nothing fell through the cracks here!

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -930,11 +931,28 @@ defmodule DateTime do
...
+      ...>                hour: 23, minute: 0, second: 7, microsecond: {0, 0},
+      ...>                utc_offset: -14400, std_offset: 0, time_zone: "America/Manaus"}
+      iex> DateTime.to_iso8601(dt, :extended, 0)
+      "2000-03-01T03:00:07+00:00"
+
+      iex> dt = %DateTime{year: 2000, month: 2, day: 29, zone_abbr: "AMT",
+      ...>                hour: 23, minute: 0, second: 7, microsecond: {0, 0},
+      ...>                utc_offset: -14400, std_offset: 0, time_zone: "America/Manaus"}
+      iex> DateTime.to_iso8601(dt, :extended, "Z")
+      "2000-03-01T03:00:07Z"
```
JonRowe, line 950: This is not quite reversible as `from_iso8601` drops this information, but I can't think of a better way to disambiguate between `0` and `Z`, personally, I would just default `0` to `Z` given that would be reversible for the most common cases, but I believe this will be your preference?

josevalim: My proposal is to not allow “Z” as argument here. Instead, we will only show Z if the offset is zero and the time zone is Etc/UTC. What do you think?

josevalim: If we do this, we can default the offset to zero.

josevalim: Also, to be clear, we should add the given offset into the current offset, instead of discarding the current one. :)

JonRowe: I take the current offset into account to get to utc if the datetime is not utc, before converting to the new offset if thats what you mean? Only if the offset is 0 and the timezone is utc do I output `Z` now, did you mean something else? The test below here I also added as it shows my intented use for the library (that is to say, reversibility).

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -930,11 +931,31 @@ defmodule DateTime do
...
+      ...>                hour: 23, minute: 0, second: 7, microsecond: {0, 0},
+      ...>                utc_offset: -14400, std_offset: 0, time_zone: "America/Manaus"}
+      iex> DateTime.to_iso8601(dt, :extended, 3600)
+      "2000-03-01T04:00:07+01:00"
+
+      iex> dt = %DateTime{year: 2000, month: 2, day: 29, zone_abbr: "AMT",
+      ...>                hour: 23, minute: 0, second: 7, microsecond: {0, 0},
+      ...>                utc_offset: -14400, std_offset: 0, time_zone: "America/Manaus"}
+      iex> DateTime.to_iso8601(dt, :extended, 0)
+      "2000-03-01T03:00:07+00:00"
```
josevalim, line 944: So my comment was about this test. :) The `offset` here shouldn't be zero, it should be -14400.

lib/elixir/lib/calendar/datetime.ex
```diff
@@ -930,11 +931,31 @@ defmodule DateTime do
...
+      iex> DateTime.to_iso8601(dt, :extended, 0)
+      "2000-03-01T03:00:07Z"
+
+      iex> {:ok, dt, offset} = DateTime.from_iso8601("2000-03-01T03:00:07Z")
+      iex> "2000-03-01T03:00:07Z" = DateTime.to_iso8601(dt, :extended, offset)
   """
-  @spec to_iso8601(Calendar.datetime(), :extended | :basic) :: String.t()
-  def to_iso8601(datetime, format \\ :extended)
+  @spec to_iso8601(Calendar.datetime(), :basic | :extended, nil | integer()) :: String.t()
+  def to_iso8601(datetime, format \\ :extended, offset \\ nil)
```
josevalim, line 956: Can the default be zero?

lib/elixir/lib/calendar/datetime.ex
josevalim: To be clear, I was thinking the logic would be largely the logic we have today. The only difference is that we will add the current offset, something like this:

```diff
diff --git a/lib/elixir/lib/calendar/datetime.ex b/lib/elixir/lib/calendar/datetime.ex
index 512710635..9c2fb0c5a 100644
--- a/lib/elixir/lib/calendar/datetime.ex
+++ b/lib/elixir/lib/calendar/datetime.ex
@@ -932,9 +932,9 @@ defmodule DateTime do
 
   """
   @spec to_iso8601(Calendar.datetime(), :extended | :basic) :: String.t()
-  def to_iso8601(datetime, format \\ :extended)
+  def to_iso8601(datetime, format \\ :extended, offset \\ 0)
 
-  def to_iso8601(%{calendar: Calendar.ISO} = datetime, format)
+  def to_iso8601(%{calendar: Calendar.ISO} = datetime, format, ofset)
       when format in [:extended, :basic] do
     %{
       year: year,
@@ -952,13 +952,13 @@ defmodule DateTime do
     Calendar.ISO.date_to_string(year, month, day, format) <>
       "T" <>
       Calendar.ISO.time_to_string(hour, minute, second, microsecond, format) <>
-      Calendar.ISO.offset_to_string(utc_offset, std_offset, time_zone, format)
+      Calendar.ISO.offset_to_string(utc_offset + offset, std_offset, time_zone, format)
   end
 
-  def to_iso8601(%{calendar: _} = datetime, format) when format in [:extended, :basic] do
+  def to_iso8601(%{calendar: _} = datetime, format, offset) when format in [:extended, :basic] do
     datetime
     |> convert!(Calendar.ISO)
-    |> to_iso8601(format)
+    |> to_iso8601(format, offset)
   end
 
   @doc """
```

Is there a reason why this approach won't work? :)

josevalim:
Ping. :)

lib/elixir/lib/calendar/datetime.ex
JonRowe: When parsing the offset is from UTC, so when applied the offset should be from UTC. This makes the most sense to me  and is the least surprising to me. Changing a timezone for the date should not affect output when I supply a specific offset, as the offset is not the timezone.

lib/elixir/lib/calendar/datetime.ex
JonRowe: It's not zero so the function can tell between "override using this offset" and "use the offset for the timezone supplied".

josevalim:
@JonRowe so our discussion is basically if the given offset is an overriding or an addition. The addition is simpler at least in terms of implementation but other than that I am not sure about new cons. I would also say the addition is simpler conceptually but that's a personal argument. Why should overriding be preferred? Can you think of any pros/cons?

In any case, you can always transform the overriding into an addition or an addition into an overriding though.

josevalim:
Hi @JonRowe, just to make sure we are on the same page, the reason I think the adding the offset is better is because of this scenario:

1. Imagine you parse the timezone as UTC + offset
2. Then you convert it to -02:00 for some reason
3. Now you want to format it but you need to bring back the original offset

If we pass the original offset and it adds, then everything works even if you change the underlying offset. However, if it overrides, I will need to pass the offset like this to recover the actual value: `offset + datetime.utc_offset + datetime.std_offset`.

JonRowe:
> If we pass the original offset and it adds, then everything works even if you change the underlying offset. However, if it overrides, I will need to pass the offset like this to recover the actual value: `offset + datetime.utc_offset + datetime.std_offset`.

I'm interested in reversibility, when parsing data from an external source I capture the offset, do things, apply the offset to reply. The offset is a formatting issue in and out.

Under your suggestion I now have to work out the delta from the original offset, capture that as well, and reapply that when formatting.

JonRowe:
Offsets in this sense are purely for formatting as Elixir doesn't take them into account to produce date time, so they should override.

josevalim:
@JonRowe but in your scenario, if you simply discard the current offset, you will show the wrong time! You'd need to translate the hours under the current offset to the original offset!

JonRowe:
> @JonRowe but in your scenario, if you simply discard the current offset, you will show the wrong time! You'd need to translate the hours under the current offset to the original offset!

ISO8601 offsets reflect time off UTC. Elixir's offsets don't change UTC values. The offset can safely be re-applied to the UTC value to get the desired output.

JonRowe:
I'm happy for you to take this over or close it, unless this adds reversibility so I can return the original string I will need to maintain my work around anyway.

JonRowe:
(Sorry I have a heap of RSpec things to do at the moment, so rewriting a contribution to do something other than what I need it to do is not at the top of my priority list, I appreciate your views may differ from mine as to this functions purpose, and I thank you for your patience and time ❤️ 💚 🧡 💛 💜 )

josevalim:
@JonRowe take this example:

```
     iex> dt = %DateTime{year: 2000, month: 2, day: 29, zone_abbr: "AMT",
    ...>                hour: 23, minute: 0, second: 7, microsecond: {0, 0},
    ...>                utc_offset: -14400, std_offset: 0, time_zone: "America/Manaus"}
    iex> DateTime.to_iso8601(dt, :extended)
    "2000-02-29T23:00:07-04:00"
```

If you called it with an offset of +03:00 and it simply returned `"2000-02-29T23:00:07+03:00"`, then it would be the wrong time! So if we want to ignore the current offset, we need to readjust the whole datetime. Unless this is what this PR is doing and I completely missed it?

josevalim:
> Unless this is what this PR is doing and I completely missed it?

Apparently that seems to be the case. :D

josevalim:

JonRowe:
> So if we want to ignore the current offset, we need to readjust the whole datetime. Unless this is what this PR is doing and I completely missed it?

Yes it goes back to UTC then applies the new offset.

PR #10317, opened 2020-09-11
Added ignore module option to coverage report

danicuki:
I would love to also add a test case for that, but I really did not find a way to insert project options in the test cases.

### DIFF
--- a/lib/mix/lib/mix/tasks/test.coverage.ex
+++ b/lib/mix/lib/mix/tasks/test.coverage.ex
@@ -169,10 +169,11 @@ defmodule Mix.Tasks.Test.Coverage do
 
   defp generate_cover_results(opts) do
     {:result, ok, _fail} = :cover.analyse(:coverage, :line)
+    ignore = opts[:ignore_modules] || []
 
     modules =
       :cover.modules()
-      |> Enum.filter(&(!Enum.member?(opts[:ignore] || [], &1)))
+      |> Enum.reject(& &1 in ignore)
 
     if summary_opts = Keyword.get(opts, :summary, true) do
       summary(ok, modules, summary_opts)
--- a/lib/mix/lib/mix/tasks/test.coverage.ex
+++ b/lib/mix/lib/mix/tasks/test.coverage.ex
@@ -169,7 +169,10 @@ defmodule Mix.Tasks.Test.Coverage do
 
   defp generate_cover_results(opts) do
     {:result, ok, _fail} = :cover.analyse(:coverage, :line)
-    modules = :cover.modules()
+
+    modules =
+      :cover.modules()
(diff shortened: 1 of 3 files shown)


lib/mix/lib/mix/tasks/test.coverage.ex
```diff
@@ -169,7 +169,10 @@ defmodule Mix.Tasks.Test.Coverage do
 
   defp generate_cover_results(opts) do
     {:result, ok, _fail} = :cover.analyse(:coverage, :line)
-    modules = :cover.modules()
+
+    modules =
+      :cover.modules()
+      |> Enum.filter(&(!Enum.member?(opts[:ignore] || [], &1)))
```
josevalim, line 175:
```suggestion
      |> Enum.reject(& &1 in Keyword.get(opts, :ignore, []))
```

josevalim, line 175:
```suggestion
    ignore = opts[:ignore] || []

    modules =
      :cover.modules()
      |> Enum.reject(& &1 in ignore)
```

lib/mix/test/fixtures/umbrella_test/apps/bar/mix.exs
```diff
@@ -7,7 +7,8 @@ defmodule Bar.MixProject do
       version: "0.1.0",
       # Choose something besides *_test.exs so that these test files don't
       # get accidentally swept up into the actual Mix test suite.
-      test_pattern: "*_tests.exs"
+      test_pattern: "*_tests.exs",
+      test_coverage: [ignore_modules: [Bar.Protocol]]
```
josevalim, line 11: Can you ignore `Bar` instead of `Bar.Protocol`? Protocols has specific coverage features so we need to verify it. `Bar` is just a plain module and we already have another plain module as `Foo`. Thanks!

lib/mix/lib/mix/tasks/test.coverage.ex
```diff
@@ -169,7 +169,11 @@ defmodule Mix.Tasks.Test.Coverage do
 
   defp generate_cover_results(opts) do
     {:result, ok, _fail} = :cover.analyse(:coverage, :line)
-    modules = :cover.modules()
+    ignore = opts[:ignore_modules] || []
+
+    modules =
+      :cover.modules()
+      |> Enum.reject(& &1 in ignore)
```
josevalim, line 176:
```suggestion
      |> Enum.reject(&(&1 in ignore))
```

josevalim, line 176:
```suggestion
    ignore = opts[:ignore_modules] || []
    modules = Enum.reject(:cover.modules(), &(&1 in ignore))
```

josevalim:

PR #10318, opened 2020-09-11
Improve wording in Map module docs

eksperimental:
improves the wording of
> If `key` is present in `map` with value `value`

The problem with this phrase is that it can be interpreted that it will update a field where key and value are both matched.

beside that `value` argument no longer exists

It is a WIP because there are other sentences in the Map module to update once the wording is decided

### DIFF
--- a/lib/elixir/lib/map.ex
+++ b/lib/elixir/lib/map.ex
@@ -242,8 +242,8 @@ defmodule Map do
   @doc """
   Fetches the value for a specific `key` in the given `map`.
 
-  If `map` contains the given `key` with value `value`, then `{:ok, value}` is
-  returned. If `map` doesn't contain `key`, `:error` is returned.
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/map.ex
```diff
@@ -599,8 +599,8 @@ defmodule Map do
   @doc """
   Updates the `key` in `map` with the given function.
 
-  If `key` is present in `map` with value `value`, `fun` is invoked with
-  argument `value` and its result is used as the new value of `key`. If `key` is
+  If `key` is present in `map` then the `existing_value` is passed to `fun` and its result is
```
josevalim, line 602:
```suggestion
  If `key` is present in `map` then the existing value is passed to `fun` and its result is
```

since there is no `existing_value` argument anywhere. :)

josevalim:
Love the change, just one tiny comment!

josevalim:

PR #10319, opened 2020-09-11
Use 'gmake' instead of 'make' when compiling deps on NetBSD

niacat:
There are basically two implementations of BSD make now, the NetBSD one (also used by FreeBSD, available as 'bmake' on Linux, macOS, etc), and the OpenBSD one.

### DIFF
--- a/lib/mix/lib/mix/tasks/deps.compile.ex
+++ b/lib/mix/lib/mix/tasks/deps.compile.ex
@@ -15,8 +15,8 @@ defmodule Mix.Tasks.Deps.Compile do
     * `mix.exs` - invokes `mix compile`
     * `rebar.config` - invokes `rebar compile`
     * `Makefile.win`- invokes `nmake /F Makefile.win` (only on Windows)
-    * `Makefile` - invokes `gmake` on FreeBSD, NetBSD, and OpenBSD, invokes
-      `make` on any other operating system (except on Windows)
+    * `Makefile` - invokes `gmake` on DragonFlyBSD, FreeBSD, NetBSD, and OpenBSD,
+      invokes `make` on any other operating system (except on Windows)
 
   The compilation can be customized by passing a `compile` option
   in the dependency:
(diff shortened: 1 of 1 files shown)


eksperimental:
I think DragonFlyBSD could also be added , it is recognized by OTP as `:dragonfly`

Source: https://github.com/erlang/otp/blob/2bd3875917dffb3c43ed50286bc522542c3285e3/lib/os_mon/src/memsup.erl#L180

lib/mix/lib/mix/tasks/deps.compile.ex
```diff
@@ -15,8 +15,8 @@ defmodule Mix.Tasks.Deps.Compile do
     * `mix.exs` - invokes `mix compile`
     * `rebar.config` - invokes `rebar compile`
     * `Makefile.win`- invokes `nmake /F Makefile.win` (only on Windows)
-    * `Makefile` - invokes `gmake` on FreeBSD and OpenBSD, invokes `make`
-      on any other operating system (except on Windows)
+    * `Makefile` - invokes `gmake` on FreeBSD, NetBSD, and OpenBSD, invokes
+      `make` on any other operating system (except on Windows)
```
josevalim, line 19:
```suggestion
    * `Makefile` - invokes `gmake` on DragonFlyBSD, FreeBSD, NetBSD, and OpenBSD,
      invokes `make` on any other operating system (except on Windows)
```

lib/mix/lib/mix/tasks/deps.compile.ex
```diff
@@ -259,7 +259,7 @@ defmodule Mix.Tasks.Deps.Compile do
         {:win32, _} when makefile_win? ->
           "nmake /F Makefile.win"
 
-        {:unix, type} when type in [:freebsd, :openbsd] ->
+        {:unix, type} when type in [:freebsd, :netbsd, :openbsd] ->
```
josevalim, line 262:
```suggestion
        {:unix, type} when type in [:dragonfly, :freebsd, :netbsd, :openbsd] ->
```

josevalim:

PR #10326, opened 2020-09-12
Fix unregistration crash

sasa1977:
This PR fixes a bug which happens if a process started with `name: {:via, Registry, {registry_name, process_name, value}}` crashes during the startup. In this case, instead of the original error, a no function clause exception happens in the Registry module. Minimal example:

```
iex> Registry.start_link(keys: :unique, name: Foo)
iex> Agent.start_link(fn -> raise "foo" end, name: {:via, Registry, {Foo, :name, :value}})

** (EXIT from #PID<0.105.0>) shell process exited with reason: an exception was raised:
    ** (FunctionClauseError) no function clause matching in Registry.unregister_name/1
        (elixir 1.10.2) lib/registry.ex:264: Registry.unregister_name({Foo, :name, :value})
```

### DIFF
--- a/lib/elixir/lib/registry.ex
+++ b/lib/elixir/lib/registry.ex
@@ -261,9 +261,8 @@ defmodule Registry do
   end
 
   @doc false
-  def unregister_name({registry, key}) do
-    unregister(registry, key)
-  end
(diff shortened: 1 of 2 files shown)


josevalim:

PR #10327, opened 2020-09-13
Document all built-in types internally

josevalim:
Also document which functions must be changed once
new types are introduced and which functions must
handle them.

Make sure all of these functions comply to the spec
and improve performance for tuples.


ericmj (approved):
💜

PR #10332, opened 2020-09-14
Improve docs and argument names for Collectable module

eksperimental:
It is not clear what original mean, if the element we extract data or the element we put into.
For clarity I chose to name it the same as the module.

this change make it clear that what needs to be returned is the initial accumulator and not the collectable we are
puting data into.

```diff
-          {original, collector_fun}
+          initial_acc = map_set
+
+          {initial_acc, collector_fun}
```

mention that the MapSet implementation is a simplified version, not the real one.

The implementations in the module have been reformated and some attribute names changed, so it is easier for developers who want to read the source code to understand how protocols are implemented.

### DIFF
--- a/lib/elixir/lib/collectable.ex
+++ b/lib/elixir/lib/collectable.ex
@@ -27,8 +27,8 @@ defprotocol Collectable do
 
   ## Examples
 
-  To show how to manually use the `Collectable` protocol, let's play with its
-  implementation for `MapSet`.
+  To show how to manually use the `Collectable` protocol, let's play with a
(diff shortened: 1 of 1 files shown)


josevalim:

eksperimental:
@josevalim I would like to move the implementations for List and Map to the corresponding modules, and just leave Bitstring in this file. Is there a reason why they were not place in those modules?

josevalim:
For bootstrap reasons. If you move them, the project will most likely fail to bootstrap.

eksperimental:
I see. Thank you.

PR #10333, opened 2020-09-14
More typing improvements

josevalim:
1. Move push_expr_stack to unify and keep helpers typing free

2. Add get_var! when accessing variables so it helps us find
   variables that have not been properly processed

3. Properly process all variables found missing on step 2


josevalim:

PR #10336, opened 2020-09-14
Reorganize documentation sections

eksperimental:
1. Protocols have been split into two separate sections: "Protocols" and "Protocol Implementations"
	 It was kind of difficult to spot what are the actual modules for working with protocols, the same goes for the implementations.

2. "Calendar" section renamed "Calendars". Follows the style of the other sections which are in plural.

3. Sections have been reordered:
	- "Calendars" section moved. I have the felling that Calendar and Protocols are not as general as the other modules, so I have moved them to the bottom of the list.
	- "Code & Macros" the same for this section which IMO had to be higher up in the list.

The section order looks like this:

- Basic Types
- Collections & Enumerables
- IO & System
- Code & Macros
- Processes & Applications
- Calendars
- Protocols
- Protocol Implementations
- Deprecated
- Exceptions

4. Run the formatter on docs.exs

### DIFF
--- a/lib/elixir/docs.exs
+++ b/lib/elixir/docs.exs
@@ -2,7 +2,7 @@
 [
   extras: Path.wildcard("lib/elixir/pages/*.md") ++ ["CHANGELOG.md"],
   groups_for_functions: [
-    Guards: & &1[:guard] == true
+    Guards: &(&1[:guard] == true)
   ],
   skip_undefined_reference_warnings_on: ["lib/elixir/pages/compatibility-and-deprecations.md"],
   groups_for_modules: [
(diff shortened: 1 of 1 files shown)


josevalim:
I kept the current ordering in order to prioritize more introductory content.

Regarding protocols, calling them “protocol implementations” is confusing, because there are also
Implementations, but I understand what you are trying to do.

If we want to split those apart, then I recommend just removing the Protocol and Inspect modules and keep them at the top level. Can you please send a PR for that? The other changes I would rather keep as is, thanks!

eksperimental:
> If we want to split those apart, then I recommend just removing the Protocol and Inspect modules and keep them at the top level

Same level, after Kernel and Kernel.SpecialForms?

So renaming the sections to "Implementations" will suffice?

eksperimental:
I think I get it, you still want it to be named "Protocols"

josevalim:
> Same level, after Kernel and Kernel.SpecialForms?

Wait, I forgot we only have these too at the top. So yeah... I wouldn't do this change then. :(

PR #10337, opened 2020-09-15
Fix "mix format" test to not to error when pattern matches a directory

eksperimental:
It check that the matches are not directories,
otherwise an error will raise

> (File.Error) could not read file "path/to/dir": illegal operation on a directory

### DIFF
--- a/lib/mix/lib/mix/tasks/format.ex
+++ b/lib/mix/lib/mix/tasks/format.ex
@@ -332,9 +332,9 @@ defmodule Mix.Tasks.Format do
   defp expand_args(files_and_patterns, _dot_formatter, {formatter_opts, subs}) do
     files =
       for file_or_pattern <- files_and_patterns,
-          file <- stdin_or_wildcard(file_or_pattern),
+          files <- stdin_or_wildcard(file_or_pattern),
           uniq: true,
-          do: file
+          do: files
 
     if files == [] do
       Mix.raise(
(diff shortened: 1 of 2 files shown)


lib/mix/lib/mix/tasks/format.ex
```diff
@@ -332,9 +332,9 @@ defmodule Mix.Tasks.Format do
   defp expand_args(files_and_patterns, _dot_formatter, {formatter_opts, subs}) do
     files =
       for file_or_pattern <- files_and_patterns,
-          file <- stdin_or_wildcard(file_or_pattern),
+          files <- stdin_or_wildcard(file_or_pattern),
```
josevalim, line 335: This is only one individual file, can you please revert this change?

josevalim:
Can you please provide a rationale for this file? I would rather raise on directories than silently discard them, which may end-up hiding some behavior or accidentally leaving files unformatted. Thanks.

eksperimental:
I tried to run the formatter on all the .ex and .exs in the Elixir project, and I was hitting this error, becuase a test ends with ".exs", so a directory is created with the test name, and the formatter would crash. 
As of now there is not way to exclude certain entries.

This will not exclude the contents for matched directory, as shown in the test, but the directory itself. I would say if it matches a directory it is a coincidence as in the example above, but there is no way to bypass that from the formatter.

josevalim:
Given it is the first time we see this report and it is only within Elixir where we need to check for this corner cases, I think we can postpone this for now. Thank you.

PR #10339, opened 2020-09-17
IEX help in place with tab key

pera:
WIP for this following proposal: https://groups.google.com/forum/?oldui=1#!topic/elixir-lang-core/QcWpEm2a-HA

To try it out just type in some function in your repl and then after the `(` press the tab key. For instance:

> iex(1)> [1, 2, 3] |> Enum.map(**\<TAB>**
> map(enumerable, fun)    
> iex(1)> [1, 2, 3] |> Enum.map(

### DIFF
--- a/lib/iex/lib/iex/autocomplete.ex
+++ b/lib/iex/lib/iex/autocomplete.ex
@@ -128,32 +128,6 @@ defmodule IEx.Autocomplete do
     end
   end
 
-  defp format_definition_from_typespec(definition) do
-    case definition do
-      {:"::", _, [x, _]} -> x
-      {:when, _, [{:"::", _, [x, _]} | _]} -> x
-      x -> x
-    end
-    |> Macro.to_string()
-  end
-
-  defp get_signatures_from_specs(module, name) do
-    with {:ok, all_specs} <- Code.Typespec.fetch_specs(module) do
-      all_specs
-      |> Enum.filter(&match?({{^name, _arity}, _specs}, &1))
-      |> Enum.map(fn {{name, _arity}, specs} ->
-        Enum.map(specs, fn spec ->
-          Code.Typespec.spec_to_quoted(name, spec)
-          |> format_definition_from_typespec()
-        end)
-      end)
-      |> List.flatten()
-      |> Enum.uniq()
-    else
-      _ -> :error
-    end
-  end
-
   defp get_signatures(name, module) when is_atom(module) do
     with docs when is_list(docs) <- get_docs(module, [:function, :macro], name) do
       Enum.map(docs, fn {_, _, [signature], _, _} -> signature end)
@@ -184,16 +158,14 @@ defmodule IEx.Autocomplete do
           |> List.flatten()
 
         {:ok, {{:., _, [mod, fun]}, _, []}} when is_atom(mod) and is_atom(fun) ->
-          case get_signatures(fun, mod) do
-            [_ | _] = xs -> xs
-            _ -> get_signatures_from_specs(mod, fun)
-          end
+          get_signatures(fun, mod)
 
         _ ->
           :error
       end
 
-    with [head | tail] <- Enum.sort(signatures, &(String.length(&1) <= String.length(&2))) do
+    with [_ | _] <- signatures do
+      [head | tail] = Enum.sort(signatures, &(String.length(&1) <= String.length(&2)))
       if tail !== [], do: IO.write("\n" <> (tail |> Enum.reverse() |> Enum.join("\n")))
       yes("", [head])
     else
--- a/lib/iex/lib/iex/autocomplete.ex
+++ b/lib/iex/lib/iex/autocomplete.ex
@@ -51,7 +51,7 @@ defmodule IEx.Autocomplete do
       h in ' (' and t != [] and identifier?(hd(t)) ->
         expand_help(reduce(expr), server)
 
-      h in '[{' ->
+      h in '([{' ->
         expand('')
 
       true ->
--- a/lib/iex/lib/iex/autocomplete.ex
+++ b/lib/iex/lib/iex/autocomplete.ex
@@ -130,18 +130,10 @@ defmodule IEx.Autocomplete do
 
   defp get_signatures(name, module) when is_atom(module) do
     with docs when is_list(docs) <- get_docs(module, [:function, :macro], name) do
-      Enum.map(docs, fn {_, _, [signature], _, _} -> signature end)
+      Enum.map(docs, fn {_, _, signatures, _, _} -> Enum.join(signatures, " ") end)
     else
       _ -> :error
     end
-  end
-
-  defp get_signatures(name, module, fs) when is_atom(module) and is_list(fs) do
-    fs
-    |> Enum.filter(&match?({^name, _arity}, &1))
-    |> Enum.map(fn {name, _} -> {module, name} end)
-    |> Enum.uniq()
-    |> Enum.map(fn {module, name} -> get_signatures(name, module) end)
   end
 
   defp expand_help(expr, server) do
@@ -154,8 +146,8 @@ defmodule IEx.Autocomplete do
 
         {:ok, {fun, _, _}} when is_atom(fun) ->
           imports_from_env(server)
-          |> Enum.map(fn {module, fs} -> get_signatures(fun, module, fs) end)
-          |> List.flatten()
+          |> Enum.filter(fn {_, funs} -> List.keymember?(funs, fun, 0) end)
+          |> Enum.flat_map(fn {module, _} -> get_signatures(fun, module) end)
 
         {:ok, {{:., _, [mod, fun]}, _, []}} when is_atom(mod) and is_atom(fun) ->
           get_signatures(fun, mod)
@@ -164,12 +156,13 @@ defmodule IEx.Autocomplete do
           :error
       end
 
-    with [_ | _] <- signatures do
-      [head | tail] = Enum.sort(signatures, &(String.length(&1) <= String.length(&2)))
-      if tail !== [], do: IO.write("\n" <> (tail |> Enum.reverse() |> Enum.join("\n")))
-      yes("", [head])
-    else
-      _ -> expand('')
+    case signatures do
+      [_ | _] ->
+        [head | tail] = Enum.sort(signatures, &(String.length(&1) <= String.length(&2)))
+        if tail != [], do: IO.write("\n" <> (tail |> Enum.reverse() |> Enum.join("\n")))
+        yes("", [head])
+      _ ->
+        expand('')
     end
   end
 
--- a/lib/iex/lib/iex/autocomplete.ex
+++ b/lib/iex/lib/iex/autocomplete.ex
@@ -160,7 +160,7 @@ defmodule IEx.Autocomplete do
           end
 
         {:ok, {fun, _, _}} when is_atom(fun) ->
-          imports_with_modules_from_env(server)
+          imports_from_env(server)
           |> Enum.map(fn {module, fs} -> get_signatures(fun, module, fs) end)
           |> List.flatten()
 
@@ -316,7 +316,7 @@ defmodule IEx.Autocomplete do
 
   defp expand_variable_or_import(hint, server) do
     variables = expand_variable(hint, server)
-    imports = imports_from_env(server)
+    imports = imports_from_env(server) |> Enum.flat_map(&elem(&1, 1))
     module_funs = get_module_funs(Kernel.SpecialForms)
     funs = match_module_funs(imports ++ module_funs, hint)
     format_expansion(variables ++ funs, hint)
@@ -604,16 +604,6 @@ defmodule IEx.Autocomplete do
   ## Evaluator interface
 
   defp imports_from_env(server) do
-    with {evaluator, server} <- server.evaluator(),
-         env_fields = IEx.Evaluator.fields_from_env(evaluator, server, [:functions, :macros]),
-         %{functions: funs, macros: macros} <- env_fields do
-      Enum.flat_map(funs ++ macros, &elem(&1, 1))
-    else
-      _ -> []
-    end
-  end
-
-  defp imports_with_modules_from_env(server) do
     with {evaluator, server} <- server.evaluator(),
          env_fields = IEx.Evaluator.fields_from_env(evaluator, server, [:functions, :macros]),
          %{functions: funs, macros: macros} <- env_fields do
--- a/lib/iex/lib/iex/autocomplete.ex
+++ b/lib/iex/lib/iex/autocomplete.ex
@@ -48,7 +48,7 @@ defmodule IEx.Autocomplete do
       h == ?/ and t != [] and identifier?(hd(t)) ->
         expand_expr(reduce(t), server)
 
-      h in ' (' ->
+      h in ' (' and t != [] and identifier?(hd(t)) ->
         expand_help(reduce(expr), server)
 
       h in '[{' ->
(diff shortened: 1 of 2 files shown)


lib/iex/lib/iex/autocomplete.ex
```diff
@@ -125,6 +128,46 @@ defmodule IEx.Autocomplete do
...
+      case Code.string_to_quoted(expr) do
+        {:ok, {{:., _, [{:__aliases__, _, aliases}, fun]}, _, []}} when is_atom(fun) ->
+          yes("", [get_signatures!(fun, aliases) |> Enum.join("\n")])
+
+        {:ok, {fun, _, _}} when is_atom(fun) ->
+          yes("", [get_signatures!(fun, server) |> Enum.join("\n")])
+      end
+    rescue
+      _ -> no()
+    end
```
josevalim, line 168: try/catch should not be necessary (given we call string_to_quoted elsewhere in this module too and never wrap it). Maybe you wanted to wrap something else?

lib/iex/lib/iex/autocomplete.ex
```diff
@@ -125,6 +128,46 @@ defmodule IEx.Autocomplete do
...
+  end
+
+  defp expand_help(expr, server) do
+    try do
+      case Code.string_to_quoted(expr) do
+        {:ok, {{:., _, [{:__aliases__, _, aliases}, fun]}, _, []}} when is_atom(fun) ->
+          yes("", [get_signatures!(fun, aliases) |> Enum.join("\n")])
+
+        {:ok, {fun, _, _}} when is_atom(fun) ->
+          yes("", [get_signatures!(fun, server) |> Enum.join("\n")])
```
josevalim, line 164: What if each signature is its own entry in the list?

lib/iex/lib/iex/autocomplete.ex
```diff
@@ -125,6 +128,46 @@ defmodule IEx.Autocomplete do
...
 
+  defp get_signatures!(name, aliases) when is_list(aliases) do
+    {:docs_v1, _, _, _, _, _, docs} =
+      aliases
+      |> Module.safe_concat()
+      |> Code.fetch_docs()
+
+    docs
+    |> Enum.filter(&match?({{:function, ^name, _}, _, _, _, _}, &1))
+    |> Enum.map(fn {_, _, [signature], _, _} -> signature end)
```
josevalim, line 139: According to EEP, this can be multiple entries which you join by line or whitespace. I believe you should write this:

```suggestion
    |> Enum.map(fn {_, _, signature, _, _} -> Enum.join(signature, " ") end)
```

lib/iex/lib/iex/autocomplete.ex
```diff
@@ -125,6 +128,46 @@ defmodule IEx.Autocomplete do
     end
   end
 
+  defp get_signatures!(name, aliases) when is_list(aliases) do
```
josevalim, line 131:
```suggestion
  defp get_signatures!(name, module) when is_atom(module) do
```

So this function receives a module and you don't need to call `get_signatures!(name, [module])` below. Also note we already have some functionality for expanding aliases which you should use. See `expand_alias/2`.

lib/iex/lib/iex/autocomplete.ex
```diff
@@ -125,6 +128,46 @@ defmodule IEx.Autocomplete do
...
+    {:docs_v1, _, _, _, _, _, docs} =
```
josevalim, line 132: This line can fail, so please match on it accordingly. :)

lib/iex/lib/iex/autocomplete.ex
```diff
@@ -125,6 +128,46 @@ defmodule IEx.Autocomplete do
...
+    try do
+      case Code.string_to_quoted(expr) do
+        {:ok, {{:., _, [{:__aliases__, _, aliases}, fun]}, _, []}} when is_atom(fun) ->
+          yes("", [get_signatures!(fun, aliases) |> Enum.join("\n")])
+
+        {:ok, {fun, _, _}} when is_atom(fun) ->
+          yes("", [get_signatures!(fun, server) |> Enum.join("\n")])
+      end
+    rescue
+      _ -> no()
```
josevalim, line 167: Instead of returning `no()`, I believe we should do what we did before, which is to return `expand('')`.

josevalim:
Hi @pera! I gave this a try locally and I found it very useful! thank you! I have added some comments to push things forward.

lib/iex/lib/iex/autocomplete.ex
```diff
@@ -125,6 +128,46 @@ defmodule IEx.Autocomplete do
...
+      |> Enum.uniq()
+      |> Enum.map(fn {module, name} -> get_signatures!(name, [module]) end)
+    end)
+    |> List.flatten
+  end
+
+  defp expand_help(expr, server) do
+    try do
+      case Code.string_to_quoted(expr) do
+        {:ok, {{:., _, [{:__aliases__, _, aliases}, fun]}, _, []}} when is_atom(fun) ->
```
wojtekmach, line 160: we should also add a clause like:

```
{:ok, {{:., _, [mod, fun]}, _, []}} when is_atom(mod) and is_atom(fun) ->
```

this way `iex> :ets.insert(` <kbd>tab</kbd> should work!

kelvinst:
Oh, this feature is GREAT! Anxiously waiting for it... 👍 One suggestion would be to also add more details about the functions when you type, for example, `String.<TAB>`, pretty much like `mix help` does, printing each submodule/function on each line and a short description after.

kelvinst:
Also, you could print these short doc for each signature too, I think it can be pretty helpful...

josevalim:
@kelvinst let's please keep this discussion focused on this particular feature right now. :) `String.` can show dozens of functions, adding the text would definitely be too verbose, but if someone wants to explore it, please do so separately.

lib/iex/lib/iex/autocomplete.ex
pera: Oh yes this was just a minor personal preference, to force one signature per line:

> iex(1)> Enum.reduce(
> reduce(enumerable, acc, fun)    reduce(enumerable, fun)

lib/iex/lib/iex/autocomplete.ex
pera: Is it possible to specify how you want to join them? (like new line vs whitespace)

lib/iex/lib/iex/autocomplete.ex
josevalim: Oh, I see. I think one per line works better indeed.

lib/iex/lib/iex/autocomplete.ex
pera: Sorry I'm not sure I understand: the first `get_signatures!` is also used in `expand_help` when you already have the module after matching `{:__aliases__, _, aliases}`. It's for cases like `IEx.Introspection.open` so you get `[:IEx, :Introspection]` directly into the function call (so you don't need to search for it in the evaluator environment).

josevalim: I mean that you should expand the aliases before you call `get_signature!`. :)

pera:
@josevalim thank you so much for all the comments!
@kelvinst I actually did a few tests with specs but at times it was pretty verbose. After fixing the issues with this PR I will make another branch on my fork if you (or anyone else) wanna try it out.

lib/iex/lib/iex/autocomplete.ex
pera: ah! sorry now I saw `expand_alias`

pera:
@josevalim do you mind if I modify `imports_from_env` to make it like `imports_with_modules_from_env` and do the flat_map in `expand_variable_or_import` instead?

Note there is a problem with erlang functions (I think? maybe it's a problem with my local install).

There is also a minor aesthetic issue where an extra empty line is appended at the bottom when the help string length is greater than 35 (?) - this depends on the configured width (80 in my case). I don't know if there is a way to avoid this, but if not we could just add an extra empty line when the string is less than 36 chars as a workaround to make it looks always the same.

josevalim:
@pera can you post screenshots of the aesthetic issues? I will reply your question about refactoring when I am back at the computer.

pera:
Screenshot_2020-09-18_12-30-29

This will include an extra empty line:
`yes("", ["aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"]) # 36 a's`
...but this wont:
`yes("", ["aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"]) # 35 a's`

josevalim:
I see, I think that’s fine. Maybe if we had one suggestion per entry it
would be better in this case but it does get worse elsewhere.
-- 


pera:
@josevalim ah I think I found where this comes from (IEx.Helpers.print_table), will try check it more in detail this weekend.

josevalim:
> do you mind if I modify imports_from_env to make it like imports_with_modules_from_env and do the flat_map in expand_variable_or_import instead?

Good idea! :+1:

pera:
I was wrong about `print_table` (it's not used here at all): the source of the extra line is in OTP (I believe https://github.com/erlang/otp/blob/master/lib/stdlib/src/edlin_expand.erl#L142) and I have no idea if there is a nice way to solve this. In principle it should be possible to IO.write the help (instead of returning the {yes, '', [signatures_string]} tuple) and then somehow reprint the prompt with what we have typed so far, but I don't know anything about the Erlang shell or the edlin module. There is though an ugly way to hack this using ANSI escape sequences (like, after printing the help with IO.write return `yes("", [IO.ANSI.cursor_up() <> IO.ANSI.cursor_right(width-4)])` where _width_ is `io:columns-4`) but I'm not sure if this is portable.

josevalim:
ANSI is unfortunately not portable. I wouldn't sweat too much about it, having the newline is fine. :)

pera:
@josevalim @wojtekmach I believe I have sorted out the issue with erlang functions by taking just one part of the specs if no compiled docs are available (or at least it works now on my computer :wink: ). In this last commit there is also a simple workaround for the blank line issue, which is actually also a fix for problem when there are too many signatures (eg :erlang.spawn) and edlin tries to replace the whole help string with "..."

josevalim:
Hi @pera! Given Erlang/OTP 23+ supports the docs chunk, I would rather keep this simple and not try to traverse specs at all. So can we please rollback this particular section of the code for now? Other than that, do you believe this is ready for review? :)

pera:
Hey @josevalim, yeah I believe it's ready :)

(I do have OTP 23 but for some reason it looks like my distribution's package doesn't include them, because if I try to do for instance `h :erlang.spawn` iex prints "Module was compiled without docs. Showing only specs." so my assumption was that Erlang's docs were still not supported here)

Eiji7:
> @pera (I do have OTP 23 but for some reason it looks like my distribution's package doesn't include them, because if I try to do for instance `h :erlang.spawn` iex prints "Module was compiled without docs. Showing only specs." so my assumption was that Erlang's docs were still not supported here)

From your screenshot I see that you are using `Elixir` in version `1.10.3`. Support for such docs was added in `Erlang/OTP 23+`, but in `Elixir` support for `Erlang` documentation would be included in `1.11.0` version. It's already supported in current release candidate as well. If you have `asdf` you can install one of 2 `Elixir` versions: `1.11.0-rc.0-otp-23`  or `master-otp-23`.

lib/iex/lib/iex/autocomplete.ex
```diff
@@ -125,6 +128,51 @@ defmodule IEx.Autocomplete do
     end
   end
 
+  defp get_signatures(name, module) when is_atom(module) do
+    with docs when is_list(docs) <- get_docs(module, [:function, :macro], name) do
+      Enum.map(docs, fn {_, _, [signature], _, _} -> signature end)
```
josevalim, line 133: Signature can actually be multiple lines according to the spec. So do this instead:

```suggestion
      Enum.map(docs, fn {_, _, signatures, _, _} -> Enum.join(signatures, " ") end)
```

lib/iex/lib/iex/autocomplete.ex
```diff
@@ -125,6 +128,51 @@ defmodule IEx.Autocomplete do
...
+      case Code.string_to_quoted(expr) do
+        {:ok, {{:., _, [{:__aliases__, _, mod}, fun]}, _, []}} when is_atom(fun) ->
+          with {:ok, mod} <- expand_alias(mod, server) do
+            get_signatures(fun, mod)
+          end
+
+        {:ok, {fun, _, _}} when is_atom(fun) ->
+          imports_from_env(server)
+          |> Enum.map(fn {module, fs} -> get_signatures(fun, module, fs) end)
+          |> List.flatten()
```
josevalim, line 158:
```suggestion
          |> Enum.flat_map(fn {module, fs} -> get_signatures(fun, module, fs) end)
```

lib/iex/lib/iex/autocomplete.ex
```diff
@@ -125,6 +128,51 @@ defmodule IEx.Autocomplete do
...
+    signatures =
+      case Code.string_to_quoted(expr) do
+        {:ok, {{:., _, [{:__aliases__, _, mod}, fun]}, _, []}} when is_atom(fun) ->
+          with {:ok, mod} <- expand_alias(mod, server) do
+            get_signatures(fun, mod)
+          end
+
+        {:ok, {fun, _, _}} when is_atom(fun) ->
+          imports_from_env(server)
+          |> Enum.map(fn {module, fs} -> get_signatures(fun, module, fs) end)
```
josevalim, line 157: `fs` is not really explicit. Maybe `funs`?

```suggestion
          |> Enum.map(fn {module, funs} -> get_signatures(fun, module, funs) end)
```

lib/iex/lib/iex/autocomplete.ex
```diff
@@ -125,6 +128,51 @@ defmodule IEx.Autocomplete do
...
+    end
+  end
+
+  defp get_signatures(name, module, fs) when is_atom(module) and is_list(fs) do
+    fs
+    |> Enum.filter(&match?({^name, _arity}, &1))
+    |> Enum.map(fn {name, _} -> {module, name} end)
+    |> Enum.uniq()
+    |> Enum.map(fn {module, name} -> get_signatures(name, module) end)
+  end
```
josevalim, line 145: fs -> funs for readability.

Also, we can only have one matching name, so I would write this as:

```suggestion
  defp get_signatures(name, module, funs) when is_atom(module) and is_list(funs) do
    if List.keymember?(funs, name, 0) do
      get_signatures(name, module)
    else
      []
    end
  end
```

lib/iex/lib/iex/autocomplete.ex
```diff
@@ -125,6 +128,51 @@ defmodule IEx.Autocomplete do
...
+          |> List.flatten()
+
+        {:ok, {{:., _, [mod, fun]}, _, []}} when is_atom(mod) and is_atom(fun) ->
+          get_signatures(fun, mod)
+
+        _ ->
+          :error
+      end
+
+    with [_ | _] <- signatures do
```
josevalim, line 167: This will be cleaner as a `case` since it has a single clause. Can you please rewrite?

lib/iex/lib/iex/autocomplete.ex
```diff
@@ -125,6 +128,51 @@ defmodule IEx.Autocomplete do
...
+      [head | tail] = Enum.sort(signatures, &(String.length(&1) <= String.length(&2)))
+      if tail !== [], do: IO.write("\n" <> (tail |> Enum.reverse() |> Enum.join("\n")))
```
josevalim, line 169: We generally use `!=` in Elixir, unless `!==` is necessary.

```suggestion
      if tail != [], do: IO.write("\n" <> (tail |> Enum.reverse() |> Enum.join("\n")))
```

josevalim (commented):
Looks great! I have added some comments. We just need some basic tests and we should be good to go! See the existing suite for examples.

pera:
@josevalim muito obrigado pelos comentarios ^_^ I have added a few test cases, please let me know if you think of anything else

josevalim:

PR #10346, opened 2020-09-18
Unify private function in Keyword.replace*

eksperimental:
Additionally tests have been added.


josevalim:
The current names are fine and do_ is used elsewhere in the module. I’d also prefer to not assert on the order, as it is not guaranteed. Thank you.

eksperimental:
> The current names are fine and do_ is used elsewhere in the module. I’d also prefer to not assert on the order, as it is not guaranteed. Thank you.

what about the removal of the code duplication?

PR #10347, opened 2020-09-18
Mention in fun doc that Keyword.delete/2 is inlined by the compiler

eksperimental:
[skip CI]

### DIFF
--- a/lib/elixir/lib/keyword.ex
+++ b/lib/elixir/lib/keyword.ex
@@ -524,6 +524,8 @@ defmodule Keyword do
   Use `delete_first/2` to delete just the first entry in case of
   duplicated keys.
 
+  Inlined by the compiler.
+
   ## Examples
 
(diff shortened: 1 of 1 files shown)


josevalim:
This function is not inlined by the compiler. :)

eksperimental:
> This function is not inlined by the compiler. :)

So what does this mean then?

https://github.com/elixir-lang/elixir/blob/5d5b842168a5e6ab807e9652036aebebb2037e9c/lib/elixir/lib/keyword.ex#L539-L540

josevalim:
This is internal to the module, it does not really leak to users, while the other ones change their external representation.

josevalim:
To be clear, this applies only to callers within the same module.

eksperimental:
I can see the difference now. Thanks for clarifying it. I think it should be documented.

PR #10349, opened 2020-09-19
 Optimize map udpates when there is no need for an update

eksperimental:
I have meassured the changes in Map.replace/3 and when a new value
is different than the current value, there is no penalty,
but when the new value is the same as the current value, the existing
implementation is 50% slower than the one in this commit.

### DIFF
--- a/lib/elixir/lib/map.ex
+++ b/lib/elixir/lib/map.ex
@@ -322,7 +322,10 @@ defmodule Map do
   @spec replace(map, key, value) :: map
   def replace(map, key, value) do
     case map do
-      %{^key => _value} ->
+      %{^key => ^value} ->
+        map
+
+      %{^key => _current_value} ->
         put(map, key, value)
 
       %{} ->
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -2182,6 +2182,7 @@ defmodule Kernel do
 
       {key, val}, acc ->
         case acc do
+          %{^key => ^val} -> acc
           %{^key => _} -> %{acc | key => val}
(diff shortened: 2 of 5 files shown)


josevalim:
Can you please include the benchmark results and which versions you used when benchmarking? I recently sent a PR to the VM so when trying to put something in a map that is already there, it is a noop, so I am wondering if you are running into those optimizations or not.

lib/elixir/lib/map.ex
```diff
@@ -322,7 +322,10 @@ defmodule Map do
   @spec replace(map, key, value) :: map
   def replace(map, key, value) do
     case map do
-      %{^key => _value} ->
+      %{^key => ^value} ->
+        map
+
+      %{^key => _current_value} ->
         put(map, key, value)
```
josevalim, line 329:
```suggestion
        %{map | key => value}
```

josevalim:
Here is the PR for those interested: https://github.com/erlang/otp/commit/a2c92e7a9313b1174da626a3590d9f318f78bb77

It is available since Erlang/OTP 22. So it would be nice to look at the benchmark scripts and results. :)

michalmuskala:
This could be fairly expensive if the value is really large and almost similar - the comparison could take a long time

josevalim:
@michalmuskala exactly, this assumes the comparison between values is fast, which is not true for large values, that's why the VM optimization looks only at the actual identity, and I would rather rely on the VM optimizing it for us. In any case, thanks @eksperimental !

eksperimental:
> This could be fairly expensive if the value is really large and almost similar - the comparison could take a long time

You are totally right @michalmuskala

PR #10354, opened 2020-09-20
Rename struct attribute in defstruct

### LINKED ISSUE #10352: Document reserved @struct attribute
defining `@struct` before defstruct makes defstruct to raise.

```elixir
defmodule Foo do
  @struct config: false
  @enforce_keys Keyword.keys(@struct)
  defstruct @struct
end
```

```
== Compilation error in file lib/foo.ex ==
** (ArgumentError) defstruct has already been called for Foo, defstruct can only be called once per module
```

This reserved struct is not documented. I think it could be added before the "Typespec attributes" section in the `Module` module.
Additionally it could also be documented `@enforce_keys`

I can contribute a PR once agreed on

eksperimental:
Rename `@struct` to `@__struct__`.

Closes #10352

### DIFF
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -4531,7 +4531,7 @@ defmodule Kernel do
             case @enforce_keys do
               [] ->
                 def __struct__(kv) do
-                  Enum.reduce(kv, @struct, fn {key, val}, map ->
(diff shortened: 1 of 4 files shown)


josevalim:
Beautifully done! :green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

PR #10356, opened 2020-09-21
Remove fresh context to optimize in-body guards

josevalim:


ericmj (approved):
💜

josevalim:

lib/elixir/lib/module/types/unify.ex
```diff
@@ -472,6 +497,41 @@ defmodule Module.Types.Unify do
...
+    do: Enum.reduce(args, acc, &collect_var_indexes(&1, context, &2))
+
+  def collect_var_indexes({:union, args}, context, acc),
+    do: Enum.reduce(args, acc, &collect_var_indexes(&1, context, &2))
+
+  def collect_var_indexes({:list, arg}, context, acc),
+    do: collect_var_indexes(arg, context, acc)
+
+  def collect_var_indexes({:map, pairs}, context, acc) do
+    Enum.reduce(pairs, acc, fn {_, key, value} ->
```
michallepicki, line 528: Dialyzer complains about this line. Looks like the function passed to `reduce` is missing an accumulator argument (that would maybe be `context` here?). It seems that it would crash at runtime.

josevalim: Fixed, thanks!

PR #10358, opened 2020-09-21
Remove duplicated assert in quote test

gusaiani:
Hi all,

this PR removes what seems to be just a duplicated assert.

Cheers!


josevalim:

Issue #10361, opened 2020-09-22
Compiler crash on list operators in guards


michalmuskala:
### Environment

* Elixir & Erlang/OTP versions (elixir --version): 
```
Erlang/OTP 23 [erts-11.0.2] [source] [64-bit] [smp:16:16] [ds:16:16:10] [async-threads:1] [hipe] [dtrace]

Elixir 1.10.4 (compiled with Erlang/OTP 23)
```
* Operating system: macOS

### Current behavior

Given the following module:
```elixir
# foo.ex
defmodule Foo do
  def check(value) do
    case value do
      list when list ++ [1] -> list
    end
  end
end
```

running `elixir foo.ex` crashes with
```
** (ArgumentError) argument error
    (stdlib 3.13) :unicode.characters_to_binary(Foo)
    (elixir 1.10.4) src/elixir_utils.erl:125: :elixir_utils.characters_to_binary/1
    (elixir 1.10.4) src/elixir_erl_compiler.erl:76: anonymous fn/1 in :elixir_erl_compiler.format_errors/1
    (stdlib 3.13) lists.erl:1338: :lists.foreach/2
    (elixir 1.10.4) src/elixir_erl_compiler.erl:12: anonymous fn/3 in :elixir_erl_compiler.spawn/2
```
### Expected behavior

This program should be rejected with a nice error, Erlang does not support `++` and `--` in guards (nor does it allow `!`).

eksperimental:
@michalmuskala I have found the bug, it is describe in this comment https://github.com/elixir-lang/elixir/pull/10363#issuecomment-697111564

michalmuskala:
Closed by https://github.com/elixir-lang/elixir/pull/10363

PR #10363, opened 2020-09-22
Raise instead of crashing on when trying to use list operators in guards

goalves:
https://github.com/elixir-lang/elixir/issues/10361.

This pull request adds the functionality to raise instead of crashing when trying to use list operators in guard clauses.

```
❯ bin/iex                                                                                                                                                                                                                                                                                                     19:57:43
Erlang/OTP 23 [erts-11.0.2] [source] [64-bit] [smp:12:12] [ds:12:12:10] [async-threads:1] [hipe] [dtrace]

Interactive Elixir (1.12.0-dev) - press Ctrl+C to exit (type h() ENTER for help)
iex(1)> defmodule Foo do
...(1)>   def check(value) do
...(1)>     case value do
...(1)>       list when list ++ [1] -> list
...(1)>     end
...(1)>   end
...(1)> end
** (ArgumentError) invalid expression in guard, ++ is not allowed in guards. To learn more about guards, visit: https://hexdocs.pm/elixir/patterns-and-guards.html
    (elixir 1.12.0-dev) lib/kernel.ex:5662: Kernel.invalid_guard!/1
    (elixir 1.12.0-dev) expanding macro: Kernel.++/2
    iex:4: Foo.check/1
```

### DIFF
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -1354,9 +1354,12 @@ defmodule Kernel do
       [1, 2 | 3]
 
   """
-  @spec list ++ term :: maybe_improper_list
-  def left ++ right do
-    :erlang.++(left, right)
+  @spec list ++ list :: list
+  defmacro left ++ right do
+    case __CALLER__.context do
+      :guard -> invalid_guard!("++")
+      _ -> quote(do: :erlang.++(unquote(left), unquote(right)))
+    end
   end
 
   @doc """
@@ -1396,8 +1399,11 @@ defmodule Kernel do
 
   """
   @spec list -- list :: list
-  def left -- right do
-    :erlang.--(left, right)
+  defmacro left -- right do
+    case __CALLER__.context do
+      :guard -> invalid_guard!("--")
+      _ -> quote(do: :erlang.--(unquote(left), unquote(right)))
+    end
   end
 
   @doc """
@@ -5644,16 +5650,9 @@ defmodule Kernel do
 
   defp assert_no_match_or_guard_scope(context, exp) do
     case context do
-      :match ->
-        invalid_match!(exp)
-
-      :guard ->
-        raise ArgumentError,
-              "invalid expression in guard, #{exp} is not allowed in guards. " <>
-                "To learn more about guards, visit: https://hexdocs.pm/elixir/patterns-and-guards.html"
-
-      _ ->
-        :ok
+      :match -> invalid_match!(exp)
+      :guard -> invalid_guard!(exp)
+      _ -> :ok
     end
   end
 
@@ -5661,6 +5660,12 @@ defmodule Kernel do
     raise ArgumentError,
           "invalid expression in match, #{exp} is not allowed in patterns " <>
             "such as function clauses, case clauses or on the left side of the = operator"
+  end
+
+  defp invalid_guard!(exp) do
+    raise ArgumentError,
+          "invalid expression in guard, #{exp} is not allowed in guards. " <>
+            "To learn more about guards, visit: https://hexdocs.pm/elixir/patterns-and-guards.html"
   end
 
   # Helper to handle the :ok | :error tuple returned from :elixir_interpolation.unescape_tokens
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -1354,12 +1354,9 @@ defmodule Kernel do
       [1, 2 | 3]
 
   """
-  @spec list ++ list :: list
-  defmacro left ++ right do
-    case __CALLER__.context do
-      :guard -> invalid_guard!("++")
-      _ -> quote(do: :erlang.++(unquote(left), unquote(right)))
-    end
+  @spec list ++ term :: maybe_improper_list
+  def left ++ right do
+    :erlang.++(left, right)
   end
 
   @doc """
@@ -1399,11 +1396,8 @@ defmodule Kernel do
 
   """
   @spec list -- list :: list
-  defmacro left -- right do
-    case __CALLER__.context do
-      :guard -> invalid_guard!("--")
-      _ -> quote(do: :erlang.--(unquote(left), unquote(right)))
-    end
+  def left -- right do
+    :erlang.--(left, right)
   end
 
   @doc """
@@ -5650,9 +5644,16 @@ defmodule Kernel do
 
   defp assert_no_match_or_guard_scope(context, exp) do
     case context do
-      :match -> invalid_match!(exp)
-      :guard -> invalid_guard!(exp)
-      _ -> :ok
+      :match ->
+        invalid_match!(exp)
+
+      :guard ->
+        raise ArgumentError,
+              "invalid expression in guard, #{exp} is not allowed in guards. " <>
+                "To learn more about guards, visit: https://hexdocs.pm/elixir/patterns-and-guards.html"
+
+      _ ->
+        :ok
     end
   end
 
@@ -5660,12 +5661,6 @@ defmodule Kernel do
     raise ArgumentError,
           "invalid expression in match, #{exp} is not allowed in patterns " <>
             "such as function clauses, case clauses or on the left side of the = operator"
-  end
-
-  defp invalid_guard!(exp) do
-    raise ArgumentError,
-          "invalid expression in guard, #{exp} is not allowed in guards. " <>
-            "To learn more about guards, visit: https://hexdocs.pm/elixir/patterns-and-guards.html"
   end
 
   # Helper to handle the :ok | :error tuple returned from :elixir_interpolation.unescape_tokens
--- a/lib/elixir/src/elixir_utils.erl
+++ b/lib/elixir/src/elixir_utils.erl
@@ -27,9 +27,9 @@ guard_op('orelse', 2) ->
 guard_op(Op, Arity) ->
   try erl_internal:op_type(Op, Arity) of
     arith -> true;
-    list  -> true;
     comp  -> true;
     bool  -> true;
+    list  -> false;
     send  -> false
   catch
     _:_ -> false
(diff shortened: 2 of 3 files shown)


goalves:
Trying to figure out why it failed on CI and was working locally. Looks like I compiled everything once and kept doing incremental changes and forgot to clean and compile from scratch again.

lib/elixir/lib/kernel.ex
```diff
@@ -1331,7 +1331,7 @@ defmodule Kernel do
   If the `right` operand is not a proper list, it returns an improper list.
   If the `left` operand is not a proper list, it raises `ArgumentError`.
 
-  Inlined by the compiler.
+  Not allowed in guard clauses. Inlined by the compiler.
```
eksperimental, line 1334: We never mention when it is not allowed. Only when it is

goalves: I thought it was being used because of these:

https://github.com/elixir-lang/elixir/blob/75726d5611413ee45cb5235b1698944b72efa244/lib/elixir/lib/kernel.ex#L1743

https://github.com/elixir-lang/elixir/blob/75726d5611413ee45cb5235b1698944b72efa244/lib/elixir/lib/kernel.ex#L3496

https://github.com/elixir-lang/elixir/blob/75726d5611413ee45cb5235b1698944b72efa244/lib/elixir/lib/kernel.ex#L3536

I can remove the documentation changes if they are needed though.

eksperimental: That is because the strickly boolean operator counter part is allowed.

goalves: I reverted the change that added the doc changes, thanks!

goalves:
I need some help to fix the issue of not building with the `make compile` task. :(

It does not find the function `invalid_guard!`? It looks like it is looking for this function on the master version of the file instead of in the one I am currently compiling (caching? not recompiling the new one in the makefile task somehow?)

```
❯ make compile 
==> bootstrap (compile)
error: undef
stacktrace: [{'Elixir.Kernel','invalid_guard!',1,[]},
```

I looked in the commit that added the `invalid_match!` functionality (https://github.com/elixir-lang/elixir/commit/4416854a9d5920311151bd7111339cb9619cad8f) and it looks like no other changes were required to make it work.

Makefile was changed on this commit 9b6ec7bae79ececfb9ed481c780febb6a5b293e8, could this line that was removed be the possible cause of the my issue?

```
-	$(Q) cd lib/elixir && ../../$(ELIXIRC) "lib/kernel.ex" -o ebin;
```

eksperimental:
> I need some help to fix the issue of not building with the `make compile` task. :(
> 
> It does not find the function `invalid_guard!`? It looks like it is looking for this function on the master version of the file instead of in the one I am currently compiling (caching? not recompiling the new one in the makefile task somehow?)
> 
> ```
> ❯ make compile 
> ==> bootstrap (compile)
> error: undef
> stacktrace: [{'Elixir.Kernel','invalid_guard!',1,[]},
> ```
> 
> I looked in the commit that added the `invalid_match!` functionality ([4416854](https://github.com/elixir-lang/elixir/commit/4416854a9d5920311151bd7111339cb9619cad8f)) and it looks like no other changes were required to make it work.
> 
> Makefile was changed on this commit [9b6ec7b](https://github.com/elixir-lang/elixir/commit/9b6ec7bae79ececfb9ed481c780febb6a5b293e8), could this line that was removed be the possible cause of the my issue?
> 
> ```
> -	$(Q) cd lib/elixir && ../../$(ELIXIRC) "lib/kernel.ex" -o ebin;
> ```

that is a private function when when the macro gets expanded, and it gets called at run time it will not available.
There is already a macro for this, you can check the implementation for `&&/2`

```elixir
  defmacro left && right do
    assert_no_match_or_guard_scope(__CALLER__.context, "&&")
    ...
```

But still I don't think this is the right approach. Look in the `elixir_rewrite.erl` file,
it will refer to the file `elixir_utils.erl`, function `allowed_guard/2`.. I think your answer is there. I haven't fully looked into it.

eksperimental:
in `elixir_rewrite.erl` look for functions `guard_rewrite/5` and `match_rewrite/5`

goalves:
> in `elixir_rewrite.erl` look for functions `guard_rewrite/5` and `match_rewrite/5`

I had deleted the comment by the time you answered to it because I decided to test the makefile changes and it looked like it wasn't the real issue, sorry.

I had previously looked at the `elixir_rewrite.erl` file and thought some change would be needed there but it looked like it was not the case at first because the `invalid_match!` function does not have an existing counterpart and was working fine.

Will take a look at the other places you mentioned, thanks for the advice! :)

eksperimental:
@goalves I found the culprit.
https://github.com/elixir-lang/elixir/blob/0bbcb8f280af07c075b37822ba766ec5fd28bf08/lib/elixir/src/elixir_utils.erl#L23-L36

```elixir
iex(3)> :erl_internal.op_type(:++, 2)    
:list
iex(4)> :erl_internal.op_type(:--, 2)
:list
```
I don't really understand what that means, because the documentation for that modules does not explain much, so instead of changing that line in case to `false`, we can add two clauses as in:

```erlang
guard_op('andalso', 2) ->
  true;
guard_op('orelse', 2) ->
  true;
guard_op('++', 2) ->
  false;
guard_op('--', 2) ->
  false;
guard_op(Op, Arity) ->
  try erl_internal:op_type(Op, Arity) of
    arith -> true;
    list  -> true;
    comp  -> true;
    bool  -> true;
    send  -> false
  catch
    _:_ -> false
  end.
```


You can see it works.

```elixir
Interactive Elixir (1.11.0-rc.0) - press Ctrl+C to exit (type h() ENTER for help)
iex(1)> defmodule Foo do; def foo(x, y) when x ++ y == [] > 0, do: :positive; end
    (stdlib 3.13.1) :unicode.characters_to_binary(Foo)
    (elixir 1.11.0-rc.0) src/elixir_utils.erl:125: :elixir_utils.characters_to_binary/1
    (elixir 1.11.0-rc.0) src/elixir_erl_compiler.erl:76: anonymous fn/1 in :elixir_erl_compiler.format_errors/1
    (stdlib 3.13.1) lists.erl:1338: :lists.foreach/2
    (elixir 1.11.0-rc.0) src/elixir_erl_compiler.erl:12: anonymous fn/3 in :elixir_erl_compiler.spawn/2

iex(2)> defmodule Foo do; def foo(x, y) when x -- y == [] > 0, do: :positive; end
    (stdlib 3.13.1) :unicode.characters_to_binary(Foo)
    (elixir 1.11.0-rc.0) src/elixir_utils.erl:125: :elixir_utils.characters_to_binary/1
    (elixir 1.11.0-rc.0) src/elixir_erl_compiler.erl:76: anonymous fn/1 in :elixir_erl_compiler.format_errors/1
    (stdlib 3.13.1) lists.erl:1338: :lists.foreach/2
    (elixir 1.11.0-rc.0) src/elixir_erl_compiler.erl:12: anonymous fn/3 in :elixir_erl_compiler.spawn/2
```

```elixir
Interactive Elixir (1.12.0-dev) - press Ctrl+C to exit (type h() ENTER for help)
iex(2)> defmodule Foo do; def foo(x, y) when x ++ y == [] > 0, do: :positive; end
** (CompileError) iex:2: cannot invoke remote function :erlang.++/2 inside guards
    (stdlib 3.13.1) lists.erl:1354: :lists.mapfoldl/3

iex(2)> defmodule Foo do; def foo(x, y) when x -- y == [] > 0, do: :positive; end 
** (CompileError) iex:2: cannot invoke remote function :erlang.--/2 inside guards
    (stdlib 3.13.1) lists.erl:1354: :lists.mapfoldl/3

```

goalves:
@eksperimental

Interesting... I was testing a few minutes ago and made it work just by not using `if` and using `case` (?) and it now compiles properly using `make clean_elixir` and `make compile`.

I tested compiling everything before by editing the `list -> true` to `list -> false` in the `elixir_utils.erl` file but I was still not able to make it behave properly.

Will try to replicate your changes but in the mean time I pushed an update that made things work (at least for me).

eksperimental:
> @eksperimental
> 
> Interesting... I was testing a few minutes ago and made it work just by not using `if` and using `case` (?) and it now compiles properly using `make clean_elixir` and `make compile`.
> 
> I tested compiling everything before by editing the `list -> true` to `list -> false` in the `elixir_utils.erl` file but I was still not able to make it behave properly.
> 
> Will try to replicate your changes but in the mean time I pushed an update that made things work (at least for me).

after doing a bit of research, I came to comment that it is safe to change the list line to false since i think ++ and -- are the only two list erlang operators.
I have compiled it and tested the IEx code and it works just fine

eksperimental:
I think changing true to false, and probably moving it to the bottom (right above send) and adding two test for these guards will suffice for this PR

Additionally you can add these tests which are related.
```elixir
Interactive Elixir (1.12.0-dev) - press Ctrl+C to exit (type h() ENTER for help)
iex(1)> [] ++ [] = []
[]
iex(2)> [] -- [] = []
** (CompileError) iex:2: cannot invoke remote function :erlang.--/2 inside a match
```

eksperimental:
the changes in Kernel are not needed.

lib/elixir/lib/kernel.ex
```diff
@@ -1354,9 +1354,12 @@ defmodule Kernel do
       [1, 2 | 3]
 
   """
-  @spec list ++ term :: maybe_improper_list
-  def left ++ right do
-    :erlang.++(left, right)
+  @spec list ++ list :: list
+  defmacro left ++ right do
```
josevalim, line 1358: Since this function is inlining by the compiler, it should not be a macro. The changes on elixir_utils.erl should be enough, can you please revert this? Thanks!

eksperimental (approved):

lib/elixir/lib/kernel.ex
goalves: Looks like you already did it! Thanks! :)

josevalim:

PR #10367, opened 2020-09-24
Use timer function instead of manually calculating milliseconds

pdgonzalez872:
If we use the function, we don't need the comment explaining that the millis make up for 2 hours.

### DIFF
--- a/lib/elixir/lib/gen_server.ex
+++ b/lib/elixir/lib/gen_server.ex
@@ -242,8 +242,7 @@ defmodule GenServer do
         end
 
         defp schedule_work do
-          # In 2 hours
-          Process.send_after(self(), :work, 2 * 60 * 60 * 1000)
+          Process.send_after(self(), :work, :timer.hours(2))
         end
       end
 
--- a/lib/elixir/lib/gen_server.ex
+++ b/lib/elixir/lib/gen_server.ex
@@ -242,7 +242,9 @@ defmodule GenServer do
         end
 
         defp schedule_work do
-          Process.send_after(self(), :work, :timer.hours(2))
+          # We schedule the work to happen in 2 hours (written in milliseconds).
+          # Alternatively, one might write :timer.hours(2)
+          Process.send_after(self(), :work, 2 * 60 * 60 * 1000)
         end
       end
 


josevalim:
Thanks for the PR!

I understand the pros here but given this is an example, I think it is also important to show the original value is in milliseconds and the explicit computation makes that explicit. It also doesn't introduce a dependency on a separate module the reader would have to get acquainted to.

Thoughts?

pdgonzalez872:
I didn't create an issue for this, since it is not an issue. Please let me know if you want me to create one and I will.

Also, feel free to not accept the suggestion, because it is just that, a suggestion. The current suggestion works, this is just an alternative.

pdgonzalez872:
Ah, @josevalim you are too quick! :)

I see your point. I'd still would prefer to point people to the helpers, because I myself didn't know about them for a little while. Maybe I can add to the comment? `in 2 hours. Can also be created with :timer.hours(2)`

Unsure if it adds any value though... Your call ❤️

pdgonzalez872:
Closing this for now ❤️

josevalim:
I think we can leave it open for a while for discussion and merge based on feedback or close if there is no activity. :)

lib/elixir/lib/gen_server.ex
```diff
@@ -242,8 +242,7 @@ defmodule GenServer do
         end
 
         defp schedule_work do
-          # In 2 hours
-          Process.send_after(self(), :work, 2 * 60 * 60 * 1000)
+          Process.send_after(self(), :work, :timer.hours(2))
```
eksperimental, line 245:
```suggestion
          # In 2 hours (expressed in milliseconds)
          Process.send_after(self(), :work, 2 * 60 * 60 * 1000) # time can also be calculated with :timer.hours(2)
```

eksperimental: It is also important to mention that the time is in milliseconds because it appears for the first time in the section bellow

lonamiaec:
IMHO the original example is really explicit as Jose said. 
For me, something important is consistency across documentation,  and if we are not used to adding this kind of comments on methods requiring time arguments we would end up breaking this consistency.

eksperimental:
> IMHO the original example is really explicit as Jose said.
> For me, something important is consistency across documentation, and if we are not used to adding this kind of comments on methods requiring time arguments we would end up breaking this consistency.

Then I think my suggestion is still valid, and I think the user of :timer to simplify the calculation of time in milliseconds could be added to the `Process` module

michalmuskala:
Calculating it directly makes it more obvious what is the unit IMHO. Using something like `:timer.hours` you have no idea if it outputs seconds, milliseconds, microseconds, etc.

I agree the additional comment about milliseconds would be very helpful.

lib/elixir/lib/gen_server.ex
josevalim, line 245:
```suggestion
          # We schedule the work to happen in 2 hours (written in milliseconds).
          # Alternatively, one might write :timer.hours(2)
          Process.send_after(self(), :work, 2 * 60 * 60 * 1000)
```

josevalim:

PR #10369, opened 2020-09-25
Fix copy paste mistake in String.capitalize docs

tomekowal:
String.capitalize uppercases first letter.
Looks like a copy-paste mistake from the function above it :)

### DIFF
--- a/lib/elixir/lib/string.ex
+++ b/lib/elixir/lib/string.ex
@@ -874,7 +874,7 @@ defmodule String do
 
   `mode` may be `:default`, `:ascii` or `:greek`. The `:default` mode considers
   all non-conditional transformations outlined in the Unicode standard. `:ascii`
-  uppercases only the letters A to Z. `:greek` includes the context sensitive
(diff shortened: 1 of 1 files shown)


hassanshaikley (approved):
I prefer the verb "capitalizes" to "uppercases" because it sounds unnatural but I understand that "lowercases" is used as a verb in the other fn. 

Just my $.02

lib/elixir/lib/string.ex
```diff
@@ -874,7 +874,7 @@ defmodule String do
 
   `mode` may be `:default`, `:ascii` or `:greek`. The `:default` mode considers
   all non-conditional transformations outlined in the Unicode standard. `:ascii`
-  lowercases only the letters A to Z. `:greek` includes the context sensitive
+  uppercases only the letters A to Z. `:greek` includes the context sensitive
```
hassanshaikley, line 877:
```suggestion
  capitalizes only the letters A to Z. `:greek` includes the context sensitive
```

I understand if you don't want this. Leaving it here anyways!

josevalim:

PR #10370, opened 2020-09-25
ArgumentError example for binary_part

hassanshaikley:
I thought this would be nice to have.

### DIFF
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -305,6 +305,11 @@ defmodule Kernel do
       iex> binary_part("Hello", 5, -3)
       "llo"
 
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/kernel.ex
```diff
@@ -305,6 +305,11 @@ defmodule Kernel do
       iex> binary_part("Hello", 5, -3)
       "llo"
 
+  An ArgumentError is raised when the length is outside of the binary.
+
+      iex> binary_part("Hello", 0, -1)
```
josevalim, line 310:
```suggestion
  An `ArgumentError` is raised when the length is outside of the binary:

      iex> binary_part("Hello", 0, 10)
```

josevalim:

PR #10372, opened 2020-09-28
Give capability to take over the shell without asking the user

wrachwal:
`iex` v1.10 (v1.9 was oldest I tried) had a capability to takeover in a way that a user was not prompted.

After changes in commit 9413dcc57336a9bbac1a6d12ad2c0e32c929754b this stopped working for me, and the question is always asked.

In the code before above change when `identifier=true` i got the desired effect in a useful trick allowing to integrate elixir shell with legacy (non-elixir) interfaces, all in one terminal.

For interested and to motivate having such capability in IEx code in short how the idea was realized:
When `iex` starts, a line in `.iex.exs` calls some bootstrap code to plug a custom evaluator. That evaluator is really a bridge between the iex server and the standard elixir evaluator. When that bridge gets an expression to evaluate starting with an invalid character (.e.g. dot) for elixir syntax, it directs such expression to non-elixir evaluator associated with the legacy interface.

NOTE: I do not use `IEx.Broker.take_over/2` directly, but I do similar what is does.

Changes in short:
- retained name, the _identifier_ (_take_identifier_ in server code), can either be `iodata` (`[location, whereami]`) or `boolean`
- explicit use of boolean in `IEx.Server.take_over?/4`

### DIFF
--- a/lib/iex/lib/iex/broker.ex
+++ b/lib/iex/lib/iex/broker.ex
@@ -73,16 +73,18 @@ defmodule IEx.Broker do
 
   @doc """
   Client requests a takeover.
+
+  To takeover without prompting the user, use boolean identifier as auto response.
   """
-  @spec take_over(binary, iodata, keyword) ::
+  @spec take_over(iodata | boolean, keyword) ::
           {:ok, server :: pid, group_leader :: pid} | {:error, :no_iex | :refused}
-  def take_over(location, whereami, opts) do
+  def take_over(identifier, opts) do
     case GenServer.whereis(@name) do
       nil ->
         {:error, :no_iex}
 
       _pid ->
-        GenServer.call(@name, {:take_over, location, whereami, opts}, :infinity)
+        GenServer.call(@name, {:take_over, identifier, opts}, :infinity)
     end
   end
 
@@ -99,7 +101,7 @@ defmodule IEx.Broker do
   end
 
   @impl true
-  def handle_call({:take_over, location, whereami, opts}, {_, ref} = from, state) do
+  def handle_call({:take_over, identifier, opts}, {_, ref} = from, state) do
     case servers(state) do
       [] ->
         {:reply, {:error, :no_iex}, state}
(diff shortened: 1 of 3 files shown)


wrachwal:
Hi @josevalim, I sketched the whole story just to show the context, but i don't expect guarantees in the future.

If my wish boils down to that in the subject (i know iex itself doesn't utilize it, but it itself is not big feature, right?), sounds it reasonable? If so, forget what I wrote about the bridge etc. assessing only that silent/quiet option.

I am aware that in the future I would have to adopt my code to changing iex stuff (or finding completely new way), but foremost I'd want to not to patch elixir right from the start.

I don't know how "a customizable IEx line parser or similar" could solve my case: behind mentioned legacy interface stands a big c++ program and what i do is direct command to it, get information for TAB expansion, and print what it wants to be printed.

josevalim:
> I don't know how "a customizable IEx line parser or similar" could solve my case: behind mentioned legacy interface stands a big c++ program and what i do is direct command to it, get information for TAB expansion, and print what it wants to be printed.

My idea is that the parser receives a string and emits Elixir quote expressions. Therefore, if the line starts with "$" (let's assume `$` is the command used for custom instructions), you could convert this:

```elixir
$somemagiccommand
```

to the following quoted expression:

```elixir
quote do
  YourWrapperToTheCPPCodebase.some_function("somemaagiccommand")
end
```

Do you see what I mean?

josevalim:
That was just an example. We will give you the whole string inputed by the user, what to do with it is up to you. :) you shouldn’t need to pass with the evaluator at all.

wrachwal:
Ah, the concept is being refined. That's great! So, in `v1.11` i will patch the elixir itself (one time, no harm), and later (likely `v1.12`) get rid off this bridge. Thanks a lot!

wrachwal:
Not `v1.12`, if notified, I can promise to utilize/check it as soon as it get into `v1.12-dev`

josevalim:
A description of the solution to this problem can be found here: #10372. Thanks!

PR #10376, opened 2020-09-29
Add :turkic mode option to String case functions

altdsoy:
EDIT: ~This PR fixes a little error in the current unconditional mapping of the "Latin capital letter i with dot above" in SpecialCasing.txt which has an error.~ SpecialCasing.txt is left to its original since its generated by Unicode.

It also adds proper handling of case manipulation of the letter i and its dotless version existing in the Turkish language.
Handling is now done according to the SpecialCasing.txt from here http://www.unicode.org/Public/UCD/latest/ucd/SpecialCasing.txt

### DIFF
--- a/lib/elixir/lib/string.ex
+++ b/lib/elixir/lib/string.ex
@@ -286,7 +286,7 @@ defmodule String do
   @typedoc "Pattern used in functions like `replace/4` and `split/3`"
   @type pattern :: t | [t] | :binary.cp()
 
-  @conditional_mappings [:greek]
+  @conditional_mappings [:greek, :turkic]
 
   @doc """
   Checks if a string contains only printable characters up to `character_limit`.
@@ -759,10 +759,10 @@ defmodule String do
   @doc """
   Converts all characters in the given string to uppercase according to `mode`.
 
-  `mode` may be `:default`, `:ascii` or `:greek`. The `:default` mode considers
+  `mode` may be `:default`, `:ascii`, `:greek` or `:turkic`. The `:default` mode considers
   all non-conditional transformations outlined in the Unicode standard. `:ascii`
   uppercases only the letters a to z. `:greek` includes the context sensitive
-  mappings found in Greek.
+  mappings found in Greek. `:turkic` properly handles the letter i with the dotless variant.
 
   ## Examples
 
@@ -782,8 +782,16 @@ defmodule String do
       iex> String.upcase("olá", :ascii)
       "OLá"
 
+  And `:turkic` properly handles the letter i with the dotless variant:
+
+      iex> String.upcase("ıi")
+      "II"
+
+      iex> String.upcase("ıi", :turkic)
+      "Iİ"
+
   """
-  @spec upcase(t, :default | :ascii | :greek) :: t
+  @spec upcase(t, :default | :ascii | :greek | :turkic) :: t
   def upcase(string, mode \\ :default)
 
   def upcase("", _mode) do
@@ -811,10 +819,10 @@ defmodule String do
   @doc """
   Converts all characters in the given string to lowercase according to `mode`.
 
-  `mode` may be `:default`, `:ascii` or `:greek`. The `:default` mode considers
+  `mode` may be `:default`, `:ascii`, `:greek` or `:turkic`. The `:default` mode considers
   all non-conditional transformations outlined in the Unicode standard. `:ascii`
   lowercases only the letters A to Z. `:greek` includes the context sensitive
-  mappings found in Greek.
+  mappings found in Greek. `:turkic` properly handles the letter i with the dotless variant.
 
   ## Examples
 
@@ -834,7 +842,7 @@ defmodule String do
       iex> String.downcase("OLÁ", :ascii)
       "olÁ"
 
-  And `:greek` properly handles the context sensitive sigma in Greek:
+  The `:greek` mode properly handles the context sensitive sigma in Greek:
 
       iex> String.downcase("ΣΣ")
       "σσ"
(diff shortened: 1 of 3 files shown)


lib/elixir/unicode/SpecialCasing.txt
```diff
@@ -66,7 +66,7 @@
 
 # Preserve canonical equivalence for I with dot. Turkic is handled below.
 
-0130; 0069 0307; 0130; 0130; # LATIN CAPITAL LETTER I WITH DOT ABOVE
+0130; 0069; 0130; 0130; # LATIN CAPITAL LETTER I WITH DOT ABOVE
```
josevalim, line 69: Those files are generated by Unicode and we should not modify them. I believe you may need to add this value to `@conditional_downcase` in properties.ex. Thanks!

lib/elixir/lib/string.ex
```diff
@@ -286,7 +286,7 @@ defmodule String do
   @typedoc "Pattern used in functions like `replace/4` and `split/3`"
   @type pattern :: t | [t] | :binary.cp()
 
-  @conditional_mappings [:greek]
+  @conditional_mappings [:greek, :turkish]
```
josevalim, line 289: I believe we should it `:turkic` as it applies to turkic languages (az, tr) but it has been a while since I have been through Unicode documents.

lib/elixir/unicode/SpecialCasing.txt
altdsoy: Ok..
Because currently, doing `String.downcase("İ") == "i"` returns `false`.
In fact I noticed that, when the test failed.

lib/elixir/test/elixir/string_test.exs
```diff
@@ -185,6 +189,10 @@ defmodule StringTest do
     assert String.downcase("OLÁ", :ascii) == "olÁ"
   end
 
+  test "downcase/1 with turkish" do
+    assert String.downcase("Iİ", :turkish) == "ıi"
```
josevalim, line 193: The unicode standard defines two tests here ([scroll to the bottom of this page](http://www.unicode.org/Public/UCD/latest/ucd/SpecialCasing.txt)).  I am copying here for convenience:

```elixir
# Turkish and Azeri

# I and i-dotless; I-dot and i are case pairs in Turkish and Azeri
# The following rules handle those cases.

0130; 0069; 0130; 0130; tr; # LATIN CAPITAL LETTER I WITH DOT ABOVE
0130; 0069; 0130; 0130; az; # LATIN CAPITAL LETTER I WITH DOT ABOVE

# When lowercasing, remove dot_above in the sequence I + dot_above, which will turn into i.
# This matches the behavior of the canonically equivalent I-dot_above

0307; ; 0307; 0307; tr After_I; # COMBINING DOT ABOVE
0307; ; 0307; 0307; az After_I; # COMBINING DOT ABOVE

# When lowercasing, unless an I is before a dot_above, it turns into a dotless i.

0049; 0131; 0049; 0049; tr Not_Before_Dot; # LATIN CAPITAL LETTER I
0049; 0131; 0049; 0049; az Not_Before_Dot; # LATIN CAPITAL LETTER I

# When uppercasing, i turns into a dotted capital I

0069; 0069; 0130; 0130; tr; # LATIN SMALL LETTER I
0069; 0069; 0130; 0130; az; # LATIN SMALL LETTER I
```

Can you please make sure we have both lowercase tests?

lib/elixir/unicode/SpecialCasing.txt
josevalim: Yes, we want to fix that, but only for `:turkic`. :)

altdsoy: Regarding the @condtional_downcase mapping (currently having greek sigma), I thought it's about mapping that are conditional within the language, while in Turkic, there aren't conditional, they apply everytime...

altdsoy: Oh..
I got it..
It's in order to bypass the default generation..

josevalim: @altdsoy I remember now! The reason why we do it with `@conditional_mapping` is because we didn't want to match on `mode == :greek` because it may affect performance. We just want to check the mode once we are inside the character that has a special case and the `@conditional_downcase` then avoids a warning. For now, leave this particular part of the code as is, I can refactor and run benchmarks later. :)

lib/elixir/test/elixir/string_test.exs
altdsoy: Does it mean that the actual SpecialCasing.txt is truncated?
I don't see the `#EOF` flag..
I'll add this one too, if you don't mind...

josevalim: It is truncated because the bottom of it needs to be implemented by hand, we cannot do it ourselves. Generally speaking, we don't touch the SpecialCasing.txt unless when updating Unicode.

altdsoy: I noticed that it's the conditional mappings that are truncated..
So the one we should take care manually..
I'll try to implement that later this day..
Thanks for the hints..

kipcole9: I'm happy to see Turkic language support for casing making a comeback - it gladdens my Unicode heart!  That leaves only Lithuanian in `SpecialCasing.txt` not supported after this PR.

altdsoy: Hello @josevalim 
I revert back the SpecialCasing.txt

And I added support for downcasing in both case (with and without the combining dot above character).
As how I understand, I left the canonical equivalence of the dot for the default.

@kipcole9 
I'll be happy to add those little Lithuanian use cases..
Just give me a little time to dig into the details of how the accents are handled according to http://www.unicode.org/Public/UCD/latest/ucd/SpecialCasing.txt
In a couple of days it should be done..

altdsoy: Do I need to merge the commits, or make a new PR?
I'm not used to contribute in public repos (but I'll be more than happy to do it more often!)

altdsoy: @josevalim I did squash the commits

josevalim:
Thank you @altdsoy! Btw, do you speak any turkic languages? In case you don't, I will also ask for some external sanity checking. :green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

josevalim:

altdsoy:
> Thank you @altdsoy! Btw, do you speak any turkic languages? In case you don't, I will also ask for some external sanity checking.

I do :)

I'm going to also implement the Lithuanian Special Casing..
For this, it'll be good to check indeed..

PR #10378, opened 2020-09-29
Handle value vars in open maps without singleton keys

josevalim:


lib/elixir/lib/module/types.ex
```diff
@@ -349,6 +349,7 @@ defmodule Module.Types do
 
   defp format_traces(traces, simplify?) do
     traces
+    |> Enum.uniq()
```
josevalim, line 352: Maps currently accumulate traces as we do multiple passes, this removes the duplicate traces. Not sure if there is a better solution... but I would rather do the extra work when formatting errors than rather as we unify.

ericmj (approved):
💜

josevalim:

PR #10381, opened 2020-10-01
Update README.md

joyskmathew:


josevalim:
We plan to keep things as is as we don't think the bold is helpful, thanks.

PR #10384, opened 2020-10-01
Take --dot-iex into account after shell respawn

wrachwal:
When the shell is respawned, as a result of `respawn/0`,` continue/0`, or taking a signal, the file given as `--dot-iex` argument is not evaluated. Worse, the default file (`./iex.exs` or `~/.iex.exs`) will be used if present.

The behaviour it is not consistent with normal use (no option and the default file).


josevalim:

PR #10389, opened 2020-10-03
Enforce enforce keys

eksperimental:


lib/elixir/lib/kernel/utils.ex
```diff
@@ -108,7 +108,18 @@ defmodule Kernel.Utils do
...
+    case enforce_keys -- :maps.keys(struct) do
+      [] ->
+        {struct, enforce_keys, Module.get_attribute(module, :derive)}
+
+      error_keys ->
+        raise CompileError,
+          description:
+            "@enforce_keys required key/s (#{inspect(error_keys)}) that are not defined in defstruct: #{
+              inspect(fields)
+            }"
```
josevalim, line 121:
```suggestion
        raise ArgumentError,
              "@enforce_keys required keys (#{inspect(error_keys)}) that are not defined in defstruct: " <>
                "#{inspect(fields)"
```

Changes:

1. CompileError is typically raised only by the compiler and when it has line+column info. ArgumentError is more appropriate here

2. Format the error message to less lines

3. Remove "key/s" idiom which is not used anywhere

josevalim:
Thanks for the PR @eksperimental!

Note there is something wrong in your "errors_test.exs" file as it has been fully formatted without taking into account the .formatter.exs configuratin. Perhaps you formatted it from lib/elixir instead of the project root?

eksperimental:
> Thanks for the PR @eksperimental!
> 
> Note there is something wrong in your "errors_test.exs" file as it has been fully formatted without taking into account the .formatter.exs configuratin. Perhaps you formatted it from lib/elixir instead of the project root?

that is because my test editor is running the formatter automatically.

josevalim:
> that is because my text editor is running the formatter automatically.

It looks like it is not picking up the .formatter.exs from the project root. Once you revert the formatting and amend the tests, we should be good to go. Thank you!

eksperimental:
fixed. please do not squash the commits as they cover two separate things

josevalim:

eksperimental:
> > that is because my text editor is running the formatter automatically.
> 
> It looks like it is not picking up the .formatter.exs from the project root. Once you revert the formatting and amend the tests, we should be good to go. Thank you!

I have come up with a solution.
If anybody else is running into this, run these commands in the `elixir` root folder
It creates symbolic links to the formatter file, and excludes them from your Git's view

```console
for file in lib/{elixir,eex,ex_unit,iex,logger,mix}/.formatter.exs; do
  ln -s ../../.formatter.exs ${file} &&
  echo $file >> .git/info/exclude;
done
```

PR #10391, opened 2020-10-03
Add default value for List.first/1 and List.last/1

btkostner:
Adds a default field for `List.first/1` and `List.last/1`. Should be pretty simple

### DIFF
--- a/lib/elixir/lib/list.ex
+++ b/lib/elixir/lib/list.ex
@@ -255,12 +255,15 @@ defmodule List do
   end
 
   @doc """
-  Returns the first element in `list` or `nil` if `list` is empty.
+  Returns the first element in `list` or `default` if `list` is empty.
 
   ## Examples
 
       iex> List.first([])
       nil
+
+      iex> List.first([], 1)
+      1
 
       iex> List.first([1])
       1
@@ -269,18 +272,22 @@ defmodule List do
       1
 
   """
-  @spec first([]) :: nil
-  @spec first([elem, ...]) :: elem when elem: var
-  def first([]), do: nil
-  def first([head | _]), do: head
+  @spec first([], any) :: any
+  @spec first([elem, ...], any) :: elem when elem: var
+  def first(list, default \\ nil)
+  def first([], default), do: default
+  def first([head | _], _default), do: head
 
   @doc """
-  Returns the last element in `list` or `nil` if `list` is empty.
+  Returns the last element in `list` or `default` if `list` is empty.
 
   ## Examples
 
       iex> List.last([])
       nil
+
+      iex> List.last([], 1)
+      1
 
       iex> List.last([1])
       1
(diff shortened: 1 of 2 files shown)


josevalim:

eksperimental:
what is the best way to deal with the `@doc since` value in this case?

wojtekmach:
@eksperimental I believe the only way would be to break it up:

```diff
- @doc ...
- def first(list, default \\ nil)
- ...
+ @doc ...
+ def first(list) do
+   first(list, nil)
+ end
+
+ @doc ...
+ @doc since: "1.12.0"
+ def first(list, default) do
+ ...
```

not sure it's worth it though.

Eiji7:
@wojtekmach At first sight I think exactly same … However I would like to mention that there is already a similar case:
https://github.com/elixir-lang/elixir/blob/1145dc01680aab7094f8a6dbd38b65185e14adb4/lib/elixir/lib/enum.ex#L404-L425

That's said in mentioned code `:discard` was never part of public `API` and function has changed its name. On the other hand we have old `chunk/2` and `chunk/3` (now deprecated) and new `chunk_every/2` and `chunk_every/3` with `since` added.

Whatever would be decided I'm really interested see opinions of other people.

eksperimental:
It defeats the purpose to split it into two functions. I think we should add `@doc since: "1.12.0"` and override the missing doc since (probable v1.0.0).

Eiji7:
@eksperimental Oh, I see. I would not be surprised if people (especially new in `Elixir`) would be confused asking: `"Why it's working on 1.10.4 when it's since is set to 1.12.0?"`. I'm a bit worried that some of them may open new issues thinking it's a bug/documentation fix report. What do you think?

eksperimental:
> @eksperimental Oh, I see. I would not be surprised if people (especially new in `Elixir`) would be confused asking: `"Why it's working on 1.10.4 when it's since is set to 1.12.0?"`. I'm a bit worried that some of them may open new issues thinking it's a bug/documentation fix report. What do you think?

`List.last/2` won't be working on 1.10.4, it will be `List.last/1`, but the since attribute it for last/2

Eiji7:
@eksperimental I know that `List.last/2` would not work … I mean that `List.last/1` would be working on `1.10.4` and new developer **may think** that it was **always** `List.last/2` (always with default argument). From such point `since` would be seen as invalid. Obviously we have a version switch in documentation - that's why I'm more than ok with your solution.

Just thinking if it would not be confusing for those new developers which does not need to check previous versions … Also switching to (one) previous version is not a big deal, but let's imagine that we have `>= 1.20.0` version and we need to take a look at multiple previous documentations to find that some function previously have less arguments and we still have no idea about its (old) `since`.

That's why if I would need to decide I would go with more clear way of having both `List.last/1` and `List.last/2`.

eksperimental:
> @eksperimental I know that `List.last/2` would not work … I mean that `List.last/1` would be working on `1.10.4` and new developer **may think** that it was **always** `List.last/2` (always with default argument). From such point `since` would be seen as invalid. Obviously we have a version switch in documentation - that's why I'm more than ok with your solution.
> 
> Just thinking if it would not be confusing for those new developers which does not need to check previous versions … Also switching to (one) previous version is not a big deal, but let's imagine that we have `>= 1.20.0` version and we need to take a look at multiple previous documentations to find that some function previously have less arguments and we still have no idea about its (old) `since`.
> 
> That's why if I would need to decide I would go with more clear way of having both `List.last/1` and `List.last/2`.

@Eiji7 I have created a PR addressing this issue

PR #10393, opened 2020-10-04
Remove compilation warning in doctest in Kernel.binary_part/3

eksperimental:
> warning: this expression will fail with ArgumentError
>   (for doctest at) lib/kernel.ex:310

It was introduced in  538a9a3 (#10370)


josevalim:

Issue #10390, opened 2020-10-03
Make IEx parser customizable


josevalim:
This is a proposal to address #10372.

The parser is a MFArgs which receives the current input, the parsing options, and the buffer. It returns either `{:ok, expr, buffer}` or `{:incomplete, buffer}`. Buffer always starts as an empty string. Here is the default implementation:

```elixir
@break_trigger "#iex:break\n"

def parse(@break_trigger, _opts, "") do
  {:incomplete, ""}
end

def parse(@break_trigger, opts, _buffer) do
  :elixir_errors.parse_error([line: opts[:line]], opts[:file], "incomplete expression", "")
end

def parse(input, opts, buffer) do
  input = buffer <> input
  
  case Code.string_to_quoted(input, opts) do
    {:ok, forms} ->
      {:ok, forms, ""}

    {:error, {_, _, ""}} ->
      {:incomplete, input}

    {:error, {location, error, token}} ->
      :elixir_errors.parse_error(location, opts[:file], error, token)
  end
end
```

This code was extracted [from the evaluator](https://github.com/elixir-lang/elixir/blob/master/lib/iex/lib/iex/evaluator.ex#L204), which should be changed to call a configurable MFArgs. The options are `[file: "iex", line: iex_state.counter]`. If the parser raises, the buffer is reset to an empty string. 

The parser shall be configurable via `IEx.configure`. A PR is welcome!

wojtekmach:
I believe this can be closed thanks to #10394.

PR #10394, opened 2020-10-05
Make IEx parser customizable

jrusso1020:
Fixes issue #10390 

Makes the IEx parser configurable through the IEx config so that
you can define your own parser if you wish.

### DIFF
--- a/lib/iex/lib/iex.ex
+++ b/lib/iex/lib/iex.ex
@@ -356,6 +356,7 @@ defmodule IEx do
     * `:continuation_prompt`
     * `:alive_prompt`
     * `:alive_continuation_prompt`
+    * `:parser`
 
   They are discussed individually in the sections below.
 
@@ -452,6 +453,18 @@ defmodule IEx do
     * `%prefix`  - a prefix given by `IEx.Server`
     * `%node`    - the name of the local node
 
+  ## Parser
+
+  This is an option determining the parser to use for IEx.
+
+  The parser is a "mfargs", which is a tuple with three elements:
+  the module name, the function name, and extra arguments to
+  be appended. The parser receives at least three arguments, the
+  current input as a string, the parsing options as a keyword list,
+  and the buffer as a string. It must return `{:ok, expr, buffer}`
+  or `{:incomplete, buffer}`.
+
+  If the parser raises, the buffer is reset to an empty string.
   """
   @spec configure(keyword()) :: :ok
   def configure(options) do
--- a/lib/iex/lib/iex/evaluator.ex
+++ b/lib/iex/lib/iex/evaluator.ex
@@ -213,33 +213,61 @@ defmodule IEx.Evaluator do
   # The first two clauses provide support for the break-trigger allowing to
   # break out from a pending incomplete expression. See
   # https://github.com/elixir-lang/elixir/issues/1089 for discussion.
-  @break_trigger '#iex:break\n'
+  @break_trigger "#iex:break\n"
 
-  defp eval(code, iex_state, state) do
-    try do
-      do_eval(String.to_charlist(code), iex_state, state)
-    catch
-      kind, error ->
-        print_error(kind, error, __STACKTRACE__)
-        {%{iex_state | cache: ''}, state}
+  @doc false
+  def parse(@break_trigger, _opts, "") do
+    {:incomplete, ""}
+  end
+
+  @doc false
+  def parse(@break_trigger, opts, _buffer) do
+    :elixir_errors.parse_error([line: opts[:line]], opts[:file], "incomplete expression", "")
+  end
+
+  @doc false
+  def parse(input, opts, buffer) do
+    input = buffer <> input
+
+    case Code.string_to_quoted(input, opts) do
+      {:ok, forms} ->
+        {:ok, forms, ""}
+
+      {:error, {_, _, ""}} ->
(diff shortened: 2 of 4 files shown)


jrusso1020:
Still need to add new tests but @josevalim just wanted to see if I was going in the right direction with this before I spent some time writing tests for it

lib/iex/lib/iex.ex
```diff
@@ -452,6 +453,14 @@ defmodule IEx do
...
     * `%node`    - the name of the local node
 
+  ## Parser
+  This is an option determining the parser to use for IEx
+
+  The parser is a MFArgs({M :: module(), F :: atom(), A :: [term()] | undefined})
+  which receives the current input, the parsing options,
+  and the buffer. It returns either `{:ok, expr, buffer}` or `{:incomplete, buffer}`.
+
+  If the parser raises, the buffer is reset to an empty string.
```
josevalim, line 463:
```suggestion
  ## Parser
  
  This is an option determining the parser to use for IEx.

  The parser is a "mfargs", which is a tuple with three elements:
  the module name, the function name, and extra arguments to
  be appended. The parser receives at least three argument, the
  current input as a string, the parsing options as keyword list,
  and the buffer as a string. It must return `{:ok, expr, buffer}`
  or `{:incomplete, buffer}`.

  If the parser raises, the buffer is reset to an empty string.
```

lib/iex/lib/iex/evaluator.ex
```diff
@@ -209,12 +209,6 @@ defmodule IEx.Evaluator do
...
   # Returns updated state.
-  #
-  # The first two clauses provide support for the break-trigger allowing to
-  # break out from a pending incomplete expression. See
-  # https://github.com/elixir-lang/elixir/issues/1089 for discussion.
-  @break_trigger '#iex:break\n'
-
   defp eval(code, iex_state, state) do
     try do
       do_eval(String.to_charlist(code), iex_state, state)
```
josevalim, line 214: Don't call `String.to_charlist(code)` here. Instead, pass the `code` as a string to the parser. Note you will have to change the default value of the `cache` key in the state to be `""` instead of `''`. I also recommend changing the key from `cache` to `buffer`.

lib/iex/lib/iex/evaluator.ex
```diff
@@ -225,21 +219,21 @@ defmodule IEx.Evaluator do
...
+    {parser_module, parse_method, additional_args} = IEx.Config.parser()
+    arguments = [latest_input, [line: line, file: "iex"], iex_state.cache | additional_args]
+
+    case apply(parser_module, parse_method, arguments) do
+      {:ok, forms, ""} ->
+        handle_eval({:ok, forms}, code, line, iex_state, state)
+
+      {:incomplete, _input} ->
+        handle_eval(:incomplete, code, line, iex_state, state)
+    end
```
josevalim, line 236: This should rather be something like:

```suggestion
    case apply(parser_module, parse_method, arguments) do
      {:ok, forms, ""} ->
        iex_state = %{iex_state | buffer: buffer, counter: iex_state.counter + 1}
        {iex_state, handle_eval(forms, code, line, state)}

      {:incomplete, buffer} ->
        {%{iex_state | buffer: buffer}, state}
    end
```

ANd then `handle_eval` can be simplified a bit. :)

lib/iex/lib/iex/parser.ex
```diff
@@ -0,0 +1,36 @@
...
+  The base implementation of the parser we use before evaluating iex code
+
+  This parser can be replaced by setting the `:parser` attribute of the iex
+  config
+  """
+
+  # The first two clauses provide support for the break-trigger allowing to
+  # break out from a pending incomplete expression. See
+  # https://github.com/elixir-lang/elixir/issues/1089 for discussion.
+  @break_trigger '#iex:break\n'
```
josevalim, line 12: Note that this needs to be a string:

```suggestion
  @break_trigger "#iex:break\n"
```

lib/iex/lib/iex/parser.ex
```diff
@@ -0,0 +1,36 @@
+defmodule IEx.Parser do
```
josevalim, line 1: There is no need for this to be a separate module. You can keep this function in the `IEx.Evaluator` module as long as it is public with `@doc false`. :)

josevalim (commented):
Awesome job @jrusso1020! I have added some comments to tidy this up. :)

lib/iex/lib/iex/evaluator.ex
wrachwal: hi @josevalim  @jrusso1020 , some parsers may be oriented more to give side effects than produce any sensible elixir values (they even may not have much common with elixir). Returning `IEx.dont_display_result/0` does it, but bringing it to API would not be nice. Maybe consider to replace `:ok` tuple with `:value` and `:no_value` result tuples (you may propose better names)?
Seems the parser feature will become available sooner than later. :+1:

josevalim: I would prefer to explicitly return `IEx.dont_display_result` for now. Especially because I think you should avoid actually doing work in the parser. Instead you should return quoted expressions.

jrusso1020:
@josevalim Thanks for all the feedback! I think I've addressed all the issues you brought up, should be ready for another review

josevalim:

PR #10398, opened 2020-10-06
Add @doc :since to List.{first,last}/1-2

eksperimental:
It addresses the issues mentioned in https://github.com/elixir-lang/elixir/pull/10391#issuecomment-703944085

### DIFF
--- a/lib/elixir/lib/list.ex
+++ b/lib/elixir/lib/list.ex
@@ -255,15 +255,12 @@ defmodule List do
   end
 
   @doc """
-  Returns the first element in `list` or `default` if `list` is empty.
+  Returns the first element in `list` or `nil` if `list` is empty.
 
   ## Examples
 
       iex> List.first([])
       nil
-
-      iex> List.first([], 1)
-      1
 
       iex> List.first([1])
       1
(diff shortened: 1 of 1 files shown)


Eiji7:
@eksperimental If we want to be `strict` on `since` then `1.0.0` should be changed to `0.12.2`:
https://github.com/elixir-lang/elixir/blob/ddc6b48a4988ea596456eb7b2129269b33d0a99f/lib/elixir/lib/list.ex#L119-L134

Related commit: https://github.com/elixir-lang/elixir/commit/42fc70a0eb66315f79acbc491d61d23e600d2a11

> Deprecate Enum.first/1 in favor of Enum.at/2 and List.first/1
> 
> The main reason for this change is because `Enum.first/1` does
not communicate that enumeration will happen, in contrast to
functions like Enum.at/2 and Enum.fetch/2 which are meant to
be used with non-indexed collections and the need for enumeration
is made a bit more obvious.

eksperimental:
> @eksperimental If we want to be **strict** on first version then `1.0.0` should be changed to `0.12.2`:
> 
> https://github.com/elixir-lang/elixir/blob/ddc6b48a4988ea596456eb7b2129269b33d0a99f/lib/elixir/lib/list.ex#L119-L134
> 
> > ### Deprecate Enum.first/1 in favor of Enum.at/2 and List.first/1
> > The main reason for this change is because `Enum.first/1` does
> > not communicate that enumeration will happen, in contrast to
> > functions like Enum.at/2 and Enum.fetch/2 which are meant to
> > be used with non-indexed collections and the need for enumeration
> > is made a bit more obvious.
> > Source: [42fc70a](https://github.com/elixir-lang/elixir/commit/42fc70a0eb66315f79acbc491d61d23e600d2a11)

We don't document versions prior to 1.0.0

Eiji7:
ok, thanks - will remember that :+1:

Eiji7 (approved):

josevalim:
I wouldn't want to break this function apart because of since so I would at best add a note to the docs. "The default argument was added on Elixir v1.12". In any case, thanks!

PR #10399, opened 2020-10-06
Move printing test seed to beginning of test run

jrusso1020:
Came across this proposal on the forums to move printing the seed of the
test suite to the beginning of the test run. This was actually
something I wished for recently when encountering a weird issue
in our codebase which caused all of our non async tests to fail and
each synchronous test would take 60 seconds before timing out. With
a few hundred synchronous tests this took quite some time for the
suite to finish and I wanted to grab the seed to replicate it.

https://groups.google.com/g/elixir-lang-core/c/sP8DiY1zMFQ

## Example
```
==> ex_unit (ex_unit)

Randomized with seed 411332
................................................................................................................................................................................................................................................................................................................................................................................
```

### DIFF
--- a/lib/ex_unit/lib/ex_unit/cli_formatter.ex
+++ b/lib/ex_unit/lib/ex_unit/cli_formatter.ex
@@ -28,6 +28,7 @@ defmodule ExUnit.CLIFormatter do
   end
 
   def handle_cast({:suite_started, _opts}, config) do
+    IO.puts("\nRandomized with seed #{config.seed}")
     {:noreply, config}
   end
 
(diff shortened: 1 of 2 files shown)


smaximov:
I think it would be better if ex_unit printed the seed both at the beginning and at the end of test suite. For instance, if you have a bunch of failing tests, you wouldn't want to scroll all the way back to where tests started to find the seed (assuming the test run has already finished).

josevalim:
Thanks @jrusso1020 for reminding me of this issue. I have gone with a different approach in 285a6a6f8479fb14100ab8be2786f7ec88e5b9ab. If your test suite gets stuck, ExUnit now intercepts sigquit and it will show which tests have been aborted as well as a summary of the run alongside the seed. :) In any case, thanks for the PRs! :heart:

PR #10400, opened 2020-10-06
Update path.ex

princemaple:
fix invalid path

### DIFF
--- a/lib/elixir/lib/path.ex
+++ b/lib/elixir/lib/path.ex
@@ -207,7 +207,7 @@ defmodule Path do
 
       Path.type("D:/usr/local/bin") #=> :absolute
       Path.type("usr/local/bin")    #=> :relative
-      Path.type("D:bar.ex")         #=> :volumerelative
+      Path.type("D:/bar.ex")         #=> :volumerelative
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/path.ex
```diff
@@ -207,7 +207,7 @@ defmodule Path do
 
       Path.type("D:/usr/local/bin") #=> :absolute
       Path.type("usr/local/bin")    #=> :relative
-      Path.type("D:bar.ex")         #=> :volumerelative
+      Path.type("D:/bar.ex")         #=> :volumerelative
```
michallepicki, line 210:
```suggestion
      Path.type("D:/bar.ex")        #=> :volumerelative
```

princemaple: Good eyes.

josevalim: `D:bar.ex` is actually valid on Windows. More info here: https://docs.microsoft.com/en-us/dotnet/standard/io/file-path-formats

josevalim:
Thanks for the PR but the path *is* valid. :)

princemaple:
Woah, TIL. Thanks.

PR #10410, opened 2020-10-08
Fix returning errors in Module.Types.Expr.of_expr

michallepicki:
Fixes small mistake detected by dialyzer:

```
lib/module/types/expr.ex:257: The pattern 
          {'error', _reason@1} can never match the type 
          'error'

lib/module/types/expr.ex:281: The pattern 
          {'error', _reason@1} can never match the type 
          'error'
```


josevalim:

PR #10414, opened 2020-10-08
Improves the docs of config/runtime.exs and Config.config_env/0

### LINKED ISSUE #10413: Unclear docs about config/runtime.exs
The current docs for `config/runtime.exs` are

> For runtime configuration, you can use the config/runtime.exs file. It is executed after your Mix project is compiled and also before a release (assembled with mix release) starts.

I do not understand this sentence:

> It is executed after your Mix project is compiled

when a project is compiled, as in `mix compile` or when building a Docker image, anything executed is executed in the void, you see what I mean :).

Does that mean "before the application starts"?

fxn:
Fixes #10413.

### DIFF
--- a/lib/elixir/lib/config.ex
+++ b/lib/elixir/lib/config.ex
@@ -160,6 +160,8 @@ defmodule Config do
   end
 
   @doc """
+  Returns the environment this configuration file is executed on.
+
   In Mix projects this function returns the environment this configuration
   file is executed on. In releases, the environment when `mix release` ran.
 
--- a/lib/elixir/lib/config.ex
+++ b/lib/elixir/lib/config.ex
@@ -64,8 +64,8 @@ defmodule Config do
   ## config/runtime.exs
 
   For runtime configuration, you can use the `config/runtime.exs` file.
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/config.ex
```diff
@@ -160,7 +160,8 @@ defmodule Config do
   end
 
   @doc """
-  Returns the environemnt this configuration file is executed on.
+  In Mix projects this function returns the environment this configuration
```
josevalim, line 163: The first line should always be a short summary, as that is shown in tables and stuff, so what do you think about this?

```suggestion
  Returns the environment this configuration file is executed on.

  In Mix projects this function returns the environment this configuration
```

fxn: Ah, forgot that! Sure, let me rebase in a while!

josevalim: There is no need to rebase, we always squash it anyway. :) Thanks!

josevalim:

fxn:
❤️

PR #10416, opened 2020-10-10
Rename a variable in Inspect documentation

qcam:
I think, to avoid confusion of MapSet being a Dict.


josevalim:

PR #10419, opened 2020-10-11
Implements Enum.zip_with and Stream.zip_with

Adzz:
This is my first stab at this so comments very welcome. 

Feature Request thread is here: https://groups.google.com/u/1/g/elixir-lang-core/c/mwBRUILsVHA

My basic approach was that the current zip/1 and zip/2 functions can be implemented by zip_with by doing this:

```elixir
zip(enums) do
  zip_with(enums, &List.to_tuple/1)
end
```

To implement zip_with/2 the zip_fun you pass to it will get a list of each of the elements from each enumerable we are zipping. This seemed okay, but it differs from what zip/3 does, where each element from the collection being zipped is a different argument to the zipping fun.

I think that's okay but happy to change them to both be consistent with each other.

### DIFF
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -3214,6 +3214,54 @@ defmodule Enum do
     |> :lists.reverse()
   end
 
+  @doc """
+  Zips two collections together by passing the corresponding element from each collection to the
+  provided fun. Returns a new list that contains the result of that fun for each pair of elements.
+
+  The zipping finishes as soon as any enumerable in the given collection completes.
+
+  ## Examples
+
+      iex> Enum.zip_with([1, 2], [3, 4], fn x, y -> x + y end)
+      [4, 6]
+
+      iex> Enum.zip_with([1, 2], [3, 4, 5, 6], fn x, y -> x + y end)
+      [4, 6]
+
+      iex> Enum.zip_with([1, 2, 5, 6], [3, 4], fn x, y -> x + y end)
+      [4, 6]
+  """
+  @spec zip_with(t, t, (term, term -> term)) :: [term]
+  def zip_with(enumerable1, enumerable2, fun)
+      when is_list(enumerable1) and is_list(enumerable2) do
+    zip_list(enumerable1, enumerable2, fun)
+  end
+
+  def zip_with(enumerable1, enumerable2, fun) do
+    zip_with([enumerable1, enumerable2], fun)
+  end
+
+  @doc """
+  Zips corresponding elements from a finite collection of enumerables into a new
+
+  Zips two collections together by passing the corresponding element from each collection to the
+  provided fun. Returns a new list that contains the result of that fun for each pair of elements.
+
+  The zipping finishes as soon as any enumerable in the given collection completes.
+
+  ## Examples
+
+      iex> Enum.zip_with([[1, 2], [3, 4], [5, 6]], fn [x, y, z] -> x + y + z end)
+      [9, 12]
+  """
+  @spec zip_with(t, (term, term -> term)) :: [term]
+  def zip_with([], _fun), do: []
+
+  def zip_with(enumerables, fun) do
+    Stream.zip_with(enumerables, fun)
+    |> Enum.to_list()
+  end
+
   ## Helpers
 
   @compile {:inline, entry_to_string: 1, reduce: 3, reduce_by: 3, reduce_enumerable: 3}
@@ -3744,13 +3792,16 @@ defmodule Enum do
   end
 
   ## zip
-
-  defp zip_list([h1 | next1], [h2 | next2]) do
-    [{h1, h2} | zip_list(next1, next2)]
+  defp zip_list(enumerable1, enumerable2) do
+    zip_list(enumerable1, enumerable2, fn x, y -> {x, y} end)
   end
 
-  defp zip_list(_, []), do: []
-  defp zip_list([], _), do: []
+  defp zip_list([h1 | next1], [h2 | next2], fun) do
+    [fun.(h1, h2) | zip_list(next1, next2, fun)]
+  end
+
+  defp zip_list(_, [], _fun), do: []
+  defp zip_list([], _, _fun), do: []
 end
 
 defimpl Enumerable, for: List do
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -3242,12 +3242,9 @@ defmodule Enum do
   end
 
   @doc """
-  Zips corresponding elements from a finite collection of enumerables into a new
-
-  Zips two collections together by passing the corresponding element from each collection to the
-  provided fun. Returns a new list that contains the result of that fun for each pair of elements.
-
-  The zipping finishes as soon as any enumerable in the given collection completes.
+  Zips corresponding elements from a finite collection of enumerables into a new enumerable which
+  is built by calling zip_fun with the first element from each enumerable, then the with the
+  second element and so on until any one of the enums runs out of elements.
 
   ## Examples
 
@@ -3257,8 +3254,8 @@ defmodule Enum do
   @spec zip_with(t, (term, term -> term)) :: [term]
   def zip_with([], _fun), do: []
 
-  def zip_with(enumerables, fun) do
-    Stream.zip_with(enumerables, fun)
+  def zip_with(enumerables, zip_fun) do
+    Stream.zip_with(enumerables, zip_fun)
     |> Enum.to_list()
   end
 
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -3215,10 +3215,13 @@ defmodule Enum do
   end
 
   @doc """
-  Zips two collections together by passing the corresponding element from each collection to the
-  provided fun. Returns a new list that contains the result of that fun for each pair of elements.
+  Zips corresponding elements from two enumerables into a new enumerable, transforming them with
+  zip_fun as it goes.
 
-  The zipping finishes as soon as any enumerable in the given collection completes.
+  The corresponding elements from each collection are passed to the provided zip_fun in turn.
+  Returns a new list that contains the result of calling zip_fun for each pair of elements.
+
+  The zipping finishes as soon as any enumerable runs out of elements.
 
   ## Examples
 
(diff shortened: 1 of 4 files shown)


lib/elixir/lib/enum.ex
```diff
@@ -3214,6 +3214,51 @@ defmodule Enum do
...
+
+      iex> Enum.zip_with([[1, 2], [3, 4], [5, 6]], fn [x, y, z] -> x + y + z end)
+      [9, 12]
+  """
+  @spec zip_with(t, (term, term -> term)) :: [term]
+  def zip_with([], _fun), do: []
+
+  def zip_with(enumerables, zip_fun) do
+    Stream.zip_with(enumerables, zip_fun)
+    |> Enum.to_list()
```
Adzz, line 3259: I think this is equivalent, but the Stream code lost me a bit...

krainboltgreene:
It would be useful to note that this is an endofunctor (the container type you put in is the container type you pull out).

lib/elixir/lib/enum.ex
```diff
@@ -3214,6 +3214,51 @@ defmodule Enum do
...
+      [4, 6]
+  """
+  @spec zip_with(t, t, (term, term -> term)) :: [term]
+  def zip_with(enumerable1, enumerable2, fun)
+      when is_list(enumerable1) and is_list(enumerable2) do
+    zip_list(enumerable1, enumerable2, fun)
+  end
+
+  def zip_with(enumerable1, enumerable2, fun) do
+    zip_with([enumerable1, enumerable2], fun)
```
josevalim, line 3241: The following example will fail:

    iex> Enum.zip_with([1, 2, 5, 6], 3..4, fn x, y -> x + y end)

That's because we call `zip_with` and `zip_with` with pass a list to the `fn`.  You can fix it by calling it like this:

```suggestion
    zip_with([enumerable1, enumerable2], &apply(fun, &1))
```

You can also take this approach on `Stream.zip_with/3` so it calls `fun` with two args.

josevalim: Btw, please add tests mixing lists with ranges and similar.

josevalim:
This looks great, awesome job! Please note that the first paragraph of `@doc` must always be a one-line summary explaining what the function does. Can you please update them accordingly? Thanks.

lib/elixir/lib/enum.ex
```diff
@@ -3214,6 +3214,51 @@ defmodule Enum do
     |> :lists.reverse()
   end
 
+  @doc """
+  Zips two collections together by passing the corresponding element from each collection to the
```
eksperimental, line 3218: We use the term "enumerables" in this module, instead of collections.

lib/elixir/lib/stream.ex
```diff
@@ -1137,46 +1155,67 @@ defmodule Stream do
   @doc since: "1.4.0"
   @spec zip(enumerables) :: Enumerable.t() when enumerables: [Enumerable.t()] | Enumerable.t()
   def zip(enumerables) do
-    &prepare_zip(enumerables, &1, &2)
+    zip_fun = fn yielded_elems -> List.to_tuple(yielded_elems) end
+    zip_with(enumerables, zip_fun)
+  end
+
+  @doc """
+  Lazily passes a list to zip_fun containing the first element from each of the enums in enumerables.
```
eksperimental, line 1163:
```suggestion
  Lazily passes a list to function `zip_fun` containing the first element from each of the enumerable from the `enumerables` list.
```

lib/elixir/lib/stream.ex
```diff
@@ -1137,46 +1155,67 @@ defmodule Stream do
...
+  Then calls zip_fun with the second element from each of the enums and so on until any one of the
```
eksperimental, line 1164:
```suggestion
  Then calls `zip_fun` with the second element from each of the enumerables and so on until any one of the
```

lib/elixir/lib/stream.ex
```diff
@@ -1137,46 +1155,67 @@ defmodule Stream do
...
+  enums runs out of elements. A new enumerable is lazily built up in this way containing the
+  result of calling that zip_fun repeatedly.
```
eksperimental, line 1166:
```suggestion
  result of calling that `zip_fun` repeatedly.
```

lib/elixir/lib/enum.ex
```diff
@@ -3214,6 +3214,51 @@ defmodule Enum do
...
+  provided fun. Returns a new list that contains the result of that fun for each pair of elements.
```
eksperimental, line 3219:
```suggestion
  provided fun. Returns a new list that contains the result of `fun` for each pair of elements.
```

lib/elixir/lib/enum.ex
```diff
@@ -3744,13 +3789,16 @@ defmodule Enum do
...
 
   ## zip
+  defp zip_list(enumerable1, enumerable2) do
+    zip_list(enumerable1, enumerable2, fn x, y -> {x, y} end)
+  end
 
-  defp zip_list([h1 | next1], [h2 | next2]) do
-    [{h1, h2} | zip_list(next1, next2)]
+  defp zip_list([h1 | next1], [h2 | next2], fun) do
+    [fun.(h1, h2) | zip_list(next1, next2, fun)]
```
eksperimental, line 3797: `head1` and `head2` instead of `h1` and `h2`

eksperimental (changes requested):

lib/elixir/lib/enum.ex
```diff
@@ -3214,6 +3214,51 @@ defmodule Enum do
...
+    zip_list(enumerable1, enumerable2, fun)
+  end
+
+  def zip_with(enumerable1, enumerable2, fun) do
+    zip_with([enumerable1, enumerable2], fun)
+  end
+
+  @doc """
+  Zips corresponding elements from a finite collection of enumerables into a new enumerable which
+  is built by calling zip_fun with the first element from each enumerable, then the with the
```
eksperimental, line 3246: "then the with the second element"
I think a word or more are missing

lib/elixir/lib/enum.ex
```diff
@@ -3214,6 +3214,51 @@ defmodule Enum do
...
+  second element and so on until any one of the enums runs out of elements.
+
+  ## Examples
+
+      iex> Enum.zip_with([[1, 2], [3, 4], [5, 6]], fn [x, y, z] -> x + y + z end)
+      [9, 12]
+  """
+  @spec zip_with(t, (term, term -> term)) :: [term]
```
eksperimental, line 3254: I think we should defined first zip_with/2 and then zip_with/3

eksperimental (changes requested):

lib/elixir/lib/enum.ex
Adzz: Do you mean move the zip_with/2 function above the zip_with/3 definition in this file?

josevalim: Don't worry, the order is fine as is. 👍 

@eksperimental, please don't worry too much about these small details in PRs. I would rather focus on the correctness of the implementation and quality of the documentation, like you did above, than ordering of functions. :)

eksperimental: > Do you mean move the zip_with/2 function above the zip_with/3 definition in this file?

Yes, but as José said it does not really matter for now.

lib/elixir/lib/enum.ex
Adzz: Nice spot! Will do.

What's your thoughts on having zip_with/2's zipping fun receive a list instead?
Essentially turn this line:  https://github.com/elixir-lang/elixir/pull/10419/files#diff-6881431a92cd4e3ea0de82bf2338f8eaR3797
into this:
```elixir 
  defp zip_list([h1 | next1], [h2 | next2], fun) do
   # pass a list here to fun
    [fun.([h1, h2]) | zip_list(next1, next2, fun)]
  end
```

I feel like it will be more consistent

lib/elixir/lib/enum.ex
whatyouhide, line 3254:
```suggestion
  @doc since: "1.12.0"
  @spec zip_with(t, (term, term -> term)) :: [term]
```

lib/elixir/lib/enum.ex
```diff
@@ -3214,6 +3214,51 @@ defmodule Enum do
...
+      iex> Enum.zip_with([1, 2], [3, 4], fn x, y -> x + y end)
+      [4, 6]
+
+      iex> Enum.zip_with([1, 2], [3, 4, 5, 6], fn x, y -> x + y end)
+      [4, 6]
+
+      iex> Enum.zip_with([1, 2, 5, 6], [3, 4], fn x, y -> x + y end)
+      [4, 6]
+  """
+  @spec zip_with(t, t, (term, term -> term)) :: [term]
```
whatyouhide, line 3234:
```suggestion
  @doc since: "1.12.0"
  @spec zip_with(t, t, (term, term -> term)) :: [term]
```

lib/elixir/lib/stream.ex
```diff
@@ -1120,6 +1120,24 @@ defmodule Stream do
...
+  and so on until either one of the enumerable's completes.
+
+  ## Examples
+
+      iex> concat = Stream.concat(1..3, 4..6)
+      iex> Stream.zip_with(concat, concat, fn [a, b] -> a + b end) |> Enum.to_list()
+      [2, 4, 6, 8, 10, 12]
+
+  """
+  @spec zip_with(Enumerable.t(), Enumerable.t(), (Enumerable.t() -> term)) :: Enumerable.t()
```
whatyouhide, line 1138:
```suggestion
  @doc since: "1.12.0"
  @spec zip_with(Enumerable.t(), Enumerable.t(), (Enumerable.t() -> term)) :: Enumerable.t()
```

lib/elixir/lib/stream.ex
```diff
@@ -1137,46 +1155,67 @@ defmodule Stream do
...
+  enums runs out of elements. A new enumerable is lazily built up in this way containing the
+  result of calling that zip_fun repeatedly.
+
+  ## Examples
+
+      iex> concat = Stream.concat(1..3, 4..6)
+      iex> Stream.zip_with([concat, concat], fn [a, b] -> a + b end) |> Enum.to_list()
+      [2, 4, 6, 8, 10, 12]
+  """
+  @spec zip_with(enumerables, (Enumerable.t() -> term)) :: Enumerable.t()
```
whatyouhide, line 1174:
```suggestion
  @doc since: "1.12.0"
  @spec zip_with(enumerables, (Enumerable.t() -> term)) :: Enumerable.t()
```

Adzz:
@krainboltgreene I don't think it is an endofunctor... take this example, you zip two maps but get returned a list. Like lots of the enum functions.

```elixir
Enum.zip_with(%{a: 1}, %{a: 2}, fn {_, left}, {_, right} -> left + right end)
# [3]
```

josevalim:

PR #10424, opened 2020-10-13
Fix spec for Enum.zip_with/2

michallepicki:
Wrong spec was detected by Dialyzer:

    lib/enum.ex:3248: The call 'Elixir.Enum':zip_with
             ([any(), ...],
              fun((_) -> any())) breaks the contract 
              (t(), fun((term(), term()) -> term())) -> [term()]


josevalim:

PR #10428, opened 2020-10-14
Sort Stream.{zip,zip_with} functions

eksperimental:
As they were interspersed before.


josevalim:
It is fine as is, thank you! :)

PR #10429, opened 2020-10-15
Map.get_indifferent/3 returns a key's value, whether the key is…

JamesLavin:
…present as a string or an atom

Long-time Elixir user/fan. First-time caller. I want to thank everyone involved in creating & maintaining Elixir!!! You have greatly improved my programming life.

I found myself writing some messy code today because I didn't know whether the data structure I was given had string or atom keys. I hope this might be useful to others.

I have tried to minimize the probability that calling this function would create a new atom, but I would totally understand if this PR gets rejected for this reason.

There is no way to create a `Keyword` equivalent for this function because keyword lists have only atoms as their first tuple elements.

Cheers,

James

### DIFF
--- a/lib/elixir/lib/map.ex
+++ b/lib/elixir/lib/map.ex
@@ -439,7 +439,7 @@ defmodule Map do
   @doc """
   Gets the value for a specific `key` in `map`.
 
-  If `key` is present in `map` then its value `value` is
+  If `key` is present in `map`, then its value `value` is
   returned. Otherwise, `default` is returned.
 
   If `default` is not provided, `nil` is used.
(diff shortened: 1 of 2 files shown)


NickNeck:
Hello @JamesLavin 
 
At first, there is a mailing list https://groups.google.com/forum/#!forum/elixir-lang-core for suggestions and proposals.

I think not knowing which type of key is used in a map is an edge case. Therefore a function like `get_indifferent` is not needed in the standard library.

You are using in your code `String.to_atom`, that can leads to some problems. See the warning in the docs:
> Warning: this function creates atoms dynamically and atoms are not garbage-collected. Therefore, string should not be an untrusted value, such as input received from a socket or during a web request. Consider using to_existing_atom/1 instead.

josevalim:
Hi @JamesLavin! Thanks for the PR! As @NickNeck suggested, this is better if discussed in the mailing list, although I believe this has been discussed in the past and rejected. My suggestion is for you to define a MapIndifferent module (or even a lib - although one may already exist) that you can reuse without a need to add it directly to Elixir. Thanks!

PR #10432, opened 2020-10-16
Fix invalid example for using erlang logger handler

albertored:
The example for using a custom erlang logger handler is invalid, the handler must be enclosed in a tuple.
In version 1.11 it is already fixed for people still using 1.10 the documentation can be misreading.


josevalim:

PR #10434, opened 2020-10-17
Set compiler options from Mix project when running tests

wojtekmach:
Ref: https://github.com/elixir-lang/elixir/pull/10255

Turns out that when `mix test` starts, it has the default compiler options, ignoring `:elixirc_options`. Notably, setting:

    elixirc_options: [warnings_as_errors: true]

doesn't have the desired effect of causing the task to fail when it encounter warnings _in tests_.

(Unfortunately, turns out the advice given in linked PR, `MIX_ENV=test mix do compile --warnings-as-errors, test` doesn't work either.)

This patch tries to fix that.

### DIFF
--- a/lib/mix/lib/mix/tasks/test.ex
+++ b/lib/mix/lib/mix/tasks/test.ex
@@ -330,6 +330,7 @@ defmodule Mix.Tasks.Test do
 
     Mix.Task.run("compile", args)
     project = Mix.Project.config()
+    Code.compiler_options(project[:elixirc_options] || [])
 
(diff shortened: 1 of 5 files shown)


wojtekmach:
Spoken with José and there was a concern about `elixirc_options` leaking to other parts of the system. We'll try the `mix test --warnings-as-errors` route that we discussed before but we'd document explicitly that the option is only for tests and if you want to warn on your compiled code, you'd need to `mix compile --warnings-as-errors`.

PR #10435, opened 2020-10-17
Add --warnings-as-errors option to mix test

wojtekmach:
Another stab at #10255 (also see https://github.com/elixir-lang/elixir/pull/10434#issuecomment-710783006)


lib/mix/lib/mix/tasks/test.ex
```diff
@@ -328,7 +358,15 @@ defmodule Mix.Tasks.Test do
     # Load ExUnit before we compile anything
     Application.ensure_loaded(:ex_unit)
 
+    old_warnings_as_errors = Code.compiler_options()[:warnings_as_errors]
```
josevalim, line 361:
```suggestion
    old_warnings_as_errors = Code.get_compiler_option(:warnings_as_errors)
```

lib/mix/test/mix/tasks/test_test.exs
```diff
@@ -408,6 +408,41 @@ defmodule Mix.Tasks.TestTest do
     end
   end
 
+  describe "--warnings-as-errors" do
```
josevalim, line 411: We probably don't need those tests to go through the CLI.

lib/mix/test/mix/tasks/test_test.exs
```diff
@@ -408,6 +408,26 @@ defmodule Mix.Tasks.TestTest do
...
+    test "fail on warning in tests" do
+      in_fixture("test_stale", fn ->
+        File.write!("test/warning_test.exs", """
```
josevalim, line 414: Let's also write a file in lib with:

    unused_compile = 13

And we will check it warns but still run tests. :)

josevalim:

PR #10439, opened 2020-10-17
Do not use private and removed ram_file:compress/1

josevalim:


josevalim:

PR #10442, opened 2020-10-18
Fix stop_supervised/1 for temporary children

### LINKED ISSUE #10440: `ExUnit.Callbacks.stop_supervised!` fails if the child is temporary
### Environment

* Elixir & Erlang/OTP versions (elixir --version): Elixir 1.11.1, OTP 23.1
* Operating system: Ubuntu

### Current behavior

`ExUnit.Callbacks.stop_supervised!` fails if the child is temporary:

```elixir
test "foo" do
  start_supervised!({Agent, fn -> :ok end}, id: :foo, restart: :temporary)
  stop_supervised!(:foo)
end

# ** (RuntimeError) could not stop child ID :foo because it was not found
```

### Expected behavior

Since a child is running, I'd expect `stop_supervised!` to succeed. I traced the problem to [this line](https://github.com/elixir-lang/elixir/blob/v1.11.1/lib/ex_unit/lib/ex_unit/callbacks.ex#L467). I presume that the supervisor is dynamic (or `:simple_one_for_one`), in which case a spec of the temporary child is removed once the child is stopped. Therefore I think that ignoring the result of `delete_child` should be enough to solve the problem.

sasa1977:
resolves #10440

### DIFF
--- a/lib/ex_unit/lib/ex_unit/callbacks.ex
+++ b/lib/ex_unit/lib/ex_unit/callbacks.ex
@@ -463,9 +463,12 @@ defmodule ExUnit.Callbacks do
         {:error, :not_found}
 
       {:ok, sup} ->
-        with :ok <- Supervisor.terminate_child(sup, id),
-             :ok <- Supervisor.delete_child(sup, id),
-             do: :ok
+        with :ok <- Supervisor.terminate_child(sup, id) do
+          # If the terminated child was temporary, delete_child returns {:error, :not_found}.
+          # Since the child was successfully terminated, we treat this result as a success.
+          true = Supervisor.delete_child(sup, id) in [:ok, {:error, :not_found}]
+          :ok
+        end
 
       :error ->
         raise ArgumentError, "stop_supervised/1 can only be invoked from the test process"
--- a/lib/ex_unit/test/ex_unit/supervised_test.exs
+++ b/lib/ex_unit/test/ex_unit/supervised_test.exs
@@ -94,6 +94,12 @@ defmodule ExUnit.SupervisedTest do
     refute Process.alive?(pid)
   end
 
+  test "stops a temporary supervised process" do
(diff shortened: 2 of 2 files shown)


lib/ex_unit/lib/ex_unit/callbacks.ex
```diff
@@ -463,9 +463,12 @@ defmodule ExUnit.Callbacks do
         {:error, :not_found}
 
       {:ok, sup} ->
-        with :ok <- Supervisor.terminate_child(sup, id),
-             :ok <- Supervisor.delete_child(sup, id),
-             do: :ok
+        with :ok <- Supervisor.terminate_child(sup, id) do
+          # If the terminated child was temporary, delete_child returns {:error, :not_found}.
+          # Since the child was successfully terminated, we treat this result as a success.
+          true = Supervisor.delete_child(sup, id) in [:ok, {:error, :not_found}]
```
josevalim, line 469: This changes the behavior of stop_supervised (without bang), as in that it case could return one of the other error status and now it will raise. What do you think? Even if we want to raise, the error message now is a bit worse.

sasa1977: I did it to be consistent with the typespec, which states we can only return `:ok | {:error, :not_found}`. I also don't see any scenario where other kinds of `delete_child` errors can be returned. As [docs](https://erlang.org/doc/man/supervisor.html#delete_child-2) state:

> If successful, the function returns ok. If the child specification identified by Id exists but the corresponding child process is running or is about to be restarted, the function returns {error,running} or {error,restarting}, respectively. If the child specification identified by Id does not exist, the function returns {error,not_found}.

So this is basically just a sanity check. If this raises, I think it can only be due to some bug/wrong assumption on our behalf, and I don't see what kind of useful message could we produce in this case.

josevalim:

Issue #9835, opened 2020-02-22
Starting release results in: No such file or directory error


axelclark:
### Environment

* Elixir & Erlang/OTP versions (elixir --version): 
```
$ asdf current
elixir         1.10.1-otp-22 (set by /Users/jackclark/.tool-versions)
erlang         22.2.6   (set by /Users/jackclark/.tool-versions)
```
* Operating system: `MacOS Catalina 10.15.3`

### Current behavior

Create new project & verify test passes:
```bash
$ mix new hello_release
* creating README.md
* creating .formatter.exs
* creating .gitignore
* creating mix.exs
* creating lib
* creating lib/hello_release.ex
* creating test
* creating test/test_helper.exs
* creating test/hello_release_test.exs

Your Mix project was created successfully.
You can use "mix" to compile it, test it, and more:

    cd hello_release
    mix test

Run "mix help" for more commands.
$ cd hello_release 

$ mix test
Compiling 1 file (.ex)
Generated hello_release app
..

Finished in 0.04 seconds
1 doctest, 1 test, 0 failures

Randomized with seed 567536
```

Create release & try to start:
```bash
$ MIX_ENV=prod mix release
Compiling 1 file (.ex)
Generated hello_release app
* assembling hello_release-0.1.0 on MIX_ENV=prod
* skipping runtime configuration (config/releases.exs not found)

Release created at _build/prod/rel/hello_release!

    # To start your system
    _build/prod/rel/hello_release/bin/hello_release start

Once the release is running:

    # To connect to it remotely
    _build/prod/rel/hello_release/bin/hello_release remote

    # To stop it gracefully (you may also send SIGINT/SIGTERM)
    _build/prod/rel/hello_release/bin/hello_release stop

To list all commands:

    _build/prod/rel/hello_release/bin/hello_release

$ _build/prod/rel/hello_release/bin/hello_release start
cut: /Users/jackclark/workspace/hello_release/_build/prod/rel/hello_release
/Users/jackclark/workspace/hello_release/_build/prod/rel/hello_release/releases/start_erl.data: No such file or directory
_build/prod/rel/hello_release/bin/hello_release: line 14: /Users/jackclark/workspace/hello_release/_build/prod/rel/hello_release
/Users/jackclark/workspace/hello_release/_build/prod/rel/hello_release/releases//env.sh: No such file or directory
```
Two additional notes:
[...]

PR #10445, opened 2020-10-20
Unset CDPATH while finding RELEASE_ROOT

garthk:
I keep running into #9835 when testing releases locally. This change fixes it without user intervention.

I considered piping the output of `cd` into `/dev/null`, but it struck me `CDPATH=''` was cleaner. WDYT?


josevalim:

PR #10447, opened 2020-10-20
Improve Config docs

cw789:
Add missing config_env/0 & config_target/0
to the list of imports of `import Config`.

But I'm not 100% sure about that...

### DIFF
--- a/lib/elixir/lib/config.ex
+++ b/lib/elixir/lib/config.ex
@@ -16,7 +16,8 @@ defmodule Config do
       import_config "#{config_env()}.exs"
 
   `import Config` will import the functions `config/2`, `config/3`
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/config.ex
```diff
@@ -16,7 +16,8 @@ defmodule Config do
       import_config "#{config_env()}.exs"
 
   `import Config` will import the functions `config/2`, `config/3`
-  and `import_config/1` to help you manage your configuration.
+  `config_env/0`, `config_target/0` and `import_config/1`
```
josevalim, line 19:
```suggestion
  `config_env/0`, `config_target/0`, and `import_config/1`
```

josevalim:

PR #10454, opened 2020-10-25
Make Code.ensure_loaded/compiled error return consistent on CS/CI FS

lukaszsamson:
On case sensitive filesystems
```
Code.ensure_loaded(Elixir)
{:error, :nofile}
```
On case insensitive filesystems
```
Code.ensure_loaded(Elixir)
{:error, :badfile}
12:33:13.307 [error] Loading of /Users/elixir/bin/../lib/elixir/ebin/Elixir.beam failed: :badfile

 
12:33:13.307 [error] beam/beam_load.c(1440): Error loading module 'Elixir':
  BEAM file exists but it defines a module named elixir

```

### DIFF
--- a/lib/elixir/lib/code.ex
+++ b/lib/elixir/lib/code.ex
@@ -1224,9 +1224,14 @@ defmodule Code do
       iex> Code.ensure_loaded(DoesNotExist)
       {:error, :nofile}
 
+      iex> Code.ensure_loaded(Elixir)
+      {:error, :nofile}
+
   """
   @spec ensure_loaded(module) ::
           {:module, module} | {:error, :embedded | :badfile | :nofile | :on_load_failure}
+  def ensure_loaded(Elixir), do: {:error, :nofile}
+
   def ensure_loaded(module) when is_atom(module) do
     :code.ensure_loaded(module)
   end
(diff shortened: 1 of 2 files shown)


josevalim:
I am not sure if we should do this, honestly. While we are fixing it for Elixir, any other module can still fail in different ways. You may be trying to avoid generating the warning by ExDoc, but that feels like something that should be tackled at the doc/ExDoc front. I still don't know the best way to do it though.

lukaszsamson:
True, it won't fix it for all cases though the current code works consistently for other elixir modules - both existing and not. `Elixir` is a special module and IMO it deserves special handling. Not having it fixed in the standard library  means it needs to be fixed separately in client code, e.g. https://github.com/elixir-lsp/elixir_sense/blob/473bd3722e133e92e22fc70a5790a48df36b1ad6/lib/elixir_sense/core/introspection.ex#L49

josevalim:
> True, it won't fix it for all cases though the current code works consistently for other elixir modules

Unfortunately that's not true:

```elixir
iex(2)> Code.ensure_loaded :USER
{:error, :badfile}
iex(3)>
14:37:12.965 [error] Loading of /Users/jose/.kerl/24.jit/lib/kernel-7.1/ebin/USER.beam failed: :badfile


14:37:12.965 [error] beam/beam_load.c(162): Error loading module 'USER':
  module name in object code is user


iex(3)> Code.ensure_loaded Uri

14:37:47.741 [error] beam/beam_load.c(162): Error loading module 'Elixir.Uri':
  module name in object code is 'Elixir.URI'


{:error, :badfile}

14:37:47.741 [error] Loading of /Users/jose/OSS/elixir/bin/../lib/elixir/ebin/Elixir.Uri.beam failed: :badfile

```

:)

PR #10455, opened 2020-10-25
update function convert_time_unit document

ray-sh:
1.Remove the additional word 'can'.
2.Add the System.convert_time_unit(1, :second, :native) return value so people have clear idea about  the two units.

### DIFF
--- a/lib/elixir/lib/system.ex
+++ b/lib/elixir/lib/system.ex
@@ -940,7 +940,7 @@ defmodule System do
   unit before you display them to humans.
 
   To determine how many seconds the `:native` unit represents in your current
-  runtime, you can can call this function to convert 1 second to the `:native`
+  runtime, you can call this function to convert 1 second to the `:native`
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/system.ex
```diff
@@ -940,8 +940,8 @@ defmodule System do
   unit before you display them to humans.
 
   To determine how many seconds the `:native` unit represents in your current
-  runtime, you can can call this function to convert 1 second to the `:native`
-  time unit: `System.convert_time_unit(1, :second, :native)`.
+  runtime, you can call this function to convert 1 second to the `:native`
+  time unit: `System.convert_time_unit(1, :second, :native)` which will return 1_000_000_000.
```
josevalim, line 944: The result of the `native` conversion is dependent on the OS, so I don't think we should include this precise example, unfortunately.

ray-sh: Thanks, learned new thing. Just removed the  example.

josevalim:

PR #10458, opened 2020-10-27
mix release: anyone who can read can execute

myobie:
In the case where a mix release's results are owned by a different user than the one booting it, the binaries should be executable by anyone who can also read them.

### DIFF
--- a/lib/mix/lib/mix/tasks/release.ex
+++ b/lib/mix/lib/mix/tasks/release.ex
@@ -1423,7 +1423,7 @@ defmodule Mix.Tasks.Release do
     end
   end
 
-  defp executable!(path), do: File.chmod!(path, 0o744)
+  defp executable!(path), do: File.chmod!(path, 0o755)
 
   # Helper functions
 


myobie:
I'm not sure why the window CI is timing out, but it seems unrelated to the changes in this PR.

fertapric (approved):
I don't agree with the premise "the binaries should be executable by anyone who can also read them" but I think 755 is a better default 👍 🙂

josevalim:
To be honest, I am not sure which approach we should take. Could we accidentally allow someone who is not supposed to run the VM?

eksperimental:
I also agree that the GROUP should be able to read and execute but why would you want others to even read the file?
I think the permission should be set to 750.

josevalim:
I don’t see a problem allowing others to read it, as there are no secrets or similar.

fertapric:
I believe 755 are the default permissions when you install executables with apk or apt.

eksperimental:
> I believe 755 are the default permissions when you install executables with apk or apt.

true.

myobie:
I can rephrase the commmit to be "match apk or apt installs" if it's desired. 

The reason I originally phrased it the way I did is that a workaround for my scenario is to `cp` the release files and then `chmod` the copy. If one can read then one can ultimately execute, so I figured it doesn't change anything by making it executable by default.

josevalim:

PR #10463, opened 2020-10-29
Allow setting custom exit code on Mix error

hauleth:
It is sometimes useful and desirable to have different exit code on
different errors (for example 1 on server error, 2 on connection error,
and so on). This change adds new field to the `Mix.Error` exception -
`code` that will be then used as an exit code value.

### DIFF
--- a/lib/mix/lib/mix.ex
+++ b/lib/mix/lib/mix.ex
@@ -430,10 +430,14 @@ defmodule Mix do
 
   @doc """
   Raises a Mix error that is nicely formatted.
+
+  ## Options
+
+  - `:exit_code` - defines exit code value, defaults to `1`
   """
-  @spec raise(binary, keyword) :: no_return
+  @spec raise(binary, [exit_code: non_neg_integer()]) :: no_return
   def raise(message, opts \\ []) when is_binary(message) do
-    Kernel.raise(Mix.Error, mix: Keyword.get(opts, :code, 1), message: message)
+    Kernel.raise(Mix.Error, mix: Keyword.get(opts, :exit_code, 1), message: message)
   end
 
   @doc """
--- a/lib/mix/lib/mix.ex
+++ b/lib/mix/lib/mix.ex
@@ -431,9 +431,9 @@ defmodule Mix do
   @doc """
   Raises a Mix error that is nicely formatted.
   """
-  @spec raise(binary) :: no_return
-  def raise(message) when is_binary(message) do
-    Kernel.raise(Mix.Error, mix: true, message: message)
+  @spec raise(binary, keyword) :: no_return
+  def raise(message, opts \\ []) when is_binary(message) do
+    Kernel.raise(Mix.Error, mix: Keyword.get(opts, :code, 1), message: message)
   end
 
   @doc """
--- a/lib/mix/lib/mix.ex
+++ b/lib/mix/lib/mix.ex
@@ -433,7 +433,7 @@ defmodule Mix do
 
   ## Options
 
-  - `:exit_code` - defines exit code value, defaults to `1`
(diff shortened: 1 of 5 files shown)


lib/mix/lib/mix.ex
```diff
@@ -431,9 +431,9 @@ defmodule Mix do
   @doc """
   Raises a Mix error that is nicely formatted.
   """
-  @spec raise(binary) :: no_return
-  def raise(message) when is_binary(message) do
-    Kernel.raise(Mix.Error, mix: true, message: message)
+  @spec raise(binary, pos_integer) :: no_return
+  def raise(message, code \\ 1) when is_binary(message) do
+    Kernel.raise(Mix.Error, mix: true, message: message, code: code)
```
josevalim, line 436: I wonder if we should instead repurpose the `:mix` key. Something like this:

  def raise(message, opts \\ []) when is_binary(message) do
    Kernel.raise(Mix.Error, mix: Keyword.get(opts, :exit_code, 1), message: message)

And in the CLI, we check if mix is set. `true` is the same as 1.

lib/mix/test/mix/cli_test.exs
```diff
@@ -19,6 +19,40 @@ defmodule Mix.CLITest do
...
+        defp error(_args, opts), do: Mix.raise("oops", opts)
+      end
+      """)
+
+      assert {_, 1} = mix_code(~w[regular])
+      assert {_, 0} = mix_code(~w[no_error])
+      assert {_, 99} = mix_code(~w[custom])
+      assert {_, 1} = mix_code(~w[too_high])
+      assert {_, 1} = mix_code(~w[negative])
+    end)
```
josevalim, line 53: Those tests are very expensive so I think we are being too extensive in our coverage. I think a simple test that raises a custom code, such as `13` or whatever, should be enough.

lib/mix/lib/mix.ex
```diff
@@ -431,9 +431,9 @@ defmodule Mix do
   @doc """
   Raises a Mix error that is nicely formatted.
   """
-  @spec raise(binary) :: no_return
-  def raise(message) when is_binary(message) do
-    Kernel.raise(Mix.Error, mix: true, message: message)
+  @spec raise(binary, keyword) :: no_return
+  def raise(message, opts \\ []) when is_binary(message) do
+    Kernel.raise(Mix.Error, mix: Keyword.get(opts, :code, 1), message: message)
```
josevalim, line 436: Can we please name `exit_code` for clarity? Also please update the function docs.

eksperimental (changes requested):

lib/mix/lib/mix.ex
josevalim: Ping! :)

lib/mix/lib/mix/cli.ex
```diff
@@ -86,16 +86,25 @@ defmodule Mix.CLI do
...
+            simplified_exception(exception, code)
+
+          _ ->
+            reraise exception, __STACKTRACE__
         end
     end
   end
 
+  defp simplified_exception(%name{} = exception, code) do
+    mod = name |> Module.split() |> Enum.at(0, "Mix")
```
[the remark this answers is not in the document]
hauleth: I just copied it from the old code. Updated.

lib/mix/lib/mix.ex
```diff
@@ -430,10 +430,14 @@ defmodule Mix do
 
   @doc """
   Raises a Mix error that is nicely formatted.
+
+  ## Options
+
+  - `:exit_code` - defines exit code value, defaults to `1`
```
josevalim, line 436:
```suggestion
    * `:exit_code` - defines exit code value, defaults to `1`
```

josevalim:

PR #10479, opened 2020-11-03
Propose Pin Operator with Unused Overridden Variable

### LINKED ISSUE #10478: warning: variable "me" is unused (if the variable is not meant to be used, prefix it with an underscore)
Erlang/OTP 21 [erts-10.3] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1]
Elixir 1.11.1 (compiled with Erlang/OTP 21)

```elixir
test "returns %Game{} given valid args" do
  me = self()
  assert %Game{
               name: "Aveline",
               player1: %Player{name: "Jordan", gender: :m, pid: me}
             } = Game.new("Aveline", "Jordan", :m, me)
  end
```

See https://hex.pm/packages/islands_game

maennchen:
When rebinding a variable that is not later used, the warning
proposes to use the pin operator.

Fixes #10478

### DIFF
--- a/lib/elixir/src/elixir_env.erl
+++ b/lib/elixir/src/elixir_env.erl
@@ -92,8 +92,8 @@ reset_unused_vars(#{unused_vars := {_Unused, Version}} = E) ->
   E#{unused_vars := {#{}, Version}}.
 
 check_unused_vars(#{unused_vars := {Unused, _Version}} = E) ->
-  [elixir_errors:form_warn([{line, Line}], E, ?MODULE, {unused_var, Name}) ||
-    {{Name, _}, Line} <- maps:to_list(Unused), Line /= false, is_unused_var(Name)],
+  [elixir_errors:form_warn([{line, Line}], E, ?MODULE, {unused_var, Name, Overridden}) ||
+    {{Name, _}, {Line, Overridden}} <- maps:to_list(Unused), is_unused_var(Name)],
   E.
 
 merge_and_check_unused_vars(E, #{unused_vars := {ClauseUnused, Version}}) ->
@@ -111,10 +111,11 @@ merge_and_check_unused_vars(Current, Unused, ClauseUnused, E) ->
 
       %% The parent doesn't know it and we didn't use it
       #{} when ClauseValue /= false ->
+        {Line, Overridden} = ClauseValue,
         case is_unused_var(Name) of
           true ->
-            Warn = {unused_var, Name},
-            elixir_errors:form_warn([{line, ClauseValue}], E, ?MODULE, Warn);
+            Warn = {unused_var, Name, Overridden},
+            elixir_errors:form_warn([{line, Line}], E, ?MODULE, Warn);
 
           false ->
             ok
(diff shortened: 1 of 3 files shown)


lib/elixir/src/elixir_env.erl
```diff
@@ -138,10 +139,14 @@ is_compiler_var([$_]) -> true;
...
+  format_error({unused_var, atom_to_list(Name), VariableWithSameNameExistsInContext});
+
+format_error({unused_var, "_" ++ _ = Name, _VariableWithSameNameExistsInContext}) ->
+  io_lib:format("unknown compiler variable \"~ts\" (expected one of __MODULE__, __ENV__, __DIR__, __CALLER__, __STACKTRACE__)", [Name]);
+
+format_error({unused_var, Name, false}) ->
+  io_lib:format("variable \"~ts\" is unused (if the variable is not meant to be used, prefix it with an underscore)", [Name]);
+
+format_error({unused_var, Name, true}) ->
+  io_lib:format("variable \"~ts\" is unused (there is a variable with the same name in the context, did you mean to use the pin-operator?)", [Name]).
```
maennchen, line 152: Should the message include the `underscore` hint from the non-overridden warning as well?

lib/elixir/src/elixir_env.erl
```diff
@@ -92,8 +92,8 @@ reset_unused_vars(#{unused_vars := {_Unused, Version}} = E) ->
   E#{unused_vars := {#{}, Version}}.
 
 check_unused_vars(#{unused_vars := {Unused, _Version}} = E) ->
-  [elixir_errors:form_warn([{line, Line}], E, ?MODULE, {unused_var, Name}) ||
-    {{Name, _}, Line} <- maps:to_list(Unused), Line /= false, is_unused_var(Name)],
+  [elixir_errors:form_warn([{line, Line}], E, ?MODULE, {unused_var, Name, Overridden}) ||
+    {{Name, _}, {Line, Overridden} = Details} <- maps:to_list(Unused), Details /= false, is_unused_var(Name)],
```
josevalim, line 96: Details can never be false now :)

```suggestion
    {{Name, _}, {Line, Overridden}} <- maps:to_list(Unused), is_unused_var(Name)],
```

lib/elixir/src/elixir_env.erl
```diff
@@ -138,10 +139,14 @@ is_compiler_var([$_]) -> true;
...
 
-format_error({unused_var, Name}) ->
-  case atom_to_list(Name) of
-    "_" ++ _ ->
-      io_lib:format("unknown compiler variable \"~ts\" (expected one of __MODULE__, __ENV__, __DIR__, __CALLER__, __STACKTRACE__)", [Name]);
-    _ ->
-      io_lib:format("variable \"~ts\" is unused (if the variable is not meant to be used, prefix it with an underscore)", [Name])
-  end.
+format_error({unused_var, Name, VariableWithSameNameExistsInContext}) when is_atom(Name) ->
+  format_error({unused_var, atom_to_list(Name), VariableWithSameNameExistsInContext});
```
josevalim, line 143: Instead of calling yourself recursively, I think a `case` on the result of `atom_to_list` would be clearer. You can match on `_VariableWithSameNameExistsInContext` in a guard, like this:

    Name when VariableWithSameNameExistsInContext ->

I also think `VariableWithSameNameExistsInContext` is slightly wrong. We should probably call it `Overridden`?

lib/elixir/src/elixir_env.erl
josevalim: What do you think about: (there is a variable with the same name in the context, use the pin operator (^) to match on it or prefix this variable with underscore if it is not meant to be used) ?

maennchen: I like it :+1:

lib/elixir/src/elixir_env.erl
maennchen: Done

lib/elixir/src/elixir_env.erl
maennchen: Done

PR #10488, opened 2020-11-10
Explain REF format in `mix.deps` output

richardkmichael:


lib/mix/lib/mix/tasks/deps.ex
```diff
@@ -144,6 +144,10 @@ defmodule Mix.Tasks.Deps do
       [locked at REF]
       STATUS
 
+  For dependencies satisfied by Hex, `REF` is the package checksum.
+
+  For dependencies satisfied by git, `REF` is a commit SHA1, and may include branch or tag information.
```
chulkilee, line 149: It would be better to use the official git term ["object name"](https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-aiddefobjectnameaobjectname), since [git is going to support SHA-256 hash as well](https://git-scm.com/docs/hash-function-transition/).

Not sure `mix` assumes anything (e.g. SHA-256 hash will be 64 chars compared to 40 chars of SHA-1)

josevalim, line 149:
```suggestion
  For dependencies satisfied by git, `REF` is the commit object name,
  and may include branch or tag information.
```

?

v0idpwn: lgtm

josevalim:

lib/mix/lib/mix/tasks/deps.ex
richardkmichael: Yes, agreed, thanks for commenting.  Jose fixed it :)

PR #10490, opened 2020-11-11
Add :compressed to stream_modes spec

dominicletz:
`:compressed` is already passed through and supported here so that gzipped files can be streamed with e.g. `File.stream!("log.gz", [:compressed])`
Adding it to the spec also makes dialyzer happy when actively using it.


v0idpwn (approved):

josevalim:

PR #10507, opened 2020-11-22
Add limit to comprehensions

AlexJuca:
This PR implements :limit option for comprehensions to restrict the amount of results returned.

### DIFF
--- a/lib/elixir/src/elixir_expand.erl
+++ b/lib/elixir/src/elixir_expand.erl
@@ -258,7 +258,7 @@ expand({for, Meta, [_ | _] = Args}, E) ->
         {Args, []}
     end,
 
-  validate_opts(Meta, for, [do, into, uniq, reduce], Block, E),
+  validate_opts(Meta, for, [do, into, uniq, reduce, limit], Block, E),
 
   {Expr, Opts} =
     case lists:keytake(do, 1, Block) of
@@ -273,7 +273,7 @@ expand({for, Meta, [_ | _] = Args}, E) ->
   assert_generator_start(Meta, ECases, E),
 
   {EExpr, EE} =
-    case validate_for_options(EOpts, false, false, false) of
+    case validate_for_options(EOpts, false, false, false, false) of
       {ok, MaybeReduce} -> expand_for_do_block(Meta, Expr, EC, MaybeReduce);
       {error, Error} -> form_error(Meta, E, ?MODULE, Error)
     end,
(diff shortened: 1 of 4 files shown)


lib/elixir/src/elixir_expand.erl
```diff
@@ -702,17 +702,21 @@ generated_case_clauses([{do, Clauses}]) ->
...
 
-validate_for_options([{into, _} = Pair | Opts], _Into, Uniq, Reduce) ->
-  validate_for_options(Opts, Pair, Uniq, Reduce);
-validate_for_options([{uniq, Boolean} = Pair | Opts], Into, _Uniq, Reduce) when is_boolean(Boolean) ->
-  validate_for_options(Opts, Into, Pair, Reduce);
-validate_for_options([{uniq, Value} | _], _, _, _) ->
+validate_for_options([{into, _} = Pair | Opts], _Into, Uniq, Reduce, Limit) ->
+  validate_for_options(Opts, Pair, Uniq, Reduce, Limit);
+validate_for_options([{limit, Integer} = Pair | Opts], Into, _Uniq, Reduce, Limit) when is_integer(Integer) ->
+validate_for_options(Opts, Into, Pair, Reduce, Limit);
```
josevalim, line 708:
```suggestion
  validate_for_options(Opts, Into, Pair, Reduce, Limit);
```

AlexJuca: Looks good.

benwilson512:
At the risk of bike shedding, would `take` as the option name be better since it is serving the same function as the already existing `Stream.take/2` and `Enum.take/2`? If there was a mailing list thread that went through this I apologize for missing it!=

josevalim:
The reason why I avoided `take` is because you have an impression that the take happens after the computation (although that's not generally true for Stream.take). `limit` is clearer that no further computations or items will be taken once you reach the limit. But maybe we should indeed stick with take.

benwilson512:
> that the take happens after the computation
 
I hear that concern. The thing is, we already have to educate folks with `Stream.take` and `Enum.take` that this does not happen. If we keep the `for` option as `take` we get to go "take halts computation when it has enough" as a generic rule, vs "for `:limit` is the same as `Stream/Enum.take`, they stop computation when it has enough".

josevalim:
Ack, I will update the original issue, thank you.

Also closing this as it has gone stale with this change.

PR #10513, opened 2020-11-25
added guard clauses to functions on Integer module.

Papillon6814:
Hello.
I added guard clauses in Integer module.


josevalim:
Hi @Papillon6814, since these functions are inlined by the compiler, the guard doesn't actually make any difference, so we will skip this one for now. Thank you!

v0idpwn:
@josevalim where can I find more info about which functions are inlined by the compiler?

josevalim:
It always has a line in their documentation.

PR #10515, opened 2020-11-27
added colons.

Papillon6814:
Hello!
I added colons on comments in algebra.ex and utils.ex!


lib/elixir/lib/kernel/utils.ex
```diff
@@ -189,7 +189,7 @@ defmodule Kernel.Utils do
   that checks for its presence in a guard, then unquotes the variable references as
   appropriate.
 
-  The following code
+  The following code:
```
josevalim, line 192: No colon here because the expression continues after the example. :)

```suggestion
  The following code
```

PR #10516, opened 2020-11-29
Raise if enumerating erl_eval functions

### LINKED ISSUE #10511: Deprecate Function implementation for Enumerable

josevalim:
Closes #10511.


lib/elixir/lib/protocol.ex
```diff
@@ -60,13 +60,32 @@ defmodule Protocol do
...
     * `Any` (see below)
 
+  ## Security considerations for `Function` implementations
+
+  If you are going to implement a given protocol for `Function`,
+  you must do so with care. For example, take this code:
+
+      Enum.map(enumerable, fn x -> x * 2 end)
+
+  `enumerable` uses the `Enumerable` protocl, which can be a list,
```
fertapric, line 77:
```suggestion
  `enumerable` uses the `Enumerable` protocol, which can be a list,
```

ericmj (approved):
This is great! 💜

lib/elixir/lib/protocol.ex
```diff
@@ -60,13 +60,32 @@ defmodule Protocol do
...
+
+  `enumerable` uses the `Enumerable` protocol, which can be a list,
+  a map... or a function. If for some reason the `enumerable` comes
+  from an external source, for instance, as the result of
+  `:erlang.binary_to_term/2`, one can carefully craft a function
+  using `:erl_eval` that executes custom code. Therefore, it is
+  discouraged to use implement protocols for functions, unless those
+  protocols are not directly invoking the function (such as the
+  `Inspect` protocol, which prints function metadata, but does not
+  invoke it). If you must invoke a function, use
```
voltone, line 85:
```suggestion
  using `:erl_eval` that executes custom code. Therefore,
  protocol implementation for functions is discouraged unless the
  protocol ensures the function is never invoked (such as the
  `Inspect` protocol, which prints function metadata, but does not
  invoke it). If you must invoke a function, use
```

lib/elixir/lib/protocol.ex
```diff
@@ -60,13 +60,32 @@ defmodule Protocol do
     * `BitString`
     * `Integer`
     * `Float`
-    * `Function`
+    * `Function` - see security considerations section
```
eksperimental, line 63:
```suggestion
    * `Function` - see "Security considerations for Function implementations" section
```

lib/elixir/lib/protocol.ex
```diff
@@ -60,13 +60,32 @@ defmodule Protocol do
...
 
+  ## Security considerations for `Function` implementations
+
+  If you are going to implement a given protocol for `Function`,
+  you must do so with care. For example, take this code:
+
+      Enum.map(enumerable, fn x -> x * 2 end)
+
+  `enumerable` uses the `Enumerable` protocol, which can be a list,
+  a map... or a function. If for some reason the `enumerable` comes
```
eksperimental, line 78:
```suggestion
  a map, or a function. If for some reason the `enumerable` comes
```

voltone (commented):
Definitely a big step in the right direction. A skilful attacker may still be able to trigger function calls that have dangerous side effects, but they're going to have to 'live off the land'.

This could be hardened a bit more by also disallowing any funs that have a non-empty value for `env`: someone could manually create an ETF binary that referred to a function with an arity >2 and set `env` to a list of implicit arguments. With this capability, an attacker would have a larger range of function they could target (arity 2+ rather than 2) and they would have full control over the additional arguments. I do not suppose this capability to set implicit arguments is used in practice anywhere other than by `erl_eval`, so I think it should be safe to restrict it.

Small suggestion for avoiding a double negative in the docs.

lib/elixir/test/elixir/enum_test.exs
```diff
@@ -2089,4 +2089,14 @@ defmodule EnumTest.Function do
...
     end
   end
+
+  test "raises if given a function from erl_eval" do
+    {:ok, tokens, _} = :erl_scan.string('fun(A, B) -> A + B end.')
+    {:ok, [expr]} = :erl_parse.parse_exprs(tokens)
+    {:value, fun, _} = :erl_eval.expr(expr, [])
+
+    assert_raise ArgumentError,
+                 "cannot enumerate evaluated functions (from erl_eval) for security reasons",
```
eksperimental, line 2099:
```suggestion
                 "cannot enumerate evaluated functions (from :erl_eval) for security reasons",
```

lib/elixir/lib/stream.ex
```diff
@@ -1682,12 +1682,21 @@ defimpl Enumerable, for: Stream do
...
       {:cont, [acc | _]} -> {reason, acc}
       {:halt, [acc | _]} -> {:halted, acc}
       {:suspend, [acc | _]} -> {:suspended, acc, &{:done, elem(&1, 1)}}
     end
   end
+
+  defp no_eval!(fun) do
+    if :erlang.fun_info(fun, :module) == {:module, :erl_eval} do
+      raise ArgumentError,
+            "cannot enumerate evaluated functions (from erl_eval) for security reasons"
```
eksperimental, line 1695:
```suggestion
            "cannot enumerate evaluated functions (from :erl_eval module) for security reasons"
```

lib/elixir/lib/enum.ex
```diff
@@ -3874,7 +3874,14 @@ defimpl Enumerable, for: Function do
   def member?(_function, _value), do: {:error, __MODULE__}
   def slice(_function), do: {:error, __MODULE__}
 
-  def reduce(function, acc, fun) when is_function(function, 2), do: function.(acc, fun)
+  def reduce(function, acc, fun) when is_function(function, 2) do
```
eksperimental, line 3877:
```suggestion
  def reduce(function, acc, fun_element) when is_function(function, 2) do
```

It would be good to rename `fun` to something that can differentiate from the `function` argument

josevalim:
@voltone unfortunately :env is used any time there is a closure, so we can't forbid it.

Your comment just made me realize that this does not really fix anything. Now, all someone needs to do is to get the result of `Stream.each(..., &List.flatten/1)`, convert to binary, and then replace the `List.flatten/1` function in the environment by one that uses eval and we can't really check this one (I mean, we can, but it means checking all anonymous function calls in the Stream module and I am not sure we want to go this route).

I think the best course of action is indeed to not serialize/deserialize functions in the first place. If you allow functions, then the fact streams exist mean they may be eventually enumerated on.

voltone:
> @voltone unfortunately :env is used any time there is a closure, so we can't forbid it.

Hmm, correct me if I'm wrong, but aren't closures always handled by `erl_eval`?

> I think the best course of action is indeed to not serialize/deserialize functions in the first place. If you allow functions, then the fact streams exist mean they may be eventually enumerated on.

I would [continue to recommend that](https://erlef.github.io/security-wg/secure_coding_and_deployment_hardening/serialisation), even if the mitigation in this PR were implemented. But still, it would be nice to have a safety net that would make it a little harder to exploit.

josevalim:
> Hmm, correct me if I'm wrong, but aren't closures always handled by erl_eval?

No, only the ones created in the shell/IEx. Other closure are never evaled but rather compiled.

voltone:
> > Hmm, correct me if I'm wrong, but aren't closures always handled by erl_eval?
> 
> No, only the ones created in the shell/IEx. Other closure are never evaled but rather compiled.

If it gets compiled, doesn't the closed value get compiled into the function? For example:

```elixir
defmodule Test do
  def test do
    x = "Hello"  
    f = fn -> IO.inspect(x) end
    IO.inspect(:erlang.fun_info(f))
    f
  end
end
```

This produces:

```elixir
iex(1)> Test.test
[
  pid: #PID<0.0.0>,
  module: Test,
  new_index: 0,
  new_uniq: <<67, 112, 219, 249, 192, 124, 98, 76, 153, 183, 148, 48, 247, 222,
    224, 65>>,
  index: 0,
  uniq: 35358431,
  name: :"-test/0-fun-0-",
  arity: 0,
  env: [],
  type: :local
]
#Function<0.35358431/0 in Test.test/0>
iex(2)> :erlang.fun_info(v())
[
  pid: #PID<0.0.0>,
  module: Test,
  new_index: 0,
  new_uniq: <<67, 112, 219, 249, 192, 124, 98, 76, 153, 183, 148, 48, 247, 222,
    224, 65>>,
  index: 0,
  uniq: 35358431,
  name: :"-test/0-fun-0-",
  arity: 0,
  env: [],
  type: :local
]
```

Here `env` is not used. In the BEAM bytecode, the `apply_fun` instruction appends the contents of `env` as implicit arguments, so if the caller specified 2 arguments and `env` is a list with two values, the function being called is expected to take 4 arguments. In `erl_eval` this is used to pass the bindings as an implicit argument, which is how closures are handled. I haven't seen `env` used in any other context, but if one were to set it somehow, it would just result in additional arguments being passed (like some sort of reverse partial application).

josevalim:
@voltone that's because in your example the variable is a literal, so the compiler is smart and pass that across. But if you make it an expression, you should see the value in `env`.

voltone:
Ah, ok, I see. Thanks. TIL. So this...

```elixir
defmodule Test do        
  def test(x) do
    fn y -> x + y end
  end
end
```

...actually creates a 2-arity anonymous function in the module, and when calling `Test.test(1)` it puts the 1 in `env` so it gets passed implicitly when the returned fun is called. But that does not allow the kind of environment manipulation you were referring to earlier, do it?

josevalim:
It does. The env is a list of values. The values can be anything. :(

josevalim:
I will go ahead and close this, unfortunately nothing we can do here. Thanks.

PR #10521, opened 2020-12-01
add note about optional mix dependencies not being started

foggy1:
This has been a repeated source of confusion and while the feature makes sense there's no note of the fact that this behavior occurs anywhere. Let me know if I should flesh this out a little more.


josevalim:

PR #10523, opened 2020-12-02
Use one-based generated arguments everywhere

eksperimental:
It complemments changes introduced in 30373382ec2183d316f14b6ec6a99225855c6d8d

### DIFF
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -3883,7 +3883,7 @@ defmodule Kernel do
   end
 
   defp ensure_evaled_var(elem, {index, ast}) do
-    var = {String.to_atom("arg" <> Integer.to_string(index)), [], __MODULE__}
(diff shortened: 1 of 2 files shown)


lib/iex/lib/iex/pry.ex
```diff
@@ -482,7 +482,7 @@ defmodule IEx.Pry do
 
     # Have an extra binding per argument for case matching.
     case_vars =
-      for id <- tl(Enum.to_list(0..length(args))) do
+      for id <- tl(Enum.to_list(1..(length(args) + 1))) do
```
josevalim, line 485: This change is incorrect. We were skipping the zero already due to the `tl` call. With this change, it will start counting from 2.

eksperimental: Yes, there was no test. I was waiting for your input. I will revert it.

josevalim:

PR #10528, opened 2020-12-03
Add Integer.pow/2 and Float.pow/2

josevalim:


lib/elixir/lib/float.ex
```diff
@@ -46,6 +46,41 @@ defmodule Float do
   @precision_range 0..15
   @type precision_range :: 0..15
 
+  @doc """
+  Computes `base` raised to power of `exponent`.
+
+  `base` must be a float and `exponent` can be any number.
+  However, if a negative base and a fractional exponent
+  are given, it raises argument error.
```
chulkilee, line 54: Do we use "argument error" to refer all argument errors? Or only `ArgumentError`?  I found it is captured as `ArithmeticError` in Elixir 1.1.2 / OTP 23.1.2

```elixir
defmodule Foo do
  def pow(base, exponent) when is_float(base) and is_number(exponent),
    do: :math.pow(base, exponent)
end

Foo.pow(-1.0, 1.1)
    (stdlib 3.13.2) :math.pow(-1.0, 1.1)
    iex:3: Foo.pow/2
```

josevalim:
```suggestion
  are given, it raises `ArithmeticError`.
```

lib/elixir/lib/float.ex
```diff
@@ -46,6 +46,41 @@ defmodule Float do
...
+      iex> Float.pow(3.0, 1.5)
+      5.196152422706632
+
+      iex> Float.pow(-2.0, 3)
+      -8.0
+      iex> Float.pow(-2.0, 4)
+      16.0
+
+  """
+  def pow(base, exponent) when is_float(base) and is_number(exponent),
```
fertapric, line 81:
```suggestion
  """
  @doc since: "1.12.0"
  @spec pow(float, number) :: float
  def pow(base, exponent) when is_float(base) and is_number(exponent),
```

lib/elixir/lib/integer.ex
```diff
@@ -64,6 +64,49 @@ defmodule Integer do
...
+      iex> Integer.pow(4, 3)
+      64
+
+      iex> Integer.pow(-2, 3)
+      -8
+      iex> Integer.pow(-2, 4)
+      16
+
+  """
+  def pow(base, exponent) when is_integer(base) and is_integer(exponent) and exponent >= 0 do
```
fertapric, line 100:
```suggestion
  """
  @doc since: "1.12.0"
  @spec pow(integer, pos_integer) :: integer
  def pow(base, exponent) when is_integer(base) and is_integer(exponent) and exponent >= 0 do
```

fertapric (approved):
🎉

josevalim:

PR #10531, opened 2020-12-04
Be specific about lists when they are proper lists

eksperimental:
It was not clear to me when we use the notation `[...]` whether those were proper lists or may be improper lists as well.

### DIFF
--- a/lib/elixir/lib/keyword.ex
+++ b/lib/elixir/lib/keyword.ex
@@ -1,6 +1,6 @@
 defmodule Keyword do
   @moduledoc """
-  A keyword list is a list that consists exclusively of two-element tuples.
+  A keyword list is a proper list that consists exclusively of two-element tuples.
 
   The first element of these tuples is known as the *key*, and it must be an atom.
   The second element, known as the *value*, can be any term.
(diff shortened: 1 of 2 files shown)


josevalim:
The convention is that all lists are proper unless it is said they are explicitly improper. I really don't want to add the "proper" prefix to everything because I think it will be verbose and potentially even confusing. Thank you!

eksperimental:
> The convention is that all lists are proper unless it is said they are explicitly improper. I really don't want to add the "proper" prefix to everything because I think it will be verbose and potentially even confusing. Thank you!

Is this convention formally written somewhere in the docs?

eksperimental:
> The convention is that all lists are proper unless it is said they are explicitly improper. I really don't want to add the "proper" prefix to everything because I think it will be verbose and potentially even confusing. Thank you!

I agree with not mentioning it everywhere, so the change in the Keyword module can be ignored, but after so all these years I was not 100% if the spec `[type]` would include improper lists. So I went to read the Typespec page and is not clear: I had to take my conclusions by making reading reference after reference: `[type]` -> `[...]` -> nonempty_list(type) -> "non-empty proper list"

PR #10533, opened 2020-12-05
Indent specs after line breaks

eksperimental:
Related to https://github.com/elixir-lang/ex_doc/pull/1296

### DIFF
--- a/lib/iex/lib/iex/introspection.ex
+++ b/lib/iex/lib/iex/introspection.ex
@@ -735,9 +735,12 @@ defmodule IEx.Introspection do
     string = "@#{kind} #{Macro.to_string(definition)}"
 
     try do
+      whitespace = String.duplicate(" ", nesting)
+
       string
       |> Code.format_string!(line_length: IEx.width() - 2 * nesting)
       |> IO.iodata_to_binary()
+      |> String.replace("\n", "\n#{whitespace}")
     rescue
       _ -> string
     end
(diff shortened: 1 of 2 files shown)


lib/iex/lib/iex/introspection.ex
```diff
@@ -735,9 +735,12 @@ defmodule IEx.Introspection do
     string = "@#{kind} #{Macro.to_string(definition)}"
 
     try do
+      whitespace = String.duplicate(" ", nesting)
+
       string
       |> Code.format_string!(line_length: IEx.width() - 2 * nesting)
       |> IO.iodata_to_binary()
+      |> String.replace("\n", "\n#{whitespace}")
```
josevalim, line 743: I am afraid this is going to be indented twice, because we already indent after the try/catch block.

lib/iex/test/iex/helpers_test.exs
```diff
@@ -382,6 +382,15 @@ defmodule IEx.HelpersTest do
       assert capture_io(fn -> h(def) end) =~ def_h
     end
 
+    test "indents after line break in long specs" do
+      spec =
+        "\n" <>
+          "  @spec ascii_printable?([...], limit) :: boolean()\n" <>
+          "          when limit: :infinity | pos_integer()\n"
```
josevalim, line 389: Also note that Elixir doesn't indent `when`. This is how Elixir formats this code:

```elixir
@spec ascii_printable?([...], limit) :: boolean()
      when limit: :infinity | pos_integer()
```

josevalim: If we consider Elixir's indention, the code is actually correct. So I will close this. Thank you!

eksperimental:
I agree. The issue is only related to ExDoc.

PR #10537, opened 2020-12-07
Default the mix target to :host when empty

fhunleth:
It seems that it's an easy mistake to set `$MIX_TARGET` to be empty
rather than unset it. The intention is to compile the project for the
host and it happens when switching between target and host. While this
seems like it should be easy to catch, it can trip up Nerves users since
Mix targets aren't well used in other contexts.

This commit also protects against `$MIX_ENV` being set to the empty
string for consistency.

Finally, tests were added for verifying that the mix targets gets set
and for the case that this code change addresses.

### DIFF
--- a/lib/mix/test/mix/cli_test.exs
+++ b/lib/mix/test/mix/cli_test.exs
@@ -198,6 +198,30 @@ defmodule Mix.CLITest do
     System.delete_env("MIX_EXS")
   end
 
+  test "env config and target use defaults when empty", %{tmp_dir: tmp_dir} do
+    File.cd!(tmp_dir, fn ->
+      File.write!("custom.exs", """
+      defmodule P do
+        use Mix.Project
+        def project, do: [app: :p, version: "0.1.0"]
+      end
+      """)
+
+      System.put_env("MIX_ENV", "")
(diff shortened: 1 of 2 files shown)


lib/mix/test/mix/cli_test.exs
```diff
@@ -198,6 +198,67 @@ defmodule Mix.CLITest do
...
+      System.put_env("MIX_EXS", "custom.exs")
+
+      output = mix(["run", "-e", "IO.inspect {Mix.target(), System.argv()}", "--", "a", "b", "c"])
+
+      assert output =~ ~s({:host, ["a", "b", "c"]})
+    end)
+  after
+    System.delete_env("MIX_TARGET")
+    System.delete_env("MIX_EXS")
+  end
```
josevalim, line 260: Those tests are expensive so we tend to have few of them. Can you please keep only this last test and remove the other two? While at it, please check for `System.put_env("MIX_ENV", "")` too. Thank you!

fhunleth: Done.

josevalim:

PR #10539, opened 2020-12-07
Add :as_tuples option to keyword inspect

BobbyMcWho:
Something someone was asking if there was the ability to do in the Elixir Slack, and I was surprised there wasn't a way to do.

### DIFF
--- a/lib/elixir/lib/inspect.ex
+++ b/lib/elixir/lib/inspect.ex
@@ -175,6 +175,7 @@ defimpl Inspect, for: List do
     %Inspect.Opts{
       charlists: lists,
       char_lists: lists_deprecated,
+      keywords: keywords,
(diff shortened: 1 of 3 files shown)


josevalim:
Hi @BobbyMcWho! I am wondering what is the underlying use case for this? We support this for other data structures because the mapping between a struct and its raw representation can be miles apart... but for keywords it is always relatively straightforward.

BobbyMcWho:
I asked the OP, but to be honest I'm not entirely sure what their use case was.
I got curious if it was something already supported by the inspect module, so it gave me a good excuse to dig into Elixir source :) 

image

josevalim:
Closing this per the above. If we have evidence of where this could be useful, please let us know and we will gladly reconsider. Thanks!

PR #10542, opened 2020-12-08
Be clearer about Enumerable.count/1 implementations

eksperimental:
It should return {:ok, count} when we can count the number of elements
without traversing the enumerable.


josevalim:

PR #10543, opened 2020-12-08
Add warning for blocks without = mark on EEx

### LINKED ISSUE #10536: EEx compiler should provide a warning if a block is used without <%=
As I described on [elixir-lang-core](https://groups.google.com/g/elixir-lang-core/c/FKA9PancUm0), I had a bug that happened because my eyes slipped over the following bug:

```eex
<% if @password do %>
  Temporary password: <%= @password %>
<% else %>
  You will need to log in with your password.
<% end %>
```

The error is easily corrected:

```eex
<%= if @password do %>
  Temporary password: <%= @password %>
<% else %>
  You will need to log in with your password.
<% end %>
```

It seems to me that the EEx compiler _should_ be able to provide a compile warning if a block is used when `<% … %>` is used instead of `<%= … %>`.

This would introduce one possible incompatibility with the current compiler:

```eex
<% somevar = for item <- list do %>
  <% if(item % 2 == 0, do: item * 2, else: item * 3) %>
<% end %>
```

This can be resolved by combining these into a single tag:

```eex
<% somevar = form item <- list do
  if(item % 2 == 0, do: item * 2, else: item * 3)
end %>
```

As requested, I’ve opened an issue.

v0idpwn:
Closes #10536. Adds a small step which checks for blocks outside
assigns on unmarked tags. Warning to be defined, I've put a placeholder.

### DIFF
--- a/lib/eex/lib/eex/engine.ex
+++ b/lib/eex/lib/eex/engine.ex
@@ -221,7 +221,9 @@ defmodule EEx.Engine do
   end
 
   defp check_for_blocks(ast) do
-    Macro.traverse(ast, %{inside_assign: false, block_count: 0},
+    Macro.traverse(
+      ast,
+      %{inside_assign: false, block_count: 0},
       fn
         {:__block__, _, _} = ast, %{inside_assign: false} = acc ->
           {ast, %{acc | block_count: acc.block_count + 1}}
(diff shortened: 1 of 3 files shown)


v0idpwn:
I'm not really experienced with metaprogramming, so I believe maybe it is not ideal. I'm still opening for discussion.

josevalim:
Hi @v0idpwn! I believe the best place to add this warning is in the tokenizer: https://github.com/elixir-lang/elixir/blob/master/lib/eex/lib/eex/tokenizer.ex

If the marker is an empty string and you have a `start_expr`, we can emit a warning. :)

v0idpwn:
I'll be closing this PR and opening another one as soon as possible

PR #10552, opened 2020-12-09
Add support for functions on Enum.with_index/2

v0idpwn:
Makes  `Enum.with_index/2` support both functions and offsets by setting
a default function `fn element, index -> {element, index} end` and
a default offset `0` and letting the user change one of them in the
second argument.

### DIFF
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -627,6 +627,95 @@ defmodule Enum do
   end
 
   @doc """
+  Counts the enumerable stopping at `limit`.
+
+
+  This is useful for checking certain properties of the count of an enumerable
+  without having to actually count the entire enumerable. For example, if you
+  wanted to check that the count was exactly, at least, or more than a value.
+
+  If the enumerable implements `c:Enumerable.count/1`, the enumerable is
+  not traversed and we return the lower of the two numbers. To force
+  enumeration, use `count_until/3` with `fn _ -> true end` as the second
+  argument.
+
+  ## Examples
+
+      iex> Enum.count_until(1..20, 5)
+      5
+      iex> Enum.count_until(1..20, 50)
+      20
+      iex> Enum.count_until(1..10, 10) == 10 # At least 10
+      true
+      iex> Enum.count_until(1..11, 10 + 1) > 10 # More than 10
+      true
+      iex> Enum.count_until(1..5, 10) < 10 # Less than 10
+      true
+      iex> Enum.count_until(1..10, 10 + 1) == 10 # Exactly ten
+      true
+
+  """
+  @doc since: "1.12.0"
+  @spec count_until(t, pos_integer) :: non_neg_integer
+  def count_until(enumerable, limit) when is_integer(limit) and limit > 0 do
+    stop_at = limit - 1
+
+    case Enumerable.count(enumerable) do
+      {:ok, value} ->
+        Kernel.min(value, limit)
+
+      {:error, module} ->
+        enumerable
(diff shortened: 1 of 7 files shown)


Eiji7 (commented):
Looks nice. I have added suggestions in case we want to be more strict.

v0idpwn:
@Eiji7 thanks for taking your time to review! 

The support for functions should allow you to use the indexes however you want. For instance, you could change the tuple order (scenario which motivated this PR, see https://groups.google.com/g/elixir-lang-core/c/lZyKHBJOVHc/), this implies that the result spec could be any list()

lib/elixir/lib/enum.ex
```diff
@@ -3226,7 +3226,13 @@ defmodule Enum do
   Returns the `enumerable` with each element wrapped in a tuple
   alongside its index.
 
-  If an `offset` is given, we will index from the given offset instead of from zero.
+  May receive a function or an offset.
```
josevalim, line 3229:
```suggestion
  May receive a function or an integer offset.
```

lib/elixir/lib/enum.ex
```diff
@@ -3236,21 +3242,31 @@ defmodule Enum do
...
+  @spec do_with_index(list, integer, function) :: list
+  defp do_with_index([], _offset, _fun) do
     []
   end
 
-  defp do_with_index([head | tail], index) do
-    [{head, index} | do_with_index(tail, index + 1)]
+  defp do_with_index([head | tail], index, fun) do
+    [fun.(head, index) | do_with_index(tail, index + 1, fun)]
   end
```
josevalim, line 3270: The implementation we have here is not very efficient unfortunately. For things that are not lists, it traverses everything twice. I think we should write it like this:

```elixir
def with_index(enumerable, offset) when is_integer(offset) do
  enumerable
  |> Enumerable.map_reduce(offset, fn x, i -> {{x, i}, i+1} end)
  |> elem(0)
end

def with_index(enumerable, fun) when is_function(fun) do
  enumerable
  |> Enumerable.map_reduce(offset, fn x, i -> {fun.(x, i), i+1} end)
  |> elem(0)
end
```

Can you please update accordingly?

Thank you.

lib/elixir/lib/enum.ex
```diff
@@ -3236,21 +3242,31 @@ defmodule Enum do
       iex> Enum.with_index([:a, :b, :c], 3)
       [a: 3, b: 4, c: 5]
 
+      iex> Enum.with_index([:a, :b, :c], fn element, index -> {index, element} end)
+      [{0, :a}, {1, :b}, {2, :c}]
   """
-  @spec with_index(t, integer) :: [{element, index}]
-  def with_index(enumerable, offset \\ 0) do
+  @spec with_index(t, integer | (element, index -> any)) :: list
```
josevalim, line 3248:
```suggestion
  @spec with_index(t, integer) :: [{term, integer}]
  @spec with_index(t, (element, index -> value)) :: [value] when value: any
```

lib/elixir/lib/enum.ex
v0idpwn: Sure!

lib/elixir/lib/enum.ex
v0idpwn: Didn't even knew you could write specs this way T_T

lib/elixir/lib/enum.ex
v0idpwn: Just pushed the updated version, @josevalim :)

josevalim:

PR #10554, opened 2020-12-09
Update Tuple.sum and .product docs

wojtekmach:


lib/elixir/lib/tuple.ex
```diff
@@ -130,15 +130,16 @@ defmodule Tuple do
...
-      iex> Tuple.sum(tuple)
+
+      iex> Tuple.sum({255, 255})
       510
       iex> Tuple.sum({})
       0
   """
   @doc since: "1.12.0"
-  @spec sum(tuple) :: integer()
+  @spec sum(tuple) :: number()
```
josevalim, line 142: Good call. Can you please add a test for floats on both? Thanks!

josevalim:

PR #10555, opened 2020-12-09
Add offset option to Regex.scan/3 and Regex.run/3

antedeguemon:
This MR adds the [offset option](https://erlang.org/doc/man/re.html#run-3) to be internally used by `:re.run/3` in `Regex.scan/3` and `Regex.run/3` functions. 

The offset option specifies the starting position of the string in which the regex should start matching. It is different from simply slicing the string because of the `^` matcher:

```elixir
str = "foobar"
sliced_str = String.slice(str, 3, 3)

Regex.run(~r{^bar}, sliced_str) == ["bar"]
Regex.run(~r{^bar}, str, offset: 3) == nil
``` 

[(A more detailed example from python re docs, the offset is equivalent to the `pos` parameter there.)](https://docs.python.org/3/library/re.html#re.Pattern.search)

The offset option is somewhat already present in `Regex.scan` and `Regex.run` - it is just always implicitly defaulted to zero.

### DIFF
--- a/lib/elixir/lib/regex.ex
+++ b/lib/elixir/lib/regex.ex
@@ -296,7 +296,7 @@ defmodule Regex do
       Defaults to `:binary`.
     * `:capture` - what to capture in the result. Check the moduledoc for `Regex`
       to see the possible capture values.
-    * `:offset` - specifies the starting offset to match in the given string.
+    * `:offset` - (since v1.12.0) specifies the starting offset to match in the given string.
       Defaults to zero.
 
   ## Examples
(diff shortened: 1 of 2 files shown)


v0idpwn (approved):

lib/elixir/lib/regex.ex
```diff
@@ -296,6 +296,8 @@ defmodule Regex do
       Defaults to `:binary`.
     * `:capture` - what to capture in the result. Check the moduledoc for `Regex`
       to see the possible capture values.
+    * `:offset` - specifies the starting offset to match in the given string.
```
thiamsantos, line 299:
```suggestion
    * `:offset` - (since v1.12.0) specifies the starting offset to match in the given string.
```

lib/elixir/lib/regex.ex
```diff
@@ -425,6 +428,8 @@ defmodule Regex do
...
```
josevalim, line 431:
```suggestion
    * `:offset` - (since v1.12.0) specifies the starting offset to match in the given string.
```

josevalim:

PR #10557, opened 2020-12-09
bin/elixir: Rename erl() function to erl_set()

lanodan:
It conflicted with the erl executable for Korn-derived shells where a function
can be used in `exec command`.

It isn't clear which behavior is expected from reading the POSIX specification
but "Shell Command Language § 2.9.1 Simple Commands" clears the usage of
the term `command` quite well.
And even with excluding functions from the `command` operand of `exec`,
why are aliases accepted? (in most if not all shells)

See: https://bugs.gentoo.org/729964


josevalim:

PR #10560, opened 2020-12-10
Add left out clarification in Enumerable count

eksperimental:
Mention that when we say that "you can count" is
without traversing the whole collection.
This was left out in 4a1e4481a3714cc48597b6f55242ca6413cade4d


josevalim:

PR #10561, opened 2020-12-10
Add String.chunk_by/2

J3RN:
### Motivation

I posted this concept to the [Elixir mailing list](https://groups.google.com/g/elixir-lang-core/c/4pUwugz9Oa0), but unfortunately did not elicit much feedback.

The goal is to add a generalized version of `String.chunk/2` that can be used to succinctly describe breaking a string into groups of codepoints that share some characteristic. This makes it inherently very similar to [`Enum.chunk_by/1`](https://hexdocs.pm/elixir/Enum.html#chunk_by/2). `String.chunk/2` presently only supports chunking based on the `String.valid?/1` and `String.printable?/2` functions.

Some examples of problems that benefit from this function include:
- Wrapping text on word boundaries, while preserving the whitespace between words.
- Finding the longest sequence of repeated characters.
- Breaking a string into it's ASCII and non-ASCII parts.

### Approach

- There is very little new code here, as most of the code that forms `String.chunk_by/2` was previously in `String.chunk/2`.
- `String.chunk/2` now simply finds it's chunking function based on the second parameter and calls `String.chunk_by/2`.
- The logic inside of `do_chunk` was lightly modified such that the `flag` does not need to be a boolean, but can be anything comparable.

### DIFF
--- a/lib/elixir/lib/string.ex
+++ b/lib/elixir/lib/string.ex
@@ -1756,30 +1756,48 @@ defmodule String do
 
   """
   @spec chunk(t, :valid | :printable) :: [t]
-
-  def chunk(string, trait)
-
-  def chunk("", _), do: []
-
   def chunk(string, trait) when is_binary(string) and trait in [:valid, :printable] do
-    {cp, _} = next_codepoint(string)
-    pred_fn = make_chunk_pred(trait)
-    do_chunk(string, pred_fn.(cp), pred_fn)
+    chunk_by(string, make_chunk_pred(trait))
   end
 
-  defp do_chunk(string, flag, pred_fn), do: do_chunk(string, [], <<>>, flag, pred_fn)
+  @doc ~S"""
+  Splits the string on every character for which `fun` returns a new value.
+
+  Returns a list of binaries in which each character produces the same output
+  when passed to `fun`.
+
+  ## Examples
+
+      iex> String.chunk_by(" foo   bar ", fn cp -> cp =~ "\s" end)
+      [" ", "foo", "   ", "bar", " "]
+
+      iex> String.chunk_by("Mississippi", &Function.identity/1)
+      ["M", "i", "ss", "i", "ss", "i", "pp", "i"]
+
+      iex> String.chunk_by("Ńaïve", fn cp -> cp < "~" end)
+      ["Ń", "a", "ï", "ve"]
+
+  """
+  @spec chunk_by(t, (codepoint -> any)) :: [t]
+  def chunk_by("", _), do: []
+
+  def chunk_by(string, fun) do
+    {cp, _} = next_codepoint(string)
+    do_chunk(string, fun.(cp), fun)
+  end
+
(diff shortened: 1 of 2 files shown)


Eiji7:
@J3RN Please correct me if I'm wrong …

In your implementation we call **2 times** `fun.(cp)` where `cp` is **first** `codepoint` of `string`. Instead we could simply use **first** `cp` as `acc` and use `rest` (currently unbound variable `_`) instead of `string`.

lib/elixir/lib/string.ex
```diff
@@ -1756,30 +1756,48 @@ defmodule String do
...
 
-  def chunk(string, trait)
+  Returns a list of binaries in which each character produces the same output
+  when passed to `fun`.
 
-  def chunk("", _), do: []
+  ## Examples
 
-  def chunk(string, trait) when is_binary(string) and trait in [:valid, :printable] do
+      iex> String.chunk_by(" foo   bar ", fn cp -> cp =~ "\s" end)
```
josevalim, line 1771:
```suggestion
      iex> String.chunk_by(" foo   bar ", fn cp -> cp == "\s" end)
```

lib/elixir/lib/string.ex
```diff
@@ -1756,30 +1756,48 @@ defmodule String do
...
+      [" ", "foo", "   ", "bar", " "]
+
+      iex> String.chunk_by("Mississippi", &Function.identity/1)
+      ["M", "i", "ss", "i", "ss", "i", "pp", "i"]
+
+      iex> String.chunk_by("Ńaïve", fn cp -> cp < "~" end)
```
josevalim, line 1777: I think we should remove this example. Comparison < is most likely not going to behave how people expect, especially for Unicode.

josevalim:
Hi @J3RN,

Historically speaking, the reason why we have avoided adding enumeration functions to the String module is because there are two ways to traverse strings: by codepoint and by graphemes.

For example, for the two first examples that you gave (word boundaries and repeated chars), code points is not the correct choice, so this function should not actually be used to perform it.

For this reason, I don’t think we should pursue this functionality, at least not as proposed.

The fact not much discussion happened in the mailing list makes me believe it is not generally needed either, but lack of data is always inconclusive. :)

J3RN:
Hi @josevalim! Thanks for taking the time to review my PR!

> Historically speaking, the reason why we have avoided adding enumeration functions to the String module is because there are two ways to traverse strings: by codepoint and by graphemes.

This is an excellent point. It occurred to me while writing the proposal that one could use `String.codepoints(str) |> Enum.chunk_by(fn x -> ... end)` to achieve a similar outcome, but I still felt that this contribution was valuable. That said, the fact that there are multiple ways to enumerate a string mean that developers would most likely be better off calling `String.codepoints(str) |> Enum.chunk_by(...` or `String.graphemes(str) |> Enum.chunk_by(...` depending on their needs.

> The fact not much discussion happened in the mailing list makes me believe it is not generally needed either

Another good point, and I'll keep this in mind :grin:

PR #10568, opened 2020-12-13
Make comment unambiguous

eksperimental:
The v2.0 could be interpreted as if it's referring to the defstruct version, which by the way it is already number 2.


josevalim:

PR #10570, opened 2020-12-13
chore(Enum): :lists.foreach already returns an :ok atom, so we don't …

tomciopp:
…need to add another one

As per the erlang docs, this function already returns an `:ok` atom, so we don't need to do it twice.

https://erlang.org/doc/man/lists.html#foreach-2


josevalim:

PR #10571, opened 2020-12-13
perf(Integer.gcd): If the two integers are equal to one another, we c…

tomciopp:
…an immediately exit


josevalim:

michalmuskala:
I'm not sure this is an actual optimisation. This will only remove at most once iteration (since `rem(x, x)` will return 0 for the next iteration that will terminate the loop), but adds constant overhead to every iteration to perform the comparison.

It would be good to see numbers that this actually improves things.

josevalim:
This is going to be an improvement as long as rem(x, x) is also fast. But then, if rem(x, x) is not fast, I would rather make it fast, so I agree with @michalmuskala here. Reverting. :)

PR #10574, opened 2020-12-13
Update Task.async docs

wojtekmach:
before it read:

> Another consideration to have in mind is that tasks are always linked to their callers

This isn't true for e.g. `Task.start`. However, since it's docs for `await/2`, I guess it is implied that we're talking about tasks started by `async/1`, I thought it might be worth clarifying but maybe it's unnecessary :)


josevalim:

PR #10577, opened 2020-12-14
Remove trailing spaces from Markdown

jbampton:


josevalim:
It is there intentionally for people writing the reports. Or do they end-up being more of a bother?

PR #10581, opened 2020-12-16
Add documentation for Keyword lists with default arguments

achempion:
Hi,

I found using `Keyword` lists to pass grouped parameters is the [preferred choice](https://elixirforum.com/t/passing-in-options-maps-vs-keyword-lists/1963) over `Map` because of keyword list benefits

 * keys must be atoms
 * keys can be given more than once
 * keys are ordered, as specified by the developer

I think new Elixir developers will benefit from extended documentation how to define functions using keyword lists with default attributes.

### DIFF
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -4295,6 +4295,28 @@ defmodule Kernel do
   Note that `\\` can't be used with anonymous functions because they
   can only have a sole arity.
 
+  ### Keyword lists with default arguments
+
+  Functions containing many arguments can benefit from using `Keyword`
+  lists to group and pass attributes as a single value.
+
+     defmodule MyConfiguration do
+       @default_opts [storage: "local"]
+
+       def configure(resource, opts \\ []) do
+         opts = Keyword.merge(@default_opts, opts)
+         storage = opts[:storage]
+         # ...
+       end
+     end
+
+  The difference between using `Map` and `Keyword` to store many
+  arguments is `Keyword`'s keys:
+
+    * must be atoms
+    * can be given more than once
+    * ordered, as specified by the developer
+
   ## Function and variable names
 
   Function and variable names have the following syntax:


josevalim (approved):

lib/elixir/lib/kernel.ex
```diff
@@ -4295,6 +4295,28 @@ defmodule Kernel do
...
+         opts = Keyword.merge(@default_opts, opts)
+         storage = opts[:storage]
+         # ...
+       end
+     end
+
+  The difference between using `Map` and `Keyword` to store many
+  arguments is `Keyword`'s keys:
+
+    * must me atoms
```
chulkilee, line 4316: must "be"

chulkilee:
I'm wondering whether we can keep this it under the existing "Default arguments" section - since it's just a special case of default arguments.

Also listing the difference between keyword list and map in this section seems unnecessary since the basic is covered at https://elixir-lang.org/getting-started/keywords-and-maps.html (probably better to integrate the guide and doc?)

Since there is no "one answer" - it would be better to have own section for "keyword vs map for multiple arguments" - not under "Keyword lists with default arguments" - to compare different use cases.

whatyouhide:
Thanks @achempion! 💟

lib/elixir/lib/kernel.ex
```diff
@@ -4295,6 +4295,28 @@ defmodule Kernel do
   Note that `\\` can't be used with anonymous functions because they
   can only have a sole arity.
 
+  ### Keyword lists with default arguments
+
+  Functions containing many arguments can benefit from using `Keyword`
+  lists to group and pass attributes as a single value.
```
chulkilee, line 4301: [the guide](https://elixir-lang.org/getting-started/keywords-and-maps.html) focuses on using keyword lists to pass a group of options - not just multiple attributes.

The current change sounds like "oh, you have 5 attributes? you can use one keyword list to pass them at once!" *blindly*. I think the code is okay, but this sentence should be updated to clarify the use case.

Instead of explaining using keyword list for grouping options, I think it's better to show how `Keyword.merge/2` can be used to implement the default arguments for keyword list.

.. then we may include `Map.merge/2` example for map arguments.. under the existing section (Default arguments) as more implementation examples, instead of adding too specific section here.

chulkilee (commented):
I found it's already merged.. but I left some reviews here.

I'm also wondering whether we can have a link from doc to guides, since guide works better for cross-module feature explanation (like keyword list vs map in this case)

whatyouhide:
@chulkilee thanks for the retroactive review, can you make a PR with the changes you suggested?

PR #10585, opened 2020-12-18
Helper macro for importing config if file exists

hauleth:
This can be useful in development and CI where we want to configure parts of application that cannot be easily represented via environment variables or require additional variables (for example setting `ecto.Repo.socket_dir` when using `Postgrex` when one uses socket for local development).

### DIFF
--- a/lib/elixir/lib/config.ex
+++ b/lib/elixir/lib/config.ex
@@ -235,6 +235,25 @@ defmodule Config do
     end
   end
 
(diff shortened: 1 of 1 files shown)


josevalim:
We don't want to encourage conditional compilation because it doesn't work with the new config/runtime.exs and config/releases.exs. If you really need to do it, you can do it with File.exists?+if. Thank you!

PR #10586, opened 2020-12-21
Set evaluation to empty string to fix string replace on batch

### LINKED ISSUE #10495: elixir -e "" fails on windows
### Environment

* Elixir & Erlang/OTP versions (elixir --version): 1.11.2, otp 23.1.2
* Operating system: windows 10 

### Current behavior

While working on https://github.com/elixir-lsp/vscode-elixir-ls/pull/154 I discovered a small issue in command line parameter handling

```
elixir -e ""
** (TokenMissingError) nofile:1:3 missing terminator: "
```
It works as expected on macOS and linux.

I realise that it's not a very useful feature. In vscode-elixir-ls it is used to test if elixir command is in PATH and behaves properly. In this particular case it could be easily replaced with e.g. `elixir --version` but looking into `elixir.bat` i can see other params are being handled in the same way.

### Expected behavior

Proper handling of empty strings

josevalim:
Closes #10495.

### DIFF
--- a/bin/elixir.bat
+++ b/bin/elixir.bat
@@ -99,18 +99,21 @@ if !par!=="+elixirc" (set parsElixir=!parsElixir! +elixirc && set runMode="elixi
 rem ******* EVAL PARAMETERS ************************
 if ""==!par:-e=! (
   set "VAR=%~1"
+  if not defined VAR (set VAR= )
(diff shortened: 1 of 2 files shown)


josevalim:

PR #10591, opened 2020-12-23
ensure that a manifest is generated, even with no source code

### LINKED ISSUE #10590: Mix.Utils.extract_stale/2 marks all source files as stale if the manifest doesn't exist
### Environment

```
Erlang/OTP 23 [erts-11.0.2] [source] [64-bit] [smp:16:16] [ds:16:16:10] [async-threads:1] [hipe]
Elixir 1.12.0-dev (86e2384) (compiled with Erlang/OTP 23)
```
(Though it also happens on elixir 1.11.x)

* Operating system: MacOs Catalina (10.15.7)

### Current behavior

In an umbrella project, if there is an app with no source code, then 
```
    manifests = Mix.Tasks.Compile.Elixir.manifests()
    configs = Mix.Project.config_files()
    stale = Mix.Utils.extract_stale(configs, manifests)
```
marks all of config files as stale. (this snippet is taken from phoenix: https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/code_reloader/server.ex#L210-L230)

This prevents phoenix's code reloading from working, since the manifest file (`test_empty_app_umbrella/_build/dev/lib/common_config/.mix/compile.elixir`) doesn't exist, which causes its last updated to be the unix epoch, and so thus any source files are marked as being stale. See https://github.com/baseballlover723/test_empty_app for a reproducible test project.

### Expected behavior

Source files (from the umbrella) are not marked as stale if a child app has no source files.

### Notes
I would consider this to be a bug in mix and not in phoenix, since `Mix.Tasks.Compile.Elixir.manifests()` returns a manifest file that may or may not exist. In my opinion, either `Mix.Tasks.Compile.Elixir.manifests()` should check and make sure that the manifest file exists, or manifest files should just always be generated, even if there isn't any source code.

baseballlover723:
fixes #10590

I'm not too sure what the wider implications are of always generating manifest files, but I would presume that an empty manifest file would behave similarly or better then a non existent manifest file.


lib/mix/lib/mix/compilers/elixir.ex
```diff
@@ -119,6 +119,9 @@ defmodule Mix.Compilers.Elixir do
         write_manifest(manifest, modules, sources, all_local_exports, timestamp)
         {:ok, warning_diagnostics(sources)}
 
+      !File.exists?(manifest) ->
+        write_manifest(manifest, [], [], %{}, timestamp)
+
       true ->
```
josevalim, line 125:
```suggestion
      true ->
        write_manifest(manifest, [], [], %{}, timestamp)
```

lib/mix/test/fixtures/no_source_code/mix.exs
```diff
@@ -0,0 +1,10 @@
+defmodule TestStale.MixProject do
```
josevalim, line 1: We don't need a new fixture, you can pick one of the existing ones, such as `compile_erlang`. :)

lib/mix/lib/mix/compilers/elixir.ex
```diff
@@ -120,6 +120,7 @@ defmodule Mix.Compilers.Elixir do
         {:ok, warning_diagnostics(sources)}
 
       true ->
+        write_manifest(manifest, [], [], %{}, timestamp)
```
josevalim, line 123: Actually, you were right. If we always do this, it means we will always trigger downstream compilers. :D Sorry!

```suggestion
        if not File.exists?(manifest) do
          write_manifest(manifest, [], [], %{}, timestamp)
        end
```

baseballlover723: do you prefer this or changing it back to being part of the cond?

josevalim: This is more appropriate because we still want to return no-op. :)

lib/mix/test/mix/tasks/compile.elixir_test.exs
```diff
@@ -109,6 +109,22 @@ defmodule Mix.Tasks.Compile.ElixirTest do
...
   end
 
+  test "generates module-application manifest" do
+    in_fixture("no_mixfile", fn ->
+      Mix.Tasks.Compile.Elixir.run([])
+
+      assert File.regular?("_build/dev/lib/sample/.mix/compile.elixir")
+    end)
+  end
+
```
josevalim, line 119:
```suggestion
```

lib/mix/test/mix/tasks/compile.elixir_test.exs
```diff
@@ -109,6 +109,22 @@ defmodule Mix.Tasks.Compile.ElixirTest do
...
+  test "generates module-application manifest even if there isn't any Elixir source code" do
```
josevalim, line 112:
```suggestion
  test "generates manifest even if there isn't any Elixir source code" do
```

josevalim:

PR #10595, opened 2020-12-26
Improve Enum.chunk_while/4 docs to clarify `after_fun` semantics

madlep:
I found trying to understand `Enum.chunk_while/4` quite confusing, and needed to resort to reading the code to understand how the `after_fun` param semantics are intended to work. Have added my understanding to the docs for it, and made a couple of improvements to code readability, and added a couple of test cases covering extra cases.

- Updated docs to clarify
  - add `{:halt, acc}` tuple return doc for `chunk_fun`
  - explain `after_fun` semantics in more detail
  - format return tuples for `chunk_fun` and `after_fun` as markdown lists
  - added note that `acc` in `after_fun` return tuple is ignored
  - state `Enum.chunk_while/4` returns a list of *chunks*, not a list of lists (emitted chunks can be anything, not necessarily just lists)
- added example showing `after_fun`'s behaviour more clearly
- renamed variable names used in `chunk_fun` / `after_fun` matches to make them consistent with each other, and with the typespec of the functions
- added extra tests for `after_fun`
  - more explicitly exercising behaviour when acc has accumulated more elements
  - handling returning types other than just lists

### DIFF
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -485,12 +485,12 @@ defmodule Enum do
   as part of an accumulator, but were not emited as a chunk by `chunk_fun`.
   It must return:
 
-  * `{:cont, chunk, acc}` to emit a chunk. The chunk will be appended to the
-     list of already emitted chunks.
-  * `{:cont, acc}` to not emit a chunk
+    * `{:cont, chunk, acc}` to emit a chunk. The chunk will be appended to the
+      list of already emitted chunks.
+    * `{:cont, acc}` to not emit a chunk
 
-  (`acc` in `after_fun` return tuple is required by `chunk_every/4` to match
-  on the result, but is ignored.)
+  The `acc` in `after_fun` is required in order to mirror the tuple format
+  from `chunk_fun` but it will be discarded since the traversal is complete.
 
   Returns a list of emitted chunks.
 
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -509,6 +509,8 @@ defmodule Enum do
       ...> end
       iex> Enum.chunk_while(1..10, [], chunk_fun, after_fun)
       [[1, 2], [3, 4], [5, 6], [7, 8], [9, 10]]
+      iex> Enum.chunk_while([1, 2, 3, 5, 7], [], chunk_fun, after_fun)
+      [[1, 2], [3, 5, 7]]
 
   """
   @doc since: "1.5.0"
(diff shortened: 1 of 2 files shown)


lib/elixir/lib/enum.ex
```diff
@@ -513,15 +525,15 @@ defmodule Enum do
...
+          {:cont, chunk, acc} -> {:cont, {[chunk | buffer], acc}}
           {:cont, acc} -> {:cont, {buffer, acc}}
           {:halt, acc} -> {:halt, {buffer, acc}}
         end
       end)
 
     case after_fun.(acc) do
       {:cont, _acc} -> :lists.reverse(res)
-      {:cont, elem, _acc} -> :lists.reverse([elem | res])
+      {:cont, chunk, _acc} -> :lists.reverse([chunk | res])
```
madlep, line 536: This was the part that initially confused me, and took a while to grok what is going on here. The confusing part is that `_acc` is required in the returned tuple from `after_fun`, but it's ignored.

I'm wondering if it makes sense to deprecate that, and change the return shape to look something more like:

```elixir
case after_fun.(acc) do
  {:chunk, chunk} ->
    :lists.reverse([chunk | res])
  :ok ->
    :lists.reverse(res)
  {:cont, chunk, _acc} ->
    IO.warn("Enum.chunk_while/4 after_fun callback function returning {:cont, chunk, acc} is deprecated, please return {:chunk, chunk} instead")
    :lists.reverse(res)
  {:cont, _acc} ->
    IO.warn("Enum.chunk_while/4 after_fun callback function returning {:cont, acc} is deprecated, please return :ok instead")
    :lists.reverse(res)
end
```

Maybe also changing the return shape from `chunk_fun` to something similar too.

lib/elixir/lib/enum.ex
```diff
@@ -474,15 +474,25 @@ defmodule Enum do
...
+  finished (or `halt`ed) to handle any trailing elements that were returned
+  as part of an accumulator, but were not emited as a chunk by `chunk_fun`.
+  It must return:
+
+  * `{:cont, chunk, acc}` to emit a chunk. The chunk will be appended to the
+     list of already emitted chunks.
+  * `{:cont, acc}` to not emit a chunk
+
+  (`acc` in `after_fun` return tuple is required by `chunk_every/4` to match
+  on the result, but is ignored.)
```
josevalim, line 493:
```suggestion
    * `{:cont, chunk, acc}` to emit a chunk. The chunk will be appended to the
      list of already emitted chunks.
    * `{:cont, acc}` to not emit a chunk

  The `acc` in `after_fun` is required in order to mirror the tuple format
  from `chunk_fun` but it will be discarded since the traversal is complete.
```

josevalim:

PR #10600, opened 2020-12-29
Format interpolations on one line

### LINKED ISSUE #10587: Do not break lines inside interpolation in the formatter
Today, if an interpolation goes over the line limit, it is broken like this:

```
  "foobar #{
    x
  } bazbat"
```

My hope would be that people would like at this and rewrite it but unfortunately it didn't happen, so I think we should keep it as a single line and let it be rewritten by hand (most likely because it goes over a line limit).

jrusso1020:
Allow interpolated strings to be on the same line if it's longer
than the max length

fixes #10587

### DIFF
--- a/lib/elixir/lib/code/formatter.ex
+++ b/lib/elixir/lib/code/formatter.ex
@@ -1380,7 +1380,7 @@ defmodule Code.Formatter do
   defp list_interpolation_to_algebra([entry | entries], escape, state, acc, last) do
     {{:., _, [Kernel, :to_string]}, meta, [quoted]} = entry
     {doc, state} = block_to_algebra(quoted, line(meta), closing_line(meta), state)
-    doc = surround("\#{", doc, "}")
+    doc = interpolate_doc(doc)
     list_interpolation_to_algebra(entries, escape, state, concat(acc, doc), last)
   end
 
@@ -1397,13 +1397,16 @@ defmodule Code.Formatter do
   defp interpolation_to_algebra([entry | entries], escape, state, acc, last) do
     {:"::", _, [{{:., _, [Kernel, :to_string]}, meta, [quoted]}, {:binary, _, _}]} = entry
     {doc, state} = block_to_algebra(quoted, line(meta), closing_line(meta), state)
-    doc = surround("\#{", doc, "}")
+    doc = interpolate_doc(doc)
     interpolation_to_algebra(entries, escape, state, concat(acc, doc), last)
   end
 
   defp interpolation_to_algebra([], _escape, state, acc, last) do
     {concat(acc, last), state}
   end
+
+  defp interpolate_doc({:doc_force, _t} = doc), do: surround("\#{", doc, "}")
+  defp interpolate_doc(doc), do: surround("\#{", doc, "}") |> format_to_string() |> string()
 
   ## Sigils
 
--- a/lib/elixir/lib/inspect/algebra.ex
+++ b/lib/elixir/lib/inspect/algebra.ex
@@ -189,6 +189,7 @@ defmodule Inspect.Algebra do
   @container_separator ","
   @tail_separator " |"
   @newline "\n"
+  @interpolation_identifier "\#{"
   @next_break_fits :enabled
 
   # Functional interface to "doc" records
@@ -966,6 +967,8 @@ defmodule Inspect.Algebra do
   defp fits?(w, k, b?, [{i, m, doc_nest(x, j, _)} | t]),
     do: fits?(w, k, b?, [{apply_nesting(i, k, j), m, x} | t])
 
+  defp fits?(_w, _k, _b?, [{_i, _m, doc_cons(@interpolation_identifier, _y)} | _t]),
+    do: true
   defp fits?(w, k, b?, [{i, m, doc_cons(x, y)} | t]),
     do: fits?(w, k, b?, [{i, m, x}, {i, m, y} | t])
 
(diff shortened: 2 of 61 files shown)


jrusso1020:
I gave this a try but it's not perfect

One test is failing which probably shouldn't be
```
1) test interpolation with comment inside before and after (Code.Formatter.CommentsTest)
     test/elixir/code_formatter/comments_test.exs:243
     Assertion with == failed
     code:  assert IO.iodata_to_binary(Code.format_string!(good, opts)) == String.trim(good)
     left:  "IO.puts(\"Hello \#{# comment\nworld}\")"
     right: "IO.puts(\n  \"Hello \#{\n    # comment\n    world\n  }\"\n)"
     stacktrace:
       test/elixir/code_formatter/comments_test.exs:244: (test)
```

Open to other ideas on how to fix this but I was trying to avoid changing the algebra returned by `Code.Formatter.to_algebra!` and instead make a change in the `Inspect.Algebra.format` function. Let me know if that's a bad idea. 

At first glance a solution was to say interpolated strings `fit` by looking at the ``{:doc_cons, "\#{", ...}`, which looks like it works for cases that don't involve comments within the interpolation

lib/elixir/lib/inspect/algebra.ex
```diff
@@ -189,6 +189,7 @@ defmodule Inspect.Algebra do
   @container_separator ","
   @tail_separator " |"
   @newline "\n"
+  @interpolation_identifier "\#{"
```
josevalim, line 192: Inspect.Algebra is general purpose and it should not know about the formatter. :) We also need to make sure that any expression inside the interpolation doesn't change either.

I think you can solve this fully inside the formatter by calling `doc |> format_to_string |> string`  in the formatter where `doc` is the interpolation contents.

jrusso1020: @josevalim yeah makes sense to keep it in the formatter, not sure why but when I was initially following the code I couldn't find `\#{` in the formatter 😆 

Gave this another shot, I initially tried doing the below but it didn't seem to solve the issue (from your explanation it seemed you recommended trying to do it on the contents of the interpolation which is why I tried this)
```
doc = doc |> format_to_string() |> string()
doc = surround("\#{", doc, "}")
``` 

Next I tried doing the below, but this had a similar issue to my first attempt where it messed up the comments in the interpolation and put everything on one line
```
surround("\#{", doc, "}") |> format_to_string() |> string()
``` 

So I took a look at what `doc` was and noticed a `{:doc_force, ...}` at the start of the algebra in the comment inside the interpolation case so I had to handle the edge case of not putting it all on one line. Let me know if you have any better ideas!

lib/elixir/lib/code/formatter.ex
```diff
@@ -1397,14 +1397,17 @@ defmodule Code.Formatter do
...
+    doc = interpolate_doc(doc)
     interpolation_to_algebra(entries, escape, state, concat(acc, doc), last)
   end
 
   defp interpolation_to_algebra([], _escape, state, acc, last) do
     {concat(acc, last), state}
   end
 
+  defp interpolate_doc({:doc_force, _t} = doc), do: surround("\#{", doc, "}")
+  defp interpolate_doc(doc), do: surround("\#{", doc, "}") |> format_to_string() |> string()
```
josevalim, line 1409: Surround is the done adding the breaks. Instead of handling force, try this:

```suggestion
  defp interpolate_doc(doc), do: concat(["\#{", doc, "}"]) |> format_to_string() |> string()
```

josevalim: Also, please add a test where you make sure something like:

```elixir
"""
foo
#{concat(foo, bar)}
bar
"""
```

doesn't break lines either.

jrusso1020: When using `concat` directly instead of `surround` and handling `:doc_force`  the following test fails
```
test "with comment inside before and after" do
      assert_same ~S"""
      IO.puts(
        "Hello #{
          # comment
          world
        }"
      )
      """

      assert_same ~S"""
      IO.puts(
        "Hello #{
          world
          # comment
        }"
      )
      """
end
```

with the following message
```
1) test interpolation with comment inside before and after (Code.Formatter.CommentsTest)
     test/elixir/code_formatter/comments_test.exs:243
     Assertion with == failed
     code:  assert IO.iodata_to_binary(Code.format_string!(good, opts)) == String.trim(good)
     left:  "IO.puts(\"Hello \#{# comment\nworld}\")"
     right: "IO.puts(\n  \"Hello \#{\n    # comment\n    world\n  }\"\n)"
     stacktrace:
       test/elixir/code_formatter/comments_test.exs:244: (test)
```

This is similar when using `surround` as well. My thought was that this test should continue to pass with these changes because we want to keep the formatting as is when comments are in the interpolation

Is my understanding incorrect? Do we want this test to be updated as well?

josevalim:
Oh, I see. Yeah, the comment makes things much more complex. Unfortunately I am not sure what is the best solution, I will have to investigate it in depth, sorry.

jrusso1020:
@josevalim makes sense! I pushed up the additional tests you asked for, tests are all passing. I'll also see if I can come up with a better solution than checking for `{:doc_force, ...}` but thanks for the help!

josevalim:
Hi @jrusso1020, I spent most of the day trying a solution and I think I was able to come up with something. I have pushed it to master. In any case, thank you very much for your PR which helped explore some of those ideas! :)

PR #10601, opened 2020-12-29
Fix `Path.relative_to/2` when referencing self

aerosol:
Hello, 

Unless my reasoning is wrong, there might be a bug
in `Path.relative_to/2` implementation.

The assumption I'm making is `relative_to` should always
return a path of type `:relative`.

Thus, if `relative_to` argument points at `path`, the
result should be relative.

Currently it's:

    iex(1)> Path.relative_to('/foo/bar', '/foo/bar') |> Path.type()
    :absolute

This patch fixes it, so it's:

    iex(1)> Path.relative_to('/foo/bar', '/foo/bar')
    "."
    iex(2)> Path.type(v(1))
    :relative

Similar code in python:

    In [1]: os.path.relpath('/foo/bar', start='/foo/bar')
    Out[1]: '.'

And in ruby:

    > Pathname.new('/foo/bar').relative_path_from(Pathname.new('/foo/bar')).to_s
    => "."

### DIFF
--- a/lib/elixir/lib/path.ex
+++ b/lib/elixir/lib/path.ex
@@ -313,11 +313,18 @@ defmodule Path do
       iex> Path.relative_to("/usr/local/foo", "/etc")
       "/usr/local/foo"
 
+      iex> Path.relative_to("/usr/local/foo", "/usr/local/foo")
+      "."
+
   """
   @spec relative_to(t, t) :: binary
   def relative_to(path, from) do
     path = IO.chardata_to_string(path)
     relative_to(split(path), split(from), path)
+  end
+
+  defp relative_to(path, path, _original) do
+    "."
   end
 
   defp relative_to([h | t1], [h | t2], original) do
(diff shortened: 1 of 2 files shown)


aerosol:
The [CI error](https://github.com/elixir-lang/elixir/pull/10601/checks?check_run_id=1621488950) I'm getting seems unrelated:

```
   1) test raises error message about correct usage (Mix.Tasks.IexTest)
     test/mix/tasks/iex_test.exs:6
     stacktrace:
       (stdlib 3.12) :ets.next(:ac_tab, {:loaded, :eex})
       (stdlib 3.12) ets.erl:780: :ets.do_filter/5
       test/test_helper.exs:57: anonymous fn/0 in MixTest.Case.__ex_unit_setup_0/1
       (ex_unit 1.12.0-dev) lib/ex_unit/on_exit_handler.ex:143: ExUnit.OnExitHandler.exec_callback/1
       (ex_unit 1.12.0-dev) lib/ex_unit/on_exit_handler.ex:129: ExUnit.OnExitHandler.on_exit_runner_loop/0
```

josevalim:

Issue #10602, opened 2020-12-29
Add Kernel.tap/2 and Kernel.then/2


josevalim:
tap/2 is for side-effects in pipelines, then/2 is for anonymous functions:

```elixir
def tap(x, fun) do
  fun.(x)
  x
end

def then(x, fun), do: fun.(x)
```

We should probably improve the error message in Macro.pipe to link to refer to `Kernel.then/2`.

v0idpwn:
Can I take this one?

Eiji7:
Sorry @v0idpwn, I started working on `Kernel.then/2` before you asked. :smiling_imp:

lukaszsamson:
I wonder if those functions should handle awaitable `%Task{}` specially.
```
Task.async(fn -> some_io_bound_op() end)
|> tap(fn res -> do_sth(res) end)
|> then(fn res -> Task.async(fn -> do_sth(res) end) end)
|> Task.await
# awaited do_sth(res)
```

josevalim:
Please see mailing list discussion @lukaszsamson. Overall, It is not the goal of these functions to be polymorphic, modules could provide variants but I don’t think there are reasonable implementations for those in task.

PR #10603, opened 2020-12-29
Adds Kernel.then/2

Eiji7:
Related issue: #10602

### DIFF
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -2473,7 +2473,7 @@ defmodule Kernel do
   """
   @doc since: "1.12.0"
   @spec then(x, (x -> result)) :: result when result: any(), x: any()
-  def then(x, fun), do: fun.(x)
+  def then(x, fun) when is_function(fun, 1), do: fun.(x)
 
   @doc """
   Gets a value from a nested structure.
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -2472,7 +2472,7 @@ defmodule Kernel do
       2
   """
   @doc since: "1.12.0"
-  @spec then(x, (x -> result)) :: result when result: any, x: var
+  @spec then(x, (x -> result)) :: result when result: var, x: var
   def then(value, fun) when is_function(fun, 1), do: fun.(value)
 
   @doc """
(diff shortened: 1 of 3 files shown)


lib/elixir/lib/kernel.ex
```diff
@@ -2455,6 +2455,26 @@ defmodule Kernel do
...
+
+  This function makes such code easy to read.
+
+  ### Examples
+
+      iex> 1 |> then(fn x -> x * 2 end)
+      2
+  """
+  @doc since: "1.12.0"
+  @spec then(x, (x -> result)) :: result when result: any, x: var
```
josevalim, line 2475:
```suggestion
  @spec then(x, (x -> result)) :: result when result: var, x: var
```

lib/elixir/lib/kernel.ex
```diff
@@ -2455,6 +2455,26 @@ defmodule Kernel do
...
   end
 
+  @doc """
+  Pipes a value into the function.
+
+  Using pipe `|>` operator you cannot use anonymous functions directly.
+  To make them work you need to wrap them into parentheses
+  and call such function like:
+
+      iex> 1 |> (fn x -> x * 2 end).()
```
wojtekmach, line 2465: this approach was often discouraged (in favour of e.g. extracting private function) and with the addition of then/1 it's unnecessary. Thus, I wouldn't even talk about it. Might be worth to talk about then/1 in the pipe docs though.

Eiji7: :ok_hand: no problem

Eiji7:
@wojtekmach Let me know if documentation looks good for you now.

lib/elixir/lib/kernel.ex
```diff
@@ -2455,6 +2455,21 @@ defmodule Kernel do
     end
   end
 
+  @doc """
+  Pipes a value into the function.
+
+  Using pipe `|>` operator you cannot use anonymous functions directly.
+  This function makes such code easy to read.
```
josevalim, line 2462:
```suggestion
  Pipes `value` into the given `fun`.

  In other words, it invokes `fun` with `value` as argument.
  This is most commonly used in pipelines, allowing you
  to pipe a value to a function outside of its first argument.
```

josevalim:

PR #10604, opened 2020-12-29
Add Kernel.tap/2

v0idpwn:
Missing tests and examples, will be adding soon. Does half of #10602

### DIFF
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -1141,10 +1141,10 @@ defmodule Kernel do
     %{a: 1}
   """
   @doc since: "1.12.0"
-  @spec tap(var, (var -> any)) :: var when var: term
-  def tap(var, fun) when is_function(fun, 1) do
-    fun.(var)
-    var
+  @spec tap(x, (x -> any)) :: x when x: var
+  def tap(value, fun) when is_function(fun, 1) do
+    fun.(value)
+    value
   end
 
   @doc """
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -1127,6 +1127,27 @@ defmodule Kernel do
   end
 
   @doc """
+  Applies a function over a value returns the initial value.
+
+  Useful for running synchronous side effects when you don't want their
+  result to affect a pipe chain.
+
+  ## Examples
+
+    iex> tap(1, fn x -> x + 1 end)
+    1
+
+    iex> tap(%{a: 1}, &IO.inspect(Map.keys(&1)))
+    %{a: 1}
+  """
+  @doc since: "1.12.0"
+  @spec tap(var, (var -> any)) :: var when var: term
+  def tap(var, fun) when is_function(fun, 1) do
+    fun.(var)
+    var
+  end
+
+  @doc """
   A non-local return from a function.
 
   Check `Kernel.SpecialForms.try/1` for more information.
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -1127,18 +1127,24 @@ defmodule Kernel do
   end
 
   @doc """
-  Applies a function over a value returns the initial value.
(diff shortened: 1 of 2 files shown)


lib/elixir/lib/kernel.ex
```diff
@@ -1126,6 +1126,17 @@ defmodule Kernel do
     :erlang.spawn_monitor(module, fun, args)
   end
 
+  @doc """
+  Applies a function over a term and returns term
+
+  Useful for running side effects
+  """
+  @spec tap(x :: any, (x -> any)) :: x
```
Eiji7, line 1134:
```suggestion
  @spec tap(x, (x -> any)) :: x when x: any
```

Eiji7 (changes requested):

Eiji7:
@v0idpwn You should add also:
```elixir
@doc since: "1.12.0"
```
and some examples in documentation and test (check my PR for a reference).

lib/elixir/lib/kernel.ex
josevalim: Use “var” instead of any. Also please add examples, @doc since, and tests.

lib/elixir/lib/kernel.ex
```diff
@@ -1126,6 +1126,17 @@ defmodule Kernel do
     :erlang.spawn_monitor(module, fun, args)
   end
 
+  @doc """
+  Applies a function over a term and returns term
+
+  Useful for running side effects
+  """
+  @spec tap(x, (x -> any)) :: x when x: term
```
josevalim, line 1134: Use “var” instead of any. Also please add examples, @doc since, and tests.

josevalim: Also please use “value” instead of “term” and add periods at the end of sentences.

v0idpwn: Sorry, @josevalim, I opened the PR before finishing just to signalize work was being done

v0idpwn:
Just pushed a first version. 

I need to use some kind of side effect on the doctests, but using IO.inspect() will print during tests. Do you have any other idea of illustrative tests which would serve for this purpose?

lib/elixir/lib/kernel.ex
```diff
@@ -1126,6 +1126,27 @@ defmodule Kernel do
...
+  Useful for running synchronous side effects when you don't want their
+  result to affect a pipe chain.
+
+  ## Examples
+
+    iex> tap(1, fn x -> x + 1 end)
+    1
+
+    iex> tap(%{a: 1}, &IO.inspect(Map.keys(&1)))
+    %{a: 1}
```
v0idpwn, line 1141: This doctest prints during tests. I want a side-effect that doesn't print but could be useful for the reader. Nothing comes in mind.

maennchen: How about a `send/2` to somewhere?

lib/elixir/lib/kernel.ex
```diff
@@ -1126,6 +1126,27 @@ defmodule Kernel do
     :erlang.spawn_monitor(module, fun, args)
   end
 
+  @doc """
+  Applies a function over a value returns the initial value.
+
+  Useful for running synchronous side effects when you don't want their
+  result to affect a pipe chain.
```
josevalim, line 1133:
```suggestion
  Pipes `value` to the given `fun` and returns the `value` itself.

  Useful for running synchronous side effects in a pipeline.
```

lib/elixir/lib/kernel.ex
josevalim, line 1141:
```suggestion
      iex> tap(1, fn x -> x + 1 end)
      1

  Most commonly, this is used in pipelines. For example,
  let's suppose you want to inspect part of a data structure.
  You could write:

      %{a: 1}
      |> Map.update!(:a, & &1 + 2)
      |> tap(&IO.inspect(&1.a))
      |> Map.update!(:a, & &1 * 2)

```

josevalim:

PR #10605, opened 2020-12-30
Extend functions/macros not allowed inside Protocols

eksperimental:
Add to the list: defexception/1, defdelegate/2, defguard/1, defguardp/1

### DIFF
--- a/lib/elixir/test/elixir/protocol_test.exs
+++ b/lib/elixir/test/elixir/protocol_test.exs
@@ -262,6 +262,33 @@ defmodule ProtocolTest do
                    end
                  end
   end
+
+  test "cannot define function definitions inside protocol" do
+    assert_raise CompileError, ~r"undefined function def/2", fn ->
+      defprotocol Foo do
+        def bar(arg), do: arg
+      end
+    end
+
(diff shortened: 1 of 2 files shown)


lib/elixir/test/elixir/protocol_test.exs
```diff
@@ -262,6 +262,33 @@ defmodule ProtocolTest do
...
+        end
+      end
+    end
+
+    assert_raise CompileError, ~r"undefined function defguard/1", fn ->
+      defprotocol Foo do
+        defguard is_something(arg) when arg == :something
+      end
+    end
+  end
```
josevalim, line 291:
```suggestion
```

josevalim: We don't need those tests as they are basically testing the declarative aspect of the code.

eksperimental:
This is probable for another PR. I was trying to defined a __using__/1 macro iside a protocol, and I was scratching my head becuase it said defmacro/2 was not defined. I had to go to the source code of protocol and figure out what was going on.
I think we should override these functions and emit a warning when called, to give the user a hint of what is going on.
On the other hand, there is no documentation that backs this behaviour.

josevalim:

PR #10608, opened 2020-12-30
Stremaline specs in Kernel.{tap, then}/2

eksperimental:
x is not used anywhere in the specs. Use name of argument instead.

/cc @Eiji7, @v0idpwn


v0idpwn (approved):

josevalim:

PR #10612, opened 2020-12-31
Return :error on decomposing regular tuples as calls

am-kantox:
Since [`Macro.decompose_call/1`](https://hexdocs.pm/elixir/Macro.html?#decompose_call/1) is exported and documented, it would make sense to explicitly handle the case of passing quoted regular tuple as an argument, returning `:error` as it does for any other garbage (neither local nor remote call.)

### DIFF
--- a/lib/elixir/lib/macro.ex
+++ b/lib/elixir/lib/macro.ex
@@ -548,7 +548,6 @@ defmodule Macro do
   @spec decompose_call(t()) :: {atom, [t()]} | {t(), atom, [t()]} | :error
   def decompose_call(ast)
 
-  # regular tuples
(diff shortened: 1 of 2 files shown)


lib/elixir/lib/macro.ex
```diff
@@ -541,10 +541,16 @@ defmodule Macro do
...
       :error
 
+      iex> Macro.decompose_call(quote(do: {:foo, [], []}))
+      :error
+
   """
   @spec decompose_call(t()) :: {atom, [t()]} | {t(), atom, [t()]} | :error
   def decompose_call(ast)
 
+  # regular tuples
```
josevalim, line 551:
```suggestion
```

josevalim:

PR #10615, opened 2020-12-31
Add Collectable.into/1 to Deprecation list & update Enum.into/2 doc

eksperimental:


lib/elixir/lib/enum.ex
```diff
@@ -1328,7 +1328,7 @@ defmodule Enum do
 
   Note that passing a non-empty list as the `collectable` is deprecated. If you're collecting
   into a non-empty keyword list, consider using `Keyword.merge/2`. If you're collecting into a
-  non-empty list, consider something like `to_list(enumerable) ++ collectable`.
+  non-empty list, consider something like `collectable ++ to_list(enumerable)`.
```
eksperimental, line 1331: This behaves the same way Enum.into/2 does.

josevalim: The reason why we are doing something different is precisely because the current deprecated behavior is wrong, so I would point to the correct behavior.

eksperimental: I cannot see why the current behavior is wrong.
Maybe if you can link to discussion of the decision.

```elixir
iex(11)> collectable = [1, 2, 3]                
[1, 2, 3]
iex(12)> enumerable = 4..6                      
4..6
iex(13)> Enum.into(enumerable, collectable)     
warning: the Collectable protocol is deprecated for non-empty lists. The behaviour of things like Enum.into/2 or "for" comprehensions with an :into option is incorrect when collecting into non-empty lists. If you're collecting into a non-empty keyword list, consider using Keyword.merge/2 instead. If you're collecting into a non-empty list, consider concatenating the two lists with the ++ operator.
  (elixir 1.12.0-dev) lib/collectable.ex:98: Collectable.List.into/1
  (elixir 1.12.0-dev) lib/enum.ex:1382: Enum.into_protocol/2
  (stdlib 3.13.2) erl_eval.erl:680: :erl_eval.do_apply/6
  (elixir 1.12.0-dev) src/elixir.erl:280: :elixir.recur_eval/3
  (elixir 1.12.0-dev) src/elixir.erl:265: :elixir.eval_forms/3
  (iex 1.12.0-dev) lib/iex/evaluator.ex:276: IEx.Evaluator.handle_eval/3

[1, 2, 3, 4, 5, 6]
iex(14)> collectable ++ Enum.to_list(enumerable)
[1, 2, 3, 4, 5, 6]
iex(15)> Enum.into(enumerable, collectable, &Function.identity/1)
[1, 2, 3, 4, 5, 6]
```

josevalim: Given both keyword lists and lists are lists, we need a semantic that works for both. And the semantic that works for both is if the enumerable comes before the collectable (so the enumerable overrides collectable).

eksperimental: Sorry, I see it. When you insert into a list, you prepend, you do not append.

eksperimental: So when you call Keword.get/2 you get the key from enumerable?

lib/elixir/lib/enum.ex
```diff
@@ -1327,8 +1327,8 @@ defmodule Enum do
   Inserts the given `enumerable` into a `collectable`.
 
   Note that passing a non-empty list as the `collectable` is deprecated. If you're collecting
-  into a non-empty keyword list, consider using `Keyword.merge/2`. If you're collecting into a
-  non-empty list, consider something like `to_list(enumerable) ++ collectable`.
+  into a non-empty keyword list, consider using `Keyword.merge(collectable, Enum.to_list(enumerable))`.
+  If you're collecting into a non-empty list, consider something like `Enum.to_list(enumerable) ++ collectable`.
```
eksperimental, line 1331: @josevalim could you please have a look at these new changes.

josevalim:

PR #10617, opened 2021-01-02
Correct protocol error, when @fallback_to_any true and no implementation for Any

eksperimental:
given
```elixir
defprotocol WithFallBackToAnyFalse do
  def foo(term)
end

defprotocol WithFallBackToAnyTrue do
  @fallback_to_any true

  def foo(term)
end

```
before the fix:

```
iex> WithFallBackToAnyFalse.foo([])
** (Protocol.UndefinedError) protocol WithFallBackToAnyFalse not implemented for [] of type List. There are no implementations for this protocol.

iex> WithFallBackToAnyTrue.foo([])
** (UndefinedFunctionError) function nil.foo/1 is undefined

```
after the fix:

```
iex> WithFallBackToAnyTrue.foo([])
** (Protocol.UndefinedError) protocol WithFallBackToAnyTrue not implemented for [] of type List, while @fallback_to_any has been set to true, the implementation for Any is missing. There are no implementations for this protocol.
```

### DIFF
--- a/lib/elixir/lib/protocol.ex
+++ b/lib/elixir/lib/protocol.ex
@@ -767,12 +767,27 @@ defmodule Protocol do
       @spec impl_for!(term) :: atom
       if any_impl_for do
         Kernel.def impl_for!(data) do
-          impl_for(data)
+          try do
+            impl_for(data)
+          rescue
+            UndefinedFunctionError ->
+              raise_impl_for(data)
+          else
+            nil ->
+              raise_impl_for(data)
+
+            result ->
+              result
+          end
         end
       else
         Kernel.def impl_for!(data) do
-          impl_for(data) || raise(Protocol.UndefinedError, protocol: __MODULE__, value: data)
+          impl_for(data) || raise_impl_for(data)
         end
+      end
+
+      Kernel.defp raise_impl_for(value) do
+        raise(Protocol.UndefinedError, protocol: __MODULE__, value: value)
       end
 
       # Internal handler for Structs
--- a/lib/elixir/lib/protocol.ex
+++ b/lib/elixir/lib/protocol.ex
@@ -720,7 +720,9 @@ defmodule Protocol do
               unquote(__MODULE__.Any).__impl__(:target)
             rescue
               UndefinedFunctionError ->
-                nil
+                reraise Protocol.UndefinedError,
+                        [protocol: unquote(__MODULE__), value: :how_do_i_pass_this_value?],
+                        __STACKTRACE__
             end
           end
         else
--- a/lib/elixir/lib/protocol.ex
+++ b/lib/elixir/lib/protocol.ex
@@ -715,7 +715,14 @@ defmodule Protocol do
     quote bind_quoted: [built_in: __built_in__()] do
       any_impl_for =
         if @fallback_to_any do
-          quote do: unquote(__MODULE__.Any).__impl__(:target)
+          quote do
+            try do
+              unquote(__MODULE__.Any).__impl__(:target)
+            rescue
+              UndefinedFunctionError ->
+                nil
+            end
+          end
         else
           nil
         end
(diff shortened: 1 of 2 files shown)


josevalim:
Looks great, can you please add a test? I also think we can remove the special description from the error message... the `Any` protocol is usually implemented by the person that defines the protocol itself and this error is most likely to be read by a user of the protocol. Thanks!

eksperimental:
> Looks great, can you please add a test? I also think we can remove the special description from the error message... the `Any` protocol is usually implemented by the person that defines the protocol itself and this error is most likely to be read by a user of the protocol. Thanks!

Sure. Can you please tell me what the final message should read like?

josevalim:
It is the default message you get if you don't add a description key.

eksperimental:
@josevalim please have a look at this new implementation, as the previous one was wrong.

lib/elixir/lib/protocol.ex
```diff
@@ -767,14 +767,29 @@ defmodule Protocol do
...
         Kernel.def impl_for!(data) do
-          impl_for(data)
+          try do
+            impl_for(data)
+          rescue
+            UndefinedFunctionError ->
+              raise_impl_for(data)
+          else
+            nil ->
+              raise_impl_for(data)
```
eksperimental, line 777: I am not sure whether this will ever return `nil`, if not this can be removed.

eksperimental:
> Looks great, can you please add a test? I also think we can remove the special description from the error message... the `Any` protocol is usually implemented by the person that defines the protocol itself and this error is most likely to be read by a user of the protocol. Thanks!

I think we could emit a warning at compile time when we set @fallback_to_any to true, and no implementation for Any is defined.

lib/elixir/lib/protocol.ex
eksperimental: this is solved in the latest implementation

eksperimental:
I took a third approach which I think is simpler and correct

lib/elixir/lib/protocol.ex
```diff
@@ -715,7 +715,14 @@ defmodule Protocol do
...
       any_impl_for =
         if @fallback_to_any do
-          quote do: unquote(__MODULE__.Any).__impl__(:target)
+          quote do
+            try do
+              unquote(__MODULE__.Any).__impl__(:target)
+            rescue
+              UndefinedFunctionError ->
+                nil
+            end
```
josevalim, line 724: This is a change in behaviour. If you fallback to any, then it is always required, we never potentially return nil.

eksperimental: I see. Also note that we would still be raising in `impl_for/1`  if `@fallback_to_any true` and no implementation for Any exists

lib/elixir/lib/protocol.ex
```diff
@@ -720,7 +720,9 @@ defmodule Protocol do
               unquote(__MODULE__.Any).__impl__(:target)
             rescue
               UndefinedFunctionError ->
-                nil
+                reraise Protocol.UndefinedError,
+                        [protocol: unquote(__MODULE__), value: :how_do_i_pass_this_value?],
```
eksperimental, line 724: I tried for quiet some time to pass value to here to no avail. What' s the way to solve this?

josevalim: Right, because here is not really the correct place to raise. This could rather return `nil` so the parent raises it but as mentined, it would be a semantic change on behaviour.

eksperimental: I converted any_impl_for to an anonymous function of arity-1. The trouble that i have when it is currently called `unquote(any_impl_for)`  I tried to need to unquote `data` inside another `unquote` call, ending up with `unquote(any_impl_for.(unquote(data))`).

The other solution is modify `Protocol.UndefinedError` to accept `:message` and pass only this attribute.

josevalim: Right. The other thing to keep in mind is that this code is very performance sensitive. Wrapping the call in a try/catch can negatively affect performance. I think it is best to drop this. :/

eksperimental: How about something like this?
```elixir
      any_impl_for =
        if @fallback_to_any do
          quote do
            if function_exported?(unquote(__MODULE__.Any), :__impl__, 1) do
              unquote(__MODULE__.Any).__impl__(:target)
            else
              raise Protocol.UndefinedError,
                message: "protocol #{inspect(__MODULE__)} has no implementation for Any"
            end
          end
        else
          nil
        end
```

josevalim: It is not enough because the module may not have been loaded yet. You would need to add a Code.ensure_loaded? check which, once again, can be expensive. Given the performance concern and change of behaviour, we should drop this. Thank you!

eksperimental: Fair enough.

PR #10618, opened 2021-01-02
Document __impl__(:target)

eksperimental:


josevalim:
Target is an implementation detail, so I don’t think we should expose it for now unless we have an use case. Thanks!

PR #10626, opened 2021-01-05
Make copy-paste of piped multiline expressions to IEx work

am-kantox:
This is a very dubious PR, but I still think it brings a lot of value, specifically for newcomers.

Inspired by: https://stackoverflow.com/questions/65573781/using-pipe-on-next-line-in-iex-syntax-error-before

We already have `( ... )` and/or `fn -> ... end.()` in IEx to copy-paste the multiline pipelines. It might be handy sometimes to copy-paste them as is ‹and review results on each step›. Also, `|> IO.inspect()` might be handy, assuming the history queue is not empty.

I started with an implementation of macro `|>/1`, defined in `Helpers`, but then I realized it fits the `Evaluator` way better. No code update whatsoever, we just prepend `v()` to the input if the first line of it starts with `|>`.


josevalim:
I like the general idea but we need to make it more general if we are to apply it:

1. It needs to apply to every binary operator (except +/- which would be ambiguous with the binary ones)

2. You cannot trim whitespace in case there is no pipe operator, otherwise you change the semantics of things like spaces inside strings/heredocs

3. You should not look for the pipe operator if the current buffer is not empty

In other words, if you polish this a bit I think it can be a positive addition. :) Thank you!

am-kantox:
Sure, I like 1), I understand 2) and 3) is already in place, the clause has hardcoded `""` pattern-match for buffer.

lib/iex/test/iex/helpers_test.exs
```diff
@@ -1111,6 +1111,13 @@ defmodule IEx.HelpersTest do
     end
   end
 
+  describe "iex|1> |>" do
```
josevalim, line 1114:
```suggestion
  describe "iex> |>" do
```

josevalim:
> I understand 2) and 3) is already in place, the clause has hardcoded "" pattern-match for buffer.

Good call. That should be enough!

am-kantox:
Done.

The only tricky part I could think of was `< 5` to be distinguished from `<<rest::binary>>`. I might be missing something, though.

lib/iex/lib/iex/evaluator.ex
```diff
@@ -58,17 +58,37 @@ defmodule IEx.Evaluator do
   # https://github.com/elixir-lang/elixir/issues/1089 for discussion.
   @break_trigger "#iex:break\n"
 
+  @space " "
+  @leading_binary_operators ~w(|> * / ++ -- <= >= <> > < && || === !== == != =~ or and in)
```
josevalim, line 62: There are some operators missing, such as `<<<`, which is going to complicate your leading binary check. :) See elixir_parser.yrl for a complete list.

am-kantox:
```elixir
["not in" | ~w(-> <- when :: | => & = || ||| or && &&& and == != =~ === !==
 < > <= >= |> <<< >>> <<~ ~>> <~ ~> <~> <|> in ^^^ ++ -- .. <>
+++ --- + - * / + - ! ^ not ~~~ . @)] -- 
["not in", "or", "and", "in"] ++ ~w(
    ^^^ <<< >>> <<~ ~>> <~ ~> <~> <|> &&& ||| === !==
    |> ++ -- <= >= <> > < && || == != =~
    * /)
#⇒ ["->", "<-", "when", "::", "|", "=>", "&", "=", "..", "+++", "---",
#   "+", "-",
#   "+", "-", "!", "^", "not", "~~~", ".", "@"]
```

lib/iex/lib/iex/evaluator.ex
```diff
@@ -58,17 +58,49 @@ defmodule IEx.Evaluator do
   # https://github.com/elixir-lang/elixir/issues/1089 for discussion.
   @break_trigger "#iex:break\n"
 
+  @space " "
+  @leading_binary_operators ["not in", "or", "and", "in"] ++ ~w(
```
josevalim, line 62: This can lead to false positives if a variable is named `orion`. You will need to check if those are followed by a space. :)

lib/iex/lib/iex/evaluator.ex
```diff
@@ -58,17 +58,49 @@ defmodule IEx.Evaluator do
...
+    ^^^ <<< >>> <<~ ~>> <~ ~> <~> <|> &&& ||| === !==
+    |> ++ -- <= >= <> > < && || == != =~
+    * /)
```
josevalim, line 65:
```suggestion
    * /
  )
```

lib/iex/lib/iex/evaluator.ex
am-kantox: Followed by a space is easy, we might simply put a space into the pattern itself. But `:foo in[:foo]` is somehow valid syntax, as well as `true or+([:foo])`. I surely could check if the next symbol is an allowed-for-variable-name Unicode character, but I am not sure if we need to be as captious here.

So I personally would go with
```suggestion
  @leading_binary_operators ["not in ", "or ", "and ", "in "] ++ ~w(
```

Thoughts?

josevalim: I see. Then we are reimplementing the parser considerably which is a bit too much. I have plans to rewrite the parser so we could actually return this information more easily but this is quite some time ahead...

josevalim: I think a reasonable compromise for now is to support this only for the pipe operators in the .yrl file. WDYT?

am-kantox: Well, the main purpose of this PR is still to enable copy-paste blocks as is. The more I think the more hardly could I see why one would do `++ [:foo]` (and if needed, there is still `v() ++ [:foo]`.) So I would accept only what `Code.format_string!/2` welcomes at the beginning of the line. That might be tested as:

1. produce a bunch of snippets with the operator on another line
2. run `mix format` on it
3. see what goes where (I am almost sure only pipe operators are not pulled up to the previous line, but I have to check)
```

josevalim: I like the formatter idea. You can find the list of operators that can go on the next line in lib/code/formatter.ex: https://github.com/elixir-lang/elixir/blob/master/lib/elixir/lib/code/formatter.ex#L21:L25

josevalim: Actually, because when is only allowed in patterns, we don’t need to include it. There is probably no need to include | either, so just go with the first module attribute!

am-kantox: Maybe it’d make sense to internally export (`@doc false`) this list from `lib/code/formatter.ex` to borrow it here, but because of the aforementioned parser rewrite, I think it’s good enough as it is, at least for now.

josevalim:

PR #10627, opened 2021-01-05
Enhance Calendar.strftime docs

supersimple:
Make day of month documentation more clear
Show example while removing padding from day

At first glance, I saw the example in the days column as "01, 12" which made more sense as a month.
Also, I thought it would be nice to add an example with the padding removed.


josevalim:

PR #10630, opened 2021-01-07
clarify docs: terminate/2 not called on crash

davidsulc:


josevalim:
Can you please define what crashes mean? terminate is definitely called on errors within the GenServer itself, which you can verify by running this small program:

```elixir
defmodule G do
  def init(:ok) do
    {:ok, :ok}
  end

  def handle_call(_, _, _) do
    1 + :foo
  end

  def terminate(foo, bar) do
    IO.inspect {:terminate, foo, bar}
  end
end

{:ok, pid} = GenServer.start_link(G, :ok)
GenServer.call(pid, :foo)
```

davidsulc:
It looks like I was misinformed, or misunderstood: I was looking into what guarantees exist for `terminate/2` being executed on GenServer exit when I came across https://elixirforum.com/t/retrieving-state-from-exited-process/30428/3 where it's stated that "It is not called when the process crashes". Is this information obsolete, or am I misunderstanding the context of the reply?

josevalim:
I believe the question you need to ask in that thread is what the author meant by *crash*. If by crash they mean a broken link, then it is not called unless trapping exits. If they mean errors, then it is called. This has always been the behaviour.

PR #10634, opened 2021-01-08
Log when .iex.exs file is loaded when starting IEx.

TheFirstAvenger:
This simple enhancement came out of [this Phoenix PR](https://github.com/phoenixframework/phoenix/pull/4124) and the discussion about letting people know 1) that the `.iex.exs` file is even a feature and 2) that it has been loaded in the current session.

This PR simply outputs "Loaded .iex.exs" when a file is loaded. If the file is in a custom location, it will print the path relative to the current working directory.


josevalim:
I believe this is probably best places directly in the file. :) then you have complete control over what it does and prints!

TheFirstAvenger:
@josevalim Do you think there is a security or end-user-confusion case to be made for *always* noting when commands from .iex.exs is being loaded?

josevalim:
Yes and no. Invoking elixir commands will always load scripts around. Mix invoked mix.exs, IEx invoked this, the formatter invoked another file, etc.

TheFirstAvenger:
Ahh yes, great point!

PR #10635, opened 2021-01-08
Update String.printable? character_limit typespec

qcam:


josevalim:

lib/elixir/lib/string.ex
```diff
@@ -310,7 +310,7 @@ defmodule String do
 
   """
   @spec printable?(t, 0) :: true
-  @spec printable?(t, pos_integer | :infinity) :: boolean
+  @spec printable?(t, non_neg_integer | :infinity) :: boolean
```
qcam, line 313: @josevalim Actually I was wrong on this one 🙈 . `0` has already been covered in the spec above.

josevalim: And I saw the spec above and misread this one. :D

PR #10637, opened 2021-01-09
Support compression on Registry

benwilson512:
The original use case here is enabling a tuning option for Absinthe subscriptions. In testing this can improve memory usage by 3x for workloads where individual users submit a lot of subscriptions and therefore there is a lot of duplicate context content. Chatted w/ Jose about this a month or so ago, finally getting around to messing with it.

### DIFF
--- a/lib/elixir/lib/registry.ex
+++ b/lib/elixir/lib/registry.ex
@@ -354,13 +354,20 @@ defmodule Registry do
             "expected :listeners to be a list of named processes, got: #{inspect(listeners)}"
     end
 
+    compressed = Keyword.get(options, :compressed, false)
+
+    unless is_boolean(compressed) do
+      raise ArgumentError,
+            "expected :compressed to be a boolean, got: #{inspect(compressed)}"
+    end
+
     # The @info format must be kept in sync with Registry.Partition optimization.
     entries = [
       {@all_info, {keys, partitions, nil, nil, listeners}},
       {@key_info, {keys, partitions, nil}} | meta
     ]
 
-    Registry.Supervisor.start_link(keys, name, partitions, listeners, entries)
+    Registry.Supervisor.start_link(keys, name, partitions, listeners, entries, compressed)
   end
 
   @doc false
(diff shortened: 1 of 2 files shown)


lib/elixir/lib/registry.ex
```diff
@@ -354,13 +354,20 @@ defmodule Registry do
             "expected :listeners to be a list of named processes, got: #{inspect(listeners)}"
     end
 
+    compressed = Keyword.get(options, :compressed, false)
```
benwilson512, line 357: This is a minor deviation from the `:ets` API which just does `[:compressed]` but that isn't conventional Elixir.

lib/elixir/lib/registry.ex
```diff
@@ -1339,20 +1346,20 @@ defmodule Registry.Supervisor do
   @moduledoc false
   use Supervisor
 
-  def start_link(kind, registry, partitions, listeners, entries) do
-    arg = {kind, registry, partitions, listeners, entries}
+  def start_link(kind, registry, partitions, listeners, entries, compressed) do
+    arg = {kind, registry, partitions, listeners, entries, compressed}
```
benwilson512, line 1350: Seems like we could do some further refactoring to generate this tuple on line 370 and pass it through instead of having the proliferation of args.

lib/elixir/lib/registry.ex
```diff
@@ -1435,17 +1442,18 @@ defmodule Registry.Partition do
...
-      read_concurrency: true,
-      write_concurrency: true
-    ])
+  defp init_key_ets(:duplicate, key_partition, compressed) do
+    opts = [:duplicate_bag, :public, read_concurrency: true, write_concurrency: true]
+    :ets.new(key_partition, compression_opt(opts, compressed))
+  end
+
+  defp compression_opt(opts, compressed) do
+    if compressed, do: [:compressed] ++ opts, else: opts
```
bruteforcecat, line 1456: nit:  `[:compressed | opts]` should be slightly more performant and idiomatic?

josevalim: Both are equivalent when the left side of `++` is a literal. :)

josevalim:

PR #10642, opened 2021-01-11
Simplify code in Enum.into/2

eksperimental:


josevalim:
Given pattern matching can be more efficiently optimized, so we will stick with it. Thank you!

eksperimental:
> Given pattern matching can be more efficiently optimized, so we will stick with it. Thank you!

Good to know.
Is this a general rule, or it is specific to `is_struct` ?
So if I do %`{} = term`, is faster than `is_map(term)` ?

PR #10644, opened 2021-01-12
Refactor replace/4 to remove guards on replacement parameter

gcauchon:
#10500 was shared in _Elixir Weekly, issue 232_ on December 24th, so I read the story and looked at the PR more than a month after it was merged! Looking at the diff, I realized the `is_function` guard was now useless because it was transferred to the `precompile_replacement/1` function.

We can get rid of the guard, the function header and the private `do_replace/4` all at ounce.

🎩 to @dsdshcym for encouraging me to push my first (of I hope many) PR to the language!

### DIFF
--- a/lib/elixir/lib/regex.ex
+++ b/lib/elixir/lib/regex.ex
@@ -649,19 +649,7 @@ defmodule Regex do
 
   """
   @spec replace(t, String.t(), String.t() | (... -> String.t()), [term]) :: String.t()
-  def replace(regex, string, replacement, options \\ [])
-
-  def replace(regex, string, replacement, options)
-      when is_binary(string) and is_binary(replacement) and is_list(options) do
-    do_replace(regex, string, replacement, options)
-  end
-
-  def replace(regex, string, replacement, options)
-      when is_binary(string) and is_function(replacement) and is_list(options) do
-    do_replace(regex, string, replacement, options)
-  end
-
-  defp do_replace(%Regex{} = regex, string, replacement, options) do
+  def replace(%Regex{} = regex, string, replacement, options \\ []) do
     opts = if Keyword.get(options, :global) != false, do: [:global], else: []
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/regex.ex
```diff
@@ -649,19 +649,7 @@ defmodule Regex do
...
-    do_replace(regex, string, replacement, options)
-  end
-
-  def replace(regex, string, replacement, options)
-      when is_binary(string) and is_function(replacement) and is_list(options) do
-    do_replace(regex, string, replacement, options)
-  end
-
-  defp do_replace(%Regex{} = regex, string, replacement, options) do
+  def replace(%Regex{} = regex, string, replacement, options \\ []) do
```
josevalim, line 652:
```suggestion
  def replace(%Regex{} = regex, string, replacement, options \\ [])
      when is_binary(string) and is_list(options) do
```

josevalim:

lib/elixir/lib/regex.ex
gcauchon: @josevalim what are the advantages of adding guards on `string` and `options` when there is only one function signature?

josevalim: Better error messages if you give a bad arg!

gcauchon: This is what I though, based on the _new_ compiler check added in `1.11`!

josevalim: The compiler checks help who are writing the code but not those calling the functions. The improve error message for callers was added on v1.5 IIRC. :)

gcauchon: TRUE! Would it be considered a community convention/guideline to add more guards to "public" APIs, especially the ones from in the stdlib?

I’m asking more as a consistency guideline. Since many other public functions of the Regex module are not defined with guards.

PR #10645, opened 2021-01-12
Make mix.release script rand() work with busybox

chrta:
Combining dd, od and awk with the common arguments found in busybox and GNU coreutils as well BSD the function rand() can make correct id.

This is based on
https://github.com/erlware/relx/commit/200e12815ef665338cea1280244e9c4f9dfa06b9
and
https://github.com/erlware/relx/pull/828

I ran into this when adding a mix release to a minimal yocto image using https://github.com/meta-erlang/meta-erlang and a busybox with undefined CONFIG_DESKTOP.

### DIFF
--- a/lib/mix/lib/mix/tasks/release.init.ex
+++ b/lib/mix/lib/mix/tasks/release.init.ex
@@ -104,7 +104,7 @@ defmodule Mix.Tasks.Release.Init do
     export RELEASE_BOOT_SCRIPT_CLEAN="${RELEASE_BOOT_SCRIPT_CLEAN:-"start_clean"}"
 
     rand () {
-      od -t xS -N 2 -A n /dev/urandom | tr -d " \n"
+      dd count=1 bs=2 if=/dev/urandom 2> /dev/null | od -x | awk 'NR==1{print $2}'
     }
 
     release_distribution () {


chrta:
@tsloughter @hauleth @joaohf

lib/mix/lib/mix/tasks/release.init.ex
```diff
@@ -104,7 +104,7 @@ defmodule Mix.Tasks.Release.Init do
     export RELEASE_BOOT_SCRIPT_CLEAN="${RELEASE_BOOT_SCRIPT_CLEAN:-"start_clean"}"
 
     rand () {
-      od -t xS -N 2 -A n /dev/urandom | tr -d " \n"
+      dd count=1 bs=2 if=/dev/urandom 2> /dev/null | od -x  | head -n1 | awk '{print $2}'
```
hauleth, line 107: Beware that this will change the output a little, as earlier version didn't printed newline at the end of output. Additionally I would use `cut` instead of `awk`:

```sh
dd count=1 bs=2 if=/dev/urandom 2> /dev/null | od -x  | head -n1 | cut -d' ' -f2
```

 or reduce amount of the commands ran via:

```sh
dd count=1 bs=2 if=/dev/urandom 2> /dev/null | od -x  | awk 'NR==1{printf("%s",$2)}'
```

chrta: > Beware that this will change the output a little, as earlier version didn't printed newline at the end of output. 

Yes i saw that, but it seems to work on my pc and on the target (and is also used in relx).

> dd count=1 bs=2 if=/dev/urandom 2> /dev/null | od -x  | head -n1 | cut -d' ' -f2

This works on my linux pc, but not on the target using busybox. cut seems also be different there. So on the target this works (but it does not work on my pc)
```
dd count=1 bs=2 if=/dev/urandom 2> /dev/null | od -x  | head -n1 | cut -d' ' -f6
```

> dd count=1 bs=2 if=/dev/urandom 2> /dev/null | od -x  | awk 'NR==1{printf("%s",$2)}'

This works on both, the target and the pc.

chrta: > Beware that this will change the output a little, as earlier version didn't printed newline at the end of output.

According to posix https://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html#tag_02_06_03 newlines are removed at the end of command substitutions. This also works in busybox, so the additional `\n` at the end should not make a difference in this script.

tsloughter:
Just a note, not something that should be part of this particular PR: Eventually you can probably get rid of the random generation altogether since Erlang supports use of `undefined` as a node name to get a randomly generated name. This is only supported for both `name` and `sname` as off erts 11.1 but the relx script uses it if available https://github.com/erlware/relx/blob/master/priv/templates/extended_bin#L289-L292

lib/mix/lib/mix/tasks/release.init.ex
chrta: To sum this up, we have three choices here:
1. Close this pr and do not merge it -> mix releases do not work with busybox in a minimal config :worried:
2. Use the pr as it is now (see discussion above) :smile:
3. Change the pr to use `dd count=1 bs=2 if=/dev/urandom 2> /dev/null | od -x | awk 'NR==1{printf("%s",$2)}'` (see discussion above) :smile:

joaohf: Hi,

I'm in favor of option 3 or 2. When I sent the relx patch related to rand(), I did not explore the awk enough to get rid the head command.

hauleth: If newline at the end of output doesn't matter, then we can simplify the AWK call:

```
dd count=1 bs=2 if=/dev/urandom 2> /dev/null | od -x | awk 'NR==1{print $2}'
```

josevalim: @chrta does the command from @hauleth above work for you? If so, please update the PR and we can ship it!

chrta: Very good, it works fine.

josevalim:

josevalim:
Thank you!

Btw, for something called busybox, it looks quite empty in executables!

*cues laugh track*

chrta:
:) Thank you all for the fast feedback and merge, even if this is some niche use-case.

PR #10650, opened 2021-01-12
Add System.trap_signal/3 and System.untrap_signal/2

josevalim:
I decided to call them `trap_signal` instead of `on_signal` because we are overriding the default OS behaviour. Thanks @lukaszsamson for catching this.


lib/elixir/lib/system.ex
```diff
@@ -409,15 +425,127 @@ defmodule System do
...
+        {:error, :already_registered}
+      else
+        :ok =
+          :gen_event.add_handler(:erl_signal_server, {SignalHandler, gen_id}, {signal, handler})
+
+        :os.set_signal(signal, :handle)
+        {:ok, id}
+      end
+    end)
+  end
```
fertapric, line 517: To match the documentation:

```suggestion
  def trap_signal(signal, id \\ make_ref(), fun)
      when signal in @signals and is_function(fun, 0) do
    :elixir_config.serial(fn ->
      gen_id = {signal, id}

      if {SignalHandler, gen_id} in signal_handlers() do
        {:error, :already_registered}
      else
        :ok =
          :gen_event.add_handler(:erl_signal_server, {SignalHandler, gen_id}, {signal, fun})

        :os.set_signal(signal, :handle)
        {:ok, id}
      end
    end)
  end
```

lib/elixir/lib/system.ex
```diff
@@ -409,15 +425,127 @@ defmodule System do
...
+  Therefore, if you add traps to the signals above, the
+  default behaviour above will be executed after all user
+  signals.
+
+  **Important**: all signals run from a single process,
+  therefore, blocking the `fun` will block subsequent traps.
+  It is also not possible to add or remove traps from within
+  a trap itself.
+  """
+  @spec trap_signal(signal, (() -> :ok)) :: {:ok, reference()}
```
fertapric, line 499:
```suggestion
  """
  @doc since: "1.12.0"
  @spec trap_signal(signal, (() -> :ok)) :: {:ok, reference()}
```

lib/elixir/lib/system.ex
```diff
@@ -409,15 +425,127 @@ defmodule System do
...
+        :os.set_signal(signal, :handle)
+        {:ok, id}
+      end
+    end)
+  end
+
+  @doc """
+  Removes a previously registered `signal` with `id`.
+  """
+  def untrap_signal(signal, id) when signal in @signals do
```
fertapric, line 522:
```suggestion
  """
  @doc since: "1.12.0"
  @spec untrap_signal(signal, id) :: :ok | {:error, :not_found} when id: term
  def untrap_signal(signal, id) when signal in @signals do
```

lib/elixir/lib/system.ex
```diff
@@ -409,15 +425,127 @@ defmodule System do
...
+    :elixir_config.serial(fn ->
+      gen_id = {signal, id}
+
+      case :gen_event.delete_handler(:erl_signal_server, {SignalHandler, gen_id}, :delete) do
+        :ok ->
+          if not trapping?(signal) do
+            :os.set_signal(signal, :default)
+          end
```
michalmuskala, line 533: Does this work correctly if there were other handlers not coming from this API?

fertapric (approved):

lib/elixir/lib/system.ex
josevalim: Not really. We should probably document it relies on `:os.set_signal` and we may have conflicts if others are changing it directly?

michalmuskala: Yes, I do think it needs to be at least documented. I don't think there's a good solution here. Hopefully the signal API is not widely used in libraries

sorentwo: I agree that it should be documented and probably discouraged in the "Writing Libraries" documentation.

josevalim: @michalmuskala @sorentwo I have added a big disclaimer about usage of this function and a section on implementation notes. Please review. :)

lib/elixir/lib/system.ex
```diff
@@ -81,6 +81,22 @@ defmodule System do
...
+          :sighup
+          | :sigquit
+          | :sigabrt
+          | :sigalrm
+          | :sigterm
+          | :sigusr1
+          | :sigusr2
+          | :sigchld
+          | :sigstop
+          | :sigtstp
```
eksperimental, line 94: They can be sorted alphabetically:
```elixir
  @type signal ::
          :sigabrt
          | :sigalrm
          | :sigchld
          | :sighup
          | :sigquit
          | :sigstop
          | :sigterm
          | :sigtstp
          | :sigusr1
          | :sigusr2
```

lib/elixir/lib/system.ex
```diff
@@ -409,15 +425,159 @@ defmodule System do
...
+  acceptable for libraries to set their own traps is when
+  using Elixir in script mode, such as in `.exs` files and
+  via Mix tasks.
+
+  An optional `id` that uniquely identifies the function
+  can be given, otherwise a unique one is automatically
+  generated. If a previously registered `id` is given,
+  this function returns an error tuple. The `id` can be
+  used to remove a registered signal by calling
+  `untrap_signal/1`.
```
wojtekmach, line 480:
```suggestion
  `untrap_signal/2`.
```

lib/elixir/lib/system.ex
josevalim, line 94:
```suggestion
  @type signal ::
          :sigabrt
          | :sigalrm
          | :sigchld
          | :sighup
          | :sigquit
          | :sigstop
          | :sigterm
          | :sigtstp
          | :sigusr1
          | :sigusr2
```

josevalim:

PR #10658, opened 2021-01-15
[Proposal] Make tap/2 and then/2 macros

sabiwara:
This PR is a proposal for discussion. The original discussion on the mailing list can be found [here](
https://groups.google.com/u/1/g/elixir-lang-core/c/OhtFggrEHfU/m/BO6x5XleDAAJ)

The idea is to make `Kernel.tap/2` and `Kernel.then/2` macros instead of functions, in order to remove the runtime overhead.
This might be an overkill optimization, but since it is in the Kernel I thought it could make sense.

Looking forward to knowing your thoughts!

### DIFF
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -1147,10 +1147,11 @@ defmodule Kernel do
 
   """
   @doc since: "1.12.0"
-  @spec tap(value, (value -> any)) :: value when value: var
-  def tap(value, fun) when is_function(fun, 1) do
-    fun.(value)
-    value
+  defmacro tap(value, fun) do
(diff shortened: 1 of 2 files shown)


michalmuskala:
This might be specifically interesting because of the compiler optimisation where immediately called funs are never actually created and are lowered into a call to the internal function representing the lambda.
TBH I don't expect it to be significant in real-world code, but it might show up in benchmarks since it would avoid allocating the fun.

josevalim:
@michalmuskala I just thought the same. Although I can't recall if it has fully landed in Erlang/OTP.

The other thing is that a macro provides less guarantees, so it may be a better starting point.

josevalim:

PR #10664, opened 2021-01-21
add Enum.every?

rands0n:
Returns `true` if `fun.(element)` is truthy for all elements in `enumerable`.

Which differs from `Enum.all?` and `Enum.any?`. Example:

`Enum.all?` - Returns `true` when a empty enumerable is given;
`Enum.any?` - Returns `false` when a empty enumerable is given, but
every member on the list needs to return `true`, otherwise is `false`.

I think we can cover edge cases like this one:

https://github.com/elixir-lang/elixir/issues/1508

### DIFF
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -374,6 +374,58 @@ defmodule Enum do
   end
 
   @doc """
+  Returns `true` if `fun.(element)` is truthy for all elements in `enumerable`.
+
+  Iterates over the `enumerable` and invokes `fun` on each element. When an invocation
+  of `fun` returns a falsy value (neither `false` nor `nil`) iteration stops
+  immediately and `false` is returned. In all other cases `true` is returned.
+
+  ## Examples
+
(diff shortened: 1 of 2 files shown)


josevalim:
Hi @rands0n , can you please send this to the mailing list for discussion? I think this can be welcome but I am not sure this distinction between functions is clear in the name. I would like to hear about prior art and alternatives, such as all_non_empty? Or perhaps an anonymous function that customized the empty behavior, as done elsewhere in enum.

Thank you!

lib/elixir/lib/enum.ex
```diff
@@ -3538,6 +3590,24 @@ defmodule Enum do
     false
   end
 
+  ## every?
+
+  defp every_list([_], _) do
```
JonRowe, line 3595: I know this is closed, but if this gets reopened, wouldn't this function cause `every?([false])` to return true? You don't need a single match clause as `[false]` would be matches as `[false | []]` on 3599

PR #10665, opened 2021-01-21
Draft approach to conditionally define functions within module

am-kantox:
This is by no mean a PR-proposed-to-merge, this is just sharing the idea. I feel easier explaining thoughts with code rather than plain English, that’s the only reason you see this as a PR, not an issue.

**Proposal:** a built-in module attribute allowing to conditionally include the next function to the compiled beam, based on the compile-time availability of some other module.

---

Consider the library that might conditionally rely on other libraries. For the sake of simplicity, let’s write the universal rounding tool.

```elixir
def round(i) when is_integer(i), do: i
def round(f) when is_float(f), do: Float.round(f)
```

Now we want to support rounding from `Decimal` but we don’t want to silently bring `Decimal` to the consumer of our library. So we do somewhat like

```elixir
if {:module, Decimal} == Code.ensure_compiled(Decimal) do
  def round(%Decimal{} = d), do: Decimal.round(d)
end
```

The above breaks indenting and might actually be written as

```elixir
@if_available Decimal
def round(%Decimal{} = d), do: Decimal.round(d)
```

The code in this PR is a draft implementation of this. Spec included to demonstrate the behaviour.

---

**Drawbacks**

There are many. 

1. To gracefully handle `spec` for this function, too many changes are needed; the provided code silly prints a warning for it
1. I was unable to figure out how to handle warnings about duplicate `@doc`
1. It might look a bit counter-intuitive that the function happens to appear in the code while it actually is not included in the beam.

---

I would be happy to discuss the general idea, better implementations or listen to “this is absolutely not needed.”

### DIFF
--- a/lib/elixir/src/elixir_def.erl
+++ b/lib/elixir/src/elixir_def.erl
@@ -162,28 +162,51 @@ store_definition(Kind, CheckClauses, Call, Body, Pos) ->
 store_definition(Meta, Kind, CheckClauses, Name, Arity, DefaultsArgs, Guards, Body, File, ER) ->
   Module = ?key(ER, module),
   Tuple = {Name, Arity},
-  E = env_for_expansion(Kind, Tuple, ER),
 
-  {Args, Defaults} = unpack_defaults(Kind, Meta, Name, DefaultsArgs, E),
-  Clauses = [elixir_clauses:def(Clause, E) ||
-             Clause <- def_to_clauses(Kind, Meta, Args, Guards, Body, E)],
+  case is_definition_needed(Module, Tuple, Meta) of
+    false ->
+      ok;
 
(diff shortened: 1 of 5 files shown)


josevalim:
Hi @am-kantox, please submit this for discussion on the mailing list. You can either link to this PR or link to a GitHub diff.

My initial thought is that you don't need to use Core.ensure_compiled as it is only for modules within the same project, ensure_loaded? is fine. Therefore, we are asking to add all of this work to replace this:

```elixir
if Code.ensure_loaded?(Decimal) do
  ...
end
```

Which is clearer, just pure Elixir, so I personally don't think it is worth the trouble. Feel free to submit this to the mailing list for general feedback.

Thank you!

PR #10667, opened 2021-01-22
Explain why Enum.all?/2 returns true for empty enumerables

fxn:
As per the [discussion](https://groups.google.com/g/elixir-lang-core/c/8VfA1cc7qsw) in `elixir-core`.

### DIFF
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -300,7 +300,7 @@ defmodule Enum do
   As the last example shows, `Enum.all?/2` returns `true` if `enumerable` is
   empty, regardless of `fun`. In an empty enumerable there is no element for
   which `fun` returns a falsy value, so the result must be `true`. This is a
-  standard mathematical argument for empty collections.
+  well-defined logical argument for empty collections.
 
   If no function is given, the truthiness of each element is checked during iteration.
   When an element has a falsy value (`false` or `nil`) iteration stops immediately and
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -282,9 +282,9 @@ defmodule Enum do
   @doc """
   Returns `true` if `fun.(element)` is truthy for all elements in `enumerable`.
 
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/enum.ex
```diff
@@ -294,9 +294,14 @@ defmodule Enum do
...
       false
 
-      iex> Enum.all?([], fn x -> x > 0 end)
+      iex> Enum.all?([], fn x -> nil end)
       true
 
+  As the last example shows, `Enum.all?/2` returns `true` if `enumerable` is
+  empty, regardless of `fun`. In an empty enumerable there is no element for
+  which `fun` returns a falsy value, so the result must be `true`. This is a
+  standard mathematical argument for empty collections.
```
josevalim, line 303: What do you think about?

```suggestion
  which `fun` returns a falsy value, so the result must be `true`. This mimics
  the mathematical definition of similar quantifiers for empty collections.
```

Or at least, we can replace "standard" by "classical" or "well-defined". The issue is that "standard" has multiple meanings and it may not be clear which one we had in mind. In particular, I want to avoid the interpretation of "very familiar" or "well-established", which may lead a reader to think they should be aware with the mathematical interpretation, which is not strictly a requirement.

Another option is to replace mathematical by logical, if you think it is suitable. :)

fxn: Good point. Let me think.

fxn: What about 

> This is a standard logical argument for empty collections.

I am looking for something informal and lightweight that resonates to anyone without sounding "academic", to say it somehow.

josevalim: My biggest issue is with the word standard. Can we use one of "well-defined", "accepted", "classical", "common"? My preference is "well-defined" so we keep authoritative but without being too academic, as you said.

In other words:

```suggestion
  which `fun` returns a falsy value, so the result must be `true`. This is a
  well-defined logical argument for empty collections.
```

fxn: Perfect :).

josevalim:

PR #10680, opened 2021-01-29
Add functions for RFC 3986 compliant space encoding and decoding

florish:
This pull request is based on a proposal discussed and accepted here:

https://groups.google.com/g/elixir-lang-core/c/gMqIFBr4OTU

Please note that this is my first contribution to Elixir itself.

### DIFF
--- a/lib/elixir/lib/uri.ex
+++ b/lib/elixir/lib/uri.ex
@@ -82,17 +82,18 @@ defmodule URI do
 
   You can specify one of the following `encoding` strategies:
 
-    * `:rfc_3986` - (recommended) keys and values are encoded following
-      [RFC 3986](https://tools.ietf.org/html/rfc3986) specifications.
     * `:www_form` - (default) keys and values are URL encoded as per
-      `encode_www_form/1`. Same as `:rfc_3986`, except for encoding " " as "+"
-      instead of "%20".
+      `encode_www_form/1`. This is the format typically used by browsers
+      on query strings and form data. It encodes " " as "+".
+
+    * `:rfc_3986` - the same as `:www_form` except it encodes " " as
+      "%20" according [RFC 3986](https://tools.ietf.org/html/rfc3986).
+      This is the best option if you are encoding a non-http related
+      scheme, since encoding spaces as "+" can be ambiguous to URI
+      parsers. This can inadvertently lead to spaces being interpreted
+      as literal plus signs.
 
   Encoding defaults to `:www_form` for backward compatibility.
-
-  The main reason for recommending `:rfc_3986` is that encoding spaces as "+"
-  can be ambiguous to URI parsers. This can inadvertently lead to spaces being
-  interpreted as literal plus signs.
 
   ## Examples
 
@@ -322,6 +323,10 @@ defmodule URI do
   @doc """
   Encodes `string` as "x-www-form-urlencoded".
 
+  Note "x-www-form-urlencoded" is not specified as part of
+  RFC 3986. However, it is a commonly used format to encode
+  query strings and form data by browsers.
+
   ## Example
 
       iex> URI.encode_www_form("put: it+й")
@@ -368,6 +373,10 @@ defmodule URI do
 
   @doc """
   Decodes `string` as "x-www-form-urlencoded".
+
+  Note "x-www-form-urlencoded" is not specified as part of
+  RFC 3986. However, it is a commonly used format to encode
+  query strings and form data by browsers.
 
   ## Examples
 
--- a/lib/elixir/lib/uri.ex
+++ b/lib/elixir/lib/uri.ex
@@ -71,19 +71,7 @@ defmodule URI do
   end
 
   @doc """
-  Encodes an enumerable into a query string.
-
-  This function is deprecated. Use `encode_query/2` with `:www_form` as second
-  argument value to achieve the same result.
-  """
-  @doc deprecated: "Use URI.encode_query/2 instead"
-  @spec encode_query(Enum.t()) :: binary
-  def encode_query(enumerable) do
-    encode_query(enumerable, :www_form)
-  end
-
-  @doc """
-  Encodes an enumerable into a query string.
+  Encodes `enumerable` into a query string using `encoding`.
 
   Takes an enumerable that enumerates as a list of two-element
   tuples (for instance, a map or a keyword list) and returns a string
@@ -92,12 +80,15 @@ defmodule URI do
   Keys and values can be any term that implements the `String.Chars`
   protocol with the exception of lists, which are explicitly forbidden.
 
-  You must specify one of the following `encoding` strategies:
+  You can specify one of the following `encoding` strategies:
 
     * `:rfc3986` - (recommended) keys and values are encoded following
       [RFC3986](https://tools.ietf.org/html/rfc3986) specifications.
-    * `:www_form` - keys and values are URL encoded as per `encode_www_form/1`.
-      Same as RFC3986, except for encoding " " as "+" instead of "%20".
+    * `:www_form` - (default) keys and values are URL encoded as per
+      `encode_www_form/1`. Same as RFC3986, except for encoding " " as "+"
+      instead of "%20".
+
+  Encoding defaults to `:www_form` for backward compatibility.
 
   The main reason for recommending `:rfc3986` is that encoding spaces as "+" can
   be ambiguous to URI parsers. This can inadvertently lead to spaces being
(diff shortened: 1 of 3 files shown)


florish:
Hmm, this is odd. Checks are failing, but on the formatting of a file I haven't touched...

I'm not sure how to proceed, help is very much appreciated!

florish:
@josevalim Seems like this commit on master is causing the build to fail (and with that this PR's build):

https://github.com/elixir-lang/elixir/commit/e8678f339b2014d27cb1819ff86f3d5d3aa8c5e9#diff-3998c6417caf542fb1232839150458ebf885a73a4b53c3ff17ebcd0d66c0cf74R175

Just let me know once this has been fixed, and I'll merge master back into my forked branch (if rebasing on master is preferred, let me know).

lib/elixir/lib/uri.ex
```diff
@@ -73,42 +73,73 @@ defmodule URI do
...
   Encodes an enumerable into a query string.
 
+  This function is deprecated. Use `encode_query/2` with `:www_form` as second
+  argument value to achieve the same result.
+  """
+  @doc deprecated: "Use URI.encode_query/2 instead"
+  @spec encode_query(Enum.t()) :: binary
+  def encode_query(enumerable) do
+    encode_query(enumerable, :www_form)
+  end
```
dotdotdotpaul, line 83: Why deprecate, and not just make `:www_form` the default when called as a 1-arity function?

florish: Well, what you propose is what was my original proposal. Deprecation was suggested by someone else and this suggestion was seen as an improvement by @josevalim, specifically in this message:

https://groups.google.com/g/elixir-lang-core/c/gMqIFBr4OTU/m/uue1MSulDQAJ

I'm not specifically for of against deprecation, I'm happy to leave this decision to the Elixir core team.

josevalim:
No need to deprecate. :)

In theory we also need to support the same flag on decode_query, because we
shouldn’t touch + according to the RFC.
-- 


florish:
> No need to deprecate. :) In theory we also need to support the same flag on decode_query, because we shouldn’t touch + according to the RFC.

Sure! Question: how to proceed with the code? I could remove `encode_query/1` and make `:www_form` the default:

```
def encode_query(enumerable, encoding \\ :www_form)
```

But I think this would lead to `encode_query/1` to disappear from the Functions list on hexdocs.pm, which might not be a good thing.

lib/elixir/lib/uri.ex
florish: Ok, so no deprecation it is, per https://github.com/elixir-lang/elixir/pull/10680#issuecomment-770075143 👍

christhekeele:
> Sure! Question: how to proceed with the code? I could remove `encode_query/1` and make `:www_form` the default for `encode_query/2`:
> 
> ```
> def encode_query(enumerable, encoding \\ :www_form)
> ```
> 
> But I think this would lead to `encode_query/1` to disappear from the Functions list on hexdocs.pm, which might not be a good thing.

I think it's okay as we are trying to steer people to be explicit about this choice, and aware of its consequences. Trying to look up `encode_query/1` and being required to understand the options makes sense to me.

lib/elixir/lib/uri.ex
```diff
@@ -73,42 +73,73 @@ defmodule URI do
...
-  values are URL encoded as per `encode_www_form/1`.
+  in the form of `key1=value1&key2=value2...`.
 
   Keys and values can be any term that implements the `String.Chars`
   protocol with the exception of lists, which are explicitly forbidden.
 
+  You must specify one of the following `encoding` strategies:
+
+    * `:rfc3986` - (recommended) keys and values are encoded following
+      [RFC3986](https://tools.ietf.org/html/rfc3986) specifications.
```
christhekeele, line 98: Consideration: is `:rfc_3986` a little easier to parse visually? Or is there prior art here with other such identifiers within the stdlib?

florish: Good catch! I was under the impression that RFCs were spelled like this, but the page headers on [RFC 3986](https://tools.ietf.org/html/rfc3986) tell me otherwise. Will fix!

lib/elixir/lib/uri.ex
florish: This should be fixed now.

lib/elixir/lib/uri.ex
florish: Should be fixed.

florish:
Thanks for the quick feedback! I think I've processed all of it, and in the process also adjusted the documentation to reflect the changes. Should be good to go for another review round.

christhekeele:
LGTM, though should this note be addressed in the same PR @florish + @josevalim ?

> In theory we also need to support the same flag on decode_query, because we shouldn’t touch + according to the RFC.

josevalim:
Yes, we need to add decode_query as part of this PR for completeness. :)

lib/elixir/lib/uri.ex
```diff
@@ -71,44 +71,67 @@ defmodule URI do
...
 
+  You can specify one of the following `encoding` strategies:
+
+    * `:www_form` - (default) keys and values are URL encoded as per
+      `encode_www_form/1`. This is the format typically used by browsers
+      on query strings and form data. It encodes " " as "+".
+
+    * `:rfc_3986` - the same as `:www_form` except it encodes " " as
+      "%20" according [RFC 3986](https://tools.ietf.org/html/rfc3986).
+      This is the best option if you are encoding a non-http related
```
fertapric, line 91: Nitpicks:

```suggestion
    * `:rfc_3986` - the same as `:www_form` except it encodes " " as
      "%20" according to [RFC 3986](https://tools.ietf.org/html/rfc3986).
      This is the best option if you are encoding a non-HTTP related
```

fertapric: @josevalim should we add the `(since v1.12.0)` comments on the flags?

josevalim: Yes, we should. :) Please add suggestions? :)

lib/elixir/lib/uri.ex
```diff
@@ -71,44 +71,67 @@ defmodule URI do
...
   Keys and values can be any term that implements the `String.Chars`
   protocol with the exception of lists, which are explicitly forbidden.
 
+  You can specify one of the following `encoding` strategies:
+
+    * `:www_form` - (default) keys and values are URL encoded as per
+      `encode_www_form/1`. This is the format typically used by browsers
+      on query strings and form data. It encodes " " as "+".
+
+    * `:rfc_3986` - the same as `:www_form` except it encodes " " as
```
fertapric, line 89:
```suggestion
    * `:www_form` - (default, since v1.12.0) keys and values are URL encoded as per
      `encode_www_form/1`. This is the format typically used by browsers
      on query strings and form data. It encodes " " as "+".

    * `:rfc_3986` - (since v1.12.0) the same as `:www_form` except it encodes " " as
```

lib/elixir/lib/uri.ex
florish, line 91: Considearion: "non-http scheme" might give the impression that "+" space encoding is OK for all http(s) situations. I ran into a "+ interpreted as literal plus" situation while connecting to an HTTP(S) web API. If I understand the specs correctly, even for HTTP URI's, this is technically correct behaviour.

Maybe something like "non-browser situation" would be better?

florish:
> Yes, we need to add decode_query as part of this PR for completeness. :)

Sure! I'm looking at the `decode_query/2` code right now and the `IO.warn("URI.decode_query/2 is deprecated, please use URI.decode_query/1")` statements here confuse me a bit:

* https://github.com/florish/elixir/blob/uri-encode-query-2/lib/elixir/lib/uri.ex#L161
* https://github.com/florish/elixir/blob/uri-encode-query-2/lib/elixir/lib/uri.ex#L170

As this points me to a function with arity 1, but I don't really see a `URI.encode_query/1` function anywhere in the code, I'm not sure how to move forward with adding an extra `encoding \\ :www_form` argument here.

Any pointers would be really appreciated!

lib/elixir/lib/uri.ex
josevalim: Sounds good, please change! :)

josevalim:
@florish functions in Elixir are indentified by name and arity, which means we are defining both /1 and /2 versions at once.

We should change the warning to say that passing anything but a map as second argument is deprecated. And the :www_form and :rfc_3986 are the third argument.

florish:
> @florish functions in Elixir are indentified by name and arity, which means we are defining both /1 and /2 versions at once.

Ah, thanks! I kind of knew, but the `IO.warn` statement confused my understanding a bit.

> We should change the warning to say that passing anything but a map as second argument is deprecated.

I'll leave that to another PR if that's OK, as it is outside the scope of the current feature.

> And the :www_form and :rfc_3986 are the third argument.

Clear. I'm working on this right now, and it seems like I also need to extend `URI.query_decoder/1` to accept a second argument (`:www_form | :rfc_3986`) as this function is using the same logic as `decode_query/2` under the hood:

```
  def query_decoder(query) when is_binary(query) do
    Stream.unfold(query, &decode_next_query_pair/1)
  end
```

Will then become:

```
  def query_decoder(query, encoding \\ :www_form) when is_binary(query) do
    Stream.unfold(query, &decode_next_query_pair(&1, encoding))
  end
```

Feels like a logical change to me, but just checking here to be sure.

josevalim:
No, the atoms should be added as a third argument. The second argument will continue to default to a map - it is already supported today, which is why the warning is wrong. :)

florish:
> No, the atoms should be added as a third argument. The second argument will continue to default to a map - it is already supported today, which is why the warning is wrong. :)

Yes, this is true for `decode_query/3`, but I was referring to`query_decoder/1`, which is a different function (see the two code snippets).

josevalim:
Yes, you are right in regards to the query_decoder!

lib/elixir/lib/uri.ex
florish: Fixed in https://github.com/elixir-lang/elixir/pull/10680#discussion_r567232540

florish: Changed in https://github.com/elixir-lang/elixir/pull/10680/commits/941c76bb2d3b8ca59e0f88ce1a6ee9a9061943e2

florish:
I think I've processed all feedback, so... time for another review round :)

josevalim:

PR #10681, opened 2021-01-30
chore: fix callbacks amount in GenServer docs

milmazz:
I guess that the original text was introduced before
`handle_continue/2`, but now the `GenServer` behaviour define 8
callbacks:

* code_change/3
* format_status/2
* handle_call/3
* handle_cast/2
* handle_continue/2
* handle_info/2
* init/1
* terminate/2


josevalim:

PR #10682, opened 2021-01-31
Fix `@impl` warning when defdelegate is used

### LINKED ISSUE #10593: `@impl` is skipping the warnings when defdelegate is used
### Environment

* Elixir & Erlang/OTP versions (elixir --version): Elixir master, Erlang/OTP 23
* Operating system: Linux

### Current behavior
```elixir
defprotocol ImplDefdelegate do
  @moduledoc """
  Documentation for `ImplDefdelegate`.
  """
  def foo(term)

  def bar(term)
end

defimpl ImplDefdelegate, for: List do
  @impl true
  def foo(_) do
    :list
  end

  defdelegate bar(list), to: __MODULE__, as: :baz

  def baz(_) do
    :list
  end
end
```

```
$ mix test
Compiling 2 files (.ex)
Generated impl_defdelegate app


Finished in 0.02 seconds
0 failures

```
no warnings,

### Expected behavior

A warning ("module attribute @impl was not set for function bar/1 callback") should be given

eksperimental:
Fixes #10593

Fixing bug #10593 unveiled another bug, which is that we are not setting the context as the one where `defdelegate/2` is being called from so since the macro is defined in Kernel, this is the module that was being set.
That is fixed, but the commit fixes the initial issue publish, but when using a behaviour instead of a protocol, it will not work.

The best solution which will be more performant is if we can annotate the AST code generated by the `defdelegate/2` macro with the keyword `delegated: true, but updating the meta field in the AST when returning did not work.

### DIFF
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -5248,7 +5248,9 @@ defmodule Kernel do
 
         @doc delegate_to: {target, as, :erlang.length(as_args)}
 
-        def unquote(name)(unquote_splicing(args)) do
+        # Build the call AST by hand so it doesn't get a
+        # context and it warns on things like missing @impl
+        def unquote({name, [line: __ENV__.line], args}) do
           unquote(target).unquote(as)(unquote_splicing(as_args))
         end
       end
--- a/lib/elixir/test/elixir/kernel/impl_test.exs
+++ b/lib/elixir/test/elixir/kernel/impl_test.exs
@@ -16,6 +16,11 @@ defmodule Kernel.ImplTest do
     on_exit(fn -> purge(Kernel.ImplTest.ImplAttributes) end)
   end
 
+  defprotocol AProtocol do
+    def foo(term)
+    def bar(term)
+  end
+
   defmodule Behaviour do
     @callback foo() :: any
   end
@@ -495,11 +500,15 @@ defmodule Kernel.ImplTest do
       capture_err(fn ->
         Code.eval_string("""
         defmodule Kernel.ImplTest.ImplAttributes do
+          @behaviour Kernel.ImplTest.Behaviour
           use Kernel.ImplTest.UseBehaviourWithoutImpl
 
           @impl true
           def bar_without_impl(), do: :overridden
           def baz_without_impl(), do: :overridden
+
+          defdelegate foo(), to: __MODULE__, as: :baz
+          def baz(), do: :ok
         end
         """)
       end)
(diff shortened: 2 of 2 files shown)


josevalim:
Hi @eksperimental! Looks good although I am not sure if this is the best fix... can you please add a test so I can check it out? Thank you! :)

eksperimental:
> Hi @eksperimental! Looks good although I am not sure if this is the best fix... can you please add a test so I can check it out? Thank you! :)

I agree. I am not sure about the fix,  but it is the first stab.
Please have a look at the test (the protocol test is passing, the behaviour test is not)

eksperimental:
Note: if we annotate the AST with `:delegated`, in Module.missing_impl_in_context? we don't need to pass the target, and we can delete `meta[:context]` to mimic the locally defined functions, and we don't need to worry about whether defdelegate was used in a behaviour, a protocol, and we get rid of` target_from_defs/1`

josevalim:
Thank you @eksperimental. I have played with this and it seems the best fix for this is to *not* add a context to defdelegate. I did this diff:

```diff
diff --git a/lib/elixir/lib/kernel.ex b/lib/elixir/lib/kernel.ex
index a8a40aaa1..8ddb73a29 100644
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -5225,6 +5225,7 @@ defmodule Kernel do
           opts
       end
 
+    ast =
     quote bind_quoted: [funs: funs, opts: opts] do
       target =
         Keyword.get(opts, :to) || raise ArgumentError, "expected to: to be given as argument"
@@ -5253,6 +5254,8 @@ defmodule Kernel do
         end
       end
     end
+
+    Macro.prewalk(ast, &Macro.update_meta(&1, fn meta -> Keyword.delete(meta, :context) end))
   end
 
   ## Sigils

```

And it also solves the problem. However, we cannot simply erase all of the context, because that may affect other areas, such as variable hygiene and so on. The quickest workaround I found was this diff:

```diff
diff --git a/lib/elixir/lib/kernel.ex b/lib/elixir/lib/kernel.ex
index a8a40aaa1..0a43603fa 100644
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -5248,7 +5248,9 @@ defmodule Kernel do
 
         @doc delegate_to: {target, as, :erlang.length(as_args)}
 
-        def unquote(name)(unquote_splicing(args)) do
+        # Build the call AST by hand so it doesn't get a
+        # context and it warns on things like missing @impl
+        def unquote({name, [line: __ENV__.line], args}) do
           unquote(target).unquote(as)(unquote_splicing(as_args))
         end
       end

```

I.e. we build the call AST by hand so it doesn't have the context. it is not pretty but it should work. So I would recommend going down this route for now. Thoughts?

eksperimental:
I am looking into this.
Quick question: I took a similar approach, and this this at the end defdelegate
`Macro.prewalk(ast, &Macro.update_meta(&1, fn x -> Keyword.put(x, :delegated, true) end)) `
but the :delegated key was never added to meta. Do you know why?

josevalim:
I believe we filter the AST nodes that we pass forward at some point.

eksperimental:
> I believe we filter the AST nodes that we pass forward at some point.

So, what would be the way to do that?

eksperimental:
> I.e. we build the call AST by hand so it doesn't have the context. it is not pretty but it should work. So I would recommend going down this route for now. Thoughts?

I think this is the easiest way to go. I have updated the PR

josevalim:

eksperimental:
Awesome

PR #10683, opened 2021-02-01
Fix context in defdelagate/2 AST

eksperimental:
Add context as the current module where it is defined.
Related issue: https://github.com/elixir-lang/elixir/issues/10593


josevalim:
Quick question: why is this important? What changes if we don't merge this vs we do merge this? Generally speaking I would prefer if we don't mess with contexts. :)

eksperimental:
If we are not taking the route of adding the context to fix the other bug, it changes nothing.

PR #10686, opened 2021-02-03
Interpolate module attributes in match assertions

### LINKED ISSUE #10675: Show module attribute values in ExUnit match diffs
### Environment

```
Erlang/OTP 23 [erts-11.1.3] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [hipe]

Elixir 1.11.2 (compiled with Erlang/OTP 23)
```

* Operating system: Linux Mint

### Current behavior

Right now if I write the following tests:

```elixir
defmodule MyMod do
  use ExUnit.Case, async: true
  
  @num 3
  test "failure example" do
    string = "other"
    assert %{num: @num, other: string} == %{num: 4, other: "other"}
  end

  test "other example" do
    string = "other"
    assert %{num: @num, other: ^string} = %{num: 4, other: "other"}
  end
end
```

I can see the value of `@num` in the diff for the assertion using `==`, but not for the assertion using `=`.

```
  1) test failure example (MyMod)
     test/my_test.exs:5
     match (=) failed
     The following variables were pinned:
       string = "other"
     code:  assert %{num: @num, other: ^string} = %{num: 4, other: "other"}
     left:  %{num: @num, other: ^string}
     right: %{num: 4, other: "other"}

  2) test other example (MyMod)
     test/my_test.exs:10
     Assertion with == failed
     code:  assert %{num: @num, other: string} == %{num: 4, other: "other"}
     left:  %{num: 3, other: "other"}
     right: %{num: 4, other: "other"}
```

### Expected behavior

I think it would be nice to either interpolate the value of `@num` in the `left` field in the diff like we do for the `==` diff, or to show it in a way similar to what we do with `pinned_variables`. The most helpful would be to interpolate directly, so that would look something like this:

```
  1) test failure example (MyMod)
     test/my_test.exs:5
     match (=) failed
     The following variables were pinned:
       string = "other"
     code:  assert %{num: @num, other: ^string} = %{num: 4, other: "other"}
     left:  %{num: 3, other: ^string}
     right: %{num: 4, other: "other"}
```

but just showing the value in the diff would also be helpful, so something like this:

```
  1) test failure example (MyMod)
     test/my_test.exs:5
     match (=) failed
     The following variables were pinned:
       string = "other"
       @num = 3
     code:  assert %{num: @num, other: ^string} = %{num: 4, other: "other"}
     left:  %{num: @num, other: ^string}
     right: %{num: 4, other: "other"}
```

Having not looked too deeply into it just yet, I'm thinking it may not be entirely possible to do this given how tricky just showing the diffs for match errors is, but if it is possible it would sure be nice!

devonestes:
This will interpolate the value of module attributes in match assertions
in ExUnit. Basically, the old diff would look like this:

```
  1) test assert match with module attribute (ExUnit.AssertionsTest)
     test/ex_unit/assertions_test.exs:186
     match (=) failed
     code:  assert {2, @test_mod_attribute} = Value.tuple()
     left:  {2, @test_mod_attribute}
     right: {2, 1}
     stacktrace:
       test/ex_unit/assertions_test.exs:187: (test)
```

And now it will look like this:

```
  1) test assert match with module attribute (ExUnit.AssertionsTest)
     test/ex_unit/assertions_test.exs:186
     match (=) failed
     code:  assert {2, @test_mod_attribute} = Value.tuple()
     left:  {2, 2}
     right: {2, 1}
     stacktrace:
       test/ex_unit/assertions_test.exs:187: (test)
```

Resolves #10675

### DIFF
--- a/lib/ex_unit/lib/ex_unit/assertions.ex
+++ b/lib/ex_unit/lib/ex_unit/assertions.ex
@@ -696,6 +696,9 @@ defmodule ExUnit.Assertions do
       {:__aliases__, _, _} = expr ->
         Macro.expand(expr, caller)
 
+      {:@, _, [{attribute, _, _}]} ->
+        caller.module |> Module.get_attribute(attribute) |> Macro.escape()
+
       {left, meta, right} = expr ->
         case Macro.expand(expr, caller) do
           ^expr -> expr
--- a/lib/ex_unit/test/ex_unit/assertions_test.exs
+++ b/lib/ex_unit/test/ex_unit/assertions_test.exs
@@ -182,6 +182,16 @@ defmodule ExUnit.AssertionsTest do
     assert vec(x: ^x, y: ^y) = vec(x: x, y: y, z: z)
   end
 
+  @test_mod_attribute %{key: :value}
+  test "assert match with module attribute" do
+    try do
+      assert {@test_mod_attribute, 1} = Value.tuple()
+    rescue
+      error in [ExUnit.AssertionError] ->
+        assert "{%{key: :value}, 1}" == Macro.to_string(error.left)
+    end
+  end
+
   test "assert match with pinned variable" do
     a = 1
     {2, 1} = assert {2, ^a} = Value.tuple()


ericmj (approved):

josevalim:

PR #10687, opened 2021-02-04
Support parsing of ordinal ISO-8601 dates

christhekeele:


lib/elixir/lib/calendar/iso.ex
```diff
@@ -210,25 +234,62 @@ defmodule Calendar.ISO do
...
+  end
+
+  defp validate_parsed_date(year, month, day) do
+    if valid_date?(year, month, day) do
+      {:ok, {year, month, day}}
+    else
+      {:error, :invalid_date}
+    end
+  end
+
```
christhekeele, line 267: Is there a better name for the function below? Is it possibly a valuable addition to the calendar public interface protocol at large? Alternatively, could it be added to the top-level `Calendar` public API since it should work for anything supporting the `Calendar` protocol?

lib/elixir/lib/calendar/iso.ex
```diff
@@ -210,25 +234,62 @@ defmodule Calendar.ISO do
...
+  defp calendar_day_of_year(year, month, ordinal_day) do
+    days_in_month = days_in_month(year, month)
+
+    if ordinal_day > days_in_month do
+      calendar_day_of_year(year, month + 1, ordinal_day - days_in_month)
+    else
+      {year, month, ordinal_day}
+    end
+  end
+
```
christhekeele, line 281: Is the function below possibly a valuable addition to the calendar public interface protocol at large? Alternatively, could it be added to the top-level `Calendar` public API since it should work for anything supporting the `Calendar` protocol?

Notably, function comes up in the discussion [here](https://elixirforum.com/t/how-to-support-multiple-week-calendars-in-elixir/18783/16) around the [cldr_dates_times](https://github.com/elixir-cldr/cldr_dates_times) library, but I can't actually find an implementation in the source anymore.

lib/elixir/lib/calendar/iso.ex
```diff
@@ -280,50 +357,92 @@ defmodule Calendar.ISO do
...
-          {:error, :invalid_time}
+  defp parse_naive_datetime(
+         <<unquote(match_ordinal_date), sep, unquote(match_time), rest::binary>>,
+         multiplier
+       )
+       when unquote(guard_ordinal_date) and sep in @sep and unquote(guard_time) do
+    with {microsecond, rest} <- parse_microsecond(rest),
+         {_offset, ""} <- parse_offset(rest) do
+      {year, ordinal_day} = unquote(read_ordinal_date)
+
```
christhekeele, line 385: Worth noting: the implementation of `calendar_day_of_year` returns the year as well, and could trivially "roll over" the year if given an ordinal day beyond its range (my original impl did this). However, this is against the ISO-8601 spec, so we handle it here instead, and similarly in the other parsing functions.

lib/elixir/lib/calendar/iso.ex
```diff
@@ -210,25 +234,62 @@ defmodule Calendar.ISO do
...
+    months_in_year(year)
+    |> Range.new(1)
+    |> Enum.reduce(0, fn month, days -> days + days_in_month(year, month) end)
+  end
+
   @doc """
   Parses a naive datetime string.
 
+  The date part of the string supports both ISO-8601 calendar dates and ordinal dates.
+
```
christhekeele, line 292: This function's examples are rapidly becoming a test suite, and probably too expansive to be helpful documentation. I'd like to break them out into an actual suite and leave behind mostly instructive examples, but am unsure which ones are most valuable to keep.

lib/elixir/lib/calendar/iso.ex
```diff
@@ -210,25 +234,62 @@ defmodule Calendar.ISO do
...
+  defp validate_parsed_date(year, month, day) do
+    if valid_date?(year, month, day) do
+      {:ok, {year, month, day}}
+    else
+      {:error, :invalid_date}
+    end
+  end
+
+  defp calendar_day_of_year(year, ordinal_day) do
```
christhekeele, line 268: Is there a better name for this function? Is it possibly a valuable addition to the calendar public interface protocol at large? Alternatively, could it be added to the top-level `Calendar` public API since it should work for anything supporting the `Calendar` protocol?

lib/elixir/lib/calendar/iso.ex
```diff
@@ -210,25 +234,62 @@ defmodule Calendar.ISO do
...
+    days_in_month = days_in_month(year, month)
+
+    if ordinal_day > days_in_month do
+      calendar_day_of_year(year, month + 1, ordinal_day - days_in_month)
+    else
+      {year, month, ordinal_day}
+    end
+  end
+
+  defp days_in_year(year) do
```
christhekeele, line 282: Is this function possibly a valuable addition to the calendar public interface protocol at large? Alternatively, could it be added to the top-level `Calendar` public API since it should work for anything supporting the `Calendar` protocol?

Notably, function comes up in the discussion [here](https://elixirforum.com/t/how-to-support-multiple-week-calendars-in-elixir/18783/16) around the [cldr_dates_times](https://github.com/elixir-cldr/cldr_dates_times) library, but I can't actually find an implementation in the source anymore.

lib/elixir/lib/calendar/iso.ex
```diff
@@ -280,50 +357,92 @@ defmodule Calendar.ISO do
...
+  defp parse_naive_datetime(
+         <<unquote(match_ordinal_date), sep, unquote(match_time), rest::binary>>,
+         multiplier
+       )
+       when unquote(guard_ordinal_date) and sep in @sep and unquote(guard_time) do
+    with {microsecond, rest} <- parse_microsecond(rest),
+         {_offset, ""} <- parse_offset(rest) do
+      {year, ordinal_day} = unquote(read_ordinal_date)
+
+      if ordinal_day > days_in_year(year) do
```
christhekeele, line 386: Worth noting: the implementation of `calendar_day_of_year` returns the year as well, and could trivially "roll over" the year if given an ordinal day beyond its range (my original impl did this). However, this is against the ISO-8601 spec, so we handle it here instead, and similarly in the other parsing functions.

lib/elixir/lib/calendar/iso.ex
radar: It's referred to as the ordinal day in the spec (iirc), so probably `ordinal_day`.

christhekeele: The variable here is `ordinal_day`, and the function is meant to turn that into what the spec calls a "calendar day" (year, month, day). It's a little awkward tho b/c nothing else in the module refers to this construct as a "calendar day".

lib/elixir/lib/calendar/iso.ex
josevalim: Yes, that's a great idea regardless of this PR.

lib/elixir/lib/calendar/iso.ex
```diff
@@ -102,6 +102,19 @@ defmodule Calendar.ISO do
       ]
     end
 
+  [match_ordinal_date, guard_ordinal_date, read_ordinal_date] =
+    quote do
+      [
+        <<y1, y2, y3, y4, ?-, d1, d2, d3>>,
+        y1 >= ?0 and y1 <= ?9 and y2 >= ?0 and y2 <= ?9 and y3 >= ?0 and y3 <= ?9 and y4 >= ?0 and
+          y4 <= ?9 and d1 >= ?0 and d1 <= ?9 and d2 >= ?0 and d2 <= ?9 and d3 >= ?0 and d3 <= ?9,
```
Eiji7, line 110: How about using making it short with `in`:
```suggestion
        y1 in ?0..?9 and y2 in ?0..?9 and y3 in ?0..?9 and y4 in ?0..?9 and
          d1 in ?0..?9 and d2 in ?0..?9 and d3 in ?0..?9,
```
or even:
```suggestion
        y1 in '09' and y2 in '09' and y3 in '09' and y4 in '09' and
          d1 in '09' and d2 in '09' and d3 in '09',
```

lib/elixir/lib/calendar/iso.ex
```diff
@@ -102,6 +102,19 @@ defmodule Calendar.ISO do
...
+        {
+          (y1 - ?0) * 1000 + (y2 - ?0) * 100 + (y3 - ?0) * 10 + (y4 - ?0),
+          (d1 - ?0) * 100 + (d2 - ?0) * 10 + (d3 - ?0)
+        }
```
Eiji7, line 114: We have a function for that:
```suggestion
        {List.to_integer([y1, y2, y3, y4]), List.to_integer([d1, d2, d3])}
```

Eiji7 (commented):
Just 2 small changes making code shorter and easier to read. I wrote it by hand, so please test those changes.

lib/elixir/lib/calendar/iso.ex
christhekeele: Interestingly, the second form `?5 in '09' doesn't seem to work for me, but I do like the former.

There might be a performance implication here though--is list inclusion faster or slower than two equality checks? These are guards and compile-time lists so I assume either is awfully fast.

josevalim: We originally wrote in this format because it skips an `is_integer` check compared to ranges. But it is likely that the is_integer check is optimized away today. @michalmuskala would probably know. :)

Eiji7: @christhekeele Ah, I see … as said I wrote it by hand … Should be:
```elixir
iex> ?5 in '0123456789'
true
```
In that case `0..9` range is shorter.

lib/elixir/lib/calendar/iso.ex
christhekeele: I'll do this in a separate PR, by describing what formats we expect/support a little better textually in the moduledoc (do moduledocs support doctests?), making sure all current examples are represented in the test suite, then tightening these up for all functions in the module to cover just happiest-path and examples of distinct errors. Then rebase this off it.

There's enough in-flight discussion on the mailing list about what we want to support here, though, that I'll wait for that to reach a conclusion.

PR #10688, opened 2021-02-04
Fix spelling mistake in release docs

mackshkatz:


fertapric:
Thanks @mackshkatz ! 💜

PR #10689, opened 2021-02-05
Improve Calendar.ISO documentation.

christhekeele:
This makes it more clear how we've chosen to support ISO 3601.

Other changes targeted for this PR, per discussion on [the mailing list](https://groups.google.com/g/elixir-lang-core/c/CcXpeMQhsmU/m/o2b2_HM6DwAJ) and predecessor #10687:

- [x] Support leading `+` signs before years
- [x] Add tests and examples for leading `+` signs before years
- [x] ~Simplify parsing examples to essentials and move excess into a test suite [per this note](https://github.com/elixir-lang/elixir/pull/10687#discussion_r569878601)~ (implemented in #10692 instead)
- [x] ~Simplify macro-injected string-to-integer parsing [per this note](https://github.com/elixir-lang/elixir/pull/10687#discussion_r570018035)~ (only if this code gets touched in a future PR)
- [x] ~Simplify macro-injected guard expressions [per this note](https://github.com/elixir-lang/elixir/pull/10687#discussion_r570016660) (maybe, if there are no perf regressions)~ (only if this code gets touched in a future PR)

### DIFF
--- a/lib/elixir/lib/calendar/iso.ex
+++ b/lib/elixir/lib/calendar/iso.ex
@@ -33,7 +33,7 @@ defmodule Calendar.ISO do
   `Calendar.strftime/2` allows you to format datetimes however else you desire.
 
   Other optional ISO 8601 features; such as ordinal dates, week dates,
-  durations, time intervals, truncated representations, and reduced precision;
+  durations, time intervals, and reduced precision;
   are not supported by the parser or formatters.
 
   ### Extensions
--- a/lib/elixir/lib/calendar/iso.ex
+++ b/lib/elixir/lib/calendar/iso.ex
@@ -237,6 +237,9 @@ defmodule Calendar.ISO do
   def parse_date("-" <> string) when is_binary(string),
     do: parse_date(string, -1)
 
+  def parse_date("+" <> string) when is_binary(string),
+    do: parse_date(string, 1)
+
   def parse_date(string) when is_binary(string),
     do: parse_date(string, 1)
 
@@ -307,6 +310,9 @@ defmodule Calendar.ISO do
   def parse_naive_datetime("-" <> string) when is_binary(string),
     do: parse_naive_datetime(string, -1)
 
+  def parse_naive_datetime("+" <> string) when is_binary(string),
+    do: parse_naive_datetime(string, 1)
+
   def parse_naive_datetime(string) when is_binary(string),
     do: parse_naive_datetime(string, 1)
 
@@ -370,6 +376,9 @@ defmodule Calendar.ISO do
   @impl true
   def parse_utc_datetime("-" <> string) when is_binary(string),
     do: parse_utc_datetime(string, -1)
+
+  def parse_utc_datetime("+" <> string) when is_binary(string),
+    do: parse_utc_datetime(string, 1)
 
   def parse_utc_datetime(string) when is_binary(string),
     do: parse_utc_datetime(string, 1)
--- a/lib/elixir/lib/calendar/iso.ex
+++ b/lib/elixir/lib/calendar/iso.ex
@@ -30,11 +30,12 @@ defmodule Calendar.ISO do
 
   ### Features
 
-  The standard library supports a minimalistic set of possible ISO 8601 features.
+  The standard library supports a minimal set of possible ISO 8601 features.
   Specifically, the parser only supports calendar dates and the extended format.
 
   However, you can still format datetimes with `NaiveDateTime.to_iso8601/2`
   and `DateTime.to_iso8601/2` to produce either basic or extended formatted strings.
+  `Calendar.strftime/2` allows you to format datetimes however else you desire.
 
   Other optional ISO 8601 features; such as ordinal dates, week dates,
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/calendar/iso.ex
```diff
@@ -1,22 +1,52 @@
 defmodule Calendar.ISO do
   @moduledoc """
-  A calendar implementation that follows to ISO 8601.
+  The default calendar implementation, a Gregorian calendar following ISO 8601.
```
kipcole9, line 3: a proleptic Gregorian calendar following .....

lib/elixir/lib/calendar/iso.ex
```diff
@@ -1,22 +1,52 @@
...
+  This notation is not supported by Elixir.
+
+  ### Additions
+
+  ISO 8601 does not allow a whitespace instead of `T` as a separator
+  between date and times, both when parsing and formatting.
+  This is a common enough representation, Elixir allows it during parsing.
+
+  The formatting of dates in `NaiveDateTime.to_iso8601/1` and `DateTime.to_iso8601/1`
+  do produce specification-compliant sting representations using the `T` separator.
```
kipcole9, line 29: sting => string

lib/elixir/lib/calendar/iso.ex
```diff
@@ -1,22 +1,52 @@
...
+
+  ### Features
+
+  The standard library supports a minimalistic set of possible ISO 8601 features.
```
kipcole9, line 33: minimalistic => minimal ?

lib/elixir/lib/calendar/iso.ex
christhekeele: I wasn't sure if `proleptic` was super helpful for the summary description, wdyt?

lib/elixir/lib/calendar/iso.ex
christhekeele: I went with minimalistic because it tends to have "by design" associations, but I also try to keep documentation in as plain english as possible, so am a fan of the simplification

lib/elixir/lib/calendar/iso.ex
kipcole9: Yes, you are right, I'm being too pedantic give it is mentioned in the next line (which I hadn't read at the time)

josevalim:
Looks great @christhekeele! Let's break this into multiple PRs. In particular:

> Add tests and examples for leading + signs before years

Can be part of this PR.

> Simplify parsing examples to essentials and move excess into a test suite per this note

Please make it another PR.

Don't send a PR for the other two. Unless we are touching the code, there is no reason to modify it. :)

lib/elixir/lib/calendar/iso.ex
```diff
@@ -1,22 +1,52 @@
...
+  The formatting of dates in `NaiveDateTime.to_iso8601/1` and `DateTime.to_iso8601/1`
+  do produce specification-compliant string representations using the `T` separator.
+
+  ### Features
+
+  The standard library supports a minimalistic set of possible ISO 8601 features.
+  Specifically, the parser only supports calendar dates and the extended format.
+
+  However, you can still format datetimes with `NaiveDateTime.to_iso8601/2`
+  and `DateTime.to_iso8601/2` to produce either basic or extended formatted strings.
```
josevalim, line 37: We should mention that there is `strftime` for you to format in any way you want.

christhekeele:
AFAIK that build failure on master is unrelated!

lib/elixir/lib/calendar/iso.ex
```diff
@@ -1,22 +1,53 @@
...
-  24:00:00 as the zero hour of the next day, this notation is not
-  supported by Elixir.
+  The ISO 8601 specification is feature-rich, but allows applications
+  to selectively implement most parts of it. The choices Elixir makes here
+  are catalogued below.
+
+  ### Deviations
+
+  ISO 8601 allows times and datetimes to specify 24:00:00 as the zero hour of the next day.
+  This notation is not supported by Elixir.
```
christhekeele, line 20: I just noticed that the 2019 update to ISO 3601 actually removes support for this behaviour, so I don't think we need to call it out at all anymore.

lib/elixir/lib/calendar/iso.ex
```diff
@@ -1,22 +1,53 @@
...
+
+  The standard library supports a minimal set of possible ISO 8601 features.
+  Specifically, the parser only supports calendar dates and the extended format.
+
+  However, you can still format datetimes with `NaiveDateTime.to_iso8601/2`
+  and `DateTime.to_iso8601/2` to produce either basic or extended formatted strings.
+  `Calendar.strftime/2` allows you to format datetimes however else you desire.
+
+  Other optional ISO 8601 features; such as ordinal dates, week dates,
+  durations, time intervals, truncated representations, and reduced precision;
```
christhekeele, line 36: Upon review, truncated representations were also removed from the spec.

josevalim:

PR #10693, opened 2021-02-06
Support basic datetime format in Calendar.ISO parsing functions

christhekeele:
This PR adds support for an optional second parameter to `Calendar.ISO` parsing functions, allowing them to parse the `:basic` format as well as the (default) `:extended`.

### DIFF
--- a/lib/elixir/lib/calendar/iso.ex
+++ b/lib/elixir/lib/calendar/iso.ex
@@ -126,7 +126,7 @@ defmodule Calendar.ISO do
       {:ok, {-2015, 1, 23, 23, 50, 7, {0, 0}}, 0}
       iex> Calendar.ISO.parse_utc_datetime("+2015-01-23 23:50:07Z")
       {:ok, {2015, 1, 23, 23, 50, 7, {0, 0}}, 0}
-      
+
   """
 
   @behaviour Calendar
@@ -214,9 +214,9 @@ defmodule Calendar.ISO do
           y4 <= ?9 and m1 >= ?0 and m1 <= ?9 and m2 >= ?0 and m2 <= ?9 and d1 >= ?0 and d1 <= ?9 and
           d2 >= ?0 and d2 <= ?9,
         {
-          List.to_integer([y1, y2, y3, y4]),
-          List.to_integer([m1, m2]),
-          List.to_integer([d1, d2])
+          (y1 - ?0) * 1000 + (y2 - ?0) * 100 + (y3 - ?0) * 10 + (y4 - ?0),
+          (m1 - ?0) * 10 + (m2 - ?0),
+          (d1 - ?0) * 10 + (d2 - ?0)
         }
       ]
     end
@@ -229,9 +229,9 @@ defmodule Calendar.ISO do
         h1 >= ?0 and h1 <= ?9 and h2 >= ?0 and h2 <= ?9 and i1 >= ?0 and i1 <= ?9 and i2 >= ?0 and
           i2 <= ?9 and s1 >= ?0 and s1 <= ?9 and s2 >= ?0 and s2 <= ?9,
         {
-          List.to_integer([h1, h2]),
-          List.to_integer([i1, i2]),
-          List.to_integer([s1, s2])
+          (h1 - ?0) * 10 + (h2 - ?0),
+          (i1 - ?0) * 10 + (i2 - ?0),
+          (s1 - ?0) * 10 + (s2 - ?0)
         }
       ]
     end
@@ -339,7 +339,7 @@ defmodule Calendar.ISO do
 
       iex> Calendar.ISO.parse_date("2015-01-23")
       {:ok, {2015, 1, 23}}
-      
+
       iex> Calendar.ISO.parse_date("2015:01:23")
       {:error, :invalid_format}
       iex> Calendar.ISO.parse_date("2015-01-32")
--- a/lib/elixir/lib/calendar/iso.ex
+++ b/lib/elixir/lib/calendar/iso.ex
@@ -38,11 +38,13 @@ defmodule Calendar.ISO do
   ### Features
 
   The standard library supports a minimal set of possible ISO 8601 features.
-  Specifically, the parser only supports calendar dates and the extended format.
+  Specifically, the parser only supports calendar dates.
 
-  However, you can still format datetimes with `NaiveDateTime.to_iso8601/2`
-  and `DateTime.to_iso8601/2` to produce either basic or extended formatted strings.
-  `Calendar.strftime/2` allows you to format datetimes however else you desire.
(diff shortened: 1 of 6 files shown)


lib/elixir/lib/calendar/iso.ex
```diff
@@ -113,31 +205,33 @@ defmodule Calendar.ISO do
...
         y1 >= ?0 and y1 <= ?9 and y2 >= ?0 and y2 <= ?9 and y3 >= ?0 and y3 <= ?9 and y4 >= ?0 and
           y4 <= ?9 and m1 >= ?0 and m1 <= ?9 and m2 >= ?0 and m2 <= ?9 and d1 >= ?0 and d1 <= ?9 and
           d2 >= ?0 and d2 <= ?9,
         {
-          (y1 - ?0) * 1000 + (y2 - ?0) * 100 + (y3 - ?0) * 10 + (y4 - ?0),
-          (m1 - ?0) * 10 + (m2 - ?0),
-          (d1 - ?0) * 10 + (d2 - ?0)
+          List.to_integer([y1, y2, y3, y4]),
+          List.to_integer([m1, m2]),
+          List.to_integer([d1, d2])
```
josevalim, line 219: Please keep the previous implementation as it is more efficient. :)

josevalim:
Hi @christhekeele, can we please go only with the `:basic` option without `:any`? With `with` you can easily try both or even other formats. :) Thank you.

christhekeele:
@josevalim `:any` removed! 🙂

christhekeele:
Should I add something to the changelog about this within this PR, or is that handled during release?

josevalim:
Thanks for asking, changelogs are handled later on, to avoid conflicts!

wojtekmach:
I'm not sure if this was already mentioned, but this is a nice way of solving [[Proposal] Add ability to parse ISO8601 "basic" format](https://groups.google.com/g/elixir-lang-core/c/0ms6UARgQfk/m/sBt51QGOBQAJ). In that thread, it wasn't clear how the API would look like, ideally it would be option 1 & 2 which is ambiguous.

```
Date.from_iso8601(string, format)
Date.from_iso8601(string, calendar)
Date.from_iso8601(string, calendar, format)
```

Well, it's Calendar.ISO.parse_date(string, format)` which is unambiguous!

josevalim:
@wojtekmach don't celebrate too much because next step is to figure out a why to surface that upstream. :D But I think we can look at the second argument and decide accordingly. So it will be:

```
Date.from_iso8601(string, format_or_calendar)
Date.from_iso8601(string, format, calendar)
```

christhekeele:
Yeah, the `Date.from_iso8601` function signatures are a little funky.

I've always been a little confused about the use case of the signature of `Date.from_iso8601(string, calendar)`. It seems to me that `iso8601` specifies the calendar already, and basic/extended formats are specifically defined only within the context of that calendar.

AndrewDryga:
As a side topic would RFC1123 datetime format (`"D, j M Y G:i:s O"` ex. `Wed, 21 Oct 2015 19:19:53 +0000`) decoder would be accepted to Elixir? There is at least one large service (Twilio) that uses it and parsing it is a pain.

christhekeele:
@AndrewDryga I think it's worth bringing up on the core mailing list--this PR is mostly for parsing/formatting parity--but my suspicion is that that's better suited for an external library.

My way of thinking about this is that the core `Calendar` protocol is really about providing a way to work with dates and times ***correctly***, and since we need some means of conveying data in and out of that format, the default calendar supports some standard ISO interfaces to constructing them. But the parsing stuff takes a backseat and is more focused on being fast than flexible, since it must support the sigil literals for the language.

lib/elixir/lib/calendar/iso.ex
```diff
@@ -289,33 +342,64 @@ defmodule Calendar.ISO do
...
+
+      iex> Calendar.ISO.parse_date("20150123", :basic)
+      {:ok, {2015, 1, 23}}
+      iex> Calendar.ISO.parse_date("20150123", :extended)
+      {:error, :invalid_format}
+
+  """
+  @doc since: "1.12.0"
+  def parse_date("-" <> string, format) when is_binary(string),
+    do: do_parse_date(string, -1, format)
```
josevalim, line 366: Negative dates are only allowed in the extended format, so we should handle the +/- within :extended. :)

lib/elixir/lib/calendar/iso.ex
```diff
@@ -289,33 +342,64 @@ defmodule Calendar.ISO do
   """
   @doc since: "1.10.0"
   @impl true
-  def parse_date("-" <> string) when is_binary(string),
-    do: parse_date(string, -1)
+  def parse_date(string) when is_binary(string),
+    do: parse_date(string, :extended)
```
josevalim, line 346: Shall we use default arguments here instead? :)

josevalim (commented):
It looks great! I have added only two comments but they will also apply for naive dt stuff too!

lib/elixir/lib/calendar/iso.ex
christhekeele: Good point!

lib/elixir/lib/calendar/iso.ex
christhekeele: I didn't use defaults so that 1) we can have different docs/examples for each, and 2) we could have separate `@since` tags, but if neither of those are valuable I can rewrite them into defaults!

josevalim: Ah, the since tag is a good one. But I think the docs are actually worse spread apart. So maybe just a short doc saying it is equivalent to /2 with extended? (If that’s already not the case)

christhekeele: That's pretty much what I have in mind: `parse_string/1` with common `:extended` usage docs and `parse_string/2` showing `:basic`. Here's what it looks like rendered locally:

Screen Shot 2021-02-08 at 11 27 06 AM

christhekeele: (The `ISO 8601` link takes you up to the extensive moduledoc section we've added previously)
Screen Shot 2021-02-08 at 11 33 04 AM

lib/elixir/lib/calendar/iso.ex
christhekeele: I just revised the spec on this--there are two extended formats in play.

One is the general "extended" (as opposed to "basic") format--ie with separators or not. That's what the `:basic` and `:extended` flags here entail.

The other we have decided to support, to work with a wider range of dates, is the "year extension". But our support for that is independent of basic vs extended separator format, and should work with both.

TL;DR I believe this is correct as writ

josevalim: You are correct. I confused expanded with extended. :)

josevalim:

Issue #10695, opened 2021-02-09
File.cd!/2 not parallel friendly


hrubi:
### Environment

* Elixir & Erlang/OTP versions (elixir --version): 1.10.4 / 22.3.2
* Operating system: Gentoo

### Current behavior

Function `File.cd!/2` looks to me like a [bracket pattern](https://wiki.haskell.org/Bracket_pattern) which I would expect to safely execute the specified function in target directory. However, it is not. It calls `:file.set_cwd/1` and temporarily changes the global state of the Erlang file server (its CWD) for the timespan when the function is executed.

`File.cd!/2` is not parallel friendly. If some code uses relative paths and relies on the CWD of the file server to be the one the VM was started in, it can result in an error. See bug in Phoenix https://github.com/phoenixframework/phoenix/issues/4204.

Also when two `File.cd!/2` functions are called in parallel, there's a chance that the CWD won't be restored to the original one prior to those calls. See https://gist.github.com/hrubi/f1880665787d861fb84a20f491c0ee34.

### Expected behavior

Either the `File.cd!/2` is made parallel safe, or there's a warning in the documentation that it is not parallel safe.

wojtekmach:
Please send a PR to improve `File.cd!/2` docs, we already mention concurrency concerns on `File.cd/1`.

PR #10696, opened 2021-02-09
Document that File.cd!/2 affects the BEAM globally (#10695)

hrubi:


lib/elixir/lib/file.ex
```diff
@@ -1542,6 +1542,12 @@ defmodule File do
   executes the given function and then reverts back
   to the previous path regardless of whether there is an exception.
 
+  The current working directory is temporarilly set for the BEAM globally. This
```
whatyouhide, line 1545:
```suggestion
  The current working directory is temporarily set for the BEAM globally. This
```

whatyouhide (approved):

lib/elixir/lib/file.ex
whatyouhide: Ah, too late!

josevalim: It is a copy and paste of File.cd!, so better to fix both separately.

hrubi: Well, it was a copy paste, but I've added this one word.. and messed up. Here's the fix https://github.com/elixir-lang/elixir/pull/10697

PR #10699, opened 2021-02-09
add :minute, :hour, and :day to time units

btkostner:
Discussion and proposal: https://groups.google.com/g/elixir-lang-core/c/FEEwA2Pspzs/m/3_i7AKxCBQAJ

This PR adds `:day`, `:hour`, and `:minute` to `time_unit`, allowing it to be used for time conversion. Ideally this would be in the erlang/otp code, but after trying to find my way around that code base and realizing I would have to touch C code (and OS libraries), I decided this would be easier. It took me a couple tries to figure out the easiest / cleanest way to do this, and I'm pretty happy with this result.

### DIFF
--- a/lib/elixir/lib/system.ex
+++ b/lib/elixir/lib/system.ex
@@ -65,8 +65,12 @@ defmodule System do
   @typedoc """
   The time unit to be passed to functions like `monotonic_time/1` and others.
 
-  The `:second`, `:millisecond`, `:microsecond` and `:nanosecond` time
-  units controls the return value of the functions that accept a time unit.
+  The `:day`, `:hour`, `:minute`, `:second`, `:millisecond`, `:microsecond`
+  and `:nanosecond` time units controls the return value of the functions
+  that accept a time unit.
+
+  Note that `:day`, `:hour`, and `:minute` where added in Elixir v1.11.4 and
+  will throw an error in earlier versions.
 
   A time unit can also be a strictly positive integer. In this case, it
   represents the "parts per second": the time will be returned in `1 /
@@ -75,7 +79,10 @@ defmodule System do
   in 1/1000 seconds - milliseconds).
   """
   @type time_unit ::
-          :second
+          :day
+          | :hour
+          | :minute
+          | :second
           | :millisecond
           | :microsecond
           | :nanosecond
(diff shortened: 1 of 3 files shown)


lib/elixir/lib/system.ex
```diff
@@ -65,8 +65,12 @@ defmodule System do
   @typedoc """
   The time unit to be passed to functions like `monotonic_time/1` and others.
 
-  The `:second`, `:millisecond`, `:microsecond` and `:nanosecond` time
-  units controls the return value of the functions that accept a time unit.
+  The `:day`, `:hour`, `:minute`, `:second`, `:millisecond`, `:microsecond`
+  and `:nanosecond` time units controls the return value of the functions
+  that accept a time unit.
+
+  Note that `:day`, `:hour`, and `:minute` where added in Elixir v1.11.4 and
```
btkostner, line 72: I wasn't sure if there was a good way to document this with `@since`, so I just made a note here. Version might need changing depending when this gets merged.

lib/elixir/lib/system.ex
```diff
@@ -1103,9 +1112,18 @@ defmodule System do
...
+    {to_conversion_ratio, to_erlang_unit} = normalize_time_unit(to_unit)
+
+    divide_conversion_ratio(
+      :erlang.convert_time_unit(time, from_erlang_unit, to_erlang_unit) * from_conversion_ratio,
+      to_conversion_ratio
+    )
   end
 
+  defp divide_conversion_ratio(time, 1), do: time
+  defp divide_conversion_ratio(time, conversion_ratio), do: trunc(time / conversion_ratio)
```
btkostner, line 1125: Without this conditionally doing division, I was getting some test errors in `DateTime`. It appeared to be some rounding issues.

josevalim:
Hi, this was submitted [here](https://github.com/elixir-lang/elixir/pull/10185) and then [reverted here](https://github.com/elixir-lang/elixir/pull/10199#issuecomment-659525151). So we are probably not going ahead with this at this level. Thanks!

lib/elixir/test/elixir/calendar/datetime_test.exs
```diff
@@ -941,6 +941,25 @@ defmodule DateTimeTest do
                  microsecond: {123_456, 6}
                }
     end
+
+    test "add/2 with non erlang time_units" do
```
btkostner, line 945: Added this test to ensure the code ends up with the actual intended result down the code path

btkostner:
Whoops. Missed that PR in my GH search. Thanks for the quick reply!

PR #10700, opened 2021-02-11
Make Enum.any?/1 and Enum.all?/1 faster

sabiwara:
Hi! This might be a controversial PR, I'm opening it for discussion.

I found out that the use of the identity function as a default param (`fun \\ fn x -> x end`) for functions that are iterating over a collection, especially in the `Enum` module, can be adding quite some overhead in some cases.

For most functions which are building some new data structures, like `Enum.group_by/2`, this overhead is fairly small.
But for functions that are mostly reading, the speedup obtained by declaring different arity functions separately is pretty huge:
- [a dedicated Enum.all?/1 is about 3x faster than the current one](https://github.com/sabiwara/elixir_benches/blob/main/bench/fast_enum_all.results.txt) for lists
- [:lists.max/1 is about 4.5x faster than Enum.max/1](https://github.com/sabiwara/elixir_benches/blob/main/bench/enum_max.results.txt) for lists, hinting that implementing `Enum.max/1` separately could make it much faster
- [Enum.uniq/1 speedup](https://github.com/sabiwara/elixir_benches/blob/main/bench/fast_enum_uniq.results.txt) depends on whether there are many duplicates or not. If yes, the speedup can be quite substantive as well since this is mostly read-only

There are however cons to this approach:
- Code duplication
- Makes it a different function in the doc, which might be less clear?

Would there be any smart alternative way to benefit from the speedup without these drawbacks, using meta programming or optimizing the `fun \\ fn x -> x end` pattern at the compiler level?

I decided to open a PR for `Enum.any?` and `Enum.all?`. I think these might actually be clearer as separate functions, since the documentation was already split in two blocks.
What do you think?

### DIFF
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -280,6 +280,36 @@ defmodule Enum do
   end
 
   @doc """
+  Returns `true` if  all elements in `enumerable` are truthy.
+
+  When an element has a falsy value (`false` or `nil`) iteration stops immediately
+  and `false` is returned. In all other cases `true` is returned.
+
+  ## Examples
+
+      iex> Enum.all?([1, 2, 3])
+      true
+
+      iex> Enum.all?([1, nil, 3])
+      false
+
+      iex> Enum.all?([])
+      true
+
+  """
+  @spec all?(t) :: boolean
+  def all?(enumerable) when is_list(enumerable) do
+    all_list(enumerable)
+  end
+
+  def all?(enumerable) do
+    Enumerable.reduce(enumerable, {:cont, true}, fn entry, _ ->
+      if entry, do: {:cont, true}, else: {:halt, false}
+    end)
+    |> elem(1)
+  end
+
+  @doc """
   Returns `true` if `fun.(element)` is truthy for all elements in `enumerable`.
 
   Iterates over `enumerable` and invokes `fun` on each element. If `fun` ever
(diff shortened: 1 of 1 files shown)


josevalim:
Hi @sabiwara, great job!

I think this is something we should definitely consider. It is our job to be as fast as we can so you don't have to micro-optimize your code.

However, I do wonder what is the impact of those changes under the JIT. Can you benchmark with Erlang/OTP master? Because if the JIT reduces the difference considerably, then we can probably skip it. Thank you!

sabiwara:
> Can you benchmark with Erlang/OTP master?

Sure, will do! :+1:

sabiwara:
Interestingly, it seems that the separate version benefits from the JIT even more: https://github.com/sabiwara/elixir_benches/compare/jit?expand=1
(unlike the previous ones, these benchmarks are done within a docker image, which might explain why the figures are overall smaller)

josevalim:

sabiwara:
Thanks for merging!
I'll try to open a similar PR for `Enum.max/1` / `Enum.min/1` in the same fashion, WDYT?

Just to develop a bit more what I had in mind as an example possible compiler optimization (probably overkill, just mentioning):
- identify `fn x -> x end` or `& &1` declarations and make them return always a given function, e.g. `&Kernel.identity/1`
- compare `fun` against `&Kernel.identity/1` to decide if should invoke the optimized version or not

josevalim:
>  I'll try to open a similar PR for Enum.max/1 / Enum.min/1 in the same fashion, WDYT?

Yes, please!

PR #10701, opened 2021-02-11
Speedup Enum.min/1 and Enum.max/1 for lists

sabiwara:
Hi!
This change should yield a ~4.5x speedup for lists according to [this benchmark](https://github.com/sabiwara/elixir_benches/blob/main/bench/enum_max.results.txt).

It is a bit tricky to add clauses to these functions given how the double defaults are set, but I think this change is quite small and optimizes for the more important use cases (non-empty lists, with or without `empty_fallback`).
If we want to optimize other enumerables, it should be possible as well but it would be a bigger change with much duplication (and probably a less noticeable impact).

### DIFF
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -1666,6 +1666,14 @@ defmodule Enum do
   end
 
   @doc false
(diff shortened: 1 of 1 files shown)


josevalim:

PR #10703, opened 2021-02-12
Support throw in DocTest

chulkilee:
## What it does

Currently DocTest does not support `throw` case. It detects the string format in the expectation, but it try to use `Exception.throw` as exception module, not treating it as throw.

With this PR, it supports `throw` correctly. For example, the following can be used in the doctest.

```elixir
iex> (fn () -> throw :bad end).()
** (throw) :bad
```

Probably in most cases throw is only for local return inside a function call (and its private function calls), not across public function calls... but since `throw` is a feature actually, I guess it would be a good addition.

## To do / Questions

- I added a separate function body for `{:error, :throw, message}` instead of making changes in the existing function for `{:error, exception, message}`
- I added test cases but 1) I couldn't figure out how to fix tests (e.g. it needs updating line numbers somewhere) and 2) Not sure I added them to the right modules.

Feel free to edit the PR.

### DIFF
--- a/lib/ex_unit/lib/ex_unit/doc_test.ex
+++ b/lib/ex_unit/lib/ex_unit/doc_test.ex
@@ -374,6 +374,44 @@ defmodule ExUnit.DocTest do
     end
   end
 
+  defp test_case_content(expr, {:error, :throw, message}, location, stack, formatted) do
+    doctest = "\n" <> formatted <> "\n** (throw) #{inspect(message)}"
+    expr_ast = string_to_quoted(location, stack, expr, doctest)
+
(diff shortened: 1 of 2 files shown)


josevalim:
Thanks @chulkilee! Unless this is actually needed for some reason, I would prefer to not add it. Exactly because doctests are about docs and you should never really document that something throws, as it is meant to be handled internally.

chulkilee:
That was my guess :) - probably we can update [ExUnit.DocTest](https://hexdocs.pm/ex_unit/ExUnit.DocTest.html). Closing this PR

PR #10707, opened 2021-02-12
Fix Git warning

eksperimental:
It was partially solved in 74855f052d6a0d8a015f1956426b0de27e3572c8

PR #10708, opened 2021-02-13
Remove usage of master and master/slave terminology

eksperimental:
Implements suggestions in https://github.com/elixir-lang/elixir/pull/10707#issuecomment-778421314


eksperimental:
It is up to another PR the final push if we want to rename the default branch "master" to "main".
I've looked it up and there seems to be an "official" guide for doing this, https://github.com/github/renaming

josevalim:

PR #10709, opened 2021-02-13
Preserve file mode when copying non-beam ebin files

fhunleth:
If a non-beam file in the ebin directory had execute permissions set,
those were lost when making a release. This switches the copy function
to `File.cp!/3` to preserve permissions.

Specifically, this affects the use of zigler with Nerves. Zigler
generates NIF shared libaries and stores them in the `ebin` directory.
The shared libraries had their execute mode bit set, but lost it on the
copy to the release. This affected the Nerves post-processing of all
firmware executables since Nerves uses the execute bit to identify
files that should be processed.

This also switches the copy call to the raising version like is done
when writing `.beam` files for consistency.

### DIFF
--- a/lib/mix/lib/mix/release.ex
+++ b/lib/mix/lib/mix/release.ex
@@ -780,7 +780,10 @@ defmodule Mix.Release do
              {:ok, binary} <- strip_beam(File.read!(source_file), strip_options) do
           File.write!(target_file, binary)
         else
-          _ -> File.copy(source_file, target_file)
+          _ ->
+            # Use File.cp!/3 to preserve file mode for any executables stored
+            # in the ebin directory.
+            File.cp!(source_file, target_file)
         end
       end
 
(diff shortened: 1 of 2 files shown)


josevalim:
Beautiful! Can we add a test or at least a comment explaining why the use of cp? :)

fhunleth:
I added a comment since that was easy. 

As for a test, the current release tests re-use the eex application. It has a non-beam file so the copy is being tested. It doesn't have any executable files and I'm not seeing an easy way to add them, but maybe I'm not thinking creatively enough.

fhunleth:
Hang on... maybe this isn't that hard..

fhunleth:
@josevalim There's a unit test now.

josevalim:
Ah, good call on the unit test. :)

josevalim:

PR #10716, opened 2021-02-15
Simplify guards by using ranges in character matching

eksperimental:


josevalim:
The reason why we didn't use ranges is because this is performance sensitive code and the range does an extra is_integer check. I expect the Erlang compiler to be able to optimize those at this point - someone needs confirm it - but given the PR does not really change the behaviour, we can keep things as is as well. Thank you. :)

eksperimental:
True! I forgot about the is_integer check

michalmuskala:
In cases where the value is extracted from binary, the compiler should easily optimise this check away, in case of traversing lists, it's rather unlikely.

eksperimental:
> In cases where the value is extracted from a binary, the compiler should easily optimise this check away, in case of traversing lists, it's rather unlikely to be optimised away.

Thank you for the clarification. So only the changes in the String module would be optimized.

eksperimental:
> > In cases where the value is extracted from a binary, the compiler should easily optimise this check away, in case of traversing lists, it's rather unlikely to be optimised away.
> 
> Thank you for the clarification. So only the changes in the String module would be optimized.
actually, in Calendar.ISO  as well.

PR #10719, opened 2021-02-16
Fix translating erl compiler errors to diagnostics on OTP 24

wojtekmach:
The compiler previously emitted just `line` and now emits `{line, column}`.

The diagnostic struct accepts as position either:

    nil
    line
    {start_line, start_col, end_line, end_col}

so we could have used the `column` to create the 4-tuple but we don't have enough information to do that correctly, we don't know where the column or line ends.


wojtekmach:
I guess we could emit `{line, column - 1, line, column - 1}`? (columns are supposed to be 0-indexed)

josevalim:
I think this is definitely good enough for now, we can revisit in the future. :)

josevalim:

PR #10723, opened 2021-02-17
Improve syntax docs

eksperimental:


lib/elixir/pages/unicode-syntax.md
```diff
@@ -34,19 +34,21 @@ Elixir does not allow the use of ZWJ or ZWNJ in identifiers and therefore does n
...
 ### Variables
 
 Variables in Elixir follow the identifier rule above with the following modifications:
 
-  * `<Start>` includes the code point `_` (005F)
-  * `<Start>` must not include Lu (letter uppercase) and Lt (letter titlecase) characters
+  * `<Start>` additionally includes the code point `_` (005F)
+  * `<Start>` additionally excludes Lu (letter uppercase) and Lt (letter titlecase) characters
+
+In set notation: `[\u{005F}\p{Ll}\p{Lm}\p{Lo}\p{Nl}\p{Other_ID_Start}-\p{Pattern_Syntax}-\p{Pattern_White_Space}]`.
```
eksperimental, line 51: No sure if we should quote it as the other ones above

lib/elixir/pages/syntax-reference.md
```diff
@@ -21,7 +21,7 @@ Integers (`1234`) and floats (`123.4`) in Elixir are represented as a sequence o
 
 ### Atoms
 
-Unquoted atoms start with a colon (`:`) which must be immediately followed by an underscore or a Unicode letter. The atom may continue using a sequence of Unicode letters, numbers, underscores, and `@`. Atoms may end in `!` or `?`. See [Unicode Syntax](unicode-syntax.md) for a formal specification. Valid unquoted atoms are: `:ok`, `:ISO8601`, and `:integer?`.
+Unquoted atoms start with a colon (`:`) which must be immediately followed by a Unicode letter or an underscore. The atom may continue using a sequence of Unicode letters, numbers, underscores, and `@`. Atoms may end in `!` or `?`. See [Unicode Syntax](unicode-syntax.md) for a formal specification. Valid unquoted atoms are: `:ok`, `:ISO8601`, and `:integer?`.
```
eksperimental, line 24: I changed the order here as it more common an atom starting with a letter than with an underscore

lib/elixir/pages/unicode-syntax.md
```diff
@@ -34,19 +34,21 @@ Elixir does not allow the use of ZWJ or ZWNJ in identifiers and therefore does n
 
 ### Atoms
 
-Unicode atoms in Elixir follow the identifier rule above with the following modifications:
+Unquoted unicode identifier-based atoms in Elixir follow the identifier rule above with the following modifications:
```
josevalim, line 37: I think this is more precise but without specifying that it is, I don't think it helps the user understand it. So I would propose to revert it and clarify it below.

josevalim:
```suggestion
Unicode atoms in Elixir follow the identifier rule above with the following modifications:
```

lib/elixir/pages/unicode-syntax.md
```diff
@@ -34,19 +34,21 @@ Elixir does not allow the use of ZWJ or ZWNJ in identifiers and therefore does n
...
 
-  * `<Start>` includes the code point `_` (005F)
-  * `<Continue>` includes the code point `@` (0040)
+  * `<Start>` additionally includes the code point `_` (005F)
+  * `<Continue>` additionally includes the code point `@` (0040)
 
 > Note that all Elixir operators are also valid atoms. Therefore `:+`, `:@`, `:|>`, and others are all valid atoms. The full description of valid atoms is available in the Syntax Reference, this document covers only the rules for identifier-based atoms.
```
josevalim, line 42:
```suggestion
> Note that all Elixir operators are also valid atoms. Therefore `:+`, `:@`, `:|>`, and others are all valid atoms. Atoms can also be quoted, which allows any character, such as `:"hello world"`. The full description of valid atoms is available in the Syntax Reference, this document covers only the rules for identifier-based atoms.
```

josevalim:

lib/elixir/pages/unicode-syntax.md
```diff
@@ -34,12 +34,12 @@ Elixir does not allow the use of ZWJ or ZWNJ in identifiers and therefore does n
 
 ### Atoms
 
-Unquoted unicode identifier-based atoms in Elixir follow the identifier rule above with the following modifications:
+Unicode atoms in Elixir follow the identifier rule above with the following modifications:
```
eksperimental, line 37: I like this clarification. But my intention with that change ( I was also not bought by the idea of over-specifying it), was to not to rely on the note for mentioning that is this ruly only apply for identifier-based atoms.
I guess a good compromise could be just to mention "Unquoted atoms in Elixir".... at the opening of the parapgrah. WDTY @josevalim?

josevalim: unquote is ambiguous in Elixir (it can be about macros) - so unless we clearly specify what unquoted means, then a reference to it most likely won't help - because it probably makes sense only if you know what it means.

eksperimental: You are right. I have submitted a PR addressing this.

Issue #10722, opened 2021-02-16
Issues with question and exclamation marks in atoms


eksperimental:
I have found some issues with messages that were not fully complying with the syntax.

1. This message should mention that atoms may end with ! or ?

```elixir
iex(21)> :"a!"
warning: found quoted atom "a!" but the quotes are not required. Atoms made exclusively of ASCII letters, numbers, and underscore do not require quotes
:a!

iex(22)> :"a?"
warning: found quoted atom "a?" but the quotes are not required. Atoms made exclusively of ASCII letters, numbers, and 
underscore do not require quotes
:a?
```

2. This one does not emit a warning in either case, but the atom is built. It should emit a warning when not using quotes.
```elixir
iex(25)> :"!"
:!
iex(26)> :!  
:!
```

3. This one fails to build the atom when quotes are not used.
```elixir
iex(29)> :"?"
:"?"
iex(30)> :?  
** (SyntaxError) iex:30:1: unexpected token: ":" (column 1, code point U+003A)
```

I can provide a fix for no. 1 and 2 once agreed on

eksperimental:
I noticed that they were operators that and came to fix no2, and 3, to see your comment :rofl: 

Still, the warning message in no 1 needs to be fixed, since it is contradictory.
Maybe add ", or that end in ! or ? do not need to be quoted."

josevalim:
I see, good call!

eksperimental:
I noticed that the whole message had to be revamped to be strict to what the syntax expresses.

how about this warning?

> warning: found quoted atom "a!" but the quotes are not required. Atoms starting with a Unicode letter or an underscore, which may be followed by a sequence of Unicode letters, numbers, underscores, and @, and ending in ! or ? do not require quotes.

josevalim:
This warning is to avoid people writing ["foo": bar] and thinking "foo" will be a string. It is actually ok if you want to write :"josé" to make it clearer that it is in Unicode - we just want to catch the basic ASCII cases.

eksperimental:
then this?

> Atoms made exclusively of ASCII letters, numbers, and underscores, wich may end with a ! or a ?, do not require quotes

josevalim:
I would rather no longer warn if it ends with `?` and `!`. WDYT?

josevalim:
Although that may make it too much of a special case, it is probably best to just follow the identifier rule.

eksperimental:
[...]

PR #10724, opened 2021-02-17
Fix warning message mentioning quoted atoms may end with ! or ?

eksperimental:
Closes https://github.com/elixir-lang/elixir/issues/10722

### DIFF
--- a/lib/elixir/src/elixir_tokenizer.erl
+++ b/lib/elixir/src/elixir_tokenizer.erl
@@ -438,7 +438,7 @@ tokenize([$:, H | T] = Original, Line, Column, Scope, Tokens) when ?is_quote(H) 
         true ->
           WarnMsg = io_lib:format(
             "found quoted atom \"~ts\" but the quotes are not required. "
-            "Atoms made exclusively of ASCII letters, numbers, and underscores "
+            "Atoms made exclusively of ASCII letters, numbers, and underscores which may optionally end with ! or ?"
             "do not require quotes",
             [hd(Parts)]
           ),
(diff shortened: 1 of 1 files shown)


lib/elixir/src/elixir_tokenizer.erl
```diff
@@ -438,7 +438,7 @@ tokenize([$:, H | T] = Original, Line, Column, Scope, Tokens) when ?is_quote(H)
         true ->
           WarnMsg = io_lib:format(
             "found quoted atom \"~ts\" but the quotes are not required. "
-            "Atoms made exclusively of ASCII letters, numbers, and underscores "
+            "Atoms made exclusively of ASCII letters, numbers, and underscores which may optionally end with ! or ?"
```
josevalim, line 441:
```suggestion
            "Atoms made exclusively of ASCII letters, numbers, and underscores, which may optionally end with ! or ?, "
```

?

josevalim: Or:

```suggestion
            "Atoms made exclusively of ASCII letters, numbers, underscores, and optionally end with ! or ? "
```

michallepicki, line 441: I like the second option more but I think it needs `ing` in `ending` to form a valid sentence, like this:
```suggestion
            "Atoms made exclusively of ASCII letters, numbers, and underscores, and optionally ending with ! or ? "
```

eksperimental: Correct. I was about to suggest the same thing.

eksperimental:
I guess this is ready to be merged.

josevalim:

PR #10726, opened 2021-02-17
Remove unnecessary words from links, such as: in, the

eksperimental:


josevalim:
I think it is fine as is, unless there is a guide that clearly instructs otherwise. Thank you! :)

PR #10730, opened 2021-02-17
Add RELEASE_PROG to releases, closes #10728

### LINKED ISSUE #10728: Environment variables to set in shell scripts
### Current behavior

`PROGNAME` is an environment variable set by the `erl` shell script and the `relx` start scripts. In `iex` this variable isn't set so it continues to be `erl` when read:

```
> :os.getenv('PROGNAME')
'erl'
```

### Expected behavior

I think this should be either `iex` or `elixir` instead.

In addition it would be useful if the start script generated by `mix release` would set `PROGNAME` to the name of that script -- this is currently done by relx.

Two other useful environment variables would be `REL_NAME` and `REL_VSN`. Currently relx's script only exports `REL_NAME` but I'll be making a PR to also export `REL_VSN`.

All three of these variables are to be used by OpenTelemetry resource detection. In cases that the `release_handler` module isn't available or the release server isn't started the resource detector falls back on checking for those environment variables.

The release name and version are used for the `service.name` and `service.version` attributes while `PROGNAME` is used for `process.executable.name`.

josevalim:


tsloughter (approved):

josevalim:

PR #10734, opened 2021-02-17
Shorten is_struct/2 Macro Expansion for Atom Literals

J3RN:
The macro expansion for `is_struct/2` includes a `case` that checks for whether the second argument (`name`) is an atom, and raises `ArgumentError` if not. This is sometimes, maybe even usually, known at compile time, and the second clause of the generated case is simply dead code. Dialyzer warns about this.

```
The pattern
:variable_

can never match, because previous clauses completely cover the type
MyApp.MyModule.
```

The change is simply to check to see whether "name" is known to be an atom at compile-time, and if so not to generate the case at all but simply perform the map `__struct__` key check.

### DIFF
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -2373,17 +2373,22 @@ defmodule Kernel do
   defmacro is_struct(term, name) do
     case __CALLER__.context do
       nil ->
-        quote do
-          case unquote(name) do
-            name when is_atom(name) ->
-              case unquote(term) do
-                %{__struct__: ^name} -> true
-                _ -> false
-              end
+        case Macro.expand(name, __CALLER__) do
+          name when is_atom(name) ->
+            quote do
+              match?(%{__struct__: unquote(name)}, unquote(term))
+            end
 
-            _ ->
-              raise ArgumentError
-          end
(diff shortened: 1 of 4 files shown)


josevalim:
Hi @J3RN! In this case, it is better to mark the clauses as generated, so Dialyzer stops warning. That's because there are other ways this can be an atom and dialyzer will know it is an atom without the compiler knowing - so using generated is the most guaranteed way to do so.

josevalim:
There is also a dialyzer suite you can use to add a test related to this. :) Let me know if you have any questions and thanks!

J3RN:
Hey @josevalim, thanks for your feedback! You're absolutely right about Dialyzer being able to infer an atom past just the literal being present. I've made the corresponding change, and added Dialyzer tests.

One additional thought: Do you think there's merit in the macro expansion eliminating the "dead code" second `case` clause when it detects that a atom literal is being used as the second argument? Theoretically there could be a very small (perhaps negligible) runtime optimization by removing the check from the expansion, but it comes at the cost of the macro being slightly more complex.

michalmuskala:
This check should be easily eliminated by further passes of the compiler

josevalim:

PR #10736, opened 2021-02-18
Migrate Erlang module URLs to Erlang module links

eksperimental:


eksperimental:
The only caveat is the way IEx display these links.

Using URL

```
If the :debug option is present, the corresponding function in the :sys module
(http://www.erlang.org/doc/man/sys.html) will be invoked.
```

Using module link
```
If the :debug option is present, the corresponding function in the :sys module
(:sys) will be invoked.
```

Maybe there is something on the IEx side that we can do to improve the user experience...

lib/elixir/lib/agent.ex
```diff
@@ -241,7 +241,7 @@ defmodule Agent do
   and the start function will return `{:error, :timeout}`.
 
   If the `:debug` option is present, the corresponding function in the
-  [`:sys` module](http://www.erlang.org/doc/man/sys.html) will be invoked.
+  [`:sys` module](`:sys`) will be invoked.
```
wojtekmach, line 244: +1 for this and similar changes.

I think it was useful to see the erlang.org URL before, but now that `iex> h :sys` works, there's imho less of a need for that.

In terms of rendering `` [`:sys` module](`:sys`) `` link in IEx, I think we should show it as just `` `:sys` module ``, stripping the "url" part. This is something we should do for another reason and that is OTP docs chunks contain their own internal links like `` stdlib:sys#gen_server ``:

image

we should either transform the links to something more clear or strip them, I think it's totally reasonable to just strip them.

Even links to named anchors like `` [whatever](#foo) `` are arguably not useful to show, so I'd strip the URL part too.

Putting this together, I'd consider only displaying the URL if it's a http/https URL.

WDYT?

josevalim:

PR #10741, opened 2021-02-20
Optimize Enum.dedup/1

sabiwara:
HI!

Similarly to previous PRs on `Enum.max/1` and `Enum.any?/1`, I noticed that `Enum.dedup/1` could also be implemented in a much more efficient way, since `Enum.dedup_by/2` needs to:
- call identities `fun` at every step
- use a tuple as `acc` instead of just a list

My [benchmarks](https://github.com/sabiwara/elixir_benches/blob/main/bench/fast_enum_dedup.results.txt) suggest a significant performance boost disregarding (either if there are duplicates or not), and the boost seems to be even more important [with the JIT](https://github.com/sabiwara/elixir_benches/blob/2f4e6bab25ee4a9ed97b0af108d65190e6ebcaef/bench/fast_enum_dedup.results.txt).

### DIFF
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -780,8 +780,18 @@ defmodule Enum do
 
   """
   @spec dedup(t) :: list
+  def dedup(enumerable) when is_list(enumerable) do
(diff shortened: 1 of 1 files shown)


josevalim:

PR #10744, opened 2021-02-21
Optimize Enum.scan/2 and Enum.scan/3 for lists

sabiwara:
Hi!
This PR optimizes `Enum.scan/2` and `Enum.scan/3` for lists.
The speedup seems to be around 1.7x~2x according to [this benchmark](https://github.com/sabiwara/elixir_benches/blob/main/bench/fast_enum_scan.results.txt), but this drops down to ~1.6x [with the JIT](https://github.com/sabiwara/elixir_benches/blob/jit/bench/fast_enum_scan.results.txt) (which is still a nice improvement though I think).

### DIFF
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -2494,6 +2494,15 @@ defmodule Enum do
 
   """
   @spec scan(t, (element, any -> any)) :: list
+  def scan(enumerable, fun)
+
+  def scan([], _fun), do: []
+
+  def scan([elem | rest], fun) do
+    scanned = scan_list(rest, elem, fun)
+    [elem | scanned]
+  end
+
   def scan(enumerable, fun) do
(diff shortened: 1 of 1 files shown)


josevalim:

sabiwara:
Hi @josevalim! Thank you for the quick merges :blue_heart: 
If you want I can submit a couple more PRs in a similar fashion, however I expect it to be with smaller speedups at this point (like ~1.2x). Would this still be interesting?

Example:
```
Name                            ips        average  deviation         median         99th %
FastEnum.with_index/2        1.04 M        0.97 μs  ±2077.76%        0.81 μs        1.42 μs
Enum.with_index/2            0.84 M        1.20 μs  ±1821.88%        1.02 μs        1.94 μs

Comparison: 
FastEnum.with_index/2        1.04 M
Enum.with_index/2            0.84 M - 1.24x slower +0.23 μs
```

josevalim:
If it is within 20% i think we can probably skip it. :)

PR #10745, opened 2021-02-22
Optimize Enum.intersperse/2

sabiwara:
Hi! `Enum.map_intersperse/3` had an optimized version for lists, but `Enum.interspese/2` didn't.
This optimization can speedup by a factor of ~2x according to [this benchmark](https://github.com/sabiwara/elixir_benches/blob/jit/bench/fast_enum_intersperse.results.txt) (JIT).
I also renamed the `element` param to `separator` to be consistent with `map_intersperse` (which is also clearer IMO).


josevalim:

PR #10746, opened 2021-02-22
Optimize Enum.join/3 for lists

sabiwara:
Hi again. I tried to see if `Enum.join` could be improved as well for lists, and it seems that this version can be ~40% faster [with the JIT](https://github.com/sabiwara/elixir_benches/blob/jit/bench/fast_enum_join.results.txt), ~30% [with OTP23](https://github.com/sabiwara/elixir_benches/blob/main/bench/fast_enum_join.results.txt).
Note: I first tried [with body recursion](https://github.com/sabiwara/elixir_benches/commit/ed9c364a5c41a8beed8190898d3a3602c8ab8375), but the result was not so convincing: TCO seems to make a noticeable difference, especially the JIT seems to like it.

### DIFF
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -1508,14 +1508,14 @@ defmodule Enum do
   @spec join(t, String.t()) :: String.t()
   def join(enumerable, joiner \\ "")
 
-  def join(enumerable, joiner) when is_list(enumerable) and is_binary(joiner) do
-    join_list(enumerable, joiner)
-  end
-
   def join(enumerable, "") do
     enumerable
     |> map(&entry_to_string(&1))
     |> IO.iodata_to_binary()
+  end
+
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/enum.ex
```diff
@@ -1508,6 +1508,10 @@ defmodule Enum do
   @spec join(t, String.t()) :: String.t()
   def join(enumerable, joiner \\ "")
 
+  def join(enumerable, joiner) when is_list(enumerable) and is_binary(joiner) do
+    join_list(enumerable, joiner)
+  end
```
josevalim, line 1513: Perhaps move this clause after `join(enumerable, "")` because we likely need to optimize only the one where the string is not empty!

sabiwara: Got it! Pushed :slightly_smiling_face:

josevalim:

PR #10747, opened 2021-02-22
Refine docstring for String.duplicate

bessbd:
Duplicate `n` times might sound unusual to some. This commit changes the docstring for `String.duplicate` to use repeat.

See also https://lodash.com/docs/4.17.15#repeat .

It might be worth renaming / aliasing the method in the future.

### DIFF
--- a/lib/elixir/lib/string.ex
+++ b/lib/elixir/lib/string.ex
@@ -1595,7 +1595,7 @@ defmodule String do
   @compile {:inline, duplicate: 2}
 
   @doc """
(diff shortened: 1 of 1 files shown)


bessbd:
Looking at https://erlang.org/doc/man/lists.html#duplicate-2 and https://erlang.org/doc/man/string.html#copies-2 it may not be as straightforward as I thought first.

josevalim:

bessbd:
Thank you for the review and merge, @josevalim !

PR #10750, opened 2021-02-24
Add Link to erlang doc and update http to https

tedhexaflow:


lib/elixir/lib/file.ex
```diff
@@ -1352,7 +1352,7 @@ defmodule File do
 
     * `:delayed_write`, `:raw`, `:ram`, `:read_ahead`, `:sync`, `{:encoding, ...}`,
       `{:read_ahead, pos_integer}`, `{:delayed_write, non_neg_integer, non_neg_integer}` -
-      for more information about these options see `:file.open/2`.
+      for more information about these options see [`:file.open/2`](https://erlang.org/doc/man/file.html#open-2).
```
josevalim, line 1355: You don't need to link functions, those are automatically linked. Only modules aren't because they are ambiguous.

```suggestion
      for more information about these options see `:file.open/2`.
```

lib/mix/lib/mix/tasks/profile.fprof.ex
```diff
@@ -91,7 +91,7 @@ defmodule Mix.Tasks.Profile.Fprof do
   ## Caveats
 
   You should be aware that the code being profiled is running in an anonymous
-  function which is invoked by [`:fprof` module](http://wwww.erlang.org/doc/man/fprof.html).
+  function which is invoked by [`:fprof` module](https://www.erlang.org/doc/man/fprof.html).
```
josevalim, line 94:
```suggestion
  function which is invoked by [`:fprof` module](https://erlang.org/doc/man/fprof.html).
```

lib/mix/lib/mix/tasks/profile.eprof.ex
```diff
@@ -82,7 +82,7 @@ defmodule Mix.Tasks.Profile.Eprof do
   ## Caveats
 
   You should be aware that the code being profiled is running in an anonymous
-  function which is invoked by [`:eprof` module](http://wwww.erlang.org/doc/man/eprof.html).
+  function which is invoked by [`:eprof` module](https://www.erlang.org/doc/man/eprof.html).
```
josevalim, line 85:
```suggestion
  function which is invoked by [`:eprof` module](https://erlang.org/doc/man/eprof.html).
```

josevalim:

lib/elixir/lib/file.ex
tedhexaflow: Hi @josevalim,

While I'm reading a doc on https://hexdocs.pm/elixir/File.html#open/2, I came across this and notice I cannot click to go to erlang doc. Hence, I've added a link.

I'm not sure it is intedned. 


image

josevalim: Thanks, I will investigate!

tedhexaflow: Hi @josevalim,

Is it because of the recent changes?

I run make docs locally on master branch and it seems the link is working.

The one that I read on https://hexdocs.pm/elixir/File.html#open/2 is v.1.11.3.

josevalim: It may be related to the Erlang version being used to generate the docs!

chulkilee:
If we use the latest ex_doc, then we can use ex_doc link.

https://hexdocs.pm/ex_doc/readme.html

> This also allows to refer to OTP modules, e.g.: <code>\[\`:array\`\](\`:array\`)</code>.

See https://github.com/elixir-lang/ex_doc/pull/1335 which fixes the url.

tedhexaflow:
> If we use the latest ex_doc, then we can use ex_doc link.
> 
> https://hexdocs.pm/ex_doc/readme.html
> 
> > This also allows to refer to OTP modules, e.g.: `` [`:array`](`:array`) ``.
> 
> See [elixir-lang/ex_doc#1335](https://github.com/elixir-lang/ex_doc/pull/1335) which fixes the url.

Thanks @chulkilee 

That's why it does not work on https://hexdocs.pm/elixir/File.html#open/2 and it does work on local machine with latest ex_doc master branch.

Is it possible to regenreate the current v.1.11.3 with latest ex_doc on hexdocs.pm?
or we will have to wait till next elixir release.

Issue #10433, opened 2020-10-16
Support Erlang/OTP 24 new features


josevalim:
To be implemented when Erlang 24 is released:

  * [x] Use `os:env/0` in `System.get_env/0` (requires a `function_exported?` check)
  * [x] The OTP breaking change to stacktraces requires rebar3 to ship multiple compiled binaries, we need to align with this on `mix local.rebar`
  * [x] Support EEP 54
  * [x] Support float-16
  * [x] Use maps on erl_eval
  * [x] Emit error_info from Elixir exceptions (https://github.com/erlang/otp/pull/4764)
  * [x] `maps:from_keys/2`

To be implemented when we require Erlang 24 exclusively (from v1.15):

  * [x] ~Use from_abstr+no_lint (see https://github.com/erlang/otp/pull/3019)~ (not possible, overrides debug_info)
  * [x] Use sets (v2) in `MapSet`
  * [x] Generate `:optional_applications` key in `.app`
  * [x] Add `nonempty_binary` and `nonempty_bitstring` to typespecs
  * [x] Add `Process.alias/1` and `Process.monitor/2` (EEP 53)
  * [x] ~See if we can integrate `:gen_server.wait_response/check_response/receive_response` as tasks~ (not possible, it requires custom gen replies)
  * [x] Support shutdown of zero (#11836)

hauleth:
I think that thanks to @DianaOlympos we can also use float formatters from Erlang in OTP 24, the implementations should now be exactly the same in both languages.

josevalim:
@hauleth I believe there is nothing to be done on that front since we already use the relevant APIs, i.e. `io_lib_format:fwrite_g`. :)

DianaOlympos:
Yeah even if it is a private API that is undocumented .... :smile:

hauleth:
@DianaOlympos and I was doing my best to avoid non-public API in Logger… Ehhhh…

Jokes aside - additional point can be added - translating `:console` logger to be Erlang handler instead of Elixir backend (also tracked in #9465).

DianaOlympos:
@hauleth technically you can use the `io:fwrite` API with the correct options, but people just called `:io_lib_format:fwrite_g/1` instead. Which will be fixed soon as i work on making it a BIF anyway.

ThePuzzlemaker:
Erlang/OTP 24 [is now official released.](https://www.erlang.org/news/148)

lukaszsamson:
[...]

PR #10753, opened 2021-02-25
Use :os.env/0 in System.get_env if available

mtarnovan:
Ref https://github.com/elixir-lang/elixir/issues/10433

`:os.env/0` is available in [OTP 24](https://erlang.org/download/OTP-24.0-rc1.README) as OTP-16793

Note that the CI matrix doesn't include OTP 24 yet.

### DIFF
--- a/lib/elixir/lib/system.ex
+++ b/lib/elixir/lib/system.ex
@@ -604,12 +604,21 @@ defmodule System do
   Variable names and their values are strings.
   """
   @spec get_env() :: %{optional(String.t()) => String.t()}
-  def get_env do
-    Enum.into(:os.getenv(), %{}, fn var ->
-      var = IO.chardata_to_string(var)
-      [k, v] = String.split(var, "=", parts: 2)
-      {k, v}
-    end)
+
+  if function_exported?(:os, :env, 0) do
+    def get_env do
+      Map.new(:os.env(), fn {k, v} ->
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/system.ex
```diff
@@ -604,12 +604,21 @@ defmodule System do
...
   """
   @spec get_env() :: %{optional(String.t()) => String.t()}
-  def get_env do
-    Enum.into(:os.getenv(), %{}, fn var ->
-      var = IO.chardata_to_string(var)
-      [k, v] = String.split(var, "=", parts: 2)
-      {k, v}
-    end)
+
+  if function_exported?(:os, :env, 0) do
```
josevalim, line 608: You want to do it inside the function, otherwise only people who precompiled Elixir on Erlang/OTP 24+ would get the benefits. :) Also please add a TODO:

    # TODO: Remove this once we require Erlang/OTP 24+

josevalim:
Looks great, I have added one comment (plus the CI uses master which should be enough).

lib/elixir/lib/system.ex
mtarnovan: I think this will trigger an error since we're using ` --warnings-as-errors`

mtarnovan: Oh, nvm. I found how to disable the compiler error :)

josevalim:

PR #10754, opened 2021-02-26
Fix docs example in ExUnit.Case

dvic:


josevalim:

PR #10756, opened 2021-02-27
docs: GenServer.start_link/3 instead of GenServer.start_link/2

hoangbits:
as We were saying: `Although in the example above we have used` `GenServer.start_link/3` at: https://github.com/elixir-lang/elixir/blob/3e453df186f474a3a74223ebba972d2006f34fc4/lib/elixir/lib/gen_server.ex#L70
And I didn't see `GenServer.start_link/2` exists.

### DIFF
--- a/lib/elixir/lib/gen_server.ex
+++ b/lib/elixir/lib/gen_server.ex
@@ -52,7 +52,7 @@ defmodule GenServer do
       GenServer.call(pid, :pop)
       #=> :world
 
-  We start our `Stack` by calling `start_link/2`, passing the module
+  We start our `Stack` by calling `start_link/3`, passing the module
   with the server implementation and its initial argument (a list
   representing the stack containing the element `:hello`). We can primarily
   interact with the server by sending two types of messages. **call**


lib/elixir/lib/gen_server.ex
```diff
@@ -52,7 +52,7 @@ defmodule GenServer do
       GenServer.call(pid, :pop)
       #=> :world
 
-  We start our `Stack` by calling `start_link/2`, passing the module
+  We start our `Stack` by calling `start_link/3`, passing the module
```
wojtekmach, line 55: `GenServer.start_link` is defined as:

```
def start_link(module, init_arg, options \\ [])
```

so referring to it as either `start_link/2` or `start_link/3` is totally fine. Since we happen to use the 2-arity variant in the code snippet above, I think it's fine if we keep it as is.

hoangbits: yes, Thank you. I was a bit confused when reading at line 70 as we mentioned back to line 55 but with different arity.
https://github.com/elixir-lang/elixir/blob/3e453df186f474a3a74223ebba972d2006f34fc4/lib/elixir/lib/gen_server.ex#L70

josevalim:
Closing per the above, thank you!

PR #10757, opened 2021-02-27
Copy upstream documentation change about regex ascii character class deficiency

LostKobrakai:
See:
https://github.com/erlang/otp/pull/4551/files


josevalim:

PR #10775, opened 2021-03-09
Add arities to `is_function` guards

sabiwara:
Hi! I spotted several `is_function` guards that are not checking for the arity, but where the actual function signature clearly expects a specific arity. This PR makes these guards more specific.


josevalim:
We discussed this in the past and we agreed at the time not checking for the arity actually led to better error messages when invoking the function, because it would say the arguments that you passed and besides the expected arity, beyond a function clause error. So I think we should keep it as is for now, thank you!

sabiwara:
Got it, thank you for the explanation!

PR #10780, opened 2021-03-10
Clarify what type1 and type2 mean in maybe_improper_list typespec

doorgan:
The typespecs for `maybe_improper_list` and `nonempty_improper_list` have little documentation, both in Elixir and Erlang docs. While the name is self-explanatory, the meaning of `type1` amd `type2` in the expression `maybe_improper_list(type1, type2)` is not clarified.

The [Erlang docs on typespecs](https://erlang.org/doc/reference_manual/typespec.html) have the following comments:
```erlang
| maybe_improper_list(Type1, Type2)    %% Type1=contents, Type2=termination
| nonempty_improper_list(Type1, Type2) %% Type1 and Type2 as above
```

This PR just mirrors these two comments in the Elixir docs.

### DIFF
--- a/lib/elixir/pages/typespecs.md
+++ b/lib/elixir/pages/typespecs.md
@@ -62,8 +62,8 @@ The notation to represent the union of types is the pipe `|`. For example, the t
                                                         ## Lists
           | list(type)                                  # proper list ([]-terminated)
           | nonempty_list(type)                         # non-empty proper list
-          | maybe_improper_list(type1, type2)           # proper or improper list
-          | nonempty_improper_list(type1, type2)        # improper list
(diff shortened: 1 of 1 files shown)


eksperimental:
Wouldn't it be more readable to do 

```
          | maybe_improper_list(content_type, termination_type)             # proper or improper list
          | nonempty_improper_list(content_type, termination_type)        # improper list
          | nonempty_maybe_improper_list(content_type, termination_type)  # non-empty proper or improper list
```

Please note that `nonempty_maybe_improper_list/2` also needs to be updated

doorgan:
I agree, that's more readable than my propose changes. Should I open a new PR?

eksperimental:
> I agree, that's more readable than my proposed changes. Should I open a new PR?

I would say so.

PR #10782, opened 2021-03-10
Improve improper lists typespecs readability

doorgan:
This is a follow up of #10780

As noted by @eksperimental, it would be more readable if `type1` and `type2` would be renamed instead of adding additional comments to clarify their meaning.
Additionally, `nonempty_maybe_improper_list/2` was left untouched in the previous PR.


josevalim:

eksperimental:
Awesome! Thank you very much @doorgan

PR #10786, opened 2021-03-11
Make default handler process Elixir logs

hauleth:
By default `default` handler will process messages coming only from `[opt, sasl]` and `[]` (empty) domain. This commit adds configuration option that will allow setting Logger to add additional filter to `default` handler, so Elixir messages will be logged as well.

Current workaround is to call

```elixir
:logger.add_handler_filter(
  :default,
  :elixir_filter,
  {&:logger_filters.domain/2, {:log, :sub, [:elixir]}
)
```

Somewhere in your application during startup, but that makes it lost log messages that happened between Logger startup and application startup, so for example early boot logs from all dependencies can be lost.

Now if user want to handle Elixir messages in `default` handler, then they need to use:

```elixir
config :logger,
  handle_otp_reports: false,
  use_erlang_default_handler: true
```

### DIFF
--- a/lib/logger/lib/logger.ex
+++ b/lib/logger/lib/logger.ex
@@ -249,7 +249,7 @@ defmodule Logger do
   `Logger.metadata/1` applies to both, and so on.
 
   Elixir also supports formatting Erlang reports using Elixir syntax.
-  This can be controlled with two configurations:
+  This can be controlled with three configurations:
 
     * `:handle_otp_reports` - redirects OTP reports to `Logger` so
       they are formatted in Elixir terms. This effectively disables
@@ -262,6 +262,10 @@ defmodule Logger do
       in Erlang syntax until the Logger application kicks in.
       Defaults to `false`. This option only has an effect if
       `:handle_otp_reports` is true.
+
+    * `:use_erlang_default_handler` - add logging filter to `:default`
+      handler defined by the OTP. This option only has an effect if
+      `:handle_otp_reports` is false.
 
   For example, to configure `Logger` to redirect all Erlang messages using a
   `config/config.exs` file:
--- a/lib/logger/lib/logger/app.ex
+++ b/lib/logger/lib/logger/app.ex
@@ -9,6 +9,7 @@ defmodule Logger.App do
   def start(_type, _args) do
     start_options = Application.get_env(:logger, :start_options)
     otp_reports? = Application.fetch_env!(:logger, :handle_otp_reports)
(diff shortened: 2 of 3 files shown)


josevalim:
Thanks! This still seems a bit brittle as custom domains still won't be logged, no?

josevalim:
It feels like this should be a feature in OTP so you can customize the default handler to consider other domains.

hauleth:
@josevalim Elixir prepends the `:elixir` domain to the provided one (which can be a little surprising on it's own), so this one will work, as it allows all subdomains of `:elixir`.

> It feels like this should be a feature in OTP so you can customize the default handler to consider other domains.

Technically it is, in OTP you can set:

```erlang
[{kernel, [{logger, [
  {handler, default, logger_std_h, #{
    filters => [
      {remote_gl, {fun logger_filters:remote_gl/2, stop}},
      {domain, {fun logger_filters:domain/2, {log, super, [otp, sasl]}}},
      {no_domain, {fun logger_filters:domain/2, {log, equal, []}}},
      {my_domain, {fun logger_filters:domain/2, {log, sub, [my_domain]}}}
    ]
  }
]}]}].
```

In your `sys.config` and it will work. Even in Rebar3 AFAIK as they are doing some manual handling to reconfigure `logger` after loading configuration file. So the problem there is that Elixir do not allow to configure `kernel.logger` directly, and we need to go through such workarounds instead.

josevalim:
Is your concern mostly in releases or also Dev?

hauleth:
Actually - both. I know that in release you can do it via `config/config.exs` (even if there is big warning that it is no-no), but I think that it still can be pretty confusing to the developers that they need to go through some workarounds in dev that aren't needed in the prod.

---

Honestly, I think that it would be the best, if we would allow configuring `kernel.logger` in the development as well. I know that there are concerns that this would be "special", as other `kernel` configuration options will not work in the development, but personally I think it is the cleanest solution out there, that will provide cleaner API if we decide to encourage people to use Erlang handlers in favour (or next to) Elixir backends (as right now there is some data missing in Elixir backends, like for example support for all 7 log levels or direct access to structured logging).

josevalim:
Per our conversation on IRC, we could merge something like this although I would make it more general to match `:logger.add_handlers` functionality. Perhaps call it `:erlang_logger_add_handlers`. @hauleth also mentioned a proof of concept of integrating the options more, which we could look into. But I don't think we should merge this particular option because it is very fine grained and we may need to add others in the future. Thank you!

PR #10788, opened 2021-03-12
Introduce `reshape/3` macro for pipelines

am-kantox:
As usual, this is easier to show rather than to explain.

Quite often people need to break a pipeline to reshape the outcome, because the next function in the chain inspect it slighly different. For instance:

```elixir
{:ok, content} = 
  ...
  |> File.read() # I know about File.read!/2, that’s an example
File.write(content)
```

or

```elixir
[:a, :b]
|> Enum.with_index()
|> Enum.map(fn {k, v} -> {v, k} end)
|> ...
```

The introduced macro `reshape/3` would, well, reshape the value to fit in the pipeline. The examples above would become:

```elixir
...
|> File.read()
|> reshape({:ok, content}, content)
|> File.write(content)
```

and 

```elixir
[:a, :b]
|> Enum.with_index()
|> Enum.map(&reshape(&1, {k, v}, {v, k}))
|> ...
```


josevalim:
Hi @am-kantox, given we have recently added then/1, I propose to use that instead:

...
|> File.read()
|> then(fn {:ok, content} -> content end)
|> File.write(content)
```

Issue #3152, opened 2015-03-05
Protocol consolidation seems incomplete/inconsistent


gamepoet:
After doing protocol consolidation, not all protocols appear to be consolidated as reported by `Protocol.consolidated?/1`. I've tried this with both 1.0.2 and master 7d81909bc74bd3ba812853a24365507614f1be64.

### Steps

```
$ mix new protocol_test
$ cd protocol_test
$ mix compile && mix compile.protocols
$ iex -pa _build/dev/consolidated -S mix
Erlang/OTP 17 [erts-6.3.1] [source] [64-bit] [smp:8:8] [async-threads:10] [hipe] [kernel-poll:false] [dtrace]

Interactive Elixir (1.1.0-dev) - press Ctrl+C to exit (type h() ENTER for help)
iex(1)> :code.get_path |> Protocol.extract_protocols |> Enum.each fn(x) -> IO.puts "#{x}: #{Protocol.consolidated?(x)}" end
Elixir.Access: false
Elixir.Collectable: true
Elixir.Enumerable: true
Elixir.Inspect: true
Elixir.List.Chars: false
Elixir.Range.Iterator: true
Elixir.String.Chars: true
Elixir.Access: false
Elixir.Collectable: true
Elixir.Enumerable: true
Elixir.Inspect: true
Elixir.List.Chars: false
Elixir.Range.Iterator: true
Elixir.String.Chars: true
:ok
```

Note that both `Access` and `List.Chars` did not get consolidated. I imagine every protocol is in the list twice because they're in the code path twice. Now to make things more interesting, add dependencies for  `ecto ~> 0.8.0` and `postgrex >= 0.0.0`.

```
$ mix deps.get && mix deps.compile && mix compile && mix compile.protocols
$ iex -pa _build/dev/consolidated -S mix
Erlang/OTP 17 [erts-6.3.1] [source] [64-bit] [smp:8:8] [async-threads:10] [hipe] [kernel-poll:false] [dtrace]

Interactive Elixir (1.1.0-dev) - press Ctrl+C to exit (type h() ENTER for help)
iex(1)> :code.get_path |> Protocol.extract_protocols |> Enum.each fn(x) -> IO.puts "#{x}: #{Protocol.consolidated?(x)}" end
Elixir.Access: false
Elixir.Collectable: false
Elixir.Ecto.Queryable: true
Elixir.Enumerable: true
Elixir.Inspect: true
Elixir.List.Chars: false
[...]

PR #10790, opened 2021-03-13
Fix old consolidated protocols being used in test coverage when protocol consolidation is disabled.

baseballlover723:
It appears that running tests with code coverage will use consolidated protocols (if they exist) regardless if protocol consolidation is enabled. I have a repo that reliably reproduces this here: https://github.com/baseballlover723/mix_consolidate_protocols.

I only tested this with turning protocol consolidation on and off, but I would presume that this would cause issues if the protocol is changed.

I'm not sure if https://github.com/elixir-lang/elixir/issues/3152 (https://github.com/elixir-lang/elixir/commit/6b7bf013f5b40f8fe05df201c52f23fd0f0f0944) is still an issue, or if its just vestigial at this point.

Another course of action could be to modify `mix clean` to clean the consolidated protocols, because it currently doesn't for umbrella projects (unsure about non umbrella projects)

### DIFF
--- a/lib/mix/lib/mix/project.ex
+++ b/lib/mix/lib/mix/project.ex
@@ -137,7 +137,7 @@ defmodule Mix.Project do
     [
       build_embedded: config[:build_embedded],
       build_per_environment: config[:build_per_environment],
-      consolidate_protocols: false,
+      consolidate_protocols: config[:consolidate_protocols],
       consolidation_path: consolidation_path(config),
       deps_path: deps_path(config),
       env_path: build_path(config)
--- a/lib/mix/lib/mix/tasks/test.coverage.ex
+++ b/lib/mix/lib/mix/tasks/test.coverage.ex
@@ -76,7 +76,7 @@ defmodule Mix.Tasks.Test.Coverage do
     config = Mix.Project.config()
     test_coverage = config[:test_coverage] || []
     {cover_paths, compile_paths} = apps_paths(config, test_coverage)
-    pid = cover_compile(compile_paths)
+    pid = cover_compile(config, compile_paths)
 
     case Enum.flat_map(cover_paths, &Path.wildcard(Path.join(&1, "*.coverdata"))) do
       [] ->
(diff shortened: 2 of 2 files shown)


lib/mix/lib/mix/tasks/test.coverage.ex
```diff
@@ -166,7 +166,7 @@ defmodule Mix.Tasks.Test.Coverage do
       end
 
     for file <- File.ls!(dir), Path.extname(file) == ".beam" do
-      dir = if file in consolidated, do: consolidation_dir, else: dir
+      dir = if config[:consolidate_protocols] and file in consolidated, do: consolidation_dir, else: dir
```
baseballlover723, line 169: this could probably be short circuited earlier.

lib/mix/lib/mix/project.ex
```diff
@@ -137,7 +137,7 @@ defmodule Mix.Project do
     [
       build_embedded: config[:build_embedded],
       build_per_environment: config[:build_per_environment],
-      consolidate_protocols: false,
+      consolidate_protocols: config[:consolidate_protocols],
```
josevalim, line 140: This should be false because we never want to consolidate it on the children!

josevalim:
Looks great! I have dropped one comment. Also, please make sure you run `bin/elixir bin/mix format` :) Thank you!

lib/mix/lib/mix/project.ex
baseballlover723: by keeping this as false, this PR will make it so that it never uses consolidated protocols with code coverage. So if this needs to remain false, then we should close this PR and turn it into an issue, since it's dependent on reading the `consolidate_protocols` value from the config from inside the context of the child application.

josevalim: I see. The issue is that by setting this to true, we will change other parts of Mix and cause it to consolidate when it wouldn't have before. I think the best option here is to check that the consolidation path is indeed in the load path. I will push a fix, thanks!

PR #10791, opened 2021-03-13
Reconfigure Erlang logger on `loadconfig`

hauleth:
This will reconfigure Erlang logger on `mix loadconfig` to provide unified way to configure early-startup logging in development and production environments.

It can cause some confusion for some users, that would try to configure other `kernel` options in `config/config.ex`, however I think that these cases are rare and are limited to the power-users, who already know the limitations of this approach.

Of course there is an issue, that in the future Erlang can expand supported configuration options that are available in `kernel` (I think that the nearest thing that can be done there are global metadata).  However in my humble opinion it is price worth paying, as it will make the logger configuration more straightforward and more integrated in BEAM community as a whole.

### DIFF
--- a/lib/mix/lib/mix/tasks/loadconfig.ex
+++ b/lib/mix/lib/mix/tasks/loadconfig.ex
@@ -21,8 +21,6 @@ defmodule Mix.Tasks.Loadconfig do
   multiple times to load different configs.
   """
 
-  @reserved_apps [:stdlib, :kernel]
-
   @impl true
   def run(args) do
     Mix.Task.reenable("loadconfig")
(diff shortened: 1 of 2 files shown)


hauleth:
Configuring [logger proxy](https://erlang.org/doc/man/logger_chapter.html#proxy) is currently not supported, as I do not know how Elixir is supposed to handle that, as this is available only in OTP 21.3+. If someone has any guidance how to handle such case, then I am willing to update this PR.

josevalim:
I have two concerns about this approach:

1. we are special casing some of kernel configuration
2. we Have to reimplement and handle Erlang internals, leading to inconsistencies 

unless there is a public API in Erlang to do this, then I would prefer to not go down this road. Thank you!

PR #10793, opened 2021-03-13
Update new.ex

chriseyre2000:
Ignore the /.elixir_ls cache in gitignore.


josevalim:
Hi @chriseyre2000, this should be added to your global gitignore, alongside any other editor specific configuration.

PR #10796, opened 2021-03-15
Add timeout: :infinity and ordered: false in compile.erlang task

lukaszsamson:
It looks like it was missed in https://github.com/elixir-lang/elixir/pull/10789#event-4458704647


josevalim:

PR #10804, opened 2021-03-17
Added Path.expand tilde expansion for certain user

matildeopbravo:
I am of the opinion that `~user` should expand to `/home/user` so I added a  function which, first verifies whether `/home/user` exists and then expands if it does exist.
Since this is an expand  command and, the way I see it, it's supposed to mimic shell behaviour, I believe this should be the default behaviour.

I attach a screenshot of the comparison between shell expansion and default expansion with Path.expand before my added changes.
image

### DIFF
--- a/lib/elixir/lib/path.ex
+++ b/lib/elixir/lib/path.ex
@@ -685,14 +685,9 @@ defmodule Path do
 
   defp resolve_home(rest) do
     case {rest, major_os_type()} do
-      {"\\" <> _, :win32} ->
-        System.user_home!() <> rest
-
-      {"/" <> _, _} ->
-        System.user_home!() <> rest
-
-      _ ->
-        expand_user_home(rest)
+      {"\\" <> _, :win32} -> System.user_home!() <> rest
+      {"/" <> _, _} -> System.user_home!() <> rest
+      _ -> expand_user_home(rest)
     end
   end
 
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/path.ex
```diff
@@ -687,7 +687,19 @@ defmodule Path do
...
       {"/" <> _, _} -> System.user_home!() <> rest
-      _ -> "~" <> rest
+      _ -> expand_user_home(rest)
+    end
+  end
+
+  defp expand_user_home(from_user) do
+    [user | rest] = String.split(from_user, "/")
+    rest = Enum.join(rest, "/")
+    user_home = "/home/" <> user
```
josevalim, line 697: We probably need to use `System.user_home!` instead of "/home/".

josevalim:
Hi @matildeopbravo, thanks for the PR! Last time we discussed this, we decided to skip this behaviour because it leads to conditional expansion of path, which can be surprising/error prone. For example, maybe someone shares a script using "~foo", which would work on their machine, but not generally.

So while we think this is a good feature to have in your shell, we do have reservations about supporting it inside the language.

lib/elixir/lib/path.ex
matildeopbravo: I was trying to get the home for any user, not only for the current one.

matildeopbravo:
> Hi @matildeopbravo, thanks for the PR! Last time we discussed this, we decided to skip this behaviour because it leads to conditional expansion of path, which can be surprising/error prone. For example, maybe someone shares a script using "~foo", which would work on their machine, but not generally.
> 
> So while we think this is a good feature to have in your shell, we do have reservations about supporting it inside the language.

I understand

PR #10807, opened 2021-03-23
Update warnings for OTP master

wojtekmach:


wojtekmach:
There's one test failure:

```
  1) test with and do clauses emit errors, else clauses do not (Kernel.WarningTest)
     lib/elixir/test/elixir/kernel/warning_test.exs:1200
     Assertion with =~ failed
     code:  assert capture_err(fn -> Code.compile_string("with {:first, int} when is_integer(int) <- {:second, Integer.gcd(2, 4)} do\n  int\nend\n") end) =~ "this clause cannot match"
     left:  ""
     right: "this clause cannot match"
     stacktrace:
       lib/elixir/test/elixir/kernel/warning_test.exs:1201: (test)
```

(https://github.com/elixir-lang/elixir/blob/a362e11e20b03b5ca7430ec1f6b4279baf892840/lib/elixir/test/elixir/kernel/warning_test.exs#L1201:L1207)

and I'm curious if it was a false positive and thus we should remove the test?

wojtekmach:
Oh, I noticed that this file:

```elixir
# a.ex
with {:first, int} when is_integer(int) <- {:second, Integer.gcd(2, 4)} do
  int
end
```

produces no warnings, but this one:

```elixir
# a.ex
defmodule M do
  def f do
    with {:first, int} when is_integer(int) <- {:second, Integer.gcd(2, 4)} do
      int
    end
  end
end
```

does:

```
$ elixir a.ex
warning: this clause cannot match because of different types/sizes
  a.ex:3

```

wojtekmach:
I confirmed that this file:

```
# a.ex
with {:first, int} when is_integer(int) <- {:second, Integer.gcd(2, 4)} do
  int
end
```

does produce warning on OTP 23.x and OTP 24.0-rc1 but does _not_ on OTP master.

ericmj (approved):
💜

lib/elixir/src/elixir_erl_compiler.erl
```diff
@@ -97,6 +97,8 @@ handle_file_warning(true, _File, {_Line, sys_core_fold, {nomatch_shadow, _}}) ->
 
 %% Those we implement ourselves
 handle_file_warning(_, _File, {_Line, v3_core, {map_key_repeated, _}}) -> ok;
+handle_file_warning(_, _File, {_Line, sys_core_fold, {ignored, useless_building}}) -> ok;
+%% TODO: remove when we require OTP 24
```
josevalim, line 101:
```suggestion
%% TODO: remove when we require Erlang/OTP 24
```

lib/elixir/src/elixir_erl_compiler.erl
```diff
@@ -130,12 +132,20 @@ custom_format(sys_core_fold, {no_effect, {erlang, F, A}}) ->
   end,
   io_lib:format(Fmt, Args);
 
-%% Rewrite nomatch_guard to be more generic it can happen inside if, unless, and the like
-custom_format(sys_core_fold, nomatch_guard) ->
+%% Rewrite nomatch to be more generic, it can happen inside if, unless, and the like
+custom_format(sys_core_fold, {nomatch, X}) when X == guard orelse X == no_clause ->
```
josevalim, line 136:
```suggestion
custom_format(sys_core_fold, {nomatch, X}) when X == guard; X == no_clause ->
```

PR #10810, opened 2021-03-24
Add support for stepped ranges

josevalim:
Proposal: https://gist.github.com/josevalim/da8f1630e5f515dc2b05aefdc5d01af7

Mailing list discussion: https://groups.google.com/g/elixir-lang-core/c/U5EhplEqda4

TODO:

  * [x] Add coverage for stepped ranges on Enum
  * [x] Fix Range.disjoint? to consider steps

Other notes:

  * With this PR, we can now support stepped ranges on `Enum.slice/2` and `String.slice/2`. For example:
    ```elixir
    iex> String.slice("Elixir", 0..-1//2)
    "Eii"
    ```
    This is not done in this PR. We could even support slice+reverse by having negative steps, such as:
    ```elixir
    iex> String.slice("Elixir", -1..0//-2)
    "rxl"
    ```
    but unfortunately we can't do so right now because 3..-1 has a negative step. Therefore, in order to support slice+reverse, we first need to raise for negative steps on Elixir v2.0 (which is not scheduled nor planned at the moment) and then add this feature later on.

### DIFF
--- a/lib/elixir/lib/code/formatter.ex
+++ b/lib/elixir/lib/code/formatter.ex
@@ -12,7 +12,7 @@ defmodule Code.Formatter do
   @empty empty()
   @ampersand_prec Code.Identifier.unary_op(:&) |> elem(1)
 
-  # Operator that are composed of multiple binary operators
+  # Operators that are composed of multiple binary operators
   @multi_binary_operators [:..//]
 
   # Operators that do not have space between operands
--- a/lib/elixir/lib/code/formatter.ex
+++ b/lib/elixir/lib/code/formatter.ex
@@ -13,7 +13,7 @@ defmodule Code.Formatter do
   @ampersand_prec Code.Identifier.unary_op(:&) |> elem(1)
 
   # Operators that do not have space between operands
-  @no_space_binary_operators [:..]
+  @no_space_binary_operators [:.., :"//"]
 
   # Operators that do not have newline between operands (as well as => and keywords)
   @no_newline_binary_operators [:\\, :in]
@@ -319,7 +319,7 @@ defmodule Code.Formatter do
     {next_eol, comments, doc}
   end
 
-  # Special AST nodes from compiler feedback.
+  # Special AST nodes from compiler feedback
 
   defp quoted_to_algebra({{:special, :clause_args}, _meta, [args]}, _context, state) do
     {doc, state} = clause_args_to_algebra(args, state)
@@ -510,6 +510,11 @@ defmodule Code.Formatter do
   # left not in right
   defp quoted_to_algebra({:not, meta, [{:in, _, [left, right]}]}, context, state) do
     binary_op_to_algebra(:in, "not in", meta, left, right, context, state)
+  end
+
+  # 1..2//3
+  defp quoted_to_algebra({:..//, meta, [left, middle, right]}, context, state) do
+    quoted_to_algebra({:"//", meta, [{:.., meta, [left, middle]}, right]}, context, state)
   end
 
   defp quoted_to_algebra({:fn, meta, [_ | _] = clauses}, _context, state) do
@@ -2027,8 +2032,6 @@ defmodule Code.Formatter do
     {wrap_in_parens_if_operator(doc, ast), state}
   end
 
-  # TODO: We can remove this workaround once we remove
-  # ?rearrange_uop from the parser on v2.0.
   defp wrap_in_parens_if_operator(doc, {:__block__, _, [expr]}) do
     wrap_in_parens_if_operator(doc, expr)
   end
--- a/lib/elixir/lib/code/formatter.ex
+++ b/lib/elixir/lib/code/formatter.ex
@@ -12,6 +12,9 @@ defmodule Code.Formatter do
   @empty empty()
   @ampersand_prec Code.Identifier.unary_op(:&) |> elem(1)
 
+  # Operator that are composed of multiple binary operators
+  @multi_binary_operators [:..//]
+
   # Operators that do not have space between operands
   @no_space_binary_operators [:.., :"//"]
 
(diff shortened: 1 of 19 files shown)


lib/elixir/lib/kernel.ex
```diff
@@ -3631,58 +3631,120 @@ defmodule Kernel do
...
 
-  If last is larger than first, the range will be increasing from
-  first to last. If first is larger than last, the range will be
-  decreasing from first to last. If first is equal to last, the range
-  will contain one element, which is the number itself.
+  If last is more than first, the range will be increasing from
+  first to last. If first is equal to last, the range will contain
+  one element, which is the number itself.
+
+  If first is less than last, the range will be decreasing from first
```
stefanchrobot, line 3640: Shouldn't this be "If first is more than last"?

lib/elixir/lib/kernel.ex
```diff
@@ -3631,58 +3631,120 @@ defmodule Kernel do
...
+      false
+      iex> 2 in 1..3//1
+      true
+      iex> 2 in 1..3//2
+      false
+
+      iex> Enum.to_list(1..3//1)
+      [1, 2, 3]
+      iex> Enum.to_list(1..3//2)
+      [1, 2, 3]
```
stefanchrobot: Should this be `[1, 3]`?

lib/elixir/lib/kernel.ex
```diff
@@ -3631,58 +3631,120 @@ defmodule Kernel do
...
+      iex> Enum.to_list(3..1//-1)
+      [3, 2, 1]
```
stefanchrobot, line 3705: Personally, I find the negative step a bit surprising - I'd say that `[3, 2, 1]` is a range that "goes down by 1" rather than a range that "goes down by -1".

josevalim: How we express things in code will not always translate to a natural language in all possible natural languages. But I believe this definition to be logically quite good, especially because once we added steps I was able to simplify some of our traversal code.

lib/elixir/lib/range.ex
```diff
@@ -40,9 +40,13 @@ defmodule Range do
   not materialize the whole list of integers.
   """
 
-  defstruct first: nil, last: nil
+  @enforce_keys [:first, :last, :step]
+  defstruct first: nil, last: nil, step: nil
 
-  @type t :: %__MODULE__{first: integer, last: integer}
+  @type first :: integer
```
kipcole9, line 56: I wonder if this opens the opportunity to consider ranges that are floats? Given that the step is now explicit it would seem to remove the ambiguity associated with enumerating float ranges?

lib/elixir/test/elixir/kernel/parser_test.exs
```diff
@@ -15,6 +15,46 @@ defmodule Kernel.ParserTest do
...
+                 1,
+                 {:foo, [line: 1], [[do: {:__block__, [], []}]]},
+                 {:bar, [line: 1], [{:bat, [line: 1], nil}, {:baz, [line: 1], nil}]}
+               ]
+             }
+    end
+
+    test "errors" do
+      msg =
+        ~r/the range step operator \(\/\/\) must immediatelly follow the range definition operator \(\.\.\)/
```
esl-ktec, line 50: immediately

wojtekmach, line 50:
```suggestion
        ~r/the range step operator \(\/\/\) must immediately follow the range definition operator \(\.\.\)/
```

josevalim:
This should be ready to go!

lib/elixir/lib/code/formatter.ex
```diff
@@ -12,8 +12,11 @@ defmodule Code.Formatter do
   @empty empty()
   @ampersand_prec Code.Identifier.unary_op(:&) |> elem(1)
 
+  # Operator that are composed of multiple binary operators
```
richardkmichael, line 15: ` # Operators`

lib/elixir/lib/integer.ex
```diff
@@ -463,6 +463,58 @@ defmodule Integer do
   defp gcd_positive(integer1, 0), do: integer1
   defp gcd_positive(integer1, integer2), do: gcd_positive(integer2, rem(integer1, integer2))
 
+  @doc """
+  Returns the extended greatest common divisor of the two given integers.
+
+  It uses the Extended Euclidean algorithm to return a three-element with the `gcd`
```
richardkmichael, line 469: `three-element tuple with`

lib/elixir/lib/range.ex
```diff
@@ -3,21 +3,31 @@ defmodule Range do
...
+  auto-imported from `Kernel`:
+
+      iex> Enum.to_list(1..3)
+      [1, 2, 3]
+      iex> Enum.to_list(1..3//2)
+      [1, 3]
+      iex> Enum.to_list(3..1//-1)
+      [3, 2, 1]
+
+  Intenrally, ranges are represented as structs:
```
richardkmichael, line 18: Internally

lib/elixir/lib/range.ex
```diff
@@ -79,90 +168,130 @@ defmodule Range do
...
+
+      cond do
+        last2 < first1 or last1 < first2 ->
+          true
+
+        abs(step1) == 1 and abs(step2) == 1 ->
+          false
+
+        true ->
+          # We need to find the first intersection of two arithmethetical
```
richardkmichael, line 212: arithmetic

lib/elixir/lib/code/formatter.ex
josevalim, line 15:
```suggestion
  # Operators that are composed of multiple binary operators
```

lib/elixir/lib/integer.ex
josevalim, line 469:
```suggestion
  It uses the Extended Euclidean algorithm to return a three-element tuple with the `gcd`
```

lib/elixir/lib/range.ex
josevalim, line 18:
```suggestion
  Internally, ranges are represented as structs:
```

lib/elixir/lib/range.ex
josevalim, line 212:
```suggestion
          # We need to find the first intersection of two arithmetic
```

josevalim:

Issue #10808, opened 2021-03-23
ExUnit generates an exception on a pattern match with assert


kpanic:
### Environment

* Elixir & Erlang/OTP versions (elixir --version): 
```bash
$ elixir --version
Erlang/OTP 23 [erts-11.1.7] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1] [hipe]

Elixir 1.11.4 (compiled with Erlang/OTP 23)
```

* Operating system: 

Linux

### Current behavior

```elixir
defmodule MatchBubuTest do
  use ExUnit.Case
  doctest MatchBubu

  test "greets the world" do
    assert [%_{payload: %{"hello" => "world"}}] = []
  end
end
```

if I run this test, I get back this not descriptive exception:

```elixir
19:53:08.244 [error] GenServer #PID<0.143.0> terminating
** (FunctionClauseError) no function clause matching in Code.Identifier.inspect_as_atom/1
    (elixir 1.11.4) lib/code/identifier.ex:135: Code.Identifier.inspect_as_atom({:_, [line: 6], nil})
    (ex_unit 1.11.4) lib/ex_unit/diff.ex:919: ExUnit.Diff.safe_to_algebra/2
    (ex_unit 1.11.4) lib/ex_unit/diff.ex:1022: ExUnit.Diff.wrap_on_diff/3
    (elixir 1.11.4) lib/enum.ex:1411: Enum."-map/2-lists^map/1-0-"/2
    (ex_unit 1.11.4) lib/ex_unit/diff.ex:986: ExUnit.Diff.container_to_algebra/5
    (ex_unit 1.11.4) lib/ex_unit/formatter.ex:407: ExUnit.Formatter.format_sides/6
    (ex_unit 1.11.4) lib/ex_unit/formatter.ex:395: ExUnit.Formatter.format_context/4
    (ex_unit 1.11.4) lib/ex_unit/formatter.ex:160: ExUnit.Formatter.format_exception/6
    (ex_unit 1.11.4) lib/ex_unit/formatter.ex:131: anonymous fn/7 in ExUnit.Formatter.format_test_failure/5
    (elixir 1.11.4) lib/enum.ex:1508: anonymous fn/2 in Enum.map_join/3
    (elixir 1.11.4) lib/enum.ex:3449: Enum.map_intersperse_list/3
    (elixir 1.11.4) lib/enum.ex:1508: Enum.map_join/3
[...]

PR #10811, opened 2021-03-24
Fix ex_unit exception

kpanic:
This PR tries to address issue https://github.com/elixir-lang/elixir/issues/10808
I am not happy about this solution.
Comments and suggestions about a better approach are welcome!


josevalim:
Hi @kpanic, thanks for the PR!

We should definitely not change the code identifier for this, as it is used in many other places. We should instead likely change if the argument given to code identifier inside ExUnit.Diff is an atom or not and act accordingly.

kpanic:
hi @josevalim !
I added a new `safe_to_algebra/2` match that does not use `struct_to_algebra/2` but provides directly in the algebra `_`
How does it look now?

thanks!

josevalim:

PR #10822, opened 2021-03-27
Streamline get_and_update specs for Access for Kernel

eksperimental:
In the same fashion of PRs #10228 and #10223

Replaces:
- `get_values` with `current_value`
- `update_value` with `new_value`
- `data` with `new_data` when suitable

Adds nil as an argument term in the function spec.


josevalim:

PR #10825, opened 2021-03-27
Add MIX_INSTALL_DIR and use it in Mix.install/2

wojtekmach:


lib/mix/lib/mix.ex
```diff
@@ -470,12 +472,13 @@ defmodule Mix do
 
   The given `deps` should be in the same format as defined in a regular Mix
   project. See `mix help deps` for more information. As a shortcut, an atom
-  can be given as dependency to mean the latest version, e.g.: specifying
+  can be given as dependency to mean the latest version, i.e.: specifying
```
josevalim, line 475: Let’s not use e.g or I.e.

lib/mix/lib/mix/utils.ex
```diff
@@ -32,6 +32,23 @@ defmodule Mix.Utils do
...
+  Gets possible location of Mix cache.
+
+  Possible locations:
+
+   * `XDG_CONFIG_HOME/mix` (if `MIX_XDG` is set)
+   * `:filename.basedir(:user_cache, "mix")`
+
+  """
+  def mix_cache do
+    if System.get_env("MIX_XDG") in ["1", "true"] do
```
josevalim, line 45: We can always use the MIX_XDG style since this is a new directory.

we should also pass the os based on os.type.

lib/mix/lib/mix.ex
```diff
@@ -251,6 +251,8 @@ defmodule Mix do
     * `MIX_EXS` - changes the full path to the `mix.exs` file
     * `MIX_HOME` - path to Mix's home directory, stores configuration files and scripts used by Mix
       (default: `~/.mix`)
+    * `MIX_INSTALL_DIR` - (since v1.12.0) specifies directory where `Mix.install/2` keeps
+      installs cache.
```
josevalim, line 255:
```suggestion
      installs cache
```

lib/mix/lib/mix.ex
```diff
@@ -517,10 +520,15 @@ defmodule Mix do
       Mix.raise("Mix.install/2 can only be called once")
     end
 
+    installs_root =
+      case System.fetch_env("MIX_INSTALL_DIR") do
+        {:ok, dir} -> dir
+        :error -> Path.join(Mix.Utils.mix_cache(), "installs")
+      end
```
josevalim, line 527:
```suggestion
      System.fetch_env("MIX_INSTALL_DIR") ||
        Path.join(Mix.Utils.mix_cache(), "installs")
```

lib/mix/lib/mix/utils.ex
wojtekmach: Unless MIX_XDG is set, I think we should use OS default location, on macOS that will be ~/Library/Caches.

Couple other notes:

1. :filename.basedir/2 automatically infers the OS
2. :filename.basedir/2 will use `XDG_CACHE_HOME` but only if you happen to be on Linux (or you force linux using :filename.basedir/3)

```
~% XDG_CACHE_HOME=/cache elixir -e 'IO.inspect :filename.basedir(:user_cache, "mix")'
"/Users/wojtek/Library/Caches/mix"
~% XDG_CACHE_HOME=/cache elixir -e 'IO.inspect :filename.basedir(:user_cache, "mix", %{os: :linux})'
"/cache/mix"
```

I tried clarifying this in https://github.com/elixir-lang/elixir/pull/10825/commits/f301eaabe888ac9c6f162f197a96c60624216497.

josevalim:

jonatanklosko:
@wojtekmach this is perfect, thank you :heart: :cat: :ship:

PR #10828, opened 2021-03-27
Zip reduce and Zip reduce while

Adzz:
This is a work in progress but I wanted feedback on the approach I took.

~I took heavy inspiration from the way we do reduce / reduce_while in Enum. The way that works is we have `Enumerable.reduce` which acts as a lower level implementation for `Enum.reduce` and `Enum.reduce_while`.
When zipping we also special case the zipping of two lists to provide a faster implementation.~

~So what I did was implement two (awfully named) functions:~

~* `enumerable_zip_reduce_while`. This is analagous to `Enumerable.reduce` but for zipping.~
~* `fast_enumerable_zip_reduce_while`. This is the special cased "zipping two lists" faster implementation, also analagous to `Enumerable.reduce`~

~The questions I have are:~

~1. Do we want either of those to be public functions?~
~2. Should we instead define a new protocol ? `Zippable` or something?~
~3. What is a more sensible name for those functions.~

Additionally I haven't touched the `Stream` module yet. In my head you could make `enumerable_zip_reduce_while` lazy and get `Stream.zip_reduce` but I have a feeling like someone is going to tell me that's not possible for some reason...

### What's Still Needed

~1. Re-implement existing functions in terms of the `Enum.zip_reduce` / `Enum.zip_reduce_while` (eg `Enum.`)~
2. Moar tests!
3. Improve docs and make them read more good.

### DIFF
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -3432,8 +3432,9 @@ defmodule Enum do
   Zips corresponding elements from two enumerables into a list, transforming them with
   the `zip_fun` function as it goes.
 
-  The corresponding elements from each collection are passed to the provided 2-arity `zip_fun` function in turn.
-  Returns a list that contains the result of calling `zip_fun` for each pair of elements.
+  The corresponding elements from each collection are passed to the provided 2-arity `zip_fun`
+  function in turn. Returns a list that contains the result of calling `zip_fun` for each pair of
+  elements.
 
   The zipping finishes as soon as either enumerable runs out of elements.
 
@@ -3467,9 +3468,10 @@ defmodule Enum do
   Zips corresponding elements from a finite collection of enumerables into list, transforming them with
   the `zip_fun` function as it goes.
 
-  The first element from each of the enums in `enumerables` will be put into a list which is then passed to
-  the 1-arity `zip_fun` function. Then, the second elements from each of the enums are put into a list and passed to
-  `zip_fun`, and so on until any one of the enums in `enumerables` runs out of elements.
+  The first element from each of the enums in `enumerables` will be put into a list which is then
+  passed to the 1-arity `zip_fun` function. Then, the second elements from each of the enums are
+  put into a list and passed to `zip_fun`, and so on until any one of the enums in `enumerables`
+  runs out of elements.
 
   Returns a list with all the results of calling `zip_fun`.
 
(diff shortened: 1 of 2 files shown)


lib/elixir/lib/enum.ex
```diff
@@ -3492,6 +3494,279 @@ defmodule Enum do
     |> :lists.reverse()
   end
 
+  @doc """
+  Reduces a over two enumerables halting as soon as either enumerable is empty.
+
+  ### Zipping Maps
```
Adzz, line 3500: This felt important to highlight.

lib/elixir/lib/enum.ex
```diff
@@ -3492,6 +3494,279 @@ defmodule Enum do
...
+      iex> enums = [[1, 2], %{a: 3, b: 4}, [5, 6]]
+      ...> Enum.zip_reduce(enums, [], fn elements, acc ->
+      ...>   [List.to_tuple(elements) | acc]
+      ...> end)
+      [{2, {:b, 4}, 6}, {1, {:a, 3}, 5}]
+  """
+  def zip_reduce(enums, acc, reducer) do
+    # turn the enums into Enumerable fns that take consume the thing off of it
+    non_stop_reducer = &{:cont, reducer.(&1, &2)}
+    enumerable_zip_reduce_while(enums, {:cont, acc}, non_stop_reducer) |> elem(1)
```
josevalim, line 3574: Instead of implementing our own traversal, you should do like we do in `zip`, which is building a stream zip and then traversing. This should considerably simplify the implementation:

```elixir
   Stream.zip(enumerables).({:cont, []}, &{:cont, [&1 | &2]})
    |> elem(1)
    |> :lists.reverse()
```

lib/elixir/lib/enum.ex
```diff
@@ -3492,6 +3494,279 @@ defmodule Enum do
...
+  the element at the index from each list in turn. If we zip two maps together it's tempting to
+  think that you will get the given key in the left map and the matching key in the right map, but
+  there is no such guarantee because map keys are not ordered! Consider the following:
+
+  ```elixir
+  left =  %{:a => 1, 1 => 3}
+  right = %{:a => 1, :b => :c}
+  Enum.zip(left, right)
+  # [{{1, 3}, {:a, 1}}, {{:a, 1}, {:b, :c}}]
+  ```
```
josevalim, line 3512: Please indent with four spaces instead and don't use the triple backticks. :)

lib/elixir/lib/enum.ex
```diff
@@ -3492,6 +3494,279 @@ defmodule Enum do
...
+  As you can see `:a` does not get paired with `:a`. If this is what you want, you should use
+  `Map.merge/3`
+
+  ### Examples
```
josevalim, line 3516: Please use `##` instead of `###`.

lib/elixir/lib/enum.ex
```diff
@@ -3492,6 +3494,279 @@ defmodule Enum do
...
+
+  @doc """
+  Reduces over two enumerables halting if the accumulator returns `{:halt, value}` or if
+  either of the enumerables is empty.
+
+  The reducer will receive 3 args, the left enumerable's element, the right enumberable's
+  element and the accumulator. It should return one of:
+
+    * `{:halt, value}` - This will halt the reduction and return `{:halted, value}`
+    * `{:suspend, value}` - This will halt the reduction returning `{:suspended, value, continuation}`
```
josevalim, line 3585: We don't allow suspend in reduce_while and we shouldn't allow it here. :)

lib/elixir/lib/enum.ex
```diff
@@ -3492,6 +3494,279 @@ defmodule Enum do
...
+  end
+
+  @doc """
+  Reduces over two enumerables halting if the accumulator returns `{:halt, value}` or if
+  either of the enumerables is empty.
+
+  The reducer will receive 3 args, the left enumerable's element, the right enumberable's
+  element and the accumulator. It should return one of:
+
+    * `{:halt, value}` - This will halt the reduction and return `{:halted, value}`
```
josevalim, line 3584: We don't return halted, do we? We simply return the value.

lib/elixir/lib/enum.ex
```diff
@@ -3492,6 +3494,279 @@ defmodule Enum do
...
+      ...>   end)
+      ...> {:suspended, acc, cont} = cont.({:cont, acc})
+      ...> cont.({:cont, acc})
+      {:done, [6, 4]}
+  """
+
+  # This is analagous to Enumerable.reduce_while. It's quicker than the general solution
+  # so we can get a speed up if we know we are working on lists exclusively.
+
+  def fast_enumerable_zip_reduce_while(_left, _right, {:halt, acc}, _), do: {:halted, acc}
```
josevalim, line 3747: We definitely should not expose this. :)

lib/elixir/test/elixir/enum_test.exs
```diff
@@ -2,7 +2,7 @@ Code.require_file("test_helper.exs", __DIR__)
 
 defmodule EnumTest do
   use ExUnit.Case, async: true
-  doctest Enum
+  doctest Enum, import: true
```
josevalim, line 5: Please don't use imports on doctests. :)

josevalim:
> When zipping we also special case the zipping of two lists to provide a faster implementation. So what I did was implement two (awfully named) functions:

We have this pattern everywhere in Enum, so please check how it is called. But I think we simply use zip_while_reduce_list (and the other is zip_while_reduce).

> Do we want either of those to be public functions?

Nope, because they are implementation details.

> Should we instead define a new protocol ? Zippable or something?

Nope, zipping works fine through the current protocol.

> Additionally I haven't touched the Stream module yet. In my head you could make enumerable_zip_reduce_while lazy and get Stream.zip_reduce but I have a feeling like someone is going to tell me that's not possible for some reason...

Stream functions can't reduce, because once you are reducing you are not longer streaming but accumulating. So there is nothing to do for the Stream module.

lib/elixir/lib/enum.ex
Adzz: Ah yes, true that.

lib/elixir/lib/enum.ex
Adzz: Right. Haha, well that's even quicker than what I did somehow, though I thought my traverse did less reversing!

I'll try that then.

Adzz:
Thanks for the feedback!

> once you are reducing you are not longer streaming but accumulating.

I haven't got the difference clear in my head. The mental model I have is that Streams are like lazy iterators I can imagine our reducing function being "stepped through" in a similar way. Is that wrong? What's the difference between a stream and a lazy reduce?

josevalim:

PR #10832, opened 2021-03-29
Introduce Range.t/3 type

eksperimental:
We should probably deprecate t/2 when we hard deprecate the old range format.


josevalim:
I would like to wait for a use case because I can’t see why someone would like to further restrict the step. :) thank you!

eksperimental:
Well, the most common use case would be to represent a range of consecutive numbers with step `1`  or `-1`.

PR #10833, opened 2021-03-29
Fix t:Range.t/2

eksperimental:
We need to be explicit and define the step field, otherwise IEx and ExDoc will spec it as "term()"

    (1)> t Range.t/2
    @type t(first, last) :: %Range{first: first, last: last, step: term()


josevalim:

PR #10835, opened 2021-03-29
Simplify Range's step error message

eksperimental:


josevalim:
I thought about non-zero but when with a more detailed version for clarify. Thoughts?

eksperimental:
"Different than" does not sound natural to me.
We can use "other than" instead.

josevalim:

PR #10836, opened 2021-03-29
Restrict step in t:Range.t/2

eksperimental:
It improves the changes introduced in 95a6792 (#10833)
As a range defined with only first and last fields should be a consecutive sequence of integers.


josevalim:
The current version is correct. The type API does not have to mirror the definition API. Thanks!

PR #10838, opened 2021-03-30
Do not make Range.empty?/1 public

eksperimental:
`Range.empty?/1` is renamed `Range.__empty__?/1`.

The reason is that `Range.empty?/`1 is overlapped by` Enum.empty?/`1, additionally
it can be checked with Range.size(range) == 0.

`Enum.empty?/1` is optimized to call this function.

`Range.size/1` is simplified by calling this function.

### DIFF
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -958,6 +958,10 @@ defmodule Enum do
     enumerable == []
   end
 
+  def empty?(_.._//_ = range) do
+    Range.__empty__?(range)
+  end
+
   def empty?(enumerable) do
     case Enumerable.slice(enumerable) do
       {:ok, value, _} ->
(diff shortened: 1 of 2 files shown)


josevalim:
If we want to make empty? private, then we should make it truly private. Otherwise it is not fair for us to make it private but continue using it, which means people writing similar code to us cannot.

I would even go as far as removing it altogether and inline the logic inside Range.size. Can you please change accordingly?

josevalim:
I will go ahead and do the change since we are working on a RC, hopefully today! Good call, thank you!

eksperimental:
> If we want to make empty? private, then we should make it truly private. Otherwise it is not fair for us to make it private but continue using it, which means people writing similar code to us cannot.
> 
> I would even go as far as removing it altogether and inline the logic inside Range.size. Can you please change accordingly?

Sorry, I was :sleeping_bed:

PR #10840, opened 2021-03-30
Allow empty sigils as a syntactic sugar for UTC “now”

am-kantox:
This PR provides syntactic sugar for date/time sigils as shown below

```elixir
# ~U[]s
DateTime.truncate(DateTime.utc_now(), :second)

# ~T[]m
Time.truncate(Time.utc_now(), :millisecond)

# ~D[]
Date.utc_today()
```

### DIFF
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -5635,8 +5635,16 @@ defmodule Kernel do
       iex> ~D[2015-01-13]
       ~D[2015-01-13]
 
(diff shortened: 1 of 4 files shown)


josevalim:
Can you please send this up for discussion on the mailing list? I wouldn’t merge it for the empty sigil though. I would prefer ~D[today] or similar, but even then I am not convinced, especially because sigils are often compilation time, so some may expect it to mean the date it was compiled.

Thanks!

Issue #8414, opened 2018-11-21
Support Erlang/OTP 22 new features


josevalim:
To be implemented when Erlang 22 is released:

To be implemented when we require Erlang 22 exclusively:

  * [x] Use `cover:local_only/0`
  * [x] Use `system_info(:ets_count)` instead of `length(:ets.all())` in `IEx.Helpers.runtime_info`
  * [ ] Add `Logger.put_application_level/2` and `Logger.delete_application_level/2`
  * [x] Rely less on MapSet.difference and use `--` instead (as it has been heavily optimized)
  * [x] `RELEASE_NODE` no longer needs the hostname bit (see https://github.com/erlang/otp/pull/2219)
  * [x] Remove TODOs in the codebase that require Erlang/OTP 22+, especially around the Logger
  * [x] Move elixir's static config to persistent terms
  * [x] Update examples to :exsss pseudo-random algo (see #9609)
  * [ ] Introduce :inspect_redact callbacks using persistent terms ([see discussion](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/elixir-lang-core/UFPMvHBg7pA/3VUdnJTkAQAJ))

RobertDober:
Am I wrong? Seems the first issue is already resolved in master (talking about `MapSet.difference`)

josevalim:
@RobertDober not yet. Basically what that feature means is to revisit the use `MapSet` in our codebase and we still use it in certain places.

RobertDober:
Tried to figure out how to help and seem to be unable of finding the code, great for starters 😓 I guess I'll leave it to the experts, sorry for the noise.

josevalim:
@RobertDober none of those issues can be acted on right now anyway since they would require us to depend on Erlang/OTP 22+ exclusively, and we are likely 2 years away from doing so.

RobertDober:
Just out of pure academic interest and only if you do not object to this conversation in this issue @josevalim . 
Would it be 
[] possible
[] simple
[] incredibly much work
[] difficult
[] impossible
to install Elixir with a feature set depending on the Erlang version it is compiled with?

josevalim:
It depends on the feature. For example, leveraging stuff at compilation time, such as changes to the compiler, is relatively easier because we need to perform checks only when the code is compiled.

But for things like the MapSet, then the check needs to happen at runtime, and I just don't think the code ugliness is worth the benefit. It is simpler to wait. It is not like the code is slow today.

[...]

PR #10841, opened 2021-03-30
Use :erlang.system_info(:ets_count) introduced in OTP 22

lukaszsamson:
Addresses https://github.com/elixir-lang/elixir/issues/8414


josevalim:

PR #10842, opened 2021-03-30
Set correct value returned by `get_fun/2` callback

hauleth:
By looking at source of `Access.all/0` it seems that returned value by this function is `container()` instead of tuple `{val, container()}`.


josevalim:

PR #10844, opened 2021-03-30
Simplify specs in Range by introducing t:limit/0

eksperimental:


eksperimental:
There is a second PR doing a copy-edit of all specs and docs once this one is merged

josevalim:
First and last were already public types, so I would prefer to not remove them. Thanks.

eksperimental:
> First and last were already public types, so I would prefer to not remove them. Thanks.

Sorry, they were not
https://hexdocs.pm/elixir/Range.html#types

josevalim:
Ah, so I added it recently? Makes sense.

eksperimental:
> Ah, so I added it recently? Makes sense.

This makes me think if we should be adding the `@since` attribute to public and opaque types.

I know they are not part of the API but it can be useful to keep track of when they had been introduced

PR #10845, opened 2021-03-30
Copyedit Range specs and docs

eksperimental:


lib/elixir/lib/range.ex
```diff
@@ -5,7 +5,7 @@ defmodule Range do
 
   Ranges are always inclusive and they may have custom steps.
   The most common form of creating and matching on ranges is
-  via the `start..stop` and `start..stop//step` notations,
+  via the [`start..stop`](`../2`) and [`start..stop//step`](`..///3`) notations,
```
wojtekmach, line 8: > via the [`start..stop`](`../2`) and [`start..stop//step`](`..///3`) notations,
> defined respectively as the `../2` and `..///3` macros

given we refer to the appropriate macros in the very next line I think the links are redundant but I guess it doesn't hurt to include them.

eksperimental: > > via the [`start..stop`](%60../2%60) and [`start..stop//step`](%60..///3%60) notations,
> > defined respectively as the `../2` and `..///3` macros
> 
> given we refer to the appropriate macros in the very next line I think the links are redundant but I guess it doesn't hurt to include them.

I thought exactly the same, but as you said, it does not hurt.

josevalim:

PR #10850, opened 2021-03-30
Correct and improve definition of decreasing and increasing ranges

eksperimental:


lib/elixir/lib/range.ex
```diff
@@ -40,14 +40,14 @@ defmodule Range do
...
 
   An increasing range `first..last//step` is a range from
-  `first` to `last` increasing by `step` where all values
-  `v` must be `first <= v and v <= last`. Therefore, a range
+  `first` to `last` increasing by `step` where  `step` must be a positive
+  integer and all values `v` must be `first <= v and v <= last`. Therefore, a range
   `10..0//1` is an empty range because there is no value `v`
   that is `10 <= v and v <= 0`.
 
-  Similarly, a decreasing range `first..last//-step` is a range
```
eksperimental, line 48: Here having `-step` is wrong to imply that is a negative integer

PR #10852, opened 2021-03-30
Pass xref's compile args to compile task

szajbus:
`mix xref` task's module doc lists some arguments that are supposed to affect compilation, but they get stripped off by option parser before passing `args` to `compile` task.

In result, `compile` task is run with default arguments and project compilation is always _attempted_.

The PR fixes this.

### DIFF
--- a/lib/mix/lib/mix/tasks/xref.ex
+++ b/lib/mix/lib/mix/tasks/xref.ex
@@ -189,20 +189,12 @@ defmodule Mix.Tasks.Xref do
     min_cycle_size: :integer
   ]
 
-  @compile_args [
-    "--no-compile",
-    "--no-deps-check",
-    "--no-archives-check",
-    "--no-elixir-version-check"
-  ]
-
   @impl true
   def run(args) do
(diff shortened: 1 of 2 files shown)


lib/mix/lib/mix/tasks/xref.ex
```diff
@@ -189,10 +189,19 @@ defmodule Mix.Tasks.Xref do
     min_cycle_size: :integer
   ]
 
+  @compile_args [
+    "--no-compile",
+    "--no-deps-check",
+    "--no-archives-check",
+    "--no-elixir-version-check"
+  ]
+
```
josevalim, line 198: Good find! For the fix, it is probably best to list those options as part of @switches too:

```elixir
    compile: :boolean,
    deps_check: :boolean,
    archives_check: :boolean,
    elixir_version_check: :boolean,
```

josevalim: Move the line that calls compile before we parse them then!

josevalim:

PR #10854, opened 2021-03-31
Add functions for application wise log levels

hauleth:
Ref #8414

### DIFF
--- a/lib/logger/lib/logger.ex
+++ b/lib/logger/lib/logger.ex
@@ -633,7 +633,7 @@ defmodule Logger do
 
   """
   @doc since: "1.11.0"
-  @spec put_module_level(module(), level() | :all | :none) :: :ok | {:error, term()}
+  @spec put_module_level(module() | [module()], level() | :all | :none) :: :ok | {:error, term()}
   defdelegate put_module_level(mod, level), to: :logger, as: :set_module_level
 
   @doc """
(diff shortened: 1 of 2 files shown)


hauleth:
I am preparing PR with fix for the test issues right now.

lib/logger/lib/logger.ex
```diff
@@ -661,6 +661,32 @@ defmodule Logger do
...
+  Puts logging level for modules in given application.
+
+  This will take priority over the primary level set, so it can be
+  used to increase or decrease verbosity of some parts of the project.
+
+  Equivalent of:
+
+      appname |> Application.spec(:modules) |> Logger.put_module_level(level)
+  """
+  @doc since: "1.12.0"
```
josevalim, line 674:
```suggestion
  @doc since: "1.13.0"
```

lib/logger/lib/logger.ex
```diff
@@ -661,6 +661,32 @@ defmodule Logger do
...
+  defdelegate put_application_level(appname, level), to: :logger, as: :set_application_level
+
+  @doc """
+  Deletes logging level for all modules in given application to primary level
+
+  Equivalent of:
+
+      appname |> Application.spec(:modules) |> Logger.delete_module_level()
+  """
+  @doc since: "1.12.0"
```
josevalim, line 685:
```suggestion
  @doc since: "1.13.0"
```

josevalim:
@hauleth the failures you see here seem to be heisenbugs, since they don't fail on master. git is also emit warnings but that does not have an impact on the test suite.

In any case, this looks great, thanks! We just need tests and we can ship it!

hauleth:
@josevalim I do not really have an idea how we could test `Logger.put_application_level/2`, as this would require to have example `app.src` that could be used. Any idea how I can do such stunt?

josevalim:
@hauleth you can use the ExUnit app as an examples (or the EEx one).

josevalim:
To be clear, for the put_application_level, you can simply write a unit test that checks it will set `put_module_level` correctly, as you have tested the put_module_level version. :)

josevalim:

Issue #10857, opened 2021-03-31
Backward imcompatible syntax introduced with new range operator


eksperimental:
I get hit by this a few times already so I am reporting it.
The AST atom representation of the new range operator `:"..//"` is formatted like that.
Now in Elixir v1.12+ the quotes are omitted, 

```elixir
Interactive Elixir (1.13.0-dev) - press Ctrl+C to exit (type h() ENTER for help)
iex(1)> :"..//"
:..//
iex(2)> :..//  
:..//
```

The problem is that prior to v1.12, omitting the quotes will generate a syntax error.

```elixir
Interactive Elixir (1.11.4) - press Ctrl+C to exit (type h() ENTER for help)
iex(1)> :"..//"
:"..//"

iex(1)> :..//      
** (SyntaxError) iex:1:4: syntax error before: '/'
```

So code formatted with `mix format` in v1.12+ may not be backward compatible with earlier versions.
This is because all operators in Elixir are valid atoms and do not need to be quoted.
I think we should make the exception here, and implement the rule in v1.17 once 1.11 is no longer maintained.

PS: I can provide a PR for this

josevalim:
@eksperimental good catch. Let's fix this in the formatter and add a TODO saying we can remove the automatic quotes v1.16. :) PRs welcome!

eksperimental:
> > This is because all operators in Elixir are valid atoms and do not need to be quoted.
> 
> I found one exception to this rule! (well not exception to the rule per se, but different inspect behaviour)
> 
> ```
> iex(1)> :::
> :"::"
> ```
> 
> and I wonder if similar backwards compatibility concerns were the reason.

`:"//"` is the other one.
Here' s the reason behind (taken from the `Code.Identifier` module)

>     * `:not_atomable` - callable operators that must be wrapped in quotes when
>       defined as an atom. For example, `::` must be written as `:"::"` to avoid
>       the ambiguity between the atom and the keyword identifier

josevalim:
`:"//"` is not an actual operator though. It exists in the internal inspection tools but it never surfaces as an operator to developers.

PR #10858, opened 2021-03-31
Quote range operator

eksperimental:
Closes https://github.com/elixir-lang/elixir/issues/10857

I think it is good to keep the separate commits, so when we need to revert this change, we can clearly see what to change. Or at least merge the first two

### DIFF
--- a/lib/elixir/lib/code/identifier.ex
+++ b/lib/elixir/lib/code/identifier.ex
@@ -83,7 +83,7 @@ defmodule Code.Identifier do
     charlist = Atom.to_charlist(atom)
 
     cond do
-      atom in [:%, :%{}, :{}, :<<>>, :..., :.., :., :..//, :->] ->
+      atom in [:%, :%{}, :{}, :<<>>, :..., :.., :., :"..//", :->] ->
         :not_callable
 
       atom in [:"::", :"//"] ->
--- a/lib/elixir/lib/code/formatter.ex
+++ b/lib/elixir/lib/code/formatter.ex
@@ -13,7 +13,7 @@ defmodule Code.Formatter do
   @ampersand_prec Code.Identifier.unary_op(:&) |> elem(1)
 
   # Operators that are composed of multiple binary operators
-  @multi_binary_operators [:..//]
+  @multi_binary_operators [:"..//"]
 
   # Operators that do not have space between operands
   @no_space_binary_operators [:.., :"//"]
@@ -516,7 +516,7 @@ defmodule Code.Formatter do
   end
 
   # 1..2//3
-  defp quoted_to_algebra({:..//, meta, [left, middle, right]}, context, state) do
+  defp quoted_to_algebra({:"..//", meta, [left, middle, right]}, context, state) do
     quoted_to_algebra({:"//", meta, [{:.., meta, [left, middle]}, right]}, context, state)
   end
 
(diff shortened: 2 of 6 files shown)


lib/iex/test/iex/interaction_test.exs
```diff
@@ -19,6 +19,11 @@ defmodule IEx.InteractionTest do
     assert capture_iex("1 + 2") == "3"
   end
 
+  test "quoted operators" do
```
eksperimental, line 22: It may not be the right place t o test this, but Inspect.Algebra does not have any related test

lib/iex/test/iex/interaction_test.exs
```diff
@@ -19,6 +19,11 @@ defmodule IEx.InteractionTest do
...
+    assert capture_iex(~S{:"..//"}) == ~S{:"..//"}
+    assert capture_iex(~S{:"::"}) == ~S{:"::"}
+  end
+
```
josevalim, line 26:
```suggestion
```

lib/elixir/test/elixir/code_identifier_test.exs
```diff
@@ -2,5 +2,14 @@ Code.require_file("test_helper.exs", __DIR__)
 
 defmodule Code.IdentifierTest do
   use ExUnit.Case, async: true
+
+  import CodeFormatterHelpers
+
```
josevalim, line 7:
```suggestion
```

lib/elixir/test/elixir/code_identifier_test.exs
```diff
@@ -2,5 +2,14 @@ Code.require_file("test_helper.exs", __DIR__)
...
   doctest Code.Identifier
+
+  test "quoted operators" do
+    assert_same ~S[:"::"]
+    assert_same ~S[:"..//"]
+    assert_format ~S[:..//], ~S[:"..//"]
+  end
```
josevalim, line 14:
```suggestion
```

lib/elixir/lib/code/identifier.ex
```diff
@@ -138,6 +138,9 @@ defmodule Code.Identifier do
     Atom.to_string(atom)
   end
 
+  # TODO: Remove this clause in v1.16 when we no longer quote operator :..//
+  def inspect_as_atom(:"..//"), do: ":\"..//\""
+
```
josevalim, line 143: I don't think we should add this clause. Inspecting is fine, the issue is only on formatted code.

```suggestion
```

lib/elixir/lib/code/identifier.ex
```diff
@@ -177,6 +180,9 @@ defmodule Code.Identifier do
   @doc """
   Inspects the given identifier as a function name.
   """
+  # TODO: Remove this clause in v1.16 when we no longer quote operator :..//
+  def inspect_as_function(:"..//"), do: "..//"
+
```
josevalim, line 185:
```suggestion
```

josevalim:
@eksperimental I have dropped some comments. I don't want to change the language behaviour, it should act correctly on inspect. The issue is only with the formatter because we want to support earlier versions. We also need to make sure we will format `[..//: :foo]` as `["..//": :foo]`.

eksperimental:
> @eksperimental I have dropped some comments. I don't want to change the language behaviour, it should act correctly on inspect. The issue is only with the formatter because we want to support earlier versions. We also need to make sure we will format `[..//: :foo]` as `["..//": :foo]`.

done.,

PR #10859, opened 2021-04-01
Allow default inspect fun to be set globally

josevalim:


lib/elixir/lib/inspect/algebra.ex
```diff
@@ -105,6 +105,50 @@ defmodule Inspect.Opts do
...
+  Builds an `Inspect.Opts` struct.
+  """
+  def new(opts) do
+    struct(%Inspect.Opts{inspect_fun: default_inspect_fun()}, opts)
+  end
+
+  @doc """
+  Returns the default inspect fun.
+  """
+  def default_inspect_fun do
```
fertapric, line 119: Nitpick: I would use "function" instead of "fun" for the function name, since we don't have "fun" in any other function name and we have `is_function`. There are `*_fun` as types though.

lib/elixir/lib/inspect/algebra.ex
```diff
@@ -105,6 +105,50 @@ defmodule Inspect.Opts do
...
+
+      previous_fun = Inspect.Opts.default_inspect_fun()
+
+      Inspect.Opts.default_inspect_fun(fn
+        %{address: _} = map, opts ->
+          fun.(%{map | address: "[REDACTED]"}, opts)
+
+        value, opts ->
+          fun.(map, opts)
+      end)
```
fertapric, line 147:
```suggestion
      previous_fun = Inspect.Opts.default_inspect_fun()

      Inspect.Opts.default_inspect_fun(fn
        %{address: _} = map, opts ->
          previous_fun.(%{map | address: "[REDACTED]"}, opts)

        map, opts ->
          previous_fun.(map, opts)
      end)

```

fertapric:
I would also add `@spec` and `@doc since:` info 🙂

lib/elixir/lib/inspect/algebra.ex
josevalim: Unfortunately we call it `inspect_fun` in Inspect.Opts. :(

lib/elixir/lib/inspect/algebra.ex
josevalim: The other clause will match any value, no only maps.

fertapric: Updated

v0idpwn (approved):

lib/elixir/lib/inspect/algebra.ex
```diff
@@ -105,6 +105,57 @@ defmodule Inspect.Opts do
...
+  Sets the default inspect function.
+
+  Set this option with care as it will change how all values
+  in the system are inspected. The main use of this functionality
+  is to provide an entry point to filter inspected values,
+  in order for entities to comply with rules and legislations
+  on data security and data privacy.
+
+  It is extremely discouraged for libraries to set their own
+  function as this must be controlled by applications.
```
josevalim, line 137:
```suggestion
  It is **extremely discouraged** for libraries to set their own
  function as this must be controlled by applications. Libraries
  should instead define their own structs with custom inspect
  implementations. If a library must change the default inspect
  function, then it is best to define to ask users of your library
  to explicitly call `default_inspect_fun/1` with your function of
  choice.
```

josevalim:

PR #10866, opened 2021-04-03
Sync unicode data to Unicode 13.0.0

henrythebuilder:
Hi,
`UnicodeData.txt` seam to be alligated to Unicode 10.0, this PR would sync all unicode data source to 13.0.0 
The only change needed in the source code is in the `tokenizer.ex` file to handle new CJK range (3400-4DBF), but for this I ask for help to understand if it is enough.

Enrico


lib/elixir/unicode/tokenizer.ex
```diff
@@ -129,7 +129,7 @@ defmodule String.Tokenizer do
 
   defp unicode_start?(_), do: false
 
-  unless {13312, 19893} in range do
+  unless {13312, 19903} in range do
     raise "CHECK: CJK Ideograph not in range"
```
josevalim, line 133:
```suggestion
    raise "CHECK: CJK Ideograph not in range. Make sure all properties are listed."
```

josevalim:

henrythebuilder:
thanks!

eksperimental:
Thank you @henrythebuilder for the PR!

@josevalim should we backport this (based on Unicode 12) to Elixir v1.12 and probably v1.11 before its last minor release?

josevalim:
I went ahead and backported Unicode 13 to v1.12.

eksperimental:
> I went ahead and backported Unicode 13 to v1.12.

what about 1.11 since we are still supporting that version

josevalim:
It is unlikely we will do a new v1.11 release. If we do, then I will backport just before.

PR #10868, opened 2021-04-03
Mention single-element ranges in range definition

eksperimental:
It is kind of confusing to think of single-element ranges either as increasing and decreasing as per the previous definition.

It got me thinking when `5..5//1` had one element, why `5..5//-1` had one element too.

### DIFF
--- a/lib/elixir/lib/range.ex
+++ b/lib/elixir/lib/range.ex
@@ -49,6 +49,10 @@ defmodule Range do
       iex> -10..-10//1 |> Enum.to_list()
       [-10]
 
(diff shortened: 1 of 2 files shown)


eksperimental:
I am also using this format in the examples,
`iex> 5..5//1 |> Enum.to_list()`

as I think it is visually easier to identify to show the range which is what matters as the beginning of the line.

josevalim:
I will add more examples later but the current definition makes it clear one range elements are supported given there is at least one element where that is between first and last. Thank you.

PR #10869, opened 2021-04-03
Mention update to Unicode 13.0 in CHANGELOG

eksperimental:


josevalim:
We need to start a new version entry, I will do that later. Thank you.

PR #10871, opened 2021-04-03
Change v1.12 docs canonical version

wojtekmach:
We unfortunately had a discrepancy between what the docs said was the canonical version and what was the actual URL:

    % curl --silent https://hexdocs.pm/elixir/1.12/Kernel.html | grep 'rel="canonical"'
          <link rel="canonical" href="https://hexdocs.pm/elixir/v1.12/Kernel.html" />
    % curl --silent https://hexdocs.pm/elixir/1.11/Kernel.html | grep 'rel="canonical"'
          <link rel="canonical" href="https://hexdocs.pm/elixir/v1.11/Kernel.html" />
    % curl --silent https://hexdocs.pm/elixir/1.10/Kernel.html | grep 'rel="canonical"'
          <link rel="canonical" href="https://hexdocs.pm/elixir/v1.10/Kernel.html" />

Notice `elixir/1.12` vs `elixir/v1.12`. The latter 404s!

This seems the easiest way to fix it.

Alternatively, we can handle it on hexdocs/bob side, move the docs to the correct URL and add redirects. That would require  more effort but then again we wouldn't have to re-apply this same patch on all the `v1.*` branches.


josevalim:

PR #10875, opened 2021-04-04
Mix.install/2: Allow calling with the same dependencies and don't generate .app file

wojtekmach:


josevalim:

PR #10876, opened 2021-04-04
Add example to Task.async_stream documentation

thiamsantos:
## Motivation

A common scenario when using tasks is try to trigger the result of the first task to complete, given a list of tasks.

We can accomplish that using Task.async_stream + Stream API, however is not very clear for a newcomer on how to use it.

## Proposed solution

Add example of using Task.async_stream to get the result of the first async task to complete.

Mailing list thread: https://groups.google.com/g/elixir-lang-core/c/ZIFsisK12CM/m/-mDgg-DmCAAJ

### DIFF
--- a/lib/elixir/lib/task.ex
+++ b/lib/elixir/lib/task.ex
@@ -553,7 +553,7 @@ defmodule Task do
       |> Task.async_stream(fn fun -> fun.() end, ordered: false)
       |> Stream.filter(&match?({:ok, _}, &1))
       |> Stream.take(1)
-      |> Enum.to_list()
+      |> Enum.at(0)
 
   First use `Stream.filter/2` to filter the completed tasks, and then use `Stream.take/2` to take the first item of the stream.
 
--- a/lib/elixir/lib/task.ex
+++ b/lib/elixir/lib/task.ex
@@ -540,6 +540,23 @@ defmodule Task do
   If for some reason you cannot take the elements before hand,
   you can use `:max_concurrency` to limit how many elements
   may be over processed at the cost of reducing concurrency.
+
+  ## First async task to complete
+
+  Given a list of async functions, you can get the result of the first async task to complete.
+
+      [
+        &heavy_call_1/0,
+        &heavy_call_2/0,
+        &heavy_call_3/0
+      ]
+      |> Task.async_stream(fn fun -> fun.() end, ordered: false)
+      |> Stream.filter(&match?({:ok, _}, &1))
+      |> Stream.take(1)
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/task.ex
```diff
@@ -540,6 +540,23 @@ defmodule Task do
...
+
+      [
+        &heavy_call_1/0,
+        &heavy_call_2/0,
+        &heavy_call_3/0
+      ]
+      |> Task.async_stream(fn fun -> fun.() end, ordered: false)
+      |> Stream.filter(&match?({:ok, _}, &1))
+      |> Stream.take(1)
+      |> Enum.to_list()
```
benwilson512, line 556: Maybe I'm missing something, but isn't this equivalent to

```
|> Task.async_stream(fn fun -> fun.() end, ordered: false)
|> Stream.filter(&match?({:ok, _}, &1))
|> Enum.take(1)
```

Or even better an `|> Enum.at(0)` if you want `value | nil`?

thiamsantos: @benwilson512 The `Stream.take(1)` as otherwise we may cause more items than requested to be processed. https://hexdocs.pm/elixir/Task.html#async_stream/5-attention-async-take

However `Enum.at(0)` in the place of `Enum.to_list()` would make more sense as we want to get the value directly.

michalmuskala: I think the point in the docs is placing the `take` call before the `async_stream` call, less about the `Enum` vs `Stream` module - to limit the number of tasks that are launched, rather then the number of results collected.

thiamsantos: Yeah, you both are completely right, I misread the example.

lib/elixir/lib/task.ex
```diff
@@ -540,6 +540,22 @@ defmodule Task do
...
+      [
+        &heavy_call_1/0,
+        &heavy_call_2/0,
+        &heavy_call_3/0
+      ]
+      |> Task.async_stream(fn fun -> fun.() end, ordered: false)
+      |> Stream.filter(&match?({:ok, _}, &1))
+      |> Enum.at(0)
+
+  First use `Stream.filter/2` to filter the completed tasks, and then use `Stream.take/2` to take the first item of the stream.
```
benwilson512, line 557:
```suggestion
  First use `Stream.filter/2` to filter the completed tasks, and then use `Enum.at/2` to retrieve the first item of the returned from the stream.
```

josevalim:

PR #10880, opened 2021-04-05
use better example for `Kernel.tap/1`

msz:
The current example for `Kernel.tap/1` is showing using it for calling
`IO.inspect/1`. That's not the most practical example, because it is
already possible to "tap" with `IO.inspect/1` directly without needing a
separate function.

This changes the example to use `Logger.info/1` instead.


eksperimental:
How would you inspect the value of the` :a` key in that example without tap?

msz:
Ah, you're right. It does make sense after all :D

PR #10881, opened 2021-04-05
Upgrade FreeBSD to 13.0 ALPHA in Cirrus CI

eksperimental:
I am upgrading even it is not a stable release yet, to see if test pass

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -23,9 +23,10 @@ test_freebsd_task:
   env:
     CHECK_REPRODUCIBLE: true
     LC_ALL: en_US.UTF-8
+    PATH: $PATH:/usr/local/lib/erlang22/bin
 
   install_script:
-    - pkg install -y erlang git gmake
+    - pkg install -y erlang-runtime22 git gmake
     - rm -rf .git
     - gmake compile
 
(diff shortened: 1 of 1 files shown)


eksperimental:
The reason it is failing it is because the latest FreeBSD port of Erlang is OTP 21.
Any suggestion?

josevalim:
@eksperimental can we use an alternate package manager for FreeBSD?

nifoc:
FreeBSD has other Erlang versions under the erlang-runtimeX package, where X is 20 to 23.

eksperimental:
> FreeBSD has other Erlang versions under the erlang-runtimeX package, where X is 20 to 23.

There we go.
Your suggestion saved the day @nifoc

josevalim:
So we don’t need the alpha version anymore either, right?

eksperimental:
The other question is, do we want to test it against Erlang/OTP 23?

eksperimental:
> So we don’t need the alpha version anymore either, right?

We can leave it as is, and when 13 gets stable, it will automatically use that one. Or we can set it back to 12.

josevalim:

Issue #10865, opened 2021-04-02
Deprecate URI.parse


josevalim:
I was planning to move `URI.parse/1` to use Erlang's `:uri_string.parse/1` and it seems our implementation returns valid results for invalid URLs, such as:

    URI.parse("https:?foo")

Our implementation only uses a Regex, instead of a proper parser as done in Erlang, so we should probably migrate over. This is also a good opportunity to deprecate the :authority field as it causes ambiguity issues. The big question is what should be the new API:

  1.  `URI.new(string | t) :: {:ok, t} | {:error, reason}`
  2.  `URI.dissect(string | t) :: {:ok, t} | {:error, reason}`

Any other options?

wojtekmach:
`URI.new/1` returning `{:ok, t} | {:error, reason}` is consistent with `Date.new/3` (but along with it is inconsistent with `Map.new/0` & `MapSet.new/0` and others)

Some other ideas are `URI.parse_string`, `URI.from_string`, `URI.for`, `URI.build`.

wojtekmach:
Is the idea that `URI.new(t)` would return it as is or it would perform validation on fields?

josevalim:
It would return it as is, similar to how URI.parse behaves today, which can be quite handy. That's why I would like to avoid parse_string or from_string. :)

kelvinst:
> `Date.new/3` (but along with it is unfortunately inconsistent with `Map.new/0` & `MapSet.new/0` and others)

Little off-topic but, do we plan to kill that current inconsistency? I mean, if introducing a new one is a problem, I think we should reevaluate the existent one too. But if we are not doing that, then maybe we don't care as much to inconsistencies in the API. 🤷‍♂️ 

My point being, one inconsistency already makes the API inconsistent. If we worry causing confusion on people, they can get confused with the current inconsistency already, by introducing a new one, we would just raise the probability of that happening. There is even the probability of people that only used `Date.new/3` as it is now, but not `Map.new/0` and `MapSet.new/0`, so those wouldn't even get confused by `URI.new/1` return value.

All that to say: I like `URI.new(string | t) :: {:ok, t} | {:error, reason}`, and if we really are caring for making the API consistent, I vote for deprecating `Map.new/0` and `MapSet.new/0` in favor of them turning into banged functions: `Map.new!/0` and `MapSet.new!/0`.

kelvinst:
[...]

PR #10884, opened 2021-04-08
Require scheme in URIs

dogatuncay:
I ran into an issue involving unexpected behavior from the URI module. The URI module does not appear to follow the specification with respect to the scheme of a URL. This can cause inconsistencies when operating with other tools and libraries that accept URIs.

According to the [spec](https://tools.ietf.org/html/rfc3986#section-3):
> The scheme and path components are required, though the path may be empty (no characters).

However, the URI module doesn't check if the scheme is provided, nor does it set a default scheme. If you don't specify a manual scheme, it comes out like `//abc.com`. I think it should throw an `ArgumentError` to match the existing behavior for checking the correctness of the path component before turning the URI into a string.

### DIFF
--- a/lib/elixir/test/elixir/string/chars_test.exs
+++ b/lib/elixir/test/elixir/string/chars_test.exs
@@ -90,9 +90,6 @@ defmodule String.Chars.URITest do
   test "uri" do
     uri = URI.parse("http://google.com")
     assert String.Chars.to_string(uri) == "http://google.com"
-
-    uri_no_host = URI.parse("/foo/bar")
-    assert String.Chars.to_string(uri_no_host) == "/foo/bar"
   end
 end
 
--- a/lib/elixir/lib/uri.ex
+++ b/lib/elixir/lib/uri.ex
@@ -740,7 +740,7 @@ defimpl String.Chars, for: URI do
             "got: #{inspect(uri)}"
   end
 
-  def to_string(%{scheme: nil}) do
+  def to_string(%{scheme: scheme}) when is_nil(scheme) or scheme == "" do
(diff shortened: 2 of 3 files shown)


chulkilee:
What about empty string? (`""`) Should it be handled like `nil`?

We may allow empty string to allow users to build `//example.com` string... but that's against the spec I guess.

dogatuncay:
> What about empty string? (`""`) Should it be handled like `nil`?
> 
> We may allow empty string to allow users to build `//example.com` string... but that's against the spec I guess.

that's a good point. I'll update my PR.

lib/elixir/test/elixir/string/chars_test.exs
```diff
@@ -90,9 +90,6 @@ defmodule String.Chars.URITest do
   test "uri" do
     uri = URI.parse("http://google.com")
     assert String.Chars.to_string(uri) == "http://google.com"
-
-    uri_no_host = URI.parse("/foo/bar")
-    assert String.Chars.to_string(uri_no_host) == "/foo/bar"
```
dogatuncay, line 95: removed this test, because it doesn't make sense in the new change any more. The path by itself isn't a valid URI.

wojtekmach:
Per URI.parse/1 docs

> Note this function expects a well-formed URI and does not perform any validation. See the "Examples" section below for examples of how URI.parse/1 can be used to parse a wide range of URIs

So making the function stricter would be a breaking change. For this reason, we are discussing deprecating it in favour of a new function that would perform the proper validations, see #10865

josevalim:
The URI module can parse both relative and absolute URLs, see section 4.2. I will clarify this in the docs to avoid further confusion. :) Thank you!

PR #10886, opened 2021-04-08
Fixing bug in Enum.dedup/1

sabiwara:
Hi. I just noticed that `Enum.dedup/1` isn't working properly on non-list enumerables in the 1.12-rc.0.

```elixir
[1, 1, 0, 2] |> Stream.map(& &1) |> Enum.dedup()
[1, 1, 0, 2]
```

After investigating, I realized this was a bug I introduced myself in a previous PR :see_no_evil:.
Really sorry for that.

### DIFF
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -787,7 +787,7 @@ defmodule Enum do
   def dedup(enumerable) do
     Enum.reduce(enumerable, [], fn x, acc ->
       case acc do
-        [^x, _] -> acc
+        [^x | _] -> acc
         _ -> [x | acc]
       end
     end)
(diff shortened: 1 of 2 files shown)


josevalim:
Great find, can you please add a test? Thank you!

sabiwara:
@josevalim sure! Should I introduce a `EnumTest.Stream`? Current tests are basically mostly lists, ranges and maps (and this won't work to reproduce this bug)

josevalim:
There is no need, just add a "dedup with stream" and we should be good!

sabiwara:
Since it is a separate implementation, I just replicated the whole `dedup/1` case with streams.
If we cherry pick assertions, I wasn't sure which one to pick or not.
What do you think?

ericmj (approved):

josevalim:

PR #10891, opened 2021-04-11
shellcheck: Double quote to prevent globbing and word splitting

hi-rustin:
Refine shellcheck issue.


josevalim:
The usage here is correct: both are meant to contain multiple args that should not be treated as a single one. Thank you!

PR #10892, opened 2021-04-11
Added `default` option to `Mix.Shell.IO.yes?`

darraghenright:
Hello 👋 

As [proposed](https://groups.google.com/g/elixir-lang-core/c/yMn-Z4SqhEc) in the core mailing list, this is a change that adds an option to change the default behaviour of the `Enter` key as input for `Mix.Shell.IO.yes?`.

Currently, a user can simply press`Enter` to accept the prompt. The proposed change allows this this action to reject the prompt instead. Its rationale is to provide the option to enforce an explicit confirmation from the user, which may be a desirable safety net if accepting the prompt is for a potentially dangerous or irreversible operation.

As suggested by @josevalim the option would be `default: :yes | :no`; e.g:

```elixir
if Mix.Shell.IO.yes?("Delete everything? [yN]", default: :no) do
  delete_everything!()
end

if Mix.Shell.IO.yes?("Accept non-destructive action? [Yn]", default: :yes) do
  # ...
end
```

As well as changing the default behaviour of `Enter` the prompt also changes according to convention; i.e: `[Yn]` or `[yN]`

The implicit behaviour of this option is `:yes` to preserve existing functionality.

A couple of questions:

1. I added a spec, there are no other specs in this module however, so I can remove this if required (or add specs to other functions?)
2. Accepted values are `["y", "Y", "yes", "YES", "Yes"]`. Should this be case sensitive? I am currently operating under the assumption that this is an explicit allowlist for a reason.

This is my first Elixir PR so I welcome all feedback. Thanks!

### DIFF
--- a/lib/mix/lib/mix/shell.ex
+++ b/lib/mix/lib/mix/shell.ex
@@ -44,7 +44,7 @@ defmodule Mix.Shell do
   @doc """
   Prompts the user for confirmation.
   """
-  @callback yes?(message :: binary) :: boolean
+  @callback yes?(message :: binary, options :: keyword) :: boolean
 
   @doc """
   Prints the current application to the shell if
--- a/lib/mix/lib/mix/shell/io.ex
+++ b/lib/mix/lib/mix/shell/io.ex
@@ -46,10 +46,20 @@ defmodule Mix.Shell.IO do
   end
 
   @doc """
-  Prints a message and asks the user if they want to proceed.
+  Prints a message and asks the user to confirm if they
+  want to proceed. The user must type and submit one of
+  "y", "yes", "Y", "YES" or "Yes".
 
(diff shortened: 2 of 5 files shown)


josevalim:
There are no specs because those functions are callbacks. You should change Mix.Shell and see if you need to change any other callback implementation besides the IO module!

darraghenright:
Ah, I missed that, clearly :) I'll have a look.

darraghenright:
Thanks for the pointers José. I removed the spec from `Mix.Shell.IO.yes?/2` and updated its callback signature in `Mix.Shell`.

I found two other modules that implement the `Mix.Shell` behaviour: `Mix.Shell.Process` and `Mix.Shell.Quiet` and I believe I updated those accordingly. Dialyzer and `make test_mix` seem satisfied with the changes.

josevalim:

PR #10893, opened 2021-04-11
Fix optional callback blame hint

eksperimental:
Previously it was wrongly hinting that the
behaviour Behaviour expected a private or undefined function to be present,
when the callback was optional.


josevalim:

PR #10895, opened 2021-04-12
IEx path autocompletion

### LINKED ISSUE #10888: Do path autocompletion in IEx
@fhunleth, can you kindly remind me what are the conditions to trigger this? Whenever we are at the beginning of a string? Or does the string have to start with `/` or `.`?

kelvinst:
Supporting path autocompletion on strings that start with `".` or `"/`.

Thanks @fhunleth, this PR is basically a hard copy of what you have on your toolshed project.

Closes #10888

### DIFF
--- a/lib/iex/lib/iex/autocomplete.ex
+++ b/lib/iex/lib/iex/autocomplete.ex
@@ -603,7 +603,7 @@ defmodule IEx.Autocomplete do
   defp expand_path(path_fragment, possible_paths) do
     expansions =
       Enum.map(possible_paths, fn {path, dir?} ->
-        {completion_part(path_fragment, path, dir?), to_charlist(Path.basename(path))}
+        {completion_part(path_fragment, path, dir?), String.to_charlist(Path.basename(path))}
       end)
 
     case expansions do
@@ -611,12 +611,12 @@ defmodule IEx.Autocomplete do
         no()
 
       [{unique, _}] ->
-        {:yes, to_charlist(unique), []}
+        {:yes, String.to_charlist(unique), []}
 
       list ->
         {completions, filenames} = Enum.unzip(list)
         common_prefix_size = :binary.longest_common_prefix(completions)
-        hint = completions |> hd() |> binary_part(0, common_prefix_size) |> to_charlist()
+        hint = completions |> hd() |> binary_part(0, common_prefix_size) |> String.to_charlist()
         {:yes, hint, filenames}
     end
   end
--- a/lib/iex/lib/iex/autocomplete.ex
+++ b/lib/iex/lib/iex/autocomplete.ex
@@ -482,7 +482,7 @@ defmodule IEx.Autocomplete do
   ## Ad-hoc conversions
 
   defp to_entries(%{kind: kind, name: name})
-       when kind in [:map_key, :module, :variable] do
+       when kind in [:map_key, :module, :variable, :dir, :file] do
     [name]
   end
 
@@ -491,7 +491,7 @@ defmodule IEx.Autocomplete do
   end
 
   defp to_uniq_entries(%{kind: kind})
-       when kind in [:map_key, :module, :variable] do
+       when kind in [:map_key, :module, :variable, :dir, :file] do
     []
   end
 
@@ -507,8 +507,16 @@ defmodule IEx.Autocomplete do
     format_hint(name, hint) <> "."
   end
 
+  defp to_hint(%{kind: :dir, name: name}, hint) when name == hint do
+    format_hint(name, name) <> "/"
+  end
+
+  defp to_hint(%{kind: :file, name: name}, hint) when name == hint do
+    format_hint(name, name) <> "\""
+  end
+
   defp to_hint(%{kind: kind, name: name}, hint)
-       when kind in [:function, :map_key, :module, :variable] do
+       when kind in [:function, :map_key, :module, :variable, :dir, :file] do
     format_hint(name, hint)
   end
 
@@ -571,60 +579,45 @@ defmodule IEx.Autocomplete do
     :error
   end
 
-  defp path_fragment(expr), do: collect_path_fragment(expr, [])
-
-  # collects the fragment and store it on `acc`, calls `return_path_fragment` 
-  # when it finds a `"`, discards `acc` otherwise
-  defp collect_path_fragment([], _acc), do: []
-  defp collect_path_fragment([?{, ?# | _rest], _acc), do: []
-  defp collect_path_fragment([?", ?\\ | t], acc), do: collect_path_fragment(t, [?\\, ?" | acc])
-  defp collect_path_fragment([?/, ?., ?" | _], acc), do: [?., ?/ | acc]
-  defp collect_path_fragment([?/, ?" | _], acc), do: [?/ | acc]
-  defp collect_path_fragment([?" | _], _acc), do: []
-  defp collect_path_fragment([h | t], acc), do: collect_path_fragment(t, [h | acc])
+  defp path_fragment(expr), do: path_fragment(expr, [])
+  defp path_fragment([], _acc), do: []
+  defp path_fragment([?{, ?# | _rest], _acc), do: []
+  defp path_fragment([?", ?\\ | t], acc), do: path_fragment(t, [?\\, ?" | acc])
+  defp path_fragment([?/, ?., ?" | _], acc), do: [?., ?/ | acc]
+  defp path_fragment([?/, ?" | _], acc), do: [?/ | acc]
+  defp path_fragment([?" | _], _acc), do: []
+  defp path_fragment([h | t], acc), do: path_fragment(t, [h | acc])
 
   defp expand_path(path_fragment) do
-    path_fragment = List.to_string(path_fragment)
-    possible_paths = find_possible_paths(path_fragment)
-    expand_path(path_fragment, possible_paths)
+    path = List.to_string(path_fragment)
+    dir = Path.dirname(path_fragment)
+
+    path
+    |> ls_prefix(dir)
+    |> Enum.map(fn path -> 
+      %{
+        kind: (if File.dir?(path), do: :dir, else: :file), 
+        name: Path.basename(path)
+      } 
+    end)
+    |> format_expansion(path_hint(path, dir))
   end
 
-  defp expand_path(path_fragment, possible_paths) do
-    expansions =
-      Enum.map(possible_paths, fn {path, dir?} ->
-        {completion_part(path_fragment, path, dir?), String.to_charlist(Path.basename(path))}
-      end)
+  defp path_hint(path, dir) do
+    path_size = byte_size(path)
+    dir_size = byte_size(dir)
 
-    case expansions do
-      [] ->
-        no()
-
-      [{unique, _}] ->
-        {:yes, String.to_charlist(unique), []}
-
-      list ->
-        {completions, filenames} = Enum.unzip(list)
-        common_prefix_size = :binary.longest_common_prefix(completions)
-        hint = completions |> hd() |> binary_part(0, common_prefix_size) |> String.to_charlist()
-        {:yes, hint, filenames}
-    end
-  end
-
-  defp completion_part(already_entered, full_path, dir?) do
-    part = String.replace_prefix(full_path, already_entered, "")
-
-    if dir? do
-      part <> "/"
+    if path_size == dir_size do
+      ""
     else
-      part <> "\""
+      binary_part(path, dir_size + 1, path_size - (dir_size + 1))
     end
   end
 
   defp prefix_from_dir(".", <<c, _::binary>>) when c != ?., do: ""
   defp prefix_from_dir(dir, _fragment), do: dir
 
-  defp ls_prefix(path_fragment) do
-    dir = Path.dirname(path_fragment)
+  defp ls_prefix(path_fragment, dir) do
     prefix = prefix_from_dir(dir, path_fragment)
 
     case File.ls(dir) do
@@ -636,11 +629,5 @@ defmodule IEx.Autocomplete do
       _ ->
         []
     end
-  end
-
-  defp find_possible_paths(path_fragment) do
-    path_fragment
-    |> ls_prefix()
-    |> Enum.map(fn path -> {path, File.dir?(path)} end)
   end
 end
--- a/lib/iex/lib/iex/autocomplete.ex
+++ b/lib/iex/lib/iex/autocomplete.ex
@@ -580,15 +580,17 @@ defmodule IEx.Autocomplete do
   defp collect_path_fragment([?" | t], acc), do: return_path_fragment(t, acc)
   defp collect_path_fragment([h | t], acc), do: collect_path_fragment(t, [h | acc])
 
-  # scan the rest of the `expr`, returns `acc` if no other `"` is found, 
+  # scans the rest of the `expr`, returns `acc` if no other `"` is found, 
   # fallback to `discard_path_fragment` otherwise
   defp return_path_fragment([], acc), do: acc
+  defp return_path_fragment([?", ?\\ | t], acc), do: return_path_fragment(t, [?\\, ?" | acc])
   defp return_path_fragment([?" | t], acc), do: discard_path_fragment(t, acc)
   defp return_path_fragment([_ | t], acc), do: return_path_fragment(t, acc)
 
-  # scan the rest of the `expr`, discards `acc` if no other `"` is found, 
+  # scans the rest of the `expr`, discards `acc` if no other `"` is found, 
   # fallback to `return_path_fragment` otherwise
   defp discard_path_fragment([], _acc), do: []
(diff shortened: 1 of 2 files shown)


lib/iex/lib/iex/autocomplete.ex
```diff
@@ -643,4 +647,90 @@ defmodule IEx.Autocomplete do
...
   end
+
+  defp path_fragment(expr), do: collect_path_fragment(expr, [])
+
+  # collects the fragment and store it on `acc`, calls `return_path_fragment` 
+  # when it finds a `"`, discards `acc` otherwise
+  defp collect_path_fragment([], _acc), do: []
+  defp collect_path_fragment([?", ?\\ | t], acc), do: collect_path_fragment(t, [?\\, ?" | acc])
+  defp collect_path_fragment([?{, ?# | _rest], _acc), do: []
+  defp collect_path_fragment([?" | t], acc), do: return_path_fragment(t, acc)
```
kelvinst, line 658: Another way to actually check if the collected fragment is an open string would be to check `rem(Enum.count(t, & &1 == ?"), 2) == 0`. Do you think that would be better?

kelvinst: Actually, nevermind, just realized that would not work because we have to skip the escaped `"`

lib/iex/lib/iex/autocomplete.ex
```diff
@@ -28,8 +28,12 @@ defmodule IEx.Autocomplete do
 
   def expand([h | t] = expr, shell) do
     helper = get_helper(expr)
+    path = path_fragment(expr)
```
josevalim, line 31: We should probably try this only on the last clause of `cond`.

lib/iex/lib/iex/autocomplete.ex
```diff
@@ -643,4 +647,92 @@ defmodule IEx.Autocomplete do
...
+
+  # scans the rest of the `expr`, discards `acc` if no other `"` is found, 
+  # fallback to `return_path_fragment` otherwise
+  defp discard_path_fragment([], _acc), do: []
+  defp discard_path_fragment([?", ?\\ | t], acc), do: discard_path_fragment(t, [?\\, ?" | acc])
+  defp discard_path_fragment([?" | t], acc), do: return_path_fragment(t, acc)
+  defp discard_path_fragment([_ | t], acc), do: discard_path_fragment(t, acc)
+
+  defp expand_path(path_fragment) do
+    path_fragment = to_string(path_fragment)
```
josevalim, line 676:
```suggestion
    path_fragment = List.to_string(path_fragment)
```

lib/iex/lib/iex/autocomplete.ex
```diff
@@ -643,4 +647,92 @@ defmodule IEx.Autocomplete do
...
+    case expansions do
+      [] ->
+        {:no, '', []}
+
+      [{unique, _}] ->
+        {:yes, unique, []}
+
+      list ->
+        {completions, filenames} = Enum.unzip(list)
+        hint = Enum.reduce(completions, &common_prefix/2)
```
josevalim, line 696: See if you can use http://erlang.org/doc/man/binary.html#longest_common_prefix-1

lib/iex/lib/iex/autocomplete.ex
```diff
@@ -643,4 +647,92 @@ defmodule IEx.Autocomplete do
...
+
+  defp ls_prefix(path_fragment) do
+    dir = Path.dirname(path_fragment)
+    prefix = prefix_from_dir(dir, path_fragment)
+
+    case File.ls(dir) do
+      {:ok, list} ->
+        list
+        |> Stream.map(&Path.join(prefix, &1))
+        |> Stream.filter(&String.starts_with?(&1, path_fragment))
```
josevalim, line 726: No reason for streaming in this case as the list is relatively small. :)

lib/iex/test/iex/autocomplete_test.exs
```diff
@@ -408,4 +408,27 @@ defmodule IEx.AutocompleteTest do
     assert expand('take(') == {:yes, '', ['take(enumerable, amount)']}
     assert expand('derive(') == {:yes, '', ['derive(protocol, module, options \\\\ [])']}
   end
+
+  @tag :tmp_dir
+  test "path completion inside strings", %{tmp_dir: dir} do
```
josevalim, line 413: We need a bunch more tests given the complexity of the code. :) I would also create all the files upfront and then have an extensive assertion suite.

lib/iex/lib/iex/autocomplete.ex
kelvinst: Tested here, and it matches earlier on the cond clauses then. For example when I type `"./m<tab>` it would match for this clause: https://github.com/elixir-lang/elixir/pull/10895/files/213d13e29dba087ea2d4b9f3e444e1aa8ac9661c#diff-9e9de7d847547ae049e1297f35e4f7cd0f7d99d6f4cbcf8fa4b90c7c532ee035L45

lib/iex/test/iex/autocomplete_test.exs
kelvinst: Added more assertions, please let me know if you think we need more.

lib/iex/lib/iex/autocomplete.ex
josevalim: @kelvinst if I write this code:

```elixir
iex(1)> module = 123
123
iex(2)> "./mo
```

And hit tab, it won't auto-complete the variable. So it means we are ruling out it somewhere. I thought it was before we trigger this point?

lib/iex/lib/iex/autocomplete.ex
```diff
@@ -643,4 +647,89 @@ defmodule IEx.Autocomplete do
...
+      [] ->
+        {:no, '', []}
+
+      [{unique, _}] ->
+        {:yes, to_charlist(unique), []}
+
+      list ->
+        {completions, filenames} = Enum.unzip(list)
+        common_prefix_size = :binary.longest_common_prefix(completions)
+        hint = completions |> hd() |> binary_part(0, common_prefix_size) |> to_charlist()
```
josevalim, line 697:
```suggestion
        hint = completions |> hd() |> binary_part(0, common_prefix_size) |> String.to_charlist()
```

lib/iex/test/iex/autocomplete_test.exs
```diff
@@ -408,4 +408,50 @@ defmodule IEx.AutocompleteTest do
...
+
+      default_expand = expand('{')
+      assert expand('"./dir/#\{') == default_expand
+    after
+      File.rm("./single1")
+      File.rm("./file1")
+      File.rm("./file2")
+      File.rm("./dir/file3")
+      File.rm("./dir/file4")
+      File.rm("./dir")
```
josevalim, line 454: I don't think you need to remove them (nor the try/catch) if using the tmp_dir feature?

kelvinst: You are totally right, we don't need to remove them as it clears the tmp dir before starting the tests. Should I include `tmp` dir on `.gitignore` then? Just to make sure no one commits that?

lib/iex/lib/iex/autocomplete.ex
kelvinst: > So it means we are ruling out it somewhere.

I think it's discarded here: https://github.com/elixir-lang/elixir/blob/213d13e29dba087ea2d4b9f3e444e1aa8ac9661c/lib/iex/lib/iex/autocomplete.ex#L111-L133

So the cond actually matches opened strings like on your example, it just doesn't autocomplete because `Code.string_to_quoted(expr)` can't build the AST for it

kelvinst: What we could do is to try a `Code.string_to_quoted(reduce(expr))` first, and only go to path completion if it fails. I think that would be the optimal way because most cond clauses do run some variation of that anyway.

lib/iex/test/iex/autocomplete_test.exs
josevalim: Don’t we have a tmp directory for ExUnit yet? Or we do but it is in a separate location than the one the tag creates it?

lib/iex/lib/iex/autocomplete.ex
kelvinst: That way it might even make the opened string verification unnecessary, and we could just discard everything that's before the last `"`, removing all this: https://github.com/elixir-lang/elixir/blob/213d13e29dba087ea2d4b9f3e444e1aa8ac9661c/lib/iex/lib/iex/autocomplete.ex#L661-L673

lib/iex/test/iex/autocomplete_test.exs
josevalim: We already have this:

https://github.com/elixir-lang/elixir/blob/master/.gitignore#L4

can we configure it to use it instead?

lib/iex/lib/iex/autocomplete.ex
kelvinst: @josevalim I will test out this other way to implement, will let you know when I have something 👍

lib/iex/test/iex/autocomplete_test.exs
kelvinst: Actually, nevermind, it did create the `tmp` folder on the root cause I ran the test only for that particular file from the actual root, so everything is fine

josevalim:
Just a heads up. I am currently refactoring IEx.Autocomplete into a Code.autocomplete function. The goal is to decouple the detection of the autocompletion from the autocompletion and hopefully make the code more reusable. Once that is done, I will ask you to rebase your PR. :) Sorry for the inconvenience but it will be worth it. :)

kelvinst:
Oh, sure, no problem! Let me know when you have it.

kelvinst:
Just so you know, I also started the refactor I mentioned above, that checks `Code.string_to_quoted` in advance, but I feel like it would be possible to do a lot more with pattern matching that we actually are doing right now.

Also, maybe that's your intention with `Code.autocomplete`, but I suspect that we could also take advantage of `||` if some of the internal functions returned `nil` for when no completions apply, to then fallback to the other internal functions. I was planning to try that out Monday and merge the work here, but as you are doing your refactor I will wait your version of it.

josevalim:
@kelvinst if you want to follow it: https://github.com/elixir-lang/elixir/pull/10915

I don't think we will make it handle paths though - so the string splitting will have to be done before we call Code.autocomplete.

kelvinst:
> I don't think we will make it handle paths though - so the string splitting will have to be done before we call Code.autocomplete.

Makes sense, as path does not have much to do with `Code` module anyway

kelvinst:
Hey, I rebased my branch to master  (with the new `Code.cursor_context` function), and realized one thing: there is a slight change of behavior on autocompletion. For example, previously, if you type `"./d` and hit tab, it would not autocomplete to anything, because it would try to compile the whole expression and it would fail, but now, as it does not try to compile it, it suggests to complete it with vars or local functions...

kelvinst:
@josevalim let me know if that would be the new expected behavior ☝️Cause then I actually have to change the logic to check the path autocompletion first (I did refactor to check the code completion first and fallback to path completion if it doesn't suggest anything)

josevalim:
Yes, I think with the refactoring we should check for path autocompletion first.

The thing about using `"./` and `"/` as starts is really nice because those are not valid Elixir. So if they appear anywhere, it is most likely the beginning of a string with autocompletion, because it cannot be the end of a string as that would be invalid Elixir.

So for path autocompletion I would perhaps do something like this: start going back collecting all characters until you find a `"./` (relative path), `"/` (absolute path), `"` (ignore) or the expression is empty (ignore). I think this will be simple enough and enough to avoid false positives.

Make sure to add tests for things like: `Path.join("./foo", is_<TAB>`

lib/iex/test/iex/autocomplete_test.exs
```diff
@@ -399,4 +399,41 @@ defmodule IEx.AutocompleteTest do
     assert expand('take(') == {:yes, '', ['take(enumerable, amount)']}
     assert expand('derive(') == {:yes, '', ['derive(protocol, module, options \\\\ [])']}
   end
+
+  @tag :tmp_dir
+  test "path completion inside strings", %{tmp_dir: dir} do
+    File.cd!(dir)
```
josevalim, line 405: We can't use `File.cd!` on async: true as that is global. Furthermore, we should always revert back. If we need File.cd! for the tests, then we should move it to a separate module that does not have async: true.

lib/iex/lib/iex/autocomplete.ex
```diff
@@ -563,4 +570,95 @@ defmodule IEx.Autocomplete do
...
+        no()
+
+      [{unique, _}] ->
+        {:yes, String.to_charlist(unique), []}
+
+      list ->
+        {completions, filenames} = Enum.unzip(list)
+        common_prefix_size = :binary.longest_common_prefix(completions)
+        hint = completions |> hd() |> binary_part(0, common_prefix_size) |> String.to_charlist()
+        {:yes, hint, filenames}
```
josevalim, line 627: We already have code that does this, I believe it is called format_expansions or similar. We should just reuse that instead.

lib/iex/test/iex/autocomplete_test.exs
```diff
@@ -399,4 +399,41 @@ defmodule IEx.AutocompleteTest do
...
+    assert expand('"./dir/') == {:yes, 'file', ['file3', 'file4']}
+    assert expand('"./dir/#\{Str') == {:yes, '', ['Stream', 'String', 'StringIO']}
+
+    {:yes, [], list} = expand('"/')
+    assert is_list(list)
+
+    case :os.type() do
+      {:win32, _} ->
+        assert 'Users' in list
+        assert 'Windows' in list
```
josevalim, line 427: Instead of hardcoding, do `File.ls("/")` and check that instead.

josevalim:
> So for path autocompletion I would perhaps do something like this: start going back collecting all characters until you find a "./ (relative path), "/ (absolute path), " (ignore), #{ (ignore) or the expression is empty (ignore). Remember to deal with escaped quotes too. I think this will be simple enough and enough to avoid false positives.

Here is an actual example:

```elixir
  defp path_fragment([], _acc), do: :none
  defp path_fragment([?{, ?# | _rest], _acc), do: :none
  defp path_fragment([?", ?\\ | t], acc), do: collect_path_fragment(t, [?" | acc])
  defp path_fragment([?/, ?., ?" | _], acc), do: {:relative, acc}
  defp path_fragment([?/, ?" | _], acc), do: {:absolute, acc}
  defp path_fragment([?" | _], _acc), do: :none
  defp path_fragment([h | t], acc), do: collect_path_fragment(t, [h | acc])
```

kelvinst:
Done, I think it should be ready to merge now. Let me know if you need to me change something else. 😉

lib/iex/test/iex/autocomplete_test.exs
kelvinst: I went with actually completely removing `File.cd!`, as we don't actually need it to test the relative path completion. Looks like the tests are not passing on CI though, will investigate what's happening later today.

kelvinst:
Alright, I did make the tests a bit more OS agnostic, it made some of them pass, but I think we are still having problems with Windows tests... I don't have Windows here to test that, but hey, I'm thinking of doing a Bootcamp for gaming for a while, maybe it's time for some fun 😂

Can't promise though, if any of you have a Windows machine around that you can use to test the branch it would be cool

josevalim:

eksperimental:
One test is not passing on Windows.
https://github.com/elixir-lang/elixir/runs/2420388711#step:10:83

josevalim:
@kelvinst can you please take a look at the above? If you cannot, please let me know and I will revert until one of us can. Thank you!

kelvinst:
@josevalim doing it. Will open a new PR with the fixes

kelvinst:
@josevalim I asked before, maybe you missed that: should we then allow autocompletion for `"C:/<tab>` too?

josevalim:
We should, yes, where the letter can be upper or lowercase. :)

kelvinst:
Done on PR #10949

krisu05:
Hii whenever I typed iex -S mix in my terminal it's show invoked-expression can you please checked this
Screenshot 2024-02-02 175449

kelvinst:
> Hii whenever I typed iex -S mix in my terminal it's show invoked-expression can you please checked this
> 
> Screenshot 2024-02-02 175449
> 
> 

Hi @krisu05, seems like your issue is not related to this PR. Can you open another issue for it?

josevalim:
You need to use iex.ps1 on power shell because iex is an existing commandz

krisu05:
Screenshot 2024-02-12 181545
It shows again error

josevalim:
`bin\iex.ps1` or make sure that Elixir's `bin` directory is added to your [PATH](https://en.wikipedia.org/wiki/PATH_(variable)).

PR #10897, opened 2021-04-13
Path methods will always return UTF-8 encoded string

achempion:
In the [typespec documentation](https://hexdocs.pm/elixir/typespecs.html#the-string-type) advised to use `String.t()` for UTF-8 encoded strings and the `Path` methods will always return UTF-8 encoded strings.

### DIFF
--- a/lib/elixir/lib/path.ex
+++ b/lib/elixir/lib/path.ex
@@ -37,7 +37,7 @@ defmodule Path do
       #=> "D:/usr/local/../x"
 
   """
-  @spec absname(t) :: binary
+  @spec absname(t) :: String.t()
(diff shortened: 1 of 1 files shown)


josevalim:
That's not quite correct. :) If you give a binary to Path that is not utf-8 encoded, it will preserve it as is. We only act on some specific characters, keeping everything else as is. In any case, thank you!

achempion:
Should we remove this section from a `Path` module then?

> The functions in this module may receive a chardata as argument (i.e. a string or a list of characters / string) and will always return a string (encoded in UTF-8).

josevalim:
Yes, good catch. We should.

PR #10899, opened 2021-04-13
Improve Task docs

tedhexaflow:


lib/elixir/lib/task.ex
```diff
@@ -69,8 +69,8 @@ defmodule Task do
         {Task.Supervisor, name: MyApp.TaskSupervisor}
       ], strategy: :one_for_one)
 
-  And now you can use async/await once again passig the name of
-  the supervisor isntead of the pid:
+  And now you can use async/await by passig the name of
```
josevalim, line 72:
```suggestion
  And now you can use async/await by passing the name of
```

josevalim:

PR #10900, opened 2021-04-13
[WIP] Add callbacks for functions with default arguments defined inside protocols

eksperimental:
Previosly default arguments were accepted for function definition,
but the callbcaks for the arities with default arguments were not created.

Now it warns if the implementation is missing these.

### DIFF
--- a/lib/elixir/test/elixir/protocol_test.exs
+++ b/lib/elixir/test/elixir/protocol_test.exs
@@ -12,6 +12,9 @@ defmodule ProtocolTest do
       @deprecated "Reason"
       @spec ok(t) :: boolean
       def ok(term)
+
+      @spec ok_with_options(t, keyword) :: boolean
+      def ok_with_options(term, options \\ [])
     end
 
   @sample_binary sample_binary
@@ -70,6 +73,10 @@ defmodule ProtocolTest do
 
       def ok(struct) do
         Unknown.undefined(struct)
+      end
+
+      def ok_with_options(struct, options \\ []) do
+        Unknown.undefined({struct, options})
       end
     end
   end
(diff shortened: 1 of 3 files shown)


lib/elixir/test/elixir/protocol_test.exs
```diff
@@ -146,13 +153,19 @@ defmodule ProtocolTest do
     assert [{:type, 13, :fun, args}] = get_callbacks(@sample_binary, :ok, 1)
     assert args == [{:type, 13, :product, [{:user_type, 13, :t, []}]}, {:type, 13, :boolean, []}]
 
-    assert [{:type, 23, :fun, args}] = get_callbacks(@with_any_binary, :ok, 1)
-    assert args == [{:type, 23, :product, [{:user_type, 23, :t, []}]}, {:type, 23, :term, []}]
+    assert [{:type, 26, :fun, args}] = get_callbacks(@with_any_binary, :ok, 1)
+    assert args == [{:type, 26, :product, [{:user_type, 0, :t, []}]}, {:type, 26, :term, []}]
```
eksperimental, line 157: I don't understand here why `:user_type`  changed to `0`.

josevalim:
The current behavior is correct. The default arguments will always be expanded in the protocol call. For `def foo(bar, baz \\ [])`, the implementation never has to worry about implementing foo/1, because it will never be invoked. :)

josevalim:
However, it may be we have no tests for default arguments in test/elixir/protocol_test.exs, would you be so kind to double check and write an example if that's the case?

eksperimental:
This stems from my use case.
How do you make an implementation optional in a protocol?

josevalim:
You don’t, by design. You can do like the Enum module and allow users to return an error tuple.

eksperimental:
I was using `@optional_callbacks` and since alternative functions with default arguments are not defined as callbacks I was getting a compile error. Therefore this PR.

I think this should be explained in the docs to avoid users to fall into non-recomended and non-supported practices. The same case with defining `@callbacks` inside a protocol.

PR #10902, opened 2021-04-14
Improve URI docs & tests for IPv6 literals (#10896)

### LINKED ISSUE #10896: URI.parse/1 strips [] from IPv6 address literals in host field
### Environment

* `elixir --version: 1.11.4 Erlang/OTP 23 [erts-11.1.8] [source] [64-bit] [smp:32:32] [ds:32:32:10] [async-threads:1]  (compiled with Erlang/OTP 21)`

An IPv6 literal is the equivalent of `127.0.0.1`, e.g. `[fe80::]`. Per [RFC2732](https://tools.ietf.org/html/rfc2732) hostnames in URIs should be represented with the `[]` included.

Thanks @polymetis for the paired code spelunking to track this down.

### Current behavior

```
iex(1)> URI.parse("http://user:password@[fe80::]:1337/?foo=bar")
%URI{
  authority: "user:password@[fe80::]:1337",  <----- wrapped correctly
  fragment: nil,
  host: "fe80::", <----------- wrapping is missing
  path: "/",
  port: 1337,
  query: "foo=bar",
  scheme: "http",
  userinfo: "user:password"
}

## Erlang also does this

iex(2)> :uri_string.parse('http://user:password@[fe80::]:1337/?foo=bar')
%{
  host: 'fe80::', <------------ wrapping is missing
  path: '/',
  port: 1337,
  query: 'foo=bar',
  scheme: 'http',
  userinfo: 'user:password'
}
iex(3)> 
```

### Expected behavior

```
iex(4)>  URI.parse("http://[fe80::]/").host   
"[fe80::]"
```

### References

- [wikipedia](https://en.wikipedia.org/wiki/IPv6)  for a relatively digestable explanation

> Because IPv6 addresses contain colons, and HTTP uses colons as the port-address separator,
> RFC2732[39] specifies to use [square brackets] when using IPv6 addresses in URLs. e.g. > http://[2001:db8:4006:812::200e] (or, with a port, http://[2001:db8:4006:812::200e]:8080/path/page.html)

- normative ABNF  from [RFC3986 appendix A](https://tools.ietf.org/html/rfc3986#appendix-A) which updated and obsoleted RFC2732

- proposed deprecation of `URI.parse/1` in favour of Erlang's version https://github.com/elixir-lang/elixir/issues/10865

```
      IP-literal = "[" ( IPv6address / IPvFuture  ) "]"
```

- https://github.com/erlang/otp/issues/4731

### Background

When dealing with IPv6 only hostnames (servers that *only* have ip6 `AAAA` records in DNS), such as https://ipv6.google.com/ the impact is that downstream libraries try to treat the `IPv6 literal address` as a hostname, which obviously simply doesn't work. This fails in several places, for example, HTTP adapters used in Tesla, when the `userinfo` part of the URI struct is passed in as a header, and the ipv6 address is then used to reconstruct the URI from hostname, or any other `gen_tcp` like functionality.

### Options

- fix upstream `uri_string:parse/1` in Erlang because of proposed Elixir deprecation
- don't fix, but add an additional function to URI module, that can strip `userinfo` from `authority`
- do nothing to avoid Breaking The Internet

I'm happy to PR either case.

dch:
The URI module is not quite in compliance with the canonical RFC3986,
when handling IPv6 literal hostnames. These should be stored with
`[]` included. This is handled correctly in the authority field, but
not in the hostname field.

We can't change this behaviour now, so add doc & test.

closes #10896

```
$ bin/elixir lib/elixir/test/elixir/uri_test.exs
Excluding tags: [windows: true]

........................................................

Finished in 1.2 seconds (1.1s on load, 0.04s async, 0.00s sync)
32 doctests, 24 tests, 0 failures

Randomized with seed 555840
```

NB `gmake docs` fails here, but I think this is a previous commit. Using elixir /master/:

```
gmake docs
==> ex_doc (elixir)
CANONICAL=master/ bin/elixir ../ex_doc/bin/ex_doc "Elixir" "1.13.0-dev" "lib/elixir/ebin" --main "Kernel" --source-url "https://github.com/elixir-lang/elixir" --source-ref "2d2cf356cbcaf2607f8f7f356b42def43ad7540f"  --output doc/elixir --canonical "https://hexdocs.pm/elixir/master/" --homepage-url "https://elixir-lang.org/docs.html" --formatter "html" --config "lib/elixir/docs.exs"
** (FunctionClauseError) no function clause matching in Enumerable.Range.slice/1    
    
    The following arguments were given to Enumerable.Range.slice/1:
    
        # 1
        %Inspect.Error{message: "got FunctionClauseError with message \"no function clause matching in Inspect.Range.inspect/2\" while inspecting %{__struct__: Range, first: 48, last: 57}"}
    
        def slice(first.._//step = range)
    
    (elixir 1.13.0-dev) lib/range.ex:315: Enumerable.Range.slice/1
    (elixir 1.13.0-dev) lib/enum.ex:2251: Enum.random/1
    (elixir 1.13.0-dev) lib/enum.ex:1557: anonymous fn/3 in Enum.map/2
    (elixir 1.13.0-dev) lib/enum.ex:3898: Enum.reduce_range/5
    (elixir 1.13.0-dev) lib/enum.ex:2360: Enum.map/2
    (makeup 1.0.5) lib/makeup/lexer/groups.ex:327: Makeup.Lexer.Groups.random_prefix/1
    (makeup_elixir 0.15.0) lib/makeup/lexers/elixir_lexer.ex:596: Makeup.Lexers.ElixirLexer.lex/2
    (makeup 1.0.5) lib/makeup.ex:37: Makeup.highlight_inner_html/2
gmake: *** [Makefile:190: docs_elixir] Error 1
```

### DIFF
--- a/lib/elixir/lib/uri.ex
+++ b/lib/elixir/lib/uri.ex
@@ -477,6 +477,11 @@ defmodule URI do
   When a URI is given without a port, the value returned by
   `URI.default_port/1` for the URI's scheme is used for the `:port` field.
 
+  When a URI hostname is an IPv6 literal, it has the `[]` unwrapped before
+  being stored in the `:host` field. This doesn't match the formal grammar
+  for hostnames, which requires the `[]` to be preserved, according to
+  [RFC 3986, appendix A](https://tools.ietf.org/html/rfc3986#appendix-A)
+
   If a `%URI{}` struct is given to this function, this function returns it
   unmodified.
 
@@ -530,6 +535,17 @@ defmodule URI do
         userinfo: nil
       }
 
+      iex> URI.parse("//[fe80::]/")
+      %URI{
+        authority: "[fe80::]",
+        fragment: nil,
+        host: "fe80::",
+        path: "/",
+        port: nil,
+        query: nil,
+        scheme: nil,
+        userinfo: nil
+      }
   """
   @spec parse(t | binary) :: t
   def parse(%URI{} = uri), do: uri
@@ -588,6 +604,9 @@ defmodule URI do
   defp nillify_query(_other), do: nil
 
   # Split an authority into its userinfo, host and port parts.
+  #
+  # Note that the host field is returned *without* [] even if, according to
+  # RFC3986 grammar, a native IPv6 address requires them.
   defp split_authority("") do
     {nil, nil, nil, nil}
   end
--- a/lib/elixir/test/elixir/uri_test.exs
+++ b/lib/elixir/test/elixir/uri_test.exs
@@ -285,6 +285,8 @@ defmodule URITest do
         "1080::8:800:200C:417A",
         # multicast
         "FF01::101",
+        # link-local
+        "fe80::",
         # abbreviated
         "2607:f3f0:2:0:216:3cff:fef0:174a",
         # mixed hex case


josevalim:

josevalim:
Btw, the docs failure is because you need to do into your ex_doc checkout and call `mix compile`.

PR #10903, opened 2021-04-14
Improve docs for using Application.get_env/3

nickjj:
Hi,

I found myself wanting to access a specific attribute of a configuration option such as `Application.get_env(:hello, HelloWeb.Endpoint)[:signing_salt]` but the documentation didn't cover this syntax.

It took around half an hour of Googling, scouring the docs and ultimately finding a StackOverflow answer to get the correct answer. I think this extra context on how I arrived at a solution was necessary to include in this PR comment because the SO thread at https://stackoverflow.com/questions/35725536/how-to-read-config-variable-in-phoenix-elixir shows a bunch of upvotes and a marked answer for an incorrect solution.

So I figured I'd open a PR here to clarify the docs to demonstrate this usage example.

I broke up the existing paragraph into 2 paragraphs because it wrapped to 6 lines of text in the HTML form of the docs and started to look a bit cluttered.

### DIFF
--- a/lib/elixir/lib/application.ex
+++ b/lib/elixir/lib/application.ex
@@ -611,13 +611,18 @@ defmodule Application do
 
   Our database engine used by `:my_app` needs to know what databases exist, and
   what the database configurations are. The database engine can make a call to
-  `get_env(:my_app, :my_app_databases)` to retrieve the list of databases (specified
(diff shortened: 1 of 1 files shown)


josevalim:

PR #10909, opened 2021-04-16
Raise if local dep is unavailable while compiling

fertapric:


ericmj (approved):

lib/mix/lib/mix/tasks/deps.compile.ex
```diff
@@ -59,7 +59,7 @@ defmodule Mix.Tasks.Deps.Compile do
         # Because this command may be invoked explicitly with
         # deps.compile, we simply try to compile any available
         # dependency.
```
fertapric, line 61:
```suggestion
        # or local dependency.
```

josevalim (approved):

PR #10912, opened 2021-04-16
Warn if @callback, @macrocallback and @optional_callbacks defined inside protocol

### LINKED ISSUE #10901: Warn if @optional_callbacks is set inside protocol

eksperimental:
Closes #10901

WIP because we need to improve the warning messages.

```elixir
defprotocol MyProtocol do
  @spec with_specs(any(), keyword()) :: tuple()
  def with_specs(term, options \\ [])

  def without_specs(term, options \\ [])

  @callback foo(term) :: {:ok, term}
  @callback foo(term, 1) :: {:ok, term}

  @macrocallback bar(term) :: {:ok, term}

  @optional_callbacks [foo: 1, with_specs: 2]

  @optional_callbacks [without_specs: 2, foo: 2]
end
```

```console
$ mix
Compiling 2 files (.ex)
warning: callback foo/2 defined inside protocol, which is discourage; please define it with the def/2.
  lib/my_protocol.ex:1: MyProtocol (module)

warning: callback foo/1 defined inside protocol, which is discourage; please define it with the def/2.
  lib/my_protocol.ex:1: MyProtocol (module)

warning: macrocallback bar/1 defined inside protocol, which is discourage; please define it with the def/2.
  lib/my_protocol.ex:1: MyProtocol (module)

warning: @optional_callbacks defined inside protocol, please do something else...
  lib/my_protocol.ex:1: MyProtocol (module)

Generated protocol_callbacks app
```

### DIFF
--- a/lib/elixir/lib/protocol.ex
+++ b/lib/elixir/lib/protocol.ex
@@ -681,6 +681,8 @@ defmodule Protocol do
   def __protocol__(name, do: block) do
     quote do
       defmodule unquote(name) do
+        @before_compile Protocol
+
         # We don't allow function definition inside protocols
         import Kernel,
           except: [
@@ -712,6 +714,66 @@ defmodule Protocol do
 
         # Finalize expansion
         unquote(after_defprotocol())
+      end
+    end
+  end
+
+  def __callback_ast_to_fa__({kind, {:"::", _, [{name, _, args}, _return]}, _pos})
+      when kind in [:callback, :macrocallback] do
+    {name, length(args)}
+  end
+
+  def __warn__(message, env) do
+    IO.warn(message, Macro.Env.stacktrace(env))
+  end
+
+  # TODO: Convert the following warnings into errors future Elixir versions
+  defmacro __before_compile__(env) do
+    quote bind_quoted: [env: Macro.escape(env)] do
+      # Callbacks
+      callbacks =
+        :lists.map(
+          &Protocol.__callback_ast_to_fa__/1,
+          Module.get_attribute(__MODULE__, :callback)
+        )
+
+      functions = Module.get_attribute(__MODULE__, :functions)
+
+      :lists.map(
+        fn {name, arity} ->
+          Protocol.__warn__(
+            "cannot define @callback #{name}/#{arity} inside protocol, use def/1 to outline your protocol definition",
+            env
+          )
+        end,
+        callbacks -- functions
+      )
+
+      # Macro Callbacks
+      macrocallbacks =
+        :lists.map(
+          &Protocol.__callback_ast_to_fa__/1,
+          Module.get_attribute(__MODULE__, :macrocallback)
+        )
+
+      :lists.map(
+        fn {name, arity} ->
+          Protocol.__warn__(
+            "cannot define @macrocallback #{name}/#{arity} inside protocol, use def/1 to outline your protocol definition",
+            env
+          )
+        end,
+        macrocallbacks
+      )
+
+      # Optional Callbacks
+      optional_callbacks = Module.get_attribute(__MODULE__, :optional_callbacks)
+
+      if length(optional_callbacks) > 0 do
+        Protocol.__warn__(
+          "cannot define @optional_callbacks inside protocol, all of the protocol definitions are required",
+          env
+        )
       end
     end
   end
--- a/lib/elixir/lib/protocol.ex
+++ b/lib/elixir/lib/protocol.ex
@@ -718,12 +718,12 @@ defmodule Protocol do
     end
   end
 
-  def __callback_ast_to_fa__({kind, {:"::", _, [{name, _, args}, _return]}, _pos})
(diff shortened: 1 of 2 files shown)


lib/elixir/lib/protocol.ex
```diff
@@ -290,6 +290,8 @@ defmodule Protocol do
       # otherwise generate a dummy callback
       Module.spec_to_callback(__MODULE__, {name, arity}) ||
         @callback unquote(name)(unquote_splicing(type_args)) :: term
+
+      @__generated_callbacks__ {name, arity}
```
josevalim, line 294: This is already stored in `@functions`, no?

lib/elixir/lib/protocol.ex
```diff
@@ -681,6 +683,10 @@ defmodule Protocol do
   def __protocol__(name, do: block) do
     quote do
       defmodule unquote(name) do
+        @before_compile Protocol.Util
```
josevalim, line 686: Please define the callback in the same module, we don't need a new one. :)

josevalim:
Error messages:

```
warning: cannot define @macrocallback bar/1 inside protocol, use def/1 to outline your protocol definition
  lib/my_protocol.ex:1: MyProtocol (module)
```

```
warning: cannot define @optional_callbacks inside protocol, all of protocol definitions are required
  lib/my_protocol.ex:1: MyProtocol (module)
```

Please leave a TODO so we make those errors in future Elixir versions.

lib/elixir/lib/protocol.ex
eksperimental: Doesn't beore_compile callback impls need to be outside a module since it will not be available before compile such module is compiled?

lib/elixir/lib/protocol.ex
eksperimental: true

lib/elixir/lib/protocol.ex
eksperimental: Nevermind. I see what you mean. This is inside a macro that outputs "defmodule", I will move it inside

lib/elixir/test/elixir/kernel/warning_test.exs
```diff
@@ -1150,6 +1150,64 @@ defmodule Kernel.WarningTest do
...
+      assert message =~
+               "cannot define @macrocallback bar/2 inside protocol, use def/1 to outline your protocol definition\n  nofile:1"
+
+      assert message =~
+               "cannot define @optional_callbacks inside protocol, all of the protocol definitions are required\n  nofile:1"
+    after
+      purge([SampleWithCallbacks])
+    end
+
+    test "does not warn when callbacks are not defined" do
```
josevalim, line 1194: We don't need this test. Otherwise all other protocols in Elixir and in the test suite would warn.

eksperimental: It will be obvious to the eye due to the warnings, but "make test" will pass
I have added an element to the macrocallback list and I got no so many warnings, and the test pass

eksperimental: Anyway, I will delete it.

eksperimental:
Rebased and good to ship

lib/elixir/lib/protocol.ex
```diff
@@ -716,6 +718,66 @@ defmodule Protocol do
...
+      when kind in [:callback, :macrocallback] do
+    {name, length(args)}
+  end
+
+  def __warn__(message, env) do
+    IO.warn(message, Macro.Env.stacktrace(env))
+  end
+
+  # TODO: Convert the following warnings into errors future Elixir versions
+  defmacro __before_compile__(env) do
```
josevalim, line 731: This should be a function. I believe there is no reason to inject this code into the module.

eksperimental: Done.

lib/elixir/lib/protocol.ex
```diff
@@ -716,6 +718,64 @@ defmodule Protocol do
     end
   end
 
+  def __callback_ast_to_fa__({kind, {:"::", _, [{name, _, args}, _return]}, _pos})
```
josevalim, line 721: Those should be private now.

lib/elixir/lib/protocol.ex
```diff
@@ -716,6 +718,64 @@ defmodule Protocol do
...
+      when kind in [:callback, :macrocallback] do
+    {name, length(args)}
+  end
+
+  def __warn__(message, env) do
```
josevalim, line 726: This one too!

lib/elixir/lib/protocol.ex
eksperimental: oops

lib/elixir/lib/protocol.ex
eksperimental: done and reviewed the whole code again

josevalim:

PR #10916, opened 2021-04-17
Remove @doc false, from functions and macros starting with underscore

eksperimental:
As they are unnecesary


josevalim:
I like them for explicitness but thanks!

PR #10917, opened 2021-04-17
Simplify @functions definition inside defprotocol

eksperimental:


josevalim:
The current definition is fine as is, thanks though!

PR #10922, opened 2021-04-18
Mix.install/2: Add :elixir option

wojtekmach:
This is especially useful for scripts that depend on features from libraries that are only available on specific Elixir versions.

### DIFF
--- a/lib/mix/lib/mix.ex
+++ b/lib/mix/lib/mix.ex
@@ -498,6 +498,9 @@ defmodule Mix do
     * `:consolidate_protocols` - if `true`, runs protocol
       consolidation via the `mix compile.protocols` task (Default: `true`)
 
+    * `:elixir` - if set, ensures the current Elixir version matches the given
+      version requirement (Default: `nil`)
+
   ## Examples
 
       Mix.install([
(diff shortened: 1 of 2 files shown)


lib/mix/lib/mix.ex
```diff
@@ -516,6 +519,16 @@ defmodule Mix do
       Mix.raise("Mix.install/2 cannot be used inside a Mix project")
     end
 
+    elixir_requirement = opts[:elixir]
+    elixir_version = System.version()
+
+    if !!elixir_requirement and not Version.match?(elixir_version, elixir_requirement) do
```
josevalim, line 525: In theory we can simply pass this option to the underlying Mix project... but I guess it is best to be clear?

josevalim, line 525: In theory we can simply pass this option to the underlying Mix project... but I guess it is best to be clear?

josevalim:

lib/mix/lib/mix.ex
wojtekmach: Right, i just thought that a custom error message is better than the one coming from the mix project

PR #10923, opened 2021-04-19
Support autolinks in IO.ANSI docs

eksperimental:
This will deal with Markdown autolinks (URLs surrounded by <..>),
supported by almost all Markdown parsers,
so links are correctly displayed in IEx.

Reference:
https://spec.commonmark.org/0.29/#autolinks


josevalim:
Thanks but I would prefer to keep it as is. In a lot of places, IO.ANSI doesn't touch the markup at all and, given the lack of other visual indicators in the terminal, I would prefer to keep <...>.

PR #10924, opened 2021-04-19
Use proper Unicode reference

eksperimental:
As described in https://unicode.org/versions/Unicode13.0.0/#Version_Specification

### DIFF
--- a/lib/elixir/lib/string.ex
+++ b/lib/elixir/lib/string.ex
@@ -68,8 +68,9 @@ defmodule String do
 
   ## Code points and grapheme cluster
 
-  The functions in this module act according to the Unicode
-  Standard, version 13.0.0.
+  The functions in this module act according to
+  **The Unicode Standard, Version 13.0.0**.
+  <http://www.unicode.org/versions/Unicode13.0.0/>
 
   As per the standard, a code point is a single Unicode Character,
   which may be represented by one or more bytes.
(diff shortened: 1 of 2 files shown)


lib/elixir/lib/string.ex
```diff
@@ -68,8 +68,9 @@ defmodule String do
 
   ## Code points and grapheme cluster
 
-  The functions in this module act according to the Unicode
-  Standard, version 13.0.0.
+  The functions in this module act according to
+  **The Unicode Standard, Version 13.0.0**.
+  <http://www.unicode.org/versions/Unicode13.0.0/>
```
josevalim, line 73: I believe the context of that note is in relation to academic or official references, but it doesn't hurt to be clearer.

```suggestion
  [The Unicode Standard, Version 13.0.0](http://www.unicode.org/versions/Unicode13.0.0/).
```

lib/elixir/unicode/unicode.ex
```diff
@@ -8,7 +8,7 @@
 # 4. Replace GraphemeBreakTest.txt by copying original auxiliary/GraphemeBreakTest.txt
 # 5. Replace GraphemeBreakProperty.txt by copying original auxiliary/GraphemeBreakProperty.txt
 # 6. Append Extended_Pictographic from emoji/emoji-data.txt to the end of GraphemeBreakProperty.txt
-# 7. Update String.Unicode.version/0 and on String module docs
+# 7. Update String.Unicode.version/0 and on String module update version and link to the standard.
```
josevalim, line 11:
```suggestion
# 7. Update String.Unicode.version/0 and on String module docs (version and link)
```

josevalim:

PR #10926, opened 2021-04-19
Add Module.reserved_attributes/0

wojtekmach:
Worth mentioning that we have also a few more standard attributes:

  - `@tag` for ExUnit
  - `@shortdoc` for Mix

but since they are not part of Elixir, they are not included in this list.

cc @msaraiva

### DIFF
--- a/lib/elixir/lib/module.ex
+++ b/lib/elixir/lib/module.ex
@@ -335,6 +335,16 @@ defmodule Module do
         @vsn "1.0"
       end
 
+  ### Struct attributes
+
+    * `@derive` - derives an implementation for the given protocol for the
+      struct defined in the current module
+
+    * `@enforce_keys` - ensures the given keys are always set when building
+      the struct defined in the current module
+
+  See `Kernel.defstruct/1` for more information on building and using structs.
+
   ### Typespec attributes
 
   The following attributes are part of typespecs and are also built-in in
@@ -630,10 +640,12 @@ defmodule Module do
         doc: "Specifies which behaviour callbacks and macro behaviour callbacks are optional."
       },
       derive: %{
-        doc: "Derive the `Any` implementation of the given protocol."
+        doc:
+          "Derives an implementation for the given protocol for the struct defined in the current module."
       },
       enforce_keys: %{
-        doc: "Ensures the given keys are always set when building the struct."
+        doc:
+          "Ensures the given keys are always set when building the struct defined in the current module."
       }
     }
   end
--- a/lib/elixir/lib/module.ex
+++ b/lib/elixir/lib/module.ex
@@ -628,6 +628,12 @@ defmodule Module do
       },
       optional_callbacks: %{
         doc: "Specifies which behaviour callbacks and macro behaviour callbacks are optional."
+      },
+      derive: %{
+        doc: "Derive the `Any` implementation of the given protocol."
+      },
+      enforce_keys: %{
+        doc: "Ensures the given keys are always set when building the struct."
       }
     }
   end
--- a/lib/elixir/lib/module.ex
+++ b/lib/elixir/lib/module.ex
@@ -335,15 +335,15 @@ defmodule Module do
         @vsn "1.0"
       end
 
-  ### `@derive`
+  ### Struct attributes
 
-  Derives an implementation for the given protocol for the struct defined in
-  the current module. See `Kernel.defstruct/1 for more information.
+    * `@derive` - derives an implementation for the given protocol for the
+      struct defined in the current module
 
-  ### `@enforce_keys`
+    * `@enforce_keys` - ensures the given keys are always set when building
+      the struct defined in the current module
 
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/module.ex
```diff
@@ -543,6 +543,95 @@ defmodule Module do
...
 
+  @doc """
+  Returns information about module attributes used by Elixir.
+
+  See the "Module attributes" section in the module documentation for more
+  information on each attribute.
+
+  ## Examples
+
+      iex> map = Module.builtin_attributes()
```
eksperimental, line 554:
```suggestion
      iex> map = Module.reserved_attributes()
```

eksperimental (changes requested):

eksperimental:
> What about `fallback_to_any`, `derive`, `enforce_keys`, `vsn`? There are probably a few other reserved ones. One i can think of is `__struct__`

I came to comment on `__struct__`, but since it starts with an underscore we may not want to list it.

wojtekmach:
> What about fallback_to_any, derive, enforce_keys, vsn?

Good catches, thanks. These are specific to protocols & structs but perhaps they should be listed there as well. I listed just the ones that I found in `Module` documentation.

josevalim:
I think we should add @enforce_keys and @derive for sure because it applies to any module. Not sure about @fallback_to_any. Definitely a no for __struct__ as it is private.

eksperimental:
Protocols use the `@functions` attribute

eksperimental:
> Protocols use the `@functions` attribute

Since this is undocumented (AFAIK) we could make it private, and name it `@__functions__`

wojtekmach:
Just had an idea how we could do the same thing for ExUnit `@tag` etc, how about the following?

```elixir
defmodule Module do
  def reserved_attributes(module) do
    if function_exported?(...), do: module.__reserved_attributes__(), else: %{}
  end

  def reserved_attributes() do
    %{moduledoc: ..., doc: ..., ...}
  end
end

defmodule ExUnit.Case do
  def __reserved_attributes__() do
    %{tag: ..., moduletag: ..., describetag: ...}
  end
end

defmodule Ecto.Schema do
  def __reserved_attributes__() do
    %{primary_key: ..., timestamps_opts: ..., ...}
  end
end
```

and then the tooling upon seeing `use mod` would call `Module.reserved_attributes(mod)`.

Not sure if coupling to `__using__` like that is a good idea though, if we don't want to define `@enforce_keys` "globally", using this mechanism I'd put it on `Protocol.__reserved_attributes__()` but how one would automatically infer it, special case `defimpl`?

josevalim:
@eksperimental i was talking to @msaraiva from Elixirsense and we agreed that all private attributes should start with an underscore. Maybe `@__functions__` is too much but standardizing on `@_functions` would be fine? Do you want to send a PR for that?

I don’t think we want to make this customizable for now though. So the current definition plus the structs ones are fine by me.

wojtekmach:
> I don’t think we want to make this customizable for now though

fair enough. if we'd go with that `reserved_attributes` idea, I suppose we could add a `reserved_attributes(module \\ Module)` in a backwards compatible way :)

lib/elixir/lib/module.ex
```diff
@@ -543,6 +543,101 @@ defmodule Module do
   @callback __info__(:md5) :: binary()
   @callback __info__(:module) :: module()
 
+  @doc """
+  Returns information about module attributes used by Elixir.
+
+  See the "Module attributes" section in the module documentation for more
+  information on each attribute.
```
wojtekmach, line 550: I suppose this sentence is no longer true for `@derive` and `@enforce_keys`, how should we reword it? Or just remove it?

josevalim: Let’s add a quick sentence at the top on derive and enforce keys and point people to structs. :)

eksperimental:
> @eksperimental i was talking to @msaraiva from Elixirsense and we agreed that all private attributes should start with an underscore. Maybe `@__functions__` is too much but standardizing on `@_functions` would be fine? Do you want to send a PR for that?
> 
> I don’t think we want to make this customizable for now though. So the current definition plus the structs ones are fine by me.

I'm fine with it.
Sure I can contribute a PR

lib/elixir/lib/module.ex
```diff
@@ -543,6 +553,103 @@ defmodule Module do
...
+  ## Examples
+
+      iex> map = Module.reserved_attributes()
+      iex> Map.has_key?(map, :moduledoc)
+      true
+      iex> Map.has_key?(map, :doc)
+      true
+
+  """
+  @doc since: "1.13.0"
```
wojtekmach, line 571: @josevalim in case you want to backport it to v1.12 :)

```suggestion
  @doc since: "1.12.0"
```

josevalim:

eksperimental:
For the records, we also have `@protocol` and `@for` that only apply to protocol implementations.

PR #10927, opened 2021-04-19
Fix case when callback spec uses when clause

eksperimental:
This was causing an error:
** (FunctionClauseError) no function clause matching in Protocol.callback_ast_to_fa/1

Introduced in PR: #10912, commit: 485cc13946e5c24dd4c31ebb03e85876c33fa468

### DIFF
--- a/lib/elixir/lib/protocol.ex
+++ b/lib/elixir/lib/protocol.ex
@@ -723,6 +723,13 @@ defmodule Protocol do
     {name, length(args)}
   end
 
+  defp callback_ast_to_fa(
+         {kind, {:when, _, [{:"::", _, [{name, _, args}, _return]}, _vars]}, _pos}
+       )
+       when kind in [:callback, :macrocallback] do
+    {name, length(args)}
+  end
+
   defp warn(message, env) do
     IO.warn(message, Macro.Env.stacktrace(env))
   end
--- a/lib/elixir/test/elixir/kernel/warning_test.exs
+++ b/lib/elixir/test/elixir/kernel/warning_test.exs
@@ -1158,10 +1158,16 @@ defmodule Kernel.WarningTest do
             @spec with_specs(any(), keyword()) :: tuple()
             def with_specs(term, options \\ [])
 
+            @spec with_specs_and_when(any(), opts) :: tuple() when opts: keyword
+            def with_specs_and_when(term, options \\ [])
+
             def without_specs(term, options \\ [])
 
             @callback foo(term) :: {:ok, term}
             @callback foo(term, keyword) :: {:ok, term, keyword}
+
+            @callback foo_when(x) :: {:ok, x} when x: term
+            @callback foo_when(x, opts) :: {:ok, x, opts} when x: term, opts: keyword
 
             @macrocallback bar(term) :: {:ok, term}
             @macrocallback bar(term, keyword) :: {:ok, term, keyword}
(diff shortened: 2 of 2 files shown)


eksperimental:
@josevalim please let me know if you can think of any other case where the callback AST will not pattern match

eksperimental:
Trying to break things, I have come up with another issue.

```elixir
defprotocol AProtocol do
  def foo(term)
  @callback bar(term)
end
```

The code above will now raise:
```
== Compilation error in file lib/a_protocol.ex ==
** (FunctionClauseError) no function clause matching in Protocol.callback_ast_to_fa/1    
    
    The following arguments were given to Protocol.callback_ast_to_fa/1:
    
        # 1
        {:callback, {:bar, [line: 4], [{:term, [line: 4], nil}]}, {AProtocol, {4, 1}}}
    
        defp callback_ast_to_fa({kind, {:"::", meta, [{name, _, args}, _return]}, _pos}) when kind === :callback or kind === :macrocallback
        defp callback_ast_to_fa({kind, {:when, _, [{:"::", meta, [{name, _, args}, _return]}, _vars]}, _pos}) when kind === :callback or kind === :macrocallback
    
    (elixir 1.13.0-dev) lib/protocol.ex:721: Protocol.callback_ast_to_fa/1
    (stdlib 3.13.2) lists.erl:1243: :lists.map/2
    (elixir 1.13.0-dev) lib/protocol.ex:735: Protocol.callback_metas/2
    (elixir 1.13.0-dev) lib/protocol.ex:754: Protocol.__before_compile__/1
    (stdlib 3.13.2) lists.erl:1267: :lists.foldl/3
```

while before this new feature it will raise.:
```
== Compilation error in file lib/a_protocol.ex ==
** (CompileError) lib/a_protocol.ex:4: type specification missing return type: bar(term)
    (elixir 1.11.4) lib/kernel/typespec.ex:911: Kernel.Typespec.compile_error/2
    (stdlib 3.13.2) lists.erl:1358: :lists.mapfoldl/3
    (stdlib 3.13.2) lists.erl:1359: :lists.mapfoldl/3
    (elixir 1.11.4) lib/kernel/typespec.ex:237: Kernel.Typespec.translate_typespecs_for_module/2
    (elixir 1.11.4) src/elixir_erl_compiler.erl:12: anonymous fn/3 in :elixir_erl_compiler.spawn/2
```

should we create a catch-all clause for `Protocol.callback_ast_to_fa/1` or is there any better approach?

eksperimental:
> Right, the callback above is invalid but it has not been validated yet, so a catch-all sounds good.

Could you please have a look at the last two commits of  this branch
https://github.com/eksperimental/elixir/commits/protocol_callback_warning_catch_all

The first one implements the catch-all clause, and the second one reraises FunctionClauseError. I am worried that by using a catch-all, we will shadow any case we haven't thought of or if the syntax is updated in the future this will go unnoticed. Let me know first, if it is good practice to do it this way and whether is overkill.
Thank you.

josevalim:
@eksperimental why would we shadow? Any callback that we register should still be handled by the underlying compiler and raise later on. So having an invalid callback should only skip the warning, not its validation later on.

eksperimental:
> @eksperimental why would we shadow? Any callback that we register should still be handled by the underlying compiler and raise later on. So having an invalid callback should only skip the warning, not its validation later on.

When I referer to shadowing, I don't mean the compiler error, but the Warning due to using a callback inside a protocol.

josevalim:
We don't  need to emit a warning in this case but later on the code won't even compile anyway.

eksperimental:
> We don't need to emit a warning in this case but later on the code won't even compile anyway.

Had I just put a catch-all clause in the original implementation, I would have never found out we were not cover the case that this very same PR fixes, as it would pass silently.

josevalim:
And now the code has a bug where it doesn't compile. If we are using prior bugs as justifications, then the current bug is enough reason to justify the catch-all. :)

eksperimental:
> And now the code has a bug where it doesn't compile. If we are using prior bugs as justifications, then the current bug is enough reason to justify the catch-all. :)

Sorry José, but I am not following you here.
The PR that I link above impelments the catch-all clause, and additionally (second commit) it raises after the compilation is completed.

josevalim:
Sorry for being unclear too :) the PR above is doing too much. We don’t need to re-raise. Just do a catch-all and let the compiler fail on invalid callback.

eksperimental:
Al good. Thank you

PR #10929, opened 2021-04-20
Add line number in @callback and @macrocallback warning when inside Protocol

eksperimental:
Before we were just showing the line number for the defprotocol call:
warning: cannot define @callback default/1 inside protocol, use def/1 to outline your protocol definition
  lib/buildable.ex:1: Buildable (module)

Now:
warning: cannot define @callback default/1 inside protocol, use def/1 to outline your protocol definition
  lib/buildable.ex:82: Buildable (module)

For @optional_callbacks it still displays the defprotocol line number.

### DIFF
--- a/lib/elixir/lib/protocol.ex
+++ b/lib/elixir/lib/protocol.ex
@@ -718,55 +718,69 @@ defmodule Protocol do
     end
   end
 
-  defp callback_ast_to_fa({kind, {:"::", _, [{name, _, args}, _return]}, _pos})
+  defp callback_ast_to_fa({kind, {:"::", meta, [{name, _, args}, _return]}, _pos})
        when kind in [:callback, :macrocallback] do
-    {name, length(args)}
+    {name, length(args), meta}
   end
 
   defp callback_ast_to_fa(
-         {kind, {:when, _, [{:"::", _, [{name, _, args}, _return]}, _vars]}, _pos}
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/protocol.ex
```diff
@@ -718,55 +718,69 @@ defmodule Protocol do
...
+  defp fa_to_atom(function, arity) do
+    (:erlang.atom_to_list(function) ++ [?_ | :erlang.integer_to_list(arity)])
+    |> :erlang.list_to_atom()
+  end
+
+  defp get_callbacks_and_meta(module, kind) when kind in [:callback, :macrocallback] do
+    :lists.foldl(
+      fn elem, {fa_acc, meta_acc} ->
+        {function, arity, meta} = callback_ast_to_fa(elem)
+        {[{function, arity} | fa_acc], [{fa_to_atom(function, arity), meta} | meta_acc]}
```
josevalim, line 742: Suggestion: there is no need to precompute the meta. Most times we won't emit warnings, so we are doing unnecessary work. Instead, retrieve the line and the meta only when and if we are going to emit a warning.

eksperimental: Thank you @josevalim 
I have taken a much simpler approach. 
Please have a look again

josevalim:

PR #10933, opened 2021-04-20
Clarify mix local.hex documentation when both options are set

colinsmetz:
I believe the documentation for the `mix local.hex` task is incorrect, or at least confusing.

It is written that:

> If both options are set, --force takes precedence.

I understand it as "if --force is set, then --if-missing is not considered" (the installation is forced whether Hex is already installed or not). In practice, it is not true, `--force` only bypasses the shell prompt. If it is not what the sentence means, then I don't understand it.

I suggest a simple rewrite of that sentence to match the current implementation and avoid ambiguity.


josevalim:

PR #5488, opened 2016-11-28
Improve IEx autocomplete to support navigating map atom keys

myronmarston:
Eventually we may want to take this a step further and support autocompletion of the variable names themselves but I don't have time today to look into that and that would probably be a separate PR.

I noticed that 1.4.0.rc-0 was just released...any chance this could be included in 1.4?  This is a feature I've long wanted and I'd be sad to see it pushed out months :(.

PR #10938, opened 2021-04-21
Clean up unreachable code from IEx.Autocomplete

michallepicki:
Dialyzer found some fresh unreachable code. Code was added to support map field access autocompletion in https://github.com/elixir-lang/elixir/pull/5488 and made unreachable recently in https://github.com/elixir-lang/elixir/pull/10915/files#diff-9e9de7d847547ae049e1297f35e4f7cd0f7d99d6f4cbcf8fa4b90c7c532ee035L283

I checked and map field access seems to be still working correctly with the code removed :)

### DIFF
--- a/lib/iex/lib/iex/autocomplete.ex
+++ b/lib/iex/lib/iex/autocomplete.ex
@@ -152,7 +152,7 @@ defmodule IEx.Autocomplete do
   end
 
   defp expand_dot_path({:var, var}, shell) do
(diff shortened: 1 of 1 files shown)


josevalim:
Oh, good find. This is ditching some optimization, I need to add it back. :)

michallepicki:
Right, feel free to close without merging if that's better!

PR #10942, opened 2021-04-22
Improve readability: Fix minor detail Enumerable doc

eksperimental:
Convert the code block into an inline, so it is part of the paragraph.
Before the paragraph was split by the codeblock.

Before:
image

After: 
image


josevalim:

PR #10947, opened 2021-04-23
Document in Collectable docs the new range format

eksperimental:
It mentions `step` and link to the various types in Range

### DIFF
--- a/lib/elixir/lib/collectable.ex
+++ b/lib/elixir/lib/collectable.ex
@@ -17,8 +17,8 @@ defprotocol Collectable do
 
   This design is intentional. `Enumerable` was designed to support infinite
   collections, resources and other structures with fixed shape. For example,
-  it doesn't make sense to insert values into a range, as it has a fixed
-  shape where just the range limits are stored.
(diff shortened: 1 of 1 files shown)


eksperimental:
Please apply to branch v1.12

lib/elixir/lib/collectable.ex
```diff
@@ -17,8 +17,8 @@ defprotocol Collectable do
 
   This design is intentional. `Enumerable` was designed to support infinite
   collections, resources and other structures with fixed shape. For example,
-  it doesn't make sense to insert values into a range, as it has a fixed
-  shape where just the range limits are stored.
+  it doesn't make sense to insert values into a [range](`t:Range.t/0`), as it has a fixed
+  shape where just the range [limits](`t:Range.limit/0`) and the [step](`t:Range.step/0`) are stored.
```
josevalim, line 21:
```suggestion
  it doesn't make sense to insert values into a `Range`, as it has a
  fixed shape where only the range limits and step are stored.
```

josevalim:

PR #10949, opened 2021-04-23
Allowing absolute path completion on Windows

kelvinst:
Path autocompletion on should be also triggered for `"C:/<tab>` (where C can be any letter, upper or lower cased), so we can autocomplete absolute paths on Windows too


kelvinst:
There are still some tests failing on Windows, but it is on elixir stdlib (which I did not touched) and they were failing even before my PR got merged, so I suppose we are good to go right?

josevalim:

PR #10954, opened 2021-04-25
Allow filtering modules from coverage using regex

hauleth:
This can be handy in case of many generated modules (in my case it was `Cldr` module containing a lot of generated code) and protocol modules that aren't really interesting for our application (for example `Inspect` protocol when filtering out fields like passwords).


josevalim:
We need docs and tests. :)

josevalim:
Ping! :)

hauleth:
I will handle it today or over the weekend.

hauleth:
@josevalim done with tests.

lib/mix/lib/mix/tasks/test.coverage.ex
```diff
@@ -206,6 +206,17 @@ defmodule Mix.Tasks.Test.Coverage do
...
+  defp ignored?(_, []), do: false
+  defp ignored?(mod, [mod | _]), do: true
+
+  defp ignored?(mod, [%Regex{} = re | rest]) do
+    name = inspect(mod)
+
+    Regex.match?(re, name) || ignored?(mod, rest)
+  end
+
+  defp ignored?(mod, [_ | rest]), do: ignored?(mod, rest)
```
josevalim, line 218:
```suggestion
  defp ignored?(mod, [%Regex{} = re | rest]),
    do: Regex.match?(re, inspect(mod)) or ignored?(mod, rest)

  defp ignored?(mod, [_ | rest]), do: ignored?(mod, rest)
```

josevalim, line 218:
```suggestion
  defp ignored?(mod, ignores) do
    Enum.any?(ignores, &ignored_any?(mod, &1))
  end

  defp ignored_any?(mod, %Regex{} = re), do: Regex.match?(re, inspect(mod))
  defp ignored_any?(mod, other), do: mod == other
```

josevalim (commented):
Looks great! I have suggested one tiny refactoring. Then we need docs and we should be good to go!

hauleth:
Added docs.

lib/mix/lib/mix/tasks/test.ex
```diff
@@ -285,7 +285,8 @@ defmodule Mix.Tasks.Test do
       See `mix test.coverage` to compile a report from multiple exports
 
     * `:ignore_modules` - modules to ignore from generating reports and
-      in summaries
+      in summaries. It can be either module name as an atom or regular
+      exressions that will be matched against stringified module names.
```
josevalim, line 289:
```suggestion
      in summaries. It is a list of module names as atoms and regular
      expressions that are matched against the module names
```

josevalim:

PR #10955, opened 2021-04-26
Unify CI tests

eksperimental:
- It runs the Check Reproducible test on every operating system.
- Standardizes labels, and variable names across the different operating systems

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -12,7 +12,7 @@ test_template: &DEFAULT_TEST_SETTINGS
 test_freebsd_task:
   <<: *DEFAULT_TEST_SETTINGS
 
-  name: FreeBSD 13.0
+  name: FreeBSD 13.0, OTP-${OTP_VERSION}
   alias: FreeBSD Stable
 
   freebsd_instance:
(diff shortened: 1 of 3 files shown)


josevalim:
Thank you but there is no reason to run the reproducible check across all OSes.

eksperimental:
We are currently testing the reproducibility for Linux and FreeBSD, so the only one left out is Windows.
In this previous PR of mine https://github.com/elixir-lang/elixir/pull/9600 it seemed we were OK with checking it for Window builds.
I would like to hear @bmwiedemann opinion on this. If it is not needed, maybe we can even remove it from FreeBSD as well.

josevalim:
Not everything is about consistency. It is easy to check for consistency on FreeBSD because it is the same command. I don't think the extra infrastructure though is worth to test on Windows given this behaviour is no OS specific.

bmwiedemann:
I mostly care about Linux. I tested that openSUSE's `elixir-1.11.4` package builds reproducibly both with and without this patch.

eksperimental:
> I mostly care about Linux. I tested that openSUSE's `elixir-1.11.4` package builds reproducibly both with and without this patch.

Thank you for popping in @bmwiedemann. My concern was wether the builds could vary on Windows in some way that it does not under Linux.

PR #10956, opened 2021-04-26
Provide a more fitting example of Kernel.then

peaceful-james:


josevalim:
Can you please explain why it is more fitting? Thank you.

Eiji7:
@josevalim This PR was mentioned in his comment on forum:
https://elixirforum.com/t/elixir-v1-12-0-rc-0-released/38667/32?u=eiji

josevalim:
I see.

it is the first argument but there is no straightforward way to pipe into operators besides using Kernel.*, which is not everyone’s cup of tea.

we can add more examples though, but we should keep the current one.

thanks.

wojtekmach:
This is super anecdotal but I find then/1 especially useful for stuff like: `|> then(&File.write!("foo.txt", &1)`, `|> then(&IO.puts(:stderr, &1)`, etc.

josevalim:

PR #10961, opened 2021-05-01
Add more normalization examples

achempion:
I thought that `nfkc` example data was misspelled, it was "fi" (two chars) instead of "ﬁ" (one char) and decided instead of fix add more examples.

```elixir
iex> String.normalize("fi", :nfkc)
"fi"
```


josevalim:
I don't think those examples are less confusing because they are also vague, so I would keep it as is. Thank you!

PR #10963, opened 2021-05-03
Optimize Enum.unzip/1 for lists

pguillory:
The existing unzip implementation is generic for all enumerable types. However, large optimizations are available if the input is a list.

First, instead of doing two reverse operations at the end, we can do one at the beginning. This optimization is unavailable in the existing generic implementation in order to avoid materializing an enumerable that might be generated lazily.

Second, we can accumulate the two output lists in two separate parameters, rather than in the reduce function's single accumulator. This avoids generating a tuple per input element that would just be matched away and discarded.

Testing with Benchfella shows a roughly 2x to 3x performance improvement depending on the size of the list.

### DIFF
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -3349,6 +3349,15 @@ defmodule Enum do
 
   """
   @spec unzip(t) :: {[element], [element]}
+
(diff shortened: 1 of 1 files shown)


josevalim:

PR #10965, opened 2021-05-04
Add System.shell/2

josevalim:


lib/elixir/lib/system.ex
```diff
@@ -880,6 +880,52 @@ defmodule System do
     :init.stop(String.to_charlist(status))
   end
 
+  @doc ~S"""
+  Executes the given `command` in the current OS shell.
+
+  **Important**: Use this function with care. In particular, **never
+  pass user input to this function**, as the user would be able to
+  execute any code directly on the machine. Generally speaking,
```
v0idpwn, line 888: I think it would be good to mention that this is a "command injection" vulnerability, so the reader has a keyword to search for and better understand the implications.

lib/elixir/lib/system.ex
```diff
@@ -880,6 +880,52 @@ defmodule System do
     :init.stop(String.to_charlist(status))
   end
 
+  @doc ~S"""
+  Executes the given `command` in the current OS shell.
```
v0idpwn, line 884:
```suggestion
  Executes the given `command` in the OS shell.
```

"current" makes me think that it will read the "SHELL" env var or something similar

lib/elixir/lib/system.ex
```diff
@@ -880,6 +880,55 @@ defmodule System do
...
+  ## Examples
+
+      iex> System.shell("echo hello")
+      {"hello\n", 0}
+
+  ## Options
+
+  It accepts the same options as `cmd/3`.
+  """
+  @spec shell(binary, keyword) :: {Collectable.t(), exit_status :: non_neg_integer}
```
fertapric, line 903:
```suggestion
  """
  @doc since: "1.13.0"
  @spec shell(binary, keyword) :: {Collectable.t(), exit_status :: non_neg_integer}
```

fertapric (approved):

lib/elixir/lib/system.ex
```diff
@@ -880,6 +880,55 @@ defmodule System do
...
+        {:unix, _} ->
+          command =
+            command
+            |> String.replace("\"", "\\\"")
+            |> String.to_charlist()
+
+          'sh -c "' ++ command ++ '"'
+
+        {:win32, osname} ->
+          command = '"' ++ String.to_charlist(command) ++ '"'
```
lukaszsamson, line 920: no escaping on windows?

josevalim: According to the OTP source, no, but we shouldn't be quoting either. So I will update. Thanks! :)

josevalim, line 920:
```suggestion
          command = String.to_charlist(command)
```

ericmj (approved):

josevalim:

PR #10971, opened 2021-05-10
Fix signature expansion when no docs

### LINKED ISSUE #10970: IEx crashing on tab completion
### Environment

```
Erlang/OTP 24 [RELEASE CANDIDATE 3] [erts-12.0] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1] [jit]

Elixir 1.12.0-rc.1 (ce321f8) (compiled with Erlang/OTP 24)
```
* Operating system: Arch Linux (installed with asdf)

### Current behavior

Contents of `foo.ex`:
```ex
defmodule Foo do
  def foo(a), do: a
end
```

- Run `iex foo.ex`
- Type `Foo.foo(` and then press tab
- An error occurs:
```
iex(1)> Foo.foo(*** ERROR: Shell process terminated! (^G to start new job) ***

11:30:47.973 [error] Process #PID<0.66.0> raised an exception
** (FunctionClauseError) no function clause matching in IEx.Autocomplete.expand_signatures/2
    (iex 1.12.0-rc.1) lib/iex/autocomplete.ex:135: IEx.Autocomplete.expand_signatures(:error, #PID<0.78.0>)
    (kernel 8.0) group.erl:596: :group.get_line1/5
    (kernel 8.0) group.erl:471: :group.get_chars_loop/9
    (kernel 8.0) group.erl:181: :group.io_request/6
    (kernel 8.0) group.erl:117: :group.server_loop/3
```

This appears to happen for any user-defined functions or functions from external libraries (including ones written in Erlang), except for ones with arity 0.

### Expected behavior

IEx should not crash and a completion list should be shown.

v0idpwn:
Closes #10970

### DIFF
--- a/lib/iex/lib/iex/autocomplete.ex
+++ b/lib/iex/lib/iex/autocomplete.ex
@@ -135,7 +135,7 @@ defmodule IEx.Autocomplete do
     with docs when is_list(docs) <- get_docs(module, [:function, :macro], name) do
       Enum.map(docs, fn {_, _, signatures, _, _} -> Enum.join(signatures, " ") end)
     else
-      _ -> :error
+      _ -> []
     end
   end
 
--- a/lib/iex/test/iex/autocomplete_test.exs
+++ b/lib/iex/test/iex/autocomplete_test.exs
@@ -398,6 +398,12 @@ defmodule IEx.AutocompleteTest do
 
     assert expand('take(') == {:yes, '', ['take(enumerable, amount)']}
     assert expand('derive(') == {:yes, '', ['derive(protocol, module, options \\\\ [])']}
+
+    defmodule NoDocs do
+      def sample(a), do: a
+    end
+
+    assert {:yes, [], [_ | _]} = expand('NoDocs.sample(')
   end
 
   @tag :tmp_dir


lib/iex/lib/iex/autocomplete.ex
```diff
@@ -147,6 +147,8 @@ defmodule IEx.Autocomplete do
 
   defp expand_signatures([], shell), do: expand_local_or_var("", shell)
 
+  defp expand_signatures(_, _), do: no()
+
```
v0idpwn, line 151:
```suggestion
  defp expand_signatures(_, shell), do: expand_local_or_var("", shell)
```

The other alternative: this would suggest all the alternatives from context

josevalim: Yes, I would make it return an empty list instead. Can you please update the PR and add tests? Thanks.

v0idpwn:
Hope its correct now :)

lib/iex/test/iex/autocomplete_test.exs
```diff
@@ -398,6 +398,14 @@ defmodule IEx.AutocompleteTest do
 
     assert expand('take(') == {:yes, '', ['take(enumerable, amount)']}
     assert expand('derive(') == {:yes, '', ['derive(protocol, module, options \\\\ [])']}
+
+    eval("defmodule NoDocs do; def sample(a), do: a; end")
```
josevalim, line 402: I don't think it needs to be define in the shell. I think  you can define a function in `IEx.AutocompleteTest` itself and try to call it from the shell.

v0idpwn: Indeed. Changed another thing and it looks a lot better now.

josevalim:

PR #10975, opened 2021-05-12
Add `Kernel.then?/3` with a predicate parameter.

azizk:
Hey!

I think `Kernel.then/2` was a great addition so I figured why not add another helpful function to greatly improve the readability of some code? I chose to add the `?` as a suffix to make it clearer, but it's fine without it also.


josevalim:
Thank you but at the moment we are not planning to add more composition on the pipelines. Predicates, `if_ok`, etc would all be possible additions, but those can also be tackled in libraries or in your codebase. thanks!

azizk:
Alright, no problem! :)

PR #10976, opened 2021-05-12
Add `--shortest` option to `mix xref graph`

marcandre:
This option is available when `--source` and `--sink` are specified
and reduces the graph to the shortest path between the source and sink.

Prior discussion [in elixirforum](https://elixirforum.com/t/how-to-get-a-simple-path-from-xref-graph/39501/2)

### DIFF
--- a/lib/mix/lib/mix/tasks/xref.ex
+++ b/lib/mix/lib/mix/tasks/xref.ex
@@ -56,6 +56,9 @@ defmodule Mix.Tasks.Xref do
     * `--sink` - displays all files that reference the given file
       (directly or indirectly)
 
+    * `--shortest` - displays the shortest path between the source
+      and sink
+
     * `--min-cycle-size` - controls the minimum cycle size on formats
       like `stats` and `cycles`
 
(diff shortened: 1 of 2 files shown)


marcandre:
This is sadly not as useful as I was hoping for, because the shortest path might be uninteresting (e.g. 100% runtime), and the `--label` filtering is too aggressive as a shortest path would have to consist exclusively of the right type of label.

Putting this as draft, I'd like to find a better solution.

josevalim:
I thought that could be the case. Can you explain what you are hoping to achieve? Finding the shortest path to eliminate compile time deps? That won’t help much because all paths matter.

Other than that, the label option should keep transitive compile deps, but I can see that still being too aggressive.

marcandre:
> Can you explain what you are hoping to achieve? Finding the shortest path to eliminate compile time deps? 

Yes, I have a source that gets recompiled whenever I modify my sink, and it shouldn't be (conceptually). I'd like to see which dependencies I should sever so there aren't any compile-time deps.

> That won’t help much because all paths matter.

The graph between my source and sink is huge, and includes complete paths that have no compile-time deps. My understanding is that these do not matter at all, so I thought it would be nice to simplify it, or have a path with some compile-time dep, sever that and repeat.

I guess that I can still use the current output and just focus on any node with (compile) in it and go from there though.

PR #10981, opened 2021-05-16
Remove sub_dirs support from Rebar 2

josevalim:
Rebar 2 no longer works on Erlang/OTP 24+, which means
any project relying on it should move to Rebar 3. And Rebar 3
does not support sub_dirs as _deps_, so we remove support
for this feature to start unifying the runtimes.


josevalim:

PR #10986, opened 2021-05-17
Change formatter comments from tuple to map

doorgan:
This changes the comments in the formatter to be maps instead of tuples, as suggested in the [Expose functions from `Code.Formatter` via the `Code` module](https://groups.google.com/g/elixir-lang-core/c/-8CPorfVTxg) proposal


whatyouhide (approved):

josevalim:
@doorgan thank you so much! Note that if we extract this to a separate function the format_comment shouldn't be extracted, but everything else looks great!

josevalim:

PR #10988, opened 2021-05-18
Expose more formatter functions in the `Code` module

doorgan:
This adds `quoted_to_algebra/2` to the `Code` module, as discussed in [this proposal](https://groups.google.com/u/1/g/elixir-lang-core/c/-8CPorfVTxg).

I feel the testing can be improved, if so I would like some guidance.

### DIFF
--- a/lib/elixir/lib/code.ex
+++ b/lib/elixir/lib/code.ex
@@ -1360,6 +1360,11 @@ defmodule Code do
 
     * `:comments` - the list of comments associated with the quoted expression.
       Defaults to `[]`.
+
+    * `:escape` - when `true`, escaped sequences like `\n` will be escaped into
+      `\\n`. If the `:unescape` option was set to `false` when using
+      `string_to_quoted/2`, setting this option to `false` will prevent it from
+      escaping the sequences twice. Defaults to `true`.
   """
   @spec quoted_to_algebra(Macro.t(), keyword) :: Inspect.Algebra.t()
   def quoted_to_algebra(quoted, opts \\ []) do
--- a/lib/elixir/lib/code.ex
+++ b/lib/elixir/lib/code.ex
@@ -1100,6 +1100,11 @@ defmodule Code do
     * `:columns` - when `true`, attach a `:column` key to the quoted
       metadata. Defaults to `false`.
 
+    * `:unescape` - when `false`, preserves escaped sequences. For example,
+      `"null byte\\t\\x00"` will be kept as is instead of being converted to a
+      bit string literal. This is useful in combination with
+      `quoted_to_algebra/2`. Defaults to `true`.
+
     * `:existing_atoms_only` - when `true`, raises an error
       when non-existing atoms are found by the tokenizer.
       Defaults to `false`.
@@ -1310,8 +1315,25 @@ defmodule Code do
   @doc """
   Converts a quoted expression to an algebra document.
 
-  Supports both regular AST and the extended AST documented in
-  `string_to_quoted_with_comments/2`.
+  The elixir AST does not contain metadata for literals like strings, lists, or
+  tuples with two elements, which means that the produced algebra document will
+  not respect all of the user preferences and comments may be misplaced.
+  To get better results, you can use the `:token_metadata`, `:escape` and
+  `:literal_encoder` options to `string_to_quoted/2` to provide additional
+  information to the formatter:
+
+      [
+        literal_encoder: &{:ok, {:__block__, &2, [&1]}},
+        token_metadata: true,
+        unescape: false
+      ]
+
+  This will produce an AST that contains information such as `do` blocks start
+  and end lines or sigil delimiters, and by wrapping literals in blocks they can
+  now hold metadata like line number, string delimiter and escaped sequences, or
+  integer formatting(ie `0x2a` instead of `47`). This is especially useful if
+  you're doing source code manipulation, where it's important to preserve user
+  choices and comments placing.
 
   ## Options
 
--- a/lib/elixir/lib/code.ex
+++ b/lib/elixir/lib/code.ex
@@ -1360,7 +1360,7 @@ defmodule Code do
   @spec quoted_to_algebra(Macro.t(), keyword) :: Inspect.Algebra.t()
   def quoted_to_algebra(quoted, opts \\ []) do
     quoted
-    |> Code.Normalizer.normalize()
+    |> Code.Normalizer.normalize(opts)
     |> Code.Formatter.to_algebra(opts)
   end
 
--- a/lib/elixir/lib/code.ex
+++ b/lib/elixir/lib/code.ex
@@ -980,8 +980,25 @@ defmodule Code do
   @spec format_string!(binary, keyword) :: iodata
   def format_string!(string, opts \\ []) when is_binary(string) and is_list(opts) do
     line_length = Keyword.get(opts, :line_length, 98)
-    algebra = Code.Formatter.to_algebra!(string, opts)
-    Inspect.Algebra.format(algebra, line_length)
+
+    to_quoted_opts =
+      [
+        unescape: false,
+        warn_on_unnecessary_quotes: false,
+        literal_encoder: &{:ok, {:__block__, &2, [&1]}},
+        token_metadata: true
+      ] ++ opts
+
+    {forms, comments} = Code.string_to_quoted_with_comments!(string, to_quoted_opts)
+
+    to_algebra_opts =
+      [
+        comments: comments
+      ] ++ opts
+
+    doc = Code.Formatter.to_algebra(forms, to_algebra_opts)
+
+    Inspect.Algebra.format(doc, line_length)
   end
 
   @doc """
@@ -1277,8 +1294,8 @@ defmodule Code do
   @spec string_to_quoted_with_comments!(List.Chars.t(), keyword) :: {Macro.t(), map()}
   def string_to_quoted_with_comments!(string, opts \\ []) do
     case string_to_quoted_with_comments(string, opts) do
-      {:ok, forms_and_comments} ->
-        forms_and_comments
+      {:ok, forms, comments} ->
+        {forms, comments}
 
       {:error, {location, error, token}} ->
         :elixir_errors.parse_error(location, Keyword.get(opts, :file, "nofile"), error, token)
@@ -1344,7 +1361,7 @@ defmodule Code do
   def quoted_to_algebra(quoted, opts \\ []) do
     quoted
     |> Code.Normalizer.normalize()
-    |> Code.Formatter.quoted_expression_to_algebra(opts)
+    |> Code.Formatter.to_algebra(opts)
   end
 
   @doc """
--- a/lib/elixir/lib/code.ex
+++ b/lib/elixir/lib/code.ex
@@ -989,7 +989,7 @@ defmodule Code do
         token_metadata: true
       ] ++ opts
 
-    {forms, comments} = Code.string_to_quoted_with_comments!(string, to_quoted_opts)
+    {forms, comments} = string_to_quoted_with_comments!(string, to_quoted_opts)
 
     to_algebra_opts =
       [
--- a/lib/elixir/lib/code.ex
+++ b/lib/elixir/lib/code.ex
@@ -1308,6 +1308,24 @@ defmodule Code do
   defp previous_eol_count(_), do: 0
 
   @doc """
+  Converts a quoted expression to an algebra document.
+
+  Supports both regular AST and the extended AST documented in
+  `string_to_quoted_with_comments/2`.
+
+  ## Options
+
+    * `:comments` - the list of comments associated with the quoted expression.
+      Defaults to `[]`.
+  """
+  @spec quoted_to_algebra(Macro.t(), keyword) :: Inspect.Algebra.t()
+  def quoted_to_algebra(quoted, opts \\ []) do
+    quoted
+    |> Code.Normalizer.normalize()
+    |> Code.Formatter.quoted_expression_to_algebra(opts)
+  end
+
+  @doc """
   Evals the given file.
 
   Accepts `relative_to` as an argument to tell where the file is located.
--- a/lib/elixir/lib/code/formatter.ex
+++ b/lib/elixir/lib/code/formatter.ex
@@ -210,26 +210,42 @@ defmodule Code.Formatter do
         comments
         |> Enum.map(&format_comment/1)
         |> gather_comments()
+        |> state(opts)
 
-      {:ok, {forms, comments}}
+      {doc, _} = block_to_algebra(forms, @min_line, @max_line, state)
+      {:ok, doc}
     end
   end
 
   @doc """
-  Converts `string` to a quoted expression and a list of comments.
+  Converts `string` to an algebra document.
 
   Raises if the `string` cannot be parsed.
 
-  See `Code.string_to_quoted_with_comments!/2` for the list of options.
+  See `Code.format_string!/2` for the list of options.
   """
-  def string_to_quoted_with_comments!(string, opts \\ []) do
-    case string_to_quoted_with_comments(string, opts) do
-      {:ok, forms_and_comments} ->
-        forms_and_comments
+  def to_algebra!(string, opts \\ []) do
+    case to_algebra(string, opts) do
+      {:ok, doc} ->
+        doc
 
       {:error, {location, error, token}} ->
         :elixir_errors.parse_error(location, Keyword.get(opts, :file, "nofile"), error, token)
     end
+  end
+
+  defp state(comments, opts) do
+    force_do_end_blocks = Keyword.get(opts, :force_do_end_blocks, false)
+
+    locals_without_parens =
+      Keyword.get(opts, :locals_without_parens, []) ++ @locals_without_parens
+
+    %{
+      force_do_end_blocks: force_do_end_blocks,
+      locals_without_parens: locals_without_parens,
+      operand_nesting: 2,
+      comments: comments
+    }
   end
 
   @doc """
--- a/lib/elixir/lib/code/formatter.ex
+++ b/lib/elixir/lib/code/formatter.ex
@@ -191,47 +191,20 @@ defmodule Code.Formatter do
   end
 
   @doc """
-  Converts `string` to an algebra document.
-
-  Returns `{:ok, doc}` or `{:error, parser_error}`.
-
-  See `Code.format_string!/2` for the list of options.
+  Converts the quoted expression into an algebra document
   """
-  def to_algebra(string, opts \\ []) when is_binary(string) and is_list(opts) do
-    to_quoted_opts = [
-      unescape: false,
-      warn_on_unnecessary_quotes: false,
-      literal_encoder: &{:ok, {:__block__, &2, [&1]}},
-      token_metadata: true
-    ]
+  def to_algebra(quoted, opts \\ []) do
+    comments = Keyword.get(opts, :comments, [])
 
-    with {:ok, forms, comments} <- Code.string_to_quoted_with_comments(string, to_quoted_opts) do
-      state =
-        comments
-        |> Enum.map(&format_comment/1)
-        |> gather_comments()
-        |> state(opts)
+    state =
+      comments
+      |> Enum.map(&format_comment/1)
+      |> gather_comments()
+      |> state(opts)
 
-      {doc, _} = block_to_algebra(forms, @min_line, @max_line, state)
-      {:ok, doc}
-    end
-  end
+    {doc, _} = block_to_algebra(quoted, @min_line, @max_line, state)
 
-  @doc """
-  Converts `string` to an algebra document.
-
-  Raises if the `string` cannot be parsed.
-
-  See `Code.format_string!/2` for the list of options.
-  """
-  def to_algebra!(string, opts \\ []) do
-    case to_algebra(string, opts) do
-      {:ok, doc} ->
-        doc
-
-      {:error, {location, error, token}} ->
-        :elixir_errors.parse_error(location, Keyword.get(opts, :file, "nofile"), error, token)
-    end
+    doc
   end
 
   defp state(comments, opts) do
@@ -246,19 +219,6 @@ defmodule Code.Formatter do
       operand_nesting: 2,
       comments: comments
     }
-  end
-
-  @doc """
-  Converts the quoted expression into an algebra document
-  """
-  def quoted_expression_to_algebra(quoted, opts \\ []) do
-    comments = Keyword.get(opts, :comments, [])
-
-    state = state(comments, opts)
-
-    {doc, _} = block_to_algebra(quoted, @min_line, @max_line, state)
-
-    {:ok, doc}
   end
 
   defp format_comment(%{text: text} = comment) do
--- a/lib/elixir/lib/code/formatter.ex
+++ b/lib/elixir/lib/code/formatter.ex
@@ -210,42 +210,39 @@ defmodule Code.Formatter do
         comments
         |> Enum.map(&format_comment/1)
         |> gather_comments()
-        |> state(opts)
 
-      {doc, _} = block_to_algebra(forms, @min_line, @max_line, state)
-      {:ok, doc}
+      {:ok, {forms, comments}}
     end
   end
 
   @doc """
-  Converts `string` to an algebra document.
+  Converts `string` to a quoted expression and a list of comments.
 
   Raises if the `string` cannot be parsed.
 
-  See `Code.format_string!/2` for the list of options.
+  See `Code.string_to_quoted_with_comments!/2` for the list of options.
   """
-  def to_algebra!(string, opts \\ []) do
-    case to_algebra(string, opts) do
-      {:ok, doc} ->
-        doc
+  def string_to_quoted_with_comments!(string, opts \\ []) do
+    case string_to_quoted_with_comments(string, opts) do
+      {:ok, forms_and_comments} ->
+        forms_and_comments
 
       {:error, {location, error, token}} ->
         :elixir_errors.parse_error(location, Keyword.get(opts, :file, "nofile"), error, token)
     end
   end
 
-  defp state(comments, opts) do
-    force_do_end_blocks = Keyword.get(opts, :force_do_end_blocks, false)
+  @doc """
+  Converts the quoted expression into an algebra document
+  """
+  def quoted_expression_to_algebra(quoted, opts \\ []) do
+    comments = Keyword.get(opts, :comments, [])
 
-    locals_without_parens =
-      Keyword.get(opts, :locals_without_parens, []) ++ @locals_without_parens
+    state = state(comments, opts)
 
-    %{
-      force_do_end_blocks: force_do_end_blocks,
-      locals_without_parens: locals_without_parens,
-      operand_nesting: 2,
-      comments: comments
-    }
+    {doc, _} = block_to_algebra(quoted, @min_line, @max_line, state)
+
+    {:ok, doc}
   end
 
   defp format_comment(%{text: text} = comment) do
--- a/lib/elixir/lib/code/normalizer.ex
+++ b/lib/elixir/lib/code/normalizer.ex
@@ -13,66 +13,72 @@ defmodule Code.Normalizer do
   """
   def normalize(quoted, opts \\ []) do
     line = Keyword.get(opts, :line, 1)
+    escape = Keyword.get(opts, :escape, true)
 
-    do_normalize(quoted, line: line)
+    state = %{
+      escape: escape,
+      parent_meta: [line: line]
+    }
+
+    do_normalize(quoted, state)
   end
 
   # Skip normalized literals
-  defp do_normalize({:__block__, _, [literal]} = quoted, _parent_meta) when is_literal(literal) do
+  defp do_normalize({:__block__, _, [literal]} = quoted, _state) when is_literal(literal) do
     quoted
   end
 
   # Skip normalized charlists
-  defp do_normalize({:__block__, meta, [charlist]} = quoted, parent_meta)
+  defp do_normalize({:__block__, meta, [charlist]} = quoted, state)
        when is_list(charlist) do
     if Keyword.has_key?(meta, :delimiter) do
       quoted
     else
-      do_normalize(charlist, parent_meta)
+      do_normalize(charlist, state)
     end
   end
 
   # Only normalize the first argument of an alias if it's not an atom
-  defp do_normalize({:__aliases, meta, [first | rest]}, parent_meta) when not is_atom(first) do
-    meta = patch_meta_line(meta, parent_meta)
-    first = do_normalize(first, meta)
+  defp do_normalize({:__aliases, meta, [first | rest]}, state) when not is_atom(first) do
+    meta = patch_meta_line(meta, state.parent_meta)
+    first = do_normalize(first, %{state | paret_meta: meta})
     {:__aliases__, meta, [first | rest]}
   end
 
-  defp do_normalize({:__aliases__, _, _} = quoted, _parent_meta) do
+  defp do_normalize({:__aliases__, _, _} = quoted, _state) do
     quoted
   end
 
   # Skip captured arguments like &1
-  defp do_normalize({:&, meta, [term]}, parent_meta) when is_integer(term) do
-    meta = patch_meta_line(meta, parent_meta)
+  defp do_normalize({:&, meta, [term]}, state) when is_integer(term) do
+    meta = patch_meta_line(meta, state.parent_meta)
     {:&, meta, [term]}
   end
 
   # Ranges
-  defp do_normalize(left..right//step, parent_meta) do
-    left = do_normalize(left, parent_meta)
-    right = do_normalize(right, parent_meta)
+  defp do_normalize(left..right//step, state) do
+    left = do_normalize(left, state)
+    right = do_normalize(right, state)
 
-    meta = [line: parent_meta[:line]]
+    meta = [line: state.parent_meta[:line]]
 
     if step == 1 do
       {:.., meta, [left, right]}
     else
-      step = do_normalize(step, parent_meta)
+      step = do_normalize(step, state)
       {:"..//", meta, [left, right, step]}
     end
   end
 
   # Bit containers
-  defp do_normalize({:<<>>, meta, parts} = quoted, parent_meta) do
-    meta = patch_meta_line(meta, parent_meta)
+  defp do_normalize({:<<>>, meta, parts} = quoted, state) do
+    meta = patch_meta_line(meta, state.parent_meta)
 
     parts =
       if interpolated?(quoted) do
-        normalize_interpolation_parts(parts, meta)
+        normalize_interpolation_parts(parts, %{state | parent_meta: meta})
       else
-        Enum.map(parts, &do_normalize(&1, meta))
+        Enum.map(parts, &do_normalize(&1, %{state | parent_meta: meta}))
       end
 
     {:<<>>, meta, parts}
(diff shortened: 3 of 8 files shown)


doorgan:
All tests passed when I ran `make clean test`, but for some reason the FreeBSD CI is failing and I don't understand that stacktrace :/

lib/elixir/lib/code.ex
```diff
@@ -1192,6 +1192,101 @@ defmodule Code do
...
+  closest tokens:
+
+      iex> Code.string_to_quoted_with_comments("""
+      ...> :foo
+      ...>
+      ...> # Hello, world!
+      ...>
+      ...>
+      ...> # Some more comments!
+      ...> """)
```
josevalim, line 1242:
```suggestion
      iex> Code.string_to_quoted_with_comments("\""
      ...> :foo
      ...>
      ...> # Hello, world!
      ...>
      ...>
      ...> # Some more comments!
      ...> "\"")
```

lib/elixir/lib/code.ex
```diff
@@ -1192,6 +1192,101 @@ defmodule Code do
...
+
+    * `:file` - the filename to be reported in case of parsing errors.
+      Defaults to "nofile".
+
+    * `:line` - the starting line of the string being parsed.
+      Defaults to 1.
+  """
+  @spec string_to_quoted_with_comments(List.Chars.t(), keyword) ::
+          {:ok, {Macro.t(), map()}} | {:error, {location :: keyword, term, term}}
+  defdelegate string_to_quoted_with_comments(string, opts \\ []), to: Code.Formatter
```
josevalim, line 1258: Let's bring this particular logic from the formatter here. :)

lib/elixir/lib/code/formatter.ex
```diff
@@ -217,48 +263,45 @@ defmodule Code.Formatter do
...
-      state =
+      comments =
         Process.get(:code_formatter_comments)
         |> Enum.reverse()
         |> gather_comments()
-        |> state(opts)
 
-      {doc, _} = block_to_algebra(forms, @min_line, @max_line, state)
-      {:ok, doc}
+      {:ok, {forms, comments}}
```
josevalim, line 271:
```suggestion
      {:ok, forms, comments}
```

?

lib/elixir/lib/code/normalizer.ex
```diff
@@ -0,0 +1,246 @@
...
+    do_normalize(quoted, line: line)
+  end
+
+  # Skip already wrapped blocks
+  defp do_normalize({:__block__, _, [literal]} = quoted, _parent_meta) when is_literal(literal) do
+    quoted
+  end
+
+  defp do_normalize({:__block__, meta, args} = quoted, _parent_meta) do
+    if Keyword.has_key?(meta, :format) do
```
josevalim, line 26: Why do we skip in this case?

lib/elixir/lib/code/normalizer.ex
```diff
@@ -0,0 +1,246 @@
...
+    end
+  end
+
+  # Skip aliases so the module segment atoms don't get wrapped
+  defp do_normalize({:__aliases__, _, _} = quoted, _parent_meta) do
+    quoted
+  end
+
+  # Skip qualified tuples left hand side
+  defp do_normalize({:., _, [_, :{}]} = quoted, _parent_meta) do
```
josevalim, line 39: You may need to traverse the left of `.` anyway. For example, the code could be this: `:foo.{1, 2, 3}`.

lib/elixir/lib/code/normalizer.ex
```diff
@@ -0,0 +1,246 @@
...
+  defp do_normalize({:__block__, meta, args} = quoted, _parent_meta) do
+    if Keyword.has_key?(meta, :format) do
+      quoted
+    else
+      {:__block__, meta, Enum.map(args, &do_normalize(&1, meta))}
+    end
+  end
+
+  # Skip aliases so the module segment atoms don't get wrapped
+  defp do_normalize({:__aliases__, _, _} = quoted, _parent_meta) do
```
josevalim, line 34: You need to traverse the first element of __aliases__ in case it is not an atom. For example:

    __MODULE__.Bar.Baz

Maybe __MODULE__ will be fine but the first element of __aliases__ can be any expression.

lib/elixir/lib/code/normalizer.ex
josevalim: Maybe we can reverse this clause altogether because of line 51?

lib/elixir/lib/code/normalizer.ex
```diff
@@ -0,0 +1,246 @@
...
+    {:->, meta, [left, right]}
+  end
+
+  # Maps
+  defp do_normalize({:%{}, meta, args}, _parent_meta) do
+    args = Enum.map(args, &do_normalize(&1, meta))
+
+    args =
+      case args do
+        # Unwrap the right hand side if we're in an update syntax
```
josevalim, line 74: I would prefer to follow the same style as the remaining of code which is to avoid wrapping in the first place (instead of unwrapping later). That's because I can always understand the input (I know which shapes are valid AST) but I don't know all possible outputs for normalize. :) Would it be possible to change this?

lib/elixir/lib/code/normalizer.ex
```diff
@@ -0,0 +1,246 @@
...
+    {:%{}, meta, args}
+  end
+
+  # If a keyword list is an argument of a guard, we need to drop the block
+  # wrapping
+  defp do_normalize({:when, meta, args} = _quoted, _parent_meta) do
+    args =
+      Enum.map(args, fn
+        arg when is_list(arg) ->
+          {_, _, [arg]} = do_normalize(arg, meta)
```
josevalim, line 95: Maybe break normalize in two? `do_normalize` handles literals and then calls `do_normalize_non_literal`, which you would also invoke here?

josevalim: Another option is to make things consistent and to change the parser and make sure the args argument is also wrapped in the literal block when it is a keyword, and change the formatter to handle it accordingly. I think I might prefer that. Thoughts?

lib/elixir/lib/code/normalizer.ex
```diff
@@ -0,0 +1,246 @@
...
+    end
+  end
+
+  # Integers, floats, atoms
+  defp do_normalize(x, parent_meta) when is_literal(x) do
+    meta = [line: parent_meta[:line]]
+
+    meta =
+      if not is_atom(x) do
+        Keyword.put(meta, :token, Macro.to_string(x))
```
josevalim, line 156: Instead of defining it here, should we add a default for the formatter? Also, we probably want `inspect(x)` instead of Macro.to_string.

lib/elixir/lib/code/normalizer.ex
```diff
@@ -0,0 +1,246 @@
...
+
+    {:__block__, meta,
+     [
+       {do_normalize(left, left_parent_meta), do_normalize(right, parent_meta)}
+     ]}
+  end
+
+  # Lists
+  defp do_normalize(list, parent_meta) when is_list(list) do
+    if !Enum.empty?(list) and List.ascii_printable?(list) do
```
josevalim, line 190:
```suggestion
    if list != [] and List.ascii_printable?(list) do
```

lib/elixir/lib/code/normalizer.ex
```diff
@@ -0,0 +1,246 @@
...
+
+  # Everything else
+  defp do_normalize(quoted, _parent_meta) do
+    quoted
+  end
+
+  defp normalize_list_elements(elems, parent_meta, keyword? \\ false)
+
+  defp normalize_list_elements([[{_, _, [{_, _}]}] = first | rest], parent_meta, keyword?) do
+    # Skip already normalized 2-tuples
```
josevalim, line 210: Why would we need this? It seems we might be normalizing some things twice?

lib/elixir/test/elixir/code_normalizer_test.exs
```diff
@@ -0,0 +1,364 @@
...
+               ]}
+            ]} = normalized
+  end
+
+  test "unqualified call's form is skipped" do
+    quoted = Code.string_to_quoted!("foo()")
+    normalized = Normalizer.normalize(quoted)
+
+    assert {:foo, _, []} = normalized
+  end
```
josevalim, line 363: Those tests look great but they are all unit tests. Ultimately, the output of normalize is not important if the formatter can't understand it. I think more important would be to unit test that the quoted expressions get formatted properly. Perhaps by copying Macro.to_string suite (or alternatively we should change Macro.to_string suite to the new functionality, then we get all tests for free and we can remove these tests).

josevalim: Actually, we can't replace Macro.to_string because it supports an anonymous function as second argument, but we can copy the tests. :)

josevalim:
This looks great @doorgan, thank you! I am thinking it is probably easiest for us to have a separate PR for the string_to_quoted_with_comments, since the normalize one may take a while. WDYT?

lib/elixir/lib/code/normalizer.ex
doorgan: If we have a list like this one: `[1, {:foo, :bar}, 2, a: :b, c: :d]`, the `{:foo, :bar}` needs to be wrapped in a block because it's a 2-tuple literal, but for the last two only their contents need to be wrapped (`{{:__block__, [format: :keyword], [:key]}, {:__block__, [], [:value]}}`), but I can't decide when to do it without checking if the current element is part of a keyword list or not.

doorgan:
@josevalim I think that's very reasonable :)

lib/elixir/lib/code/normalizer.ex
doorgan: It's a leftover from a previous iteration, it can be safely removed

lib/elixir/lib/code/normalizer.ex
doorgan: Yes, it seems this clause can be removed :)

lib/elixir/lib/code/normalizer.ex
doorgan: I think consistency for those cases would be nice. I realized the problem in this case is more general than `a when b: :c`. If the keyword list has square brackets, then the arguments get wrapped in a block like a regular list, but if the brackets are omitted then the arguments don't get wrapped:
```elixir
# a when b: :c
{:when, [line: 1],
 [
   {:a, [line: 1], nil},
   [
     {{:__block__, [format: :keyword, line: 1], [:b]},
      {:__block__, [line: 1], [:c]}}
   ]
 ]}

# a when [b: :c]
{:when, [line: 1],
 [
   {:a, [line: 1], nil},
   {:__block__, [closing: [line: 1], line: 1],
    [
      [
        {{:__block__, [format: :keyword, line: 1], [:b]},
         {:__block__, [line: 1], [:c]}}
      ]
    ]}
 ]}

# foo(a, b: :c) do :d end
{:foo,
 [
   do: [line: 1],
   end: [line: 1],
   closing: [line: 1],
   line: 1 
 ],
 [
   {:a, [line: 1], nil},
   [
     {{:__block__, [format: :keyword, line: 1], [:b]},
      {:__block__, [line: 1], [:c]}}
   ],
   [
     {{:__block__, [line: 1], [:do]},
      {:__block__, [line: 1], [:d]}}
   ]
 ]}

# foo(a, [b: :c]) do :d end
{:foo,
 [
   do: [line: 1],
   end: [line: 1],
   closing: [line: 1],
   line: 1 
 ],
 [
   {:a, [line: 1], nil},
   {:__block__, [closing: [line: 1], line: 1],
    [
      [
        {{:__block__, [format: :keyword, line: 1], [:b]},
         {:__block__, [line: 1], [:c]}}
      ]
    ]},
   [
     {{:__block__, [line: 1], [:do]}, 
      {:__block__, [line: 1], [:d]}}
   ]
 ]}
```

I did the unwrapping in some places because it depends on the context(a standalone keyword list needs the wrapping but it's not needed in cases like the ones mentioned above). If the parser did that work instead, this could be avoided altogether.
But how can the formatter know the user preference in this case? With the current behavior it is trivial to know if the user omitted the square brackets or not, just check if it's wrapped in a block, but if everything is wrapped in a block there would be no way to tell the difference?

doorgan: Maybe the normalizer can just wrap everything in a block and not do any unwrapping, since it lacks the context to know if the brackets should be omitted or not anyways?

josevalim: The formatter will remove the wrapping brackets when it can, so it should be fine to always wrap it. We will know for sure once we add the more integration like tests. :)

lib/elixir/lib/code/normalizer.ex
doorgan: Done!

lib/elixir/test/elixir/code_normalizer_test.exs
doorgan: I'm a bit lost on how would I use those tests. Macro.to_string can take any quoted expression, and it's output is quite different to that of the formatter. Wouldn't it be better to copy the code_formatter tests, and assert that both Code.format_string and Code.quoted_to_algebra + some_doc_to_string_function produce the same output?

I was doing something similar in the tests for my first iteration of the normalizer but I wasn't sure if it was correct: https://github.com/doorgan/formatter/blob/main/test/integration_test.exs

doorgan: Done :)

doorgan:
I added the tests from `Macro.to_string/2` as requested. However, if we give the "normalied" AST to `quoted_to_algebra`, I'm not sure if the formatter breaks. I think it should already skip normalized literals, but we may need to add tests for that?

josevalim:
Awesome, I will review them tomorrow. Yeah, some unit tests to make sure we don’t double normalize can be good.

lib/elixir/lib/code.ex
```diff
@@ -1192,6 +1192,24 @@ defmodule Code do
     :elixir.string_to_quoted!(to_charlist(string), line, column, file, opts)
   end
 
+  @doc """
+  Converts a quoted expression to an algebra document.
+
+  Supports both regular AST and the extended AST documented in
+  `string_to_quoted_with_comments/2`.
```
josevalim, line 1199: I think this comment is now a little outdated. Instead, let's document you get better results if the quoted expression is built with token_metadata and a literal block encoder:

```elixir
    [
      literal_encoder: &{:ok, {:__block__, &2, [&1]}},
      token_metadata: true
    ]
```

lib/elixir/lib/code/formatter.ex
```diff
@@ -261,6 +257,19 @@ defmodule Code.Formatter do
     }
   end
 
+  @doc """
+  Converts the quoted expression into an algebra document
+  """
+  def quoted_expression_to_algebra(quoted, opts \\ []) do
```
josevalim, line 263: Suggestion: call this `to_algebra` and move part of the logic in the current `to_algebra` to `Code.format_string!`. And make sure that `Code.format_string!` call this function instead.

lib/elixir/lib/code/normalizer.ex
```diff
@@ -0,0 +1,380 @@
...
+    pair =
+      if keyword? and not module_atom? do
+        {_, _, [{left, right}]} = do_normalize({left, right}, parent_meta)
+        left = Macro.update_meta(left, &Keyword.put(&1, :format, :keyword))
+        {left, right}
+      else
+        left = do_normalize(left, parent_meta)
+        right = do_normalize(right, parent_meta)
+        {:__block__, [line: parent_meta[:line]], [{left, right}]}
+      end
```
josevalim, line 343:
```suggestion
    left = do_normalize(left, parent_meta)
    right = do_normalize(right, parent_meta)

    pair =
      if keyword? and not module_atom? do
        left = Macro.update_meta(left, &Keyword.put(&1, :format, :keyword))
        {left, right}
      else
        {:__block__, [line: parent_meta[:line]], [{left, right}]}
      end
```

lib/elixir/test/elixir/code_normalizer_test.exs
```diff
@@ -0,0 +1,476 @@
...
+
+    test "charlist" do
+      assert quoted_to_string(quote(do: [])) == "[]"
+      assert quoted_to_string(quote(do: 'abc')) == "'abc'"
+    end
+
+    test "string" do
+      assert quoted_to_string(quote(do: "")) == ~S/""/
+      assert quoted_to_string(quote(do: "abc")) == ~S/"abc"/
+      assert quoted_to_string(quote(do: "#{"abc"}")) == ~S/"#{"abc"}"/
```
josevalim, line 459: We need to add some tests on this. The issue is that the formatter works on *unescaped* strings. So if the formatter receives "null byte\t\x00", it will keep those as is. However, if we are working with regular quoted expressions, then that will have the literal bytes, and we need to convert them to the escaped representation. This also need to happen to charlists, atoms, and keyword keys.

Furthermore, if the goal of your changes is to make source modification easy, then we need to:

1. Document the `unescape` flag in Code.string_to_quoted

2. `Code.quoted_to_algebra` needs a flag that says if the strings are unescaped or not (either a :unescaped flag that defaults to false or a :escape flag that defaults to true)

lib/elixir/lib/code.ex
josevalim: Also, if we want to support unescape, then we need to list it too. See this comment for more info: https://github.com/elixir-lang/elixir/pull/10988/files#r636739871

josevalim (changes requested):
Fantastic work on the tests! I have added a couple more comments. Also, please rebase against master so we incorporate your and mine latest changes. I think we are *almost* there. :)

lib/elixir/lib/code/normalizer.ex
josevalim: Furthermore, :token is only needed for integer/float, so maybe we should split them into their own clause.

lib/elixir/test/elixir/code_normalizer_test.exs
josevalim: Oh, one last one, the Macro.to_string code probably already has the necessary logic to escape string/atom/charlist characters.

lib/elixir/test/elixir/code_normalizer_test.exs
```diff
@@ -0,0 +1,476 @@
...
+      assert quoted_to_string(quote(do: foo(x: y, p: q))) == "foo(x: y, p: q)"
+      assert quoted_to_string(quote(do: foo(a, x: y, p: q))) == "foo(a, x: y, p: q)"
+
+      assert quoted_to_string(quote(do: {[]})) == "{[]}"
+      assert quoted_to_string(quote(do: {[a: b]})) == "{[a: b]}"
+      assert quoted_to_string(quote(do: {x, a: b})) == "{x, [a: b]}"
+      assert quoted_to_string(quote(do: foo(else: a))) == "foo(else: a)"
+      assert quoted_to_string(quote(do: foo(catch: a))) == "foo(catch: a)"
+    end
+  end
```
josevalim, line 475: Also, let's add some tests to make sure that if we call string_to_quoted_with_comments with the token_metadata+literal_encoder and then we call quoted_to_algebra, the output will be as close as possible to the one from the formatter. In particular, here are some tests of things that should be kept as is:

* code comments
* escape on strings, such as `"\x0A"` should remain as is and not `"\n"`
* floats and integers, such as 1.0e1 and 0b101

We can probably have these tests in their own describe block.

doorgan:
In the tests I wrote a private `quoted_to_string` function. Do you think we shold expose that function as well in the `Code` module while we're at it, should we do that in a separate PR, or not at all?

josevalim:
> In the tests I wrote a private quoted_to_string function. Do you think we shold expose that function as well in the Code module while we're at it, should we do that in a separate PR, or not at all?

My goal is for that to become `Macro.to_string` but let's do it in a separate PR.

lib/elixir/test/elixir/code_normalizer_test.exs
doorgan: > 2. Code.quoted_to_algebra needs a flag that says if the strings are unescaped or not (either a :unescaped flag that defaults to false or a :escape flag that defaults to true)

Why is that needed? As it is now, just passing `unescape: false` is enough to preserve the original strings. I'm porting some of the formatter tests to assert that both `Code.format_string!` and `Code.string_to_quoted_with_comments` + `Code.quoted_to_algebra` produce the same output, and it seems to work as expected.

josevalim: If we want to say Code.quoted_to_algebra works with any quoted expression, then it needs to work with any quoted expression, not only unescaped ones. Otherwise it is really some_very_specific_quoted_to_algebra.

lib/elixir/lib/code/formatter.ex
doorgan: This would make `Code.format_string!` use the normalizer, no? If so we would get the tests suggested in this comment: https://github.com/elixir-lang/elixir/pull/10988#discussion_r636748972 for free, right?

doorgan:
I'm working on escaping strings. If I understand the requirement correctly, strings such as `"null byte\t\x00"` should be returned as is, and not turned into a bitstring like `<<110, 117, 108, 108, 32, 98, 121, 116, 101, 9, 0>>`.

It seems that most of this work is done by `Code.Identifier.escape`(which is used internally by `Macro.to_string`), and since escaped sequences like `\u0000` get converted to a single character(in this case the byte `0`), we can't go back to the original string in all cases(for this particular case we can't know if the sequence was `\u0000` or `\x00`, so the escape function will turn that character into a `\0`). The same happens for sequences like `\s` that translate to an empty space, and we can't know if the space came from the space character or an escaped sequence.

I'm using `Code.Identifier.escape` for now, but I would like to know if you think there's a better way to address this,

lib/elixir/lib/code/formatter.ex
doorgan: Done!

josevalim:
> I'm working on escaping strings. If I understand the requirement correctly, strings such as "null byte\t\x00" should be returned as is, and not turned into a bitstring like <<110, 117, 108, 108, 32, 98, 121, 116, 101, 9, 0>>.

Sorry, I was wrong. The important is for us to return the same as Macro.to_string, this means:

```elixir
iex(1)> Macro.to_string "tab: \t"
"\"tab: \\t\""
iex(2)> Macro.to_string "null: \x00"
"<<110, 117, 108, 108, 58, 32, 0>>"
```

lib/elixir/lib/code.ex
```diff
@@ -980,8 +980,25 @@ defmodule Code do
...
+
+    to_quoted_opts =
+      [
+        unescape: false,
+        warn_on_unnecessary_quotes: false,
+        literal_encoder: &{:ok, {:__block__, &2, [&1]}},
+        token_metadata: true
+      ] ++ opts
+
+    {forms, comments} = Code.string_to_quoted_with_comments!(string, to_quoted_opts)
```
josevalim, line 992:
```suggestion
    {forms, comments} = string_to_quoted_with_comments!(string, to_quoted_opts)
```

lib/elixir/lib/code.ex
```diff
@@ -1307,6 +1329,41 @@ defmodule Code do
...
 
+  @doc """
+  Converts a quoted expression to an algebra document.
+
+  The elixir AST does not contain metadata for literals like strings, lists, or
+  tuples with two elements, which means that the produced algebra document will
+  not respect all of the user preferences and comments may be misplaced.
+  To get better results, you can use the `:token_metadata`, `:escape` and
+  `:literal_encoder` options to `string_to_quoted/2` to provide additional
+  information to the formatter:
```
josevalim, line 1340:
```suggestion
  The Elixir AST does not contain metadata for literals like strings, lists, or
  tuples with two elements, which means that the produced algebra document will
  not respect all of the user preferences and comments may be misplaced.
  To get better results, you can use the `:token_metadata`, `:unescape` and
  `:literal_encoder` options to `string_to_quoted/2` to provide additional
  information to the formatter:
```

lib/elixir/lib/code.ex
```diff
@@ -1307,6 +1329,41 @@ defmodule Code do
...
+        token_metadata: true,
+        unescape: false
+      ]
+
+  This will produce an AST that contains information such as `do` blocks start
+  and end lines or sigil delimiters, and by wrapping literals in blocks they can
+  now hold metadata like line number, string delimiter and escaped sequences, or
+  integer formatting(ie `0x2a` instead of `47`). This is especially useful if
+  you're doing source code manipulation, where it's important to preserve user
+  choices and comments placing.
```
josevalim, line 1353:
```suggestion
  This will produce an AST that contains information such as `do` blocks start
  and end lines or sigil delimiters, and by wrapping literals in blocks they can
  now hold metadata like line number, string delimiter and escaped sequences, or
  integer formatting (such as `0x2a` instead of `47`). However, **note this AST is
  not valid**. If you evaluate it, it won't have the same semantics as the regular
  Elixir AST due to the `:unescape` and `:literal_encoder` options. However,
  those options are useful if you're doing source code manipulation, where it's
  important to preserve user choices and comments placing.
```

lib/elixir/lib/code.ex
```diff
@@ -1100,6 +1117,11 @@ defmodule Code do
     * `:columns` - when `true`, attach a `:column` key to the quoted
       metadata. Defaults to `false`.
 
+    * `:unescape` - when `false`, preserves escaped sequences. For example,
+      `"null byte\\t\\x00"` will be kept as is instead of being converted to a
+      bit string literal. This is useful in combination with
+      `quoted_to_algebra/2`. Defaults to `true`.
```
josevalim, line 1123:
```suggestion
    * `:unescape` (since v1.10.0) - when `false`, preserves escaped sequences.
      For example, `"null byte\\t\\x00"` will be kept as is instead of being converted
      to a bitstring literal. Note if you set this option to false, the resulting AST is no
      longer valid, but it can be useful to analyze/transform source code, typically in
      in combination with `quoted_to_algebra/2`. Defaults to `true`.
```

lib/elixir/test/elixir/test_helper.exs
```diff
@@ -83,6 +83,36 @@ defmodule CodeFormatterHelpers do
...
+      to_algebra_opts = [comments: comments] ++ opts
+
+      doc = Code.quoted_to_algebra(quoted, to_algebra_opts)
+
+      actual =
+        Inspect.Algebra.format(doc, line_length)
+        |> IO.iodata_to_binary()
+
+      assert actual == expected
+    end
```
josevalim, line 112:
```elixir
    quote bind_quoted: [good: good, opts: opts], location: :keep do
      assert IO.iodata_to_binary(Code.format_string!(good, opts)) ==
               CodeNormalizerHelpers.string_to_string(good, opts)
    end
  end

  def string_to_string(good, opts) do
    line_length = Keyword.get(opts, :line_length, 98)
    good = String.trim(good)

    to_quoted_opts =
      [
        literal_encoder: &{:ok, {:__block__, &2, [&1]}},
        token_metadata: true,
        unescape: false
      ] ++ opts

    {quoted, comments} = Code.string_to_quoted_with_comments!(good, to_quoted_opts)

    to_algebra_opts = [comments: comments] ++ opts

    quoted
    |> Code.quoted_to_algebra(to_algebra_opts)
    |> Inspect.Algebra.format(line_length)
    |> IO.iodata_to_binary()
  end
```

lib/elixir/test/elixir/code_normalizer/literals_test.exs
```diff
@@ -0,0 +1,304 @@
...
+      '''
+      one
+      "
+      #{2}
+      "
+      three
+      '''
+      """
+    end
+  end
```
josevalim, line 303: We need some tests for sigils too. You can copy the string tests and use ~S to build the strings. :)

lib/elixir/lib/code.ex
```diff
@@ -1307,6 +1331,43 @@ defmodule Code do
...
+  integer formatting (such as `0x2a` instead of `47`). However, **note this AST is
+  not valid**. If you evaluate it, it won't have the same semantics as the regular
+  Elixir AST due to the `:unescape` and `:literal_encoder` options. However,
+  those options are useful if you're doing source code manipulation, where it's
+  important to preserve user choices and comments placing.
+
+  ## Options
+
+    * `:comments` - the list of comments associated with the quoted expression.
+      Defaults to `[]`.
```
josevalim, line 1362: We need to document the :escape option.

josevalim:
Hi @doorgan, I have updated the docs and posted some final comments! I believe only the escape tests are missing, to sum up the above:

1. Escaping of strings, with printable characters and non-printable characters (we need to choose if we will show non-printable as binaries or as escaped strings). Perhaps it makes sense to keep them as strings, because if they were binaries, their AST would be different

2. Escaping of charlists, with ascii printable (keep as charlists) and non-ascii (keep as lists)

3. Escaping of atoms

4. Non-escaping of sigils. Sigils are always unescaped, so we need to add some tests that we won't ever escape sigils, regardless of the escape option

michallepicki:
Hi @doorgan
Dialyzer [found](https://github.com/michallepicki/elixir-lang-dialyzer-runs/runs/2643710827) two issues related to string_to_quoted_with_comments on master, I see you fixed one of them (bad pattern match), I have sent a PR to your branch to fix the other one (wrong return type in spec): https://github.com/doorgan/elixir/pull/1

lib/elixir/test/elixir/code_normalizer/general_test.exs
```diff
@@ -0,0 +1,616 @@
...
+
+      assert quoted_to_string(quote(do: :"foo\n#{bar}\tbaz")) == ~S[:"foo\n#{bar}\tbaz"]
+
+      assert quoted_to_string(quote(do: :"foo\"bar"), escape: false) == ~s[:"foo\\"bar"]
+      assert quoted_to_string(quote(do: :"foo\"bar")) == ~S[:"foo\\"bar"]
+
+      assert quoted_to_string(quote(do: :"one\n\"#{2}\"\nthree"), escape: false) ==
+               ~s[:"one\n\\"\#{2}\\"\nthree"]
+
+      assert quoted_to_string(quote(do: :"one\n\"#{2}\"\nthree")) == ~S[:"one\n\\"#{2}\\"\nthree"]
```
doorgan, line 527: I have a doubt regarding the escaping/unescaping of atom interpolations. It seems that some characters like `\n` are escaped correctly, but others like `\"` get the `\` escaped twice.
In these lines I had to put double backward slashes on the right hand sides because of this, but while it is the same escaping that `Macro.to_string` does, it's not exactly the same behavior as what `:unescape` option of `Code.string_to_quoted` produces.
Is this alright?

josevalim: The above is definitely wrong. This:

```elixir
:"one\n\\"#{2}\\"\nthree"
```

Is actually the same as:

```elixir
:"one\n\\"       #{2}\\"\nthree"
```

So the #{... is being treated as a comment. The one with `escape: false` is also wrong for the same reason.

lib/elixir/test/elixir/code_normalizer/general_test.exs
```diff
@@ -0,0 +1,616 @@
...
+      assert quoted_to_string(quote(do: '\x00\x01\x10'), escape: false) == ~S/[0, 1, 16]/
+      assert quoted_to_string(quote(do: '\x00\x01\x10')) == ~S/[0, 1, 16]/
+    end
+
+    test "atoms" do
+      assert quoted_to_string(quote(do: :"a\nb\tc"), escape: false) == ~s/:"a\nb\tc"/
+      assert quoted_to_string(quote(do: :"a\nb\tc")) == ~S/:"a\nb\tc"/
+    end
+
+    test "atoms with interpolations" do
```
josevalim, line 515: We also need a test of atoms with non printable characters (they are kept as strings). :)

lib/elixir/test/elixir/code_normalizer/general_test.exs
```diff
@@ -0,0 +1,616 @@
...
+      assert quoted_to_string(quote(do: :"foo\n#{bar}\tbaz"), escape: false) ==
+               ~s[:"foo\n\#{bar}\tbaz"]
+
+      assert quoted_to_string(quote(do: :"foo\n#{bar}\tbaz")) == ~S[:"foo\n#{bar}\tbaz"]
+
+      assert quoted_to_string(quote(do: :"foo\"bar"), escape: false) == ~s[:"foo\\"bar"]
```
josevalim, line 521:
```suggestion
      assert quoted_to_string(quote(do: :"foo\"bar"), escape: false) == ~S[:"foo\"bar"]
```

?

lib/elixir/lib/code/normalizer.ex
```diff
@@ -0,0 +1,456 @@
...
+  defp normalize_list_elements([], _state, _keyword?) do
+    []
+  end
+
+  defp maybe_escape_literal(string, %{escape: true}) when is_binary(string) do
+    if String.printable?(string) do
+      {string, _} = Code.Identifier.escape(string, ?")
+      IO.iodata_to_binary(string)
+    else
+      String.trim(inspect(string), "\"")
```
josevalim, line 417: Why do we need this trim? 🤔

doorgan: Actually it's not needed, I will remove it

lib/elixir/lib/code/normalizer.ex
```diff
@@ -0,0 +1,456 @@
...
+  defp do_normalize(
+         {{:., dot_meta, [:erlang, :binary_to_atom]}, call_meta, [{:<<>>, _, _} = string, :utf8]},
+         state
+       ) do
+    dot_meta = patch_meta_line(dot_meta, state.parent_meta)
+    call_meta = patch_meta_line(call_meta, dot_meta)
+
+    string =
+      if state.escape do
+        do_normalize(string, %{state | escape_interpolation: true})
```
josevalim, line 98: You need to be careful with this kind of flags because you can have this:

    :"foo#{~s/\n/}bar"

And escaping the interpolation inside the atom will now force the interpolation to be escaped inside the sigil. It is probably best to, once you know you have a `{:<<>>, ..., ...}`, you call a explicit `do_normalize_interpolation` with the escape as argument.

lib/elixir/lib/code/normalizer.ex
```diff
@@ -0,0 +1,456 @@
...
+       {do_normalize(left, left_state), do_normalize(right, state)}
+     ]}
+  end
+
+  # Lists
+  defp do_normalize(list, state) when is_list(list) do
+    if list != [] and List.ascii_printable?(list) do
+      # It's a charlist
+      list =
+        if state.escape == true do
```
josevalim, line 245:
```suggestion
        if state.escape do
```

lib/elixir/lib/code/normalizer.ex
josevalim: You also need to remember inspect has a limit of characters, so we need to disable any limit. Do we need to use inspect here? Could we always use Code.Identifier.escape instead? It is probably better to format "\u0000" as "\0" instead of a binary... especially because if we have a string, it was likely a string originally too.

doorgan: You're right, I got confused and tried to replicate `Macro.to_string` behavior and I did not consider that we want to keep them as strings to not change the semantics of the output. This should be fixed now :)

lib/elixir/lib/code/normalizer.ex
doorgan: I moved the bitstring normalization code into its own function with an explicit argument to avoid that issue

lib/elixir/test/elixir/code_normalizer/general_test.exs
doorgan: Added!

lib/elixir/test/elixir/code_normalizer/general_test.exs
doorgan: I fixed the test so it expects the correct output. It is now failing as it should.

I will try to fix this tomorrow.

doorgan: I think it its fixed now c:

lib/elixir/lib/code/normalizer.ex
```diff
@@ -0,0 +1,470 @@
...
+     ]}
+  end
+
+  # Lists
+  defp do_normalize(list, state) when is_list(list) do
+    if list != [] and List.ascii_printable?(list) do
+      # It's a charlist
+      list =
+        if state.escape do
+          {string, _} = Code.Identifier.escape(IO.chardata_to_string(list), -1)
```
josevalim, line 251: Nice trick with -1 and it makes sense too. The formatter will escape the escape character after all.

lib/elixir/lib/code/normalizer.ex
```diff
@@ -0,0 +1,470 @@
...
+  defp normalize_args([first | rest], state, keyword?) do
+    [do_normalize(first, state) | normalize_args(rest, state, keyword?)]
+  end
+
+  defp normalize_args([], _state, _keyword?) do
+    []
+  end
+
+  defp maybe_escape_literal(string, %{escape: true}) when is_binary(string) do
+    {string, _} = Code.Identifier.escape(string, nil)
```
josevalim, line 431: For consistency?

```suggestion
    {string, _} = Code.Identifier.escape(string, -1)
```

josevalim (approved):

josevalim:
Amazing work! :green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

josevalim:
Btw, I am exploring moving Macro.to_string  to the new function, I will keep you posted. :)

PR #10990, opened 2021-05-19
Add `Code.string_to_quoted_with_comments/2`

doorgan:
Split from #10988

I created this as a draft because I'm getting an `undef` error when running `make clean test` that I didn't figure out how to solve yet,

### DIFF
--- a/lib/elixir/lib/code.ex
+++ b/lib/elixir/lib/code.ex
@@ -1330,7 +1330,7 @@ defmodule Code do
   end
 
   defp previous_eol([]), do: 1
-  defp previous_eol(_), do: nil
+  defp previous_eol(_), do: 0
 
   @doc """
   Evals the given file.
--- a/lib/elixir/lib/code.ex
+++ b/lib/elixir/lib/code.ex
@@ -1206,32 +1206,6 @@ defmodule Code do
   Returns `{:ok, quoted_form, comments}` if it succeeds,
   `{:error, {line, error, token}}` otherwise.
 
-  The AST returned differs from regular AST in that literals are wrapped in
-  a block to preserve their metadata. For example, the literal `"foo"` will be
-  converted to `{:__block__, [line: 1, delimiter: "\"", token: "\"foo\""], ["foo"]}`.
-
-  In some cases the wrapping is not uniform, for example in the case of keyword
-  lists the wrapping may change if the square were ommitted:
-      iex> Code.string_to_quoted_with_comments("[foo: :bar]")
-      {:ok, {
-        :__block__, [closing: [line: 1], line: 1], [[
-          {{:__block__, [format: :keyword, line: 1], [:foo]}, {:__block__, [line: 1], [:bar]}}
-        ]]
-      }, []}
-
-      iex> Code.string_to_quoted_with_comments("a when foo: :bar")
-      {:ok, {
-        :when, [line: 1], [
-          {:a, [line: 1], nil},
-          [
-            {{:__block__, [format: :keyword, line: 1], [:foo]}, {:__block__, [line: 1], [:bar]}}
-          ]
-        ]
-      }, []}
-
-  Note that in the second example the outer wrapping of the keyword list is
-  omitted.
-
   Comments are maps containing information about the line they were found, its
   contents, and how many end of lines were found between the comment and the
   closest tokens:
@@ -1243,16 +1217,12 @@ defmodule Code do
       ...>
       ...> # Some more comments!
       ...> "\"")
-      {:ok, {:__block__, [line: 1], [:foo]}, [
+      {:ok, :foo, [
         %{line: 3, previous_eol: 2, next_eol: 3, text: "\# Hello, world!"},
         %{line: 6, previous_eol: 3, next_eol: 1, text: "\# Some more comments!"},
       ]}
 
-  ## Options
-    * `:file` - the filename to be reported in case of parsing errors.
-      Defaults to "nofile".
-    * `:line` - the starting line of the string being parsed.
-      Defaults to 1.
+  Check `string_to_quoted/2` for options information.
   """
   @spec string_to_quoted_with_comments(List.Chars.t(), keyword) ::
           {:ok, Macro.t(), map()} | {:error, {location :: keyword, term, term}}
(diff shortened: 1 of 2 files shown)


doorgan:
I can't figure out how to fix the error I'm getting when running `mix clean test`. It's the same error than the one in the failed CI tasks.
I have to checkout `master` and run `make clean test` first and then checkout the PR branch to be able to run the tests.

Is it ok to ask here how would I fix that or would it be better if I open an issue instead?

josevalim:
@doorgan it is a bootstrap issue. You can't use `|>` in the `Code` module because the functionality to make `|>` work is not yet there when we are compiling `Code` for the first time.

lib/elixir/lib/code.ex
```diff
@@ -1197,6 +1197,188 @@ defmodule Code do
...
+
+  This function is useful when performing textual changes to the source code,
+  while preserving information like comments and literals position.
+
+  Returns `{:ok, quoted_form, comments}` if it succeeds,
+  `{:error, {line, error, token}}` otherwise.
+
+  The AST returned differs from regular AST in that literals are wrapped in
+  a block to preserve their metadata. For example, the literal `"foo"` will be
+  converted to `{:__block__, [line: 1, delimiter: "\"", token: "\"foo\""], ["foo"]}`.
```
josevalim, line 1211: I don't think we should do this by default. It should be an option, like you pass to `string_to_quoted`, and we make the comments work even without this?

doorgan: So, the function would return regular ast + comments, and for the formatter use cases we would still pass the `literal_encoder` option?

josevalim: Yes. literal_encoder, token_metadata, etc should all be explicitly given (as in string_to_quoted). We will only set preserve_comments (which is private).

lib/elixir/lib/code.ex
```diff
@@ -1197,6 +1197,188 @@ defmodule Code do
...
+      ...> :foo
+      ...>
+      ...> # Hello, world!
+      ...>
+      ...>
+      ...> # Some more comments!
+      ...> "\"")
+      {:ok, {:__block__, [line: 1], [:foo]}, [
+        %{line: 3, previous_eol: 2, next_eol: 3, text: "\# Hello, world!"},
+        %{line: 6, previous_eol: 3, next_eol: 1, text: "\# Some more comments!"},
```
josevalim, line 1248: In some cases previous_eol can be nil, right? Do you know when? Should we make it default to zero instead? 🤔

doorgan: I think it's already defaulting to `2` in line 1354?

josevalim: I see. I think gather_comments and format_comment should remain in the formatter because they are formatter specific behaviour. The code itself should return the comments as close to how it was parsed as possible. WDYT?

And then we make previous_eol default to 0 and convert 0 (instead of nil) to 2.

That's the tricky part about these changes, we need to export all of those with values that make sense outside of the formatter, so we can't take as many shortcuts. :)

doorgan: I agree, I will get to it now :)

doorgan: Done!

> Do you know when?
It seems to happen when the comment is in the same line as the code, for example:
```elixir
foo # Hello!
```

lib/elixir/lib/code.ex
doorgan: Done!

lib/elixir/lib/code.ex
josevalim: Ah, so it should be 0 indeed. Perfect.

lib/elixir/lib/code.ex
```diff
@@ -1197,6 +1197,105 @@ defmodule Code do
...
+
+  This function is useful when performing textual changes to the source code,
+  while preserving information like comments and literals position.
+
+  Returns `{:ok, quoted_form, comments}` if it succeeds,
+  `{:error, {line, error, token}}` otherwise.
+
+  Comments are maps containing information about the line they were found, its
+  contents, and how many end of lines were found between the comment and the
+  closest tokens:
```
josevalim, line 1211:
```suggestion
  closest tokens:

```

lib/elixir/lib/code.ex
```diff
@@ -1197,6 +1197,105 @@ defmodule Code do
...
+      ...> :foo
+      ...>
+      ...> # Hello, world!
+      ...>
+      ...>
+      ...> # Some more comments!
+      ...> "\"")
+      {:ok, :foo, [
+        %{line: 3, previous_eol: 2, next_eol: 3, text: "\# Hello, world!"},
+        %{line: 6, previous_eol: 3, next_eol: 1, text: "\# Some more comments!"},
```
josevalim, line 1222: Let's also add an example where previous_eol is zero. It can be a separate example so this one does not grow to large. Something like: `Code.string_to_quoted_with_comments(":foo # :bar")`.

josevalim:
Perfect, I have added only some minor comments. :)

lib/elixir/lib/code.ex
```diff
@@ -1197,6 +1197,105 @@ defmodule Code do
...
+  Converts the given string to its quoted form and a list of comments.
+
+  This function is useful when performing textual changes to the source code,
+  while preserving information like comments and literals position.
+
+  Returns `{:ok, quoted_form, comments}` if it succeeds,
+  `{:error, {line, error, token}}` otherwise.
+
+  Comments are maps containing information about the line they were found, its
+  contents, and how many end of lines were found between the comment and the
```
whatyouhide, line 1210:
```suggestion
  Comments are maps containing information about the line they were found, their
  contents, and how many end of lines were found between the comment and the
```

lib/elixir/lib/code.ex
```diff
@@ -1197,6 +1197,105 @@ defmodule Code do
...
+  """
+  @spec string_to_quoted_with_comments(List.Chars.t(), keyword) ::
+          {:ok, Macro.t(), map()} | {:error, {location :: keyword, term, term}}
+  def string_to_quoted_with_comments(string, opts \\ [])
+      when is_binary(string) and is_list(opts) do
+    file = Keyword.get(opts, :file, "nofile")
+    line = Keyword.get(opts, :line, 1)
+    column = Keyword.get(opts, :column, 1)
+
+    charlist = String.to_charlist(string)
```
whatyouhide, line 1235: You're calling `String.to_charlist/1` on this so it seems like the spec should be `String.t()`, but the rest of the functions in this module use `List.Chars.t()`. Call `to_charlist/1` instead of `String.to_charlist/1` :)

```suggestion
    charlist = to_charlist(string)
```

lib/elixir/lib/code.ex
```diff
@@ -1197,6 +1197,105 @@ defmodule Code do
...
+      iex> Code.string_to_quoted_with_comments("\""
+      ...> :foo
+      ...>
+      ...> # Hello, world!
+      ...>
+      ...>
+      ...> # Some more comments!
+      ...> "\"")
+      {:ok, :foo, [
+        %{line: 3, previous_eol: 2, next_eol: 3, text: "\# Hello, world!"},
```
whatyouhide, line 1221: It took me a while to understand what `previous_eol`/`next_eol` meant. I think we need to mention the keys by name in the docs and I would suggest to rename them to something else. Some alternatives:

* `previous_eol_count`/`next_eol_count`
* `previous_eols`/`next_eols`

Thoughts?

doorgan: Maybe they could be renamed to `previous_eols`/`next_eols` and the docs could explain what each field represents:
> * `:line` - The line number the source code
> * `:text` - The full text of the comment, incluing the leading `#`
> * `:previous_eols` - How many end of lines there are between the comment and the previous node
> * `:next_eols` - How many end of lines there are between the comment and the next node

Another alternative could be `leading_eols`/`trailing_eols`?

josevalim: The docs are definitely welcome! On leading vs trailing eols, it also works. @whatyouhide any preferences?

whatyouhide: I’m fine with both but I personally read better with `previous_eol_count`, but I don't care too much :) But the docs are definitely welcome, agreed.

doorgan: Done :)

josevalim:
Beautiful work, thank you! :green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

PR #11002, opened 2021-05-23
Fix ast normalizing issue with keyword lists

### LINKED ISSUE #11001: line in comments is not used in Code.quoted_to_algebra/2
Hi, I'm too excited with latest changes and just can't wait for `release candidate` for next `Elixir` version. :smiling_imp: 

### Environment

```elixir
Erlang/OTP 23 [erts-11.2.2] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [hipe]

Elixir 1.13.0-dev (c6145d0) (compiled with Erlang/OTP 23)
```
* Operating system: 
```bash
$ uname -a
Linux archlinux 5.12.1-arch1-1 #1 SMP PREEMPT Sun, 02 May 2021 12:43:58 +0000 x86_64 GNU/Linux

$ cat /etc/os-release
NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://www.archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://bugs.archlinux.org/"
LOGO=archlinux
```

### Current behavior

`line` in `comments` is not (properly) used in `Code.quoted_to_algebra/2`, see:
```elixir
iex> source = """
defmodule Example do
  # something
  def sample, do: :ok
end
"""

iex> {:ok, quoted, comments} = Code.string_to_quoted_with_comments source
{:ok,
 {:defmodule, [line: 1],
  [
    {:__aliases__, [line: 1], [:Example]},
    [do: {:def, [line: 3], [{:sample, [line: 3], nil}, [do: :ok]]}]
  ]},
 [%{line: 2, next_eol_count: 1, previous_eol_count: 1, text: "# something"}]}

iex> algebra = Code.quoted_to_algebra quoted, comments: comments
{:doc_force,
 {:doc_cons, {:doc_group, "# something", :self},
  {:doc_cons, :doc_line,
   {:doc_group,
    {:doc_cons, {:doc_string, "defmodule", 9},
     {:doc_force,
      {:doc_cons,
       {:doc_cons,
        {:doc_group,
         {:doc_cons, {:doc_cons, " ", {:doc_nest, "Example", :cursor, :break}},
          " do"}, :self},
        {:doc_nest,
         {:doc_cons, :doc_nil,
          {:doc_cons, :doc_line,
           {:doc_group,
            {:doc_group,
             {:doc_cons, {:doc_string, "def", 3},
              {:doc_force,
               {:doc_cons,
                {:doc_cons,
                 {:doc_group,
                  {:doc_cons, {:doc_cons, " ", {:doc_nest, {...}, ...}}, " do"},
                  :self},
                 {:doc_nest,
                  {:doc_cons, :doc_nil, {:doc_cons, :doc_line, {...}}}, 2,
                  :always}}, {:doc_cons, :doc_line, "end"}}}}, :self}, :self}}},
         2, :always}}, {:doc_cons, :doc_line, "end"}}}}, :self}}}}

iex> algebra |> Inspect.Algebra.format(80) |> IO.puts()
# something
defmodule Example do
  def sample do
    :ok
  end
end
```

### Expected behavior

The output should be as same as from `source` variable.

doorgan:
Should fix #11001

### DIFF
--- a/lib/elixir/lib/code/normalizer.ex
+++ b/lib/elixir/lib/code/normalizer.ex
@@ -1,6 +1,94 @@
 defmodule Code.Normalizer do
   @moduledoc false
 
+  # Operators that are logical cannot be mixed without parens
+  @required_parens_forms [:||, :|||, :or, :&&, :&&&, :and, :{}]
+
+  @locals_without_parens [
+    # Special forms
+    alias: 1,
+    alias: 2,
+    case: 2,
+    cond: 1,
+    for: :*,
+    import: 1,
+    import: 2,
+    quote: 1,
+    quote: 2,
+    receive: 1,
+    require: 1,
+    require: 2,
+    try: 1,
+    with: :*,
+
+    # Kernel
+    def: 1,
+    def: 2,
+    defp: 1,
+    defp: 2,
+    defguard: 1,
+    defguardp: 1,
+    defmacro: 1,
+    defmacro: 2,
+    defmacrop: 1,
+    defmacrop: 2,
+    defmodule: 2,
+    defdelegate: 2,
+    defexception: 1,
+    defoverridable: 1,
+    defstruct: 1,
+    destructure: 2,
+    raise: 1,
+    raise: 2,
+    reraise: 2,
+    reraise: 3,
+    if: 2,
+    unless: 2,
+    use: 1,
+    use: 2,
+
+    # Stdlib,
+    defrecord: 2,
+    defrecord: 3,
+    defrecordp: 2,
+    defrecordp: 3,
+
+    # Testing
+    assert: 1,
+    assert: 2,
+    assert_in_delta: 3,
+    assert_in_delta: 4,
+    assert_raise: 2,
+    assert_raise: 3,
+    assert_receive: 1,
+    assert_receive: 2,
+    assert_receive: 3,
+    assert_received: 1,
+    assert_received: 2,
+    doctest: 1,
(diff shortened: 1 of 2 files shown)


doorgan:
Actually this is a bit more complicated than I thought, I will un-draft this as soon as I figure out how to fix the omitted kw list brackets issue

doorgan:
The normalizer seems to remove keyword list brackets only if they are the last argument of a call, and the call is not an operator that requires brackets to make sense, ie `[:||, :|||, :or, :&&, :&&&, :and]` and the special form `:{}`. If we want to remove brackets in every other case, we need to check for that. So I added the `@required_parens_forms` for this.

Also, there are calls that always omit the parenthesis, in the formatter code they are listed in the `@locals_without_parens` module attribute and the check is made with the private `local_without_parens?/2` function. I copied those over to the normalizer so I can properly normalize the AST.

This way it seems to work fine(it passes the tests at least). Thoughts?

josevalim:
Thank you @doorgan! Instead of hard-coding the operators, we can call Code.Identifier.unary_op/binary_op, but i will check those locally. Thanks!

josevalim:

PR #11004, opened 2021-05-25
Preserve metadata on regular blocks

doorgan:
This is work in progress. I found this bug while playing around with the idea of a `mix deps.add` task. Blocks were normalized as if they were lists, and metadata was lost, breaking the comments placement.


josevalim:

PR #11006, opened 2021-05-25
Remove dead comment

v0idpwn:
I'm assuming it was a mistake. If it wasn't, I may close this PR


josevalim:
It is for getting information about codepoints, useful when updating the unicode database. :)

PR #11007, opened 2021-05-25
Prevent 2-tuples from being normalized twice

doorgan:
I first thought that #11004 was related to blocks with a single expression being badly normalized, but the issue was that 2-tuples were normalized twice.


doorgan:
PS: do you prefer that I send many little PRs like this one as I find new bugs, or do you prefer if I wait a bit and collect them in a single PR?

josevalim:
PR as you find bugs! The most
Important are the regression tests :)

josevalim:

PR #11008, opened 2021-05-25
Compile time error when aliasing non-Elixir modules without :as

### LINKED ISSUE #11000: Misleading warning when aliasing erlang module without as
### Environment

* Elixir & Erlang/OTP versions (elixir --version): 1.12 and earlier

### Current behavior

```
defmodule My do
  alias :lists
end
```
results in
```
warning: unused alias :"Elixir.lists"
```

### Expected behavior

A more suiting warning or an error.

kelvinst:
Fixes #11000 

Not proficient erlang dev here, so a detailed review would really be welcome.

### DIFF
--- a/lib/elixir/src/elixir_expand.erl
+++ b/lib/elixir/src/elixir_expand.erl
@@ -69,7 +69,17 @@ expand({alias, Meta, [Ref, Opts]}, E) ->
 
   if
     is_atom(ERef) ->
-      {ERef, expand_alias(Meta, true, ERef, EOpts, ET)};
+      case lists:keyfind(as, 1, Opts) of
+        {as, _} ->
+          {ERef, expand_alias(Meta, true, ERef, EOpts, ET)};
+        _->
+          case atom_to_list(ERef) of
+            "Elixir." ++ ([FirstLetter | _]) when FirstLetter >= $A, FirstLetter =< $Z ->
+              {ERef, expand_alias(Meta, true, ERef, EOpts, ET)};
+            _ ->
+              form_error(Meta, E, ?MODULE, {invalid_alias_module, Ref})
+          end
+      end;
     true ->
       form_error(Meta, E, ?MODULE, {expected_compile_time_module, alias, Ref})
   end;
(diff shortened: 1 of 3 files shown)


lib/elixir/src/elixir_expand.erl
```diff
@@ -69,7 +69,17 @@ expand({alias, Meta, [Ref, Opts]}, E) ->
...
+      case lists:keyfind(as, 1, Opts) of
+        {as, _} ->
+          {ERef, expand_alias(Meta, true, ERef, EOpts, ET)};
+        _->
+          case atom_to_list(ERef) of
+            "Elixir." ++ ([FirstLetter | _]) when FirstLetter >= $A, FirstLetter =< $Z ->
+              {ERef, expand_alias(Meta, true, ERef, EOpts, ET)};
+            _ ->
+              form_error(Meta, E, ?MODULE, {invalid_alias_module, Ref})
+          end
```
josevalim, line 81: This looks good but you likely want to do it on elixir_aliases:last, by changing it to return `{:ok, _}` or `:error`, and then you change the callers of `elixir_aliases:last` accordingly.

lib/elixir/test/elixir/kernel/expansion_test.exs
```diff
@@ -154,6 +154,13 @@ defmodule Kernel.ExpansionTest do
       assert_raise CompileError, message, fn ->
         expand(quote(do: alias(:lists, as: :"Elixir.foobar")))
       end
+
+      message =
+        ~r"invalid alias module, expected an Elixir module, got: :lists. To alias other BEAM languages modules, use the :as option"
```
josevalim, line 159:
```suggestion
        ~r"invalid alias module, expected an Elixir module, got: :lists. To alias non-Elixir modules, use the :as option"
```

lib/elixir/src/elixir_expand.erl
kelvinst: Sure, will refactor it to that, thanks for the heads up!

lib/elixir/src/elixir_expand.erl
```diff
@@ -1116,6 +1126,9 @@ format_error(unhandled_arrow_op) ->
   "unhandled operator ->";
 format_error(as_in_multi_alias_call) ->
   ":as option is not supported by multi-alias call";
+format_error({invalid_alias_module, Ref}) ->
+  io_lib:format("invalid alias module, expected an Elixir module, got: ~ts. To alias other BEAM languages modules, use the :as option",
```
fertapric, line 1130: I would say something like "alias cannot be inferred automatically, please use the :as option"

kelvinst: Good one, will update

kelvinst:
All suggestions implemented, let me know if you want me to change anything else! Thanks

josevalim:

PR #11019, opened 2021-05-28
OptionParser.parse/2: Validate switch types/modifiers

### LINKED ISSUE #11017: OptionParser.parse/2 silently accepts invalid modifiers
### Environment

### Current behavior

If we use an invalid modifier, e.g.: `:atom` or `:number` which one might expect are supported but actually aren't, it is simply ignored and the option is considered valid:

```elixir
iex> OptionParser.parse(~w(--proglang foo), strict: [proglang: :atom])
{[proglang: "foo"], [], []}
```

only if we use the proper modifier and the option doesn't conform to that, the option is considered invalid:

```elixir
iex> OptionParser.parse(~w(--proglang foo), strict: [proglang: :integer])
{[], [], [{"--proglang", "foo"}]}
```

### Expected behavior

I think we should either raise an ArgumentError or if that would be considered a breaking change, at least print a warning.

wojtekmach:
Closes #11017

### DIFF
--- a/lib/elixir/lib/option_parser.ex
+++ b/lib/elixir/lib/option_parser.ex
@@ -596,9 +596,11 @@ defmodule OptionParser do
           raise ArgumentError, ":switches and :strict cannot be given together"
 
         switches = opts[:switches] ->
+          validate_switches(switches)
           {switches, false}
 
         strict = opts[:strict] ->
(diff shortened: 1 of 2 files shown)


josevalim:

PR #11024, opened 2021-05-31
Depend on Erlang for computing grapheme clusters

josevalim:
Erlang ships with its own embedding of the Unicode
Codebase for a couple releases and this release changes
Elixir to depend on it in order to compute grapheme
clusters.

The Erlang implementation was up to 2x faster in low
codepoints (such as latin1) while the Elixir one could
be faster up to 3x in high codepoints (such as emoji)
so at the end the performance results are roughly the
same. As a benefit, we no longer need to ship our copy
of the grapheme cluster algorithm, which would take up
to 250kB in disk and more than 15 seconds to compile.

Note we still keep our own String downcase and upcase
algorithms, as our version is considerably more efficient
on all cases since it works exclusively with binaries
(more than 5x faster).


lib/elixir/test/elixir/string_test.exs
```diff
@@ -881,4 +881,20 @@ defmodule StringTest do
     # FB13;FB13;FB13;0574 0576;0574 0576;
     assert String.normalize("\u0574\u0576", :nfkc) == "\u0574\u0576"
   end
+
+  # The way carriage return works makes it problematic in many
+  # cases which we test here.
```
josevalim, line 886:
```suggestion
  # Carriage return can be a grapheme cluster if followed by
  # newline so we test some corner cases here.
```

lib/elixir/unicode/unicode.ex
```diff
@@ -5,313 +5,422 @@
...
-# 4. Replace GraphemeBreakTest.txt by copying original auxiliary/GraphemeBreakTest.txt
-# 5. Replace GraphemeBreakProperty.txt by copying original auxiliary/GraphemeBreakProperty.txt
-# 6. Append Extended_Pictographic from emoji/emoji-data.txt to the end of GraphemeBreakProperty.txt
-# 7. Update String.Unicode.version/0 and on String module docs (version and link)
-# 8. make unicode
-# 9. elixir lib/elixir/unicode/graphemes_test.exs
+# 4. Update String.Unicode.version/0 and on String module docs (version and link)
+# 5. make unicode
+
+data_path = Path.join(__DIR__, "UnicodeData.txt")
```
josevalim, line 11: I have moved properties.ex to this file and renamed String.Unicode to String.Casing. Other than that, the contents are strictly the same.

fertapric: The other way around, right?

josevalim: Correct, renamed Casing to Unicode.

lib/elixir/lib/string.ex
```diff
@@ -2502,18 +2563,22 @@ defmodule String do
...
+            %{} -> Map.put(bag, gc, 1)
           end
 
         string_to_bag(rest, bag, length + 1)
 
-      nil ->
+      [] ->
         {bag, length}
+
+      # Let's not count mismatched bytes in the bag
```
fertapric, line 2579: This means the results of the new version are different than the previous one. For example, with previous Elixir versions:

```elixir
iex> String.bag_distance(<<?a, 200, 200>>, "ab")
0.33333333333333337
```

With the implementation of this PR:

```elixir
iex> String.bag_distance(<<?a, 200, 200>>, "ab")
0.6666666666666667
```

Is this intended? 🙂

josevalim: It was intentional but I could not reproduce the same behavior for jaro, so I will revert. Good catch! :)

lib/elixir/lib/string.ex
```diff
@@ -1894,15 +1939,13 @@ defmodule String do
...
 
-  defp do_last({char, rest}, _) do
-    do_last(next_grapheme(rest), char)
-  end
-
-  defp do_last(nil, last_char), do: last_char
+  defp do_last([gc | rest], _), do: do_last(:unicode_util.gc(rest), gc)
+  defp do_last([], acc) when is_binary(acc), do: acc
+  defp do_last([], acc), do: :unicode.characters_to_binary([acc])
+  defp do_last({:error, <<byte, rest::bitstring>>}, _), do: do_last(rest, <<byte>>)
```
fertapric, line 1948:
```suggestion
  defp do_last({:error, <<byte, rest::bitstring>>}, _), do: do_last(:unicode_util.gc(rest), <<byte>>)
```

josevalim: Awesome find. Fixed and regression test added.

fertapric (approved):

josevalim:

Issue #11023, opened 2021-05-30
Erlang 24 changed Path.extname behaviour, returns empty string when only file extension is given


ZombieHarvester:
### Environment

* Elixir 1.12.1 (compiled with Erlang/OTP 24)
* macOS Big Sur 11.3.1

### Current behavior

`Path.extname(".jpg") == ""`

### Expected behavior

`Path.extname(".jpg") == ".jpg"`

Behaviour from Elixir 1.12.1 (compiled with **Erlang/OTP 23**)

Upgrading to Erlang 24 changed the behaviour for `Path.extname` when only file extension is given without file name. With Erlang 23 it returns the extension, with Erlang 24 it returns empty string

josevalim:
Yes, this was considered a bug fix on Erlang/OTP side, so we are considering the same here. In particular, the property that `filename == Path.basename(filename) <> Path.extname(filename)` was not held in previous OTP versions:

```iex
iex(1)> Path.basename(".gitignore")
".gitignore"
iex(2)> Path.extname(".gitignore")
".gitignore"
iex(3)>
```

But now works as expected:

```iex
iex(1)> Path.basename(".gitignore")
".gitignore"
iex(2)> Path.extname(".gitignore")
""
iex(3)>
```

Perhaps we should add a note to the docs?

ForgottenGensym:
I started a PR to add a note to the documentation, as well as adding a couple of doctests to confirm the intended behavior. However, I realized that the Erlang/OTP bug fix for this doesn't apply to paths, only files that aren't inside any directories (relatively). 

More concretely,
### Current behavior
```elixir
Path.extname(".gitignore") == ""
Path.extname("~/foo/bar/.gitignore") == ".gitignore"
```

### Expected Behavior
```elixir
Path.extname(".gitignore") == ""
Path.extname("~/foo/bar/.gitignore") == ""
```

ForgottenGensym:
Sorry about that, that was just an issue with CI. Never mind!

PR #11025, opened 2021-05-31
Clarify `Path.extname/1` doc(test)s (#11023)

ForgottenGensym:
Documentation update for #11023 

Clarify the behavior of `Path.extname/1` in docstrings and doctests, namely how it behaves when given a filename that is just a file extension (like ".gitignore"). I also mentioned when you might want to use `Path.basename/1` instead, which was brought up by @josevalim in the associated issue (#11023).

### DIFF
--- a/lib/elixir/lib/path.ex
+++ b/lib/elixir/lib/path.ex
@@ -426,20 +426,15 @@ defmodule Path do
   @doc """
   Returns the extension of the last component of `path`.
 
-  Note that when the filename consists only of a file
-  extension (like ".gitignore"), `Path.extname/1` will
-  return an empty string (""), and not ".gitignore". 
+  The behaviour of this function changed in Erlang/OTP 24 for filenames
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/path.ex
```diff
@@ -426,13 +426,34 @@ defmodule Path do
...
       ".erl"
 
       iex> Path.extname("~/foo/bar")
       ""
+      
+      iex> Path.extname(".gitignore")
+      ""
+      
+      iex> Path.extname("~/foo/bar/.gitignore")
+      ""
```
josevalim, line 456: Unfortunately we cannot add those tests as it will break CI.

josevalim:
Thanks for getting this started, very appreciated!

josevalim:

lib/elixir/lib/path.ex
ForgottenGensym: Thanks!

PR #11034, opened 2021-06-04
Add List.append/1 and List.append/2

Matsa59:
> lists:flatten/1 builds an entirely new list. It is therefore expensive, and even more expensive than the ++ operator (which copies its left argument, but not its right argument).

source: https://erlang.org/doc/efficiency_guide/listHandling.html#deep-and-flat-lists

I think it could be usefull in with `Keyword.get_values/2` for example

```elixir
iex> [opts: [key1: 1], other_key: 2, opts: [key3: 3]]
...> |> Keyword.get_values(:opts)
...> |> List.append()
[key1: 1, key3: 3]
```

I do some benchmark to check how fast it could be, here the result

### `list:flatten/1` vs `list:append/1`

```
Operating System: macOS
CPU Information: Intel(R) Core(TM) i5-1038NG7 CPU @ 2.00GHz
Number of Available Cores: 8
Available memory: 16 GB
Elixir 1.11.3
Erlang 23.2.7

warmup: 2 s
time: 5 s
memory time: 0 ns
parallel: 1
inputs: none specified
Estimated total run time: 14 s

Benchmarking list_append...
Benchmarking list_flatten...

list_append         3.97 M      251.88 ns  ±3304.54%           0 ns        1000 ns
list_flatten        2.40 M      417.37 ns  ±6865.67%           0 ns        1000 ns

Comparison:
list_append         3.97 M
list_flatten        2.40 M - 1.66x slower +165.49 ns
```

source:

```elixir
Benchee.run(%{
  list_flatten: fn -> :lists.flatten([[1, 2], [3, 4]]) == [1, 2, 3, 4] end,
  list_append: fn -> :lists.append([[1, 2], [3, 4]]) == [1, 2, 3, 4] end
})

```

### DIFF
--- a/lib/elixir/lib/list.ex
+++ b/lib/elixir/lib/list.ex
@@ -229,6 +229,46 @@ defmodule List do
   end
 
   @doc """
+  Append the given `list` of nested lists to create one list.
+
+  ## Examples
+
+      iex> List.append([[1, 2], [3, 4]])
+      [1, 2, 3, 4]
+
+      iex> List.append([[1, 2], [3], [4, 5]])
+      [1, 2, 3, 4, 5]
+
+      iex> List.append([[], [1, 2], []])
+      [1, 2]
+
+  """
+  @spec append(deep_list) :: list when deep_list: [list()]
+  def append(list) do
+    :lists.append(list)
+  end
+
+  @doc """
+  Append `tail` list to the given `list`.
+
+  ## Examples
+
+      iex> List.append([1, 2, 3], [4, 5, 6])
+      [1, 2, 3, 4, 5, 6]
+
+      iex> List.append([], [1, 2 ,3])
+      [1, 2, 3]
+
+      iex> List.append([1, 2, 3], [])
+      [1, 2, 3]
+
+  """
+  @spec append(list(), list()) :: list()
+  def append(list, tail) do
+    :lists.append(list, tail)
+  end
+
+  @doc """
   Folds (reduces) the given list from the left with
   a function. Requires an accumulator.
 
--- a/lib/elixir/test/elixir/list_test.exs
+++ b/lib/elixir/test/elixir/list_test.exs
@@ -52,6 +52,20 @@ defmodule ListTest do
     assert List.flatten([1, [], 2], [3, [], 4]) == [1, 2, 3, [], 4]
   end
 
+  test "append/1" do
+    assert List.append([1, 2], [3, 4]) == [1, 2, 3, 4]
+    assert List.append([[1, 2], []]) == [1, 2]
+    assert List.append([[], [1, 2]]) == [1, 2]
+    assert List.append([[1, 2], [3], [4], [5]]) == [1, 2, 3, 4, 5]
+  end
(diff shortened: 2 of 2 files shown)


josevalim:
I believe `Enum.concat/1` is what you want. :) If it is not fast enough, then contributing a more efficient version is definitely welcome!

Matsa59:
Hello @josevalim, I update my benchmark to 

```elixir
Benchee.run(%{
  list_flatten: fn ->
    :lists.flatten([[1, 2], [3, 4], [5, 6], [7, 8]]) == [1, 2, 3, 4, 5, 6, 7, 8]
  end,
  list_append: fn ->
    :lists.append([[1, 2], [3, 4], [5, 6], [7, 8]]) == [1, 2, 3, 4, 5, 6, 7, 8]
  end,
  enum_concat: fn ->
    Enum.concat([[1, 2], [3, 4], [5, 6], [7, 8]]) == [1, 2, 3, 4, 5, 6, 7, 8]
  end
})
```

Here the result

```
Operating System: macOS
CPU Information: Intel(R) Core(TM) i5-1038NG7 CPU @ 2.00GHz
Number of Available Cores: 8
Available memory: 16 GB
Elixir 1.11.3
Erlang 23.2.7

warmup: 2 s
time: 5 s
memory time: 0 ns
parallel: 1
inputs: none specified
Estimated total run time: 21 s

Benchmarking enum_concat...
Benchmarking list_append...
Benchmarking list_flatten...

list_append         2.50 M      399.97 ns  ±7162.22%           0 ns        1000 ns
list_flatten        1.72 M      582.41 ns  ±6004.12%           0 ns        1000 ns
enum_concat         1.20 M      832.97 ns  ±4328.63%        1000 ns        1000 ns

Comparison:
list_append         2.50 M
list_flatten        1.72 M - 1.46x slower +182.45 ns
enum_concat         1.20 M - 2.08x slower +433.00 ns
```

I didn't expect this result.
Maybe `Enum.concat` could pattern match list also ?

josevalim:
>  Maybe Enum.concat could pattern match list also ?

Yes, definitely. We can pattern match on lists and try to provide an efficient implementation.

Matsa59:
Do you want I do it in this PR or wait this one to be merged and then create another one ?

josevalim:
Oh, sorry for not being clear. This one won't be merged, as the functionality is already in `Enum.concat/1`, we just need to make it fast. :) So it is up to you if you'd rather update this one or start a new one from scratch. Probably best to start a new one and leave this one clause for documentation purposes!

Matsa59:
Oh right so I close this one and create a new one later ;) 
Thanks for your time

Matsa59:
Doesn't need to change `Enum.concat/1`, on OTP 24 it becomes really fast (3 times faster than on OTP 23).
`lists:append/1` still faster but only for concats list of nested list (30% to 100% depending of the list size). 

Moreover, `Enum.concat/1`generate the list and add it in the same way to the result list, that make it faster than create it and append it.

Nice to know ;)

PR #11035, opened 2021-06-04
Small grammar fix for Config Providers section

J3RN:
The following phrase, which appears in the Config Providers documentation, is grammatically incorrect:
> The runtime configuration outlined in the previous section, which is handled by the Config.Reader provider.

Given the context, I think the phrase should be:
> The runtime configuration outlined in the previous section is handled by the `Config.Reader` provider.


whatyouhide (approved):

josevalim:

PR #11038, opened 2021-06-05
Add `identifier_location` and `last` metadata

doorgan:
This adds `identifier_locaiton` metadata to qualified identifiers and `last` metadata to aliases, as discussed in [this proposal](https://groups.google.com/u/1/g/elixir-lang-core/c/6APSoyKdKIc)

### DIFF
--- a/lib/elixir/src/elixir_parser.yrl
+++ b/lib/elixir/src/elixir_parser.yrl
@@ -830,7 +830,7 @@ build_alias({'alias', Location, Alias}) ->
   Meta = meta_from_location(Location),
   MetaWithExtra =
     case ?token_metadata() of
-      true -> lists:keystore(last, 1, Meta, {last, meta_from_location(Location)});
+      true -> [{last, meta_from_location(Location)} | Meta];
       false -> Meta
     end,
   {'__aliases__', MetaWithExtra, [Alias]}.
@@ -848,7 +848,7 @@ build_dot_alias(Dot, Expr, {'alias', SegmentLocation, Right}) ->
   Meta = meta_from_token(Dot),
   MetaWithExtra =
     case ?token_metadata() of
-      true -> lists:keystore(last, 1, Meta, {last, meta_from_location(SegmentLocation)});
+      true -> [{last, meta_from_location(SegmentLocation)} | Meta];
       false -> Meta
     end,
   {'__aliases__', MetaWithExtra, [Expr, Right]}.
--- a/lib/elixir/src/elixir_parser.yrl
+++ b/lib/elixir/src/elixir_parser.yrl
@@ -827,14 +827,31 @@ end_of_expression(Location) ->
 %% Dots
 
 build_alias({'alias', Location, Alias}) ->
-  {'__aliases__', meta_from_location(Location), [Alias]}.
+  Meta = meta_from_location(Location),
+  MetaWithExtra =
+    case ?token_metadata() of
+      true -> lists:keystore(last, 1, Meta, {last, meta_from_location(Location)});
+      false -> Meta
+    end,
+  {'__aliases__', MetaWithExtra, [Alias]}.
 
-build_dot_alias(_Dot, {'__aliases__', Meta, Left}, {'alias', _, Right}) ->
-  {'__aliases__', Meta, Left ++ [Right]};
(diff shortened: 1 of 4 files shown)


lib/elixir/src/elixir_parser.yrl
```diff
@@ -827,21 +827,46 @@ end_of_expression(Location) ->
...
   Meta = meta_from_token(Dot),
   {{'.', Meta, [Left, '{}']}, Extra ++ Meta, Right}.
 
-build_dot(Dot, Left, Right) ->
-  {'.', meta_from_token(Dot), [Left, extract_identifier(Right)]}.
+build_dot(Dot, Left, {_, IdentifierLocation, _} = Right) ->
+  Meta =
+    case ?token_metadata() of
+      true ->
+        IdentifierMeta = {identifier_location, meta_from_location(IdentifierLocation)},
```
josevalim, line 864: I don't think we need an additional key for this. I was thinking instead we could put it on the root directly:

    {{:., DotLocation, ...}, IdentifierLocation, _}

Maybe we should change this function to return a four/five element tuple with all metadata and then we change how build_identifier matches on build_dot?

lib/elixir/src/elixir_parser.yrl
```diff
@@ -827,21 +827,46 @@ end_of_expression(Location) ->
...
+  {'__aliases__', MetaWithExtra, Left ++ [Right]};
 build_dot_alias(_Dot, Atom, Right) when is_atom(Atom) ->
   error_bad_atom(Right);
-build_dot_alias(Dot, Expr, {'alias', _, Right}) ->
-  {'__aliases__', meta_from_token(Dot), [Expr, Right]}.
+build_dot_alias(Dot, Expr, {'alias', SegmentLocation, Right}) ->
+  Meta = meta_from_token(Dot),
+  MetaWithExtra =
+    case ?token_metadata() of
+      true -> lists:keystore(last, 1, Meta, {last, meta_from_location(SegmentLocation)});
```
josevalim, line 851: We can use this stule, as we are sure there are no entries named last anyway:

```suggestion
      true -> [{last, meta_from_location(SegmentLocation)} | Meta];
```

doorgan: Would it be better to use `Meta ++ [{last, meta_from_location(SegmentLocation)}]`? In the previous clauses I'm using `lists:keystore`, which adds the tuple at the end, using `++` would make the ordering consistent. We could also use `lists:keydelete` and then cons the new tuple in the previous one. WDYT?

josevalim: We don’t guarantee order in metadata. So we don’t need to worry about it.

doorgan:
It seems that if the root of the qualified identifier is the metadata for the identifier instead of the metadata for the dot, then the compiler tracers start reporting wrong column numbers. This test:
```
test "traces remote" do
    compile_string("""
    require Integer
    true = Integer.is_odd(1)
    {1, ""} = Integer.parse("1")
    """)

    assert_receive {{:remote_macro, meta, Integer, :is_odd, 1}, _}
    assert meta[:line] == 2
    assert meta[:column] == 15

    assert_receive {{:remote_function, meta, Integer, :parse, 1}, _}
    assert meta[:line] == 3
    assert meta[:column] == 18
  end
```
right now is failing because it's reporting column `16` in the first case, which is what I would have expected(since it now points at the identifier, not at the dot). But if I fix the test and also change the `18` to `19` in the next assertion, it fails. It seems that for remote macros it reports the identifier location but for remote functions it reports the dot location.

I'm failing to find where this is handled

josevalim:
You need to lookup for calls like this: https://github.com/elixir-lang/elixir/blob/50579b41db16100eb2ded21530830a32a09c32ec/lib/elixir/src/elixir_expand.erl#L804

both remote function and remote macro and make sure you are passing the DotMeta. Previously I worked because there was no distinction between them.

doorgan:
That seemed to be it :)

josevalim:
I was thinking we should do the opposite though: keep the remote_function as is and change the remote_macro. On the other hand... if it is an import, there is no dot, so we always start from the identifier, so I think the current behavior is more consistent?

josevalim:

doorgan:
> so I think the current behavior is more consistent?
It was just merged, but I was about to say the same, I would personally consider it more consistent than reporting sometimes the dot and sometimes the identifier

PR #11039, opened 2021-06-05
add @doc since: "1.12.0"

TORIFUKUKaiou:
Would you please do review?
If I mistake start/end point branch, please comment to me.

I think that Code.ensure_loaded!/1 is added from 1.12.0.
@doc since: "1.12.0" is unnecessary here, so I also think that there may be nothing now.
If I shall waste your time, I apologize.


josevalim:

TORIFUKUKaiou:
Thanks a lot for your merge.

PR #11042, opened 2021-06-06
Warn if ?\lettter is used but there is no such escape sequence

### LINKED ISSUE #11028: Warn if ?\lettter is used but there is no such escape sequence
### Environment

* Elixir & Erlang/OTP versions (elixir --version): v1.12 / OTP 24
* Operating system: Mac

### Current behavior

```
?\w == 119
```

### Expected behavior

The code above should print a warning: "unknown escape sequence ?\w, use ?w instead"

QuinnWilton:
This resolves #11028

I don't write a ton of Erlang, so it's entirely possible that this isn't very idiomatic code, and I'm happy to go back and rework anything as requested :)

### DIFF
--- a/lib/elixir/src/elixir_tokenizer.erl
+++ b/lib/elixir/src/elixir_tokenizer.erl
@@ -237,8 +237,16 @@ tokenize([$~, S, H | _] = Original, Line, Column, _Scope, Tokens) when ?is_upcas
 
 tokenize([$?, $\\, H | T], Line, Column, Scope, Tokens) ->
   Char = elixir_interpolation:unescape_map(H),
+  NewScope = if
+    H =:= Char, H =/= $\\ ->
+      Msg = io_lib:format("unknown escape sequence ?\\~tc, use ?~tc instead",
+                          [H, H]),
+      prepend_warning({Line, Scope#elixir_tokenizer.file, Msg}, Scope);
+    true ->
+      Scope
+  end,
   Token = {char, {Line, Column, [$?, $\\, H]}, Char},
-  tokenize(T, Line, Column + 3, Scope, [Token | Tokens]);
+  tokenize(T, Line, Column + 3, NewScope, [Token | Tokens]);
 
 tokenize([$?, Char | T], Line, Column, Scope, Tokens) ->
   NewScope = case handle_char(Char) of
--- a/lib/elixir/test/elixir/kernel/warning_test.exs
+++ b/lib/elixir/test/elixir/kernel/warning_test.exs
@@ -1249,13 +1249,11 @@ defmodule Kernel.WarningTest do
     purge(Sample)
   end
 
-  test "no warning on necessary code point escape" do
-    assert capture_err(fn ->
-             Code.eval_string("?\\n")
-           end) == ""
-  end
-
   test "warning on unnecessary code point escape" do
+    assert capture_err(fn ->
+             Code.eval_string("?\\n + ?\\\\")
+           end) == ""
+
     assert capture_err(fn ->
              Code.eval_string("?\\w")
            end) =~ "unknown escape sequence ?\\w, use ?w instead"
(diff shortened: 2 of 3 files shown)


lib/elixir/src/elixir_tokenizer.erl
```diff
@@ -237,8 +237,16 @@ tokenize([$~, S, H | _] = Original, Line, Column, _Scope, Tokens) when ?is_upcas
 
 tokenize([$?, $\\, H | T], Line, Column, Scope, Tokens) ->
   Char = elixir_interpolation:unescape_map(H),
+  NewScope = if
+    H =:= Char andalso (?is_upcase(H) orelse ?is_downcase(H)) ->
```
josevalim, line 241: Could this be rewritten to `(H =:= Char) and (H =/= $\\)`? This is the only case where `\\` is needed to return itself, right?

lib/elixir/test/elixir/kernel/warning_test.exs
```diff
@@ -1249,6 +1249,18 @@ defmodule Kernel.WarningTest do
     purge(Sample)
   end
 
+  test "no warning on necessary code point escape" do
+    assert capture_err(fn ->
+             Code.eval_string("?\\n")
+           end) == ""
+  end
+
+  test "warning on unnecessary code point escape" do
```
josevalim, line 1258:
```suggestion
  test "warning on unnecessary code point escape" do
    assert capture_err(fn ->
             Code.eval_string("?\\n")
           end) == ""

```

lib/elixir/src/elixir_tokenizer.erl
QuinnWilton: I actually originally had this check, but there was a few warnings showing up in `ExUnit`, because it does `?\"` a couple of times. While that escape isn't necessary there, it is a common escape in other contexts, so I'm not sure whether you'd prefer printing warnings in that case or not.

Do you have any thoughts about that?

josevalim: I would say let’s try to warn for all and if the push back is big, we can relax it a bit. WDYT?

QuinnWilton: That sounds good to me! I'll fix those warnings as part of this PR.

lib/elixir/src/elixir_tokenizer.erl
```diff
@@ -237,8 +237,16 @@ tokenize([$~, S, H | _] = Original, Line, Column, _Scope, Tokens) when ?is_upcas
 
 tokenize([$?, $\\, H | T], Line, Column, Scope, Tokens) ->
   Char = elixir_interpolation:unescape_map(H),
+  NewScope = if
+    H =:= Char andalso H =/= $\\ ->
```
QuinnWilton, line 241: In your original comment you suggested: `(H =:= Char) and (H =/= $\\)`

1) Is there a reason to prefer `and` to `andalso` here?
2) Would you prefer the parentheses around each clause?

QuinnWilton:
I pushed up the changes! I left one clarifying comment, too.

lib/elixir/src/elixir_tokenizer.erl
josevalim: Precedence in Erlang is not clear to me, so usually err on the side of caution. But in this case, we are inside an if, which allows guards, so I would do:

```suggestion
    H =:= Char, H =/= $\\ ->
```

QuinnWilton: I didn't even know that was an option! Pushed.

josevalim:

QuinnWilton:
This probably isn't the place to ask, but are there any other issues (either tracked or not), that you think would be worth me working on? I'm itching to write some more code 😅

josevalim:
Unfortunately there isn’t anything besides what is in the issues tracker. If something comes up I will let you know!

PR #11048, opened 2021-06-07
Add `--fail-above` option to `mix xref`

marcandre:
As proposed [on elixir-core](https://groups.google.com/g/elixir-lang-core/c/Hmg5y8S3v4E/m/FVK50RWBAQAJ)

Other PR forthcoming...

PR #11049, opened 2021-06-07
Support multiple sinks and sources in `mix xref graph`

marcandre:
As proposed [on elixir-core](https://groups.google.com/g/elixir-lang-core/c/Hmg5y8S3v4E/m/FVK50RWBAQAJ)

Looks like it does not conflict with #11048 🤞

### DIFF
--- a/lib/mix/lib/mix/tasks/xref.ex
+++ b/lib/mix/lib/mix/tasks/xref.ex
@@ -208,8 +208,8 @@ defmodule Mix.Tasks.Xref do
     include_siblings: :boolean,
     label: :string,
     only_nodes: :boolean,
-    sink: :string,
-    source: :string,
+    sink: :keep,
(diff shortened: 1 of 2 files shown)


lib/mix/lib/mix/tasks/xref.ex
```diff
@@ -478,35 +478,43 @@ defmodule Mix.Tasks.Xref do
...
+        )
 
-    if sink && is_nil(file_references[sink]) do
-      Mix.raise("Sink could not be found: #{sink}")
+      _ ->
+        :ok
     end
 
+    # remap [] to nil
+    with [] <- files, do: nil
```
josevalim, line 495:
```suggestion
    if files == [], do: nil, else: files
```

josevalim:

Issue #11050, opened 2021-06-07
Unmatched parentheses using `String.match?/2` with `~r()` trying to match parentheses


mrcasals:
### Precheck

I looked for the in the issue tracker, nothing appeared.

### Environment

* Elixir & Erlang/OTP versions (elixir --version): Erlang/OTP 22, Elixir 1.11.4
* Operating system: MacOS 11.0.1

### Current behavior

When trying to match parentheses using the `~r()` sigil with parentheses as enclosing symbols, it gives an `unmatched parentheses` error. The error does not appear when using `~r//` with slashes or `~r[]`:

```elixir
iex(1)> str = "Activity (debate) 12"
"Activity (debate) 12"
iex(2)> String.match?(str, ~r(\(debate\)))
** (Regex.CompileError) unmatched parentheses at position 8
    (elixir 1.11.4) lib/regex.ex:216: Regex.compile!/2
    (elixir 1.11.4) expanding macro: Kernel.sigil_r/2
    iex:2: (file)
iex(2)> String.match?(str, ~r/\(debate\)/)
true
iex(3)> String.match?(str, ~r[\(debate\)])
true
```

### Expected behavior

Regexp should be able to compile just fine using either `~r//` or `~r()`

josevalim:
This is expected. The `\)` in your case is escaping the closing of the sigil, not the parenthesis itself. In order to understand why, just change the sigil:

    iex> ~s(\(example\))
    "(example)"

In Ruby, which may be from where your expectation comes from, it works because %r comes from the parser. So the parser is effectively parsing a regex and knows the regex rules. Elixir, on the other hand, has a generic approach to sigil and the ~r sigil is completely user defined. The good news is that all sigils follow the same rule but it leaves less spaces for specialization. Therefore using // and "" for regexes is typically preferred.

mrcasals:
@josevalim that's correct, I'm coming from Ruby :) Thanks for taking the time to explain this!

> Therefore using // and "" for regexes is typically preferred.

Should docs examples on the `sigil_r` be updated to use `~r//`, then?

https://github.com/elixir-lang/elixir/blob/2a25a72d08e4a7cb3198e0000eb8c76b334f378b/lib/elixir/lib/kernel.ex#L5590-L5607

josevalim:
Yes! Would you like to send a PR? :)

mrcasals:
@josevalim on it!

mrcasals:
[...]

PR #11051, opened 2021-06-07
Improve docs for `sigil_r`

mrcasals:
A per #11050, this PR improves the docs on `sigil_r` to notice that using `~r()` and trying to match escaped parentheses will raise an exception, so any other enclosing character is recommended, specially `~r//` and `~r""`.

### DIFF
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -5603,7 +5603,7 @@ defmodule Kernel do
       iex> Regex.match?(~r/a#{:b}c/, "abc")
       true
 
-  While the `r` sigil allows parens and brackets to be used as delimiters,
+  While the `~r` sigil allows parens and brackets to be used as delimiters,
   it is preferred to use `"` or `/` to avoid escaping conflicts with reserved
   regex characters:
 
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -5603,16 +5603,15 @@ defmodule Kernel do
       iex> Regex.match?(~r/a#{:b}c/, "abc")
       true
 
-  Note that using `~r()` might raise an exception if you try to escape
-  parentheses, so typically `~r//` or `~r""` are preferred. `~r[]` also works
-  fine:
+  While the `r` sigil allows parens and brackets to be used as delimiters,
+  it is preferred to use `"` or `/` to avoid escaping conflicts with reserved
+  regex characters:
 
       iex> Regex.match?(~r(\(foo\)), "(foo)")
       ** (Regex.CompileError) unmatched parentheses at position 5
       iex> Regex.match?(~r/\(foo\)/, "(foo)")
       true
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/kernel.ex
```diff
@@ -5597,12 +5597,22 @@ defmodule Kernel do
...
+  Note that using `~r()` might raise an exception if you try to escape
+  parentheses, so typically `~r//` or `~r""` are preferred. `~r[]` also works
+  fine:
+
+      iex> Regex.match?(~r(\(foo\)), "(foo)")
+      ** (Regex.CompileError) unmatched parentheses at position 5
+      iex> Regex.match?(~r/\(foo\)/, "(foo)")
+      true
+      iex> Regex.match?(~r[\(foo\)], "(foo)")
+      true
```
josevalim, line 5615:
```suggestion
  While the `r` sigil allows parens and brackets to be used as delimiters,
  it is preferred to use `"` or `/` to avoid escaping conflicts with reserved
  regex characters:

      iex> Regex.match?(~r(\(foo\)), "(foo)")
      ** (Regex.CompileError) unmatched parentheses at position 5
      iex> Regex.match?(~r/\(foo\)/, "(foo)")
      true

```

lib/elixir/lib/kernel.ex
```diff
@@ -5597,12 +5597,21 @@ defmodule Kernel do
...
   ## Examples
 
-      iex> Regex.match?(~r(foo), "foo")
+      iex> Regex.match?(~r/foo/, "foo")
       true
 
       iex> Regex.match?(~r/a#{:b}c/, "abc")
       true
 
+  While the `r` sigil allows parens and brackets to be used as delimiters,
```
josevalim, line 5606:
```suggestion
  While the `~r` sigil allows parens and brackets to be used as delimiters,
```

lib/elixir/lib/kernel.ex
```diff
@@ -5597,12 +5597,21 @@ defmodule Kernel do
...
 
+  While the `~r` sigil allows parens and brackets to be used as delimiters,
+  it is preferred to use `"` or `/` to avoid escaping conflicts with reserved
+  regex characters:
+
+      iex> Regex.match?(~r(\(foo\)), "(foo)")
+      ** (Regex.CompileError) unmatched parentheses at position 5
+      iex> Regex.match?(~r/\(foo\)/, "(foo)")
+      true
+
```
josevalim, line 5614:
```suggestion
  regex characters.
```

josevalim:
Unfortunately testing this with doctest is hard, because the doctest does not even compile, not allowing the test suite to run, so I removed the examples for now. :(

mrcasals:
> Unfortunately testing this with doctest is hard, because the doctest does not even compile, not allowing the test suite to run, so I removed the examples for now. :(

Ah, I thought this could happen but hadn't had time to check it out. Thanks for the improvements on the docs!

josevalim:

PR #11060, opened 2021-06-11
Add doc example for :"Elixir.AnAtom"

anildigital:
Add doc example for atoms like module names.

### DIFF
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -566,6 +566,9 @@ defmodule Kernel do
       iex> is_atom("true")
       false
 
+      iex> is_atom(AnAtom)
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/kernel.ex
```diff
@@ -566,6 +566,9 @@ defmodule Kernel do
       iex> is_atom("true")
       false
 
+      iex> is_atom(AnAtom)
+      true
+
   """
```
v0idpwn, line 572:
```suggestion
      iex> is_atom(AnAtom)
      true
      
      iex> is_atom(AnAtom)
      true
  """
```

Keeping true and false results separate

josevalim:

PR #11061, opened 2021-06-11
Preserve user choice on parenthesis before do block

doorgan:
The normalizer adds a `:closing` line to calls even if there is a `do` block, forcing the formatter to add parenthesis. So for code like this one:
```elixir
module DemoMod do
  def foo, do: :bar
end
```
`Code.quoted_to_algebra` will produce this code:
```elixir
module(DemoMod) do
  def foo, do: :bar
end
```
While `Code.format_string!` will produce the output without parenthesis.

I added a check to not force the `:closing` metadata to calls if the call has `:do` metadata.

### DIFF
--- a/lib/elixir/lib/code/normalizer.ex
+++ b/lib/elixir/lib/code/normalizer.ex
@@ -284,7 +284,7 @@ defmodule Code.Normalizer do
       end
 
     meta =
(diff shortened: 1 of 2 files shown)


josevalim:
:green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart: love the work you are doing!

PR #11064, opened 2021-06-16
fix: concatenation errors

WLSF:
Currently the concat operator has variant responses for different scenarios, for example:

```elixir
# Using literals
iex> "" <> 1
** (ArgumentError) expected binary argument in <> operator but got: 1

# Using non-literals
iex> a = ""
iex> b = 3
iex> a <> B
** (ArgumentError) argument error while evaluating iex at line 2

# Using variables that are going to be solved at runtime
defmodule T do
  def f(a, b), do: a <> b  
end

iex> a = ""
iex> b = 2
iex> T.f(a, b)
# badarg error
* 1st argument: not a bitstring     This typically happens when calling Kernel.byte_size/1 with an invalid argument     or when performing binary concatenation with &lt;&gt; and one of the arguments is not a binary\

iex> T.f(b, a)
...
# just a generic error that doesn't fit any of the cases above
```

Because of that, I wanted to leave these changes here as a suggestion to improve the development experience.

The idea is to "double-check" the left and right side hands at the end, in order to ensure that both sides are binaries.

And since it's going add a couple of checks and validations into the generated AST, we wouldn't be able to compare concat operations with merging binaries operations, I'm not sure if this is a problem at all. (?)

Also, this is only going to affect operations that aren't in guard/match context; otherwise it's the same as it was before.

What do you think?

### DIFF
--- a/lib/elixir/lib/exception.ex
+++ b/lib/elixir/lib/exception.ex
@@ -1550,16 +1550,6 @@ defmodule ErlangError do
     {nil, nil, nil}
   end
 
-  defp error_info(:badarg, [{:erlang, :byte_size, _, _} | _]) do
-    {:ok,
-     """
-       * 1st argument: not a bitstring
-
-     This typically happens when calling Kernel.byte_size/1 with an invalid argument
-     or when performing binary concatenation with <> and one of the arguments is not a binary\
-     """}
-  end
-
   defp error_info(erl_exception, stacktrace) do
     with [{module, _, args_or_arity, opts} | _] <- stacktrace,
          %{} = error_info <- opts[:error_info] do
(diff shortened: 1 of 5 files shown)


josevalim:
Adding the checks at this point slows down some operations considerably, which is why we have decided to do a post check. My hope is that, with the new Erlang/OTP 24 error_info information, perhaps we are able to improve this directly in OTP, but unfortunately we cannot take this approach.

In any case, thanks for the PR and exploring it!

josevalim:
Discussion opened here: https://github.com/erlang/otp/issues/4971

WLSF:
Ok! Thanks @josevalim.

I had also updated a doc that was wrongly saying about `CompileError` when running LHS pattern matches on `<>/2`,
shall I open another PR for that?

josevalim:
@WLSF , yes please! I was going to suggest to reopen this one but it is probably best to keep it as is for historical purposes.

PR #11067, opened 2021-06-16
fix: docs for <>/2 operator

WLSF:
Add exception details for docs.

When using `<>/2` for pattern matching, the match should always happen on the RHS (right hand side) var, if it doesn't, then you'll get an ArgumentError with:
```
** (ArgumentError) the left argument of <> operator inside a match should always be a literal binary because its size can't be verified. Got: x
```


josevalim:

PR #11070, opened 2021-06-19
Update sigil_r use as per docs

anildigital:
As per docs https://github.com/elixir-lang/elixir/pull/11051, using `~r//` is recommended over using `~r()`. This PR updates the sigil_r delimiter used at multiple places.


josevalim:

PR #11071, opened 2021-06-19
Add parser_options to EEx.Compiler

dealloc:
As per https://groups.google.com/g/elixir-lang-core/c/Z-C2PFRwyw4
This PR adds the `:parser_options` option to `EEx.Compiler` which allows passing in custom options without having to modify the global options through `Code.put_compiler_option/2`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 #### EEx
 
+  * [Code] Added `:parser_options` to EEx functions
+
 #### Elixir
 
   * [Code] Add `Code.string_to_quoted_with_comments/2` and `Code.quoted_to_algebra/2`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 #### EEx
 
-  * [Code] Added `:parser_options` to EEx functions
+  * [Code] Add `:parser_options` to EEx functions
 
 #### Elixir
 
--- a/lib/eex/lib/eex.ex
+++ b/lib/eex/lib/eex.ex
@@ -48,6 +48,9 @@ defmodule EEx do
     * `:trim` - if `true`, trims whitespace left and right of quotation as
       long as at least one newline is present. All subsequent newlines and
       spaces are removed but one newline is retained. Defaults to `false`.
+    * `:parser_options` - allow customizing the parsed code that is generated.
+      See `Code.string_to_quoted/2` for available options. Note that the options
+      `:file`, `:line` and `:column` are ignored if passed in.
 
   ## Engine
 
--- a/lib/eex/lib/eex.ex
+++ b/lib/eex/lib/eex.ex
@@ -51,6 +51,7 @@ defmodule EEx do
     * `:parser_options` - allow customizing the parsed code that is generated.
       See `Code.string_to_quoted/2` for available options. Note that the options
       `:file`, `:line` and `:column` are ignored if passed in.
+      Defaults to `Code.get_compiler_option(:parser_options)` (which defaults to `[]` if not set).
 
   ## Engine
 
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 
 #### EEx
 
+  * [Code] Added `:parser_options` to EEx functions
```
josevalim, line 9:
```suggestion
  * [Code] Add `:parser_options` to EEx functions
```

lib/eex/lib/eex/compiler.ex
```diff
@@ -27,7 +28,7 @@ defmodule EEx.Compiler do
           quoted: [],
           start_line: nil,
           start_column: nil,
-          parser_options: Code.get_compiler_option(:parser_options)
+          parser_options: Code.get_compiler_option(:parser_options) ++ parser_options
```
josevalim, line 31:
```suggestion
          parser_options: Keyword.merge(Code.get_compiler_option(:parser_options), parser_options)
```

lib/eex/test/eex_test.exs
```diff
@@ -718,6 +718,16 @@ defmodule EExTest do
     end
   end
 
+  describe "parser options" do
+    test "are used" do
```
josevalim, line 722:
```suggestion
    test "customizes parsed code" do
```

lib/eex/lib/eex/compiler.ex
dealloc: I used `++` because it's used where the options are passed into `Code.string_to_quoted` -> https://github.com/elixir-lang/elixir/blob/24f8db1a40336e3cc0c517e048b77e75ff8123ad/lib/eex/lib/eex/compiler.ex#L59
Out of curiosity, is there a reason to prefer `Keyword.merge`?

josevalim: You can use `++` but the options that come first win, which is likely not what you want here. There is another option, which is to do:

    parser_options = opts[:parser_options] || Code.get_compiler_options(:parser_options)

Perhaps this is even preferred as the intent is clear?

josevalim:
We also need to add docs. :)

lib/eex/lib/eex/compiler.ex
dealloc: You're right, that's not what we want here.
I'll use `parser_options = opts[:parser_options] || Code.get_compiler_options(:parser_options)` as it's definitely more clear about what we intend to do.

dealloc: @josevalim since doing it this way will entirely override anything set in `Code.get_compiler_options(:parser_options)` do we want an additional test that checks `get_compiler_options` options are no longer used?

Also, isn't this technically breaking, assuming someone relies on `get_compiler_options` for their EEx templates?

josevalim: I can see someone being surprised that their global configuration is not taking effect but I can also see someone surprised their options are mixed with global ones. So I think the full override behavior is clearer and folks can merge upstream if they desire.

lib/eex/lib/eex.ex
```diff
@@ -48,6 +48,9 @@ defmodule EEx do
     * `:trim` - if `true`, trims whitespace left and right of quotation as
       long as at least one newline is present. All subsequent newlines and
       spaces are removed but one newline is retained. Defaults to `false`.
+    * `:parser_options` - allow customizing the parsed code that is generated.
+      See `Code.string_to_quoted/2` for available options. Note that the options
+      `:file`, `:line` and `:column` are ignored if passed in.
```
josevalim, line 53: We need to document what is the default value!

josevalim:

PR #11073, opened 2021-06-21
Optimize Enum.concat/1 for lists of lists

greg-rychlewski:
This is building off the closed pull request here: https://github.com/elixir-lang/elixir/pull/11034, where it was observed that `:lists.append/1` out performs `Enum.concat/1` for lists of lists. 

I've made a change so that `:lists.append/1` is used when a list of lists is passed into `Enum.concat/1`, otherwise it will default to the previous implementation. 

I've run some benchmarks of my own that confirm the insights from the previous pull request: https://github.com/greg-rychlewski/concat-test/blob/master/benchmark.exs.

Here is the output for the benchmarks:

```
##### With input large #####
Name               ips        average  deviation         median         99th %
proposal        709.45        1.41 ms    ±40.08%        1.30 ms        4.12 ms
master          376.83        2.65 ms    ±27.31%        2.49 ms        5.39 ms

Comparison: 
proposal        709.45
master          376.83 - 1.88x slower +1.24 ms

##### With input medium #####
Name               ips        average  deviation         median         99th %
proposal       13.91 K       71.90 μs    ±38.63%       64.99 μs      161.99 μs
master          5.64 K      177.17 μs    ±21.97%      159.99 μs      290.99 μs

Comparison: 
proposal       13.91 K
master          5.64 K - 2.46x slower +105.27 μs

##### With input small #####
Name               ips        average  deviation         median         99th %
proposal        1.17 M        0.86 μs  ±3090.98%        0.99 μs        0.99 μs
master          0.46 M        2.15 μs  ±1285.91%        1.99 μs        3.99 μs

Comparison: 
proposal        1.17 M
master          0.46 M - 2.51x slower +1.29 μs
```

### DIFF
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -3780,7 +3780,7 @@ defmodule Enum do
 
   defp concat_enum(enum) do
     fun = &[&1 | &2]
-    enum |> Enum.reduce([], &Enum.reduce(&1, &2, fun)) |> :lists.reverse()
+    enum |> reduce([], &reduce(&1, &2, fun)) |> :lists.reverse()
   end
 
   # dedup
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -599,8 +599,11 @@ defmodule Enum do
   """
   @spec concat(t) :: t
   def concat(enumerables) do
-    fun = &[&1 | &2]
-    enumerables |> reduce([], &reduce(&1, &2, fun)) |> :lists.reverse()
+    if is_list(enumerables) and :lists.all(&is_list/1, enumerables) do
+      concat_list(enumerables)
+    else
+      concat_enum(enumerables)
+    end
   end
 
   @doc """
@@ -625,7 +628,7 @@ defmodule Enum do
   end
 
   def concat(left, right) do
-    concat([left, right])
+    concat_enum([left, right])
   end
 
   @doc """
@@ -3767,6 +3770,17 @@ defmodule Enum do
 
   defp any_list([], _) do
     false
+  end
+
+  ## concat
+
+  defp concat_list(enumerables) do
+    :lists.append(enumerables)
+  end
+
+  defp concat_enum(enumerables) do
+    fun = &[&1 | &2]
+    enumerables |> reduce([], &reduce(&1, &2, fun)) |> :lists.reverse()
   end
 
   # dedup
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -598,12 +598,12 @@ defmodule Enum do
 
   """
   @spec concat(t) :: t
-  def concat(enumerables) do
-    if is_list(enumerables) and :lists.all(&is_list/1, enumerables) do
-      concat_list(enumerables)
-    else
-      concat_enum(enumerables)
-    end
+  def concat(list) when is_list(list) do
+    concat_list(list)
+  end
+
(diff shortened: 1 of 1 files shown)


greg-rychlewski:
It seems like the compiler doesn't like `&is_list/1` here:

```
if is_list(enumerables) and :lists.all(&is_list/1, enumerables) do
```

If I change it to this the warning disappears

```
if is_list(enumerables) and :lists.all(fn x -> is_list(x) end, enumerables) do
```

v0idpwn:
Hey! I think making ConcatTest a protocol would give us more precise benchmark results. While your approach is of course still faster, I think it would give us better history for posteriority :)

greg-rychlewski:
> Hey! I think making ConcatTest a protocol would give us more precise benchmark results. While your approach is of course still faster, I think it would give us better history for posteriority :)

Thanks! I switched it to a protocol and re-ran the tests. Here are the new results:

```
Operating System: macOS
CPU Information: Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
Number of Available Cores: 8
Available memory: 16 GB
Elixir 1.9.4
Erlang 22.0

warmup: 2 s
time: 5 s
memory time: 0 ns
parallel: 1
inputs: large, medium, small
Estimated total run time: 42 s

Benchmarking master with input large...
Benchmarking master with input medium...
Benchmarking master with input small...
Benchmarking proposal with input large...
Benchmarking proposal with input medium...
Benchmarking proposal with input small...

##### With input large #####
Name               ips        average  deviation         median         99th %
proposal        817.79        1.22 ms    ±18.28%        1.28 ms        1.73 ms
master          349.81        2.86 ms    ±42.11%        2.52 ms        7.88 ms

Comparison: 
proposal        817.79
master          349.81 - 2.34x slower +1.64 ms

##### With input medium #####
Name               ips        average  deviation         median         99th %
proposal       13.34 K       74.96 μs    ±38.21%          70 μs         173 μs
master          5.97 K      167.59 μs    ±24.77%         141 μs      311.67 μs

Comparison: 
proposal       13.34 K
master          5.97 K - 2.24x slower +92.63 μs

##### With input small #####
Name               ips        average  deviation         median         99th %
proposal        1.14 M        0.87 μs  ±3032.45%           1 μs           1 μs
master          0.51 M        1.95 μs  ±1278.05%           2 μs           3 μs

Comparison: 
proposal        1.14 M
master          0.51 M - 2.23x slower +1.07 μs
```

josevalim:
Thanks @greg-rychlewski!

I pushed a small change that inlines :lists.append implementation, so it doesn't require us to check the whole list upfront. It also has the benefit of using the fast approach until we find an element that is not a list. Your implementation would also return an improper list if the last list in concat was improper, so the change also address it.

lib/elixir/lib/enum.ex
```diff
@@ -3769,6 +3772,17 @@ defmodule Enum do
...
 
+  ## concat
+
+  defp concat_list([h | t]) when is_list(h), do: h ++ concat_list(t)
+  defp concat_list([h | t]), do: concat_enum([h | t])
+  defp concat_list([]), do: []
+
+  defp concat_enum(enum) do
+    fun = &[&1 | &2]
+    enum |> Enum.reduce([], &Enum.reduce(&1, &2, fun)) |> :lists.reverse()
```
josevalim, line 3783:
```suggestion
    enum |> reduce([], &reduce(&1, &2, fun)) |> :lists.reverse()
```

josevalim:

PR #11076, opened 2021-06-21
Update reference to IRC channel in CoC

Nicd:
There isn't currently an official Webchat for Libera, so I wanted to change the link to an IRCS protocol link. But it seems GitHub won't render it as a link even if I use an inline link style (see: https://github.com/github/markup/issues/426). So in the end I removed the channel link.


josevalim:

PR #11078, opened 2021-06-22
Don't discard comments column information

doorgan:
The comment column is required to calculate the range a node and it's surrounding comments span in the source code by inspecting the ast and the list of comments.

I'm writing some functions that allow users to perform ast based "find and replace", and the lack of column numbers in comments make it hard to map a node and its comments to a range in the source code so we can replace it


josevalim:

PR #11080, opened 2021-06-22
Rely on modification time and hash to determine modified sources

marcandre:
As discussed [on the mailing list](https://groups.google.com/g/elixir-lang-core/c/8-30JVn_8M0/m/hUWxonjwAQAJ), I think this adds the right check by hashing the content, but I'm lacking tests.

Does it look like I'm on the right track?

### DIFF
--- a/lib/mix/lib/mix/compilers/elixir.ex
+++ b/lib/mix/lib/mix/compilers/elixir.ex
@@ -1,7 +1,7 @@
 defmodule Mix.Compilers.Elixir do
   @moduledoc false
 
-  @manifest_vsn 8
+  @manifest_vsn 9
 
   import Record
 
@@ -10,6 +10,7 @@ defmodule Mix.Compilers.Elixir do
   defrecord :source,
     source: nil,
     size: 0,
+    digest: nil,
     compile_references: [],
     export_references: [],
     runtime_references: [],
@@ -201,12 +202,13 @@ defmodule Mix.Compilers.Elixir do
     # Sources that have changed on disk or
     # any modules associated with them need to be recompiled
     changed =
-      for source(source: source, external: external, size: size, modules: modules) <-
+      for source(source: source, external: external, size: size, digest: digest, modules: modules) <-
             all_sources,
           {last_mtime, last_size} = Map.fetch!(sources_stats, source),
           times = Enum.map(external, &(sources_stats |> Map.fetch!(&1) |> elem(0))),
-          size != last_size or Mix.Utils.stale?([last_mtime | times], [modified]) or
-            Enum.any?(modules, &Map.has_key?(modules_to_recompile, &1)),
+          Enum.any?(modules, &Map.has_key?(modules_to_recompile, &1)) or
+            Mix.Utils.stale?(times, [modified]) or
+            (size != last_size or (last_mtime > modified and digest != digest(source))),
           do: source
 
     changed = new_paths ++ changed
@@ -239,6 +241,12 @@ defmodule Mix.Compilers.Elixir do
         Map.put_new_lazy(map, file, fn -> Mix.Utils.last_modified_and_size(file) end)
       end)
     end)
+  end
+
+  defp digest(file) do
+    file
+    |> File.read!()
+    |> :erlang.md5()
   end
 
   defp compile_path(stale, dest, timestamp, opts) do
@@ -455,6 +463,7 @@ defmodule Mix.Compilers.Elixir do
     source =
       source(
         source,
+        digest: digest(file),
         compile_references: compile_references,
         export_references: export_references,
         runtime_references: runtime_references,
@@ -504,11 +513,11 @@ defmodule Mix.Compilers.Elixir do
   # to be recompiled (but were not changed on disk)
   defp update_stale_sources(sources, changed) do
     Enum.reduce(changed, {sources, %{}}, fn file, {acc_sources, acc_modules} ->
-      {source(size: size, modules: modules), acc_sources} =
+      {source(size: size, digest: digest, modules: modules), acc_sources} =
         List.keytake(acc_sources, file, source(:source))
 
       acc_modules = Enum.reduce(modules, acc_modules, &Map.put(&2, &1, true))
-      {[source(source: file, size: size) | acc_sources], acc_modules}
+      {[source(source: file, size: size, digest: digest) | acc_sources], acc_modules}
     end)
   end
 
--- a/lib/mix/lib/mix/compilers/elixir.ex
+++ b/lib/mix/lib/mix/compilers/elixir.ex
@@ -463,7 +463,8 @@ defmodule Mix.Compilers.Elixir do
     source =
       source(
         source,
-        digest: digest(file),
+        # We preserve the digest if the file is recompiled but not changed
+        digest: source(source, :digest) || digest(file),
         compile_references: compile_references,
         export_references: export_references,
         runtime_references: runtime_references,
--- a/lib/mix/test/mix/tasks/compile.elixir_test.exs
+++ b/lib/mix/test/mix/tasks/compile.elixir_test.exs
@@ -195,7 +195,39 @@ defmodule Mix.Tasks.Compile.ElixirTest do
     end)
   end
 
-  test "compiles mtime changed files" do
+  test "compiles mtime changed files if content changed but not length" do
+    in_fixture("no_mixfile", fn ->
+      assert Mix.Tasks.Compile.Elixir.run(["--verbose"]) == {:ok, []}
+      assert_received {:mix_shell, :info, ["Compiled lib/a.ex"]}
+      assert_received {:mix_shell, :info, ["Compiled lib/b.ex"]}
+
+      Mix.shell().flush
+      purge([A, B])
+
+      same_length_content = "lib/a.ex" |> File.read!() |> String.replace("A", "Z")
+      File.write!("lib/a.ex", same_length_content)
+      future = {{2038, 1, 1}, {0, 0, 0}}
+      File.touch!("lib/a.ex", future)
+      Mix.Tasks.Compile.Elixir.run(["--verbose"])
+
+      message =
+        "warning: mtime (modified time) for \"lib/a.ex\" was set to the future, resetting to now"
+
+      assert_received {:mix_shell, :error, [^message]}
+
+      message =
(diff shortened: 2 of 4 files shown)


josevalim:
@marcandre I propose we try a slightly different approach that does not require us to read all snapshots upfront. It can be made in two steps:

1. First store the MD5 in the manifest. You will add a [new field called md5 here](https://github.com/elixir-lang/elixir/blob/master/lib/mix/lib/mix/compilers/elixir.ex#L10) and also bump the `@manifest_vsn`. Now for each file compiled, you compute and store the MD5 here: https://github.com/elixir-lang/elixir/blob/master/lib/mix/lib/mix/compilers/elixir.ex#L447 - you can expect the file to exist at this point

2. Now in the line of code you removed in the other PR, you can compare the MD5, fetching it from disk if and only if the mtime is outdated and the size is the same

One thing to notice on step 2 is that we are comparing multiple mtimes. For example, imagine a Phoenix.View. It has several templates on the file system, so while the view is the same, we still need to recompile it if a template changes! So this check:

      for source(source: source, external: external, size: size, modules: modules) <-
            all_sources,
          {last_mtime, last_size} = Map.fetch!(sources_stats, source),
          times = Enum.map(external, &(sources_stats |> Map.fetch!(&1) |> elem(0))),
          size != last_size or Mix.Utils.stale?([last_mtime | times], [modified]) or
            Enum.any?(modules, &Map.has_key?(modules_to_recompile, &1)),
          do: source

Should become something like this:

      for source(source: source, external: external, size: size, modules: modules, md5: md5) <-
            all_sources,
          times = Enum.map(external, &(sources_stats |> Map.fetch!(&1) |> elem(0))),
          source_changed?(source_stats, source, md5, modified) or
            Mix.Utils.stale?(times, [modified]) or
            Enum.any?(modules, &Map.has_key?(modules_to_recompile, &1)),
          do: source

Where `source_changed?` has the size, mtime, and digest logic for the source file exclusively. It is important to not mix those. :)

josevalim:
Btw, a future step would be to keep the size and digests for those external files too, such as phoenix templates, but it can be a separate PR!

marcandre:
Great, thanks for these pointers 👍

lib/mix/lib/mix/compilers/elixir.ex
```diff
@@ -221,17 +223,32 @@ defmodule Mix.Compilers.Elixir do
...
-        Map.put_new_lazy(map, file, fn -> Mix.Utils.last_modified_and_size(file) end)
+        Map.put_new_lazy(map, file, fn -> snapshot(file) end)
       end)
     end)
   end
 
+  defp digest(file) do
+    case File.read(file) do
+      {:ok, content} ->
+        :crypto.hash(:md5, content)
```
josevalim, line 240:
```suggestion
        :erlang.md5(content)
```

The difference is that `:crypto` is a separate application, that we would need to start and depend on, but luckily MD5 is part of Erlang too.

marcandre:
And yes, clearly, the hashes have to be written to the manifest...

marcandre:
I'll work on that tomorrow...

josevalim:
There is absolutely no rush. I plan to do a new Elixir patch release this week with the faster app loading and so it would be too soon to include this (I typically like to run with those changes for a couple weeks).

marcandre:
This should be more like it

vans163:
> @marcandre I propose we try a slightly different approach that does not require us to read all snapshots upfront. It can be made in two steps:
> 
> 1. First store the MD5 in the manifest. You will add a [new field called md5 here](https://github.com/elixir-lang/elixir/blob/master/lib/mix/lib/mix/compilers/elixir.ex#L10) and also bump the `@manifest_vsn`. Now for each file compiled, you compute and store the MD5 here: https://github.com/elixir-lang/elixir/blob/master/lib/mix/lib/mix/compilers/elixir.ex#L447 - you can expect the file to exist at this point
> 2. Now in the line of code you removed in the other PR, you can compare the MD5, fetching it from disk if and only if the mtime is outdated and the size is the same
> 
> One thing to notice on step 2 is that we are comparing multiple mtimes. For example, imagine a Phoenix.View. It has several templates on the file system, so while the view is the same, we still need to recompile it if a template changes! So this check:
> 
> ```
>   for source(source: source, external: external, size: size, modules: modules) <-
>         all_sources,
>       {last_mtime, last_size} = Map.fetch!(sources_stats, source),
>       times = Enum.map(external, &(sources_stats |> Map.fetch!(&1) |> elem(0))),
>       size != last_size or Mix.Utils.stale?([last_mtime | times], [modified]) or
>         Enum.any?(modules, &Map.has_key?(modules_to_recompile, &1)),
>       do: source
> ```
> 
> Should become something like this:
> 
> ```
>   for source(source: source, external: external, size: size, modules: modules, md5: md5) <-
>         all_sources,
>       times = Enum.map(external, &(sources_stats |> Map.fetch!(&1) |> elem(0))),
>       source_changed?(source_stats, source, md5, modified) or
>         Mix.Utils.stale?(times, [modified]) or
>         Enum.any?(modules, &Map.has_key?(modules_to_recompile, &1)),
>       do: source
> ```
> 
> Where `source_changed?` has the size, mtime, and digest logic for the source file exclusively. It is important to not mix those. :)

Step 2 fails if filesystem mounted with nomtime or the filesystem does not support/properly set mtime. Possibly check if mtime is enabled on filesystem?

josevalim:
The whole code today already assumes a file system with mtime and it has never been N issue, especially for development. If this is an issue for someone, then we are glad to discuss options, but it is not a concern for this PR.

lib/mix/lib/mix/compilers/elixir.ex
```diff
@@ -455,6 +463,7 @@ defmodule Mix.Compilers.Elixir do
     source =
       source(
         source,
+        digest: digest(file),
```
josevalim, line 466: Your change in `update_stale_sources` means if a file is recompiled but not changed, you are keeping the digest, so I think we can leverage it here. I will push this change and see how the test suite behaves. :)

```suggestion
        # We preserve the digest if the file is recompiled but not changed
        digest: source(source, :digest) || digest(file),
```

josevalim:

josevalim:
Thank you @marcandre! As I mentioned, we don't run this check for external resources (for example, all templates in a Phoenix view are external resources), so this may be a necessary change in the future if you (or anyone else) is inclined.

marcandre:
Awesome, thanks!
I don't know when I'll have the time to look at external resources; if anyone is interested in picking it up that's great too

marcandre:
Is there a timeline for this to be released?

```
[15:31][~/super-pas((no branch, rebasing ch2579/revert-data-plan-change))]$ mix test something
Compiling 357 files (.ex)
```

josevalim:
@marcandre I found a bug in master about 10 days but it has been running smoothly since then. I will backport it to v1.12 branch and it should be out in one or two weeks.

josevalim:
@marcandre btw, if you could run elixir from the v1.12 branch, then it would be super helpful too to help catch any regressions and you would have the benefits today. :)

josevalim:
The website and readme have instructions to install/compile from source if needed.

marcandre:
> @marcandre btw, if you could run elixir from the v1.12 branch, then it would be super helpful too to help catch any regressions and you would have the benefits today. :)

Perfect, will do 👍

marcandre:
I was disappointed to get the dreaded "Compiling 392 files (.ex)" when using it locally, only to realize that my branch was modifying `config/test.exs`, which bumps `Mix.Project.config_mtime` and invalidates the whole project.

Although this happens less frequently, it would be really nice to fix this too, so that only actual changes to files invalidate the project.

A "quick" fix might be for `Mix.Project.config_mtime` to use the last stored `mtime` if the digest has not changed, but it's not clear to me where to store this information. Any pointers?

josevalim:
@marcandre we already have a manifest where we store the elixir and otp versions, and we could store the config digests in there. It is currently kept in `Mix.Dep.ElixirSCM` and called in mix/tasks/loadpaths.ex. It is used both by your project and elixir deps, so we would need to make the config digest be an empty map in the deps case.

marcandre:
Thank you, I'll have a look when I get a bit of time.

vans163:
> The whole code today already assumes a file system with mtime and it has never been an issue, especially for development. If this is an issue for someone, then we are glad to discuss options, but it is not a concern for this PR.

an example is sshfs has mtime issues

marcandre:
@josevalim May not be related (directly or at all), but I'm getting some strange compiling errors today that resolve if I recompile the whole `lib/`:

```
$ mix phx.server
Compiling 5 files (.ex)

== Compilation error in file lib/bobby_web/views/admin/account_view.ex ==
** (CompileError) lib/bobby_web/views/admin/account_view.ex:2: module Bobby is not loaded and could not be found. This may be happening because the module you are trying to load directly or indirectly depends on the current module
    (elixir 1.12.2) expanding macro: Kernel.use/2
    lib/bobby_web/views/admin/account_view.ex:2: BobbyWeb.Admin.AccountView (module)
    expanding macro: BobbyWeb.__using__/1
    lib/bobby_web/views/admin/account_view.ex:2: BobbyWeb.Admin.AccountView (module)
    (elixir 1.12.2) expanding macro: Kernel.use/2
    lib/bobby_web/views/admin/account_view.ex:2: BobbyWeb.Admin.AccountView (module)
    (elixir 1.12.2) lib/kernel/parallel_compiler.ex:319: anonymous fn/4 in Kernel.ParallelCompiler.spawn_workers/7
$ touch mix.exs
$ mix phx.server
Compiling 405 files (.ex)
Compiling lib/postgres_types.ex (it's taking more than 10s)
Compiling lib/bobby_web/views/documents/life_view.ex (it's taking more than 10s)
Generated bobby app
```

`Bobby` is our `MyApp`. Just about everywhere we do `use Bobby`.

I am not sure how to reproduce it, but it has happened 3 times today (and never before today, no notable to our codebase either 🤷)

This is running elixir from 1.12 branch (at `5189b292a`, I'm missing an unrelated commit)

Any idea what could be happening?

josevalim:
@marcandre next time it happens, please try passing the `--verbose` flag. You should see all five files printed before it fails and let's see if lib/bobby.ex is in the list of files.

josevalim:
Oh, and please see if touching only the file that defines Bobby next fixes it.

PR #11083, opened 2021-06-23
Optimize Enum.zip/2

sabiwara:
Hi! I noticed that the generic implementation of `Enum.zip/2` relying on `zip_with` with an anonymous function was significantly slower on lists than having dedicated private functions.
Some quick benchmarks suggest a speedup of 2.5~3x compared to the current implementation:
https://github.com/sabiwara/elixir_benches/blob/main/bench/fast_enum_zip.results.txt

Note: I went with a tail-recursive implementation, but I benchmarked [a body-recursive version](https://github.com/sabiwara/elixir_benches/commit/608d9fd786061e8d719331f9a706770fc8062796) as well and it seems noticeably slower (due to the JIT?) so I kept the former.

### DIFF
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -4228,13 +4228,12 @@ defmodule Enum do
 
   ## zip
 
-  defp zip_list(list1, list2, acc) when list1 == [] or list2 == [] do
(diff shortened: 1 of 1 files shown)


josevalim:

PR #11084, opened 2021-06-23
Fix normalization of partial keyword list elements

doorgan:
The Elixir.Normalizer lets us feed any arbitrary AST to the formatter, but if the user provides a list, and an element is a 2-tuple that is not one of the last elements, and the first element is a wrapped atom with `format: :keyword` metadata, then the formatter will happily print it as a keyword list element, despite it being invalid syntax, for example:
```elixir
iex> opts = [literal_encoder: &{:ok, {:__block__, &2, [&1]}}]
iex> {_, _, [[tuple]]} = Code.string_to_quoted("[a: b]", opts)
iex> Macro.to_string([tuple, :not_a_tuple, tuple])
"[a: b, :not_a_tuple, a: b]"
```
With this PR we would get the expected output:
```elixir
iex> Macro.to_string([tuple, :not_a_tuple, tuple])
"[{:a, b}, :not_a_tuple, a: b]"
```

### DIFF
--- a/lib/elixir/lib/code/normalizer.ex
+++ b/lib/elixir/lib/code/normalizer.ex
@@ -419,8 +419,28 @@ defmodule Code.Normalizer do
 
   defp normalize_kw_args(elems, state, keyword? \\ false)
 
+  defp normalize_kw_args(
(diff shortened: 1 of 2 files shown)


josevalim:

PR #11090, opened 2021-06-28
Remove unreachable code in String module

michallepicki:
The private `byte_size_remaining_at` function does not return a falsy value.


josevalim:
Absolutely beautiful! :green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

PR #11091, opened 2021-06-28
Fix Inspect.Algebra.color/3 return type in spec

michallepicki:
It returns Inspect.Algebra.t, just like concat/2 which it calls last


josevalim:

PR #11094, opened 2021-06-29
`Map.values/2` to retrieve values in predicted order

am-kantox:
`Map.values/1` delegates to `:maps.values/1` and since erlang maps are not ordered (when size is greater than 32,) the result of `Map.values(map)` might be confusing when `map_size(map) > 32`.

```elixir
list = Enum.to_list(1..33)
map = list |> Enum.zip(list) |> Map.new()

Map.values(map)   
#⇒ [33, 12, 23, 29, 30, 26, 31, 11, 9, …]
```

This PR provides `Map.values/2` function that accepts a list of keys to order returned values accordingly. This function might be particularly helpful when working with CSV-like inputs of considerable size using `Flow`, which is more friendly to maps in general and `key: {:key, name}` working with maps only in particular.

```elixir
Map.values(map, list)
#⇒ [1, 2, 3, 4, 5, 6, 7, 8, 9, …]
```

Despite this function might be easily implemented whenever needed outside of the core library, its existence would be a good hint reminding everybody that maps are not ordered and one should not rely on `Map.values/1` returning predictable and deterministic results.

### DIFF
--- a/lib/elixir/lib/map.ex
+++ b/lib/elixir/lib/map.ex
@@ -132,7 +132,10 @@ defmodule Map do
   defdelegate keys(map), to: :maps
 
   @doc """
-  Returns all values from `map`.
+  Returns all values from `map`. Please note, that maps in `erlang` and `elixir`
+  are not ordered, meaning that the order of values would not be defined.
+
+  To retrieve a list of values in a predicted order, use `values/2`.
 
   Inlined by the compiler.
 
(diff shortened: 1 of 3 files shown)


wojtekmach:
I always wanted to call such function `Map.take_values!` and knew there had to be a non-raising variant but I didn't have a use case for it. And thus never realised we have a choice between returning nil or not including the empty element. So good call on not trying to solve this particular problem. :)

PR #11095, opened 2021-06-30
Exit with a status of 1 when coverage falls below threshold

starbelly:
This commit adjusts Mix.Tasks.Test.Coverage to exit with a status of 1 when coverage falls below the default or specified threshold. The exception to this if summary in coverage options is set to false.

- Changed Mix.Tasks.Test.Coverage to exit(1) if threshold is below coverage unless summary is set to false
- Added tests and fixtures for three scenarios
    - coverage above threshold
    - coverage below threshold
    -  summary set to false
- Updated documentation in Mix.Tasks.Test


starbelly:
Note: I can only think of one other thing to possibly add here, an option that allows you to see the summary, but does not exit with a status 1. I can't think of why you would want that, but it should be considered.

lib/mix/lib/mix/tasks/test.ex
```diff
@@ -275,6 +275,8 @@ defmodule Mix.Tasks.Test do
       `:threshold`, which is a boolean or numeric value that enables coloring
       of code coverage results in red or green depending on whether the
       percentage is below or above the specified threshold, respectively.
+      If summary is true or a threshold value is set the task will exit with status
+      of 1.
       Defaults to `[threshold: 90]`
```
josevalim, line 280:
```suggestion
      Defaults to `[threshold: 90]`.  The task will exit with status of 1
      if there are modules with coverage below the threshold, unless the
      summary feature is set to false.
```

lib/mix/test/mix/tasks/test_test.exs
```diff
@@ -183,6 +183,24 @@ defmodule Mix.Tasks.TestTest do
...
+      in_fixture("umbrella_test", fn ->
+        assert {_, 0} = mix_code(["test", "--cover"])
+      end)
+    end
+
+    test "exits with signal 0 when summary is false" do
+      in_fixture("cover_threshold_off_test", fn ->
+        assert {_, 0} = mix_code(["test", "--cover"])
+      end)
+    end
```
josevalim, line 203: Those tests are very expensive, so we want to keep them to a minimum. We also generally prefer to change existing features than adding a new feature per case, so I propose:

1. For checking status of 0, please change the existing tests that already call --cover

2. For checking the status of 1, get one of the existing fixtures and change the test files so coverage is low (for example, you can remove one of the test files)

3. There is no need to test what happens when summary is false

At the end no new fixtures should be necessary!

lib/mix/lib/mix/tasks/test.coverage.ex
```diff
@@ -228,6 +228,14 @@ defmodule Mix.Tasks.Test.Coverage do
...
     module_results = Enum.sort(module_results, :desc)
     print_summary(module_results, totals, summary_opts)
+
+    case totals < threshold(summary_opts) do
+      true ->
+        exit({:shutdown, 1})
+
+      false ->
+        :ok
+    end
```
josevalim, line 238: We don't want to fail immediately because if you are in an umbrella, you want to show the coverage of all umbrella apps and then exit. So instead please do this:

```suggestion
    if totals < threshold(summary_opts) do
      System.at_exit(fn _ -> exit({:shutdown, 1}) end)
    end

    :ok
```

lib/mix/lib/mix/tasks/test.coverage.ex
```diff
@@ -278,10 +286,8 @@ defmodule Mix.Tasks.Test.Coverage do
   end
 
   defp display({percentage, name}, opts) do
-    threshold = Keyword.get(opts, :threshold, @default_threshold)
-
     Mix.shell().info([
-      color(percentage, threshold),
+      color(percentage, threshold(opts)),
```
josevalim, line 290: Let's compute the threshold before we call `display` and pass it as an argument to `display`.

josevalim (commented):
Thank you for the PR, comments added!

starbelly:
@josevalim  feedback addressed.

lib/mix/lib/mix/tasks/test.coverage.ex
```diff
@@ -227,7 +227,13 @@ defmodule Mix.Tasks.Test.Coverage do
   defp summary(results, keep, summary_opts) do
     {module_results, totals} = gather_coverage(results, keep)
     module_results = Enum.sort(module_results, :desc)
-    print_summary(module_results, totals, summary_opts)
+    print_summary(module_results, totals, threshold, summary_opts)
+
+    if totals < get_threshold(summary_opts) do
```
starbelly, line 232: @josevalim It may be worth passing the threshold from here down to print_summary, thoughts?

lib/mix/lib/mix/tasks/test.ex
```diff
@@ -275,7 +275,9 @@ defmodule Mix.Tasks.Test do
       `:threshold`, which is a boolean or numeric value that enables coloring
       of code coverage results in red or green depending on whether the
       percentage is below or above the specified threshold, respectively.
-      Defaults to `[threshold: 90]`
+      Defaults to `[threshold: 90]`.  The task will exit with status of 1 if
```
josevalim, line 278:
```suggestion
      Defaults to `[threshold: 90]`. The task will exit with status of 1 if
```

josevalim:

PR #11097, opened 2021-06-30
Allow using XDG-conforming path for IEx config

hauleth:
This adds new path - `$XDG_CONFIG_HOME/elixir/iex.exs` as a possible path for the configuration of IEx. If this file exists it will have higher priority over `$HOME/.iex.exs`.

### DIFF
--- a/lib/iex/lib/iex/evaluator.ex
+++ b/lib/iex/lib/iex/evaluator.ex
@@ -224,7 +224,8 @@ defmodule IEx.Evaluator do
       if path do
         [path]
       else
-        Enum.map([".iex.exs", "~/.iex.exs"], &Path.expand/1)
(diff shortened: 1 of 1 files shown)


josevalim:
We have no plans to add further lookups at the moment unless the Erlang VM is moving most of its config to XDG too. Thanks!

hauleth:
Erlang is using only `$HOME/.erlang.cookie` AFAIK (my searching through the Erlang source agree with me unless I missed something important, with only exception for `ssh` application for obvious reasons). [Rebar3 is using XDG-compatible paths for some time now](https://github.com/erlang/rebar3/pull/166) (however I see that they do that a little bit incorrectly, so I will probably do PR to fix that).

josevalim:
There is also .erlangrc which is not XDG aware I assume?

hauleth:
@josevalim ok, it seems so. I can check out how to make OTP use XDG (or platform-specific) approach as well.

PR #11101, opened 2021-07-01
add `Task.completed/1`

superhawk610:
This PR implements the new `Task` function `Task.completed/1` as proposed in the [elixir-lang group](https://groups.google.com/g/elixir-lang-core/c/ickqmBBoSMk/m/EGp507x9AAAJ?utm_medium=email&utm_source=footer).

This function models a task that immediately completes with a known result. As documented in the `@doc` block, it's a useful convenience when working with groups of asynchronous tasks where some inputs may have a result that can be determined without actually running the task. Here's a quick example:

```elixir
def process(data) do
  tasks =
    for entry <- data do
      if invalid_input?(entry) do
        Task.completed({:error, :invalid_input})
      else
        Task.async(fn -> further_process(entry) end)
      end
    end

  Task.await_many(tasks)
end
```

Here are a couple links to prior art in .NET:

https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.completedtask?view=net-5.0

https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.fromresult?view=net-5.0

Credit for the idea goes to @lukebakken.

### DIFF
--- a/lib/elixir/lib/task.ex
+++ b/lib/elixir/lib/task.ex
@@ -442,6 +442,48 @@ defmodule Task do
   end
 
   @doc """
+  Starts a task that immediately completes with the given `result`.
+
+  Unlike `async/1`, this task does not spawn a linked process. It can
+  be awaited or yielded like any other task.
+
+  ## Usage
+
+  In some cases, it is useful to create a "completed" task that represents
+  a task that has already run and generated a result. For example, when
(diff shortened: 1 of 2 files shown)


lukebakken (approved):
Great!

josevalim:

PR #11112, opened 2021-07-12
Augment documentation for `Kernel.match?/2`; counsel caution when using the pin operator

jmilamwalters:
## Description

This PR serves:

- To augment the annotation surrounding `Kernel.match?/2`; and specifically,
- To counsel caution when using the pin operator, namely by citing a misuse of the pin operator. 

## Context

I initially posted the [following question](https://elixir-lang.slack.com/archives/C03EPRA3B/p1626111373286800) in the `#general` channel of Elixir Slack:

> I’m curious why:
> 
> ```
> iex> match?(%{x: 1, y: 2}, %{x: 1, y: 2, z: 3})
> true
> iex> attrs = %{x: 1, y: 2}
> iex> match?(^attrs, %{x: 1, y: 2, z: 3})
> false
> ```
>
> ?

@LostKobrakai helped me to understand that I was misusing the pin operator. I've reworked @LostKobrakai's comments from Slack into the proposed annotations. 

The behaviour has snagged me a few times, and hence, I'm wondering whether we might augment the documentation surrounding `Kernel.match/2`? If only to help guide others who also misuse the pin operator? 

If you all find the enclosed annotations impertinent, I understand that, too! Thank you.

### DIFF
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -3247,20 +3247,21 @@ defmodule Kernel do
 
   Furthermore, please exercise caution when using the pin operator.
 
-      iex> match?(%{x: 1}, %{x: 1, y: 2})
-      true
-      iex> attrs = %{x: 1}
-      iex> match?(^attrs, %{x: 1, y: 2})
-      false
+      match?(%{x: 1}, %{x: 1, y: 2})
+      #=> true
+
+      attrs = %{x: 1}
+      match?(^attrs, %{x: 1, y: 2})
+      #=> false
 
   As a reminder, the pin operator binds _values,_ not _patterns_ (on the BEAM,
   patterns cannot be stored or composed in variables, as patterns are not a data
   type). Such behavior is not specific to `Kernel.match/2`. The following code
   will throw an exception:
 
-      iex> attrs = %{x: 1}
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/kernel.ex
```diff
@@ -3245,6 +3245,24 @@ defmodule Kernel do
...
+      #=> true
+
+      attrs = %{x: 1}
+      match?(^attrs, %{x: 1, y: 2})
+      #=> false
+
+  As a reminder, the pin operator binds _values,_ not _patterns_ (on the BEAM,
+  patterns cannot be stored or composed in variables, as patterns are not a data
+  type). Such behaviour is not specific to `Kernel.match/2`. The following code
+  will throw an exception:
```
josevalim, line 3260:
```suggestion
  Furthermore, remember the pin operator matches _values_, not _patterns_:

      match?(%{x: 1}, %{x: 1, y: 2})
      #=> true

      attrs = %{x: 1}
      match?(^attrs, %{x: 1, y: 2})
      #=> false

  The pin operator will check if the values are equal, using `===/2`, while
  patterns have their own rules when matching maps, lists, and so forth.
  Such behaviour is not specific to `match?/2`. The following code also
  throws an exception:
```

josevalim:

PR #11113, opened 2021-07-12
Support sparse patterns in Mix.SCM.Git

jjcarstens:
`git sparse-checkout` supports the same pattern matching as `.gitignore` files (see [this article](https://github.blog/2020-01-17-bring-your-monorepo-down-to-size-with-sparse-checkout/#cone-mode-restricted-patterns-improve-performance) for details). However, the current implementation of the `:sparse` option with mix breaks if you try to use a pattern:

```elixir
{:sparse_test, github: "jjcarstens/sparse_test", sparse: "lib/**/*.ex"}
```
 
```
> my_app √ % mix deps.get                                                                                                                                                                         
* Updating sparse_test (https://github.com/jjcarstens/sparse_test.git)
origin/HEAD set to main
** (File.Error) could not touch "/Users/jonjon/repos/my_app/deps/sparse_test/lib/**/*.ex/.fetch": no such file or directory
    (elixir 1.12.1) lib/file.ex:602: File.touch!/2
    (mix 1.12.1) lib/mix/dep/fetcher.ex:78: Mix.Dep.Fetcher.do_fetch/3
    (mix 1.12.1) lib/mix/dep/converger.ex:190: Mix.Dep.Converger.all/9
    (mix 1.12.1) lib/mix/dep/converger.ex:201: Mix.Dep.Converger.all/9
    (mix 1.12.1) lib/mix/dep/converger.ex:123: Mix.Dep.Converger.all/7
    (mix 1.12.1) lib/mix/dep/converger.ex:72: Mix.Dep.Converger.all/4
    (mix 1.12.1) lib/mix/dep/converger.ex:51: Mix.Dep.Converger.converge/4
    (mix 1.12.1) lib/mix/dep/fetcher.ex:16: Mix.Dep.Fetcher.all/3
```

This is because `:sparse` is being added to `:dest` option which later on attempts to be created via a `touch`, but patterns are not supported.

So this instead removes the update of the `:dest` option and uses the `:sparse` option only when needed. As a bonus, this also adds support for including multiple patterns as a list in `:sparse` to create more complex sparse checkouts as needed.

### DIFF
--- a/lib/mix/lib/mix/scm/git.ex
+++ b/lib/mix/lib/mix/scm/git.ex
@@ -31,10 +31,7 @@ defmodule Mix.SCM.Git do
   end
 
   def accepts_options(_app, opts) do
-    opts =
-      opts
-      |> Keyword.put(:checkout, opts[:dest])
-      |> sparse_opts()
+    opts = Keyword.put(opts, :checkout, opts[:dest])
 
     cond do
       gh = opts[:github] ->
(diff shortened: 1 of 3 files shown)


josevalim:
We need to keep the `:dest`code if we want to support subdirectories inside a repo, like [Nx/EXLA/Torchx](https://github.com/elixir-nx/nx) work right now.

If we want to introduce this, we would need to do so via another option, maybe `:sparse_patterns` (or `:filter`, kind of keeping the sparse stuff as an implementation detail). Therefore it is probably best to open up a discussion in the mailing list and include some examples of where this feature could be useful, because I can't see a use case for a pattern such as `sparse: "lib/**/*.ex"`. :)

Thanks!

PR #11116, opened 2021-07-13
Add 'significant' to child_spec()

rlipscombe:
OTP-24.0 adds the `significant` option to `supervisor:child_spec()`. This was inlined in https://github.com/elixir-lang/elixir/commit/6632cab46641a66d6c983e6b0401f37ab922bbfc, so dialyzer (via dialyxir) doesn't see it.

### DIFF
--- a/lib/elixir/lib/supervisor.ex
+++ b/lib/elixir/lib/supervisor.ex
@@ -515,6 +515,7 @@ defmodule Supervisor do
           required(:id) => atom() | term(),
           required(:start) => {module(), atom(), [term()]},
           optional(:restart) => :permanent | :transient | :temporary,
+          optional(:significant) => boolean(),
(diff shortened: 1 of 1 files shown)


rlipscombe:
Does this need extra documentation? Does it need annotating with the required OTP version? Should it be un-inlined instead?

Related: https://elixirforum.com/t/dialyxir-complains-about-supervisor-and-significant-children/41005

josevalim:
`:significant` is not supported in Elixir right now. In theory it works, because we simply use `:supervisor`, but that is an implementation detail. For example, a DynamicSupervisor does not respect the significant flag.

We typically avoid introducing versions that require certain OTP versions to avoid complicating the versioning mechanism. Therefore, if you need to use significant, I suggest to use the `:supervisor` module directly. We will officially implement this on #10433.

Thanks!

PR #11117, opened 2021-07-14
Keep backwards compatibility on Range and Date.Range

lukaszsamson:
Fixes https://github.com/elixir-lang/elixir/issues/11110
Handles cases not covered by https://github.com/elixir-lang/elixir/commit/13ba96be3cef5e4325fc3570f65e2eef4cce9314

There are 2 more usages but those look safe. If I understand it correctly they happen only on compile time
- https://github.com/elixir-lang/elixir/blob/58ca6524f88bfe7e963deafe0e7d6aa07a568003/lib/elixir/lib/kernel.ex#L4071
- https://github.com/elixir-lang/elixir/blob/58ca6524f88bfe7e963deafe0e7d6aa07a568003/lib/elixir/lib/kernel.ex#L4077

### DIFF
--- a/lib/elixir/lib/calendar/date_range.ex
+++ b/lib/elixir/lib/calendar/date_range.ex
@@ -27,13 +27,15 @@ defmodule Date.Range do
   defstruct [:first, :last, :first_in_iso_days, :last_in_iso_days, :step]
 
   defimpl Enumerable do
-    def member?(%{first: %{calendar: calendar}} = range, %Date{calendar: calendar} = date) do
-      %{
-        first_in_iso_days: first_days,
-        last_in_iso_days: last_days,
-        step: step
-      } = range
-
+    def member?(
+          %Date.Range{
+            first: %{calendar: calendar},
+            first_in_iso_days: first_days,
+            last_in_iso_days: last_days,
+            step: step
+          } = range,
+          %Date{calendar: calendar} = date
+        ) do
       {days, _} = Date.to_iso_days(date)
 
       cond do
@@ -48,22 +50,41 @@ defmodule Date.Range do
       end
     end
 
-    def member?(_, _) do
+    def member?(%Date.Range{step: _}, _) do
       {:ok, false}
+    end
+
+    # TODO: Remove me on v2.0
+    def member?(
+          %{__struct__: Date.Range, first_in_iso_days: first_days, last_in_iso_days: last_days} =
+            date_range,
+          date
+        ) do
+      step = if first_days <= last_days, do: 1, else: -1
+      member?(Map.put(date_range, :step, step), date)
     end
 
     def count(range) do
       {:ok, size(range)}
     end
 
-    def slice(range) do
-      %{
-        first_in_iso_days: first,
-        first: %{calendar: calendar},
-        step: step
-      } = range
-
+    def slice(
+          %Date.Range{
+            first_in_iso_days: first,
+            first: %{calendar: calendar},
+            step: step
+          } = range
+        ) do
       {:ok, size(range), &slice(first + &1 * step, step, &2, calendar)}
+    end
+
+    # TODO: Remove me on v2.0
+    def slice(
+          %{__struct__: Date.Range, first_in_iso_days: first_days, last_in_iso_days: last_days} =
+            date_range
+        ) do
+      step = if first_days <= last_days, do: 1, else: -1
+      slice(Map.put(date_range, :step, step))
     end
 
     defp slice(current, _step, 1, calendar) do
@@ -77,15 +98,28 @@ defmodule Date.Range do
       ]
     end
 
-    def reduce(range, acc, fun) do
-      %{
-        first_in_iso_days: first_days,
-        last_in_iso_days: last_days,
-        first: %{calendar: calendar},
-        step: step
-      } = range
-
+    def reduce(
+          %Date.Range{
+            first_in_iso_days: first_days,
+            last_in_iso_days: last_days,
+            first: %{calendar: calendar},
+            step: step
+          },
+          acc,
+          fun
+        ) do
       reduce(first_days, last_days, acc, fun, step, calendar)
+    end
+
+    # TODO: Remove me on v2.0
+    def reduce(
+          %{__struct__: Date.Range, first_in_iso_days: first_days, last_in_iso_days: last_days} =
+            date_range,
+          acc,
+          fun
+        ) do
+      step = if first_days <= last_days, do: 1, else: -1
+      reduce(Map.put(date_range, :step, step), acc, fun)
     end
 
     defp reduce(_first_days, _last_days, {:halt, acc}, _fun, _step, _calendar) do
@@ -136,6 +170,15 @@ defmodule Date.Range do
     defp size(%Date.Range{first_in_iso_days: first_days, last_in_iso_days: last_days, step: step}),
       do: abs(div(last_days - first_days, step)) + 1
 
+    # TODO: Remove me on v2.0
+    defp size(
+           %{__struct__: Date.Range, first_in_iso_days: first_days, last_in_iso_days: last_days} =
+             date_range
+         ) do
+      step = if first_days <= last_days, do: 1, else: -1
+      size(Map.put(date_range, :step, step))
+    end
+
     defp empty?(%Date.Range{
            first_in_iso_days: first_days,
            last_in_iso_days: last_days,
@@ -152,16 +195,33 @@ defmodule Date.Range do
          when step < 0 and first_days < last_days,
          do: true
 
-    defp empty?(%Date.Range{}), do: false
+    defp empty?(%Date.Range{step: _}), do: false
+
+    # TODO: Remove me on v2.0
+    defp empty?(
+           %{__struct__: Date.Range, first_in_iso_days: first_days, last_in_iso_days: last_days} =
+             date_range
+         ) do
+      step = if first_days <= last_days, do: 1, else: -1
+      empty?(Map.put(date_range, :step, step))
+    end
   end
 
   defimpl Inspect do
+    import Kernel, except: [inspect: 2]
+
     def inspect(%Date.Range{first: first, last: last, step: 1}, _) do
       "#DateRange<" <> inspect(first) <> ", " <> inspect(last) <> ">"
     end
 
     def inspect(%Date.Range{first: first, last: last, step: step}, _) do
       "#DateRange<" <> inspect(first) <> ", " <> inspect(last) <> ", #{step}>"
+    end
+
+    # TODO: Remove me on v2.0
+    def inspect(%{__struct__: Date.Range, first: first, last: last} = date_range, opts) do
+      step = if first <= last, do: 1, else: -1
+      inspect(Map.put(date_range, :step, step), opts)
     end
   end
 end
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -2066,6 +2066,12 @@ defmodule Enum do
     end
   end
 
+  # TODO: Remove me on v2.0
+  def min_max(%{__struct__: Range, first: first, last: last} = range, empty_fallback) do
+    step = if first <= last, do: 1, else: -1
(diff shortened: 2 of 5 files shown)


lib/elixir/lib/enum.ex
```diff
@@ -2066,6 +2066,12 @@ defmodule Enum do
     end
   end
 
+  # TODO: Remove me on v2.0
+  def min_max(%{__struct__: Range, first: first, last: last} = range, empty_fallback) do
```
josevalim, line 2070: I believe we don't need this clause because it will fallback to the protocol.

lib/elixir/lib/enum.ex
```diff
@@ -2390,6 +2396,12 @@ defmodule Enum do
     reduce_range(first, last, step, acc, fun)
   end
 
+  # TODO: Remove me on v2.0
+  def reduce(%{__struct__: Range, first: first, last: last} = range, acc, fun) do
```
josevalim, line 2400: I believe we don't need this clause because it will fallback to the protocol.

lib/elixir/lib/enum.ex
```diff
@@ -2662,6 +2674,12 @@ defmodule Enum do
     end
   end
 
+  # TODO: Remove me on v2.0
+  def slice(enumerable, %{__struct__: Range, first: first, last: last} = index_range) do
```
josevalim, line 2678: I believe we don't need this clause because it will fallback to the protocol.

lib/elixir/lib/enum.ex
```diff
@@ -3050,6 +3068,12 @@ defmodule Enum do
     |> div(2)
   end
 
+  # TODO: Remove me on v2.0
+  def sum(%{__struct__: Range, first: first, last: last} = range) do
```
josevalim, line 3072: I believe we don't need this clause because it will fallback to the protocol.

lib/elixir/lib/enum.ex
```diff
@@ -3671,6 +3695,12 @@ defmodule Enum do
     end
   end
 
+  # TODO: Remove me on v2.0
+  defp aggregate(%{__struct__: Range, first: first, last: last} = range, fun, empty) do
```
josevalim, line 3699: I believe we don't need this clause because it will fallback to the protocol.

lib/elixir/lib/exception.ex
```diff
@@ -261,6 +261,11 @@ defmodule Exception do
       {:%{}, meta, [__struct__: Range, first: first, last: last, step: step]} ->
         {:"..//", meta, [first, last, step]}
 
+      # TODO: Remove me on v2.0
+      {:%{}, meta, [__struct__: Range, first: first, last: last]} ->
+        step = if first <= last, do: 1, else: -1
+        {:"..//", meta, [first, last, step]}
```
josevalim, line 267: We don't need this clause because then it will be printed as the struct when formatting arguments in the error message, which is fine.

josevalim:
Thanks @lukaszsamson, however I believe those changes are not necessary, except for the daterange ones. Can you please keep only them and add some tests? thank you.

lib/elixir/lib/enum.ex
lukaszsamson:
```
iex(7)> Enum.min_max(old_range, 9)                                           
** (FunctionClauseError) no function clause matching in Enum.min_max/2    
    
    The following arguments were given to Enum.min_max/2:
    
        # 1
        10..1//-1
    
        # 2
        9
    
        def min_max(first..last//step = range, empty_fallback) when is_function(empty_fallback, 0)
        def min_max(enumerable, empty_fallback) when is_function(empty_fallback, 0)
    
    (elixir 1.12.2) lib/enum.ex:2057: Enum.min_max/2
```

lib/elixir/lib/enum.ex
lukaszsamson: You're right
```
iex(6)> Enum.reduce(old_range, 0, fn a, b -> a + b end)
55
```

lib/elixir/lib/enum.ex
lukaszsamson:
```
iex(7)> Enum.slice([1, 2, 3, 4, 5, 6, 7, 8, 9, 0, 11, 12, 13, 14], old_range)
** (FunctionClauseError) no function clause matching in Enum.slice/2    
    
    The following arguments were given to Enum.slice/2:
    
        # 1
        [1, 2, 3, 4, 5, 6, 7, 8, 9, 0, 11, 12, 13, 14]
    
        # 2
        10..1//-1
    
        def slice(enumerable, first..last//step = index_range)
    
    (elixir 1.12.2) lib/enum.ex:2649: Enum.slice/2
```

lib/elixir/lib/enum.ex
lukaszsamson: You're right
```
iex(4)> Enum.sum(old_range)    
55
```

lib/elixir/lib/enum.ex
lukaszsamson: You're right
```
iex(9)> Enum.max(old_range)                                                 
10
```

lib/elixir/lib/exception.ex
lukaszsamson: In this case Macro.prewalk invokes callback with `{:%{}, meta, [__struct__: Range, first: first, last: last, step: step]}` so it's unnecessary

lib/elixir/lib/enum.ex
josevalim: This one is failing because the second argument you are passing is invalid, not because of the range.

lib/elixir/lib/enum.ex
josevalim: You are correct, here the range is the second argument, so we need to address it!

lukaszsamson:
@josevalim I reverted not needed changes and added test coverage

lib/elixir/lib/calendar/date_range.ex
```diff
@@ -48,24 +50,43 @@ defmodule Date.Range do
...
     end
 
-    def member?(_, _) do
+    def member?(%Date.Range{step: _}, _) do
       {:ok, false}
     end
 
+    # TODO: Remove me on v2.0
+    def member?(
+          %{__struct__: Date.Range, first_in_iso_days: first_days, last_in_iso_days: last_days} =
```
v0idpwn, line 59:
```suggestion
          %Date.Range{first_in_iso_days: first_days, last_in_iso_days: last_days} =
```

lib/elixir/lib/calendar/date_range.ex
```diff
@@ -48,24 +50,43 @@ defmodule Date.Range do
...
     end
 
+    # TODO: Remove me on v2.0
+    def slice(
+          %{__struct__: Date.Range, first_in_iso_days: first_days, last_in_iso_days: last_days} =
+            date_range
+        ) do
+      step = if first_days <= last_days, do: 1, else: -1
+      slice(Map.put(date_range, :step, step))
+    end
```
v0idpwn, line 88:
```suggestion
    def slice(
          %Date.Range{first_in_iso_days: first_days, last_in_iso_days: last_days} =
            date_range
        ) do
      step = if first_days <= last_days, do: 1, else: -1
      slice(Map.put(date_range, :step, step))
    end
```

lib/elixir/lib/calendar/date_range.ex
```diff
@@ -77,17 +98,30 @@ defmodule Date.Range do
...
+    # TODO: Remove me on v2.0
+    def reduce(
+          %{__struct__: Date.Range, first_in_iso_days: first_days, last_in_iso_days: last_days} =
+            date_range,
+          acc,
+          fun
+        ) do
+      step = if first_days <= last_days, do: 1, else: -1
+      reduce(Map.put(date_range, :step, step), acc, fun)
+    end
```
v0idpwn, line 123:
```suggestion
          %Date.Range{first_in_iso_days: first_days, last_in_iso_days: last_days} =
            date_range,
          acc,
          fun
        ) do
      step = if first_days <= last_days, do: 1, else: -1
      reduce(Map.put(date_range, :step, step), acc, fun)
    end
```

lib/elixir/lib/calendar/date_range.ex
```diff
@@ -136,6 +170,15 @@ defmodule Date.Range do
...
 
+    # TODO: Remove me on v2.0
+    defp size(
+           %{__struct__: Date.Range, first_in_iso_days: first_days, last_in_iso_days: last_days} =
+             date_range
+         ) do
+      step = if first_days <= last_days, do: 1, else: -1
+      size(Map.put(date_range, :step, step))
+    end
+
```
v0idpwn, line 181:
```suggestion
    # TODO: Remove me on v2.0
    defp size(
           %Date.Range{first_in_iso_days: first_days, last_in_iso_days: last_days} =
             date_range
         ) do
      step = if first_days <= last_days, do: 1, else: -1
      size(Map.put(date_range, :step, step))
    end

```

lib/elixir/lib/calendar/date_range.ex
```diff
@@ -152,16 +195,33 @@ defmodule Date.Range do
...
+    defp empty?(%Date.Range{step: _}), do: false
+
+    # TODO: Remove me on v2.0
+    defp empty?(
+           %{__struct__: Date.Range, first_in_iso_days: first_days, last_in_iso_days: last_days} =
+             date_range
+         ) do
+      step = if first_days <= last_days, do: 1, else: -1
+      empty?(Map.put(date_range, :step, step))
+    end
```
v0idpwn, line 207:
```suggestion
    # TODO: Remove me on v2.0
    defp empty?(
           %Date.Range{first_in_iso_days: first_days, last_in_iso_days: last_days} =
             date_range
         ) do
      step = if first_days <= last_days, do: 1, else: -1
      empty?(Map.put(date_range, :step, step))
    end
```

lib/elixir/lib/calendar/date_range.ex
```diff
@@ -152,16 +195,33 @@ defmodule Date.Range do
...
 
     def inspect(%Date.Range{first: first, last: last, step: step}, _) do
       "#DateRange<" <> inspect(first) <> ", " <> inspect(last) <> ", #{step}>"
     end
+
+    # TODO: Remove me on v2.0
+    def inspect(%{__struct__: Date.Range, first: first, last: last} = date_range, opts) do
+      step = if first <= last, do: 1, else: -1
+      inspect(Map.put(date_range, :step, step), opts)
+    end
```
v0idpwn, line 225:
```suggestion

    # TODO: Remove me on v2.0
    def inspect(%Date.Range{first: first, last: last} = date_range, opts) do
      step = if first <= last, do: 1, else: -1
      inspect(Map.put(date_range, :step, step), opts)
    end
```

v0idpwn (commented):
Nitpick for consistency

lib/elixir/lib/calendar/date_range.ex
josevalim: We are not pattern matching on range on purpose because those are "old versions" of the range. In theory the pattern works, because we don't check the size, but I would prefer to keep it explicit it is a subset.

josevalim:

lib/elixir/lib/calendar/date_range.ex
v0idpwn: TIL, sorry for noise :)

Carmen675 (commented):
I DONT CODE I SONT KNOW HOW I JUST WANT WHO EVER HAS TAKEN OVER MY PHONE TO PLEASE STOP PUTTING MALWARE ON MY IPHONE AND SLOWLY TURNING MY PHONE INTO A ANDROID THATS GOING TO BE THEIRS. I HAVE HAD NO IDEA SOMEONE WAS DOING THIS TRACKING MY EVERY MOVE, RECORDING MY PHONE CALLS WHEN I USED TO GET THEM NOW NO CALLS OR TEXTS HARDLY ANYTHING AT ALL

PR #11119, opened 2021-07-15
Change exit status for test failures to 2

J3RN:
From [the mailing list conversation](https://groups.google.com/g/elixir-lang-core/c/KOrTrFvXxiM).

It is nice to be able to differentiate between different reasons for failure of a test suite, specifically whether the tests failed or `--warnings-as-errors` detected warnings. Currently, both failure reasons exit with code 1, but this change modifies the exit code of a `--warnings-as-errors` failure to 2.

I looked for a `/lib/mix/test/mix/compilers/test_test.exs` to modify, but it looks like there are no tests for this file. Let me know if you want me to add tests :grin:

### DIFF
--- a/lib/ex_unit/lib/ex_unit.ex
+++ b/lib/ex_unit/lib/ex_unit.ex
@@ -319,7 +319,6 @@ defmodule ExUnit do
     |> put_seed()
     |> put_slowest()
     |> put_max_cases()
-    |> put_exit_status()
   end
 
   @doc """
@@ -459,10 +458,6 @@ defmodule ExUnit do
     else
       opts
     end
-  end
-
-  defp put_exit_status(opts) do
-    Keyword.put_new(opts, :exit_status, 2)
   end
 
   defp max_cases(opts) do
--- a/lib/ex_unit/lib/ex_unit.ex
+++ b/lib/ex_unit/lib/ex_unit.ex
@@ -206,7 +206,7 @@ defmodule ExUnit do
           %{failures: failures} = ExUnit.Runner.run(options, time)
 
           if failures > 0 do
-            System.at_exit(fn _ -> exit({:shutdown, Keyword.fetch!(options, :exit_code)}) end)
+            System.at_exit(fn _ -> exit({:shutdown, Keyword.fetch!(options, :exit_status)}) end)
           end
 
         _ ->
@@ -245,8 +245,8 @@ defmodule ExUnit do
     * `:exclude` - specifies which tests are run by skipping tests that match the
       filter;
 
-    * `:exit_code` - specifies an alternate exit code to use when the test suite
-      fails. Defaults to 2;
+    * `:exit_status` - specifies an alternate exit status to use when the test
+      suite fails. Defaults to 2;
 
     * `:failures_manifest_file` - specifies a path to the file used to store failures
       between runs;
@@ -319,7 +319,7 @@ defmodule ExUnit do
     |> put_seed()
     |> put_slowest()
     |> put_max_cases()
-    |> put_exit_code()
+    |> put_exit_status()
   end
 
   @doc """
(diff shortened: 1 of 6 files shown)


josevalim:
Thanks @J3RN. After looking at the code, this is a bit trickier because it changes both compilation failures and warnings as errors to 2. Is this what we want? Everywhere else in Elixir compilation failures have status of 1, so maybe we should be changing the ExUnit exit_status to 2 and make that configurable via an option?

WLSF:
Perhaps it would be nice to check if the flag `--warnings-as-errors` exists in `opts` and if the `{:error, _, _}` happens exactly because warnings were identified in the test file.

josevalim:
@WLSF in my opinion the same issue still exists. Everywhere else the compiler and warnings as errors use status code of 1. So I think it is probably best to change the test suite failure to 2.

@J3RN can you please move the changes to `ExUnit.start/1` and make sure the `exit_status` is configurable, but with the default of 2?

J3RN:
@josevalim Yes, the inconsistency with compilation exit codes under different circumstances did occur to me while I was writing this. Since test failures, by contrast, only have one context, it does seem to make more sense to change that exit code than the one for compilation. I'll make this change! :+1:

J3RN:
Since `mix test` and `ExUnit.start/1` have separate exit code handling, I had to make the change in two places. Anyhow, I just pushed those changes. Let me know what you think! :grin:

lib/ex_unit/lib/ex_unit.ex
```diff
@@ -245,6 +245,9 @@ defmodule ExUnit do
     * `:exclude` - specifies which tests are run by skipping tests that match the
       filter;
 
+    * `:exit_code` - specifies an alternate exit code to use when the test suite
+      fails;
```
josevalim, line 249:
```suggestion
      fails. Defaults to 2;
```

J3RN: Good catch!

J3RN:
It looks like Windows has some kind of incompatibility with exit codes. I'll tag the failing test `:unix`.

josevalim:
I am wondering if we should call it exit_code or exit_status. A new feature in Mix called them `:exit_code` but ports call them `:exit_status` and I think we should call them `:exit_status` for consistency. /cc @hauleth

Can you please rename accordingly?

hauleth:
@josevalim sure.

josevalim:
To be clear, @J3RN, can you please rename? @hauleth you don't need to do anything for now. :)

J3RN:
@josevalim Done! :grin:

lib/mix/lib/mix/tasks/test.ex
```diff
@@ -756,6 +764,16 @@ defmodule Mix.Tasks.Test do
...
   end
 
+  defp exit_status_opts(opts) do
+    case Keyword.fetch(opts, :exit_status) do
+      {:ok, _status} ->
+        opts
+
+      :error ->
+        Keyword.put(opts, :exit_status, 2)
+    end
```
josevalim, line 774:
```suggestion
    Keyword.put_new(opts, :exit_status, 2)
```

lib/ex_unit/lib/ex_unit.ex
```diff
@@ -457,6 +461,10 @@ defmodule ExUnit do
     end
   end
 
+  defp put_exit_status(opts) do
+    Keyword.put_new(opts, :exit_status, 2)
+  end
```
josevalim, line 466: This function is not necessary. Instead, put the default value in lib/ex_unit/mix.exs. The default values in this file are only the ones that need to be computed dynamically. :)

josevalim:
Just two additional comments and we should be ready to go!

josevalim:

PR #11121, opened 2021-07-16
Fix autocomplete on atoms with `@`

### LINKED ISSUE #11120: IEx autocomplete error with erlang modules including the `@` symbol
### Environment

```
elixir --version
Erlang/OTP 24 [erts-12.0.3] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1]

```
* Operating system: macOS

### Current behavior

[This sample repo](https://github.com/drewolson/iex_autocomplete_issue) is a minimal reproduction of this bug. Run the following:

```
mix compile
iex -S mix
```

Once in an iex session, attempting to perform autocomplete on the `example@foo` erlang module results in an error:

```
iex(1)> :example@foo.foo()
"hello"
iex(2)> :example@foo.<TAB>*** ERROR: Shell process terminated! (^G to start new job) ***

09:57:44.460 [error] Process #PID<0.65.0> raised an exception
** (FunctionClauseError) no function clause matching in IEx.Autocomplete.expand_dot_path/2
    (iex 1.12.2) lib/iex/autocomplete.ex:154: IEx.Autocomplete.expand_dot_path({:module_attribute, 'foo'}, #PID<0.82.0>)
    (iex 1.12.2) lib/iex/autocomplete.ex:146: IEx.Autocomplete.expand_dot/3
    (kernel 8.0.1) group.erl:596: :group.get_line1/5
    (kernel 8.0.1) group.erl:471: :group.get_chars_loop/9
    (kernel 8.0.1) group.erl:181: :group.io_request/6
    (kernel 8.0.1) group.erl:117: :group.server_loop/3
```

### Expected behavior

Autocomplete should work.

v0idpwn:
Closes #11120.
Code.cursor_context/2 was misidentifying some atoms as module attributes.

### DIFF
--- a/lib/elixir/lib/code.ex
+++ b/lib/elixir/lib/code.ex
@@ -415,7 +415,7 @@ defmodule Code do
       {kind, _, '..' ++ _, acc} -> alias_or_local_or_var(kind, acc)
       # Module attributes
       {:alias, _, '@' ++ _, _} -> :none
-      {:identifier, _, '@' ++ _, acc} -> {:module_attribute, acc}
+      {:identifier, _, '@', acc} -> {:module_attribute, acc}
       # Everything else
       {:alias, _, '.' ++ rest, acc} -> nested_alias(rest, acc)
       {:identifier, _, '.' ++ rest, acc} -> dot(rest, acc)
@@ -452,6 +452,12 @@ defmodule Code do
 
   defp check_identifier([h | _], _acc) when h in @non_identifier, do: :none
   defp check_identifier(rest, acc), do: rest_identifier(rest, acc)
+
+  # @ needs a special clause because it is an operator only if at word start,
+  # but can be used at middle of atoms
+  defp rest_identifier([?@ = h, s | rest], acc) do
+    rest_identifier([s | rest], [h | acc])
+  end
 
   defp rest_identifier([h | rest], acc) when h not in @non_identifier do
     rest_identifier(rest, [h | acc])
--- a/lib/elixir/test/elixir/code_test.exs
+++ b/lib/elixir/test/elixir/code_test.exs
@@ -58,6 +58,7 @@ defmodule CodeTest do
       assert Code.cursor_context(":hello.wor") == {:dot, {:unquoted_atom, 'hello'}, 'wor'}
       assert Code.cursor_context(":hell@o.wor") == {:dot, {:unquoted_atom, 'hell@o'}, 'wor'}
       assert Code.cursor_context(":he@ll@o.wor") == {:dot, {:unquoted_atom, 'he@ll@o'}, 'wor'}
+      assert Code.cursor_context(":hell@@o.wor") == {:dot, {:unquoted_atom, 'hell@@o'}, 'wor'}
       assert Code.cursor_context("@hello.wor") == {:dot, {:module_attribute, 'hello'}, 'wor'}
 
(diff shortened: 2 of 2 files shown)


lib/elixir/lib/code.ex
```diff
@@ -424,6 +424,14 @@ defmodule Code do
     end
   end
 
+  defp module_attribute_or_atom(acc, rest) do
+    case Enum.reverse(rest) do
+      '' -> {:module_attribute, acc}
+      ':' ++ atom -> {:unquoted_atom, atom ++ '@' ++ acc}
+      _ -> :none
+    end
```
v0idpwn, line 432: Hey, I'm pretty sure that this function isn't implemented in the best way but I was trying to get it to work. How could I improve this fix?

lib/elixir/lib/code.ex
```diff
@@ -424,6 +424,14 @@ defmodule Code do
     end
   end
 
+  defp module_attribute_or_atom(acc, rest) do
+    case Enum.reverse(rest) do
+      '' -> {:module_attribute, acc}
+      ':' ++ atom -> {:unquoted_atom, atom ++ '@' ++ acc}
+      _ -> :none
```
josevalim, line 431: What happens when you do: `foo :bar@baz` or `foo(:bar@baz)` or `foo :bar, @baz`? I am afraid in those cases simply looking at the rest won't be enough. I think what you probably need to do is to call `rest_identifier` again. If it returns `{_, _, '@' ++ _, _}` or `{_, _, ':' ++ _, _}`, then it is atom, otherwise it is a module attribute.

josevalim: Actually, you should call rest_identifier. If you get `{_, _, '@' ++ _, _}`, then call it again until you find a `{_, _, ':' ++ _, _}`. Otherwise it is a module attribute.

v0idpwn:
Current master:

```elixir
iex(1)> foo@bs.*** ERROR: Shell process terminated! (^G to start new job) ***

21:18:36.225 [error] Process #PID<0.65.0> raised an exception
** (FunctionClauseError) no function clause matching in IEx.Autocomplete.expand_dot_path/2
    (iex 1.12.1) lib/iex/autocomplete.ex:154: IEx.Autocomplete.expand_dot_path({:module_attribute, 'bs'}, #PID<0.77.0>)
    (iex 1.12.1) lib/iex/autocomplete.ex:146: IEx.Autocomplete.expand_dot/3
    (kernel 8.0) group.erl:596: :group.get_line1/5
    (kernel 8.0) group.erl:471: :group.get_chars_loop/9
    (kernel 8.0) group.erl:181: :group.io_request/6
    (kernel 8.0) group.erl:117: :group.server_loop/3

```

The issue reported on #11120 isn't related to module attributes, but to atoms that have `@` on them. `Purerl` relies on this character for atom/module namespacing, just as Elixir relies on "Elixir."

lib/elixir/lib/code.ex
v0idpwn: you're correct about these cases, I'll try another iteration to fix them

v0idpwn:
Hey, I managed to solve the errors you pointed out, but found another corner case on which my code doesn't work.

Pushed test for reference. I think a bigger change is likely required.

josevalim:
I handled the additional case and I pushed to master. Thank you!

PR #11125, opened 2021-07-18
mix deps: Add :subdir option to git deps

wojtekmach:
This is useful when the Mix project is not at the root of the Git
checkout but you still want to do a full checkout because you want to
refer to some other files in the repo outside of your Mix project.

One particular use case is having a monorepo with different Mix projects
and path dependencies between them:

    foo/
      mix.exs
    bar/
      mix.exs # with: {:foo, path: "../foo"}

sparse checkout would not work here.

### DIFF
--- a/lib/mix/lib/mix/scm/git.ex
+++ b/lib/mix/lib/mix/scm/git.ex
@@ -35,17 +35,18 @@ defmodule Mix.SCM.Git do
       opts
       |> Keyword.put(:checkout, opts[:dest])
       |> sparse_opts()
+      |> subdir_opts()
 
     cond do
       gh = opts[:github] ->
         opts
         |> Keyword.delete(:github)
         |> Keyword.put(:git, "https://github.com/#{gh}.git")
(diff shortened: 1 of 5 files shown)


wojtekmach:
If this sounds like a good idea, I'll work on adding tests. :)

josevalim:
Looks good to me but let's call it `:subdir`. :)

josevalim:
Also please add the `@impl true` stuff in another PR.

lib/mix/lib/mix/scm/git.ex
```diff
@@ -97,6 +98,9 @@ defmodule Mix.SCM.Git do
     File.rm_rf!(path)
     File.mkdir_p!(path)
 
+    # for :subdir
+    File.mkdir_p!(opts[:dest])
```
josevalim, line 102:
```suggestion
    File.mkdir_p!(opts[:dest])
```

josevalim:

PR #11126, opened 2021-07-18
Add `@impl true` to Mix.SCM.Git and Mix.SCM.Path

wojtekmach:


josevalim:

PR #11128, opened 2021-07-20
Preserve heredoc indentation metadata

doorgan:
Indentation metadata is preserved for heredoc sigils, but it's not considered for heredoc strings or charlists. It's reasonable since they're literals anyways, but using the `literal_encoder` option knowing the indentation of the node comes in handy.


josevalim:

PR #11129, opened 2021-07-21
Add `did you mean` to error for unfound CLI option

maartenvanvliet:
I typed `mix test --fauled` repeatedly and found the error could
be improved. This change adds the `did you mean` suggestion with the
most likely match to the error message.

The `0.8` cutoff is pulled from the exceptions.ex file.

### DIFF
--- a/lib/elixir/lib/option_parser.ex
+++ b/lib/elixir/lib/option_parser.ex
@@ -828,7 +828,15 @@ defmodule OptionParser do
     if type = get_type(option, opts, types) do
       "#{option} : Missing argument of type #{type}"
     else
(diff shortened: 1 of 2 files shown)


josevalim:

lib/elixir/lib/option_parser.ex
```diff
@@ -847,4 +855,16 @@ defmodule OptionParser do
       types[key]
     end
   end
+
+  defp did_you_mean(option, types) do
+    key = String.trim_leading(option, "-")
```
josevalim, line 860: I just realized that we also need to convert "-" into "_". Can you please send another PR for that? Thanks!

PR #11132, opened 2021-07-21
Update mix release code snippets in docs

wojtekmach:


lib/mix/lib/mix/tasks/release.ex
```diff
@@ -560,9 +584,11 @@ defmodule Mix.Tasks.Release do
 
   On Windows, your `env.bat` would look like this:
 
-      IF NOT %RELEASE_COMMAND:start=%==%RELEASE_COMMAND% (
-        set ELIXIR_ERL_OPTIONS="-kernel inet_dist_listen_min %BEAM_PORT% inet_dist_listen_max %BEAM_PORT%"
-      )
+  ```text
```
josevalim, line 587:
```suggestion
  ```bash
```

josevalim (commented):
Can you please keep the text ones indented? It is cleaner :)

wojtekmach:
If we do an intended code fence we have:

IMG_5795

IMG_3145

This actually is consistent with Cmark:

image

(https://spec.commonmark.org/0.12/#example-80)

That is we can indent code fences by up to 3 spaces and it is as if there's no indent. 4 spaces indent produces a code block and the code fence is part of the output.

3 space indent is not handled by Earmark quite correctly:

image

image

there's an extra left margin and it doesn't syntax highlight the language.

I'm happy to close this until there's perhaps support for this in Earmark.

josevalim:

PR #11137, opened 2021-07-23
Overhaul of documentation for `Keyword` module

azizk:
So at first I saw that the word "duplicated" was used throughout the doc, which isn't quite correct. Better to use the adjective "duplicate".
I noticed many other issues and decided to give it a thorough read. Hope you like the changes.

### DIFF
--- a/lib/elixir/lib/keyword.ex
+++ b/lib/elixir/lib/keyword.ex
@@ -553,7 +553,7 @@ defmodule Keyword do
   @doc """
   Deletes the first entry in the keyword list under a specific `key`.
 
-  If the `key` does not exist, the keyword list is returned unchanged.
+  If the `key` does not exist, it returns the keyword list unchanged.
 
   ## Examples
 
--- a/lib/elixir/lib/keyword.ex
+++ b/lib/elixir/lib/keyword.ex
@@ -477,7 +477,7 @@ defmodule Keyword do
   catch
     element ->
       raise ArgumentError,
-            "expected a keyword list, but an entry in the list is not a two-element tuple" <>
+            "expected a keyword list, but an entry in the list is not a two-element tuple " <>
               "with an atom as its first element, got: #{inspect(element)}"
   end
 
--- a/lib/elixir/lib/keyword.ex
+++ b/lib/elixir/lib/keyword.ex
@@ -382,7 +382,7 @@ defmodule Keyword do
   end
 
   defp get_and_update!([], key, _fun, acc) when is_atom(key) do
-    raise(KeyError, key: key, term: acc)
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/keyword.ex
```diff
@@ -13,45 +13,44 @@ defmodule Keyword do
...
 
       [exit_on_close: true, active: :once, packet_size: 1024]
 
-  The two syntaxes are completely equivalent. Like atoms, keyword
-  lists keys must be composed of Unicode characters such as letters,
-  numbers, underscore, and `@`. If the keyword has a character that
-  does not belong to the category above, such as spaces, you can wrap
-  it in quotes:
+  The two syntaxes are slightly different but as values they are
+  completely equivalent.
```
josevalim, line 21:
```suggestion
  The two syntaxes build the exact same value.
```

lib/elixir/lib/keyword.ex
```diff
@@ -548,9 +552,9 @@ defmodule Keyword do
   defp delete_key([], _key), do: []
 
   @doc """
-  Deletes the first entry in the keyword list for a specific `key`.
+  Deletes the first entry in the keyword list under a specific `key`.
 
-  If the `key` does not exist, returns the keyword list unchanged.
+  If the `key` does not exist, the keyword list is returned unchanged.
```
josevalim, line 557: We should avoid passive voice:

```suggestion
  If the `key` does not exist, it returns the keyword list unchanged.
```

josevalim:
Looks great, two comments added.

josevalim:
Also please run `bin/elixir bin/mix format` to make sure the file is formatted.

azizk:
Regarding the passive voice, it was being used in some places before I made my changes. Would you like me to correct all sentences with a passive voice?

josevalim:
I don't think we need to go and change all of them but we should probably avoid adding more. :)

lib/elixir/lib/keyword.ex
```diff
@@ -462,26 +467,24 @@ defmodule Keyword do
...
+      fn
+        {key, _} when is_atom(key) -> key
+        element -> throw(element)
+      end,
+      keywords
+    )
+  catch
+    element ->
+      raise ArgumentError,
+            "expected a keyword list, but an entry in the list is not a two-element tuple" <>
```
josevalim, line 480:
```suggestion
            "expected a keyword list, but an entry in the list is not a two-element tuple " <>
```

azizk:
It's fine. Let me quickly change some passive voice sentences. Not a big deal. :)

josevalim:

PR #11138, opened 2021-07-23
Keyword: change sentences from passive to active voice.

azizk:
Was going to push it on the previous PR, but you're too fast :stuck_out_tongue_closed_eyes:. Wasn't much work, so I just made the changes. It does read better overall I think. :+1:


josevalim:

PR #11139, opened 2021-07-24
Add `with_log` and `with_io`

marcandre:
I added very minimal testing for these, but they are indirectly tested with the existing tests as `capture_log` and `capture_io` rely entirely on them. Let me know if I should add more tests.

### DIFF
--- a/lib/ex_unit/lib/ex_unit/capture_io.ex
+++ b/lib/ex_unit/lib/ex_unit/capture_io.ex
@@ -12,6 +12,14 @@ defmodule ExUnit.CaptureIO do
         test "example" do
           assert capture_io(fn -> IO.puts("a") end) == "a\n"
         end
+
+        test "another example" do
+          assert with_io(fn ->
+            IO.puts("a")
+            IO.puts("b")
+            2 + 2
+          end) == {4, "a\nb\n"}
+        end
       end
 
   """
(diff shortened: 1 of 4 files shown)


lib/ex_unit/lib/ex_unit/capture_io.ex
```diff
@@ -88,55 +96,96 @@ defmodule ExUnit.CaptureIO do
...
 
   @spec capture_io(atom(), keyword(), (() -> any())) :: String.t()
   def capture_io(device, options, fun)
       when is_atom(device) and is_list(options) and is_function(fun, 0) do
-    do_capture_io(map_dev(device), options, fun)
+    {_result, capture} = with_io(device, options, fun)
+    capture
+  end
+
+  @doc """
```
josevalim, line 140:
```suggestion
  @doc ~S"""
```

josevalim: So we don't interpret escape codes.

fertapric:
What do you think about renaming the functions to `CaptureLog.capturing_log(...)` or `CaptureLog.with_captured_log(...)`? 🙂

josevalim:
I don't think we should rename the existing functions unless we have a very good reason to. I am personally happy with `capture_io` and `with_io`. Can you find a good pairing without renaming? :)

astery:
May I propose another naming for the function?

We have similar functions `&CaptureIO.capture_io/1` and `&CaptureLog.capture_log/1`, and I think a function doing the same, but returning a different return value, should have a name that resembles that, like:

```
capture_io_and_res
capture_log_and_res
```

or

```
capture_io_with
capture_log_with
```

fertapric:
I meant renaming the function `with_io` to `capturing_io` or `with_captured_io` (and the same for the log counterpart)

astery:
> I meant renaming the function `with_io` to `capturing_io` or `with_captured_io` (and the same for the log counterpart)

Now I get it. I like `with_captured_io/with_captured_log` variant, as it resembles the original, and tells that where will be a result with captured io/log.

josevalim:
I don't like that one version presente tense (capture) and the other is the past particle (captured). I was hoping `with_io` would be bypass that. Can we articulate what is bad about `with_io`/`with_log`?

astery:
Oh, and now I understand the original `with_io/with_log` name wasn't that far, but it just not "clicked" in my mind that this is the same `capture_io` function, but have a different return value. 

And now I actually think `with_io/with_log` is good variants, they are short, with clear intent. If they were there from the start you would never thought to add `_captured_`. 

I don't know would I trade short and clear `with_io`, for slightly longer `with_captured_io` but resembling `capture_io` now. Sorry, I didn't have a strong position myself, but got into the discussion.

fertapric:
In my head,  those `with_*` functions usually mean that something is enabled/configured during the execution of the given block. ActiveRecord's [`with_connection`](https://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/ConnectionPool.html#method-i-with_connection) or [`with_lock`](https://apidock.com/rails/v5.2.3/ActiveRecord/Locking/Pessimistic/with_lock) are some examples of that. `with_io` would then mean "with IO (enabled)", which is not really what it does and doesn't make much sense.
 
After reading the comments, now I see that the meaning might be "result with IO", could that be it?

josevalim:
Yes, exactly. What if we change the docs to this:

```
@doc """
Executes the given function and returns its result with
the captured io.
"""
```

Would that help?

fertapric:
Yeah, it would help a little.

However, I would go with `with_captured_io` because it works both ways: it might mean "execute this with IO captured" or "return result with the captured IO".

lib/ex_unit/lib/ex_unit/capture_io.ex
```diff
@@ -88,55 +96,96 @@ defmodule ExUnit.CaptureIO do
...
   def capture_io(device, options, fun)
       when is_atom(device) and is_list(options) and is_function(fun, 0) do
-    do_capture_io(map_dev(device), options, fun)
+    {_result, capture} = with_io(device, options, fun)
+    capture
+  end
+
+  @doc ~S"""
+  Same functionality as `capture_io/2`, but returns a tuple
+  with the result of the evaluation and the captured output.
```
josevalim, line 142:
```suggestion
  Invokes the given `fun` and returns a tuple with its result
  and the captured output.
```

lib/ex_unit/lib/ex_unit/capture_log.ex
```diff
@@ -60,9 +65,22 @@ defmodule ExUnit.CaptureLog do
...
   """
   @spec capture_log(keyword, (() -> any)) :: String.t()
   def capture_log(opts \\ [], fun) do
+    {_, log} = with_log(opts, fun)
+    log
+  end
+
+  @doc """
+  Same functionality as `capture_log/2`, but returns a tuple
+  with the result of the evaluation and the captured output.
```
josevalim, line 80:
```suggestion
  Invokes the given `fun` and returns a tuple with its result
  and the captured logs.

  It has the same API and options as `capture_log/2`.
```

josevalim:
Between with_io and with_captured_io I definitely prefer the former, it is more concise and doesn't add potential tense confusion, so I will go ahead with this. If someone has suggestions, feel free to chime in, we can still rename things before v1.13 (expected oct/nov).

lib/ex_unit/lib/ex_unit/capture_io.ex
```diff
@@ -88,55 +96,98 @@ defmodule ExUnit.CaptureIO do
...
       when is_atom(device) and is_list(options) and is_function(fun, 0) do
-    do_capture_io(map_dev(device), options, fun)
+    {_result, capture} = with_io(device, options, fun)
+    capture
+  end
+
+  @doc ~S"""
+  Invokes the given `fun` and returns a tuple with its result
+  and the captured output.
+
```
josevalim, line 143:
```suggestion

```

josevalim:

marcandre:
Great, thanks!

PR #11143, opened 2021-07-26
Add Code.Fragment.surround_context/3

josevalim:


lib/elixir/lib/code/fragment.ex
```diff
@@ -286,72 +312,290 @@ defmodule Code.Fragment do
...
+        dot(tl(rest), dot_count + 1, acc)
+
+      true ->
+        {{:operator, acc}, count}
     end
   end
 
-  defp operator(rest, acc, _call_op?) do
+  defp operator(rest, count, acc, _call_op?) do
     case :elixir_tokenizer.tokenize(acc, 1, 1, []) do
```
josevalim, line 366: Note `:elixir_tokenizer.tokenize/4` returns `{:ok, tokens}` on Elixir v1.12. So backporting this code requires wrapping this call (which is a private API anyway).

josevalim:

PR #11149, opened 2021-07-27
feat: add keyword validate

polvalente:
This PR implements `Keyword.validate/2` and `Keyword.validate!/2` as discussed in the mailing list.

The original implementation for `validate!/2` can be found in [Nx.Defn.Kernel](https://github.com/elixir-nx/nx/blob/c8353de695a3d70dc6d518e0ab2ec832faa7df68/nx/lib/nx/defn/kernel.ex#L741)

### DIFF
--- a/lib/elixir/lib/keyword.ex
+++ b/lib/elixir/lib/keyword.ex
@@ -315,7 +315,7 @@ defmodule Keyword do
       {:error, invalid_keys} ->
         keys =
           for value <- values,
-              do: if(is_atom(value), do: value, else: Kernel.elem(value, 0))
+              do: if(is_atom(value), do: value, else: elem(value, 0)
 
         raise ArgumentError,
               "unknown keys #{inspect(invalid_keys)} in #{inspect(keyword)}, the allowed keys are: #{inspect(keys)}"
--- a/lib/elixir/lib/keyword.ex
+++ b/lib/elixir/lib/keyword.ex
@@ -215,6 +215,7 @@ defmodule Keyword do
       iex> Keyword.validate!([three: 3], [one: 1, two: 2])
       ** (ArgumentError) unknown key :three in [three: 3], expected one of [:one, :two]
   """
+  @doc since: "1.13.0"
   def validate!(keyword, values) when is_list(keyword) and is_list(values) do
     # We use two lists to avoid reversing/concatenating
     # lists in the middle of traversals.
--- a/lib/elixir/lib/keyword.ex
+++ b/lib/elixir/lib/keyword.ex
@@ -187,6 +187,193 @@ defmodule Keyword do
   end
 
   @doc """
+  Ensures the first argument is a `keyword` with the given
+  keys and default values.
+  The second argument must be a list of atoms, specifying
+  a given key, or tuples specifying a key and a default value.
+  If any of the keys in the `keyword` is not defined on
+  `values`, it raises an error.
+
+  See also: `validate/2`
+
+  ## Examples
+      iex> Keyword.validate!([], [one: 1, two: 2]) |> Enum.sort()
+      [one: 1, two: 2]
+      iex> Keyword.validate!([two: 3], [one: 1, two: 2]) |> Enum.sort()
+      [one: 1, two: 3]
+
+  If atoms are given, they are supported as keys but do not
+  provide a default value:
+
+      iex> Keyword.validate!([], [:one, two: 2]) |> Enum.sort()
+      [two: 2]
+      iex> Keyword.validate!([one: 1], [:one, two: 2]) |> Enum.sort()
+      [one: 1, two: 2]
+
+  Passing an unknown key raises:
+
+      iex> Keyword.validate!([three: 3], [one: 1, two: 2])
+      ** (ArgumentError) unknown key :three in [three: 3], expected one of [:one, :two]
+  """
+  def validate!(keyword, values) when is_list(keyword) and is_list(values) do
+    # We use two lists to avoid reversing/concatenating
+    # lists in the middle of traversals.
+    case validate!(keyword, values, [], []) do
+      {:ok, keyword} ->
+        keyword
+
+      error ->
+        keys =
+          for value <- values do
+            if is_atom(value) do
+              value
+            else
+              elem(value, 0)
+            end
+          end
+
+        case error do
+          {:badkey, key} ->
+            raise ArgumentError,
+                  "unknown key #{inspect(key)} in #{inspect(keyword)}, " <>
+                    "expected one of #{inspect(keys)}"
+
+          :badkey ->
+            raise ArgumentError,
+                  "expected a keyword list with keys #{inspect(keys)}, got: #{inspect(keyword)}"
+        end
+    end
+  end
+
+  @doc """
+  Ensures the first argument is a `keyword` with the given
+  keys and default values.
+  The second argument must be a list of atoms, specifying
+  a given key, or tuples specifying a key and a default value.
+
+  Possible return values:
+  * `{:ok, keyword}`: Success case. Returns the validated keyword with default values applied.
+  * `{:error, :not_a_keyword}`: Indicates that the validated value is not a keyword.
+  * `{:error, :invalid_expected_values}`: Indicates that the expected values list is invalid.
+  * `{:error, {:invalid_key, key}}`: Indicates that a given key is not expected.
+
+  See also: `validate!/2`
+
+  ## Examples
+      iex> {:ok, result} = Keyword.validate([], [one: 1, two: 2])
+      iex> Enum.sort(result)
+      [one: 1, two: 2]
+
+      iex> {:ok, result} = Keyword.validate([two: 3], [one: 1, two: 2])
+      iex> Enum.sort(result)
+      [one: 1, two: 3]
+
+  If atoms are given, they are supported as keys but do not
+  provide a default value:
+
+      iex> {:ok, result} = Keyword.validate([], [:one, two: 2])
+      iex> Enum.sort(result)
+      [two: 2]
+
+      iex> {:ok, result} = Keyword.validate([one: 1], [:one, two: 2])
+      iex> Enum.sort(result)
+      [one: 1, two: 2]
+
+  Passing an unknown key returns an error:
+
+      iex> Keyword.validate([three: 3], [one: 1, two: 2])
+      {:error, {:invalid_key, :three}}
+
+  Passing an invalid value in either argument also returns errors
+
+      iex> Keyword.validate(%{three: 3}, [one: 1, two: 2])
+      {:error, :not_a_keyword}
+      iex> Keyword.validate([:three], [one: 1, two: 2])
+      {:error, :not_a_keyword}
+      iex> Keyword.validate([three: 3], [{"one", 1}, :two])
+      {:error, :invalid_expected_values}
+  """
+  @spec validate(keyword :: keyword(), values :: [atom() | {atom(), term()}]) ::
+          {:ok, keyword}
+          | {:error,
+             :not_a_keyword
+             | {:invalid_key, term()}
+             | :invalid_expected_values}
+  def validate(keyword, values) do
+    with :ok <- is_atom_or_tuple_list(values),
+         {:ok, kw} <- validate!(keyword, values, [], []) do
+      {:ok, kw}
+    else
+      :badkey ->
+        {:error, :not_a_keyword}
+
+      {:badkey, key} ->
+        {:error, {:invalid_key, key}}
+
+      {:error, :not_atom_or_tuple_list} ->
+        {:error, :invalid_expected_values}
+    end
+  end
+
+  defp is_atom_or_tuple_list(values) when is_list(values) do
+    validation_fn = fn
+      v when is_atom(v) -> true
+      {k, _} when is_atom(k) -> true
+      _ -> false
+    end
+
+    if Enum.all?(values, validation_fn) do
+      :ok
+    else
+      {:error, :not_atom_or_tuple_list}
+    end
+  end
+
+  defp validate!([{key, _} = pair | keyword], values1, values2, acc) when is_atom(key) do
+    case find_key!(key, values1, values2) do
+      {values1, values2} ->
+        validate!(keyword, values1, values2, [pair | acc])
+
+      :error ->
+        case find_key!(key, values2, values1) do
+          {values1, values2} ->
+            validate!(keyword, values1, values2, [pair | acc])
+
+          :error ->
+            {:badkey, key}
+        end
+    end
+  end
+
+  defp validate!([], values1, values2, acc) do
+    {:ok, move_pairs!(values1, move_pairs!(values2, acc))}
+  end
+
+  defp validate!(_keyword, _values1, _values2, _acc) do
+    :badkey
+  end
+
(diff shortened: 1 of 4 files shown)


lib/elixir/lib/keyword.ex
```diff
@@ -186,6 +186,193 @@ defmodule Keyword do
     :lists.foldl(fun, [], Enum.reverse(pairs))
   end
 
+  @doc """
+  Ensures the first argument is a `keyword` with the given
+  keys and default values.
+  The second argument must be a list of atoms, specifying
+  a given key, or tuples specifying a key and a default value.
+  If any of the keys in the `keyword` is not defined on
+  `values`, it raises an error.
```
polvalente, line 195: This doc is almost the same as the one for `validate/2`. Is there a best-practices approach to avoid being this repetitive, or is the way this was done fine?

lib/elixir/lib/keyword.ex
```diff
@@ -186,6 +186,193 @@ defmodule Keyword do
...
+  @doc """
+  Ensures the first argument is a `keyword` with the given
+  keys and default values.
+  The second argument must be a list of atoms, specifying
+  a given key, or tuples specifying a key and a default value.
+
+  Possible return values:
+  * `{:ok, keyword}`: Success case. Returns the validated keyword with default values applied.
+  * `{:error, :not_a_keyword}`: Indicates that the validated value is not a keyword.
+  * `{:error, :invalid_expected_values}`: Indicates that the expected values list is invalid.
```
polvalente, line 257: I'm not happy with some of these error messages, especially "invalid_expected_values". Suggestions are welcome!

lib/elixir/lib/keyword.ex
```diff
@@ -186,6 +186,193 @@ defmodule Keyword do
...
+      iex> Keyword.validate!([], [:one, two: 2]) |> Enum.sort()
+      [two: 2]
+      iex> Keyword.validate!([one: 1], [:one, two: 2]) |> Enum.sort()
+      [one: 1, two: 2]
+
+  Passing an unknown key raises:
+
+      iex> Keyword.validate!([three: 3], [one: 1, two: 2])
+      ** (ArgumentError) unknown key :three in [three: 3], expected one of [:one, :two]
+  """
```
wojtekmach, line 217:
```suggestion
  """
  @doc since: "1.13.0"
```

lib/elixir/lib/keyword.ex
wojtekmach: perhaps something like this?

```
  @doc """
  Similar to `validate/2` but returns the keyword or raises an error.

  ## Examples

      iex> Keyword.validate!([two: 3], [one: 1, two: 2])
      [one: 1, two: 3]

      iex> Keyword.validate!([three: 3], [one: 1, two: 2])
      ** (ArgumentError) unknown key :three in [three: 3], expected one of [:one, :two]

  """
```

btw,  I'd define the `validate!/2` function _after_ `validate/2` given they'd show up in docs in that order

lib/elixir/lib/keyword.ex
wojtekmach: perhaps something like this?

```
  @doc """
  Similar to `validate/2` but returns the keyword or raises an error.

  ## Examples

      iex> Keyword.validate!([two: 3], [one: 1, two: 2])
      [one: 1, two: 3]

      iex> Keyword.validate!([three: 3], [one: 1, two: 2])
      ** (ArgumentError) unknown key :three in [three: 3], expected one of [:one, :two]

  """
```

btw,  I'd define the `validate!/2` function _after_ `validate/2` given they'd show up in docs in that order

lib/elixir/lib/keyword.ex
wojtekmach: for these particular cases, instead of returning an error tuple I'd raise. This would be consistent with e.g. Keyword.fetch:

```
iex(1)> Keyword.fetch!("not a keyword", :key)
** (FunctionClauseError) no function clause matching in Keyword.fetch!/2

    The following arguments were given to Keyword.fetch!/2:

        # 1
        "not a keyword"

        # 2
        :key

        def fetch!(keywords, key) when is_list(keywords) and is_atom(key)

    (elixir 1.13.0-dev) lib/keyword.ex:417: Keyword.fetch!/2
```

```
iex(2)> Keyword.fetch!([], "bad key")
** (FunctionClauseError) no function clause matching in Keyword.fetch!/2

    The following arguments were given to Keyword.fetch!/2:

        # 1
        []

        # 2
        "bad key"

        def fetch!(keywords, key) when is_list(keywords) and is_atom(key)

    (elixir 1.13.0-dev) lib/keyword.ex:417: Keyword.fetch!/2
```

wdyt?

lib/elixir/lib/keyword.ex
```diff
@@ -186,6 +186,193 @@ defmodule Keyword do
...
+      {:error, :not_a_keyword}
+      iex> Keyword.validate([three: 3], [{"one", 1}, :two])
+      {:error, :invalid_expected_values}
+  """
+  @spec validate(keyword :: keyword(), values :: [atom() | {atom(), term()}]) ::
+          {:ok, keyword}
+          | {:error,
+             :not_a_keyword
+             | {:invalid_key, term()}
+             | :invalid_expected_values}
```
wojtekmach, line 301: per the ML thread:

```suggestion
          {:ok, keyword} | {:error, unknown_keys, missing_keys}
```

josevalim: To add to @wojtekmach's comment, invalid inputs (i.e. not a keyword or invalid expected values) should just raise.

josevalim: Actually, my comment on the mailing list was wrong, there is no such thing as missing keys as we don't have required keys. So this should simply be:

    {:ok, keyword} | {:error, unknown_keys :: [atom]}

wojtekmach: Oh, good call. In that case, `validate!` should definitely raise a KeyError which will give us a really good error message.

wojtekmach: scratch that, we can't raise a KeyError as we could have multiple unknown keys.

lib/elixir/lib/keyword.ex
polvalente: Makes sense. I thought that maybe accepting invalid inputs would be desirable as well, but this makes for a better interface

lib/elixir/lib/keyword.ex
polvalente: @josevalim I've updated the code with the reviews. One thing that isn't clear to me is that the suggested spec shows a list of invalid keys.

The current implementation only returns the first invalid key. Should I change this for `validate`?

lib/elixir/lib/keyword.ex
```diff
@@ -186,6 +186,182 @@ defmodule Keyword do
...
+            raise ArgumentError,
+                  "unknown key #{inspect(key)} in #{inspect(keyword)}, " <>
+                    "expected one of #{inspect(keys)}"
+
+          :badkey ->
+            raise ArgumentError,
+                  "expected a keyword list with keys #{inspect(keys)}, got: #{inspect(keyword)}"
+        end
+    end
+  end
```
wojtekmach, line 319: ideally the raising variant uses the non-raising variant under the hood, I think it will remove a fair amount of duplication. We can even drop the guards as the underlying function has them already.

```elixir
  def validate!(keyword, values) do
    case validate(keyword, values) do
      {:ok, keyword} ->
        keyword

      {:error, missing_keys} ->
        # ...
    end
  end
```

regarding error, I think we have two options, perhaps something along the lines of either:

1. `"expected a keyword list with keys #{inspect(keys)}, got: #{inspect(keyword)}"`
2. `"unknown keys #{inspect(missing_keys)} in #{inspect(keyword)}, the allowed keys are: #{inspect(keys)}"`

what do you think?

polvalente: Yeah, I'll refactor the implementation then. Makes more sense, especially since both fail exactly on the same cases.

One thing I noticed is that although probably not intended `KeyError` accepts a list as `:key`. The message in itself becomes a little bit weird though, so I'll stick to ArgumentError

```elixir
iex(1)> raise KeyError, key: [:a, :b, :c]
** (KeyError) key [:a, :b, :c] not found
```

lib/elixir/lib/keyword.ex
```diff
@@ -186,6 +186,182 @@ defmodule Keyword do
...
+  Passing an invalid value in either argument raises
+
+      iex> Keyword.validate(%{three: 3}, [one: 1, two: 2])
+      ** (ArgumentError) expected a keyword list with keys [:one, :two], got: %{three: 3}
+
+      iex> Keyword.validate([:three], [one: 1, two: 2])
+      ** (ArgumentError) expected a keyword list with keys [:one, :two], got: [:three]
+
+      iex> Keyword.validate([three: 3], [:three, 3, :two])
+      ** (ArgumentError) expected the second argument to be a list of atoms or tuples, got item: 3
```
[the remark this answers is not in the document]
polvalente: Ok, I'll move these to unit tests

polvalente:
@wojtekmach For some reason Github is not letting me "re-request" your review. So, I've refactored the code and it's ready for another round :)

lib/elixir/lib/keyword.ex
```diff
@@ -186,6 +186,142 @@ defmodule Keyword do
...
+  @spec validate!(keyword(), values :: [atom() | {atom(), term()}]) :: keyword()
+  def validate!(keyword, values) do
+    case validate(keyword, values) do
+      {:ok, kw} ->
+        kw
+
+      {:error, invalid_keys} ->
+        keys =
+          for value <- values,
+              do: Kernel.if(is_atom(value), do: value, else: Kernel.elem(value, 0))
```
josevalim, line 318:
```suggestion
              do: if(is_atom(value), do: value, else: elem(value, 0))
```

lib/elixir/test/elixir/keyword_test.exs
```diff
@@ -205,4 +205,24 @@ defmodule KeywordTest do
...
+                 fn ->
+                   Keyword.validate([:three], one: 1, two: 2)
+                 end
+
+    assert_raise ArgumentError,
+                 "expected the second argument to be a list of atoms or tuples, got item: 3",
+                 fn ->
+                   Keyword.validate([three: 3], [:three, 3, :two])
+                 end
+  end
```
josevalim, line 227:
```suggestion
  test "validate/2 raises on invalid arguments" do
    assert_raise ArgumentError,
                 "expected a keyword list as first argument, got invalid entry: :three",
                 fn -> Keyword.validate([:three], one: 1, two: 2) end

    assert_raise ArgumentError,
                 "expected the second argument to be a list of atoms or tuples, got: 3",
                 fn -> Keyword.validate([three: 3], [:three, 3, :two]) end
  end
```

lib/elixir/lib/keyword.ex
polvalente: copy-paste slipped :grimacing:

josevalim:
@polvalente I have reverted to the original implementation since it was optimized to avoid lists traversals and kept the rest. Next, can you also add Map.validate? I am wondering what would be the best way to implement it... :)

lib/elixir/lib/keyword.ex
```diff
@@ -186,6 +186,142 @@ defmodule Keyword do
...
+  @spec validate!(keyword(), values :: [atom() | {atom(), term()}]) :: keyword()
+  def validate!(keyword, values) do
+    case validate(keyword, values) do
+      {:ok, kw} ->
+        kw
+
+      {:error, invalid_keys} ->
+        keys =
+          for value <- values,
+              do: if(is_atom(value), do: value, else: Kernel.elem(value, 0))
```
josevalim, line 318:
```suggestion
              do: if(is_atom(value), do: value, else: elem(value, 0)
```

polvalente:
> @polvalente I have reverted to the original implementation since it was optimized to avoid lists traversals and kept the rest. Next, can you also add Map.validate? I am wondering what would be the best way to implement it... :)

@josevalim Sure, no problem!

I had changed the implementation because it was easier than refactoring yours for accumulating multiple invalid keys.

Should Map.validate come in another PR? I'm thinking that perhaps there is a way to use Keyword.validate underneath it, if the stdlib itself doesn't complain about it the way it complained about Enum. I'm not sure about the cost of converting from map to keyword back to map.

Another approach I have in mind is to look at how `@enforce_keys` is implemented for structs, since it's kind of a similar behavior, as well as looking into `:maps` to see if any functions there could help.

WDYT?

lib/elixir/lib/keyword.ex
```diff
@@ -186,6 +186,142 @@ defmodule Keyword do
...
+  keys and default values.
+
+  The second argument must be a list of atoms, specifying
+  a given key, or tuples specifying a key and a default value.
+
+
+  It returns `{:ok, keyword}` in case the keyword list only has
+  the given keys with returns the validated keyword with default
+  values applied. Otherwise it returns `{:error, invalid_keys}`
+  with invalid keys.
```
wojtekmach, line 200:
```suggestion

  If the keyword list has only the given keys, it returns
  `{:ok, keyword}` with default values applied. Otherwise it
  returns `{:error, invalid_keys}` with invalid keys.
```

lib/elixir/lib/keyword.ex
```diff
@@ -186,6 +186,142 @@ defmodule Keyword do
...
+
+      iex> {:ok, result} = Keyword.validate([], [:one, two: 2])
+      iex> Enum.sort(result)
+      [two: 2]
+
+      iex> {:ok, result} = Keyword.validate([one: 1], [:one, two: 2])
+      iex> Enum.sort(result)
+      [one: 1, two: 2]
+
+  Passing an unknown key returns an error:
```
wojtekmach, line 225:
```suggestion
  Passing unknown keys returns an error:
```

lib/elixir/lib/keyword.ex
```diff
@@ -186,6 +186,142 @@ defmodule Keyword do
...
+
+  If atoms are given, they are supported as keys but do not
+  provide a default value:
+
+      iex> Keyword.validate!([], [:one, two: 2]) |> Enum.sort()
+      [two: 2]
+      iex> Keyword.validate!([one: 1], [:one, two: 2]) |> Enum.sort()
+      [one: 1, two: 2]
+
+  Passing an unknown key raises:
```
wojtekmach, line 303:
```suggestion
  Passing unknown keys raises an error:
```

lib/elixir/lib/keyword.ex
```diff
@@ -186,6 +186,140 @@ defmodule Keyword do
...
+  @spec validate!(keyword(), values :: [atom() | {atom(), term()}]) :: keyword()
+  def validate!(keyword, values) do
+    case validate(keyword, values) do
+      {:ok, kw} ->
+        kw
+
+      {:error, invalid_keys} ->
+        keys =
+          for value <- values,
+              do: if(is_atom(value), do: value, else: elem(value, 0)
```
josevalim, line 316:
```suggestion
              do: if(is_atom(value), do: value, else: elem(value, 0))
```

josevalim:
I am glad to sit on Map.validate until someone submits it or comes up with a use case. I will be merging this once CI passes then!

josevalim:

PR #11150, opened 2021-07-27
Fix Code.Fragment.surround_context/3 range for spaced identifiers

jonatanklosko:
Before:

```ex
iex(1)> Code.Fragment.surround_context(" Enum", {1, 4})
%{begin: {1, 1}, context: {:alias, 'Enum'}, end: {1, 6}}
```

After:

```ex
iex(1)> Code.Fragment.surround_context(" Enum", {1, 4})
%{begin: {1, 2}, context: {:alias, 'Enum'}, end: {1, 6}}
```

### DIFF
--- a/lib/elixir/lib/code/fragment.ex
+++ b/lib/elixir/lib/code/fragment.ex
@@ -245,20 +245,26 @@ defmodule Code.Fragment do
       {:unquoted_atom, acc, count} ->
         {{:unquoted_atom, acc}, count}
 
-      {:alias, '.' ++ rest, acc, count} when rest == [] or hd(rest) != ?. ->
-        nested_alias(rest, count + 1, acc)
+      {:alias, rest, acc, count} ->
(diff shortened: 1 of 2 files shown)


lib/elixir/lib/code/fragment.ex
```diff
@@ -312,7 +318,6 @@ defmodule Code.Fragment do
         if ?@ in extra do
           :none
         else
-          {rest, count} = strip_spaces(rest, count)
```
jonatanklosko, line 315: An alternative solution could be to conditionally consume spaces here, like this:

```elixir
# Consume spaces only if the identifier is not finished
{rest, count} =
  case strip_spaces(rest, count) do
    {'.' ++ _ = rest, count} when rest == [] or hd(rest) != ?. -> {rest, count}
    _ -> {rest, count}
  end
```

However, this doesn't seem like the proper place for this.

josevalim:

PR #11153, opened 2021-07-28
Update record reflection docs

wojtekmach:


lib/elixir/lib/record.ex
```diff
@@ -36,9 +36,10 @@ defmodule Record do
 
   ## Reflection
 
-  A list of all records in a module, if any, can be retrieved by calling
-  reading the `@__records__` module attribute. It returns a list of maps
-  with the record kind, name, tag, and fields.
+  A list of all records in a module, if any, can be retrieved by reading the
+  `@__records__` module attribute. It returns a list of maps with the record
+  kind, name, tag, and fields. The attribute is only available at the build
+  time.
```
josevalim, line 42:
```suggestion
  kind, name, tag, and fields. The attribute is only available inside the
  module definition.
```

josevalim:

PR #11154, opened 2021-07-28
`String.split/3`: special case some additional arguments

paulswartz:
- `[parts: :infinity, trim: true]` can use the `trim_all` argument to
`:binary.split/3`
- `[parts: 2, trim: false]` can use the basic `:binary.split/2` behavior


josevalim:
Thank you! `trim_all` was added in later versions OTP versions, so let's see what CI says. :)

josevalim:

PR #11159, opened 2021-07-30
feat: add Map.validate/2 and Map.validate!/2

polvalente:
This PR is a continuation of #11149, implementing the same functionality for maps

### DIFF
--- a/lib/elixir/lib/map.ex
+++ b/lib/elixir/lib/map.ex
@@ -221,6 +221,118 @@ defmodule Map do
   end
 
   @doc """
+  Ensures the first argument is a `map` with the given
+  keys and default values.
+
+  The second argument must be a list of `t:term`s, specifying
+  a given key, or tuples specifying a key and a default value.
+
+  If the map has only the given keys, it returns
+  `{:ok, keyword}` with default values applied. Otherwise it
+  returns `{:error, invalid_keys}` with invalid keys.
+
+  See also: `validate!/2`.
+
+  ## Examples
+
+      iex> Map.validate(%{}, [{"one", 1}, two: 2])
+      {:ok, %{"one" => 1, two: 2}}
+
+      iex> Map.validate(%{two: 3}, [one: 1, two: 2])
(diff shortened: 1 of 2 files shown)


lib/elixir/lib/map.ex
```diff
@@ -220,6 +220,120 @@ defmodule Map do
...
+  @doc since: "1.13.0"
+  @spec validate(map, [{term(), term()} | term()]) :: {:ok, map()} | {:error, [atom]}
+  def validate(map, values) when is_map(map) and is_list(values) do
+    # split values into the default half and the non-default half
+    {keys_with_defaults, non_default_keys} =
+      for val <- values, reduce: {[], []} do
+        {with_defaults, non_defaults} ->
+          case val do
+            {_, _} -> {[val | with_defaults], non_defaults}
+            _ -> {with_defaults, [val | non_defaults]}
```
polvalente, line 267: Should I only accept atoms and strings as the non-default keys? I've ended up relaxing this condition because maps are more versatile than keywords in this regard.

josevalim:
Thank you @polvalente! For now we decided to sit on this until there is a clearer use case. One thing to keep in mind, with maps, is that if we allow any key, then `[foo: :bar]` is ambiguous if it means accepted key `{:foo, :bar}` or accepted key `:foo` with value of `:bar`. It may be best to enforce the second element of `validate` to be maps in case of maps. Then we would just check if any key does no longer to the given map and then merge, which should be quite efficient.

polvalente:
> Thank you @polvalente! For now we decided to sit on this until there is a clearer use case. One thing to keep in mind, with maps, is that if we allow any key, then `[foo: :bar]` is ambiguous if it means accepted key `{:foo, :bar}` or accepted key `:foo` with value of `:bar`. It may be best to enforce the second element of `validate` to be maps in case of maps. Then we would just check if any key does no longer to the given map and then merge, which should be quite efficient.

No problem at all! Better than feature creep as always :sweat_smile:. 
At least there's already this exploratory PR with some caveats to keep in mind here :)

PR #11164, opened 2021-08-03
Break Macro.Env apart

josevalim:
Previously, the internal Elixir expansion pass worked directly on
`Macro.Env`. However, this poised an issue, if we want to track
more information in the pass, we ended up exposing it on
`Macro.Env`, making it larger, and potentially slowing down
operations such as `__ENV__` serialization.

This commit refactors the expansion pass to work with two
structures, the `Macro.Env` struct and a `#elixir_ex{}` record.


josevalim:

Issue #11142, opened 2021-07-25
column metadata does not consider grapheme clusters


josevalim:
Take the following code:

    Code.string_to_quoted("'josé' = 1", columns: true)

where "josé" is written in nfd format. The column for the equal sign will be 9 but it should be 8.

This issue exists on master and doesn't depend on the OTP version.

PR #11180, opened 2021-08-14
Handle grapheme clusters in `Code.string_to_quoted`'s columns metadata

felipelincoln:
The solution calculates the diference between the lengths of raw string and nfc converted string and decrease column count by this difference. Solves #11142


josevalim:
Thank you @felipelincoln! Just one note: instead of fixing the column after the fact, instead compute the column properly the first time around. For example, you will change the catch-all clause to call `unicode_util:gc(Rest)` and then increase the column by one.

felipelincoln:
I appreciate your feedback @josevalim.
Since I am still learning how elixir is structured I will need some time to work on this so I'm closing this PR for now.

PR #11183, opened 2021-08-17
mix deps.compile: Copy priv/ before compilation

wojtekmach:
With this patch, and the commit just before it (f4f44e2), we can correctly mix install both of these packages:

```elixir
Mix.install [
  :certifi,
  :erqwest
]

IO.inspect :certifi.cacertfile()

:ok = :erqwest.start_client(:default)
IO.inspect :erqwest.get(:default, "https://hex.pm/api", %{headers: [{"user-agent", "erqwest"}]})
```

### DIFF
--- a/lib/mix/lib/mix/tasks/deps.compile.ex
+++ b/lib/mix/lib/mix/tasks/deps.compile.ex
@@ -234,7 +234,7 @@ defmodule Mix.Tasks.Deps.Compile do
 
     # Copy priv/ after compilation too if it was created then
     if File.exists?(dep_priv) and not File.exists?(build_priv) do
-      Mix.Utils.symlink_or_copy(config[:build_embedded] == true, dep_priv, build_priv)
+      Mix.Utils.symlink_or_copy(config[:build_embedded], dep_priv, build_priv)
     end
 
     Code.prepend_path(Path.join(build_path, "ebin"))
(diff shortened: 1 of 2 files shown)


lib/mix/lib/mix/tasks/deps.compile.ex
```diff
@@ -198,56 +198,44 @@ defmodule Mix.Tasks.Deps.Compile do
     true
   end
 
-  defp do_rebar3(%Mix.Dep{opts: opts} = dep, config) do
+defp do_rebar3(%Mix.Dep{opts: opts} = dep, config) do
```
josevalim, line 201:
```suggestion
  defp do_rebar3(%Mix.Dep{opts: opts} = dep, config) do
```

lib/mix/lib/mix/tasks/deps.compile.ex
```diff
@@ -201,53 +201,41 @@ defmodule Mix.Tasks.Deps.Compile do
...
-    )
+    build_priv = Path.join(build_path, "priv")
+    dep_priv = Path.join(dep_path, "priv")
+
+    # Copy priv/ after compilation too if it was created then
+    if File.exists?(dep_priv) and not File.exists?(build_priv) do
+      Mix.Utils.symlink_or_copy(config[:build_embedded] == true, dep_priv, build_priv)
+    end
 
     Code.prepend_path(build_ebin)
```
josevalim, line 240:
```suggestion
    Code.prepend_path(Path.join(build_path, "ebin"))
```

josevalim:

PR #11186, opened 2021-08-17
Small improvements on docs of Regex related functions.

marciol:


lib/elixir/lib/kernel.ex
```diff
@@ -2163,6 +2163,7 @@ defmodule Kernel do
       iex> "abcd" =~ ""
       true
 
+  For more information about regular expressions, please check `Regex` docs.
```
josevalim, line 2166:
```suggestion
  For more information about regular expressions, please check the `Regex` module.
```

lib/elixir/lib/regex.ex
```diff
@@ -274,6 +274,8 @@ defmodule Regex do
       iex> Regex.match?(~r/foo/, "bar")
       false
 
+  Elixir also provides `Kernel.=~/2` and `String.match?/2` as alternative ways to test strings against regular expressions.
+
```
josevalim, line 278:
```suggestion
  Elixir also provides `Kernel.=~/2` and `String.match?/2` as
  an alternative to test strings against regular expressions and
  strings.
```

lib/elixir/lib/string.ex
```diff
@@ -2315,6 +2315,8 @@ defmodule String do
       iex> String.match?("bar", ~r/foo/)
       false
 
+  Elixir also provides `Kernel.=~/2` and `Regex.match?/2` as alternative ways to test strings against regular expressions.
+
```
josevalim, line 2319:
```suggestion
  Elixir also provides `Kernel.=~/2` and `Regex.match?/2` as
  alternatives to test strings against regular expressions.
```

josevalim:

Issue #11178, opened 2021-08-13
Do not recompile all code if mix.exs changes


josevalim:
Instead, we should make Mix.Project a compile-time dependency and only change modules that use it. We also need to handle `elixirc_options` and `elixirc_paths` and recompile whenever they change.

josevalim:
We should also make it so changing erlang modules just mark them as dirty instead of recompiling the whole project.

PR #11187, opened 2021-08-18
Make sure .beam files still exist for restored files 

### LINKED ISSUE #11176: Compilation error/interruption leave manifest intact but removes compiled code
Compilation state can be invalid following #11080:

1) Compile a project successfully.
2) Modify a file `a.ex` and introduce a compile error
3) `mix compile` => compile error
4) Revert `a.ex` to exactly what it was initially
4) `mix compile` says project is up to date, but `_build/dev/.../A.beam` is gone

Probably a more frequent way to create this (at least for me) is:
1) compile successfully
2) switch to a branch that changes `my_app.ex` to check something
3) forget to switch back, do `mix compile`, and when seeing `compiling 370 files...` thinking "oh oh, wrong branch" => abort compilation with control-C twice
4) switch back to original branch
5) state is now corrupted, as most compiled modules are gone but manifest intact.

Obvious solution is to invalidate the digest when we remove the compiled files. Making any of the errors above would imply that the compiled modules would be deleted and then recompiled.

A nicer solution would be to update the compiled files and manifest in sync, i.e. if a compilation fails or is interrupted, no change on the compiled module or manifest happen. Maybe the newly compiled modules could be written to a temporary location and moved to their final destination (overwritting previously compiled files) when compilation succeeds? In both scenarios above, no recompilation would occur on the last step but the compiled files would still be there. How feasible is that?

marcandre:
[fixes #11176]

### DIFF
--- a/lib/mix/lib/mix/compilers/elixir.ex
+++ b/lib/mix/lib/mix/compilers/elixir.ex
@@ -175,11 +175,8 @@ defmodule Mix.Compilers.Elixir do
     {[], %{}, all_paths, sources_stats}
   end
 
-  # Assume that either all .beam files are missing, or none of them are.
-  defp missing_beam_file?(dest, [module | _]) do
-    !File.exists?(beam_path(dest, module))
-  end
-
+  # Assume that either all .beam files are missing, or none of them are
+  defp missing_beam_file?(dest, [mod | _]), do: not File.exists?(beam_path(dest, mod))
   defp missing_beam_file?(_dest, []), do: false
 
   defp compiler_info_from_updated(
--- a/lib/mix/lib/mix/compilers/elixir.ex
+++ b/lib/mix/lib/mix/compilers/elixir.ex
@@ -175,6 +175,13 @@ defmodule Mix.Compilers.Elixir do
     {[], %{}, all_paths, sources_stats}
   end
 
+  # Assume that either all .beam files are missing, or none of them are.
+  defp missing_beam_file?(dest, [module | _]) do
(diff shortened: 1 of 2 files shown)


lib/mix/lib/mix/compilers/elixir.ex
```diff
@@ -175,6 +175,13 @@ defmodule Mix.Compilers.Elixir do
     {[], %{}, all_paths, sources_stats}
   end
 
+  # Assume that either all .beam files are missing, or none of them are.
+  defp missing_beam_file?(dest, [module | _]) do
+    !File.exists?(beam_path(dest, module))
+  end
+
+  defp missing_beam_file?(_dest, []), do: false
```
josevalim, line 183:
```suggestion
  # Assume that either all .beam files are missing, or none of them are
  defp missing_beam_file?(dest, [mod | _]), do: not File.exists?(beam_path(dest, mod))
  defp missing_beam_file?(_dest, []), do: false
```

josevalim:
Thank you! I was planning to do it today but I am glad you bet me to it because I would have checked ALL beam files. :D

For the .exs issues, I have opened #11178 and #11181. I plan to tackle the mix.exs one but I think the config one requires a bit more brewing.

josevalim:

marcandre:
Great, thanks.

Would you like me to backport this to the 1.12 branch?

josevalim:
@marcandre just backported. :)

marcandre:
Thanks! 🎉

PR #11190, opened 2021-08-19
Multiple improvements to file tracking

### LINKED ISSUE #11178: Do not recompile all code if mix.exs changes

josevalim:
Closes #11178.

PS: Rebase, do not squash.


josevalim:

PR #11192, opened 2021-08-19
Make Inspect.Opts.new/1 stricter

wojtekmach:


lib/elixir/lib/io.ex
```diff
@@ -416,7 +416,8 @@ defmodule IO do
   """
   @spec inspect(device, item, keyword) :: item when item: var
   def inspect(device, item, opts) when is_list(opts) do
-    label = if label = opts[:label], do: [to_chardata(label), ": "], else: []
+    {label, opts} = Keyword.pop(opts, :label)
+    label = if label, do: [to_chardata(label), ": "], else: []
     opts = Inspect.Opts.new(opts)
```
wojtekmach, line 421: changing Opts.new to use `struct!` could be considered a breaking change since before this, incorrect key names were silently ignored.

josevalim:

josevalim:
Reverted cause it broke Ecto.

wojtekmach:
I could make new/1 strict but the call site _not_ use new/1 and thus be more forgiving. This way we’d not break stuff but the new api will be "better".

otherwise if you think dropping a comment on new/1 that it is forgiving would be useful, I’d send a pr. I am happy to leave things exactly as they are now too.

josevalim:
A comment is good. We might introduce new! in the future?

PR #11196, opened 2021-08-20
Fix redundancies in Task and Task.Supervisor docs

v0idpwn:


marciol:
I think that "caller process" is better than "current caller process". Kudos @v0idpwn.

josevalim:

PR #11198, opened 2021-08-21
Update valid? for docs

seanmor5:
Small nit, but the docs for `valid?/1` render with an ambiguous `arg1` rather than following the rest of the API with `string`: https://hexdocs.pm/elixir/1.12/String.html#valid?/1


josevalim:

PR #11226, opened 2021-09-05
Add guards to IO.puts/2 and IO.inspect/3

eksperimental:
Adds is_device private guard cheking the argument is a device.

The use case is that I have wasted a lot of time trying to figure out why I wasn't able to print an IO.ANSI string.

I was executing `IO.inspect(term, printable_limit: :infinity)` on a big string, and I switched to IO.puts to see how it looked, 
`IO.puts(term, printable_limit: :infinity)`,  I was getting a cryptic CaseClause error not matching. Long story short, `IO.puts` does not take options, and the first argument is a device, not the term to inspect/print. So this guard will guide you in the right path.


josevalim:

PR #11231, opened 2021-09-06
Fix nfd cluster columns

### LINKED ISSUE #11142: column metadata does not consider grapheme clusters

WLSF:
- [x] Add tests for NFD vs NFC cases with columns
- [x] Add spaces between `Column + 2`
- [x] Add new validation to disregard `\r\n` tokens on cluster
- [x] Add `unicode_util:gc/1` to clusterize NFD symbols

Fix #11142

### DIFF
--- a/lib/elixir/src/elixir_interpolation.erl
+++ b/lib/elixir/src/elixir_interpolation.erl
@@ -52,12 +52,21 @@ extract(Line, Column, Scope, true, [$#, ${ | Rest], Buffer, Output, Last) ->
   end;
 
 extract(Line, Column, Scope, Interpol, [$\\, Char | Rest], Buffer, Output, Last) ->
-  extract(Line, Column+2, Scope, Interpol, Rest, [Char, $\\ | Buffer], Output, Last);
+  extract(Line, Column + 2, Scope, Interpol, Rest, [Char, $\\ | Buffer], Output, Last);
 
 %% Catch all clause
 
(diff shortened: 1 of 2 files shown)


lib/elixir/src/elixir_interpolation.erl
```diff
@@ -52,12 +52,21 @@ extract(Line, Column, Scope, true, [$#, ${ | Rest], Buffer, Output, Last) ->
...
-  extract(Line, Column+2, Scope, Interpol, Rest, [Char, $\\ | Buffer], Output, Last);
+  extract(Line, Column + 2, Scope, Interpol, Rest, [Char, $\\ | Buffer], Output, Last);
 
 %% Catch all clause
 
 extract(Line, Column, Scope, Interpol, [Char | Rest], Buffer, Output, Last) ->
-  extract(Line, Column + 1, Scope, Interpol, Rest, [Char | Buffer], Output, Last).
+  NewRest =
+    case Rest of
+      [$\r, $\n | _T] ->
```
WLSF, line 62: I had to add this validation, because when `unicode_util:gc/1` finds these tokens together it clusterizes them, and it was breaking the heredocs validation in the extract functions above

josevalim:

jpbrab0:
AEEEEEEEEEEEE! Parabéns Will!

PR #11233, opened 2021-09-08
IEx: Fix .iex.exs load after pry session

marciol:
Fix the loss of .iex.exs loaded code after a pry session

### DIFF
--- a/lib/iex/lib/iex/server.ex
+++ b/lib/iex/lib/iex/server.ex
@@ -2,7 +2,13 @@ defmodule IEx.State do
   @moduledoc false
   # This state is exchanged between IEx.Server and
   # IEx.Evaluator which is why it is a struct.
-  defstruct buffer: "", counter: 1, prefix: "iex", on_eof: :stop_evaluator, evaluator_options: []
+  defstruct buffer: "",
+            counter: 1,
+            prefix: "iex",
+            on_eof: :stop_evaluator,
+            evaluator_options: [],
+            previous_options: []
+
   @type t :: %__MODULE__{}
 end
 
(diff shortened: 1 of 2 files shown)


josevalim:
Hrm... this is actually a good discussion. For pry, we actually want to use the environment of the pried scope. We want to have its variables, its imports, etc. I think mixing with the dot-iex can be prone to confusion, unfortunately. Thoughts?

marciol:
I was thinking about the moment after the pry session. I think that does not make sense to carry all `.iex.exs` stuff during pry session, but it will be nice if we can get it back after the session. What do you think @josevalim?

marciol:
Ah, and it makes sense when one is using the REPL driven tools, like [Cortex](https://github.com/urbint/cortex) when working with the `iex` always running.

marciol:
Another alternative solution is to provide a new field on `IEx.State` struct, called `previous_evaluator_options`, so that we can put it in place again after the pry session.

josevalim:
@marciol yes, I think preserving it after Pry is a good call!

marciol:
I'll change the implementation and add more tests to ensure that the pried context will not get polluted with `.iex.exs` bindings.

marciol:
I worked on what I consider a better implementation. Let me know if there is something I can do to improve it.

lib/iex/lib/iex/server.ex
```diff
@@ -321,10 +344,24 @@ defmodule IEx.Server do
   defp iex_state(opts) do
     prefix = Keyword.get(opts, :prefix, "iex")
     on_eof = Keyword.get(opts, :on_eof, :stop_evaluator)
+    previous_options = Keyword.get(opts, :previous_options, [])
+
+    {previous_options, opts} =
+      cond do
+        prefix == "iex" and Enum.any?(previous_options) ->
```
josevalim, line 351: I don't like looking into the prefix to decide this because the prefix can be changed/customized. Is there anything else we can look into options to decide if we should use the previous options or not?

Or maybe we should just evaluate dot-iex again on respawn/continue?

marciol: Maybe evaluate it again can be a good option. I gonna try this.

josevalim:
Try it in another branch just in case this approach is the best one!
-- 

PR #11234, opened 2021-09-09
IEx: Fix .iex.exs load after pry session - Second Approach

marciol:
This is the second attempt to solve the loss of `.iex.exs` loaded code after an `IEx.pry` session.

The `IEx.pry` evaluator gets rid of the previous evaluator options, so to solve this problem we just hold the previous state when a pry session starts, so that it can be passed back when the pry session ends.


marciol:
I liked this approach more @josevalim. Your [original suggestion](https://github.com/elixir-lang/elixir/pull/11233#discussion_r704995000) is not possible because the evaluator is restarted during a takeover, as could be seen in all calls to `rerun` private function.

josevalim:

PR #11235, opened 2021-09-09
Fixes output of deps.unlock when dependency is not locked

supersimple:
Current behaviour:
image

Does not output empty line with * when no dependencies were unlocked
Adds test for that behaviour

### DIFF
--- a/lib/mix/test/mix/tasks/deps_test.exs
+++ b/lib/mix/test/mix/tasks/deps_test.exs
@@ -396,20 +396,13 @@ defmodule Mix.Tasks.DepsTest do
     end)
   end
 
-  test "unlocking a dep that is not locked does not output a blank line", context do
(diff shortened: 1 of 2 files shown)


supersimple:
I observed this today IRL when I was trying to get some transitive dependencies to play nicely

ericmj (approved):

lib/mix/test/mix/tasks/deps_test.exs
```diff
@@ -396,6 +396,23 @@ defmodule Mix.Tasks.DepsTest do
     end)
   end
 
+  test "unlocking a dep that is not locked does not output a blank line", context do
```
josevalim, line 399:
```suggestion
  test "unlocking a dep that is not locked is a no-op", context do
```

josevalim:

PR #11236, opened 2021-09-10
Kernel.tap/2: Explicitly ignore return value for compatibility with Dialyzer unmatched_returns

xtian:
When the `unmatched_returns` option of Dialyzer is enabled, functions which can return more than a single value must be explicitly matched against. This fixes `Kernel.tap/2` to be compatible with that option.


josevalim:

PR #11239, opened 2021-09-11
Map filter

Qqwy:
PR is based on [this proposal](https://groups.google.com/g/elixir-lang-core/c/oDIt4mKUc6A).

- [x] Implementation of `Map.filter/2` and `Map.reject/2`.
- [x] Implementation of the analogous `Keyword.filter/2` and `Keyword.reject/2`
- [x] `Map.map/2`
- [x] `Keyword.map/2`

### Questions:

- `:maps.filter/2` [supports a map iterator](https://github.com/erlang/otp/blob/master/lib/stdlib/src/maps.erl#L302) as first parameter. As we're not currently exposing map iterators through the `Map` module at all, my assumption is that this support is not required for `Map.filter/2` etc., simplifying the code a little. Is this correct?
- `:maps.filter/2`'s source code has [some error handling/decorating](https://github.com/erlang/otp/blob/master/lib/stdlib/src/maps.erl#L312). I'm not 100% sure how to convert this to Elixir code. Is this already handled by Elixir's own `FunctionClause`-handling where "function clauses that were tried" are printed? We might be able to [remove this line](https://github.com/elixir-lang/elixir/pull/11239/files#diff-17d8120ee6e35ed2a465a5e10985eeb91b64f54b38347daf429cc20d1c85df85R1050) from the implementation, maybe?

### DIFF
--- a/lib/elixir/lib/keyword.ex
+++ b/lib/elixir/lib/keyword.ex
@@ -1338,7 +1338,7 @@ defmodule Keyword do
 
   defp do_filter([], _fun), do: []
 
-  defp do_filter([entry | entries], fun) do
+  defp do_filter([{_, _} = entry | entries], fun) do
     if fun.(entry) do
       [entry | do_filter(entries, fun)]
     else
@@ -1366,7 +1366,7 @@ defmodule Keyword do
 
   defp do_reject([], _fun), do: []
 
-  defp do_reject([entry | entries], fun) do
+  defp do_reject([{_, _} = entry | entries], fun) do
     if fun.(entry) do
       do_reject(entries, fun)
     else
--- a/lib/elixir/lib/keyword.ex
+++ b/lib/elixir/lib/keyword.ex
@@ -1334,15 +1334,17 @@ defmodule Keyword do
   @doc since: "1.13.0"
   @spec filter(t, ({key, value} -> as_boolean(term))) :: t
   def filter(keywords, fun) when is_list(keywords) and is_function(fun, 1) do
-    boolean_fun =
-      fn element ->
-        case fun.(element) do
-          x when x in [false, nil] -> false
-          _ -> true
-        end
-      end
+    do_filter(keywords, fun)
+  end
 
-    :lists.filter(boolean_fun, keywords)
+  defp do_filter([], _fun), do: []
+
+  defp do_filter([element | vals], fun) do
+    if fun.(element) do
+      [element | do_filter(vals, fun)]
+    else
+      do_filter(vals, fun)
+    end
   end
 
   @doc """
@@ -1361,14 +1363,16 @@ defmodule Keyword do
   @doc since: "1.13.0"
   @spec reject(t, ({key, value} -> as_boolean(term))) :: t
   def reject(keywords, fun) when is_list(keywords) and is_function(fun, 1) do
-    boolean_fun =
-    fn element ->
-      case fun.(element) do
-        x when x in [false, nil] -> true
-        _ -> false
-      end
-    end
+    do_reject(keywords, fun)
+  end
 
-    :lists.filter(boolean_fun, keywords)
+  defp do_reject([], _fun), do: []
+
+  defp do_reject([element | vals], fun) do
+    if fun.(element) do
+      do_reject(vals, fun)
+    else
+      [element | do_reject(vals, fun)]
+    end
   end
 end
--- a/lib/elixir/lib/keyword.ex
+++ b/lib/elixir/lib/keyword.ex
@@ -1318,9 +1318,8 @@ defmodule Keyword do
   end
 
   @doc """
-  Filters `keywords` i.e. returns a keyword list containing
-  only the elements from `keywords` for which the function
-  `fun` returns a truthy value.
+  Returns a keyword list containing only the elements from `keywords`
+  for which the function `fun` returns a truthy value.
 
   See also `reject/2` which discards all elements where the function
   returns a truthy value.
--- a/lib/elixir/lib/keyword.ex
+++ b/lib/elixir/lib/keyword.ex
@@ -1334,7 +1334,7 @@ defmodule Keyword do
   @doc since: "1.13.0"
   @spec filter(t, ({key, value} -> as_boolean(term))) :: t
   def filter(keywords, fun) when is_list(keywords) and is_function(fun, 1) do
-    :lists.filter(keywords, fun)
+    :lists.filter(fun, keywords)
   end
 
   @doc """
@@ -1353,6 +1353,6 @@ defmodule Keyword do
   @doc since: "1.13.0"
   @spec reject(t, ({key, value} -> as_boolean(term))) :: t
   def reject(keywords, fun) when is_list(keywords) and is_function(fun, 1) do
-    :lists.filter(keywords, &(not fun.(&1)))
+    :lists.filter(&(not fun.(&1)), keywords)
   end
 end
--- a/lib/elixir/lib/keyword.ex
+++ b/lib/elixir/lib/keyword.ex
@@ -1350,8 +1350,7 @@ defmodule Keyword do
   Returns a keyword list excluding the entries from `keywords`
   for which the function `fun` returns a truthy value.
 
-  See also `filter/2` which discards all entries where the function
-  returns a falsy value.
+  See also `filter/2`.
 
   ## Examples
 
--- a/lib/elixir/lib/keyword.ex
+++ b/lib/elixir/lib/keyword.ex
@@ -1316,4 +1316,43 @@ defmodule Keyword do
   def size(keywords) do
     length(keywords)
   end
+
+  @doc """
+  Filters `keywords` i.e. returns a keyword list containing
+  only the elements from `keywords` for which the function
+  `fun` returns a truthy value.
+
+  See also `reject/2` which discards all elements where the function
+  returns a truthy value.
+
+  ## Examples
+
+      iex> Keyword.filter([one: 1, two: 2, three: 3], fn {_key, val} -> rem(val, 2) == 1 end)
+      [one: 1, three: 3]
+
+  """
+  @doc since: "1.13.0"
+  @spec filter(t, ({key, value} -> as_boolean(term))) :: t
+  def filter(keywords, fun) when is_list(keywords) and is_function(fun, 1) do
+    :lists.filter(keywords, fun)
+  end
+
+  @doc """
+  Returns a keyword list excluding the elements from `keywords`
+  for which the function `fun` returns a truthy value.
+
+  See also `filter/2` which discards all elements where the function
+  returns a falsy value.
+
+  ## Examples
+
+      iex> Keyword.reject([one: 1, two: 2, three: 3], fn {_key, val} -> rem(val, 2) == 1 end)
+      [two: 2]
+
+  """
+  @doc since: "1.13.0"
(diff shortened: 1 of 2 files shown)


lib/elixir/lib/keyword.ex
```diff
@@ -1316,4 +1316,43 @@ defmodule Keyword do
...
+  ## Examples
+
+      iex> Keyword.filter([one: 1, two: 2, three: 3], fn {_key, val} -> rem(val, 2) == 1 end)
+      [one: 1, three: 3]
+
+  """
+  @doc since: "1.13.0"
+  @spec filter(t, ({key, value} -> as_boolean(term))) :: t
+  def filter(keywords, fun) when is_list(keywords) and is_function(fun, 1) do
+    :lists.filter(keywords, fun)
```
josevalim, line 1337: Ah, we have to implement this in Elixir because Erlang doesn’t have our concept of truthiness.

josevalim:
About your questions, yes, we don’t need to support iterators and function clause errors are good enough!

lib/elixir/lib/keyword.ex
Qqwy: Even with `inline_list_funcs` it seems like the Erlang compiler is not able to currently optimize intermediate closures, so the final implementation is a hand-written tail-recursive implementation.

Qqwy:
I believe the implementation is finished.

Are there still places where e.g. the formatting of the documentation needs improvement?

lib/elixir/lib/keyword.ex
```diff
@@ -1316,4 +1316,87 @@ defmodule Keyword do
   def size(keywords) do
     length(keywords)
   end
+
+  @doc """
+  Filters `keywords` i.e. returns a keyword list containing
+  only the elements from `keywords` for which the function
+  `fun` returns a truthy value.
```
christhekeele, line 1323: We should make this one as terse as Reject's docs, ex:

```
Returns a keyword list containing only the elements from `keywords`
for which the function `fun` returns a truthy value.
```

v0idpwn, line 1323:
```suggestion
  Filters a keyword list, i.e. returns a keyword list containing
  only the pairs for which `fun` returns a truthy value.
```

lib/elixir/lib/keyword.ex
```diff
@@ -1316,4 +1316,87 @@ defmodule Keyword do
...
+    if fun.(element) do
+      [element | do_filter(vals, fun)]
+    else
+      do_filter(vals, fun)
+    end
+  end
+
+  @doc """
+  Returns a keyword list excluding the elements from `keywords`
+  for which the function `fun` returns a truthy value.
```
v0idpwn, line 1352:
```suggestion
  Returns a keyword list excluding the elements for which
  `fun` returns a truthy value.
```

lib/elixir/lib/keyword.ex
```diff
@@ -1316,4 +1316,87 @@ defmodule Keyword do
...
+
+  See also `filter/2` which discards all elements where the function
+  returns a falsy value.
```
v0idpwn, line 1355:
```suggestion
  See also `filter/2`.
```

lib/elixir/lib/map.ex
```diff
@@ -971,4 +971,115 @@ defmodule Map do
...
+  @doc """
+  Filters `map`, i.e. returns a map containing
+  only those elements for which the function `fun` returns a truthy value.
+
+  `fun` receives the key and value of each of the
+  elements in the map as a `{key, value}` tuple.
+
+  `Map.filter/2` is faster than using `map |> Enum.filter(fun) |> Enum.into(%{})`,
+  as no intermediate list is being built.
+
```
v0idpwn, line 984:
```suggestion
```

lib/elixir/lib/map.ex
```diff
@@ -971,4 +971,115 @@ defmodule Map do
   def size(map) do
     map_size(map)
   end
+
+  @doc """
+  Filters `map`, i.e. returns a map containing
+  only those elements for which the function `fun` returns a truthy value.
```
v0idpwn, line 977:
```suggestion
  Filters `map`, i.e. returns a map containing only those elements for
  which `fun` returns a truthy value.
```

lib/elixir/lib/map.ex
```diff
@@ -971,4 +971,115 @@ defmodule Map do
...
+    else
+      do_filter(next(iter), fun)
+    end
+  end
+
+  @doc """
+  Returns map excluding the elements from `map` for which
+  the function `fun` returns a truthy value.
+
+
```
v0idpwn, line 1017:
```suggestion

```

lib/elixir/lib/map.ex
```diff
@@ -971,4 +971,115 @@ defmodule Map do
...
+    if fun.({key, value}) do
+      [{key, value} | do_filter(next(iter), fun)]
+    else
+      do_filter(next(iter), fun)
+    end
+  end
+
+  @doc """
+  Returns map excluding the elements from `map` for which
+  the function `fun` returns a truthy value.
```
v0idpwn, line 1015:
```suggestion
  Returns map excluding the elements from `map` for which `fun` returns a truthy value.
```

lib/elixir/lib/map.ex
```diff
@@ -971,4 +971,115 @@ defmodule Map do
...
+
+
+  `Map.reject/2` is faster than using `map |> Enum.reject(fun) |> Enum.into(%{})`,
+  as no intermediate list is being built.
+
+  See also `filter/2` which discards all elements where the
+  function returns a falsy value.
```
v0idpwn, line 1022:
```suggestion
  See also `filter/2`.
```

lib/elixir/lib/keyword.ex
```diff
@@ -1316,4 +1316,87 @@ defmodule Keyword do
...
+      do_reject(vals, fun)
+    else
+      [element | do_reject(vals, fun)]
+    end
+  end
+
+  @doc """
+  Maps the function `fun` over all `{key, value}`-elements in `keywords`,
+  returning a keyword list with all the values replaced with
+  the result of the function.
```
christhekeele, line 1382: The Elixir stdlib docs are not consistent about this, but it does seem that the most common rendering of this is "key-value pairs", fully outside code blocks, whether or not the tuple literal is in the spec.

lib/elixir/lib/map.ex
christhekeele, line 977: Same thoughts about brevity as `Keyword.filter/2`

lib/elixir/lib/map.ex
```diff
@@ -971,4 +971,115 @@ defmodule Map do
...
+    if fun.({key, value}) do
+      do_reject(next(iter), fun)
+    else
+      [{key, value} | do_reject(next(iter), fun)]
+    end
+  end
+
+  @doc """
+  Maps the function `fun` over all `{key, value}`-elements in `map`, returning a map
+  with all the values replaced with the result of the function.
```
christhekeele, line 1050: Same thoughts as `Keyword.map/2`

christhekeele (changes requested):
I have just a few documentation formatting thoughts if interested, this looks great

v0idpwn (commented):
I think these changes could improve the documentation a bit.

While the moduledoc for `Keyword` specifies that functions don't guarantee ordering, I think maybe it could be good to point that in the new functions as it may be counter intuitive based on the name.

v0idpwn:
Oh no, me and @christhekeele made the same review 😄

lib/elixir/lib/keyword.ex
Qqwy: I did not make this change, because `Enum.reject/2` also contains this longer sentence.
Should we keep both long sentences, or also alter `Enum.reject/2`'s docs?

lib/elixir/lib/map.ex
Qqwy: I did not make this change, because Enum.reject/2 also contains this longer sentence.
Should we keep both long sentences, or also alter Enum.reject/2's docs?

lib/elixir/lib/keyword.ex
Qqwy: I wonder if we maybe should use the wording `entries` (rather than `elements` or `pairs`) since this is used in the rest of the `Keyword` module's documentation. What do you think?

lib/elixir/lib/keyword.ex
Qqwy: The current wording was based on `Enum.filter/2`'s:

> Filters the `enumerable`, i.e. returns only those elements
  for which `fun` returns a truthy value.

I do like your new terser wording better.

Qqwy: I chose the terser wording provided by @christhekeele 's suggestion above.

lib/elixir/lib/keyword.ex
v0idpwn: Agree. I didn't like elements and suggested pairs, but entries sounds better

lib/elixir/lib/map.ex
v0idpwn: `Enum.reject/2` is short, `Enum.filter` is long: https://github.com/elixir-lang/elixir/blob/d6d8ad419f8fe91dda7eb3f74e34041a924bd7f9/lib/elixir/lib/enum.ex#L2444 this is why I suggested

Qqwy: Ah, I see :smile: ! :+1:

christhekeele (approved):

lib/elixir/lib/keyword.ex
```diff
@@ -1316,4 +1316,85 @@ defmodule Keyword do
...
+  """
+  @doc since: "1.13.0"
+  @spec filter(t, ({key, value} -> as_boolean(term))) :: t
+  def filter(keywords, fun) when is_list(keywords) and is_function(fun, 1) do
+    do_filter(keywords, fun)
+  end
+
+  defp do_filter([], _fun), do: []
+
+  defp do_filter([entry | entries], fun) do
```
josevalim, line 1341:
```suggestion
  defp do_filter([{_, _} = entry | entries], fun) do
```

lib/elixir/lib/keyword.ex
```diff
@@ -1316,4 +1316,85 @@ defmodule Keyword do
...
+  """
+  @doc since: "1.13.0"
+  @spec reject(t, ({key, value} -> as_boolean(term))) :: t
+  def reject(keywords, fun) when is_list(keywords) and is_function(fun, 1) do
+    do_reject(keywords, fun)
+  end
+
+  defp do_reject([], _fun), do: []
+
+  defp do_reject([entry | entries], fun) do
```
josevalim, line 1369:
```suggestion
  defp do_reject([{_, _} = entry | entries], fun) do
```

josevalim:

s3cur3:
Should we document what happens when `Map.map/2` produces collisions in the keys? Given that the current implementation uses `:maps.from_list/2`, we could just use the same verbiage as `Map.new/2`:

> Duplicated keys are removed; the latest one prevails.

(I suppose we might conceivably _not_ want to make any guarantees in this area and just note that collisions in keys will cause all but one instance of the key to be dropped, hinting that maybe one shouldn’t don’t do that.)

josevalim:
@s3cur3 yes, a PR is welcome!

PR #11240, opened 2021-09-12
Use a b instead of n for Integer.extended_gcd

entertainyou:
a and b matches the documentation and this also makes the heading in
doc to be `def extended_gcd(a, b)` instead of `def extended_gcd(n, n)`.


josevalim:

PR #11241, opened 2021-09-13
Add power operator

josevalim:
Same semantics as in Nx.

Question: should the formatter format it as `left ** right` or `left**right`?

/cc @seanmor5 @polvalente @jonatanklosko


polvalente (commented):
@josevalim this is nice! I personally prefer `a ** b`, because it would be consistent with the other arithmetic infix binary operators.

Also, I think that `123**-45` is a little bit uglier than `123 ** -45`

whitepaperclip:
My 2¢ - I don't understand what the argument could be for `left**right` instead of `left ** right`. The latter is much easier to read and is consistent with the formatting of every other operator (afaik).

josevalim:
Alright, `left ** right` it is. Thanks everyone for the input!

josevalim:

PR #11243, opened 2021-09-13
Do not depend on :maps module internals in Map.{filter, reject, map}

### LINKED ISSUE #11242: New Map.{filter, map, reject} do not play nice with dialyzer
### Environment

```
$ elixir --version
Erlang/OTP 24 [erts-12.0.4] [source] [64-bit] [smp:6:6] [ds:6:6:10] [async-threads:1] [jit]
Elixir 1.13.0-dev (d74597f) (compiled with Erlang/OTP 24)
```
* Operating system: Ubuntu 20.04

### Current behavior
```elixir
defmodule Hello do
  def hello do
    filtered = Map.filter(%{hello: :world}, fn {k, _} -> k == :hello end)
    filtered.hello
  end
end
```

causes this dialyzer issue:
```
Function hello/0 has no local return.
```

This issue propagates to all functions that call this one.

### Expected behavior

No dialyzer issues that propagate when calling built-in Map functions

### Notes

Running dialyzer on Elixir itself shows those issues (erlang notation):
```
lib/map.ex:996: Function filter/2 has no local return

lib/map.ex:1027: Function reject/2 has no local return

lib/map.ex:1055: Function map/2 has no local return

lib/map.ex:1061: Function do_map/2 has no local return

lib/map.ex:1002: The pattern 
         <'none',
          __fun@1> can never match the type 
         <[any()] | {_, _, maps:iterator(_, _)},
          fun((_) -> any())>

lib/map.ex:1033: The pattern 
         <'none',
          __fun@1> can never match the type 
         <[any()] | {_, _, maps:iterator(_, _)},
          fun((_) -> any())>

lib/map.ex:1061: The pattern 
         <'none',
          __fun@1> can never match the type 
         <[any()] | {_, _, maps:iterator(_, _)},
          fun((_) -> any())>

lib/map.ex:1074: The attempt to match a term of type 
          maps:iterator(_, _) against the pattern 
          {_key@1, _val@1, _iter@1} breaks the opacity of the term

lib/map.ex:1079: The attempt to match a term of type 
          maps:iterator(_, _) against the pattern 
          'none' breaks the opacity of the term
```

michallepicki:
closes #11242

### DIFF
--- a/lib/elixir/lib/map.ex
+++ b/lib/elixir/lib/map.ex
@@ -994,8 +994,8 @@ defmodule Map do
   @doc since: "1.13.0"
   @spec filter(map, ({key, value} -> as_boolean(term))) :: map
   def filter(map, fun) when is_map(map) and is_function(fun, 1) do
-    iter = iterator(map)
-    next = next(iter)
+    iter = :maps.iterator(map)
+    next = :maps.next(iter)
     :maps.from_list(do_filter(next, fun))
   end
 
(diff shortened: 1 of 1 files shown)


josevalim:

PR #11246, opened 2021-09-14
Formatter plugins

josevalim:
This PR adds formatter plugins which customizes how the formatter behaves.
Plugins must implement the `Mix.Tasks.Format` behaviour. For example, imagine
that your project uses Markdown in two distinct ways: via a custom `~M` sigil and
via files with the `.md` and `.markdown` extensions. A custom plugin would look
like this:

```elixir
    defmodule MixMarkdownFormatter do
      @behaviour Mix.Tasks.Format

      def features(_opts) do
        [sigils: [:M], extensions: [".md", ".markdown"]]
      end

      def format(contents, opts) do
        # logic that formats markdown
      end
    end
```

Now any application can use your formatter as follows:

```elixir
    # .formatters.exs
    [
      # Define the desired plugins
      plugins: [MixMarkdownFormatter],
      # Remember to update the inputs list to include the new extensions
      inputs: ["{mix,.formatter}.exs", "{config,lib,test}/**/*.{ex,exs}", "posts/*.{md,markdown}"]
    ]
```

Remember that, when running the formatter with plugins, you must make
sure that your dependencies and your application has been been compiled,
so the relevant plugin code can be loaded. Otherwise a warning is logged.

This can be extremely useful for projects like Phoenix LiveView and Surface,
as they can now directly hook into Elixir's formatter.

### DIFF
--- a/lib/elixir/lib/code/formatter.ex
+++ b/lib/elixir/lib/code/formatter.ex
@@ -142,55 +142,6 @@ defmodule Code.Formatter do
   @do_end_keywords [:rescue, :catch, :else, :after]
 
   @doc """
-  Checks if two strings are equivalent.
-  """
-  def equivalent(string1, string2) when is_binary(string1) and is_binary(string2) do
-    quoted1 = :elixir.string_to_quoted!(to_charlist(string1), 1, 1, "nofile", [])
-    quoted2 = :elixir.string_to_quoted!(to_charlist(string2), 1, 1, "nofile", [])
-
-    case not_equivalent(quoted1, quoted2) do
-      {left, right} -> {:error, left, right}
-      nil -> :ok
-    end
-  end
-
-  defp not_equivalent({:__block__, _, [left]}, right) do
-    not_equivalent(left, right)
-  end
-
-  defp not_equivalent(left, {:__block__, _, [right]}) do
-    not_equivalent(left, right)
-  end
-
-  defp not_equivalent({:__block__, _, []}, nil) do
-    nil
-  end
-
-  defp not_equivalent(nil, {:__block__, _, []}) do
-    nil
-  end
-
-  defp not_equivalent([left | lefties], [right | righties]) do
-    not_equivalent(left, right) || not_equivalent(lefties, righties)
-  end
-
-  defp not_equivalent({left_name, _, left_args}, {right_name, _, right_args}) do
-    not_equivalent(left_name, right_name) || not_equivalent(left_args, right_args)
-  end
-
-  defp not_equivalent({left1, left2}, {right1, right2}) do
-    not_equivalent(left1, right1) || not_equivalent(left2, right2)
-  end
-
-  defp not_equivalent(side, side) do
-    nil
-  end
-
-  defp not_equivalent(left, right) do
-    {left, right}
-  end
-
-  @doc """
   Converts the quoted expression into an algebra document.
   """
   def to_algebra(quoted, opts \\ []) do
--- a/lib/elixir/lib/code/formatter.ex
+++ b/lib/elixir/lib/code/formatter.ex
@@ -203,7 +203,6 @@ defmodule Code.Formatter do
       |> state(opts)
 
     {doc, _} = block_to_algebra(quoted, @min_line, @max_line, state)
-
     doc
   end
 
@@ -227,13 +226,29 @@ defmodule Code.Formatter do
   defp state(comments, opts) do
     force_do_end_blocks = Keyword.get(opts, :force_do_end_blocks, false)
     locals_without_parens = Keyword.get(opts, :locals_without_parens, [])
+    sigils = Keyword.get(opts, :sigils, [])
+
+    sigils =
+      Map.new(sigils, fn {key, value} ->
+        with true <- is_atom(key) and is_function(value, 1),
+             [char] <- Atom.to_charlist(key),
+             true <- char in ?A..?Z do
+          {char, value}
+        else
+          _ ->
+            raise ArgumentError,
+                  ":sigils must be a keyword list with uppercased letters as key and an anonymous " <>
+                    "function expecting a single argument as value, got: #{inspect(sigils)}"
+        end
+      end)
 
     %{
       force_do_end_blocks: force_do_end_blocks,
       locals_without_parens: locals_without_parens ++ locals_without_parens(),
       operand_nesting: 2,
       skip_eol: false,
-      comments: comments
+      comments: comments,
+      sigils: sigils
     }
   end
 
@@ -1350,6 +1365,22 @@ defmodule Code.Formatter do
          [{:<<>>, _, entries}, modifiers] when is_list(modifiers) <- args,
          opening_delimiter when not is_nil(opening_delimiter) <- meta[:delimiter] do
       doc = <<?~, name, opening_delimiter::binary>>
+
+      entries =
+        case state.sigils do
+          %{^name => callback} ->
+            case callback.(hd(entries)) do
+              binary when is_binary(binary) ->
+                [binary]
+
+              other ->
+                raise ArgumentError,
+                      "expected sigil callback to return a binary, got: #{inspect(other)}"
+            end
+
+          %{} ->
+            entries
+        end
 
       if opening_delimiter in [@double_heredoc, @single_heredoc] do
         closing_delimiter = concat(opening_delimiter, List.to_string(modifiers))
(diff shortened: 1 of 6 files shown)


lib/mix/lib/mix/tasks/format.ex
```diff
@@ -87,13 +83,46 @@ defmodule Mix.Tasks.Format do
...
+  It is possible to customize how the formatter behaves. Plugins must implement
+  the `Mix.Tasks.Format` behaviour. For example, imagine that your project uses
+  Markdown in two distinct ways: via a custom `~M` sigil and via files with the
+  `.md` and `.markdown` extensions. A custom plugin would look like this:
+
+      defmodule MixMarkdownFormatter do
+        @behaviour Mix.Tasks.Format
+
+        def features(_opts) do
+          [sigils: [:W], extensions: [".md", ".markdown"]]
```
mhanberg, line 97:
```suggestion
          [sigils: [:M], extensions: [".md", ".markdown"]]
```

paulstatezny:
On behalf of the entire Surface community, a big thank you! This is going to scratch a common itch. 🙂

lib/mix/lib/mix/tasks/format.ex
```diff
@@ -52,29 +55,22 @@ defmodule Mix.Tasks.Format do
...
+      reject contributions with unformatted code. If the check fails,
+      the formatted contents are not written to disk. Keep in mind
       that the formatted output may differ between Elixir versions as
-      improvements and fixes are applied to the formatter.
-
-    * `--check-equivalent` - checks if the files after formatting have the
-      same AST as before formatting. If the ASTs are not equivalent, it is
-      a bug in the code formatter. This option is useful if you suspect you
-      have ran into a formatter bug and you would like confirmation.
+      improvements and fixes are applied to the formatter. If `--check-*`
```
fertapric, line 61:
```suggestion
      improvements and fixes are applied to the formatter.
```

lib/mix/lib/mix/tasks/format.ex
```diff
@@ -87,13 +83,46 @@ defmodule Mix.Tasks.Format do
...
+      # .formatters.exs
+      [
+        # Define the desired plugins
+        plugins: [MixMarkdownFormatter],
+        # Remember to update the inputs list to include the new extensions
+        inputs: ["{mix,.formatter}.exs", "{config,lib,test}/**/*.{ex,exs}", "posts/*.{md,markdown}"]
+      ]
+
+  Remember that, when running the formatter with plugins, you must make
+  sure that your dependencies and your application has been been compiled,
```
juanperi, line 116:
```suggestion
  sure that your dependencies and your application has been compiled,
```

lib/elixir/lib/code/formatter.ex
```diff
@@ -227,13 +177,29 @@ defmodule Code.Formatter do
...
+    sigils =
+      Map.new(sigils, fn {key, value} ->
+        with true <- is_atom(key) and is_function(value, 1),
+             [char] <- Atom.to_charlist(key),
+             true <- char in ?A..?Z do
+          {char, value}
+        else
+          _ ->
+            raise ArgumentError,
+                  ":sigils must be a keyword list with uppercased letters as key and an anonymous " <>
```
leandrocp, line 191: Maybe?

```suggestion
                  ":sigils must be a keyword list with a single uppercased letter as key and an anonymous " <>
```

lib/mix/lib/mix/tasks/format.ex
```diff
@@ -135,32 +164,64 @@ defmodule Mix.Tasks.Format do
   @manifest "cached_dot_formatter"
   @manifest_vsn 1
 
+  @doc """
+  Returns which features this plugin should plug into.
+  """
+  @callback features(Keyword.t()) :: [sigils: [atom()], extensions: [binary()]]
```
fertapric, line 170: What do you think about `inputs`, `sources`, `locations`, or `identifiers` as a replacement for `features`? For some reason, the term ["feature"](https://en.wikipedia.org/wiki/Software_feature) does not feel right to me 🙂

lib/mix/lib/mix/tasks/format.ex
```diff
@@ -191,6 +253,34 @@ defmodule Mix.Tasks.Format do
       Mix.raise("Expected :subdirectories to return a list of directories, got: #{inspect(subs)}")
     end
 
+    if not is_list(plugins) do
+      Mix.raise("Expected :plugins to return a list of directories, got: #{inspect(plugins)}")
```
fertapric, line 257:
```suggestion
      Mix.raise("Expected :plugins to return a list of modules, got: #{inspect(plugins)}")
```

lib/mix/lib/mix/tasks/format.ex
josevalim: I am not a big fan of `features` but I don't think any of the other ones make it justice. I thought maybe call it `plugin`, but I don't think it is better either. :(

lib/mix/lib/mix/tasks/format.ex
```diff
@@ -191,6 +253,34 @@ defmodule Mix.Tasks.Format do
...
+    for plugin <- plugins do
+      cond do
+        not Code.ensure_loaded?(plugin) ->
+          Mix.shell().error(
+            "Skipping formatter plugin #{inspect(plugin)} because module cannot be found"
+          )
+
+        not function_exported?(plugin, :features, 1) ->
+          Mix.shell().error(
+            "Skipping formatter plugin #{inspect(plugin)} because it does not define features/0"
```
fertapric, line 274:
```suggestion
            "Skipping formatter plugin #{inspect(plugin)} because it does not define features/1"
```

lib/mix/lib/mix/tasks/format.ex
fertapric: Oh! It could be `plug_in`.

Another alternative is to split `features` in two, `sigils` and `extensions`, but not a big fan of that option either :(

lib/mix/lib/mix/tasks/format.ex
```diff
@@ -390,10 +482,40 @@ defmodule Mix.Tasks.Format do
...
+      true ->
+        & &1
+    end
+  end
+
+  defp find_plugin_for_extension(formatter_opts, ext) do
+    plugins = Keyword.get(formatter_opts, :plugins, [])
+
+    Enum.find(plugins, fn plugin ->
+      Code.ensure_loaded?(plugin) and function_exported?(plugin, :features, 1) and
```
fertapric, line 504: Is this required? It seems the list of plugins is already validated in `eval_deps_and_subdirectories`.

lib/mix/lib/mix/tasks/format.ex
```diff
@@ -191,6 +253,34 @@ defmodule Mix.Tasks.Format do
...
+            "Skipping formatter plugin #{inspect(plugin)} because it does not define features/1"
+          )
+
+        true ->
+          :ok
+      end
+    end
+
+    formatter_opts = Keyword.put(formatter_opts, :plugins, plugins)
+
```
fertapric, line 283: What do you think about moving all this logic to a new function, like `validate_formatter_opts` or similar?

lib/mix/lib/mix/tasks/deps.loadpaths.ex
```diff
@@ -35,14 +35,11 @@ defmodule Mix.Tasks.Deps.Loadpaths do
...
 
     unless "--no-load-deps" in args do
-      load_paths =
-        for dep <- all,
-            path <- Mix.Dep.load_paths(dep) do
-          _ = Code.prepend_path(path)
-          path
-        end
-
-      prune_deps(config, load_paths, "--no-deps-check" in args)
```
fertapric, line 45: Why are deps no longer pruned? Performance?

lib/mix/lib/mix/tasks/format.ex
josevalim: They are validated BUT they are still kept because we only validate it when .formatter.exs is loaded but then the result of .formatter.exs is cached. So we need to look at the cached result and see if they have been loaded post caching or not.

lib/mix/lib/mix/tasks/deps.loadpaths.ex
josevalim: The commit message has some context but this is not necessary anymore. before we loaded everything in `_build/ENV/lib/*/ebin` but now we only load your deps, so we never load something we should not.

josevalim:

PR #11247, opened 2021-09-14
fix: string.replace with empty string in pattern

yashin5:
that is the actual behaviour when have a empty string in a list.
```elixir
iex(30)> String.replace("Elixir", [""], "")
    (stdlib 3.14.2) :binary.matches("Elixir", [""])
    (elixir 1.11.4) lib/string.ex:1491: String.replace_guarded/4
iex(30)> String.replace("Elixir", ["a", ""], "")
    (stdlib 3.14.2) :binary.matches("Elixir", ["a", ""])
    (elixir 1.11.4) lib/string.ex:1491: String.replace_guarded/4
```

if the empty string isn't in a list, it doesn't break, so I assume this isn't the expected behavior.
```elixir
iex(30)> String.replace("Elixir", "", "")       
"Elixir"
```

### DIFF
--- a/lib/elixir/lib/string.ex
+++ b/lib/elixir/lib/string.ex
@@ -1506,6 +1506,10 @@ defmodule String do
     end
   end
 
+  defp replace_guarded(subject, pattern_list, replacement, options) when is_list(pattern) and "" in pattern do
+    replace_guarded(subject, Enum.filter(pattern, &(&1 == "")), replacement, options)
+  end
+
   defp replace_guarded(subject, pattern, replacement, options) do
     if insert = Keyword.get(options, :insert_replaced) do
       IO.warn(
--- a/lib/elixir/lib/string.ex
+++ b/lib/elixir/lib/string.ex
@@ -1506,7 +1506,7 @@ defmodule String do
     end
   end
 
-  defp replace_guarded(subject, pattern_list, replacement, options) when is_list(pattern) and "" in pattern do
+  defp replace_guarded(subject, pattern, replacement, options) when is_list(pattern) and "" in pattern do
     replace_guarded(subject, Enum.filter(pattern, &(&1 == "")), replacement, options)
   end
 


lib/elixir/lib/string.ex
```diff
@@ -1506,6 +1506,10 @@ defmodule String do
     end
   end
 
+  defp replace_guarded(subject, pattern_list, replacement, options) when is_list(pattern) and "" in pattern do
```
yashin5, line 1509:
```suggestion
  defp replace_guarded(subject, pattern, replacement, options) when is_list(pattern) and "" in pattern do
```

josevalim:
This behaviour comes from Erlang, so I would prefer to document it. What if we mention in the docs that it must be a list of non-empty strings? There are probably other examples in the docs we need to update. Thank you :)

yashin5:
> This behaviour comes from Erlang, so I would prefer to document it. What if we mention in the docs that it must be a list of non-empty strings? There are probably other function docs we need to update. Thank you :)

You welcome! :) 

But @josevalim, if you prefer just document it, why make a match in pattern when it is a empty string and only return the subject?

i mean about here:
```elixir
  defp replace_guarded(subject, "", "", _) do
    subject
  end
```

as far as I know, in erlang it also breaks if it's an empty string

josevalim:
At first we raised, but someone pointed there is a good property that describes replacing `""`:

```elixir
iex(1)> String.replace "foo", "", "bar"
"barfbarobarobar"
```

Ideally we would support empty strings in lists too but, meanwhile, I think the simplest path is to document the current behaviour.

yashin5:
> At first we raised, but someone pointed there is a good property that describes replacing `""`:
> 
> ```elixir
> iex(1)> String.replace "foo", "", "bar"
> "barfbarobarobar"
> ```
> 
> Ideally we would support empty strings in lists too but, meanwhile, I think the simplest path is to document the current behaviour.

Thanks for the explanation! \o

josevalim:
@yashin5 would you like to update the docs to mention it is a list of non-empty strings? Or should I go ahead? :)

PR #11254, opened 2021-09-16
Add support for `mix xref graph --label=compile-connected --excluded=…`

marcandre:
Currently, excluded paths still count for connections. In the example test, the file `d.ex` is said to be `compile-connected` because it has a single dependency on `e.ex`. If `e.ex` is excluded, I believe that `d.ex` should no longer be considered a compile-connected dependency. This PR fixes that.

Use case:
We `use MyApp` just about everywhere, and `lib/my_app.ex` is known to not refer to anything. I would like to ask for compile-connected dependencies but exclude `lib/my_app.ex`. This way I may call `use MyApp` even in the few files that we use as compile-time dependencies.

Implementation-wise, I am wondering if it wouldn't be cleaner and simpler to filter the whole list of `file_references` once at the beginning. This might entail a bit of unnecessary processing when explicit sources/sinks are provided (because it would go through the whole graph instead of only the subgraph for the given sources/sinks), but otherwise it's not clear if there would be any other issue besides that (probably negligible) performance loss.

### DIFF
--- a/lib/mix/lib/mix/tasks/xref.ex
+++ b/lib/mix/lib/mix/tasks/xref.ex
@@ -685,7 +685,7 @@ defmodule Mix.Tasks.Xref do
       end
 
     # Filter according to non direct label
-    file_references = filter(file_references, filter)
+    file_references = filter(file_references, excluded, filter)
 
     # If a label is given, remove empty root nodes
     file_references =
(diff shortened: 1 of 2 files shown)


josevalim:
I actually think this implementation may be more expensive than filtering everything at the beginning (since `-- []` will still fully traverse the left side to see if it is a proper list) but, for all purposes, both are good enough. So I will go ahead and merge this BUT, if you want to explore another solution and send a PR, I will gladly review it too. :+1:

josevalim:

PR #11257, opened 2021-09-17
Fix Version.Requirement.new/2 spec

michallepicki:
Some other changes in this module allowed dialyzer to detect and point out spec issues

### DIFF
--- a/lib/elixir/lib/version.ex
+++ b/lib/elixir/lib/version.ex
@@ -134,7 +134,7 @@ defmodule Version do
     @compile inline: [compare: 2]
 
     @doc false
-    @spec new(String.t(), :ets.match_spec()) :: t
+    @spec new(String.t(), [atom | matchable]) :: t
     def new(source, lexed) do
       %__MODULE__{source: source, lexed: lexed}
     end


michallepicki:
dialyzer's warnings that pointed out the issue (erlang format):
```
lib/mix/tasks/archive.check.ex:19: The created fun has no local return

lib/mix/tasks/archive.check.ex:43: Function parse_archive/1 has no local return

lib/version.ex:430: Function 'parse_requirement!'/1 has no local return

lib/mix/dep/loader.ex:131: The pattern 
          {'ok', _req@2} can never match the type 
          'error'

lib/mix/hex.ex:39: The call 'Elixir.Version':'match?'
         (any(),
          <<62,61,32,48,46,49,57,46,48>>) does not have an opaque term of type 
          'Elixir.Version.Requirement':t() as 2nd argument

lib/mix/local.ex:122: The call 'Elixir.Version':'match?'
         (<<_:80>>,
          _req@1 :: binary()) does not have an opaque term of type 
          'Elixir.Version.Requirement':t() as 2nd argument

lib/mix/rebar.ex:165: The pattern 
          {'ok', _} can never match the type 
          'error'

lib/mix/tasks/archive.check.ex:45: The pattern 
          {'ok', _req@2} can never match the type 
          'error'

lib/mix/tasks/deps.loadpaths.ex:49: The pattern 
          {'ok', _req@2} can never match the type 
          'error'

lib/version.ex:409: The call 'Elixir.Version.Requirement':new
         (_string@1 :: binary(),
          _lexed@1 ::
              ['!=' | '<' | '<=' | '==' | '>' | '>=' | 'and' | 'or' |
               '~>' |
               {_, _, _, _, _},
               ...]) breaks the contract 
          ('Elixir.String':t(), ets:match_spec()) -> t()

lib/version.ex:432: The pattern 
          {'ok', _requirement@1} can never match the type 
          'error'
```

josevalim:
Thank you! All of those contributions are deeply appreciated. :green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

PR #11259, opened 2021-09-17
Fix docs for new functions in Macro.Env

smaximov:
The docs for  `fetch_alias/2` and `fetch_macro_alias/2` state incorrectly that these functions return either the alias or `nil` (instead of `{:ok, atom()} | :error`).

In case of `lookup_import/2` I believe the proposed wording is more correct (though my English is not very good so I can be wrong).


josevalim:

PR #11261, opened 2021-09-17
Ignore `MIX_TEST_PARTITION` when partitions set to 1

hauleth:
This is meant to make GitLab CI integration with partitions much easier as will allow to use

```shell
MIX_TEST_PARTITION=${CI_NODE_INDEX} mix test --cover --partitions ${CI_NODE_TOTAL}
```

As a test task instead of additionally taking care of case when the parallel testing is disabled.

### DIFF
--- a/lib/mix/lib/mix/tasks/test.ex
+++ b/lib/mix/lib/mix/tasks/test.ex
@@ -486,7 +486,7 @@ defmodule Mix.Tasks.Test do
     cover =
       if opts[:cover] do
         compile_path = Mix.Project.compile_path(project)
-        partition = opts[:partitions] && System.get_env("MIX_TEST_PARTITION")
+        partition = opts[:partitions] > 1 && System.get_env("MIX_TEST_PARTITION")
 
         cover =
           @cover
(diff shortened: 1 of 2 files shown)


hauleth:
Just FYI, it wasn't much harder to make it work in current state, as all was needed was to use Bash "fallback if not set" syntax `:-1`:

```bash
MIX_TEST_PARTITION=${CI_NODE_INDEX:-1} mix test --cover --partitions ${CI_NODE_TOTAL}
```

But I think that proposed API makes it easier to work with.

lib/mix/lib/mix/tasks/test.ex
```diff
@@ -486,7 +486,7 @@ defmodule Mix.Tasks.Test do
     cover =
       if opts[:cover] do
         compile_path = Mix.Project.compile_path(project)
-        partition = opts[:partitions] && System.get_env("MIX_TEST_PARTITION")
+        partition = opts[:partitions] > 1 && System.get_env("MIX_TEST_PARTITION")
```
josevalim, line 489: Maybe we should Delete the option if it is set to 1 so we don’t have to handle it elsewhere?

josevalim:
Looks great! Just one comment and please add a mention to the docs and we are good to go!

josevalim:

PR #11263, opened 2021-09-19
Adding documentation to the IEx.Info protocol

### LINKED ISSUE #11262: IEx.Info protocol documentation?
When calling `i/1` in IEx, general information about the term passed to it is shown.
This is implemented as the `IEx.Info` protocol.

I am interested in implementing this protocol for my own types, to improve the description "This is a struct. Structs are maps with a __struct__ key." as well as potentially add some other definitions to the list of definitions.

However, IEx.Info is not currently documented at all.

- Should we add documentation to it?
- Or is this protocol intended to be 'private' to Elixir/IEx? If so, should we add some indication that this is the case somewhere?

Qqwy:
Fixes #11262 

This is my initial stab at adding this documentation.
It's based on my own interpretation of the code of the existing `IEx.Info`-implementations as well as `IEx.Helpers.i/1`.

Please let me know in what ways this documentation can be improved. I expect that parts are too specific, and in other parts important information might be missing.

:slightly_smiling_face:

### DIFF
--- a/lib/iex/lib/iex/info.ex
+++ b/lib/iex/lib/iex/info.ex
@@ -25,11 +25,14 @@ defprotocol IEx.Info do
   All other sections of information are added (and can be overridden)
   by customized implementations of this function.
 
-  It is recommended to at least include the following sections for a custom implementation:
+  It is recommended to at least include the following sections for a
+  custom implementation:
 
-  - `"Data type"`: Name of the data type. Usually the name of the module defining the data type.
+  - `"Data type"`: Name of the data type. Usually the name of the module
+     defining the data type.
   - `"Description"`: One or a few sentences describing what the data type represents.
-  - `"Reference modules`: One or a few comma-separated module names that focus on working with this datatype.
+  - `"Reference modules`: One or a few comma-separated module names that focus
+    on working with this datatype.
 
   Other recommended sections are:
 
@@ -67,7 +70,7 @@ defimpl IEx.Info, for: Atom do
     description =
       if atom == IEx.dont_display_result() do
         description = """
-        This atom is returned by IEx when a function that should not print itsd
+        This atom is returned by IEx when a function that should not print its
         return value on screen is executed.
         """
 
--- a/lib/iex/lib/iex/info.ex
+++ b/lib/iex/lib/iex/info.ex
@@ -14,10 +14,10 @@ defprotocol IEx.Info do
   @doc """
   Returns information for the given term.
 
-  Information should be returned as a list of `info_name`-`info` pairs,
-
-  `info_name` should be short.
-  It is OK for `info` to contain newlines.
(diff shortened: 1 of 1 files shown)


lib/iex/lib/iex/info.ex
```diff
@@ -1,6 +1,45 @@
...
+  the `"Data type"`, `"Description"` and `"Reference modules"` sections.
+  """
+
+  @doc """
+  Returns information for the given term.
+
+  Information should be returned as a list of `info_name`-`info` pairs,
+
+  `info_name` should be short.
+  It is OK for `info` to contain newlines.
```
josevalim, line 20:
```suggestion
  Information should be returned as a list of `info_name`-`info` tuples,
  where both values are strings. `info_name` should be short. `info` can
  be arbitrarily long and contain newlines.
```

lib/iex/lib/iex/info.ex
```diff
@@ -1,6 +1,45 @@
...
+     defining the data type.
+  - `"Description"`: One or a few sentences describing what the data type represents.
+  - `"Reference modules`: One or a few comma-separated module names that focus
+    on working with this datatype.
+
+  Other recommended sections are:
+
+  - `"Raw representation`: showing another way of writing the passed `term`.
+    This is mostly relevant for data-structures whose `String.Chars`-implementations
+    make use of sigils or other syntactic sugar.
```
josevalim, line 41:
```suggestion
    * `"Data type"`: Name of the data type. Usually the name of the module
       defining the data type.
    * `"Description"`: One or a few sentences describing what the data type represents.
    * `"Reference modules`: One or a few comma-separated module names that focus
      on working with this datatype.

  Other recommended sections are:

    * `"Raw representation`: showing another way of writing the passed `term`.
      This is mostly relevant for data-structures whose `String.Chars`-implementations
      make use of sigils or other syntactic sugar.
```

Just standardizing with other markdown in the codebase. :)

josevalim (commented):
Two tiny suggestions and we can ship it!

lib/iex/lib/iex/info.ex
Qqwy: I'm for this change, but note that the spec:

```
  @spec info(term()) :: [{info_name :: String.Chars.t(), info :: String.t()}]
```

as well as [iex/helpers.ex:503](https://github.com/elixir-lang/elixir/blob/master/lib/iex/lib/iex/helpers.ex#L503) seem to hint that there are cases where `info_name` is not necessarily a string. 
Do we want to alter the spec and potentially remove the call to `to_string`?

josevalim: I updated the suggestion. Let me know what you think.

Qqwy:
Suggestions incorporated :+1: ! Ready to ship when you are :ship: .

josevalim:

PR #11269, opened 2021-09-22
Update documentation for `apply/3`

kipcole9:
This commit documents that `apply/3` can only
be used to call public functions in the target
module.

### DIFF
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -305,6 +305,10 @@ defmodule Kernel do
   prefer `module.function(arg_1, arg_2, ..., arg_n)` as it is clearer than
   `apply(module, :function, [arg_1, arg_2, ..., arg_n])`.
 
+  The function to be invoked must be a public function in `module`,
+  `apply/3` cannot be used to invoke private functions, even to call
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/kernel.ex
```diff
@@ -305,6 +305,10 @@ defmodule Kernel do
   prefer `module.function(arg_1, arg_2, ..., arg_n)` as it is clearer than
   `apply(module, :function, [arg_1, arg_2, ..., arg_n])`.
 
+  The function to be invoked must be a public function in `module`,
+  `apply/3` cannot be used to invoke private functions, even to call
+  private functions in the same `module` as the call to `apply/3`.
```
Qqwy, line 310: Maybe  "even to call private functions in the same `module`"
is more understandable without the " as the call to `apply/3`" part added at the end?

josevalim, line 310: Generally speaking, I am not a big fan of saying what we can't do, because there are many things any function cannot do. What do you think about keeping it short as:

```suggestion
  `apply/3` cannot be used to call private functions.
```

josevalim:

josevalim:

PR #11270, opened 2021-09-23
Improve error message on bad Mix aliases

whatyouhide:
Before this commit, we were failing with a `FunctionClauseError` a bit deeper in Mix itself. Now, if you give a bad alias to `:aliases`, we raise a nice and helpful error message.

### DIFF
--- a/lib/mix/lib/mix/task.ex
+++ b/lib/mix/lib/mix/task.ex
@@ -462,6 +462,13 @@ defmodule Mix.Task do
     run_alias(t, alias_args, proj, original_task, res)
   end
 
+  defp run_alias([h | _], _alias_args, _proj, _original_task, _res) do
+    Mix.raise(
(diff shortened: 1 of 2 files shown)


lib/mix/lib/mix/task.ex
```diff
@@ -462,6 +462,13 @@ defmodule Mix.Task do
     run_alias(t, alias_args, proj, original_task, res)
   end
 
+  defp run_alias([h | _], _alias_args, _proj, _original_task, _res) do
+    Mix.raise(
+      "Invalid Mix alias format, aliases can be either a string (representing a command) " <>
```
josevalim, line 467:
```suggestion
      "Invalid Mix alias format, aliases can be either a string (representing a Mix task with arguments) " <>
```

josevalim (approved):

PR #11271, opened 2021-09-23
Better errors when arguments reordered

hauleth:
Earlier it was throwing `ArithmeticError` which seems confusing, instead now it will fail on function clause which should give user better feedback that the arguments are in wrong order.


josevalim:

PR #11273, opened 2021-09-27
Fix IO.chardata_to_string/1 misleading in exdoc

wingyplus:
In https://hexdocs.pm/elixir/1.12/IO.html#chardata_to_string/1, the function signature tells that it accept string:

Screenshot 2021-09-27 at 20-37-29 IO — Elixir v1 12 3

It's incorrect because it accept the `chardata` which's string (binary) or list which make confusing when reading the doc. This patch fix by add bodyless clause that accept `chardata` as a first argument.

### DIFF
--- a/lib/elixir/lib/io.ex
+++ b/lib/elixir/lib/io.ex
@@ -630,6 +630,8 @@ defmodule IO do
 
   """
   @spec chardata_to_string(chardata) :: String.t()
+  def chardata_to_string(chardata)
+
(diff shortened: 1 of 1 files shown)


wingyplus:
@josevalim Do we need to backport to the v1.12 branch?

josevalim:
No need. Next release is most likely 1.13. :)

wingyplus:
Got it :)

PR #11274, opened 2021-09-28
Fix Date.beginning_of_week/2 and Date.end_of_week/2 for none ISO calendars

NickNeck:
I have played with @kipcole9 [Cldr.Calendar.Coptic](https://hex.pm/packages/ex_cldr_calendars_coptic) and got the following results:
```elixir
iex(1)> date = Date.new!(2020, 07, 05, Cldr.Calendar.Coptic)
~D[2020-07-05 Cldr.Calendar.Coptic]
iex(2)> Date.beginning_of_week(date)
~D[2020-07-07 Cldr.Calendar.Coptic]
iex(3)> date = Date.new!(2020, 07, 14, Cldr.Calendar.Coptic)
~D[2020-07-14 Cldr.Calendar.Coptic]
iex(4)> Date.end_of_week(date)
~D[2020-07-13 Cldr.Calendar.Coptic]
```
Maybe this PR can fix this. In this fix, I assume that each calendar had 7 weekdays. I am not sure if this assumption is correct.

### DIFF
--- a/lib/elixir/lib/calendar/date.ex
+++ b/lib/elixir/lib/calendar/date.ex
@@ -803,6 +803,7 @@ defmodule Date do
     %Date{calendar: Calendar.ISO, year: year, month: month, day: day}
   end
 
+  @days_per_week 7
   def beginning_of_week(%{calendar: calendar} = date, starting_on) do
     %{year: year, month: month, day: day} = date
 
@@ -811,7 +812,11 @@ defmodule Date do
         %Date{calendar: calendar, year: year, month: month, day: day}
 
       {day_of_week, first_day_of_week, _} ->
-        add(date, -(day_of_week - first_day_of_week))
+        days =
+          first_day_of_week - day_of_week -
+            if day_of_week < first_day_of_week, do: @days_per_week, else: 0
+
+        add(date, days)
     end
   end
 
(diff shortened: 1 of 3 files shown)


josevalim:
I think the Coptic calendar may be wrong here. It is returning the day of week with iso notation but it should return the day of week in regards to its own calendar. For example, if it starts on a wednesday, then wednesday must be 1, tuesday must be 7, in a way that start_day_of_week < end_day_of_week. What do you think @kipcole9? You are the expert. :)

kipcole9:
José, I too suspect the issue is in my calendar code and this issue should move to my repo (please close this issue).  I have a large refactor coming in the next couple of weeks and will build this in.  The overall challenge is that the notion of start of week and end of week is not strongly defined for many calendars (it is clear for Islamic and Hebrew, but not completely clear for Chinese, Korean and Japanese  as examples).

josevalim:
> The overall challenge is that the notion of start of week and end of week is not strongly defined for many calendars 

I wouldn't worry to much about it because the contract requires a `:default` but also supports passing its own field, so if you pick a "bad" default, people can easily change it. Also, for chinese, korean, and japanese, you most likely have the locale information somewhere for the beginning of the week, no? It is likely the locale information for that matches the calendars. WDYT?

In any case, @NickNeck, can you please report this issue on the cldr repo? Thanks!

PR #11278, opened 2021-09-30
Correct return type in Regex.scan/3 function spec

Stratus3D:
This funtion can return a list of tuples containing two integers when
the `return: :index` option is provided.


josevalim:

Issue #11281, opened 2021-10-02
Add import Mod, only: :sigils


josevalim:
It can be useful when you want to import the sigils from a module but not the remaining of its functionality.

For those interested, the code that handles `only: :functions` and `only: :macros` is here:

https://github.com/elixir-lang/elixir/blob/master/lib/elixir/src/elixir_import.erl#L11-L16

PR #11284, opened 2021-10-03
Add import Mod, only: :sigils 

lurodrigo:
Implements #11281

Tried to implement this for fun and this is actually my first time writing Erlang, so the implementation is perhaps very naive. About `import_sigil_functions` and `import_sigil_macros`, I thought this was a case where a bit of duplication was better than trying to abstract stuff.

I ran the following code in `bin/iex` and it worked as expected:

```elixir 
defmodule MySigils do
  def sigil_i(string, []), do: String.to_integer(string)
  def sigil_i(string, [?n]), do: -String.to_integer(string)
  
  def some_function, do: 1
  defmacro some_macro, do: 1
end

defmodule Success do 
  @moduledoc "Will compile"
  import MySigils, only: :sigils
  
  def ten, do: ~i'10'
  def minus_ten, do: ~i'10'n
end

defmodule FunctionsError do 
  @moduledoc "Won't compile"
  import MySigils, only: :sigils
  
  def fail, do: some_function()
end

defmodule MacrosError do 
  @moduledoc "Won't compile"
  import MySigils, only: :sigils
  
  def fail, do: some_macro()
end
```


lib/elixir/src/elixir_import.erl
```diff
@@ -46,6 +50,29 @@ import_macros(Force, Meta, Ref, Opts, E) ->
     end
   end).
 
+import_sigil_functions(Meta, Ref, Opts, E) ->
+  calculate(Meta, Ref, Opts, ?key(E, functions), ?key(E, file), fun() ->
+    filter_sigils(get_functions(Ref))
+  end).
+
+import_sigil_macros(Force, Meta, Ref, Opts, E) ->
```
josevalim, line 58: The `Force` argument is not required here. :)

lib/elixir/src/elixir_import.erl
```diff
@@ -46,6 +50,29 @@ import_macros(Force, Meta, Ref, Opts, E) ->
...
+        elixir_errors:form_error(Meta, E, ?MODULE, {no_macros, Ref});
+      error ->
+        []
+    end
+  end).
+
+filter_sigils(Key) -> 
+  lists:filter(fun({Atom, _}) -> 
+    Name = atom_to_list(Atom),
+    (string:slice(Name, 0, 6) =:= "sigil_") and (length(Name) =:= 7)
```
josevalim, line 73:
```suggestion
    case atom_to_list(Atom) of
      "sigil_" ++ [L] when L >= $a, L =< $z; L >= $A, L =< $Z -> true;
      _ -> false
    end
```

josevalim (commented):
Beautiful! I have added to tiny comments. Then we just need tests (they can be in test/elixir/kernel/import_test.exs) and we can ship it!

lurodrigo:
> Beautiful! I have added to tiny comments. Then we just need tests (they can be in test/elixir/kernel/import_test.exs) and we can ship it!

Thank you! Applied your suggestions and wrote a test.

josevalim:

PR #11285, opened 2021-10-03
Change Calendar.year_of_era/1 callback to /3

kipcole9:
### Change Calendar.year_of_era/1 callback to /3
In order to support calendars which can change
era any time during the calendar year, the callback
Calendar.year_of_era/1 is changed to Calendar.year_of_era/3.

Since Calendar.ISO.year_of_era/1 is a public function,
default arguments for month and day are set if
year_of_era/1 is called to allow compatibility with
any existing code that relies upon this function.

### DIFF
--- a/lib/elixir/lib/calendar/date.ex
+++ b/lib/elixir/lib/calendar/date.ex
@@ -943,8 +943,8 @@ defmodule Date do
   @spec year_of_era(Calendar.date()) :: {Calendar.year(), non_neg_integer()}
   def year_of_era(date)
 
-  def year_of_era(%{calendar: calendar, year: year}) do
-    calendar.year_of_era(year)
+  def year_of_era(%{calendar: calendar, year: year, month: month, day: day}) do
+    calendar.year_of_era(year, month, day)
   end
 
   @doc """
--- a/lib/elixir/lib/calendar.ex
+++ b/lib/elixir/lib/calendar.ex
@@ -189,7 +189,7 @@ defmodule Calendar do
   @doc """
   Calculates the year and era from the given `year`.
   """
-  @callback year_of_era(year) :: {year, era}
+  @callback year_of_era(year, month, day) :: {year, era}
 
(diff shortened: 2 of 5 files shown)


josevalim:
@kipcole9 should we rather fix this instead? Let's introduce `calendar.year_of_era/3` instead.

kipcole9:
José, that's preferable I think since it doesn't introduce any incompatibilities in the public API.  I'll work on this over the next couple of days.

josevalim:

lib/elixir/lib/calendar/date.ex
```diff
@@ -943,8 +943,8 @@ defmodule Date do
   @spec year_of_era(Calendar.date()) :: {Calendar.year(), non_neg_integer()}
   def year_of_era(date)
 
-  def year_of_era(%{calendar: calendar, year: year}) do
-    calendar.year_of_era(year)
+  def year_of_era(%{calendar: calendar, year: year, month: month, day: day}) do
+    calendar.year_of_era(year, month, day)
```
wojtekmach, line 947: year_of_era/1 is present on Elixir 1.12 so I think this is a breaking change. Should we go through a deprecation period, in other words something like:

```elixir
if function_exported?(calendar, :year_of_era, 3) do
  calendar.year_of_era(year, month, day)
else
  IO.warn "year_of_era/1 is deprecated in favour of year_of_era/3"
  calendar.year_of_era(year)
end
```

kipcole9:
I did a scan of Github a few weeks back looking for other calendar implementations (other than my own) and found only one (a Persian calendar implementation).  Not a complete analysis though for sure.

While you are correct that its a breaking change to the callback, the same was true for the change to `day_of_week/3` changing to `day_of_week/4` and I don't think that made a disturbance in the force?

josevalim:
I added a `function_exported?` check just in case, as it is cheap and it doesn't hurt. :)

PR #11290, opened 2021-10-04
Always show step when range is descending

### LINKED ISSUE #11289: Incorrect inspect implementation for descending ranges with a positive step
### Environment

* Elixir & Erlang/OTP versions (elixir --version): v1.12.0
* Operating system: Linux Manjaro x86-64

### Current behavior

```elixir
iex> myrange = 43..42//1
43..42
iex> 43..42
43..42//-1
iex> myrange == 43..42
false

iex> Enum.to_list(myrange)
[]
iex> Enum.to_list(43..42)
[43, 42] # or the charlist equivalent
```

These results are confusing. I encountered this issue when writing doctests, where the values you need to write in the code-lines do not match the values you need to indicate in the expectation-lines.

### Expected behavior

When inspecting a descending range with a positive step, make sure the `//1` is not dropped.

As example, `inspect(43..42//1)` should probably be `"43..42//1"`. 
This would make it clear that it is not in fact the same as writing `43..42`.

v0idpwn:
Closes #11289

### DIFF
--- a/lib/elixir/lib/range.ex
+++ b/lib/elixir/lib/range.ex
@@ -349,7 +349,7 @@ defimpl Inspect, for: Range do
   import Inspect.Algebra
   import Kernel, except: [inspect: 2]
 
-  def inspect(first..last//1, opts) do
+  def inspect(first..last//1, opts) when last >= first do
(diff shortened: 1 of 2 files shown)


josevalim:

PR #11293, opened 2021-10-05
Adding a snippet to the docs of `defoverridable/1`

### LINKED ISSUE #11292: `defoverridable` allows changing `defp` to `def`. Is this intended or not?
### Environment

* Elixir & Erlang/OTP versions (elixir --version): v1.12.0 / OTP 24
* Operating system: Linux Manjaro x86-64

### Given

```elixir
defmodule Foo do        
  defp bar(), do: 42
  # In practice, the lines below would be added by some AST returned by a macro:
  defoverridable bar: 0
  def bar(), do: super() + 1
end
```
(This issue is based on https://github.com/Qqwy/elixir-type_check/issues/64 )

### Current behavior

This code happily compiles, resulting in the module `Foo` where `bar` is a public function.

### Expected behavior

I would have expected `defoverridable` to not have worked on a private function.
Or if it did, to not allow changing it from a private function to a public function.

Is the current behaviour intended? (If so, can we improve the documentation?) 
Or is this a bug?

Qqwy:
Adding a paragraph to the docs of `Kernel.defoverridable/1` noting that changing the visibility of the function is allowed.

Fixes #11292

### DIFF
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -1,4 +1,4 @@
- elixir_bootstrap module to be able to bootstrap Kernel.
+# Use elixir_bootstrap module to be able to bootstrap Kernel.
 # The bootstrap module provides simpler implementations of the
 # functions removed, simple enough to bootstrap.
 import Kernel,
--- a/lib/elixir/lib/kernel.ex
+++ b/lib/elixir/lib/kernel.ex
@@ -5187,9 +5187,9 @@ defmodule Kernel do
   An overridable function is lazily defined, allowing a developer to override
   it.
 
-  It is possible for the overridden function to have a different visibility
-  than the original function: A public function can be overridden by a
-  private function and vice-versa.
+  It is possible for the overridden definition to have a different visibility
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/kernel.ex
```diff
@@ -1,4 +1,4 @@
-# Use elixir_bootstrap module to be able to bootstrap Kernel.
+ elixir_bootstrap module to be able to bootstrap Kernel.
```
josevalim, line 1:
```suggestion
# Use elixir_bootstrap module to be able to bootstrap Kernel.
```

lib/elixir/lib/kernel.ex
```diff
@@ -5187,6 +5187,10 @@ defmodule Kernel do
   An overridable function is lazily defined, allowing a developer to override
   it.
```
josevalim, line 5188:
```suggestion
  If the user defines a new function or macro with the same name
  and arity, then the overridable ones are discarded. Otherwise, the
  original definitions are used.
```

josevalim:

PR #11295, opened 2021-10-05
add optional recursiveness to Map.from_struct method

perzanko:
I've encountered this specific issue with struct to map conversion (described for example [here](https://elixirforum.com/t/convert-a-nested-struct-into-a-nested-map/23814/10) or [here](https://groups.google.com/g/elixir-lang-talk/c/ABpynrektMw?pli=1)) so many times in my daily work that I decided to prepare a proposal of enhancement of method `Map.from_struct/1`.

The main idea is to allow people to recursively convert structs with nested structs to maps with nested maps.

```
user = %User{
  id: 1,
  first_name: "John",
  last_name: "Doe",
  contact_details: %User.ContactDetails{
    phone_number: "19998887777",
    email: "john.doe@example.com"
  },
  permissions: [
    %Permision{
       level: :super_admin
    }
  ]
}

result = Map.from_struct(user, true)

IO.inspect(result)
%{
  id: 1,
  first_name: "John",
  last_name: "Doe",
  contact_details: %{
    phone_number: "19998887777",
    email: "john.doe@example.com"
  }
  permissions: [
    %{
       id: 1,
       level: :super_admin
    }
  ]
}
```

I'd agree that this kind of conversion is much more expensive than the previous one, but I do believe that having a method that converts only a top-level struct is also very misleading. Let's give people at least a bult-in solution after they realize how the standard version works :)

WDYT?

### DIFF
--- a/lib/elixir/lib/map.ex
+++ b/lib/elixir/lib/map.ex
@@ -925,14 +925,41 @@ defmodule Map do
       #=> %{name: "john"}
 
   """
-  @spec from_struct(atom | struct) :: map
-  def from_struct(struct) when is_atom(struct) do
-    delete(struct.__struct__(), :__struct__)
+  @type recursively :: boolean()
+  @spec from_struct(atom | struct, recursively()) :: map
+  def from_struct(struct, recursively \\ false)
+
+  def from_struct(struct, recursively) when is_atom(struct) do
+    from_struct(struct.__struct__(), recursively)
   end
 
-  def from_struct(%_{} = struct) do
+  def from_struct(%_{} = struct, false) do
     delete(struct, :__struct__)
   end
+
+  def from_struct(%_{} = struct, true) do
+    iter = :maps.iterator(struct)
+    next = :maps.next(iter)
+    :maps.from_list(do_from_struct(next))
+  end
+
+  defp do_from_struct({key, _value, iter}) when key == :__struct__ do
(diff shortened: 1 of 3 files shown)


josevalim:
Hi @perzanko, thank you for the PR but I don't believe this belongs in Elixir.

The issue is that a blank recursive check is not always going to work as you expect. For example, what happens if one of your %User{} fields contains a MapSet? A MapSet is a struct but you likely want to keep it as a struct. The same would happen for %Range{} or even for some application specific values, like PostGIS types, which you want to keep them as structs as they are values, and not convert them to maps.

Therefore, the answer of what needs to be converted to a map recursively highly depends on the application. Maybe the recursive is a safe bet for your application, where you can make certain assumptions, but I don't think it works as a catch-all pattern in Elixir.

And then there is also the issue were you don't want to convert MapSet to a map but you may want to convert the structs inside a MapSet to a map. In other words, this is as complex as JSON rendering, where you need to handle each data structure individually. So another way someone could tackle this is to introduce a Mapable protocol and then you can @derive Mapable for each struct you want to convert to a map. This way you can keep it recursively working and traversing structs. But this of course belongs in a package, rather than Elixir.

josevalim:
@perzanko btw, if you want to send a PR that clarifies the docs to say that the conversion is not recursive, it would be welcome!

PR #11302, opened 2021-10-06
Fix infinite loop in String.slice/2

sarchertech:
The 2nd slice's guard clause doesn't match the previous. If you pass anything other than a binary, you get an infinite loop.


josevalim:

Issue #7928, opened 2018-07-18
Improve Logger's gen_event message


josevalim:
Breaking changes:

  * [ ] The timestamp should be in native units instead of a data structure
  * [ ] The metadata should be a map instead of a keyword list and we should no longer add module and function keys, furthermore file should be a charlist

These changes may also require changes to the formatter, as it receives both the old time structure and old metadata. In particular, we may want to pass the metadata keys as arguments to the formatter, instead of filtering them in the backend.

We can likely implement this today by having a flag or a function_exported? on each backend. If the new backend is used, it receives and uses the message as is. If the old backend is used, we will wrap it on another GenEvent handler and convert the old message to the new one on the fly.

We should probably start working on this around v1.12, once the Erlang's logger integration solidifies.

josevalim:
Btw, we may be able to keep this backwards compatible if we copy `gen_event`'s source into Logger's codebase and push our modifications in there. Although I would only do something like this when work towards 2.0.0 is effectively started, as we don't want to maintain a fork of `gen_event` without need.

hauleth:
Noa as #9333 is merged we already have integration done. About backends then as soon as `logger_olp` will became public then we can make all current backends "old backends" and promote using `:logger` handlers as the "new backends". This will solve all tasks mentioned there.

josevalim:
Good call. We likely want to postpone this one a bit because the answer may be: "write an Erlang backend now".

hauleth:
From what I see, if we manage to provide formatter in Elixir 1.10 then it can become hard warning in lucky 1.13 or less lucky 1.14 release.

chulkilee:
I'm maintaining own formatter implementing Logger.Formatter... and here are my thoughts (with exciting Elixir 1.11 changes). 

* timestamp - should use `:microsecond`, not `:native` to match with [`:logger.timestamp/0`](https://erlang.org/doc/man/logger.html#timestamp-0)
  * Also.. erlang defines `time` in metadata - so probably we should use that instead of passing it as separate argument?
* level - should use [`:logger.level`](https://erlang.org/doc/man/logger.html#type-level) not Elixir's logger level (we may drop `erl_level` when we *remove* "old" backend)
* module/function - I guess we want to use `mfa` key as [`:logger.metadata`](https://erlang.org/doc/man/logger.html#type-metadata) defines?

hauleth:
[...]

PR #11304, opened 2021-10-08
Emit notice logs in :console backend

riccardomanfrin:
yes.. this is not very "backward compatible"..
I guess adding some config flag to enable it, and
default it to false.. would be proper (?)

### DIFF
--- a/lib/logger/lib/logger/backends/console.ex
+++ b/lib/logger/lib/logger/backends/console.ex
@@ -138,6 +138,12 @@ defmodule Logger.Backends.Console do
   def handle_event({level, _gl, {Logger, msg, ts, md}}, state) do
     %{level: log_level, ref: ref, buffer_size: buffer_size, max_buffer: max_buffer} = state
 
+    {:erl_level, level} = List.keyfind(md, :erl_level, 0, {:erl_level, level})
+    level = case level do
+      :warning -> :warn
+      _ -> level
+    end
+
     cond do
       not meet_level?(level, log_level) ->
         {:ok, state}
@@ -245,11 +251,6 @@ defmodule Logger.Backends.Console do
   end
 
   defp log_event(level, msg, ts, md, %{device: device} = state) do
-    {:erl_level, level} = List.keyfind(md, :erl_level, 0, {:erl_level, level})
-    level = case level do
-      :warning -> :warn
-      _ -> level
-    end
     output = format_event(level, msg, ts, md, state)
     %{state | ref: async_io(device, output), output: output}
   end
--- a/lib/logger/lib/logger/backends/console.ex
+++ b/lib/logger/lib/logger/backends/console.ex
@@ -245,7 +245,7 @@ defmodule Logger.Backends.Console do
   end
 
   defp log_event(_level, msg, ts, md, %{device: device} = state) do
-    {:erl_level, level} = List.keyfind(md, :erl_level, 0, :not_found)
+    {:erl_level, level} = List.keyfind(md, :erl_level, 0, {:erl_level, :not_found})
 
     output = format_event(level, msg, ts, md, state)
     %{state | ref: async_io(device, output), output: output}
--- a/lib/logger/lib/logger/backends/console.ex
+++ b/lib/logger/lib/logger/backends/console.ex
@@ -241,7 +241,7 @@ defmodule Logger.Backends.Console do
       info: Keyword.get(colors, :info, :normal),
       warning: Keyword.get(colors, :warn, :yellow),
       error: Keyword.get(colors, :error, :red),
-      notice: Keyword.get(colors, :error, :blue),
+      notice: Keyword.get(colors, :info, :normal),
       enabled: Keyword.get(colors, :enabled, IO.ANSI.enabled?())
     }
   end
--- a/lib/logger/lib/logger/backends/console.ex
+++ b/lib/logger/lib/logger/backends/console.ex
@@ -244,8 +244,8 @@ defmodule Logger.Backends.Console do
     }
   end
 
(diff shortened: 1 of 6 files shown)


riccardomanfrin:
I've added support for `:notice` in the `formatter.ex` and changed the `console.ex` backend.. I hope this is limiting the propagation of the `:notice` to only the console backend, leaving the rest of backends happy (the key was there already as `erl_level`)

Is this what you meant?

It's in any case breaking the format of preexisting log collections in case someone was logging to notice.

lib/logger/lib/logger/backends/console.ex
```diff
@@ -232,15 +232,21 @@ defmodule Logger.Backends.Console do
...
       warn: Keyword.get(colors, :warn, :yellow),
       error: Keyword.get(colors, :error, :red),
+      notice: Keyword.get(colors, :error, :blue),
       enabled: Keyword.get(colors, :enabled, IO.ANSI.enabled?())
     }
   end
 
-  defp log_event(level, msg, ts, md, %{device: device} = state) do
+  defp log_event(_level, msg, ts, md, %{device: device} = state) do
+    {:erl_level, level} = List.keyfind(md, :erl_level, 0, :not_found)
```
josevalim, line 248:
```suggestion
    {:erl_level, level} = List.keyfind(md, :erl_level, 0, {:erl_level, :not_found})
```

lib/logger/lib/logger/backends/console.ex
```diff
@@ -232,15 +232,21 @@ defmodule Logger.Backends.Console do
...
       warn: Keyword.get(colors, :warn, :yellow),
       error: Keyword.get(colors, :error, :red),
+      notice: Keyword.get(colors, :error, :blue),
       enabled: Keyword.get(colors, :enabled, IO.ANSI.enabled?())
     }
   end
 
-  defp log_event(level, msg, ts, md, %{device: device} = state) do
+  defp log_event(_level, msg, ts, md, %{device: device} = state) do
+    {:erl_level, level} = List.keyfind(md, :erl_level, 0, {:erl_level, :not_found})
```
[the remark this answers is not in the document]
josevalim: Good find!

lib/logger/lib/logger/backends/console.ex
```diff
@@ -138,6 +138,12 @@ defmodule Logger.Backends.Console do
   def handle_event({level, _gl, {Logger, msg, ts, md}}, state) do
     %{level: log_level, ref: ref, buffer_size: buffer_size, max_buffer: max_buffer} = state
 
+    {:erl_level, level} = List.keyfind(md, :erl_level, 0, {:erl_level, level})
+    level = case level do
+      :warning -> :warn
+      _ -> level
+    end
```
josevalim, line 145: No, let's not convert back. `Logger.warn` is already deprecated, so it is a good opportunity to move to `warning`. :)

lib/logger/lib/logger/backends/console.ex
```diff
@@ -243,7 +239,7 @@ defmodule Logger.Backends.Console do
       emergency: Keyword.get(colors, :debug, :red),
       debug: Keyword.get(colors, :debug, :cyan),
       info: Keyword.get(colors, :info, :normal),
-      warn: Keyword.get(colors, :warn, :yellow),
+      warning: Keyword.get(colors, :warning, :yellow),
```
josevalim, line 242:
```suggestion
      warning: Keyword.get(colors, :warn, :yellow),
```

lib/logger/test/logger/backends/console_test.exs
```diff
@@ -154,7 +166,7 @@ defmodule Logger.Backends.ConsoleTest do
     Logger.configure_backend(:console, colors: [warn: :cyan])
 
     assert capture_log(fn -> Logger.warn("hello") end) ==
-             IO.ANSI.cyan() <> "hello" <> IO.ANSI.reset()
+             IO.ANSI.yellow() <> "hello" <> IO.ANSI.reset()
```
josevalim, line 169:
```suggestion
             IO.ANSI.cyan() <> "hello" <> IO.ANSI.reset()
```

hauleth (changes requested):
Ignoring all the comments, I have different concept how we can add support for more features from Erlang logger instead.

- add new optional callback for the backends, for example `type/0` that will return `:legacy | :rich` and will default to `:legacy` if not defined
- for `:legacy` backends receives message in the current, tuple based, format
- for `:rich` (can use different name there) backend receives message in "raw" format from the Erlang (possibly with some alterations to wrap `:report_cb` for example)

This approach should give us a way to make full transition in future (possibly still in multiple steps).

What do you think @josevalim about such approach? It would give a way to have smother transition period, and possibly migrate to "new" format even without bumping Logger's major version (which would imply bumping Elixir's major version).

lib/logger/lib/logger/formatter.ex
```diff
@@ -163,10 +163,10 @@ defmodule Logger.Formatter do
   defp output(:levelpad, level, _, _, _), do: levelpad(level)
   defp output(other, _, _, _, _), do: other
 
-  defp levelpad(:debug), do: ""
+  # TODO: Deprecate me on Elixir v1.13+ or later
```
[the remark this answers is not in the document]
josevalim: levelpad is not very useful now when one of the levels is "emergency". There will be just too much padding on something like "info" (5 chars to be more precise).

hauleth:
@josevalim you probably meant #7928. I can see if I can craft something like that as a part of Hacktoberfest.

I wouldn't say that it is strictly needed to get us there, but for sure can ease transition period (unless I really will be able to craft my PR before 1.13).

PR #11306, opened 2021-10-11
Fix `--fail-above` with `--exclude`.

marcandre:
This PR deals with exclusions once, which simplifies the code.

### DIFF
--- a/lib/mix/lib/mix/tasks/xref.ex
+++ b/lib/mix/lib/mix/tasks/xref.ex
@@ -586,10 +586,16 @@ defmodule Mix.Tasks.Xref do
 
   ## Graph
 
-  defp excluded(opts) do
-    opts
-    |> Keyword.get_values(:exclude)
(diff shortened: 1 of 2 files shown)


marcandre:
Thanks! Yes, I definitely want to fix that bug, so I'll have a separate fix by then.
As I'm using `excluded` strictly with `compile_connected`, I really can't think of a use case where there would be a difference in behavior (and thus argue one way or the other).

marcandre:
I rebased this branch on #11330, to show that the particular example of 3 "linear" dependencies (as I understood it) does not change with this PR (with or without a `--source`).

Is there a different scenario that I should be testing?

marcandre:
@josevalim Looks like this was not backported correctly to the 1.13 branch.

06594f526 only includes a test, while as 7efd70b28 includes the code fix and the test. Looks partly my fault, as my test appears not to be sufficient to reveal the issue...

PR #11309, opened 2021-10-13
Fix elixir test assertion for erlang >=OTP-24.1

jendis:


josevalim:
In this case the functionality is not broken, only the tests. Hence why I propose to change the function we are testing. :) But we can update both this branch and master, yes. :)

josevalim:

PR #11313, opened 2021-10-15
Tokenizer uses static_atoms_encoder on quoted keys 3ae53ac 

### LINKED ISSUE #11312: `Code.string_to_quoted/2` `static_atoms_encoder` is not called for quoted keys
### Environment

* Elixir & Erlang/OTP versions (elixir --version): `Elixir 1.12.1 (compiled with Erlang/OTP 24)` (reproduced on master)

### Current behavior

```elixir
iex> Code.string_to_quoted(~S(["foo bar": :baz]), static_atoms_encoder: fn x, _ -> {:ok, x} end)
{:ok, ["foo bar": "baz"]}
```

The `:"foo bar"` atom actually gets created. If adding `existing_atoms_only`, it fails:

```elixir
iex> Code.string_to_quoted(~S(["foo bar 2": :baz]), static_atoms_encoder: fn x, _ -> {:ok, x} end, existing_atoms_only: true)
** (ArgumentError) errors were found at the given arguments:

  * 1st argument: invalid UTF8 encoding

    :erlang.binary_to_existing_atom("foo bar 2", :utf8)

```

### Expected behavior

The encoder should be called on all static atoms and prevent atom leaks:

```elixir
iex> Code.string_to_quoted(~S(["foo bar": :baz]), static_atoms_encoder: fn x, _ -> {:ok, x} end)
{:ok, [{"foo bar", "baz"}]}
```

Note: I have a working fix, will open a PR.

sabiwara:
Make the tokenizer call `static_atoms_encoder` on quoted keys.

Question: I re-used `kw_identifier`, should I introduce a new one like `kw_identifier_quoted` to be consistent with `atom_qoted`?

Closes #11312

### DIFF
--- a/lib/elixir/test/erlang/tokenizer_test.erl
+++ b/lib/elixir/test/erlang/tokenizer_test.erl
@@ -79,7 +79,7 @@ kw_test() ->
   [{kw_identifier, {1, 1, nil}, 'A@!'}] = tokenize("A@!: "),
   [{kw_identifier, {1, 1, nil}, 'a@!'}] = tokenize("a@!: "),
   [{kw_identifier, {1, 1, nil}, foo}, {bin_string, {1, 6, nil}, [<<"bar">>]}] = tokenize("foo: \"bar\""),
-  [{kw_identifier_unsafe, {1, 1, nil}, [<<"+">>]}, {bin_string, {1, 6, nil}, [<<"bar">>]}] = tokenize("\"+\": \"bar\"").
+  [{kw_identifier, {1, 1, nil}, '+'}, {bin_string, {1, 6, nil}, [<<"bar">>]}] = tokenize("\"+\": \"bar\"").
 
 int_test() ->
   [{int, {1, 1, 123}, "123"}] = tokenize("123"),


lib/elixir/src/elixir_tokenizer.erl
```diff
@@ -757,16 +757,26 @@ handle_strings(T, Line, Column, H, Scope, Tokens) ->
...
+            when is_function(StaticAtomsEncoder) and is_binary(Part) ->
+          case unsafe_to_atom(Part, Line, Column - 1, Scope) of
+            {ok, Atom} ->
+              Token = {kw_identifier, {Line, Column - 1, nil}, Atom},
+              tokenize(Rest, NewLine, NewColumn + 1, NewScope, [Token | Tokens]);
+            {error, Reason} ->
+              {error, Reason, Rest, Tokens}
+          end;
+
+        {{ok, Unescaped}, _} ->
```
josevalim, line 771:
```suggestion
      case unescape_tokens(Parts, Line, Column, NewScope) of
        {ok, [Part]} when is_binary(Part) ->
          case unsafe_to_atom(Part, Line, Column - 1, Scope) of
            {ok, Atom} ->
              Token = {kw_identifier, {Line, Column - 1, nil}, Atom},
              tokenize(Rest, NewLine, NewColumn + 1, NewScope, [Token | Tokens]);
            {error, Reason} ->
              {error, Reason, Rest, Tokens}
          end;

        {ok, Unescaped} ->
```

lib/elixir/test/elixir/kernel/parser_test.exs
```diff
@@ -154,6 +154,23 @@ defmodule Kernel.ParserTest do
                Code.string_to_quoted("there_is_no_such_var", static_atoms_encoder: encoder)
     end
 
+    test "encodes quoted atom keys" do
```
josevalim, line 157:
```suggestion
    test "encodes quoted keyword keys" do
```

josevalim:

PR #11316, opened 2021-10-16
Clean up a bit of unreachable code from iex/autocomplete.ex

michallepicki:
Dialyzer found an unreachable clause that can be removed


josevalim:
@michallepicki great find! Let's change it so it no longer returns an ok tuple. :)

lib/iex/lib/iex/autocomplete.ex
```diff
@@ -294,7 +294,7 @@ defmodule IEx.Autocomplete do
   defp expand_struct_fields_or_local_or_var(code, hint, shell) do
     with {:ok, quoted} <- Code.Fragment.container_cursor_to_quoted(code),
          {aliases, pairs} <- find_struct_fields(quoted),
-         {:ok, alias} <- value_from_alias(aliases, shell),
+         alias <- value_from_alias(aliases, shell),
```
josevalim, line 297:
```suggestion
         alias = value_from_alias(aliases, shell),
```

josevalim:

PR #11322, opened 2021-10-20
Expand System.stop/1 docs on binary exit status

smaximov:


lib/elixir/lib/system.ex
```diff
@@ -855,8 +855,13 @@ defmodule System do
...
   are closed before the system terminates by calling `halt/1`.
 
-  `status` must be a non-negative integer value which is returned by the
-  runtime system to the operating system.
+  `status` must be a non-negative integer or a binary.
+
+    * If an integer, the runtime system exits with the integer value which is
+      returned to the operating system.
+
+    * If a string, an Erlang crash dump is produced with status as slogan, and
```
josevalim, line 863:
```suggestion
    * If a binary, an Erlang crash dump is produced with status as slogan, and
```

PR #11329, opened 2021-10-21
Add specs and improve guards to Calendar.ISO.parse_*/2 functions

eksperimental:


eksperimental:
Please let me know if this change is desired, as the tests should be updated.
I am enforcing what is expressed in the docs, where it says that format should either be `:basic` or `:extended`, so it makes sense to raise `FunctionClauseError` is another value is passed.

eksperimental:
> Sorry, it was a misclick. :) Yeah, we should remove the `is_format`, but keep the specs. :)

But if the the function will still return, will not Dialyzer complain about it? should we update the spec to "format() | any()" ?

PR #11341, opened 2021-10-25
Add URI.new/1 and URI.new!/1

### LINKED ISSUE #10865: Deprecate URI.parse

josevalim:
Closes #10865.

### DIFF
--- a/lib/elixir/lib/uri.ex
+++ b/lib/elixir/lib/uri.ex
@@ -575,6 +575,8 @@ defmodule URI do
         userinfo: nil
       }}
   """
+  @doc since: "1.13.0"
+  @spec new(t() | String.t()) :: {:ok, t()} | {:error, String.t()}
   def new(%URI{} = uri), do: {:ok, uri}
 
   def new(binary) when is_binary(binary) do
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/uri.ex
```diff
@@ -465,89 +476,198 @@ defmodule URI do
...
+  @doc """
+  Parses a well-formed URI into its components.
+
+  This function is deprecated as it fails to raise in case of invalid URIs.
+  Use `URI.new!/1` or `URI.new/1` instead. In case you want to mimic the
+  behaviour of this function, you can do:
+
+      case URI.new(path) do
+        {:ok, uri} -> uri
+        {:error, _, _} -> %URI{path: path}
```
kipcole9, line 668: I believe this example should be `{:error, _}` not `{:error, _, _}`. A 2-tuple is returned from `URI.new/1`'; a 3-tuple is returned from `:uri_string.parse/1`.

PR #11349, opened 2021-10-27
Add Enum.slide/3

s3cur3:
[Below is my intro/pitch from the [previous discussion](https://groups.google.com/g/elixir-lang-core/c/LYmkUopaWN4/m/SwBzRt2zBQAJ) on the elixir-lang-core mailing list.]

Rotate pulls a slice of an enumerable out and replaces that slice somewhere else.

The classic use case is this: Suppose you have a list of to-do items, which the user has ordered by priority:

1. Apply to college
2. Brush the dog
3. Change the car's oil
4. Deliver flowers
5. Exchange gifts

A rotate occurs when the user selects some number of elements and drags them to a new place in the list. Let's say they selected items 3 & 4 from the preceding and dragged them above item 2. When they release the mouse, the new order should be:

1. Apply to college
2. Change the car's oil
3. Deliver flowers
4. Brush the dog
5. Exchange gifts

Doing this without the named algorithm requires 3 calls to `Enum.slice/3` (one at the insertion point, one at the start of the selected range, and one at the end of the selected range). It's easy to get the index math wrong, and it's even harder for readers of your code to grasp what's going on... and it's also substantially slower than the implementation here.

[A number of other languages](https://twitter.com/code_report/status/1419900906062204939) have a rotate algorithm, though it's still somewhat uncommon. I found [Dave Abrahams' comments](https://forums.swift.org/t/proposal-implement-a-rotate-algorithm-equivalent-to-std-rotate-in-c/491/2) valuable when this was discussed for inclusion in Swift.

### DIFF
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -2533,6 +2533,156 @@ defmodule Enum do
   end
 
   @doc """
+  Pulls out either a single element (denoted by an integer index) or a contiguous range of
+  the enumerable (given by a range) and inserts it in front of the value previously at the
+  insertion index.
+
+  The semantics of the range to be rotated match the semantics of Enum.slice/2. Specifically,
+  that means:
+
+  * Indices are normalized, meaning that negative indexes will be counted from the end
+    (for example, -1 means the last element of the enumerable). This will result in *two*
+    traversals of your enumerable on types like lists that don't provide a constant-time count.
+  * If the normalized index range's `last` is out of bounds, the range is truncated to the last element.
+  * If the normalized index range's `first` is out of bounds, the selected range for rotation
+    will be empty, so you'll get back your input list.
+  * Decreasing ranges (e.g., the range 5..0//1) also select an empty range to be rotated, so you'll
+    get back your input list.
+  * Ranges with any step but 1 will raise an error.
+
+  ## Examples
+
+      # Rotate a single element
+      iex> Enum.rotate([0, 1, 2, 3, 4, 5, 6], 5, 1)
+      [0, 5, 1, 2, 3, 4, 6]
+
+      # Rotate a range of elements backward
+      iex> Enum.rotate([0, 1, 2, 3, 4, 5, 6], 3..5, 1)
+      [0, 3, 4, 5, 1, 2, 6]
+
+      # Rotate a range of elements forward
+      iex> Enum.rotate([0, 1, 2, 3, 4, 5, 6], 1..3, 5)
+      [0, 4, 5, 1, 2, 3, 6]
+
+      # Rotate with negative indices (counting from the end)
+      iex> Enum.rotate([0, 1, 2, 3, 4, 5, 6], 3..-1//1, 2)
+      [0, 1, 3, 4, 5, 6, 2]
+      iex> Enum.rotate([0, 1, 2, 3, 4, 5, 6], -4..-2, 1)
+      [0, 3, 4, 5, 1, 2, 6]
+
+  """
+  def rotate(enumerable, range_or_single_index, insertion_index)
+
+  def rotate(enumerable, single_index, insertion_index) when is_integer(single_index) do
+    rotate(enumerable, single_index..single_index, insertion_index)
+  end
+
+  # This matches the behavior of Enum.slice/2
+  def rotate(_, _.._//step = index_range, _insertion_index) when step != 1 do
+    raise ArgumentError,
+          "Enum.rotate/3 does not accept ranges with custom steps, got: #{inspect(index_range)}"
+  end
+
+  # Normalize negative input ranges like Enum.slice/2
+  def rotate(enumerable, first..last, insertion_index) when first < 0 or last < 0 do
+    count = Enum.count(enumerable)
+    normalized_first = if first >= 0, do: first, else: first + count
+    normalized_last = if last >= 0, do: last, else: last + count
+
+    if normalized_first >= 0 and normalized_first < count and normalized_first != insertion_index do
+      normalized_range = normalized_first..normalized_last//1
+      rotate(enumerable, normalized_range, insertion_index)
+    else
+      Enum.to_list(enumerable)
+    end
+  end
+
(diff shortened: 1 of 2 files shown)


josevalim:
Just a quick note before I can review the PR, but I think `slide` may be a better name indeed. `rotate` somehow gives me an idea of mutation, but `slide` feels like a transformation. Does this make sense? WDYT?

s3cur3:
Slide works for me! Pushing the change now...

lib/elixir/lib/enum.ex
```diff
@@ -2532,6 +2532,156 @@ defmodule Enum do
...
+
+      # Slide a range of elements forward
+      iex> Enum.slide([0, 1, 2, 3, 4, 5, 6], 1..3, 5)
+      [0, 4, 5, 1, 2, 3, 6]
+
+      # Slide with negative indices (counting from the end)
+      iex> Enum.slide([0, 1, 2, 3, 4, 5, 6], 3..-1//1, 2)
+      [0, 1, 3, 4, 5, 6, 2]
+      iex> Enum.slide([0, 1, 2, 3, 4, 5, 6], -4..-2, 1)
+      [0, 3, 4, 5, 1, 2, 6]
```
josevalim, line 2571: I just realized something, by having the input list be numbers, someone can think that 1..3 is referring to the actual values and not the position. Shall we use `[:a, :b, :c, ...]` as input?

lib/elixir/lib/enum.ex
```diff
@@ -2532,6 +2532,156 @@ defmodule Enum do
     end
   end
 
+  @doc """
+  Pulls out either a single element (denoted by an integer index) or a contiguous range of
+  the enumerable (given by a range) and inserts it in front of the value previously at the
+  insertion index.
```
josevalim, line 2538:
```suggestion
  Slides a single or multiple elements given by `range_or_single_index` from `enumerable`
  to `insertion_index`.
```

lib/elixir/lib/enum.ex
```diff
@@ -2532,6 +2532,156 @@ defmodule Enum do
...
+
+  * Indices are normalized, meaning that negative indexes will be counted from the end
+    (for example, -1 means the last element of the enumerable). This will result in *two*
+    traversals of your enumerable on types like lists that don't provide a constant-time count.
+  * If the normalized index range's `last` is out of bounds, the range is truncated to the last element.
+  * If the normalized index range's `first` is out of bounds, the selected range for sliding
+    will be empty, so you'll get back your input list.
+  * Decreasing ranges (e.g., the range 5..0//1) also select an empty range to be moved, so you'll
+    get back your input list.
+  * Ranges with any step but 1 will raise an error.
```
josevalim, line 2551:
```suggestion
  The semantics of the range to be moved match the semantics of `Enum.slice/2`.
  Specifically, that means:

   * Indices are normalized, meaning that negative indexes will be counted from the end
      (for example, -1 means the last element of the enumerable). This will result in *two*
      traversals of your enumerable on types like lists that don't provide a constant-time count.

    * If the normalized index range's `last` is out of bounds, the range is truncated to the last element.

    * If the normalized index range's `first` is out of bounds, the selected range for sliding
      will be empty, so you'll get back your input list.

    * Decreasing ranges (such as `5..0//1`) also select an empty range to be moved,
      so you'll get back your input list.

    * Ranges with any step but 1 will raise an error.
```

lib/elixir/lib/enum.ex
```diff
@@ -2532,6 +2532,156 @@ defmodule Enum do
...
+        {pre, post} =
+          cond do
+            index < start -> {[item | pre], post}
+            index >= start and index < middle -> {pre, [item | post]}
+            index >= middle and index <= last -> {[item | pre], post}
+            true -> {pre, [item | post]}
+          end
+
+        {index + 1, pre, post}
+      end)
```
josevalim, line 2645: Beautiful! 💯

lib/elixir/lib/enum.ex
```diff
@@ -2532,6 +2532,156 @@ defmodule Enum do
...
+
+    :lists.reverse(pre, :lists.reverse(post))
+  end
+
+  # Like slide_any/4 above, this optimized implementation of slide for lists depends
+  # on the indices being sorted such that we're moving middle..last to be in front of start.
+  defp find_start([h | t], start, middle, last)
```
josevalim, line 2652: Let's call this `slide_list_start` and then `slide_list_middle` and `slide_list_last`?

lib/elixir/test/elixir/enum_test.exs
```diff
@@ -803,6 +803,148 @@ defmodule EnumTest do
...
+
+    test "on a single-element enumerable is the same as transforming to list" do
+      for enum <- [["foo"], [1], [%{foo: "bar"}], %{foo: :bar}, MapSet.new(["foo"]), 1..1] do
+        assert Enum.slide(enum, 0..0, 0) == Enum.to_list(enum)
+      end
+    end
+
+    test "moves a single element" do
+      for zero_to_20 <- [0..20, Enum.to_list(0..20)] do
+        expected_numbers = Enum.flat_map([0..7, [14], 8..13, 15..20], &Enum.to_list/1)
```
josevalim, line 821: Why not? :)

```suggestion
        expected_numbers = Enum.to_list(0..20)
```

josevalim:
@s3cur3 thank you for the PR! :heart: the code looks great, I have dropped some minor comments on the docs and on the function naming, everything else is ready to go!

lib/elixir/lib/enum.ex
s3cur3: Makes sense!

s3cur3:
Really appreciate the code review, José. ☺️ I've applied all your changes. 👍

josevalim:
If you are looking for a challenge, you can implement Stream.slide :) I believe the implementation would fit Stream.Reducers, so it should be a matter of converting the Enumerable implementation into the Stream.Reducers format!

s3cur3:
Sweet, I'll look into it! 👍

josevalim:

PR #11351, opened 2021-10-29
Fix formatting of map-like types in warnings

gliboc:
This came up in issue #11204:

When emitting a warning for a type unification error,
the compiler overly simplifies the formatting of types
when these are maps or unions of maps.

To address this, we recursively check for maps inside
of a union type when comparing it to another map-like type.

We also add an integration test related to issue #11204, 
showing the correctly formatted types after this fix.


lib/elixir/lib/module/types.ex
```diff
@@ -505,6 +505,10 @@ defmodule Module.Types do
   defp map_type?({:map, _}), do: true
   defp map_type?(_other), do: false
 
+  defp maplike_type?({:map, _}), do: true
+  defp maplike_type?({:union, union}), do: Enum.all?(union, &maplike_type?/1)
```
josevalim, line 509: Should this be `Enum.any?`? Any entry is enough to make it map like, right?

```suggestion
  defp maplike_type?({:union, union}), do: Enum.any?(union, &maplike_type?/1)
```

lib/elixir/test/elixir/module/types/integration_test.exs
```diff
@@ -688,6 +688,75 @@ defmodule Module.Types.IntegrationTest do
 
       assert_warnings(files, warning)
     end
+
+    test "collapses list types and fails to unify" do
```
josevalim, line 692: We should generally avoid integration tests because they become quite expensive over time. Ideally we would reduce this test case to a minimum necessary. You can check `types_test.exs` in the same directory for more unit like test. You should be able to reproduce this with:

```elixir
        warning(
          [map1, map2],
          (
            [_var1, _var2] = [map1, map2]
            %{} = map1
            %{} = map2.subkey
          )
        )
```

josevalim (commented):
@gliboc It is looking good! I have dropped two comments :)

josevalim:
Thank you for the changes @gliboc! Btw, Elixir has a code formatter, so whenever you make changes, remember to run `make format` so it formats the codebase. Depending on your editor, you may also have a shortcut for code formatting. :) I have formatted manually for now and I am merging it soon!

josevalim:

PR #11353, opened 2021-10-29
Typespecs: show expanded module name on inexistent struct field error

### LINKED ISSUE #11352: Typespec compiler not resolving __MODULE__ macro in error message
### Environment

  Erlang/OTP 24 [erts-12.0.3] [source] [64-bit] [smp:12:12] [ds:12:12:10] [async-threads:1] [jit]
* Operating system: 
  macOS BigSur 11.6

### Current behavior

We currently have a module using the `@enforce_keys` attribute and using the `__MODULE__` macro in the type definition.

```
defmodule App.Document do
  @enforce_keys ~w(
    field_1
    new_field_2
   )a

   defstruct(@enforce_keys)

  @type t :: %__MODULE__{
          field_1: String.t(),
          field_2: String.t()
  }
end
```
If I make a change to a field name in the `@enforce_keys` attribute but fail to update the `@type` I get a compiler error:
`(CompileError) lib/app/document.ex:38: undefined field :field_2 on struct __MODULE__`

### Expected behavior

The error message should be using the module name in the message:
`(CompileError) lib/app/document.ex:38: undefined field :field_2 on struct App.Document`

v0idpwn:
Closes #11352.

I believe that this behaviour it is consistent with other invalid field errors:
```
iex(1)> defmodule Foo.Bar do
...(1)>   defstruct [a: 1]
...(1)>   def bar, do: %__MODULE__{b: 1}
...(1)> end
** (KeyError) key :b not found
    expanding struct: Foo.Bar.__struct__/1
```

### DIFF
--- a/lib/elixir/lib/kernel/typespec.ex
+++ b/lib/elixir/lib/kernel/typespec.ex
@@ -573,9 +573,11 @@ defmodule Kernel.Typespec do
 
     fun = fn {field, _} ->
       unless Keyword.has_key?(struct, field) do
+        humanized_module = module |> Atom.to_string() |> String.replace("Elixir.", "")
+
         compile_error(
           caller,
-          "undefined field #{inspect(field)} on struct #{Macro.to_string(name)}"
+          "undefined field #{inspect(field)} on struct #{humanized_module}"
         )
       end
     end
--- a/lib/elixir/lib/kernel/typespec.ex
+++ b/lib/elixir/lib/kernel/typespec.ex
@@ -573,11 +573,9 @@ defmodule Kernel.Typespec do
 
     fun = fn {field, _} ->
       unless Keyword.has_key?(struct, field) do
-        humanized_module = module |> Atom.to_string() |> String.replace("Elixir.", "")
-
         compile_error(
           caller,
-          "undefined field #{inspect(field)} on struct #{humanized_module}"
+          "undefined field #{inspect(field)} on struct #{inspect(module)}"
         )
       end
     end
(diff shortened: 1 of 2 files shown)


lib/elixir/lib/kernel/typespec.ex
```diff
@@ -573,9 +573,11 @@ defmodule Kernel.Typespec do
 
     fun = fn {field, _} ->
       unless Keyword.has_key?(struct, field) do
+        humanized_module = module |> Atom.to_string() |> String.replace("Elixir.", "")
+
         compile_error(
           caller,
-          "undefined field #{inspect(field)} on struct #{Macro.to_string(name)}"
+          "undefined field #{inspect(field)} on struct #{humanized_module}"
```
josevalim, line 580:
```suggestion
          "undefined field #{inspect(field)} on struct #{inspect(module)}"
```

lib/elixir/lib/kernel/typespec.ex
```diff
@@ -573,9 +573,11 @@ defmodule Kernel.Typespec do
 
     fun = fn {field, _} ->
       unless Keyword.has_key?(struct, field) do
+        humanized_module = module |> Atom.to_string() |> String.replace("Elixir.", "")
+
```
josevalim, line 577:
```suggestion
```

josevalim:

lib/elixir/lib/kernel/typespec.ex
v0idpwn: TIL

PR #11354, opened 2021-10-30
Keep keyword format of tuple last element

doorgan:
This was originally reported and fixed by @msaraiva in https://github.com/doorgan/sourceror/pull/30

Considering a tuple like this one:
```elixir
{SomeModule, opt1: true, opt2: false}
```
The last element is a keyword list without square brackets, and the formatter preserves that style. If we use `Code.string_to_qutoed` and then `Code.quoted_to_algebra`(as `Macro.to_string` does), the keyword list gets brackets inserted:

```elixir
iex> opts = [literal_encoder: &{:ok, {:__block__, &2, [&1]}}]
iex> {:ok, quoted} = Code.string_to_quoted("{SomeModule, opt1: true, opt2: false}", opts)
iex> Macro.to_string(quoted)
"{SomeModule, [opt1: true, opt2: false]}"
```

### DIFF
--- a/lib/elixir/lib/code/normalizer.ex
+++ b/lib/elixir/lib/code/normalizer.ex
@@ -182,6 +182,11 @@ defmodule Code.Normalizer do
     end
   end
 
+  # Tuples
+  defp do_normalize({:{}, _, args} = quoted, state) when is_list(args) do
+    normalize_literal(quoted, [], state)
+  end
+
   # Calls
   defp do_normalize({_, _, args} = quoted, state) when is_list(args) do
     normalize_call(quoted, state)
@@ -227,11 +232,29 @@ defmodule Code.Normalizer do
     end
   end
 
+  # Tuples
+  defp normalize_literal({:{}, _, args} = quoted, meta, state) do
+    last_arg = List.last(args)
+
+    with [{{:__block__, key_meta, _}, _} | _] <- last_arg, :keyword <- key_meta[:format] do
+      {:__block__, meta, [quoted]}
+    else
+      _ ->
+        normalize_call(quoted, state)
+    end
+  end
+
   # 2-tuples
   defp normalize_literal({left, right}, meta, state) do
     meta = patch_meta_line(meta, state.parent_meta)
     state = %{state | parent_meta: meta}
-    {:__block__, meta, [{do_normalize(left, state), do_normalize(right, state)}]}
+
+    with [{{:__block__, key_meta, _}, _} | _] <- right, :keyword <- key_meta[:format] do
+      {:__block__, meta, [{do_normalize(left, state), right}]}
+    else
+      _ ->
+        {:__block__, meta, [{do_normalize(left, state), do_normalize(right, state)}]}
+    end
   end
 
   # Lists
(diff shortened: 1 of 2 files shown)


lib/elixir/lib/code/normalizer.ex
```diff
@@ -227,11 +232,29 @@ defmodule Code.Normalizer do
     end
   end
 
+  # Tuples
+  defp normalize_literal({:{}, _, args} = quoted, meta, state) do
+    last_arg = List.last(args)
+
+    with [{{:__block__, key_meta, _}, _} | _] <- last_arg, :keyword <- key_meta[:format] do
+      {:__block__, meta, [quoted]}
```
josevalim, line 240: We still need to normalize the contents inside `quoted`, no?

lib/elixir/lib/code/normalizer.ex
```diff
@@ -227,11 +232,29 @@ defmodule Code.Normalizer do
...
+  end
+
   # 2-tuples
   defp normalize_literal({left, right}, meta, state) do
     meta = patch_meta_line(meta, state.parent_meta)
     state = %{state | parent_meta: meta}
-    {:__block__, meta, [{do_normalize(left, state), do_normalize(right, state)}]}
+
+    with [{{:__block__, key_meta, _}, _} | _] <- right, :keyword <- key_meta[:format] do
+      {:__block__, meta, [{do_normalize(left, state), right}]}
```
josevalim, line 253: We still need to normalize the right, no? For example, we need to normalize the values of the keyword.

doorgan:
Thanks for the comments @josevalim!

You're correct! I just addressed that. I also removed the indirection from `do_normalize({:{}, _, _}, ...) ...` to `normalize_literal`, since 3+ tuples aren't literals in the ast.

lib/elixir/lib/code/normalizer.ex
```diff
@@ -182,6 +182,20 @@ defmodule Code.Normalizer do
...
   end
 
+  # Tuples
+  defp do_normalize({:{}, meta, args} = quoted, state) do
+    {last_arg, args} = List.pop_at(args, -1)
+
+    with [{{:__block__, key_meta, _}, _} | _] <- last_arg, :keyword <- key_meta[:format] do
+      args = normalize_kw_args(args, state)
+      kw_list = normalize_kw_args(last_arg, state)
+      {:__block__, [], [{:{}, meta, args ++ kw_list}]}
```
josevalim, line 192: Do we need the block here? It is not a literal… but I am not sure.

lib/elixir/lib/code/normalizer.ex
```diff
@@ -182,6 +182,20 @@ defmodule Code.Normalizer do
...
+    else
+      _ ->
+        normalize_call(quoted, state)
```
josevalim, line 195: Actually, shouldn’t normalize call already handle keywords?

josevalim:
Btw, do we handle the same in lists accordingly?

lib/elixir/lib/code/normalizer.ex
doorgan: Actually, no, I'll remove that. Literal 2-tuples do need a block, as that's how they're encoded by the literal encoder, but 3+ tuples don't need it.

doorgan:
@josevalim we are since #11084 
We have a test case for the handling of non-normalized quoted expressions, I can add a test to ensure it also preserves the user choice for already normalized expressions.

josevalim:
No need for additional ones, the previous PR handles it. Let’s address the remaining comments and we are good!

lib/elixir/lib/code/normalizer.ex
doorgan: It does handle them, but it won't process `:{}` forms the way we are doing it here due to `allow_keyword?` excluding it. We'd have to add an additional clause to the `cond` in line 314 with the same code we have here.

josevalim:

PR #11355, opened 2021-11-01
Add :all back to typespec for IO.read

kipcole9:
Elixir 1.13 replaces the flag :all with :eof although
:all remains valid (albeit marked for deprecation).

However removing :all from the typespec means that
dialyzer errors on Elixir 1.13, and if replaced with :eof
dialyzer errors on earlier releases.

This commit aligns the typespec with the implementation. However
an additional comment in the @doc could be used to note
that :all should be replaced with :eof.

### DIFF
--- a/lib/elixir/lib/io.ex
+++ b/lib/elixir/lib/io.ex
@@ -142,7 +142,7 @@ defmodule IO do
       NFS volume
 
   """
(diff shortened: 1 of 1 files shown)


josevalim:
We generally don't keep deprecations options in the typespec, as that is also way to signal users. The best is for you to handle it in your code accordingly. :)

PR #11359, opened 2021-11-02
Add note in Float.to_{string/charlist}/1

v0idpwn:
Makes it clear where to look for if you're looking to use something
other than the algorithm used there.

### DIFF
--- a/lib/elixir/lib/float.ex
+++ b/lib/elixir/lib/float.ex
@@ -512,10 +512,14 @@ defmodule Float do
   Returns a charlist which corresponds to the smallest text representation
   of the given float.
 
-  The underlying algorithm changes depending on the OTP version:
-    * For OTP >= 24, it uses the [Ryū algorithm][1]
-    * For OTP < 24, it uses the algorithm presented it "Printing Floating-Point
-      Numbers Quickly and Accurately" in Proceedings of the SIGPLAN '96
+  The underlying algorithm changes depending on the Erlang/OTP version:
+
+    * For OTP >= 24, it uses the algorithm presented in "Ryū: fast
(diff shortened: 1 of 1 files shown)


v0idpwn:
Also, I believe the statement above will be imprecise starting from OTP 25. Shall we fix it already?

josevalim:
The statement is false from OTP 24, so amending it will be welcome. Here is the paper: https://dl.acm.org/citation.cfm?doid=3296979.3192369

v0idpwn:
Though it was merged to master a while ago, it doesn't seem like it was released.

josevalim:
OTP 24 has the algorithm implemented in Erlang, which is used by io_lib. 25 brings the algorithm to C.

lib/elixir/lib/float.ex
```diff
@@ -509,13 +509,20 @@ defmodule Float do
   defp sign(1, num), do: -num
 
   @doc """
-  Returns a charlist which corresponds to the text representation
+  Returns a charlist which corresponds to the smallest text representation
```
josevalim, line 512:
```suggestion
  Returns a charlist which corresponds to the shortest text representation
```

lib/elixir/lib/float.ex
```diff
@@ -529,13 +536,20 @@ defmodule Float do
   end
 
   @doc """
-  Returns a binary which corresponds to the text representation
+  Returns a binary which corresponds to the smallest text representation
```
josevalim, line 539:
```suggestion
  Returns a binary which corresponds to the shortest text representation
```

josevalim:

PR #11360, opened 2021-11-03
Call out what happens to duplicate keys in Map.map/2 in the docs

s3cur3:
Per the discussion in #11239, this adds a note that if the function you pass to `Map.map/2` produces duplicate keys, only the last one will be kept.


josevalim:

josevalim:
Good find @garazdawi!

s3cur3:
Oh geez, I totally misunderstood. 🤦‍♂️

PR #11370, opened 2021-11-05
Support configurable formatter plugins

maartenvanvliet:
This allows libraries to provide formatter plugins whilst
allowing the user to configure them:

E.g. `plugins: [{CssPlugin, extensions: [".css", ".scss"]}]`.

It is up to the plugin if it supports any custom configuration.

### DIFF
--- a/lib/mix/lib/mix/tasks/format.ex
+++ b/lib/mix/lib/mix/tasks/format.ex
@@ -251,7 +251,7 @@ defmodule Mix.Tasks.Format do
   defp eval_deps_and_subdirectories(dot_formatter, prefix, formatter_opts, sources) do
     deps = Keyword.get(formatter_opts, :import_deps, [])
     subs = Keyword.get(formatter_opts, :subdirectories, [])
-    plugins = Keyword.get(formatter_opts, :plugins, []) |> Enum.map(&normalize_plugin(&1))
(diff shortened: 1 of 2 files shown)


lib/mix/lib/mix/tasks/format.ex
```diff
@@ -251,7 +251,7 @@ defmodule Mix.Tasks.Format do
   defp eval_deps_and_subdirectories(dot_formatter, prefix, formatter_opts, sources) do
     deps = Keyword.get(formatter_opts, :import_deps, [])
     subs = Keyword.get(formatter_opts, :subdirectories, [])
-    plugins = Keyword.get(formatter_opts, :plugins, [])
+    plugins = Keyword.get(formatter_opts, :plugins, []) |> Enum.map(&normalize_plugin(&1))
```
josevalim, line 254:
```suggestion
    plugins = Keyword.get(formatter_opts, :plugins, []) |> Enum.map(&normalize_plugin/1)
```

josevalim:
At the moment, plugins are meant to be configured at the root:

```elixir
[
  plugins: [CSSFormatter],
  css_extensions: ~w(.css .scss)
]
```

I don't think we should have both. Your approach is nicer because it is nested, but also more verbose. You may also want to access other formatter options at the root too, so maybe we should keep it as is?

maartenvanvliet:
Yes makes sense! I'll close this PR

PR #11374, opened 2021-11-08
Add `Mix.installed?/0`

wojtekmach:


lib/mix/lib/mix.ex
```diff
@@ -747,6 +747,14 @@ defmodule Mix do
     end
   end
 
+  @doc """
+  Returns whether Mix.install/2 was called in the current VM.
```
josevalim, line 751:
```suggestion
  Returns whether `Mix.install/2` was called in the current node.
```

lib/mix/lib/mix.ex
```diff
@@ -747,6 +747,14 @@ defmodule Mix do
...
+  """
+  @doc since: "1.13.0"
+  def installed? do
+    !!Mix.State.get(:installed)
```
josevalim, line 755:
```suggestion
    Mix.State.get(:installed) != nil
```

?

josevalim:

PR #11377, opened 2021-11-09
Have `mix xref --label compile-connected` ignore compile-time dependency chains

marcandre:
This way, only compile-time dependencies that lead to additional
transitive dependencies (i.e. via non-compile-time dependencies) are shown.

### DIFF
--- a/lib/mix/lib/mix/tasks/xref.ex
+++ b/lib/mix/lib/mix/tasks/xref.ex
@@ -762,11 +762,16 @@ defmodule Mix.Tasks.Xref do
 
   defp filter_fn(file_references, :compile_connected),
     do: fn {key, type} ->
-      type == :compile and match?([_ | _], file_references[key] || [])
+      type == :compile and has_non_compile_time_dependencies?(file_references[key])
     end
 
   defp filter_fn(_file_references, filter),
     do: fn {_key, type} -> type == filter end
+
+  defp has_non_compile_time_dependencies?(nil), do: false
+
+  defp has_non_compile_time_dependencies?(reference_list),
+    do: Enum.any?(reference_list, fn {_key, type} -> type != :compile end)
 
   defp filter(file_references, :all), do: file_references
 
--- a/lib/mix/test/mix/tasks/xref_test.exs
+++ b/lib/mix/test/mix/tasks/xref_test.exs
@@ -514,6 +514,24 @@ defmodule Mix.Tasks.XrefTest do
       """)
     end
 
+    @abcd_compile_time_files %{
+      "lib/a.ex" => "B.b()\ndefmodule A, do: def a, do: 42",
+      "lib/b.ex" => "C.c()\ndefmodule B, do: def b, do: 42",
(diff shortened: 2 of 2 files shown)


marcandre:
This occurred to me while reading the nice summary:

> `mix xref graph` now supports `--label` to be set to “compile-connected”, which returns all compile-time dependencies that lead to additional transitive dependencies.

This description is what we want, but current behavior lists any connection, even to another compile-time dependency.

I should have realized this earlier too, as in our codebase, we have a few compile-time-only chains and I was dealing with this manually with `--exclude`, whereby I would exclude the few end nodes I knew of and then making an extra check that `MyApp.GetText` didn't have any depency:

```
mix xref graph --label=compile-connected \
--exclude=lib/bobby.ex \
--exclude=lib/bobby_web/gettext.ex \
--exclude=lib/bobby/utilities/enum.ex \
--fail-above 0 &&
mix xref graph \
--source=lib/bobby.ex \
--source=lib/bobby_web/gettext.ex \
--source=lib/bobby/utilities/enum.ex \
--fail-above 0
```

I believe that with this change, I will not actually be able to run the straight command without any frills

josevalim:
Hrm, I am not sure. A compile time dependency is a transitive dependency too! So I'd think the current behaviour is the correct one?

marcandre:
> Hrm, I am not sure. A compile time dependency is a transitive dependency too! 

You are correct. Maybe we should have an extra label?

I'll try to articulate better why I think the current definition is not as useful as the one in this PR (later though, takeoff time)

marcandre:
The main point I want to bring is that the type of dependency that we want to find, in practice, is exactly the kind where `A` has a compile-time dependency on `B` and `B` has a non-compile-time dependency on `C`.

Maybe `--label compile-tainting` (or `compile-convert`) could be a decent name for this, as the `A=>B` dependency "taints" the `B=>C` dependency, as well as any (non-compile-time) dependency that `C` may have, "tainting" them from whatever type they are to a compile-time one. Sorry, I can't think of better names.

I feel that in practice we should only care about these cases, and never about compile-time dependency chains, so I would suggest we only introduce one kind of compound check. Alternative, we can have two, but I fail to see why anyone would use `compile-connected` if `compile-tainting` exists (or for example `export-connected`, if we added that).

josevalim:
My suggestion for now would then be for us to collect a bit more feedback, we have plenty of goodies in v1.13 (thank you!) so let's get some mileage on them before moving forward. WDYT?

marcandre:
Thank you for your patience and the feedback. You are right, and my proposal doesn't actually resolve what I wanted and requires more thought.

I checked with this PR and it doesn't fix all in our codebase, because `gettext` is an export dependency. This makes me realize that there's probably no choice than having an explicit list of "mini-chains" that are allowed but known to lead to dead ends.

Makes me think of a different approach of a list of nodes to contract. This would allow a cleaner way to achieve the right result in a single `mix xref graph` command instead of two as in my example above.

Closing this.

PR #11381, opened 2021-11-11
Adds support for Enum.sum/2

tiagonbotelho:
This PR adds a mapper function argument to the `sum` function argument.

There are at least two main benefits with this addition:
1. `sum` becomes available to a wider range of enumerables (e.g. maps)
2. Added expliciteness without requiring an Enum.reduce/3 or Enum.map/2 |> Enum.sum/1 to be done

### DIFF
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -3217,19 +3217,30 @@ defmodule Enum do
       iex> Enum.sum(1..10//2)
       25
 
-  """
-  @spec sum(t) :: number
-  def sum(enumerable)
+      iex> Enum.sum(1..10, & &1 * 2)
+      209
 
-  def sum(first..last//step = range) do
-    range
+      iex> Enum.sum([%{num: 1}, %{num: 2}], & &1.num)
+      3
+
+      iex> Enum.sum(%{num_2: 2, num_3: 3}, fn {_, v} -> v end)
+      5
+  """
+  @spec sum(t, (element -> number)) :: number
+  def sum(enumerable, fun \\ fn item -> item end)
+
+  def sum(first..last//step, fun) do
+    first = fun.(first)
+    last = fun.(last)
+
+    first..last//step
     |> Range.size()
     |> Kernel.*(first + last - rem(last - first, step))
     |> div(2)
   end
 
-  def sum(enumerable) do
-    reduce(enumerable, 0, &+/2)
+  def sum(enumerable, fun) do
+    reduce(enumerable, 0, &(fun.(&1) + &2))
   end
 
   @doc """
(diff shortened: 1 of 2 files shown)


tiagonbotelho:
I probably should've opened a discussion in the mailing list but the change felt so small that it almost didn't seem worth it. If that'd be preferred I'm more than happy to create it 😊

lib/elixir/lib/enum.ex
```diff
@@ -3217,19 +3217,30 @@ defmodule Enum do
...
+
+      iex> Enum.sum(%{num_2: 2, num_3: 3}, fn {_, v} -> v end)
+      5
   """
-  @spec sum(t) :: number
-  def sum(enumerable)
+  @spec sum(t, (element -> number)) :: number
+  def sum(enumerable, fun \\ fn item -> item end)
+
+  def sum(first..last//step, fun) do
```
tiagonbotelho, line 3232: This is probably the bit I'm more concerned about, from my testing it seems like the approach of using the `fun` on the "edges" of the range and then _coercing_ it back into a range, seems like the way to go. But I'm open to feedback on this 🙌

josevalim:
Hi @tiagonbotelho! 👋 

As you noted, a custom sum can be achieved using reduce, as you noted, and the other aggregation function with arity 2 that we have, count/2, has the second function working as a filtering. So this would add some inconsistency and it can be easily achievable in other ways, for this reason I am :-1: on this change. :)

tiagonbotelho:
Ah @josevalim I wasn't aware of `count/2` that is a good point 👍 I'd still prefer to have a straightforward `Enum.sum/2` with an explicit mapper function (similar to what we have in [`Enum.sort_by/3`](https://hexdocs.pm/elixir/1.12/Enum.html#sort_by/3) rather than an `Enum.reduce/3` which requires a bit of parsing in order to understand what is really going on.

It is just a matter of preference though, if we are confident this would lead to some inconsistency in the public API and don't feel like this is a strong enough reason to move forward with this change, I'm more than happy to drop it 😊

josevalim:
Right, sum_by may make more sense then but I am not strongly in favor of the function.

tiagonbotelho:
> Right, sum_by may make more sense then but I am not strongly in favor of the function.

Agreed 👍 , as I mentioned this might come down to a matter of preference, so I'm happy to close the PR if we think this shouldn't be added to the `Enum` public API

lib/elixir/lib/enum.ex
josevalim: This will only work if the function is linear, which we cannot guarantee. For example, &:math.sin/1 wouldn’t work as callback.

josevalim:
Hi @tiagonbotelho, I am closing the PR due to the points above. Feel free to open up a discussion in the mailing list if you want a more general discussion!

PR #11383, opened 2021-11-11
Allow Application.compile_env* to accept attributes

eksperimental:
In Elixir v1.12.0 it is giving the following error
when calling `Application.compile_env(@app, :key)`

    ** (FunctionClauseError) no function clause matching in Application."MACRO-compile_env"/4

### DIFF
--- a/lib/mix/test/mix/tasks/archive_test.exs
+++ b/lib/mix/test/mix/tasks/archive_test.exs
@@ -149,9 +149,6 @@ defmodule Mix.Tasks.ArchiveTest do
       true = Application.compile_env!(:git_repo, :archive_config)
 
       defmodule GitRepo.Archive do
-        @app :git_repo
-        true = Application.compile_env!(@app, :archive_config)
-
         def hello do
           "World"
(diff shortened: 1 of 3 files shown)


eksperimental:
Please apply to v1.13 and v1.12 branches.
Since we are deprecating Application.get_env/3 and friends, and Application.compile_env should be favored, I think it should be backported to v1.12 to promote wider adoption.

lib/mix/test/mix/tasks/archive_test.exs
```diff
@@ -149,6 +149,9 @@ defmodule Mix.Tasks.ArchiveTest do
       true = Application.compile_env!(:git_repo, :archive_config)
 
       defmodule GitRepo.Archive do
+        @app :git_repo
+        true = Application.compile_env!(@app, :archive_config)
```
josevalim, line 153: Can you please add a test to `test/elixir/application_test.exs` instead? This test is not really related to the functionality.

josevalim:

PR #11384, opened 2021-11-12
Corrections, clarifications, and copyediting to Task.Supervisor docs

s3cur3:
This fixes a few issues I ran into while digging into the `Task.Supervisor` documentation for the first time.

Most of it is pretty minor, but there are maybe two things worth calling out:

1. In 2017, 6d1deb2032d1a9b21877a6757033de1bc88987a0 migrated `Task.Supervisor` to use `DynamicSupervisor`. In doing so, it added an optional keyword list to many of the functions, but not all of the docs got updated to note the correct number of arguments on those functions.
2. The `start_link/1` docs indicated that it was possible for `async` to receive `:restart` options, which as best I can tell hasn't been true at least since that 2017 rewrite.


josevalim:

Issue #11385, opened 2021-11-12
When remsh exits its related IEx.Evaluator doesn't stop


RudolfMan:
### Environment

* Elixir & Erlang/OTP versions (elixir --version): Elixir 1.12.3 (compiled with Erlang/OTP 24)
* Operating system: observed on MacOS 11.6 BigSur as well as alpine:3.13

### Current behavior

When we quit `remsh` its IEx.Evaluator process on the main node keeps hanging "orphaned". New remsh connection just start another IEx.Evaluator.

How to reproduce:
1.  run a node `iex --name a@127.0.0.1`
2. attach remote shell and `iex --name b@127.0.0.1 --remsh a`
3. print IEx.Evaluator pid **of remsh** 
    ```elixir
    iex(a@127.0.0.1)1> self()
    #PID<0.125.0>
    ```
4. quit remsh with `Ctrl+\` (or 2 times `Ctrl+C`)
5. in the running node check if process is alive
    ```elixir
    iex(a@127.0.0.1)1> p = pid("0.125.0")
    #PID<0.125.0>
    iex(a@127.0.0.1)2> Process.alive?(p)
    true
    iex(a@127.0.0.1)3> :proc_lib.translate_initial_call(p)
    {IEx.Evaluator, :init, 4}
    ```
6. furthermore, just in case, we could check that IEx.Server that was monitoring IEx.Evaluator is dead
    ```elixir
    iex(a@127.0.0.1)4> {:dictionary, dict} = Process.info(p, :dictionary) 
    iex(a@127.0.0.1)5> Process.alive?(dict[:iex_server])
    false
    ```
 Also can be observed in Processes tab of `:observer` when multiple time start and exit remote shell, processes with `Elixir.IEx.Evaluator:loop/1` keep adding one at a time and never delete unlike `Elixir.IEx.Server:wait_input/4`

<img src="https://user-images.githubusercontent.com/53276677/141419093-30902811-1744-4292-9c05-5447fdc388c5.png" width="500" />

### Expected behavior
IEx.Evaluator stops along with its IEx.Server when user exits remsh, or at least is somehow gets monitored and not just keep hanging.

PR #11386, opened 2021-11-12
Remsh: stop IEx.Evaluator before exit IEx.Server

RudolfMan:
Stopping evaluator before exit IEx.Server seems to be fixing the issue https://github.com/elixir-lang/elixir/issues/11385
However, I might be missing some other use cases.


josevalim:

PR #11387, opened 2021-11-12
Add the option of using a custom delimiter for EEx tags

alopezz:
I have a use case where it would be very convenient to be able to customize at least to a certain extent the delimiting characters for EEx tags. So that, for example:

```elixir

iex> EEx.eval_string("foo <$= bar $>", [bar: "baz"], delimiter: ?$)
"foo baz"

iex> EEx.eval_string("<%= foo %> <$= bar $>", [bar: "baz"], delimiter: ?$)
"<%= foo %> baz"
```

I realized that the specific characters in the `<% %>` syntax are hardcoded in EEx's tokenizer, so I decided to try to change it to add this customization option; hence this PR.

Changing the tokenizer's code was simple and clean enough. The more involved part was actually to ensure that the error messages are consistent with the customized delimiter.

### DIFF
--- a/lib/eex/lib/eex.ex
+++ b/lib/eex/lib/eex.ex
@@ -48,6 +48,7 @@ defmodule EEx do
     * `:trim` - if `true`, trims whitespace left and right of quotation as
       long as at least one newline is present. All subsequent newlines and
       spaces are removed but one newline is retained. Defaults to `false`.
+    * `:delimiter` - a character to be used instead of `%` for delimiting EEx tags.
     * `:parser_options` - (since: 1.13.0) allow customizing the parsed code that is generated.
       See `Code.string_to_quoted/2` for available options. Note that the options
       `:file`, `:line` and `:column` are ignored if passed in.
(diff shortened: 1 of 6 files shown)


josevalim:
Can you please expand on the use case? My biggest concern with allowing this is that we will no longer be able to recognize what EEx is, especially if it starts differing between applications and engines.

alopezz:
I admit it may be a bit of a niche use case, and I would understand it not being enough to justify such a change).

What I'm trying to do is to generate EEx files from templates which are written in EEx themselves.

The motivation behind this is that I'm working on a project generator (something similar to [cookiecutter](https://github.com/cookiecutter/cookiecutter), where the files for a new project are generated by rendering files written in EEx. This works well in general, but if a user wants to generate a file which is intended to contain EEx tags after rendering, then they have to be very careful about quoting those (with `<%% %>`), which is a suboptimal solution.

I can probably come up with some other workaround solution that works for my project (without having to change `EEx` itself), but the cleanest solution that came to my mind was the one proposed in this PR.

josevalim:
Hi @alopezz, given escaping is a possible solution (although I understand it is not an ideal one), I would prefer to not merge this for now, given the long term impact it can have. :) Thanks!

PR #11394, opened 2021-11-16
Update `~E` docs

wojtekmach:


lib/eex/lib/eex.ex
```diff
@@ -120,6 +120,7 @@ defmodule EEx do
       "Hello world\n"
 
   '''
+  @doc since: "1.14.0"
```
wojtekmach, line 123:
```suggestion
  @doc since: "1.13.0"
```

? :)

josevalim:

PR #11395, opened 2021-11-16
Add note to `Task.Supervisor.children/1`

RudolfMan:
Clarify that `Task.Supervisor.children/1` ignores children that are restarting.

Plus note about calling the function under low memory condition, (same that applies to `DynamicSupervisor.which_children/1`)


josevalim:

PR #11397, opened 2021-11-17
Fix normalizer not preserving user choice on module attribute lists

### LINKED ISSUE #11393: Code formatted with `Code.quoted_to_algebra` does not preserve user choice on module attribute kw list
### Environment

  Erlang/OTP 23 [erts-11.1.1] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1] [hipe]

  Elixir 1.14.0-dev (39121ad) (compiled with Erlang/OTP 22)
* Operating system: Linux

### Current behavior

Parsing a string with a module attribute and a keyword list ignores the user preference on brackets:
```elixir
iex> opts = [literal_encoder: &{:ok, {:__block__, &2, [&1]}}, token_metadata: true]
iex> {:ok, quoted} = Code.string_to_quoted("""
...> @foo [
...>   a: 1,
...>   b: 2
...> ]
...> """, opts)
iex> quoted |> Code.quoted_to_algebra() |> Inspect.Algebra.format(80) |> IO.puts()
@foo a: 1,
     b: 2
:ok
```
([Reproduction gist](https://gist.github.com/doorgan/d2ca852802781e1a97ae8399674d3a19))

### Expected behavior

It should preserve the user choice on square brackets like it does with the rest of the code:
```elixir
iex> opts = [literal_encoder: &{:ok, {:__block__, &2, [&1]}}, token_metadata: true]
iex> {:ok, quoted} = Code.string_to_quoted("""
...> @foo [
...>   a: 1,
...>   b: 2
...> ]
...> """, opts)
iex> quoted |> Code.quoted_to_algebra() |> Inspect.Algebra.format(80) |> IO.puts()
@foo [
  a: 1,
  b: 2
]
:ok
```

---

I was going to prepare a PR for this one, but I'm not sure if the issue here is with the normalizer or the formatter. As far as I can tell, the only difference is that the kw list with brackets has `:closing` metadata. Whether or not this metadata is present, the formatter seems to only look at the `format: :keyword` metadata from the elements.

doorgan:
This fixes #11393 

It turns out this was another normalizer bug.

The AST for a module attribute is
```elixir
{:@, attr_meta, [
  {:name, name_meta, [value]}
]}
```
`value` here is any quoted expression.

This was handled as 2 nested calls: one pass for `:@` and one for `:name`. The second pass for `:name` won't know that it's part of a module attribute definition, so it will be handled as a function call. The formatter removes brackets whenever possible on function call, which caused the bug.

I figured adding a new clause would be more predictable than adding more fields to the `state` that we then need to clean up.

### DIFF
--- a/lib/elixir/lib/code/normalizer.ex
+++ b/lib/elixir/lib/code/normalizer.ex
@@ -197,6 +197,18 @@ defmodule Code.Normalizer do
     end
   end
 
+  # Module attributes
+  defp do_normalize({:@, meta, [{name, name_meta, [value]}]}, state) do
+    value =
+      if is_list(value) do
+        normalize_kw_args(value, state, false)
+      else
+        do_normalize(value, state)
+      end
+
+    {:@, meta, [{name, name_meta, [value]}]}
+  end
+
   # Calls
   defp do_normalize({_, _, args} = quoted, state) when is_list(args) do
     normalize_call(quoted, state)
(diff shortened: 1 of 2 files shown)


josevalim:

PR #11403, opened 2021-11-19
Improve user experience when there is a faulty Inspect implementation

eksperimental:
When a faulty implementation of the Inspect protocol is used,
the user is returned an `Inspect.Error` struct when the user tries to inspect it,
also IEx prints `Inspect.Error` upon inspection or when a struct with a fault inspect is returned.

This is the default behaviour in Elixir, because it uses the :safe  option in Inspect.Opts is to true by default.
Since Kernel.inspect/2 will return a string, an no warning is emitted, it is very likely that an error like this can go unnoticed.

This PR tries to make it as clear as possible for the user, by introducing the :stacktrace field in the Inspect.Error exception,
and by emitting a runtime warning.

Given the following faulty implementation:

```elixir
defmodule Failing do
  @enforce_keys [:name]
  defstruct @enforce_keys

  defimpl Inspect do
    def inspect(%Failing{name: name}, _) do
      Atom.to_string(name)
    end
  end
end
```

These are the examples of how it behaved before and how it does with this PR.

## Case 1) Evaluate the struct in IEx

Elixir v1.13.0-rc.1 will return:

```elixir
iex(1)> struct = %Failing{name: "Foo"}
%Inspect.Error{
  message: "got ArgumentError with message \"errors were found at the given arguments:\\n\\n  * 1st argument: not an atom\\n\" while inspecting %{__struct__: Failing, name: \"Foo\"}"
}
```

While with this PR, it will warn and return a stacktrace along with the warning and the returned `Inspect.Error` struct

```elixir
iex(1)> struct = %Failing{name: "Foo"}
warning: error when trying to inspect struct; got ArgumentError with message "errors were found at the given arguments:\n\n  * 1st argument: not an atom\n" while inspecting %{__struct__: Failing, name: "Foo"}
  :erlang.atom_to_binary("Foo", :utf8)
  (debug_inspect_error_bug 0.1.0) lib/failing.ex:7: Inspect.Failing.inspect/2
  (elixir 1.14.0-dev) lib/inspect/algebra.ex:344: Inspect.Algebra.to_doc/2
  (elixir 1.14.0-dev) lib/kernel.ex:2246: Kernel.inspect/2
  (iex 1.14.0-dev) lib/iex/evaluator.ex:336: IEx.Evaluator.io_inspect/1
  (iex 1.14.0-dev) lib/iex/evaluator.ex:313: IEx.Evaluator.handle_eval/3
  (iex 1.14.0-dev) lib/iex/evaluator.ex:285: IEx.Evaluator.do_eval/3
  (iex 1.14.0-dev) lib/iex/evaluator.ex:274: IEx.Evaluator.eval/3

%Inspect.Error{
  message: "got ArgumentError with message \"errors were found at the given arguments:\\n\\n  * 1st argument: not an atom\\n\" while inspecting %{__struct__: Failing, name: \"Foo\"}",
  stacktrace: [
    {:erlang, :atom_to_binary, ["Foo", :utf8],
     [error_info: %{module: :erl_erts_errors}]},
    {Inspect.Failing, :inspect, 2, [file: 'lib/failing.ex', line: 7]},
    {Inspect.Algebra, :to_doc, 2, [file: 'lib/inspect/algebra.ex', line: 344]},
    {Kernel, :inspect, 2, [file: 'lib/kernel.ex', line: 2246]},
    {IEx.Evaluator, :io_inspect, 1, [file: 'lib/iex/evaluator.ex', line: 336]},
    {IEx.Evaluator, :handle_eval, 3, [file: 'lib/iex/evaluator.ex', line: 313]},
    {IEx.Evaluator, :do_eval, 3, [file: 'lib/iex/evaluator.ex', line: 285]},
    {IEx.Evaluator, :eval, 3, [file: 'lib/iex/evaluator.ex', line: 274]}
  ]
}
```

## Case 2) Inspect the struct with Kernel.inspect/2

The same happens if you safely inspect the struct

Elixir v1.13.0-rc.1:

```elixir
iex(2)> inspect(struct)
"%Inspect.Error{message: \"got ArgumentError with message \\\"errors were found at the given arguments:\\\\n\\\\n  * 1st argument: not an atom\\\\n\\\" while inspecting %{__struct__: Failing, name: \\\"Foo\\\"}\"}"
```

With this PR, it will warn and return a string with the stacktrace included.

```elixir
iex(2)> inspect(struct)
warning: error when trying to inspect struct; got ArgumentError with message "errors were found at the given arguments:\n\n  * 1st argument: not an atom\n" while inspecting %{__struct__: Failing, name: "Foo"}
  :erlang.atom_to_binary("Foo", :utf8)
  (debug_inspect_error_bug 0.1.0) lib/failing.ex:7: Inspect.Failing.inspect/2
  (elixir 1.14.0-dev) lib/inspect/algebra.ex:344: Inspect.Algebra.to_doc/2
  (elixir 1.14.0-dev) lib/kernel.ex:2246: Kernel.inspect/2
  (stdlib 3.16.1) erl_eval.erl:685: :erl_eval.do_apply/6
  (elixir 1.14.0-dev) src/elixir.erl:289: :elixir.recur_eval/3
  (elixir 1.14.0-dev) src/elixir.erl:274: :elixir.eval_forms/3
  (iex 1.14.0-dev) lib/iex/evaluator.ex:310: IEx.Evaluator.handle_eval/3

"%Inspect.Error{message: \"got ArgumentError with message \\\"errors were found at the given arguments:\\\\n\\\\n  * 1st argument: not an atom\\\\n\\\" while inspecting %{__struct__: Failing, name: \\\"Foo\\\"}\", stacktrace: [{:erlang, :atom_to_binary, [\"Foo\", :utf8], [error_info: %{module: :erl_erts_errors}]}, {Inspect.Failing, :inspect, 2, [file: 'lib/failing.ex', line: 7]}, {Inspect.Algebra, :to_doc, 2, [file: 'lib/inspect/algebra.ex', line: 344]}, {Kernel, :inspect, 2, [file: 'lib/kernel.ex', line: 2246]}, {:erl_eval, :do_apply, 6, [file: 'erl_eval.erl', line: 685]}, {:elixir, :recur_eval, 3, [file: 'src/elixir.erl', line: 289]}, {:elixir, :eval_forms, 3, [file: 'src/elixir.erl', line: 274]}, {IEx.Evaluator, :handle_eval, 3, [file: 'lib/iex/evaluator.ex', line: 310]}]}"
```

## Case 3) Passing a struct as an argument to a function that raises and inspects it in the error message

This is the worst case where the error message is cryptic and misleading.

```elixir
iex(3)> Enum.to_list(%Failing{name: "Foo"})
** (Protocol.UndefinedError) protocol Enumerable not implemented for %Inspect.Error{message: "got ArgumentError with message \"errors were found at the given arguments:\\n\\n  * 1st argument: not an atom\\n\" while inspecting %{__struct__: Failing, name: \"Foo\"}"} of type Failing (a struct). This protocol is implemented for the following type(s): Date.Range, File.Stream, Function, GenEvent.Stream, HashDict, HashSet, IO.Stream, List, Map, MapSet, Range, Stream
    (elixir 1.13.0-rc.1) lib/enum.ex:1: Enumerable.impl_for!/1
    (elixir 1.13.0-rc.1) lib/enum.ex:143: Enumerable.reduce/3
    (elixir 1.13.0-rc.1) lib/enum.ex:4143: Enum.reverse/1
    (elixir 1.13.0-rc.1) lib/enum.ex:3488: Enum.to_list/1
```

With this PR, a warning is emitted and inside the error message the stacktrace is included, hinting for easier tracing.

```elixir
iex(3)> Enum.to_list(%Failing{name: "Foo"})
warning: error when trying to inspect struct; got ArgumentError with message "errors were found at the given arguments:\n\n  * 1st argument: not an atom\n" while inspecting %{__struct__: Failing, name: "Foo"}
  :erlang.atom_to_binary("Foo", :utf8)
  (debug_inspect_error_bug 0.1.0) lib/failing.ex:7: Inspect.Failing.inspect/2
  (elixir 1.14.0-dev) lib/inspect/algebra.ex:344: Inspect.Algebra.to_doc/2
  (elixir 1.14.0-dev) lib/kernel.ex:2246: Kernel.inspect/2
  (elixir 1.14.0-dev) lib/exception.ex:1267: Protocol.UndefinedError.message/1
  (elixir 1.14.0-dev) lib/exception.ex:66: Exception.message/1
  (elixir 1.14.0-dev) lib/exception.ex:117: Exception.format_banner/3
  (iex 1.14.0-dev) lib/iex/evaluator.ex:357: IEx.Evaluator.print_error/3

** (Protocol.UndefinedError) protocol Enumerable not implemented for %Inspect.Error{message: "got ArgumentError with message \"errors were found at the given arguments:\\n\\n  * 1st argument: not an atom\\n\" while inspecting %{__struct__: Failing, name: \"Foo\"}", stacktrace: [{:erlang, :atom_to_binary, ["Foo", :utf8], [error_info: %{module: :erl_erts_errors}]}, {Inspect.Failing, :inspect, 2, [file: 'lib/failing.ex', line: 7]}, {Inspect.Algebra, :to_doc, 2, [file: 'lib/inspect/algebra.ex', line: 344]}, {Kernel, :inspect, 2, [file: 'lib/kernel.ex', line: 2246]}, {Protocol.UndefinedError, :message, 1, [file: 'lib/exception.ex', line: 1267]}, {Exception, :message, 1, [file: 'lib/exception.ex', line: 66]}, {Exception, :format_banner, 3, [file: 'lib/exception.ex', line: 117]}, {IEx.Evaluator, :print_error, 3, [file: 'lib/iex/evaluator.ex', line: 357]}]} of type Failing (a struct). This protocol is implemented for the following type(s): Date.Range, File.Stream, Function, GenEvent.Stream, HashDict, HashSet, IO.Stream, List, Map, MapSet, Range, Stream
    (elixir 1.14.0-dev) lib/enum.ex:1: Enumerable.impl_for!/1
    (elixir 1.14.0-dev) lib/enum.ex:143: Enumerable.reduce/3
    (elixir 1.14.0-dev) lib/enum.ex:4143: Enum.reverse/1
    (elixir 1.14.0-dev) lib/enum.ex:3488: Enum.to_list/1
```

## Case 4) No differences when inspecting with :safe option set to false.

Both versions behave the same.

```elixir
** (Inspect.Error) got ArgumentError with message "errors were found at the given arguments:\n\n  * 1st argument: not an atom\n" while inspecting %{__struct__: Failing, name: "Foo"}
    :erlang.atom_to_binary("Foo", :utf8)
    (debug_inspect_error_bug 0.1.0) lib/failing.ex:7: Inspect.Failing.inspect/2
    (elixir 1.13.0-rc.1) lib/inspect/algebra.ex:342: Inspect.Algebra.to_doc/2
    (elixir 1.13.0-rc.1) lib/kernel.ex:2246: Kernel.inspect/2
```

## Drawbacks

1. When returning the string representation of the struct, it could be too verbose with the new stacktrace, but this way it is guaranteed that the use gets all the information to inspect the issue.

2. Warning could be anyoying too verbose, so we could introduce a new option to Inspect.Opts called `:warn`, which defaults to `true'.

## Other concerns

I wonder if, unless the exist, other ways to cheaply test the correct implementation of the inspect protocol.
Currently we either need to try/rescue which is known to be expensive, or pattern match the beginning of the returned string against "%Inspect.Error{".

### DIFF
--- a/lib/elixir/lib/exception.ex
+++ b/lib/elixir/lib/exception.ex
@@ -1633,83 +1633,55 @@ defmodule ErlangError do
 end
 
 defmodule Inspect.Error do
+  import Inspect.Utils, only: [pad: 2]
+
   @moduledoc """
   Raised when a struct cannot be inspected.
   """
-  # QUESTION: I am not sure if is it considered backward incompatible
-  # to enforce keys that did not exist before
-  @enforce_keys [:exception, :stacktrace, :struct]
-  defexception exception: nil,
-               message: nil,
-               stacktrace: nil,
-               struct: nil
+  @enforce_keys [:exception, :exception_message, :stacktrace, :struct]
+  defexception @enforce_keys
 
   @impl true
   def exception(arguments) when is_list(arguments) do
-    with {:ok, exception} <- Keyword.fetch!(arguments, :exception) |> validate(:exception),
-         {:ok, stacktrace} <- Keyword.fetch!(arguments, :stacktrace) |> validate(:stacktrace),
-         {:ok, struct} <- Keyword.fetch!(arguments, :struct) |> validate(:struct),
-         {:ok, message} <- Keyword.get(arguments, :message) |> validate(:message) do
-      message =
-        if message do
-          message
-        else
-          message(%Inspect.Error{
-            exception: exception,
-            stacktrace: stacktrace,
-            struct: struct
-          })
-        end
+    exception = Keyword.fetch!(arguments, :exception)
 
-      %Inspect.Error{
-        exception: exception,
-        message: message,
-        stacktrace: stacktrace,
-        struct: struct
-      }
-    else
-      {:error, key, term} -> raise(ArgumentError, "invalid #{key}, got: " <> inspect(term))
-    end
+    exception_inspected =
+      exception |> Map.fetch!(:__struct__) |> Inspect.Atom.inspect(%Inspect.Opts{})
+
+    exception_message = Exception.message(exception) |> String.trim_trailing("\n")
+    stacktrace = Keyword.fetch!(arguments, :stacktrace)
+    true = is_list(stacktrace)
+    struct_formatted = Keyword.fetch!(arguments, :struct) |> format_struct()
+
+    %Inspect.Error{
+      exception: exception_inspected,
+      exception_message: exception_message,
+      stacktrace: stacktrace,
+      struct: struct_formatted
+    }
   end
-
-  defp validate(term, :exception) when is_exception(term), do: {:ok, term}
-  defp validate(term, :stacktrace) when is_list(term), do: {:ok, term}
-  defp validate(term, :struct) when is_struct(term), do: {:ok, term}
-  defp validate(term, :message) when is_binary(term) or is_nil(term), do: {:ok, term}
-  defp validate(term, key), do: {:error, key, term}
 
   @impl true
-  def message(%__MODULE__{message: message}) when is_binary(message) do
-    message
-  end
+  def message(%__MODULE__{
+        exception: exception,
+        exception_message: exception_message,
+        struct: struct
+      })
+      when is_binary(exception) and is_binary(exception_message) and is_binary(struct) do
+    ~s'''
+    got #{exception} with message:
 
-  def message(%__MODULE__{exception: exception, stacktrace: stacktrace, struct: struct})
-      when is_exception(exception) and is_list(stacktrace) and is_struct(struct) do
-    "got #{inspect(exception.__struct__)} with message:\n\n" <>
-      "    \"\"\"\n" <>
-      "#{pad(String.trim_trailing(Exception.message(exception), "\n"), 4)}\n" <>
-      "    \"\"\"\n\n" <>
-      "  while inspecting:\n" <>
-      pad(format_struct(struct), 4) <>
-      "\n"
+        """
+    #{pad(exception_message, 4)}
+        """
+
+      while inspecting:
+    #{pad(struct, 4)}
+    '''
   end
 
   # Helpers
-  def pad(message, padding_length) do
-    padding = String.duplicate(" ", padding_length)
-
-    result =
-      message
-      |> String.replace("\\n", "\n")
-      |> String.split("\n")
-      |> Enum.map(&(padding <> &1))
-      |> Enum.join("\n")
-      |> String.trim("\n")
-
-    result
-  end
-
-  def format_struct(struct) do
+  defp format_struct(struct) when is_map(struct) do
     opts = %Inspect.Opts{}
 
     struct
--- a/lib/elixir/lib/exception.ex
+++ b/lib/elixir/lib/exception.ex
@@ -1631,3 +1631,90 @@ defmodule ErlangError do
   defp nth(3), do: "3rd"
   defp nth(n), do: "#{n}th"
 end
+
+defmodule Inspect.Error do
+  @moduledoc """
+  Raised when a struct cannot be inspected.
+  """
+  # QUESTION: I am not sure if is it considered backward incompatible
+  # to enforce keys that did not exist before
+  @enforce_keys [:exception, :stacktrace, :struct]
+  defexception exception: nil,
+               message: nil,
+               stacktrace: nil,
+               struct: nil
+
+  @impl true
+  def exception(arguments) when is_list(arguments) do
+    with {:ok, exception} <- Keyword.fetch!(arguments, :exception) |> validate(:exception),
+         {:ok, stacktrace} <- Keyword.fetch!(arguments, :stacktrace) |> validate(:stacktrace),
+         {:ok, struct} <- Keyword.fetch!(arguments, :struct) |> validate(:struct),
+         {:ok, message} <- Keyword.get(arguments, :message) |> validate(:message) do
+      message =
+        if message do
+          message
+        else
+          message(%Inspect.Error{
+            exception: exception,
+            stacktrace: stacktrace,
+            struct: struct
+          })
+        end
+
+      %Inspect.Error{
+        exception: exception,
+        message: message,
+        stacktrace: stacktrace,
+        struct: struct
+      }
+    else
+      {:error, key, term} -> raise(ArgumentError, "invalid #{key}, got: " <> inspect(term))
+    end
+  end
+
+  defp validate(term, :exception) when is_exception(term), do: {:ok, term}
+  defp validate(term, :stacktrace) when is_list(term), do: {:ok, term}
+  defp validate(term, :struct) when is_struct(term), do: {:ok, term}
+  defp validate(term, :message) when is_binary(term) or is_nil(term), do: {:ok, term}
+  defp validate(term, key), do: {:error, key, term}
+
+  @impl true
+  def message(%__MODULE__{message: message}) when is_binary(message) do
+    message
+  end
+
+  def message(%__MODULE__{exception: exception, stacktrace: stacktrace, struct: struct})
+      when is_exception(exception) and is_list(stacktrace) and is_struct(struct) do
+    "got #{inspect(exception.__struct__)} with message:\n\n" <>
+      "    \"\"\"\n" <>
+      "#{pad(String.trim_trailing(Exception.message(exception), "\n"), 4)}\n" <>
+      "    \"\"\"\n\n" <>
+      "  while inspecting:\n" <>
+      pad(format_struct(struct), 4) <>
+      "\n"
+  end
+
+  # Helpers
+  def pad(message, padding_length) do
+    padding = String.duplicate(" ", padding_length)
+
+    result =
+      message
+      |> String.replace("\\n", "\n")
+      |> String.split("\n")
+      |> Enum.map(&(padding <> &1))
+      |> Enum.join("\n")
+      |> String.trim("\n")
+
+    result
+  end
+
+  def format_struct(struct) do
+    opts = %Inspect.Opts{}
+
+    struct
+    |> Inspect.Map.inspect(opts)
+    |> Inspect.Algebra.format(opts.width)
+    |> IO.iodata_to_binary()
+  end
+end
--- a/lib/elixir/lib/inspect/algebra.ex
+++ b/lib/elixir/lib/inspect/algebra.ex
@@ -53,7 +53,8 @@ defmodule Inspect.Opts do
     * `:safe` - when `false`, failures while inspecting structs will be raised
       as errors instead of being wrapped in the `Inspect.Error` exception. This
       is useful when debugging failures and crashes for custom inspect
-      implementations. Defaults to `true`.
+      implementations. When `true` it will emit a warning is an error inspecting
+      the term is found. Defaults to `true`.
 
     * `:structs` - when `false`, structs are not formatted by the inspect
       protocol, they are instead printed as maps. Defaults to `true`.
@@ -364,18 +365,32 @@ defmodule Inspect.Algebra do
               res = IO.iodata_to_binary(format(res, :infinity))
 
               message =
-                "got #{inspect(caught_exception.__struct__)} with message " <>
-                  "#{inspect(Exception.message(caught_exception))} while inspecting #{res}"
+                "got #{inspect(caught_exception.__struct__, safe: false)} with message " <>
+                  "#{inspect(Exception.message(caught_exception), safe: false)} while inspecting #{res}"
 
-              exception = Inspect.Error.exception(message: message)
+              error_exception = Inspect.Error.exception(message: message)
 
-              if opts.safe do
-                Inspect.inspect(exception, %{
+              if opts.safe && not is_struct(caught_exception, Inspect.Error) do
+                # Since we are not raising, and we are returning the string representation
+                # of the Inspect.Error exception, we warn.
+                IO.warn("error when trying to inspect struct; " <> message, __STACKTRACE__)
+
+                Inspect.inspect(error_exception, %{
                   opts
                   | inspect_fun: Inspect.Opts.default_inspect_fun()
                 })
               else
-                reraise(exception, __STACKTRACE__)
+                attributes =
+                  error_exception
+                  |> Map.drop([:__struct__, :__exception__])
+                  |> Map.put(
+                    :message,
+                    error_exception.message <>
+                      " Stacktrace:\n" <> Exception.format_stacktrace(__STACKTRACE__)
+                  )
+                  |> Map.to_list()
+
+                reraise(Inspect.Error, attributes, __STACKTRACE__)
               end
             after
               Process.delete(:inspect_trap)
--- a/lib/elixir/lib/inspect/algebra.ex
+++ b/lib/elixir/lib/inspect/algebra.ex
@@ -53,8 +53,7 @@ defmodule Inspect.Opts do
     * `:safe` - when `false`, failures while inspecting structs will be raised
       as errors instead of being wrapped in the `Inspect.Error` exception. This
       is useful when debugging failures and crashes for custom inspect
-      implementations. When `true` it will emit a warning is an error inspecting
-      the term is found. Defaults to `true`.
+      implementations. Defaults to `true`.
 
     * `:structs` - when `false`, structs are not formatted by the inspect
       protocol, they are instead printed as maps. Defaults to `true`.
@@ -162,14 +161,6 @@ defmodule Inspect.Opts do
   def default_inspect_fun(fun) when is_function(fun, 2) do
     :persistent_term.put({__MODULE__, :inspect_fun}, fun)
   end
-end
-
-defmodule Inspect.Error do
-  @moduledoc """
-  Raised when a struct cannot be inspected.
-  """
-  @enforce_keys [:message]
-  defexception message: nil, stacktrace: []
 end
 
 defmodule Inspect.Algebra do
@@ -356,27 +347,15 @@ defmodule Inspect.Algebra do
             try do
               Process.put(:inspect_trap, true)
 
-              res =
-                Inspect.Map.inspect(struct, %{
-                  opts
-                  | syntax_colors: [],
-                    inspect_fun: Inspect.Opts.default_inspect_fun()
-                })
-
-              res = IO.iodata_to_binary(format(res, :infinity))
-
-              message =
-                "got #{inspect(caught_exception.__struct__, safe: false)} with message " <>
-                  "#{inspect(Exception.message(caught_exception), safe: false)} while inspecting #{res}"
-
               error_exception =
-                Inspect.Error.exception(message: message, stacktrace: __STACKTRACE__)
+                Inspect.Error.exception(
+                  exception: caught_exception,
+                  inspect_opts: opts,
+                  stacktrace: __STACKTRACE__,
+                  struct: struct
+                )
 
               if opts.safe do
-                # Since we are not raising, and we are returning the string representation
-                # of the Inspect.Error exception, we warn.
-                IO.warn("error when trying to inspect struct; " <> message, __STACKTRACE__)
-
                 opts = %{opts | inspect_fun: Inspect.Opts.default_inspect_fun()}
                 Inspect.inspect(error_exception, opts)
               else
--- a/lib/elixir/lib/inspect/algebra.ex
+++ b/lib/elixir/lib/inspect/algebra.ex
@@ -168,7 +168,8 @@ defmodule Inspect.Error do
   @moduledoc """
   Raised when a struct cannot be inspected.
   """
-  defexception [:message]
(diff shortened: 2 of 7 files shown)


josevalim:
Thank you @eksperimental! Unfortunately `warn` is not an option because it can be problematic in production systems. For example, what happens if there is a bug and then we are writing to stdout all the time, slowing down the app?

So my suggestion is to do what you did for IEx but do it generally:

1. Include the stacktrace in the Inspect.Error input
2. Make sure the Inspect.Error is nicely formatted? Perhaps by setting pretty to true and some width?

There is an alternative for 2, which is to implement the Inspect protocol for Inspect.Error and make it show up with more information. Something like:

```elixir
#Inspect.Error<
  got ArgumentError with message "errors were found at the given arguments:\n\n  * 1st argument: not an atom\n" while inspecting %{__struct__: Failing, name: "Foo"}
      :erlang.atom_to_binary("Foo", :utf8)
      (debug_inspect_error_bug 0.1.0) lib/failing.ex:7: Inspect.Failing.inspect/2
      (elixir 1.13.0-rc.1) lib/inspect/algebra.ex:342: Inspect.Algebra.to_doc/2
      (elixir 1.13.0-rc.1) lib/kernel.ex:2246: Kernel.inspect/2
>
```

eksperimental:
I agree about the stderr issue. I sorted of sensed it since we do not emit runtime warnings, but it was an initial step.

> Include the stacktrace in the Inspect.Error input
What do you mean by "input" ?

I like the idea of formatting InspectError, but is that OK when that is inside another string like in this case:

```elixir
** (Protocol.UndefinedError) protocol Enumerable not implemented for %Inspect.Error{message: "got ArgumentError with message \"errors were found at the given arguments:\\n\\n  * 1st argument: not an atom\\n\" while inspecting %{__struct__: Failing, name: \"Foo\"}", stacktrace: [{:erlang, :atom_to_binary, ["Foo", :utf8], [error_info: %{module: :erl_erts_errors}]}, {Inspect.Failing, :inspect, 2, [file: 'lib/failing.ex', line: 7]}, {Inspect.Algebra, :to_doc, 2, [file: 'lib/inspect/algebra.ex', line: 344]}, {Kernel, :inspect, 2, [file: 'lib/kernel.ex', line: 2246]}, {Protocol.UndefinedError, :message, 1, [file: 'lib/exception.ex', line: 1267]}, {Exception, :message, 1, [file: 'lib/exception.ex', line: 66]}, {Exception, :format_banner, 3, [file: 'lib/exception.ex', line: 117]}, {IEx.Evaluator, :print_error, 3, [file: 'lib/iex/evaluator.ex', line: 357]}]} of type Failing (a struct). This protocol is implemented for the following type(s): Date.Range, File.Stream, Function, GenEvent.Stream, HashDict, HashSet, IO.Stream, List, Map, MapSet, Range, Stream
    (elixir 1.14.0-dev) lib/enum.ex:1: Enumerable.impl_for!/1
    (elixir 1.14.0-dev) lib/enum.ex:143: Enumerable.reduce/3
    (elixir 1.14.0-dev) lib/enum.ex:4143: Enum.reverse/1
    (elixir 1.14.0-dev) lib/enum.ex:3488: Enum.to_list/1
```

eksperimental:
I agree about the stderr issue. I sorted of sensed it since we do not emit runtime warnings, but it was an initial step.

> Include the stacktrace in the Inspect.Error input
What do you mean by "input" ?

I like the idea of formatting InspectError, but is that OK when that is inside another string like in this case:

```elixir
** (Protocol.UndefinedError) protocol Enumerable not implemented for %Inspect.Error{message: "got ArgumentError with message \"errors were found at the given arguments:\\n\\n  * 1st argument: not an atom\\n\" while inspecting %{__struct__: Failing, name: \"Foo\"}", stacktrace: [{:erlang, :atom_to_binary, ["Foo", :utf8], [error_info: %{module: :erl_erts_errors}]}, {Inspect.Failing, :inspect, 2, [file: 'lib/failing.ex', line: 7]}, {Inspect.Algebra, :to_doc, 2, [file: 'lib/inspect/algebra.ex', line: 344]}, {Kernel, :inspect, 2, [file: 'lib/kernel.ex', line: 2246]}, {Protocol.UndefinedError, :message, 1, [file: 'lib/exception.ex', line: 1267]}, {Exception, :message, 1, [file: 'lib/exception.ex', line: 66]}, {Exception, :format_banner, 3, [file: 'lib/exception.ex', line: 117]}, {IEx.Evaluator, :print_error, 3, [file: 'lib/iex/evaluator.ex', line: 357]}]} of type Failing (a struct). This protocol is implemented for the following type(s): Date.Range, File.Stream, Function, GenEvent.Stream, HashDict, HashSet, IO.Stream, List, Map, MapSet, Range, Stream
    (elixir 1.14.0-dev) lib/enum.ex:1: Enumerable.impl_for!/1
    (elixir 1.14.0-dev) lib/enum.ex:143: Enumerable.reduce/3
    (elixir 1.14.0-dev) lib/enum.ex:4143: Enum.reverse/1
    (elixir 1.14.0-dev) lib/enum.ex:3488: Enum.to_list/1
```

josevalim:
Yes, it would also be ok. Everywhere we inspect means we are expecting an inspected representation, which means it should be ok!

eksperimental:
I have applied the suggestions. I think the PR is pretty much done.

You can see the different scenarios.
https://gist.github.com/eksperimental/ba24eb2983d866ffc89d9fb2456b7958

The Error when it raises and what we print on screen when a faulty implementation is found is pretty much the same.

image

image


image


The only thing that is left as I don't know if it is desired is to make in IEx when we safely inspect a struct with a faulty implementation, to replace the \n with new lines when `structs: false`.

image

lib/elixir/lib/inspect/algebra.ex
```diff
@@ -163,13 +163,6 @@ defmodule Inspect.Opts do
   end
 end
 
-defmodule Inspect.Error do
-  @moduledoc """
-  Raised when a struct cannot be inspected.
-  """
-  defexception [:message]
-end
```
eksperimental, line 171: Moved it to exception.ex

lib/elixir/lib/inspect/algebra.ex
```diff
@@ -354,28 +347,24 @@ defmodule Inspect.Algebra do
...
-                    inspect_fun: Inspect.Opts.default_inspect_fun()
-                })
-
-              res = IO.iodata_to_binary(format(res, :infinity))
-
-              message =
-                "got #{inspect(caught_exception.__struct__)} with message " <>
-                  "#{inspect(Exception.message(caught_exception))} while inspecting #{res}"
-
-              exception = Inspect.Error.exception(message: message)
```
eksperimental, line 370: We were defining in `Algebra` the way the `Inspect.Error` message was formatted and its contents, so I moved it to exception.ex and added the needed fields to the exception to format this message, these being `:exception`, `:stacktrace` and `:struct`.

lib/elixir/lib/exception.ex
```diff
@@ -1631,3 +1631,90 @@ defmodule ErlangError do
...
+    "got #{inspect(exception.__struct__)} with message:\n\n" <>
+      "    \"\"\"\n" <>
+      "#{pad(String.trim_trailing(Exception.message(exception), "\n"), 4)}\n" <>
+      "    \"\"\"\n\n" <>
+      "  while inspecting:\n" <>
+      pad(format_struct(struct), 4) <>
+      "\n"
+  end
+
+  # Helpers
```
josevalim, line 1697: Please make the helpers private. :)

lib/elixir/lib/exception.ex
```diff
@@ -1631,3 +1631,90 @@ defmodule ErlangError do
...
+  def pad(message, padding_length) do
+    padding = String.duplicate(" ", padding_length)
+
+    result =
+      message
+      |> String.replace("\\n", "\n")
+      |> String.split("\n")
+      |> Enum.map(&(padding <> &1))
+      |> Enum.join("\n")
```
josevalim, line 1706:
```suggestion
      |> String.replace("\n", padding <> "\n")
```

lib/elixir/lib/exception.ex
```diff
@@ -1631,3 +1631,90 @@ defmodule ErlangError do
...
+      |> String.trim("\n")
+
+    result
```
josevalim, line 1709: You can remove this var :)

lib/elixir/lib/exception.ex
```diff
@@ -1631,3 +1631,90 @@ defmodule ErlangError do
...
+  def pad(message, padding_length) do
+    padding = String.duplicate(" ", padding_length)
+
+    result =
+      message
+      |> String.replace("\\n", "\n")
+      |> String.split("\n")
+      |> Enum.map(&(padding <> &1))
+      |> Enum.join("\n")
+      |> String.trim("\n")
```
josevalim, line 1707: If you trim the input then you likely won't need to trim here nor when calling `pad`.

lib/elixir/lib/exception.ex
```diff
@@ -1631,3 +1631,90 @@ defmodule ErlangError do
...
+
+  def message(%__MODULE__{exception: exception, stacktrace: stacktrace, struct: struct})
+      when is_exception(exception) and is_list(stacktrace) and is_struct(struct) do
+    "got #{inspect(exception.__struct__)} with message:\n\n" <>
+      "    \"\"\"\n" <>
+      "#{pad(String.trim_trailing(Exception.message(exception), "\n"), 4)}\n" <>
+      "    \"\"\"\n\n" <>
+      "  while inspecting:\n" <>
+      pad(format_struct(struct), 4) <>
+      "\n"
```
josevalim, line 1694: Maybe use heredocs for building this message. Use `~s'''` to avoid conflicts.

lib/elixir/lib/exception.ex
```diff
@@ -1631,3 +1631,90 @@ defmodule ErlangError do
   defp nth(3), do: "3rd"
   defp nth(n), do: "#{n}th"
 end
+
+defmodule Inspect.Error do
+  @moduledoc """
+  Raised when a struct cannot be inspected.
+  """
+  # QUESTION: I am not sure if is it considered backward incompatible
+  # to enforce keys that did not exist before
```
josevalim, line 1640: It is not a breaking change because Inspect.Error is never returned, it is only always printed. So you can remove the `:message` field.

lib/elixir/lib/exception.ex
```diff
@@ -1631,3 +1631,90 @@ defmodule ErlangError do
...
+               message: nil,
+               stacktrace: nil,
+               struct: nil
+
+  @impl true
+  def exception(arguments) when is_list(arguments) do
+    with {:ok, exception} <- Keyword.fetch!(arguments, :exception) |> validate(:exception),
+         {:ok, stacktrace} <- Keyword.fetch!(arguments, :stacktrace) |> validate(:stacktrace),
+         {:ok, struct} <- Keyword.fetch!(arguments, :struct) |> validate(:struct),
+         {:ok, message} <- Keyword.get(arguments, :message) |> validate(:message) do
```
josevalim, line 1652: There is no reason for those validations. We should just let it crash later on if invalid. We also no longer need to handle message.

lib/elixir/lib/inspect/algebra.ex
```diff
@@ -354,28 +347,24 @@ defmodule Inspect.Algebra do
...
+              error_exception =
+                Inspect.Error.exception(
+                  exception: caught_exception,
+                  inspect_opts: opts,
```
josevalim, line 353: This option is not used.

lib/elixir/lib/inspect/algebra.ex
```diff
@@ -354,28 +347,24 @@ defmodule Inspect.Algebra do
...
+                opts = %{opts | inspect_fun: Inspect.Opts.default_inspect_fun()}
+                Inspect.inspect(error_exception, opts)
               else
-                reraise(exception, __STACKTRACE__)
+                attributes =
+                  error_exception
+                  |> Map.drop([:__struct__, :__exception__])
+                  |> Map.to_list()
+
+                reraise(Inspect.Error, attributes, __STACKTRACE__)
```
josevalim, line 367: We should not change this behaviour. `safe` should always return the original exception.

lib/elixir/lib/exception.ex
```diff
@@ -1631,3 +1631,90 @@ defmodule ErlangError do
...
+
+  @impl true
+  def message(%__MODULE__{message: message}) when is_binary(message) do
+    message
+  end
+
+  def message(%__MODULE__{exception: exception, stacktrace: stacktrace, struct: struct})
+      when is_exception(exception) and is_list(stacktrace) and is_struct(struct) do
+    "got #{inspect(exception.__struct__)} with message:\n\n" <>
+      "    \"\"\"\n" <>
```
josevalim, line 1689: We don't need the heredocs here. We can assume that this is a string. :)

josevalim (commented):
Hi @eksperimental, this looks really great. I have dropped some comment that will allow us to simplify the code but there is one very important change you need to make.

You must call `format_struct` and `Exception.message(original_exception)` inside Inspect.Error.exception because, if we fail to format them, then it needs to be handled by the rescue/catch inside `Inspect`, otherwise we can build an `Inspect.Error` that will fail to be formatted. So the `Inspect.Error` should keep already formatted data instead of doing it later on.

lib/elixir/lib/exception.ex
eksperimental: I tried different options, and since this this is aimed at developers, I ended up making it look like docstring,,
make the idea to clearly show where the message starts and where it ends,
because later we continue with "while inspecting.... stacktrace" and it is not clear wheather that is part of the the excption mentioned, or part of InspectError

lib/elixir/lib/exception.ex
eksperimental: Ok. even simpler

lib/elixir/lib/inspect/algebra.ex
eksperimental: True. This is a left-over from previous experiments. Should we raise here?  we crash if the option is not valid, but these ones will go unnoticed.

lib/elixir/lib/exception.ex
josevalim: I see, point taken. :)

lib/elixir/lib/inspect/algebra.ex
josevalim: It is fine to not raise, it is all for internal use anyway. I would keep the code tidy rather than defensive. :)

lib/elixir/lib/exception.ex
eksperimental: You do,
```iex
iex(4)> "Foo\n" |> String.split("\n") |> Enum.join("\n")       
"Foo\n"
```
and in fact, many of those message ends in a new line.

lib/elixir/lib/inspect/algebra.ex
eksperimental: I am confused about this comment, because this section is for safe=false,
anyway the previous code we return and raise Inspect.Error.
`exception = Inspect.Error.exception(message: message)`

josevalim: Sorry, my comment was confusing. I am saying you can do this: `reraise(error_exception, __STACKTRACE__)`, you don't need to extract and rebuild it with `reraise/3`.

eksperimental:
@josevalim this is ready

lib/elixir/lib/exception.ex
eksperimental: @josevalim  it is really good optimization, but I cannot apply it because I had to cover a case for empty lines, to keep them empty.
If you want to have a look, maybe there is room for optimization.
https://github.com/elixir-lang/elixir/pull/11403/commits/712010d0d7a7db10eb71183127cfa0c17682d9c4#diff-0e2e75b45f813c505f2fa6643a3cddd54e5443a43d16014eac12bc498dfbd050R8-R16

lib/elixir/lib/exception.ex
```diff
@@ -1631,3 +1631,62 @@ defmodule ErlangError do
...
+      })
+      when is_binary(exception) and is_binary(exception_message) and is_binary(struct) do
+    ~s'''
+    got #{exception} with message:
+
+        """
+    #{pad(exception_message, 4)}
+        """
+
+      while inspecting:
```
josevalim, line 1678: Is the two space indentation here desired?

josevalim:

lib/elixir/lib/exception.ex
eksperimental: Yes.
But actually add 2 to "while inspecting" and the struct bellow will look better now that you mention it.

eksperimental:
> green_heart blue_heart purple_heart yellow_heart heart

Thank you for your input @josevalim

PR #11405, opened 2021-11-20
add Registry.count_select/2

RudolfMan:
PR is based on [this proposal](https://groups.google.com/g/elixir-lang-core/c/chazhfKGmOg)

Adds `Registry.count_select/2`

I was doubting if it should accept arbitrary full match spec or only `match_head` and optional `guards` similar to `count_match/4`, considering that the third element of the spec is always `[true]`, we could hide that.

However, by accepting full match spec we enable users to make use of [ex2ms](https://github.com/ericmj/ex2ms)

### DIFF
--- a/lib/elixir/lib/registry.ex
+++ b/lib/elixir/lib/registry.ex
@@ -1227,19 +1227,21 @@ defmodule Registry do
     guards = [{:"=:=", {:element, 1, :"$_"}, {:const, key}} | guards]
     spec = [{{:_, {:_, pattern}}, guards, [true]}]
 
-    case key_info!(registry) do
-      {:unique, partitions, key_ets} ->
-        key_ets = key_ets || key_ets!(registry, key, partitions)
-        :ets.select_count(key_ets, spec)
+    count_entries(registry, spec, fn partitions ->
+      :ets.select_count(key_ets!(registry, key, partitions), spec)
+    end)
+  end
 
-      {:duplicate, 1, key_ets} ->
+  defp count_entries(registry, spec, count_entries_across_unique_partitions_fun) do
+    case key_info!(registry) do
+      {:unique, partitions, nil} ->
+        count_entries_across_unique_partitions_fun.(partitions)
+
+      {_kind, 1, key_ets} ->
         :ets.select_count(key_ets, spec)
 
       {:duplicate, partitions, _key_ets} ->
-        Enum.reduce(0..(partitions - 1), 0, fn partition_index, acc ->
-          count = :ets.select_count(key_ets!(registry, partition_index), spec)
-          acc + count
-        end)
+        count_entries_across_partitions(registry, partitions, spec)
     end
   end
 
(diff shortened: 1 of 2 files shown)


josevalim:
Looks good! Please add tests for both unique and duplicate registries. :)

lib/elixir/lib/registry.ex
```diff
@@ -1318,6 +1318,47 @@ defmodule Registry do
...
+      when is_atom(registry) and is_list(spec) do
+    spec =
+      for part <- spec do
+        case part do
+          {{key, pid, value}, guards, select} ->
+            {{key, {pid, value}}, guards, select}
+
+          _ ->
+            raise ArgumentError,
+                  "invalid match specification in Registry.select_count/2: #{inspect(spec)}"
```
josevalim, line 1346:
```suggestion
                  "invalid match specification in Registry.count_select/2: #{inspect(spec)}"
```

lib/elixir/lib/registry.ex
```diff
@@ -1318,6 +1318,47 @@ defmodule Registry do
...
+        end
+      end
```
josevalim, line 1348: Please move this to a private function that can be reused by both select and select_count.

lib/elixir/lib/registry.ex
```diff
@@ -1318,6 +1318,47 @@ defmodule Registry do
...
+    case key_info!(registry) do
+      {_kind, partitions, nil} ->
+        Enum.reduce(0..(partitions - 1), 0, fn partition_index, acc ->
+          count = :ets.select_count(key_ets!(registry, partition_index), spec)
+          acc + count
+        end)
+
+      {_kind, 1, key_ets} ->
+        :ets.select_count(key_ets, spec)
+    end
```
josevalim, line 1359: We have this exact logic in `count_match` too. In fact, the one in count_match is a bit more optimized, so please extract the one in count_match and reuse it. :)

RudolfMan: Thank you for the guidance, José!

If I understand correctly, for `count_match` when `key_info!` returns `{:unique, partitions, nil}` - the logic is slightly different. There we can rely on given `key` to find the `key_ets`. While for `count_select` that's not the case.. we have to traverse all partitions to find the total count. Am I missing something?

Here https://github.com/elixir-lang/elixir/pull/11405/commits/a35e45e4e87e9833ef1bea227ce87610a1d9014a I extracted the logic out, in order to reuse it, but since that one particular case is different - I pass the logic in anonymous function, so that `count_select` and `count_match` can define their own way to handle the case. 
What would be the better way to deal with this situation?

_P.S. I couldn't come up with better naming for private functions, went with more descriptive for now_

lib/elixir/lib/registry.ex
RudolfMan: Attemptted to do that here https://github.com/elixir-lang/elixir/pull/11405/commits/7d06e77047692b0c00881e0689e4c3e660fca23d 
Is there a better way to identify the function name of the caller than fetching it from `__ENV__.function`?

josevalim: This is good, thank you!

lib/elixir/lib/registry.ex
josevalim: Yeah, sorry I missed this nuance. I think we should revert this one. :)

RudolfMan: Done! =)

josevalim:

PR #11413, opened 2021-11-22
Add HTML summary page to test coverage report

rlipscombe:
The `mix test.cover` task generates an HTML file for each module, but it doesn't generate an `index.html` file containing a summary.

This PR attempts to resolve that. It's rough-and-ready; comments welcomed.

### DIFF
--- a/lib/mix/lib/mix/tasks/test.coverage.ex
+++ b/lib/mix/lib/mix/tasks/test.coverage.ex
@@ -1,5 +1,6 @@
 defmodule Mix.Tasks.Test.Coverage do
   use Mix.Task
+  import Mix.Generator
 
   @moduledoc """
   Build reports from exported test coverage.
@@ -268,11 +269,20 @@ defmodule Mix.Tasks.Test.Coverage do
     ignore = opts[:ignore_modules] || []
     modules = Enum.reject(:cover.modules(), &ignored?(&1, ignore))
 
+    {module_results, totals} = gather_coverage(ok, modules)
+    module_results = Enum.sort(module_results, :desc)
+
     if summary_opts = Keyword.get(opts, :summary, true) do
-      summary(ok, modules, summary_opts)
+      summary(module_results, totals, summary_opts)
     end
 
-    html(modules, opts)
+    output = Keyword.get(opts, :output, "cover")
+    File.mkdir_p!(output)
+
+    html(modules, output)
+    html_summary(module_results, totals, summary_opts, Path.join(output, "index.html"))
+
+    Mix.shell().info("Generated HTML coverage results in #{inspect(output)} directory")
   end
 
   defp ignored?(mod, ignores) do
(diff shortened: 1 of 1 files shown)


josevalim:
Hi @rlipscombe! I would like to suggest to contribute this feature to Erlang's `cover` instead. This way all BEAM projects can benefit from it. WDYT?

rlipscombe:
I'd like to, but that would involve lifting all of the `gather_coverage` stuff into Erlang as well, and we'd need to do something with `:ignore_modules`.

That is: if I pull on any one bit of this, we end up migrating most of "mix test.cover" into Erlang.

Unless you can see a different way to slice the problem...?

josevalim:
I am not sure we need to surface all of it upstream. One of the reasons why we have ignore_modules here is because we compute the results to change the CLI status code, which is not a concern for cover. I think the summary result from cover could indeed include all modules. But if they think ignore_modules could be helpful there, that's likely a worthy contribution too.

rlipscombe:
> One of the reasons why we have ignore_modules here ... is not a concern for cover.

Counterpoint: The primary reason we use ignore_modules is to exclude them from _everything_ -- not just consideration of thresholds. Some examples: If a module is generated (we use a lot of protocol buffers), we don't want to see it. If a module is hard to unit test (we use e2e tests for those), we don't want to see it. Having these appear in the list means cognitive load.

> contribute this feature to Erlang's cover instead.

I'm not sure the best way to do expose an API for this; I'll continue thinking.

josevalim:
> Counterpoint: The primary reason we use ignore_modules is to exclude them from everything -- not just consideration of thresholds.

Ah, then that's not a problem. Then we should not pass those modules to `cover`, which means that `cover` should not generate a report for them (and therefore not include them in the summary).

rlipscombe:
Unless I'm missing something, that _does_ mean pushing `gather_coverage` (and all of its ETS-finagling) upstream, and hiding all of that behind, say, `:cover.summarize_to_file(modules)`. At that point, we probably also want `:cover.summarize(modules)` to generate the data for the textual report, because it shares the logic.

Actually, it seems better to have `:cover.summarize(modules)` and convert that to text and HTML. Except that OTP already does the HTML rendering, which kinda smudges the boundaries. So I'm wondering whether `:cover.summarize_to_file(summary)` could also exist, to avoid calculating the summary twice...

josevalim:
I think that's a good route. I guess we could add both `summarize` and `summary_to_file`, where the latter receives the result of summarize.

To give a bit more context, we have been glad to add those features around `cover` so far because I personally didn't consider printing to the terminal to be a `cover` concern. However, we always avoided changes to the HTML, to avoid discrepancy in styles and because that's pretty much `cover`'s business and we generally prefer to improve upstream.

josevalim:
I will go ahead and close this for now, as we don't plan on adding this feature. Thank you!

PR #11419, opened 2021-11-25
Fixed Verbiage in documentation for Stream.transform/3

MoshePaul:
Simple fix, the sentence was just worded a little oddly so I took out the extra "and an accumulator."

### DIFF
--- a/lib/elixir/lib/stream.ex
+++ b/lib/elixir/lib/stream.ex
@@ -800,10 +800,10 @@ defmodule Stream do
   @doc """
   Transforms an existing stream.
 
-  It expects an accumulator and a function that receives each stream element.
-  It must return a tuple, where the first element is a new
-  stream (often a list) or the atom `:halt`, and the second element is the
-  accumulator to be used by the next element, if any, in both cases.
+  It expects an accumulator and a function that receives two arguments,
+  the stream element and the updated accumulator. It must return a tuple,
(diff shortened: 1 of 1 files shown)


lib/elixir/lib/stream.ex
```diff
@@ -800,8 +800,8 @@ defmodule Stream do
   @doc """
   Transforms an existing stream.
 
-  It expects an accumulator and a function that receives each stream element
-  and an accumulator. It must return a tuple, where the first element is a new
+  It expects an accumulator and a function that receives each stream element.
+  It must return a tuple, where the first element is a new
   stream (often a list) or the atom `:halt`, and the second element is the
   accumulator to be used by the next element, if any, in both cases.
```
josevalim, line 806: The function receives two arguments, the stream element and the accumulator. Therefore the current version is incomplete. What do you think about:

```suggestion
  It expects an accumulator and a function that receives two arguments,
  the stream element and the updated accumulator. It must return a tuple,
  where the first element is a new stream (often a list) or the atom `:halt`,
  and the second element is the accumulator to be used by the next element.
```

MoshePaul: Makes sense to me! I was looking at it very narrowly, but this revision seems smarter to me.

josevalim:

PR #11420, opened 2021-11-26
Disable ssa and bool passes

josevalim:
Module bodies, especially in tests, tend to be
long, which affects the performance of passes
such as beam_ssa_opt and beam_bool. This commit
disables those passes during module definition.
As an example, this makes loading Elixir's test
suite 7-8% faster.


josevalim:

PR #11426, opened 2021-11-29
mix release: Allow passing args to "start" command

wojtekmach:
Say we have this app callback module:

    defmodule Foo.Application do
      @moduledoc false
      use Application

      @impl true
      def start(_type, args) do
        IO.inspect(:init.get_plain_arguments())

        children = []
        opts = [strategy: :one_for_one, name: Foo.Supervisor]
        Supervisor.start_link(children, opts)
      end
    end

Executing

    $ _build/dev/rel/foo/bin/foo start foo bar

Before this patch would print:

    ['--no-halt']

After this patch, we have:

    ['--no-halt', '--', 'start', 'foo', 'bar']

Extracted from burrito-elixir/burrito, thanks @QuinnWilton & @ashea-code!

### DIFF
--- a/lib/mix/lib/mix/tasks/release.init.ex
+++ b/lib/mix/lib/mix/tasks/release.init.ex
@@ -181,7 +181,7 @@ defmodule Mix.Tasks.Release.Init do
 
     case $1 in
       start)
-        start "elixir" --no-halt
+        start "elixir" --no-halt -- $@
         ;;
 
       start_iex)
@@ -378,7 +378,7 @@ defmodule Mix.Tasks.Release.Init do
       --erl-config "!RELEASE_SYS_CONFIG!" ^
       --boot "!REL_VSN_DIR!\!RELEASE_BOOT_SCRIPT!" ^
       --boot-var RELEASE_LIB "!RELEASE_ROOT!\lib" ^
-      --vm-args "!RELEASE_VM_ARGS!"
+      --vm-args "!RELEASE_VM_ARGS!" -- %*
     goto end
 
     :eval


ashea-code:
First, thanks so much for bringing this patch over!

I have another question relating to this batch file:

these lines in the windows batch files (Line 286)
```batch
pushd .
cd "%~dp0\.."
set RELEASE_ROOT=%cd%
popd
 ```
 
I assume this is used to compute the current directory of the batch file so we can launch the Erlang VM properly later in the file. There appears to be a subtle bug if you call this batch file from a Windows shell prompt that is currently within another drive letter.

Here's a minimal example:

I have a test.bat file that I will save to my desktop:

```batch
@echo off

pushd .
cd "%~dp0"
set RELEASE_ROOT=%cd%
popd

echo "BAT file location with pushd: %RELEASE_ROOT%"

set RELEASE_ROOT=%~dp0

echo "BAT file location without pushd: %RELEASE_ROOT%"
```

I'll print the result of `%RELEASE_ROOT%` using the method from the Mix Release's script, and then print it again using a more simplistic method of querying the current batch file's directory.

The results are:

```
C:\>C:\Users\Test\Desktop\test.bat
"BAT file location with pushd: C:\Users\Test\Desktop"
"BAT file location without pushd: C:\Users\Test\Desktop\"
```

But if I change my prompt to another drive letter:

```
C:\>B:

B:\>C:\Users\Test\Desktop\test.bat
"BAT file location with pushd: B:\"
"BAT file location without pushd: C:\Users\Test\Desktop\"
```

My question is: is there a reason for the `pushd` and `popd` or can we just revise that segment of the batch file to use the more straightforward `set RELEASE_ROOT="%~dp0\.."`?

QuinnWilton:
> Extracted from burrito-elixir/burrito, thanks @QuinnWilton & @ashea-code!

You asked last night whether this PR was an accurate summary of the issues we ran into, so I'm just jumping in to confirm that this looks good on my end!

Aaron mentioned another issue in his last comment, but there is also some quirks we've run into with arguments being escaped differently between Windows and other platforms, but we can chat about that elsewhere if you'd like. I'll probably try to find some time to put together a minimal reproduction of that and then open an issue to get your thoughts!

josevalim:
> My question is: is there a reason for the pushd and popd or can we just revise that segment of the batch file to use the more straightforward set RELEASE_ROOT="%~dp0\.."?

We want an expanded path and that's the only way to get it. Or so I thought it was, I just found this on StackOverflow: https://stackoverflow.com/questions/1645843/resolve-absolute-path-from-relative-path-and-or-file-name :) A PR (or an issue) would definitely be welcome!

> but there are also some quirks we've run into with arguments being escaped differently between Windows and other platforms, but we can chat about that elsewhere if you'd like

Glad to chat, but to clarify, escaping does work differently in both and unfortunately that's specific to each shell. We can also be messing something up here and there.

ashea-code:
> We want an expanded path and that's the only way to get it. Or so I thought it was, I just found this on StackOverflow: https://stackoverflow.com/questions/1645843/resolve-absolute-path-from-relative-path-and-or-file-name :) A PR (or an issue) would definitely be welcome!

Sounds good to me! I'd be happy to write up a PR for this particular bit

josevalim:
Looking forward to it and feel free to reach out!

simonmcconnell:
I could not figure how to escape args passed to `eval`. I decided to write the release and elixir batch scripts in zig instead. Another priority was to hide the functionality in the batch script from the end users.

whatyouhide (approved):
LGMT!

QuinnWilton:
> Glad to chat, but to clarify, escaping does work differently in both and unfortunately that's specific to each shell. We can also be messing something up here and there.

Sorry, I saw this response come in, but I forgot to revisit it when I had time to respond! This may be a bit of a rambling reply, since I'm mostly working off of memory, and I don't have access to my development machine at the moment.

I understand that each shell has different escaping rules, but in our case it seemed like certain characters needed to be escaped multiple times, which I suspect is unintended. Here's an example of two equivalent command-line invocations, across Bash and Powershell 5:

```
# On Bash
./my_release start --filter "verb!='post'"

# On Powershell 5
.\my_release.bat start --filter '"'"verb^^^!='post'"'"'
```

I didn't spend too much time investigating, so I haven't been able to confirm, but my intuition tells me that this may be because invoking an Elixir release goes through the following flow:
1) [release.bat](https://github.com/elixir-lang/elixir/blob/043f1c49ec16376da3393558e189644021746c6f/lib/mix/lib/mix/tasks/release.init.ex#L281) is invoked, and the command line string is parsed by either `cmd.exe` or Powershell
2) [elixir.bat](https://github.com/elixir-lang/elixir/blob/main/bin/elixir.bat) is invoked, and the command line string is parsed by either `cmd.exe` or Powershell
3) The "end" of the command line string is passed to `erl.exe` as the argument to `-extra`

I believe that each of steps (1) and (2)  parse the input in such a way as to require that the users escapes things twice. I think everything behaves as expected once the input reaches `erl.exe` though. Also worth noting is that `cmd.exe`, Powershell 5, and Powershell 7 all have different escaping rules, which further complicates things for the user.

I don't know that you're necessarily doing anything wrong, and it's possible that I've made some faulty assumptions here, but I just wanted to raise these observations to you in case they point to a concern worth looking into!

Also, let me know if you'd like me to put together a reproduction of anything I've described here: I don't have access to a Windows install of Elixir at the moment, but I'm happy to do that once I'm back at work tomorrow and can access our development servers.

josevalim:
Ouch, what you describe sounds like a bug then and perhaps a very big blocker to adding this functionality. :) thanks for sharing!

QuinnWilton:
> Ouch, what you describe sounds like a bug then and perhaps a very big blocker to adding this functionality. :) thanks for sharing!

For sure! I haven't tested this behavior using a regular Elixir release yet, without Burrito, so it's possible that the problem is on our end. That being said, Burrito was receiving the correct command line arguments, and we were just passing the entire string along without doing any parsing, which seemed to suggest that the issue was somewhere between `release.bat` and `erl.exe`.

We worked around the problem by encoding the command line string as Base64 from Zig, and then decoding that from Elixir, but then Thanksgiving happened, and I haven't found a chance to get back to looking into the issue 😅 Things are calming down on my end though, so let me know if you need any additional information or code samples, and I'll see what I can put together.

wojtekmach:
I've spoken with @josevalim and we decided not to move forward with this after all. We don't want users to have to use `:init.get_plain_arguments()` and juggle the args around the `--` sentinel value. A more native solution would be one where all they'd need to do is call `System.argv()` and everything would already be in place, however getting there is challenging. Closing for now!

wojtekmach:
Here's another way to tackle this problem: https://github.com/elixir-lang/elixir/pull/12734.

PR #11428, opened 2021-12-01
Fix opaque Version.Requirement usage

michallepicki:
Do not pattern match on internal opaque type structure
outside of the module with that opaque type definition.
Fixes related dialyzer issues.


josevalim:

PR #11431, opened 2021-12-01
Fix normalization of kw list in blocks

doorgan:
Regular blocks were interpreted as a call, so if the last expression in a block was a kw list then the wrapper block would be removed, causing the formatter to crash


josevalim:

PR #11433, opened 2021-12-03
#11418 Disallow shorthand pipe after matches

am-kantox:
Please refer to #11418 for a feature discussion.

I am not sure if the error message is descriptive enough, but AFAIU for a better report we’d need to introduce a new type of error and I tend to avoid it here.

### DIFF
--- a/lib/iex/lib/iex/evaluator.ex
+++ b/lib/iex/lib/iex/evaluator.ex
@@ -64,11 +64,13 @@ defmodule IEx.Evaluator do
   @doc false
   def parse(input, opts, buffer)
 
-  def parse(@break_trigger, _opts, "") do
+  def parse(input, opts, ""), do: parse(input, opts, {"", :none})
+
+  def parse(@break_trigger, _opts, {"", _last_op}) do
     {:incomplete, ""}
   end
 
-  def parse(@break_trigger, opts, _buffer) do
+  def parse(@break_trigger, opts, {_buffer, _last_op}) do
     :elixir_errors.parse_error(
       [line: opts[:line]],
       opts[:file],
@@ -78,12 +80,11 @@ defmodule IEx.Evaluator do
     )
   end
 
-  def parse(input, opts, buffer) do
+  def parse(input, opts, {buffer, last_op}) do
     input = buffer <> input
     file = Keyword.get(opts, :file, "nofile")
     line = Keyword.get(opts, :line, 1)
     column = Keyword.get(opts, :column, 1)
-    last_op = Keyword.get(opts, :last_op, :none)
     charlist = String.to_charlist(input)
 
     result =
@@ -91,10 +92,10 @@ defmodule IEx.Evaluator do
            {:ok, adjusted_tokens} <- adjust_operator(tokens, line, column, file, opts, last_op),
            {:ok, forms} <- :elixir.tokens_to_quoted(adjusted_tokens, file, opts) do
         last_op =
-          Enum.reduce_while(tokens, :other, fn
-            {:match_op, _, _}, _ -> {:halt, :match}
-            _, acc -> {:cont, acc}
-          end)
+          case forms do
+            {:=, _, [_, _]} -> :match
+            _ -> :other
+          end
 
         {:ok, forms, last_op}
       end
@@ -112,7 +113,7 @@ defmodule IEx.Evaluator do
           file,
           "pipe shorthand is not allowed immediately after a match expression in IEx; to make it work, surround the whole pipeline with parentheses ",
           token,
-          {charlist, line, column}
+          {charlist, Keyword.get(location, :line, line), Keyword.get(location, :column, column)}
         )
 
       {:error, {location, error, token}} ->
@@ -242,6 +243,7 @@ defmodule IEx.Evaluator do
       env: env,
       server: server,
       history: history,
+      last_op: :none,
       stacktrace: stacktrace,
       ref: ref
     }
--- a/lib/iex/lib/iex/evaluator.ex
+++ b/lib/iex/lib/iex/evaluator.ex
@@ -62,15 +62,15 @@ defmodule IEx.Evaluator do
                [:three_op, :concat_op, :mult_op]
 
   @doc false
-  def parse(input, opts, buffer)
+  def parse(input, opts, parser_state)
 
   def parse(input, opts, ""), do: parse(input, opts, {"", :other})
 
-  def parse(@break_trigger, _opts, {"", last_op}) do
-    {:incomplete, {"", last_op}}
+  def parse(@break_trigger, _opts, {"", _} = parser_state) do
+    {:incomplete, parser_state}
   end
 
-  def parse(@break_trigger, opts, {_buffer, _last_op}) do
+  def parse(@break_trigger, opts, _parser_state) do
     :elixir_errors.parse_error(
       [line: opts[:line]],
       opts[:file],
@@ -243,7 +243,6 @@ defmodule IEx.Evaluator do
       env: env,
       server: server,
       history: history,
-      last_op: :other,
       stacktrace: stacktrace,
       ref: ref
     }
@@ -306,7 +305,7 @@ defmodule IEx.Evaluator do
   defp do_eval({:ok, forms, buffer}, iex_state, state) when is_binary(buffer),
     do: do_eval({:ok, forms, {buffer, :other}}, iex_state, state)
 
-  defp do_eval({:ok, forms, {buffer, last_op}}, iex_state, state) do
+  defp do_eval({:ok, forms, parser_state}, iex_state, state) do
     put_history(state)
     put_whereami(state)
     state = handle_eval(forms, iex_state.counter, state)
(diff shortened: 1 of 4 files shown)


lib/iex/lib/iex/evaluator.ex
```diff
@@ -83,20 +83,38 @@ defmodule IEx.Evaluator do
...
+        {:ok, forms, "", last_op}
 
       {:error, {_, _, ""}} ->
         {:incomplete, input}
 
+      {:error, {location, :malformed_pipe, token}} ->
+        :elixir_errors.parse_error(
+          location,
+          file,
+          "pipe shorthand is not allowed immediately after a match expression in IEx; to make it work, surround the whole pipeline with parentheses ",
```
josevalim, line 113:
```suggestion
          "pipe shorthand is not allowed immediately after a match expression in IEx. To make it work, surround the whole pipeline with parentheses",
```

lib/iex/lib/iex/evaluator.ex
```diff
@@ -83,20 +83,38 @@ defmodule IEx.Evaluator do
...
+          token,
+          {charlist, line, column}
+        )
```
josevalim, line 116: The error message is good but I think we should already return the `{location, message, token}` format from `adjust_operator` and then it falls under the clause below.

josevalim: The only change to the error message is to remove `;`. We rarely use it in Elixir error messages. :)

lib/iex/lib/iex/evaluator.ex
```diff
@@ -83,20 +83,38 @@ defmodule IEx.Evaluator do
     file = Keyword.get(opts, :file, "nofile")
     line = Keyword.get(opts, :line, 1)
     column = Keyword.get(opts, :column, 1)
+    last_op = Keyword.get(opts, :last_op, :none)
```
josevalim, line 86: This is a state that is specific to the evaluator, so I don't think it should surface all the way up to the IEx server. The trick here is to change the evaluator state. The first clause should be:

  def parse(input, opts, ""), do: parse(input, opts, {"", :none})

And now always return a tuple as the evaluation state, where the second element is the last operation.

lib/iex/lib/iex/evaluator.ex
```diff
@@ -83,20 +83,38 @@ defmodule IEx.Evaluator do
...
-      with {:ok, tokens} <- :elixir.string_to_tokens(charlist, line, column, file, opts) do
-        :elixir.tokens_to_quoted(adjust_operator(tokens, line, column, file, opts), file, opts)
+      with {:ok, tokens} <- :elixir.string_to_tokens(charlist, line, column, file, opts),
+           {:ok, adjusted_tokens} <- adjust_operator(tokens, line, column, file, opts, last_op),
+           {:ok, forms} <- :elixir.tokens_to_quoted(adjusted_tokens, file, opts) do
+        last_op =
+          Enum.reduce_while(tokens, :other, fn
+            {:match_op, _, _}, _ -> {:halt, :match}
+            _, acc -> {:cont, acc}
+          end)
```
josevalim, line 97: I would look at the forms instead. If `match?({:=, _, [_, _], forms)`, then you have a match. :)

lib/iex/lib/iex/evaluator.ex
```diff
@@ -64,11 +64,13 @@ defmodule IEx.Evaluator do
   @doc false
   def parse(input, opts, buffer)
 
-  def parse(@break_trigger, _opts, "") do
+  def parse(input, opts, ""), do: parse(input, opts, {"", :none})
```
josevalim, line 67: We probably don't need `:none` and `:other`. So perhaps we use `:other` here?

lib/iex/lib/iex/evaluator.ex
```diff
@@ -78,25 +80,42 @@ defmodule IEx.Evaluator do
...
+          end
+
+        {:ok, forms, last_op}
       end
 
     case result do
-      {:ok, forms} ->
-        {:ok, forms, ""}
+      {:ok, forms, last_op} ->
+        {:ok, forms, "", last_op}
```
josevalim, line 105: Now we should change this to be `{:ok, forms, {"", last_op}}`. The API on server should not change but we should probably rename the `buffer` variable in there to something else. Maybe `parser_state`?

lib/iex/lib/iex/evaluator.ex
```diff
@@ -220,6 +243,7 @@ defmodule IEx.Evaluator do
       env: env,
       server: server,
       history: history,
+      last_op: :other,
```
josevalim, line 246: Should no longer be here. :)

lib/iex/lib/iex/evaluator.ex
am-kantox: > Maybe `parser_state`?

Luckily enough, we pass the necessity to name things here; it makes sense to restructure it everywhere, so `buffer` stays as in `{buffer, last_op}`.

lib/iex/lib/iex/evaluator.ex
am-kantox: It’s evaluator’s state. Not `iex` state. And I need it to use a short notation `%{state | last_op: last_op}` down the rabbit hole.

lib/iex/lib/iex/evaluator.ex
```diff
@@ -279,18 +309,21 @@ defmodule IEx.Evaluator do
     end
   end
 
-  defp do_eval({:ok, forms, buffer}, iex_state, state) do
+  defp do_eval({:ok, forms, buffer}, iex_state, state) when is_binary(buffer),
+    do: do_eval({:ok, forms, {buffer, :other}}, iex_state, state)
+
+  defp do_eval({:ok, forms, {buffer, last_op}}, iex_state, state) do
```
josevalim, line 315: Sorry, to be clear, the evaluator should not know anything about `last_op`. It is a parser specific behaviour that we shouldn't leak. This should be an "opaque" term to the evaluator: 

```suggestion
  defp do_eval({:ok, forms, parser_state}, iex_state, state) do
```

My suggestion is to rename the `:buffer` key to `:parser_state`.

am-kantox: Ah. Got it. ok.

am-kantox: @josevalim could you please provide guidance on where would I store the `last_op` between different `parse/3` invocations, if not in evaluator’s state? This:
```elixir
iex|1> x = 42
iex|2> |> IO.puts()
```
results in two invocations of a pure `parse/3` function and the second invocation should take the value of `last_op` set during the first one. Am I missing something?

am-kantox: I think I finally get it.

> rename the `:buffer` key to `:parser_state`

`:buffer` key belongs to `IEx.State`. I was under the impression you didn’t want to amend `IEx.State` and keep things private to the evaluator. But we can move it there, of course.

am-kantox: I think I finally get it.

> rename the `:buffer` key to `:parser_state`

`:buffer` key belongs to `IEx.State`. I was under the impression you didn’t want to amend `IEx.State` and keep things private to the evaluator. But we can move it there, of course.

josevalim: I couldn't answer your question without trying the code, so I just pushed a change that makes the parser state opaque to the server: https://github.com/elixir-lang/elixir/commit/abc23d00c4a9bbafcf08237ed64f93e175196bfa

Please rebase and you should be able to address this by only changing the parse function. :)

lib/iex/lib/iex/evaluator.ex
```diff
@@ -279,6 +302,9 @@ defmodule IEx.Evaluator do
     end
   end
 
+  defp do_eval({:ok, forms, buffer}, iex_state, state) when is_binary(buffer),
+    do: do_eval({:ok, forms, {buffer, :other}}, iex_state, state)
```
josevalim, line 306: This change should no longer be required. :)

lib/iex/lib/iex/evaluator.ex
josevalim: Only this comment left and we should be good. In a nutshell, I think you should return the message above instead of :malformed_pipe in `adjust_operator`.

lib/iex/lib/iex/evaluator.ex
```diff
@@ -108,13 +127,17 @@ defmodule IEx.Evaluator do
     end
   end
 
-  defp adjust_operator([{op_type, _, _} | _] = tokens, line, column, file, opts)
+  defp adjust_operator([{op_type, _, token} | _] = _tokens, line, column, _file, _opts, :match)
+       when op_type in @op_tokens,
+       do: {:error, {[line: line, column: column], :malformed_pipe, "('#{token}')"}}
```
josevalim, line 132: There are no parens around the token in the error message!

```suggestion
       do: {:error, {[line: line, column: column], :malformed_pipe, "'#{token}'"}}
```

lib/iex/lib/iex/evaluator.ex
am-kantox: @josevalim ✓ Done.

josevalim:

PR #11436, opened 2021-12-03
Update microsecond typespec

LostKobrakai:
…to show which part of the tuple is the value and which the precision


josevalim:

PR #11438, opened 2021-12-03
Adds System.fetch_mandatory_env!/1 utility

piacsek:
# Proposal

This PR aims to add a utility function to `System` module to make it easier to fetch environment variables that *need* to have a non-blank value.

# Motivation

At SumUp, we have our South American online banking solution written in Elixir. We used to rely on `System.fetch_env!/1` to retrieve environment variables up until we had an infrastructure incident that made some environment variables to be set with an empty string as value. The system booted because `System.fetch_env!/1` did not raise an exception, so we ended up with a running system without some of its necessary data.

### DIFF
--- a/lib/elixir/lib/system.ex
+++ b/lib/elixir/lib/system.ex
@@ -700,6 +700,27 @@ defmodule System do
   end
 
   @doc """
+  Fetches an environment variable.
+  Raises an `ArgumentError` in case the variable is not set or its trimmed value is blank.
+
+  Example usage in a config file
+
+  config :my_app,
+    MyModule,
+    my_mandatory_value: System.fetch_mandatory_env!("ENV_THAT_CANNOT_BE_BLANK")
+  """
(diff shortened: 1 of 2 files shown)


josevalim:
Hi @piacsek, thanks for the PR! Can you please open up a discussion on the Elixir core mailing list first? I would say the real issue here is lack of validation. And while blank validation can be helpful, you can still set a host to "example" instead of "example.com", a port to "8O" instead of "80", and so forth. So I would like some discussion around it, because I personally don't find it sufficient. :)

fuelen:
Hey @piacsek
I agree, that the real issue here is lack of validation. `System.fetch_mandatory_env!/1` solves only 2 cases, while real world application can have a much more. I'd recommend defining a custom module with helpers and validations that are needed for your app.
At proSapient, we had similar issues, and when Elixir 1.11 was released, I wrote a [utility](https://github.com/prosapient/confispex) which goes far beyond checking nil/blank string. Check Motivation section and Getting started guide, may be helpful.

PR #11439, opened 2021-12-04
Add guards and specs to Identifier.escape/4

eksperimental:
Note: I am not 100% sure on the `fun` specs

### DIFF
--- a/lib/elixir/lib/code/identifier.ex
+++ b/lib/elixir/lib/code/identifier.ex
@@ -214,8 +214,11 @@ defmodule Code.Identifier do
   @doc """
   Escapes the given identifier.
   """
-  def escape(other, char, count \\ :infinity, fun \\ &escape_map/1) do
-    escape(other, char, count, [], fun)
+  @spec escape(binary(), char(), :infinity | non_neg_integer, (char() -> iolist() | false)) ::
+          {escaped :: iolist(), remaining :: binary()}
+  def escape(binary, char, limit \\ :infinity, fun \\ &escape_map/1)
+      when limit == :infinity or (is_integer(limit) and limit >= 0) do
+    escape(binary, char, limit, [], fun)
   end
 
   defp escape(<<_, _::binary>> = binary, _char, 0, acc, _fun) do


lib/elixir/lib/code/identifier.ex
```diff
@@ -214,8 +214,11 @@ defmodule Code.Identifier do
   @doc """
   Escapes the given identifier.
   """
-  def escape(other, char, count \\ :infinity, fun \\ &escape_map/1) do
-    escape(other, char, count, [], fun)
+  @spec escape(binary(), char(), :infinity | non_neg_integer, (char() -> iolist() | false)) ::
```
josevalim, line 217:
```suggestion
  @spec escape(binary(), char(), :infinity | non_neg_integer, (char() -> char() | false)) ::
```

eksperimental: Look José.
https://github.com/elixir-lang/elixir/blob/846147285b614f66677ce38d89d00e5038774d53/lib/elixir/lib/code/identifier.ex#L276-L286

eksperimental: could be `charlist()` instead, the thing is that chars are not part of io_lists, so it should be [byte()].. That is why iolist made more sense  to me.

josevalim: But it can return chars too, so i think it is `char() | iolist() | false`. WDYT?

josevalim: And good catch btw!

eksperimental: Off the top of my head, I think if you return a non-byte char, I think IO.iodata_to_binary() will crash. I think that is why I decided to spec the functions because i could not fully understand how it worked.
Or maybe we should modify the code to deal with chars too?

josevalim: No, you are correct. My bad. :)

josevalim:

eksperimental:
Awesome, thank you.

PR #11441, opened 2021-12-04
Colorize Version.Requirement source in Inspect protocol

eksperimental:
This is to avoid ambiguity in cases like

    iex> Version.parse_requirement!("~> 1.2.3 or < 3.4.5 and > 5.6.7")
    #Version.Requirement<~> 1.2.3 or < 3.4.5 and > 5.6.7>

Before:
image

After:
image

### DIFF
--- a/lib/elixir/lib/version.ex
+++ b/lib/elixir/lib/version.ex
@@ -677,7 +677,12 @@ defimpl String.Chars, for: Version.Requirement do
 end
 
 defimpl Inspect, for: Version.Requirement do
-  def inspect(%Version.Requirement{source: source}, _opts) do
-    "#Version.Requirement<" <> source <> ">"
+  def inspect(%Version.Requirement{source: source}, opts) do
+    formatted =
+      Inspect.Algebra.color(source, :atom, opts)
+      |> Inspect.Algebra.format(opts.width)
+      |> IO.iodata_to_binary()
+
+    "#Version.Requirement<" <> formatted <> ">"
   end
 end
--- a/lib/elixir/lib/version.ex
+++ b/lib/elixir/lib/version.ex
@@ -403,7 +403,7 @@ defmodule Version do
 
       iex> {:ok, requirement} = Version.parse_requirement("== 2.0.1")
       iex> requirement
-      #Version.Requirement<== 2.0.1>
+      #Version.Requirement<"== 2.0.1">
 
       iex> Version.parse_requirement("== == 2.0.1")
       :error
@@ -425,7 +425,7 @@ defmodule Version do
   ## Examples
 
       iex> Version.parse_requirement!("== 2.0.1")
-      #Version.Requirement<== 2.0.1>
+      #Version.Requirement<"== 2.0.1">
 
       iex> Version.parse_requirement!("== == 2.0.1")
       ** (Version.InvalidRequirementError) invalid requirement: "== == 2.0.1"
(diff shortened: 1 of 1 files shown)


eksperimental:
Initially I tried to escape, not to colorized, but I did not like the idea.
Now i realized that when interpolating the output it is still ambiguous,

```elixir
iex(7)> IO.puts "requirement: #{inspect(vr)}"   
requirement: #Version.Requirement<~> 1.2.3 or < 3.4.5 and > 5.6.7>
```


Maybe we should escape and colorize?

michallepicki:
@eksperimental Orange text background on the Before screenshot is your text selection, correct? And you mean that the user could easily miss the first `~>`?

eksperimental:
> @eksperimental Orange text background on the Before screenshot is your text selection, correct? And you mean that the user could easily miss the first `~>`?

Sorry, i know my description was not clear enough. What is ambiguous is the use of  `>` in the source of the version requirement, which closes the opening `<` in  `#Version.Requirement<`

josevalim:
@eksperimental what if we put it between quotes and escape it as a string?

    #Version.Requirement<"~> 2.0">

The string is more likely the shape we will see the requirement and it aligns with how we render some other stuff like `MapSet`.

eksperimental:
> @eksperimental what if we put it between quotes and escape it as a string?
> 
> ```
> #Version.Requirement<"~> 2.0">
> ```
> 
> The string is more likely the shape we will see the requirement and it aligns with how we render some other stuff like `MapSet`.

Yes! I guess that is the way to got. In fact t:Version.requirement/0  is `String.t()` | `Version.Requirement.t()`

wojtekmach:
would brackets help?

`#Version.Requirement[~> 2.0]`

we use those for calendar structs.

eksperimental:
> would brackets help?
> 
> `#Version.Requirement[~> 2.0]`
> 
> we use those for calendar structs.

So which one is clearer when doing string interpolation?

square brackets:
> requirement: #Version.Requirement[~> 1.2.3 or < 3.4.5 and > 5.6.7]

double quotes:
> requirement: #Version.Requirement<"~> 1.2.3 or < 3.4.5 and > 5.6.7">

josevalim:
Things like doctests assume `#...<`, so we should stick with `<`.

eksperimental:
This is what I've got so far:

image

I am not escaping the double-quotes, because they are not allowed in versions, and I don't see how they will ever have a double quote, but let me know if you think there is an edge case

lib/elixir/lib/version.ex
```diff
@@ -677,7 +677,13 @@ defimpl String.Chars, for: Version.Requirement do
...
-  def inspect(%Version.Requirement{source: source}, _opts) do
-    "#Version.Requirement<" <> source <> ">"
+  def inspect(%Version.Requirement{source: source}, opts) do
+    formatted =
+      ("\"" <> source <> "\"")
+      |> Inspect.Algebra.color(:string, opts)
+      |> Inspect.Algebra.format(:infinity)
+      |> IO.iodata_to_binary()
+
+    "#Version.Requirement<" <> formatted <> ">"
```
josevalim, line 687: You should not render inside inspect. You want more something like this (untested):

```suggestion
    inner = Inspect.Algebra.color("\"" <> source <> "\"", :string, opts)
    Inspect.Algebra.concat(["#Version.Requirement<", inner, ">"])
```

eksperimental: Got you. Fixed

josevalim:

eksperimental:
Thank you everyone for chiming in.

PR #11447, opened 2021-12-05
Validate options for all Base encode/decode functions

sorentwo:
Incorrect options, e.g., `pad: false` rather than `padding: false` were silently ignored. This mistake bit me recently, and I was surprised to find that encode/decode functions didn't validate the options, and the typespecs were loose enough that dialyzer didn't catch it either.  This mistake seems prevalent enough that the "identity" test case erroneously used `pad` as an option.

The validations are doubled up for functions with a bang variant because `Keyword.validate!/2` raises an `ArgumentError`, and `rescue` would translate it into a generic `:error` tuple. There is a potential refactoring here where the bang variant calls the non-bang variant and raises an `{:error, reason}` tuple. However, that seemed too invasive and I didn't include it in this PR.

### DIFF
--- a/lib/elixir/test/elixir/base_test.exs
+++ b/lib/elixir/test/elixir/base_test.exs
@@ -799,13 +799,36 @@ defmodule BaseTest do
         |> Enum.shuffle()
         |> IO.iodata_to_binary()
 
+      allowed_opts =
+        encode
+        |> Function.info()
+        |> Keyword.fetch!(:name)
+        |> case do
+          :encode16 -> [:case]
+          :encode64 -> [:padding]
+          :url_encode64 -> [:padding]
+          _ -> [:case, :padding]
+        end
+
       expected =
         data
-        |> encode.(case: encode_case, pad: pad?)
-        |> decode.(case: decode_case, pad: pad?)
+        |> encode.(Keyword.take([case: encode_case, padding: pad?], allowed_opts))
+        |> decode.(Keyword.take([case: decode_case, padding: pad?], allowed_opts))
 
       assert data == expected,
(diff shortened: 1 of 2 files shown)


lib/elixir/test/elixir/base_test.exs
```diff
@@ -799,13 +799,36 @@ defmodule BaseTest do
...
     end
   end
+
+  test "unknown encode options raise an argument error" do
+    for encode <- [&encode16/2, &encode32/2, &encode64/2, &hex_encode32/2, &url_encode64/2] do
+      assert_raise ArgumentError, fn -> encode.("foobar", pad: false) end
+    end
+  end
+
+  test "unknown decode options raise an argument error" do
```
sorentwo, line 829: This test and the one above could easily be combined into one.

lib/elixir/test/elixir/base_test.exs
```diff
@@ -799,13 +799,36 @@ defmodule BaseTest do
         |> Enum.shuffle()
         |> IO.iodata_to_binary()
 
+      allowed_opts =
+        encode
+        |> Function.info()
```
sorentwo, line 804: Injecting this into the existing identity test preserved the test history. Alternatively, I could break this test into three cases (*16, *32, *64) and extract some helpers. Thoughts?

josevalim:
Thanks for the PR!

Unfortunately, this can be a breaking change. If someone does this:

```
@doc """

  :padding - controls padding

"""
def some_funtion_that_calls_pad(..., opts \\ []) do
  Base.something(opts)
end
```

Then the code will no longer work. And, to be honest, I am not sure we want to go in this direction either. However, we should definitely keep the spec updates. Can you please send a PR only with them and the test fixes?

PR #11449, opened 2021-12-05
Specific base typespecs

sorentwo:
Each `Base` function had `keyword` as the option type, which didn't help guard against typos or mismatched options.

This also includes a small fix to the `identity` spec, which passed unused options in many cases and used `pad:` rather than `padding:` where appropriate.

### DIFF
--- a/lib/elixir/lib/base.ex
+++ b/lib/elixir/lib/base.ex
@@ -92,6 +92,9 @@ defmodule Base do
 
   """
 
+  @type encode_case :: :upper | :lower
+  @type decode_case :: :upper | :lower | :mixed
+
   b16_alphabet = '0123456789ABCDEF'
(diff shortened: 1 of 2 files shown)


josevalim:

lib/elixir/lib/base.ex
```diff
@@ -397,7 +400,7 @@ defmodule Base do
       {:ok, "foob"}
 
   """
-  @spec decode64(binary, keyword) :: {:ok, binary} | :error
+  @spec decode64(binary, ignore: :whitespace, padding: boolean) :: {:ok, binary} | :error
```
fuelen, line 403: Is it valid to use such notation to cover reversed order of arguments `decode("...", padding: true, ignore: :whitespace)`?

sorentwo: Great point. I've opened a PR to handle that situation.

PR #11455, opened 2021-12-07
Add `Enum.sum/2`

hauleth:
The behaviour of new function is similar to `map/2` followed by `sum/1`.

This behaves similarly to:

- `Enum.all?/2`
- `Enum.any?/2`
- `Enum.count/2`
- All `*_by/2` functions

This should simplify a lot of cases when we want to sum values stored in nested manner like summing values in list of tuples, list of maps, or even in maps itself.

### DIFF
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -3071,6 +3071,31 @@ defmodule Enum do
   end
 
   @doc """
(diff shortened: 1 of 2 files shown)


v0idpwn:
Related: https://github.com/elixir-lang/elixir/pull/11381

josevalim:
Closed per the previous discussion. Thanks!

PR #11456, opened 2021-12-07
Add Stream.duplicate/1

camilleryr:
@josevalim  mentioned in [this twitch stream](https://www.twitch.tv/videos/1225900187) (around minute 37) that Stream.duplicate/2 would be a good addition.  Thoughts?

```
      iex> stream = Stream.duplicate([1, 2], 3)
      iex> Enum.to_list(stream)
      [[1, 2], [1, 2], [1, 2]]
```

### DIFF
--- a/lib/elixir/lib/stream.ex
+++ b/lib/elixir/lib/stream.ex
@@ -442,6 +442,7 @@ defmodule Stream do
       iex> Enum.to_list(stream)
       [[1, 2], [1, 2], [1, 2]]
   """
+  @doc since: "1.14.0"
   @spec duplicate(any, non_neg_integer) :: Enumerable.t()
   def duplicate(value, n) when is_integer(n) and n >= 0 do
     unfold(n, fn
--- a/lib/elixir/lib/stream.ex
+++ b/lib/elixir/lib/stream.ex
@@ -418,6 +418,39 @@ defmodule Stream do
   end
 
   @doc """
+  Duplicates the given element `n` times in a stream.
+
+  `n` is an integer greater than or equal to `0`.
+
+  If `n` is `0`, an empty stream is returned.
+
+  ## Examples
+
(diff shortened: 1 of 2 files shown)


josevalim:
Sorry, I wasn't clear enough. We shuold have `Stream.duplicate(elem, n)` where `n >= 0`, that behaves like `List.duplicate/2` and `Tuple.duplicate/2`. Implementation wise, I think it can be implemented with `unfold`. :)

maennchen:
@josevalim Even though there seems to be some prior similar implementation, I think `duplicate` is not actually a great name for this functionality. Duplicate implies to make two out of one and in this case, we're making any number of entries. Wouldn't something like `multiply` be a better name?

josevalim:
Multiply is definitely ambiguous. I would say the better name would be repeat but I think we should stay with the stdlib convention at this point. "Duplicate something N times" also seem to be relatively common.

josevalim:
To further add, we tend to `repeat` for actions (i.e. repeat an anonymous function), so I will take a step back and say "duplicate n times" is better as it reveals the static nature of the argument.

camilleryr:
Updated the changes to reflect your comments, now has `Stream.duplicate/2` as implemented with `unfold`.  Thanks!

lib/elixir/lib/stream.ex
```diff
@@ -417,6 +417,39 @@ defmodule Stream do
...
+      ["hi"]
+
+      iex> stream = Stream.duplicate("bye", 2)
+      iex> Enum.to_list(stream)
+      ["bye", "bye"]
+
+      iex> stream = Stream.duplicate([1, 2], 3)
+      iex> Enum.to_list(stream)
+      [[1, 2], [1, 2], [1, 2]]
+  """
```
josevalim, line 444:
```suggestion
  """
  @doc since: "1.14.0"
```

josevalim:

PR #11462, opened 2021-12-08
Fix formatting of lists in module attribues

### LINKED ISSUE #11459: Macro.to_string/1 drops square brackets in module attributes
### Environment

* Elixir & Erlang/OTP versions (elixir --version):  **1.13.0 / OTP 24**
* Operating system:  **Arch linux**

### Current behavior

Macro.to_string/1 drops the square brackets on a module attribute declaration.

```
iex(3)> Code.string_to_quoted!("@param [1]")  |> Macro.to_string()
"@param 1"
```
Note that it can be worked around with using `Macro.to_string/2`:

```
iex(7)> Code.string_to_quoted!("@param [1]")  |> Macro.to_string(fn _ast, s -> s end)
"@param([1])"
```
Although this gives a warning (#11373)

### Expected behavior

It would be expected that the result of `Macro.to_string` returns an equivalent AST when parsed again; with 1.13 this is not the case.

On Elixir 1.12 it works correctly:

```
iex(5)> Code.string_to_quoted!("@param [1]")  |> Macro.to_string()                                   
"@param([1])"
```

doorgan:
Fixes #11459

### DIFF
--- a/lib/elixir/lib/code/normalizer.ex
+++ b/lib/elixir/lib/code/normalizer.ex
@@ -200,10 +200,15 @@ defmodule Code.Normalizer do
   # Module attributes
   defp do_normalize({:@, meta, [{name, name_meta, [value]}]}, state) do
     value =
-      if is_list(value) do
-        normalize_kw_args(value, state, false)
-      else
-        do_normalize(value, state)
+      cond do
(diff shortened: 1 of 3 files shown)


josevalim:

PR #11466, opened 2021-12-09
Don't show snippet for EOF line

v0idpwn:
Restriction added only for "missing terminator" error + empty line.

We could inform the user that EOF was reached, but not all evaluated elixir code is in a file. For example:

```sh
elixir -e "defmodule Foo do"
```

should show: 
```
** (TokenMissingError) nofile:1:17: missing terminator: end (for "do" starting at line 1)
    |
  1 | defmodule Foo do
    |                 ^
    (elixir 1.14.0-dev) lib/code.ex:403: Code.validated_eval_string/3
```

So I didn't add it. Let me know if you think it is worth it, shouldn't be a big of a change.

### DIFF
--- a/lib/elixir/src/elixir_errors.erl
+++ b/lib/elixir/src/elixir_errors.erl
@@ -87,9 +87,12 @@ snippet(InputString, Location, StartLine, StartColumn) ->
   case lists:keyfind(column, 1, Location) of
     {column, Column} ->
       Lines = string:split(InputString, "\n", all),
-      Snippet = elixir_utils:characters_to_binary(lists:nth(Line - StartLine + 1, Lines)),
+      Snippet = (lists:nth(Line - StartLine + 1, Lines)),
       Offset = if Line == StartLine -> Column - StartColumn; true -> Column - 1 end,
-      #{content => Snippet, offset => Offset};
+      case string:trim(Snippet, leading) of
+        [] -> nil;
+        _ -> #{content => elixir_utils:characters_to_binary(Snippet), offset => Offset}
+      end;
 
     false ->
       nil
--- a/lib/elixir/test/elixir/kernel/errors_test.exs
+++ b/lib/elixir/test/elixir/kernel/errors_test.exs
@@ -468,6 +468,21 @@ defmodule Kernel.ErrorsTest do
     assert_eval_raise TokenMissingError, ~r/nofile:1:3: invalid escape \\ at end of file/, '1 \\'
   end
 
+  test "show snippet on missing tokens" do
+    assert_eval_raise TokenMissingError,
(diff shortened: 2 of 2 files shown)


lib/elixir/src/elixir_errors.erl
```diff
@@ -98,12 +98,24 @@ snippet(InputString, Location, StartLine, StartColumn) ->
...
+  case Error of
+    <<"syntax error before: ">> ->
+      Message = <<"syntax error: expression is incomplete">>,
+      Snippet = snippet(InputString, Location, StartLine, StartColumn),
+      raise(Location, File, 'Elixir.TokenMissingError', Message, Snippet);
+
+    <<"missing terminator: ", _/binary>> ->
+      case snippet(InputString, Location, StartLine, StartColumn) of
+        #{content := <<"">>} ->
+          raise(Location, File, 'Elixir.TokenMissingError', Error);
```
josevalim, line 110: Do we want to have this check inside "missing terminator: "? Or do we want to not show the snippet every time `content := <<"">>`? Maybe we should have `snippet/4` return `nil` if the content is `<<"">>` and then in this function we simply don't care about handling the snippet?

josevalim:
Looks great, I dropped some comments about potential refactorings. We also need a test! :)

josevalim:
On the other hand, I wonder if we should flip this message around? Instead of showing the snippet at the position it errored, maybe the snippet should show on the terminator that started? So we would do this:

```
** (TokenMissingError) nofile:1:15: unmatched "do" (expected a matching "end" but none was found)
    |
  1 | defmodule Foo do
    |               ^
    (elixir 1.14.0-dev) lib/code.ex:403: Code.validated_eval_string/3
```

josevalim:
It is a bit of a longer PR though, so let's merge this as a fix that I can backport and then we work on this new format for main. :)

josevalim:
Actually, ignore me. Take this code:

```elixir
defmodule Foo do
  def bar do

  enda

  def baz do

  end
end
```

It is accusing line 1 but the error is really on line 2. Pointing to line 1 can be confusing here, so I think this PR is the way to go. Let's just see if the refactoring I suggest above makes sense.

lib/elixir/src/elixir_errors.erl
josevalim: I think in general snippet on empty lines are not helpful, as there is nothing to point to. But also unlikely. So I think we can generalize this, yeah. Possible also rule it out if it only has whitespaces. So if `string:trim(String, leading)` is an empty binary, we skip it. WDYT?

v0idpwn: Agreed! Pushed a commit that follows this rule. Also added tests for this behaviour.

josevalim:

PR #11470, opened 2021-12-11
`MapSet.put_all/2`

balexand:
Hi everyone. I feel like this function would be a nice addition to the stdlib. The best alternative that I can find for adding keys in bulk is to call `MapSet.union(set, MapSet.new([3,4]))`. `MapSet.put_all(set, [3,4])` is more concise and readable. Thank you and please let me know if there's anything that I can do to help get this merged.

### DIFF
--- a/lib/elixir/lib/map_set.ex
+++ b/lib/elixir/lib/map_set.ex
@@ -275,6 +275,24 @@ defmodule MapSet do
   end
 
   @doc """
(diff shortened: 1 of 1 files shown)


josevalim:
I would say union is the way to go in this case. There is very little overhead and it also equally gets us there. :) Thank you!

PR #11471, opened 2021-12-11
Set a max line_length for Macro.to_string

### LINKED ISSUE #11469: Wrong new line placement for anonymous function
Take this code:

```elixir
Macro.to_string(
  quote do
    capture_log(fn ->
      {:ok, pid} = Repl.start_link({self(), opts})
      assert Exception.message(error) =~ msg
    end) =~ msg
  end
)
```

It will render as:

```elixir
capture_log(fn -> {:ok, pid} = Repl.start_link({self(), opts})
  assert Exception.message(error) =~ msg end) =~ msg
```

Note the lack of breaks after `->`.

@doorgan, in case you want to take a look! Otherwise I will do it later on. :heart:

doorgan:
Fixes #11469

I'm not 100% about where would be the best place to add a test, if one is needed

### DIFF
--- a/lib/elixir/lib/macro.ex
+++ b/lib/elixir/lib/macro.ex
@@ -957,7 +957,7 @@ defmodule Macro do
   """
   @spec to_string(t()) :: String.t()
   def to_string(tree) do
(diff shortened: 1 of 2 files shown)


josevalim:

PR #11480, opened 2021-12-13
Optimize even further Enum.into implementation

eksperimental:
This commit builds on 79d9c70.

It reduces two levels of indirection,
and it does not call the implementation for lists inside a try/catch.

### DIFF
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -1457,34 +1457,26 @@ defmodule Enum do
   defp into_map(enumerable, collectable),
     do: Enum.reduce(enumerable, collectable, fn {key, val}, acc -> Map.put(acc, key, val) end)
 
-  defp into_protocol(enumerable, collectable) do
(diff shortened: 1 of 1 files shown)


eksperimental:
> We need to keep the list implementation inside a try-catch because calling the into callback itself can fail.

The current implementation is calling `Collectable.into(collectable)` outside the try/catch block. Should we move it in?


> And in the second code the transform function can fail too.

I thought of this, but in other parts of the current implementation we are calling transform without catching it. I thought it was alright to raise there. What should we do?

josevalim:
It is not about catching it, it is making sure the collectable resource has been halt in case of errors. In the cases you mention, there is no "resource".

> The current implementation is calling Collectable.into(collectable) outside the try/catch block. Should we move it in?

We can't do this because the resource we need to call halt on is the one returned by Collectable.

eksperimental:
I see now what you mean.
Anyway I forced pushed the commit leaving just the changes that reduce the indirection, as I found it myself a bit difficult to follow what was going one with the code.

PR #11487, opened 2021-12-15
feat: do not write to failures manifest upon warnings

### LINKED ISSUE #11484: mix test --warnings-as-errors and --failed flags with unexpected interaction
### Environment

```shell
Erlang/OTP 24 [erts-12.1.2] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [jit]

Elixir 1.12.3 (compiled with Erlang/OTP 24)
```

* Operating system: Ubuntu 20.04 (but issue also occurs on CI machine with Alpine 3.14.2)

### Current behavior

Running `mix test --warnings-as-errors || mix test --warnings-as-errors --failed` can cause warnings to fall through the CI workflow.
This is because the `--warnings-as-errors` ends up not running any tests at all, which in turn causes the `--failed` flag on the second command to not return any files to be compiled, resulting in a false success.

For what is worth and future reference, the current local fix is to run it as: `mix test --warnings-as-errors || if [[ $? = 1 ]]; then (exit 1); else mix test --warnings-as-errors --failed; fi`

### Expected behavior

In my opinion this is not desirable behaviour, and perhaps could be fixed by keeping track of a failure due to warnings, effectively marking all related files as failed for the next command.

Thanks in advance!

polvalente:
closes #11484

### DIFF
--- a/lib/ex_unit/lib/ex_unit.ex
+++ b/lib/ex_unit/lib/ex_unit.ex
@@ -251,8 +251,6 @@ defmodule ExUnit do
     * `:failures_manifest_file` - specifies a path to the file used to store failures
       between runs;
 
-    * `:skip_failures_manifest_file` - skips writing to the failures manifest file;
-
     * `:formatters` - the formatters that will print results,
       defaults to `[ExUnit.CLIFormatter]`;
 
--- a/lib/ex_unit/lib/ex_unit.ex
+++ b/lib/ex_unit/lib/ex_unit.ex
@@ -442,11 +442,14 @@ defmodule ExUnit do
       ExUnit.FailuresManifest.read(manifest_filename |> IO.inspect(label: "manifest path"))
       |> IO.inspect(label: "manifest before")
 
-    failed_tests
-    |> IO.inspect(label: "failed tests")
-    |> Enum.reduce(manifest, &ExUnit.FailuresManifest.put_test(&2, &1))
-    |> IO.inspect(label: "manifest after")
-    |> ExUnit.FailuresManifest.write!(manifest_filename)
+    :ok =
+      failed_tests
+      |> IO.inspect(label: "failed tests")
+      |> Enum.reduce(manifest, &ExUnit.FailuresManifest.put_test(&2, &1))
+      |> IO.inspect(label: "manifest after")
+      |> ExUnit.FailuresManifest.write!(manifest_filename)
+
+    {:ok, length(failed_tests)}
   end
 
   # Persists default values in application
--- a/lib/ex_unit/lib/ex_unit.ex
+++ b/lib/ex_unit/lib/ex_unit.ex
@@ -442,14 +442,11 @@ defmodule ExUnit do
       ExUnit.FailuresManifest.read(manifest_filename |> IO.inspect(label: "manifest path"))
       |> IO.inspect(label: "manifest before")
 
-    :ok =
-      failed_tests
-      |> IO.inspect(label: "failed tests")
-      |> Enum.reduce(manifest, &ExUnit.FailuresManifest.put_test(&2, &1))
-      |> IO.inspect(label: "manifest after")
-      |> ExUnit.FailuresManifest.write!(manifest_filename)
-
-    {:ok, length(failed_tests)}
+    failed_tests
+    |> IO.inspect(label: "failed tests")
+    |> Enum.reduce(manifest, &ExUnit.FailuresManifest.put_test(&2, &1))
+    |> IO.inspect(label: "manifest after")
+    |> ExUnit.FailuresManifest.write!(manifest_filename)
   end
 
   # Persists default values in application
--- a/lib/ex_unit/lib/ex_unit.ex
+++ b/lib/ex_unit/lib/ex_unit.ex
@@ -438,14 +438,10 @@ defmodule ExUnit do
         |> Enum.map(&%{&1 | state: {:failed, []}})
       end)
 
-    manifest =
-      ExUnit.FailuresManifest.read(manifest_filename |> IO.inspect(label: "manifest path"))
-      |> IO.inspect(label: "manifest before")
+    manifest = ExUnit.FailuresManifest.read(manifest_filename)
 
     failed_tests
-    |> IO.inspect(label: "failed tests")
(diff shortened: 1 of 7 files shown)


lib/ex_unit/lib/ex_unit.ex
```diff
@@ -251,6 +251,8 @@ defmodule ExUnit do
     * `:failures_manifest_file` - specifies a path to the file used to store failures
       between runs;
 
+    * `:skip_failures_manifest_file` - skips writing to the failures manifest file;
```
polvalente, line 254: This should probably have a better name, because we only skip writing the file, not reading it

lib/mix/test/mix/tasks/test_test.exs
```diff
@@ -510,7 +510,9 @@ defmodule Mix.Tasks.TestTest do
         output = mix(["test", "--warnings-as-errors"])
         assert output =~ "2 failures"
         refute output =~ "Test suite aborted after successful execution"
-        output = mix(["test", "--failed"])
+        # missing assertion here
```
polvalente, line 513: I could not figure out how to test that the actual Mix Task  is calling ExUnit with the correct options

lib/mix/lib/mix/compilers/test.ex
```diff
@@ -45,17 +45,19 @@ defmodule Mix.Compilers.Test do
...
         try do
-          failed? =
+          compilation_failed? =
             case Kernel.ParallelCompiler.require(test_files, parallel_require_callbacks) do
               {:ok, _, [_ | _]} when warnings_as_errors? -> true
               {:ok, _, _} -> false
               {:error, _, _} -> exit({:shutdown, 1})
             end
 
+          :ok = ExUnit.configure(skip_failures_manifest_file: compilation_failed?)
```
josevalim, line 55: Unfortunately there is a race condition here. Maybe the tests finished before we set this option.

I am thinking for now the simplest option may be to get rid of the failed manifest altogether indeed. :( Otherwise we will need to expose full control of the manifest via ExUnit APIs.

polvalente: So we're gonna need something like ExUnit.clear_failed_manifest() to call after everything stops?

Btw, what's the actual race condition?

polvalente: Oh, because await_run wait for the tests to finish, but they are started above?

Could we move the ParallelCompiler code to before the task is started? This should enable this idea to work

josevalim: > Btw, what's the actual race condition?

The race condition is that the tests finish and write to the manifest file before this call happens. :)

> So we're gonna need something like ExUnit.clear_failed_manifest() to call after everything stops?

I think we pass the manifest file to ExUnit, so we can just go ahead and delete it if it is available.

josevalim: > Could we move the ParallelCompiler code to before the task is started? This should enable this idea to work

Then we no longer load and run tests concurrently, it will make the suite slower.

polvalente: Yes. IIRC it resides in the :failed_manifest_file option

polvalente: What if we set the files with warnings as failed in the manifest, instead of deleting it?

josevalim: I like that! The best option so far and it should be just one small API addition.

lib/ex_unit/lib/ex_unit.ex
```diff
@@ -428,6 +428,30 @@ defmodule ExUnit do
     end
   end
 
+  def mark_files_as_failed(modules, allowed_filenames, manifest_filename) do
```
polvalente, line 431: @josevalim This code is currently working as expected through manual tests (I created a local project using the compiled mix/ex_unit versions).

Any pointers on how we could test this? Or is this even the proper way to do it?

polvalente: There's a failing test which I suspect is what needs to be changed, but I'm not sure if it breaking is a bug or indeed just a consequence of the current change.

lib/mix/lib/mix/compilers/test.ex
```diff
@@ -45,24 +45,43 @@ defmodule Mix.Compilers.Test do
...
+                # IO.inspect(second, label: "second")
+                files_with_warnings = Enum.map(warnings, &elem(&1, 0))
+                # IO.inspect(warnings, label: "warnings")
+                # IO.inspect(files_with_warnings, label: "files_with_warnings")
+                :ok  =
+                  ExUnit.mark_files_as_failed(
+                    modules,
+                    files_with_warnings,
+                    opts[:failures_manifest_file]
+                  )
```
josevalim: To avoid the race condition, you need to do this after the suite runs (after `await_run`). Also, is `failures_manifest_file` guaranteed to exist? Or can the user set it to nil somehow?

lib/mix/lib/mix/compilers/test.ex
```diff
@@ -45,24 +45,43 @@ defmodule Mix.Compilers.Test do
...
-            if failed? do
-              message =
-                "\nERROR! Test suite aborted after successful execution due to warnings while using the --warnings-as-errors option"
+          IO.inspect(results, label: "results")
+          IO.inspect(test_files, label: "test_files")
 
-              IO.puts(:stderr, IO.ANSI.format([:red, message]))
-              exit({:shutdown, 1})
-            end
+          if compilation_failed? do
```
josevalim, line 74: The test is failing because before we only emitted this warning if the suite passed and the warnings were emitted. Now you are always printing this warning. If you revert, you should be good.

lib/mix/lib/mix/compilers/test.ex
josevalim: After another look, I am afraid this still won't be enough. The warning might have come from a support file and no test will be marked as dirty.

I am starting to think the best option is to indeed check on the status code and mark this as unresolved. :(

polvalente: Yeah, given all the options, the simplest/cleanest solution seems to be the one I suggested.
WDYT about adding a note on this issue on the help text for `--warnings-as-errors` and/or `--failed`?

PR #11493, opened 2021-12-16
Add MapSet.map/2, MapSet.filter/2, and MapSet.reject/2

palexanderm:


lib/elixir/lib/map_set.ex
```diff
@@ -358,6 +358,106 @@ defmodule MapSet do
...
+
+      iex> MapSet.map(MapSet.new(1..3), fn x -> x * 2 end)
+      #MapSet<[2, 4, 6]>
+
+      iex> MapSet.map(MapSet.new(~w(a b c)a), &to_string/1)
+      #MapSet<["a", "b", "c"]>
+
+  """
+  @doc since: "1.14.0"
+  @spec map(t, (any -> any)) :: t
```
v0idpwn, line 377: Perhaps the typespecs could be made more specific?
```suggestion
  @spec map(t(a), (a -> b)) :: t(b) when a: value, b: value
```

palexanderm: How about `@spec map(t(value), (value -> value)) :: t(value)` to be more consistent with other areas of the code?

josevalim:
```suggestion
  @spec map(t(a), (a -> b)) :: t(b) when a: value, b: value
```

lib/elixir/lib/map_set.ex
```diff
@@ -358,6 +358,106 @@ defmodule MapSet do
...
+
+      iex> MapSet.filter(MapSet.new(1..5), fn x -> x > 3 end)
+      #MapSet<[4, 5]>
+
+      iex> MapSet.filter(MapSet.new(["a", :b, "c"]), &is_atom/1)
+      #MapSet<[:b]>
+
+  """
+  @doc since: "1.14.0"
+  @spec filter(t, (any -> as_boolean(term))) :: t
```
josevalim, line 408:
```suggestion
  @spec filter(t(a), (a -> as_boolean(term))) :: t(a) when a: value
```

lib/elixir/lib/map_set.ex
```diff
@@ -358,6 +358,106 @@ defmodule MapSet do
...
+
+      iex> MapSet.reject(MapSet.new(1..5), fn x -> rem(x, 2) != 0 end)
+      #MapSet<[2, 4]>
+
+      iex> MapSet.reject(MapSet.new(["a", :b, "c"]), &is_atom/1)
+      #MapSet<["a", "c"]>
+
+  """
+  @doc since: "1.14.0"
+  @spec reject(t, (any -> as_boolean(term))) :: t
```
josevalim, line 443:
```suggestion
  @spec reject(t(a), (a -> as_boolean(term))) :: t(a) when a: value
```

josevalim:

PR #11498, opened 2021-12-17
Improve specs and docs in Macro.underscore/camelize

eksperimental:


lib/elixir/lib/macro.ex
```diff
@@ -1815,12 +1815,13 @@ defmodule Macro do
   end
 
   @doc """
-  Converts the given atom or binary to underscore format.
+  Converts `module` to a string with underscore-slash format.
```
josevalim, line 1818: I don't think we should use `module`. If you say it is a module, I would expect it to be something previous defined with `defmodule`. Not only that, it also accept `:FooBar` or "FooBar", and none of them are modules. It is really an atom or a string.

eksperimental: I went with `module_name` but then decided to go with module.

Should we go with `atom_or_string`?

josevalim: > Should we go with atom_or_string?

Sounds good to me!

eksperimental: Please have a look at the changes.

josevalim:

PR #11499, opened 2021-12-17
Remove double when on TODO comment

WLSF:
I'm not sure if this is right, so I'm leaving it here as a suggestion, but it seems like there is a double "when" on the TODO comment.


josevalim:

eksperimental:
Thank you @WLSF

PR #11500, opened 2021-12-17
Make Base.decode16's odd-length message more helpful

maco:
As discussed on today's livestream, what to do about "odd-length string" may not be immediately obvious.

### DIFF
--- a/lib/elixir/lib/base.ex
+++ b/lib/elixir/lib/base.ex
@@ -349,7 +349,8 @@ defmodule Base do
   end
 
   def decode16!(string, _opts) when is_binary(string) do
-    raise ArgumentError, "odd-length string"
+    raise ArgumentError,
+          "odd-length string:\nAn even number of bytes was expected, but got #{byte_size(string)}.\nDo you need to String.trim?"
   end
 
   @doc """
@@ -928,7 +929,8 @@ defmodule Base do
           <<main::bits, unquote(fun)(c1)::4, unquote(fun)(c2)::4>>
 
         <<_::8>> ->
-          raise ArgumentError, "odd-length string"
+          raise ArgumentError,
+                "odd-length string:\nAn even number of bytes was expected, but got #{byte_size(string)}.\nDo you need to String.trim?"
 
         <<>> ->
           main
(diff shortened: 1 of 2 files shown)


lib/elixir/lib/base.ex
```diff
@@ -349,7 +349,8 @@ defmodule Base do
   end
 
   def decode16!(string, _opts) when is_binary(string) do
-    raise ArgumentError, "odd-length string"
+    raise ArgumentError,
+          "odd-length string:\nAn even number of bytes was expected, but got #{byte_size(string)}.\nDo you need to String.trim?"
```
josevalim, line 353:
```suggestion
          "string given to decode has wrong length. An even number of bytes was expected, got: #{byte_size(string)}. " <>
            "Double check your string for unwanted characters or pad it accordingly"
```

lib/elixir/lib/base.ex
```diff
@@ -928,7 +929,8 @@ defmodule Base do
           <<main::bits, unquote(fun)(c1)::4, unquote(fun)(c2)::4>>
 
         <<_::8>> ->
-          raise ArgumentError, "odd-length string"
+          raise ArgumentError,
+                "odd-length string:\nAn even number of bytes was expected, but got #{byte_size(string)}.\nDo you need to String.trim?"
```
josevalim, line 933:
```suggestion
                "string given to decode has wrong length. An even number of bytes was expected, got: #{byte_size(string)}. " <>
                  "Double check your string for unwanted characters or pad it accordingly"
```

lib/elixir/test/elixir/base_test.exs
```diff
@@ -79,9 +79,11 @@ defmodule BaseTest do
...
 
   test "decode16!/1 errors odd-length string" do
-    assert_raise ArgumentError, "odd-length string", fn ->
-      decode16!("666")
-    end
+    assert_raise ArgumentError,
+                 "odd-length string:\nAn even number of bytes was expected, but got 3.\nDo you need to String.trim?",
+                 fn ->
+                   decode16!("666")
+                 end
```
josevalim, line 86:
```suggestion
    assert_raise ArgumentError, ~r/string given to decode has wrong length/, fn ->
      decode16!("666")
    end
```

josevalim:

josevalim:
Thank you! I didn't go with the String.trim suggestion because it is really hard to say if that's the correct fix. It can, for example, be a new line in the middle of the string, which String.trim wouldn't catch either. :)

PR #11502, opened 2021-12-18
Remove special case for symlinks on rm_rf

josevalim:


josevalim:

PR #11503, opened 2021-12-18
Giving anonymous blame output a meaningful comparison to what was given

krainboltgreene:
So as per the test you can see failing in this pull request whenever you define a pattern or guard for an anonymous function, a message comes along that doesn't match, and no match is found you get a blame message that is woefully under-descriptive, to the point where when this happened to me I had almost no idea where it was happening.

If this is desired I'll finish up the passing portion of this.

### DIFF
--- a/lib/elixir/test/elixir/exception_test.exs
+++ b/lib/elixir/test/elixir/exception_test.exs
@@ -446,6 +446,22 @@ defmodule ExceptionTest do
 
                  def fetch(-%module{} = container-, key)
              """
+
+      assert blame_message(Access, & &1.fetch(:foo, :bar)) =~ """
+             no function clause matching in Access.fetch/2
+
+             The following arguments were given to Access.fetch/2:
+
+                 # 1
+                 :foo
+
+                 # 2
+                 :bar
+
+             Attempted function clauses (showing 5 out of 5):
+
+                 x
+             """
     end
 
     test "annotates undefined function error with suggestions" do
--- a/lib/elixir/test/elixir/exception_test.exs
+++ b/lib/elixir/test/elixir/exception_test.exs
@@ -448,9 +448,9 @@ defmodule ExceptionTest do
              """
 
       assert blame_message(%{"a" => "b"}, fn %{"b" => "c"} -> "d" end) =~ """
-             no function clause matching in anonymous fn/1 in ExceptionTest.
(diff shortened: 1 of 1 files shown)


josevalim:
Hi!

Unfortunately this is very hard to achieve in practice. The way this works for function definitions is that we need to traverse the AST and find where the function is defined. This works somewhat easily for functions because they are always at the root and they are identified by their name and arity. Anonymous functions do not have those properties, so it would require changes to the compiler to somehow tag them, in a way we can recover them at runtime in a somewhat fast fashion. For example, doing so per line wouldn’t be enough: we would need to traverse the whole file to find the line and there could be more than one.

So if you want to send a PR, it is welcome, but bear in mind it is a non trivial amount of compiler work. I will leave this open for the next 24h or so, in case someone does have a suggestion to make this trivial, but I hope this explains why this feature doesn’t currently exist and why it is complex to do so.

krainboltgreene:
@josevalim So I did discover that, however I noticed that `Function.info(x, :env)` does give *something* of the arguments, but in a weird structure. I can feasibly transform that datastructure into something resembling an AST for `Macro.to_string()`, thoughts?

josevalim:
In IEx it doesn’t “count”. The code there is evaluated , so it keeps the AST for evaluation. In actual code, there is no AST in the env, so a very different solution is needed.

krainboltgreene:
Does anyone have an example of similar "tagging" work that I could base mine on?

josevalim:
One idea I had was to change the Elixir->Erlang pass to generate Erlang's named anonymous functions and investigate if the name sticks to the stacktrace. The name should have the parent function and a counter. If it works, then we can try to reconstruct the function name from Exception.blame/2 and traverse the AST to find it.

krainboltgreene:
Feel free to close for now, I'm going to investigate this on my own, but that's going to require learning how elixir-lang/elixir works.

josevalim:
After trying a couple implementations of this approach, I concluded the implementation is unfortunately non trivial, especially because anonymous functions are closures and then doing the blame on top of the closure state can get very complicated, unfortunately. The same applies to `case` and other constructs, so I don't believe we can add this for now. :(

krainboltgreene:
Yeah the only solution I came up with was 🪄 magic comments 🌟  and that felt like a nightmare for the future.

PR #11506, opened 2021-12-19
Allow bypassing application mode validation in release spec

thalesmg:
Today, there is a mode validation check when doing a Mix Release that
prevents a parent application that has an application mode of
`:permanent`, for example, while a child application has mode `:load`,
as it might be unsafe.

However, some complex applications may need more control over the
application load/start order.  For such cases, the user would need a
way to tell Mix.Release to don't be strict while constructing the
`.rel` file.

Here, we add an option `:skip_mode_validation_for` to the release
options (default: `[]`) which disables such strict check to specified
applicaations.  The default value ensures the current behavior is
preserved.

### DIFF
--- a/lib/mix/lib/mix/release.ex
+++ b/lib/mix/lib/mix/release.ex
@@ -626,7 +626,7 @@ defmodule Mix.Release do
       for {app, mode} <- modes do
         properties = Map.get(apps, app) || throw({:error, "Unknown application #{inspect(app)}"})
         children = Keyword.get(properties, :applications, [])
-        app not in skip_mode_validation_for && validate_mode!(app, mode, modes, children)
+        app in skip_mode_validation_for || validate_mode!(app, mode, modes, children)
         build_app_for_release(app, mode, properties)
       end
 
--- a/lib/mix/lib/mix/release.ex
+++ b/lib/mix/lib/mix/release.ex
@@ -617,7 +617,7 @@ defmodule Mix.Release do
       options: options
     } = release
 
-    skip_mode_validation_index =
+    skip_mode_validation_for =
       options
       |> Keyword.get(:skip_mode_validation_for, [])
       |> MapSet.new()
@@ -626,10 +626,7 @@ defmodule Mix.Release do
       for {app, mode} <- modes do
         properties = Map.get(apps, app) || throw({:error, "Unknown application #{inspect(app)}"})
         children = Keyword.get(properties, :applications, [])
-
-        if not MapSet.member?(skip_mode_validation_index, app),
-          do: validate_mode!(app, mode, modes, children)
-
+        app not in skip_mode_validation_for && validate_mode!(app, mode, modes, children)
         build_app_for_release(app, mode, properties)
       end
 
--- a/lib/mix/lib/mix/release.ex
+++ b/lib/mix/lib/mix/release.ex
@@ -609,13 +609,21 @@ defmodule Mix.Release do
   end
 
   defp build_release_spec(release, modes) do
-    %{name: name, version: version, erts_version: erts_version, applications: apps} = release
+    %{
+      name: name,
+      version: version,
+      erts_version: erts_version,
+      applications: apps,
+      options: options
+    } = release
+
+    validate_application_mode? = Keyword.get(options, :validate_application_mode?, true)
 
     rel_apps =
       for {app, mode} <- modes do
         properties = Map.get(apps, app) || throw({:error, "Unknown application #{inspect(app)}"})
         children = Keyword.get(properties, :applications, [])
-        validate_mode!(app, mode, modes, children)
+        if validate_application_mode?, do: validate_mode!(app, mode, modes, children)
         build_app_for_release(app, mode, properties)
       end
 
--- a/lib/mix/lib/mix/release.ex
+++ b/lib/mix/lib/mix/release.ex
@@ -617,13 +617,19 @@ defmodule Mix.Release do
       options: options
     } = release
 
-    validate_application_mode? = Keyword.get(options, :validate_application_mode?, true)
+    skip_mode_validation_index =
+      options
+      |> Keyword.get(:skip_mode_validation_for, [])
+      |> MapSet.new()
 
     rel_apps =
       for {app, mode} <- modes do
         properties = Map.get(apps, app) || throw({:error, "Unknown application #{inspect(app)}"})
         children = Keyword.get(properties, :applications, [])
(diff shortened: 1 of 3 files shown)


josevalim:
Hi @thalesmg, thank you for the PR. Can you please give detailed context of when this option may be necessary and why?

A release that violates this may not boot at all.

thalesmg:
> Hi @thalesmg, thank you for the PR. Can you please give detailed context of when this option may be necessary and why?
> 
> A release that violates this may not boot at all.

Hi @josevalim ! Sure!  Let me try to give a minimal example.
 
Say our release consists of several application, and let's pick a few special ones: `:common_app`, `:booter`, `later_app`.  `:later_app` must be started last, and it depends on `:common_app` (so `:common_app` should appear in its `:application` key when consulted by `:application.get_key(:later_app, :applications)`).  But, due to the complexity of the application, it's `:booter` that takes care of properly starting up `:common_app`.

So, under the current strict validation, if our release has:

```elixir
[
  relase_foo: [
    applications: [
      # ...
      common_app: :load,     # <-- only loaded
      # ...
      booter: :permanent,    # <-- starts `:common_app` during its boot
      # ...
      later_app: :permanent  # <-- depends on `:common_app`
    ],
    # ...
```

While _in general_ it indeed would not be safe to allow such configuration, some complex applications may work like that.  In this example, even though `:latter_app` is `:permanent` and depends on `:common_app` (which is `:load), it'll boot fine because `:booter` will have started `:common_app` by the time `:later_app` is booted.  

(More concretely, I'm trying to "mixify" [emqx](https://github.com/emqx/emqx/blob/b2e7a6a24938743d41963b2bc1177fdeb9dc3c5d/rebar.config.erl#L251-L261), and it is such case where the startup of some application is controlled by intermediary ones in the middle of the release boot order.  I have just disabled the strict validation as proposed in this PR, and the application booted fine using the boot order I linked.)

thalesmg:
While it's generally unsafe to allow such configurations or it may be hard to statically verify that the release does boot, the idea here is more about giving the user the possibility of skipping this strict check if they really want to.

josevalim:
Thank you. What about a `skip_mode_validation_for: [app]` and you pass a list of apps you want to skip this on instead of a blanket disable all check.

thalesmg:
> Thank you. What about a `skip_mode_validation_for: [app]` and you pass a list of apps you want to skip this on instead of a blanket disable all check.

Good idea!  That can be used to achieve the same effect with a higher degree of control.

I'll change this PR to reflect this.

lib/mix/lib/mix/release.ex
```diff
@@ -617,13 +617,19 @@ defmodule Mix.Release do
...
 
     rel_apps =
       for {app, mode} <- modes do
         properties = Map.get(apps, app) || throw({:error, "Unknown application #{inspect(app)}"})
         children = Keyword.get(properties, :applications, [])
-        if validate_application_mode?, do: validate_mode!(app, mode, modes, children)
+
+        if not MapSet.member?(skip_mode_validation_index, app),
+          do: validate_mode!(app, mode, modes, children)
+
```
josevalim, line 632:
```suggestion
        if app not in skip_mode_validation_index, do: validate_mode!(app, mode, modes, children)
```

lib/mix/lib/mix/release.ex
```diff
@@ -609,13 +609,27 @@ defmodule Mix.Release do
...
 
     rel_apps =
       for {app, mode} <- modes do
         properties = Map.get(apps, app) || throw({:error, "Unknown application #{inspect(app)}"})
         children = Keyword.get(properties, :applications, [])
-        validate_mode!(app, mode, modes, children)
+
+        if not MapSet.member?(skip_mode_validation_index, app),
+          do: validate_mode!(app, mode, modes, children)
+
```
josevalim, line 632:
```suggestion
        if app not in skip_mode_validation_index, do: validate_mode!(app, mode, modes, children)
```

lib/mix/lib/mix/release.ex
```diff
@@ -609,13 +609,27 @@ defmodule Mix.Release do
...
-    %{name: name, version: version, erts_version: erts_version, applications: apps} = release
+    %{
+      name: name,
+      version: version,
+      erts_version: erts_version,
+      applications: apps,
+      options: options
+    } = release
+
+    skip_mode_validation_index =
```
josevalim, line 620:
```suggestion
    skip_mode_validation =
```

lib/mix/lib/mix/release.ex
```diff
@@ -609,13 +609,24 @@ defmodule Mix.Release do
...
+      options
+      |> Keyword.get(:skip_mode_validation_for, [])
+      |> MapSet.new()
 
     rel_apps =
       for {app, mode} <- modes do
         properties = Map.get(apps, app) || throw({:error, "Unknown application #{inspect(app)}"})
         children = Keyword.get(properties, :applications, [])
-        validate_mode!(app, mode, modes, children)
+        app not in skip_mode_validation_for && validate_mode!(app, mode, modes, children)
```
josevalim, line 629:
```suggestion
        app in skip_mode_validation_for || validate_mode!(app, mode, modes, children)
```

lib/mix/lib/mix/tasks/release.ex
```diff
@@ -455,6 +455,12 @@ defmodule Mix.Tasks.Release do
     * `:steps` - a list of steps to execute when assembling the release. See
       the "Steps" section for more information.
 
+    * `:skip_mode_validation_for` - a list of application names
+      (atoms) specifying applications to skip strict validation of
+      "unsafe" modes. An "unsafe" case is when a parent application
+      mode is `:permanent` but one of the applications it depends on
+      is set to `:load`, for example. Defaults to `[]`.
```
josevalim, line 462:
```suggestion
      is set to `:load`. Use this with care, as a release with
      invalid modes may no longer boot without additional tweaks.
      Defaults to `[]`.
```

josevalim:

thalesmg:
Thanks for accepting the PR and for the discussion (especially during a Sunday night)! :beers: :beers:

thalesmg:
By the way, should I port this to some older versions (1.12, 1.11, ..., 1.9)?

josevalim:
New features generally make only to the next minor release but I can backport this to v1.13.

thalesmg:
> New features generally make only to the next minor release but I can backport this to v1.13.

Awesome!  I can do that, if you want.  I'll prepare a PR against `v1.13` soon, then.  :beers:

josevalim:
I already did it, no worries!

PR #11507, opened 2021-12-19
Check plugins first so they can format `.ex` and `.exs` files

marcandre:
Would you consider allowing plugins to override the builtin formatter for `.ex`/`.exs` files?


whatyouhide:
I like this idea, after all most users won't have custom plugins for `.ex` and `.exs` files so this is essentially not gonna change anything for them.

josevalim:

PR #11517, opened 2021-12-22
Streamline t:Enumerable.t/1 docs

eksperimental:
1. It replaces the your of "I" pronoun with "You" which is the standard in the docs.
2. It replaces "value" with "element" which is the terminology and type used in Enum.
3. It replace "ints" with "integers" to make example more readable.

/cc @whatyouhide


whatyouhide:
Thanks @eksperimental! 💟

PR #11519, opened 2021-12-22
Warn on 0-arity callbacks inside protocols

thiamsantos:
## Motivation

During the upgrade to elixir v1.13 I got a compilation error caused by a 0-arity callback defined inside a protocol. Since  #10912 the expected behavior in these scenarios is to show a warning for the callbacks inside the protocol, seems that we don't yet support 0-arity callbacks defined without parens.

Compilation error:
```
** (ArgumentError) errors were found at the given arguments:

  * 1st argument: not a list

    :erlang.length(nil)
    (elixir 1.13.1) lib/protocol.ex:726: Protocol.callback_ast_to_fa/1
    (stdlib 3.17) lists.erl:1254: :lists.flatmap/2
    (elixir 1.13.1) lib/protocol.ex:742: Protocol.callback_metas/2
    (elixir 1.13.1) lib/protocol.ex:761: Protocol.__before_compile__/1
    (stdlib 3.17) lists.erl:1267: :lists.foldl/3
```

Example of a module that triggers the compilation error:

```elixir
defprotocol MyApp.Foo do
  @callback bar :: String.t()
end
````

## Proposed solution

Use `List.wrap/1` when building the warning of callbacks inside a protocol, in order to handle the scenario when `args` is `nil` caused by a `callback` defined without parens.

### DIFF
--- a/lib/elixir/lib/protocol.ex
+++ b/lib/elixir/lib/protocol.ex
@@ -731,14 +731,14 @@ defmodule Protocol do
 
   defp callback_ast_to_fa({kind, {:"::", meta, [{name, _, args}, _return]}, _pos})
        when kind in [:callback, :macrocallback] do
-    [{{name, length(args)}, meta}]
+    [{{name, length(List.wrap(args))}, meta}]
   end
 
   defp callback_ast_to_fa(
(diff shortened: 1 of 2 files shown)


josevalim:
Beautifully done! :green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

PR #11520, opened 2021-12-22
Add t:Enum.t/1

eksperimental:
This is to follow along with the recent introduction of `t:Enumerable.t/1`.

I think it is needed as some people may currently be relying on `t:Enum.t/0` instead of directly calling `t:Enumerable.t/1` ([we were doing it ourselves](https://github.com/elixir-lang/elixir/commit/b1414ee1d0d78a6681303d24a07e2254bcb76e39)) it will have people mixing up `t:Enumerable.t/1` and `t:Enum.t/0` within the same module provided they want to use `t:Enumerable.t/1`, and it could end up confusing both for developers and end-users.

So let's let  developers rely on `t:Enum.t/0` and `t:Enum.t/1` for convenience and minimum code change.

/cc @whatyouhide

### DIFF
--- a/lib/elixir/lib/enum.ex
+++ b/lib/elixir/lib/enum.ex
@@ -270,6 +270,14 @@ defmodule Enum do
   @compile :inline_list_funcs
 
   @type t :: Enumerable.t()
+
+  @typedoc """
+  An enumerable of elements of type `element`.
+
+  See `t:Enumerable.t/1` for information."
+  """
+  @type t(element) :: Enumerable.t(element)
+
   @type acc :: any
   @type element :: any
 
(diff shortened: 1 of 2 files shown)


josevalim:
I think adding Enum.t was a mistake but not a big enough to justify deprecating it. For most protocols, people will depend on the protocol, and that’s what we should mirror here.

eksperimental:
I think its deprecation should be considered then.
As, IMO, it is the GO TO module where everyone goes and look at the source code along with Kernel.
I  had taken myself Enum as an inspiration for a protocol for other project and I have replicated this pattern.

eksperimental:
Well, deprecating will be too much and we will be hated for that.
Maybe leave a comment in the `@typedoc` saying that using `t:Enumerable.t/0` directly is preferable. WDYT?

josevalim:
I think it is fine as is. I consider it more of a shortcut so you don’t have to type Enumerable.t all the time and that’s the best way we can do this shortcut at the moment.

Other languages have the concept of type aliases for this.

Issue #11318, opened 2021-10-18
Elixir builds on GitHub Actions


wojtekmach:
Currently, on every Elixir GitHub release, there's a `Precompiled.zip` build. That build is against an oldest supported OTP version for that release.

I'd like to propose to build Elixir against all major OTP versions that a given release supports to make sure we can use all the OTP features.

For example, for Elixir v1.13, we'd have:

```
elixir-1.13.0-otp-22.zip
elixir-1.13.0-otp-23.zip
elixir-1.13.0-otp-24.zip
```

We should always build against OTP MAJOR.0, given, say, code compiled on OTP 24.1 may not work optimally on OTP 24.0.

We'd attach these builds to every Elixir release going forward. We could even do it for Elixir master, even compile it against OTP master, but it is a bit tricky because we need a tag and a release to attach these to so let's leave that out of the scope for now. (One idea is we'd have a `master-nightly` tag that we update nightly.)

The release assets would be added after the release is published. Here's how we can do that on GitHub Actions: https://github.com/elixir-nx/xla/blob/9307f1a4eaeeb72a3474b6fa4461e44d40b15dbc/.github/workflows/release.yml#L37:L39. We'd need to make sure we update the release notes with the SHAs of the uploaded artifacts.

Worth mentioning that Hex Bob is already providing [Elixir builds](https://github.com/hexpm/bob#elixir-builds). Bob will keep making Elixir builds but that functionality is becoming deprecated.

princemaple:
There is an official github action for uploading assets, if you want to separate building and uploading.

https://github.com/elixir-lsp/elixir-ls/blob/master/.github/workflows/release-asset.yml#L78-L87

michallepicki:
> There is an official github action for uploading assets
The [actions/upload-release-asset](https://github.com/actions/upload-release-asset) action repository is archived and the action is not maintained anymore.

> The release assets would be added after the release is published
Alternatively, instead of having the workflow be triggered by a new release like Nx does, a pushed tag could be the release workflow trigger. Then the workflow itself could create a *draft* release and upload the built artifacts there. When it is done someone from the team with write access can see the draft on the `/releases` page, fill in release notes and hit Publish. The advantage of such approach is that when it is published, it already has the artifacts. [Gleam](https://github.com/gleam-lang/gleam/blob/16ea8364407f8e4d748c1ba07f12db9bfb3dad2a/.github/workflows/release.yaml#L21) has such workflow set up.

[...]

PR #11522, opened 2021-12-23
Pre-built Elixir release

wingyplus:
Trying to tackle #11318. This PR create a new workflow to build and publish asset when GitHub Release is created. Here it's summarized: 

* When the tag is created, the workflow will create a draft release start building Elixir.
* Build Elixir with 3 OTP versions, 22, 23, and 24. And upload assets to the release. Those OTP releases is the latest of OTP in each release.
* Build Docs.zip against OTP 24 and upload it to the release.
* Generate `sha1.txt` and `sha256.txt` for the checksum of zip files.

Here is the workflow [1]: 

Screen Shot 2564-12-29 at 23 35 34

And the release [2]:

Screen Shot 2564-12-29 at 23 38 20

[1] https://github.com/wingyplus/elixir/actions/runs/1634690409
[2] https://github.com/wingyplus/elixir/releases/tag/v1.14.0-dev

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -23,15 +23,18 @@ jobs:
             --notes '' \
             --draft \
             ${{ github.ref_name }}
-  release_pre_built:
+  build_zips:
     needs: create_draft_release
     strategy:
       fail-fast: true
       matrix:
         include:
-          - otp_release: 22
-          - otp_release: 23
-          - otp_release: 24
+          - otp: 22
+            otp_version: 22.3
+          - otp: 23
+            otp_version: 23.3
+          - otp: 24
+            otp_version: 24.1
             build_docs: build_docs
     runs-on: ubuntu-18.04
     steps:
@@ -40,25 +43,18 @@ jobs:
           fetch-depth: 50
       - uses: erlef/setup-beam@v1
         with:
-          # The otp-version is the latest version of the release. It doesn't guarantee
-          # the minor (or patch) version when running this workflow.
-          otp-version: ${{ matrix.otp_release }}
+          otp-version: ${{ matrix.otp_version }}
+          version-type: strict
       - name: Build Elixir Release
         run: |
           vsn=$(cat VERSION)
           make Precompiled.zip
-          mv Precompiled-v${vsn}.zip elixir-${vsn}-otp-${{ matrix.otp_release }}.zip
-          shasum -a 1 elixir-${vsn}-otp-${{ matrix.otp_release }}.zip > elixir-${vsn}-otp-${{ matrix.otp_release }}.zip.sha1
-          shasum -a 256 elixir-${vsn}-otp-${{ matrix.otp_release }}.zip > elixir-${vsn}-otp-${{ matrix.otp_release }}.zip.sha256
+          mv Precompiled-v${vsn}.zip elixir-otp-${{ matrix.otp }}.zip
           echo "$PWD/bin" >> $GITHUB_PATH
-      - name: Upload Pre-built
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: |
-          vsn=$(cat VERSION)
-          gh release upload --clobber "v${vsn}" \
-            elixir-${vsn}-otp-${{ matrix.otp_release }}.zip \
-            elixir-${vsn}-otp-${{ matrix.otp_release }}.zip.sha{1,256}
+      - uses: actions/upload-artifact@v2
+        with:
+          name: elixir-otp-${{ matrix.otp }}.zip
+          path: elixir-otp-${{ matrix.otp }}.zip
       - uses: actions/checkout@v2
         with:
           repository: elixir-lang/ex_doc
@@ -76,14 +72,53 @@ jobs:
         run: |
           vsn=$(cat VERSION)
           make Docs.zip
-          shasum -a 1 "Docs-v${vsn}.zip" > Docs-v${vsn}.zip.sha1
-          shasum -a 256 "Docs-v${vsn}.zip" > Docs-v${vsn}.zip.sha256
-      - name: Upload Docs
+          mv Docs-v${vsn}.zip Docs.zip
+      - uses: actions/upload-artifact@v2
+        with:
+          name: Docs.zip
+          path: Docs.zip
         if: ${{ matrix.build_docs }}
+  upload_assets:
+    needs: build_zips
+    runs-on: ubuntu-18.04
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 50
+      - uses: actions/download-artifact@v2
+        with:
+          name: elixir-otp-22.zip
+      - uses: actions/download-artifact@v2
+        with:
+          name: elixir-otp-23.zip
+      - uses: actions/download-artifact@v2
+        with:
+          name: elixir-otp-24.zip
+      - uses: actions/download-artifact@v2
+        with:
+          name: Docs.zip
+      - name: Rename files
+        run: |
+          vsn=$(cat VERSION)
+          mv elixir-otp-22.zip elixir-${vsn}-otp-22.zip
+          mv elixir-otp-23.zip elixir-${vsn}-otp-23.zip
+          mv elixir-otp-24.zip elixir-${vsn}-otp-24.zip
+      - name: Generate checksum
+        run: |
+          vsn=$(cat VERSION)
+          for algo in 1 256
+          do
+            shasum -a ${algo} elixir-${vsn}-otp-22.zip elixir-${vsn}-otp-23.zip elixir-${vsn}-otp-24.zip Docs.zip > "sha${algo}sums.txt"
+          done
+      - name: Upload assets
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           vsn=$(cat VERSION)
-          gh release upload --clobber "v${vsn}" \
-            Docs-v${vsn}.zip \
-            Docs-v${vsn}.zip.sha{1,256}
+          gh release upload --clobber v${vsn} \
+            elixir-${vsn}-otp-22.zip \
+            elixir-${vsn}-otp-23.zip \
+            elixir-${vsn}-otp-24.zip \
+            Docs.zip \
+            sha1sums.txt \
+            sha256sums.txt
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -23,18 +23,15 @@ jobs:
             --notes '' \
             --draft \
             ${{ github.ref_name }}
-  build_zips:
+  release_pre_built:
     needs: create_draft_release
     strategy:
       fail-fast: true
       matrix:
         include:
-          - otp: 22
-            otp_version: 22.3
-          - otp: 23
-            otp_version: 23.3
-          - otp: 24
-            otp_version: 24.1
+          - otp_release: 22
+          - otp_release: 23
+          - otp_release: 24
             build_docs: build_docs
     runs-on: ubuntu-18.04
     steps:
@@ -43,18 +40,25 @@ jobs:
           fetch-depth: 50
       - uses: erlef/setup-beam@v1
         with:
-          otp-version: ${{ matrix.otp_version }}
-          version-type: strict
+          # The otp-version is the latest version of the release. It doesn't guarantee
+          # the minor (or patch) version when running this workflow.
+          otp-version: ${{ matrix.otp_release }}
       - name: Build Elixir Release
         run: |
           vsn=$(cat VERSION)
           make Precompiled.zip
-          mv Precompiled-v${vsn}.zip elixir-otp-${{ matrix.otp }}.zip
+          mv Precompiled-v${vsn}.zip elixir-${vsn}-otp-${{ matrix.otp_release }}.zip
+          shasum -a 1 elixir-${vsn}-otp-${{ matrix.otp_release }}.zip > elixir-${vsn}-otp-${{ matrix.otp_release }}.zip.sha1
+          shasum -a 256 elixir-${vsn}-otp-${{ matrix.otp_release }}.zip > elixir-${vsn}-otp-${{ matrix.otp_release }}.zip.sha256
           echo "$PWD/bin" >> $GITHUB_PATH
-      - uses: actions/upload-artifact@v2
-        with:
-          name: elixir-otp-${{ matrix.otp }}.zip
-          path: elixir-otp-${{ matrix.otp }}.zip
+      - name: Upload Pre-built
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          vsn=$(cat VERSION)
+          gh release upload --clobber "v${vsn}" \
+            elixir-${vsn}-otp-${{ matrix.otp_release }}.zip \
+            elixir-${vsn}-otp-${{ matrix.otp_release }}.zip.sha{1,256}
       - uses: actions/checkout@v2
         with:
           repository: elixir-lang/ex_doc
@@ -72,53 +76,14 @@ jobs:
         run: |
           vsn=$(cat VERSION)
           make Docs.zip
-          mv Docs-v${vsn}.zip Docs.zip
-      - uses: actions/upload-artifact@v2
-        with:
-          name: Docs.zip
-          path: Docs.zip
+          shasum -a 1 "Docs-v${vsn}.zip" > Docs-v${vsn}.zip.sha1
+          shasum -a 256 "Docs-v${vsn}.zip" > Docs-v${vsn}.zip.sha256
+      - name: Upload Docs
         if: ${{ matrix.build_docs }}
-  upload_assets:
-    needs: build_zips
-    runs-on: ubuntu-18.04
-    steps:
-      - uses: actions/checkout@v2
-        with:
-          fetch-depth: 50
-      - uses: actions/download-artifact@v2
-        with:
-          name: elixir-otp-22.zip
-      - uses: actions/download-artifact@v2
-        with:
-          name: elixir-otp-23.zip
-      - uses: actions/download-artifact@v2
-        with:
-          name: elixir-otp-24.zip
-      - uses: actions/download-artifact@v2
-        with:
-          name: Docs.zip
-      - name: Rename files
-        run: |
-          vsn=$(cat VERSION)
-          mv elixir-otp-22.zip elixir-${vsn}-otp-22.zip
-          mv elixir-otp-23.zip elixir-${vsn}-otp-23.zip
-          mv elixir-otp-24.zip elixir-${vsn}-otp-24.zip
-      - name: Generate checksum
-        run: |
-          vsn=$(cat VERSION)
-          for algo in 1 256
-          do
-            shasum -a ${algo} elixir-${vsn}-otp-22.zip elixir-${vsn}-otp-23.zip elixir-${vsn}-otp-24.zip Docs.zip > "sha${algo}sums.txt"
-          done
-      - name: Upload assets
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           vsn=$(cat VERSION)
-          gh release upload --clobber v${vsn} \
-            elixir-${vsn}-otp-22.zip \
-            elixir-${vsn}-otp-23.zip \
-            elixir-${vsn}-otp-24.zip \
-            Docs.zip \
-            sha1sums.txt \
-            sha256sums.txt
+          gh release upload --clobber "v${vsn}" \
+            Docs-v${vsn}.zip \
+            Docs-v${vsn}.zip.sha{1,256}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -29,9 +29,12 @@ jobs:
       fail-fast: true
       matrix:
         include:
-          - otp_release: 22
-          - otp_release: 23
-          - otp_release: 24
+          - otp: 22
+            otp_version: 22.3
+          - otp: 23
+            otp_version: 23.3
+          - otp: 24
+            otp_version: 24.1
             build_docs: build_docs
     runs-on: ubuntu-18.04
     steps:
@@ -40,16 +43,15 @@ jobs:
           fetch-depth: 50
       - uses: erlef/setup-beam@v1
         with:
-          # The otp-version is the latest version of the release. It doesn't guarantee
-          # the minor (or patch) version when running this workflow.
-          otp-version: ${{ matrix.otp_release }}
+          otp-version: ${{ matrix.otp_version }}
+          version-type: strict
       - name: Build Elixir Release
         run: |
           vsn=$(cat VERSION)
           make Precompiled.zip
-          mv Precompiled-v${vsn}.zip elixir-${vsn}-otp-${{ matrix.otp_release }}.zip
-          shasum -a 1 elixir-${vsn}-otp-${{ matrix.otp_release }}.zip > elixir-${vsn}-otp-${{ matrix.otp_release }}.zip.sha1
-          shasum -a 256 elixir-${vsn}-otp-${{ matrix.otp_release }}.zip > elixir-${vsn}-otp-${{ matrix.otp_release }}.zip.sha256
+          mv Precompiled-v${vsn}.zip elixir-${vsn}-otp-${{ matrix.otp }}.zip
+          shasum -a 1 elixir-${vsn}-otp-${{ matrix.otp }}.zip > elixir-${vsn}-otp-${{ matrix.otp }}.zip.sha1sum
+          shasum -a 256 elixir-${vsn}-otp-${{ matrix.otp }}.zip > elixir-${vsn}-otp-${{ matrix.otp }}.zip.sha256sum
           echo "$PWD/bin" >> $GITHUB_PATH
       - name: Upload Pre-built
         env:
@@ -57,8 +59,8 @@ jobs:
         run: |
           vsn=$(cat VERSION)
           gh release upload --clobber "v${vsn}" \
-            elixir-${vsn}-otp-${{ matrix.otp_release }}.zip \
-            elixir-${vsn}-otp-${{ matrix.otp_release }}.zip.sha{1,256}
+            elixir-${vsn}-otp-${{ matrix.otp }}.zip \
+            elixir-${vsn}-otp-${{ matrix.otp }}.zip.sha{1,256}sum
       - uses: actions/checkout@v2
         with:
           repository: elixir-lang/ex_doc
@@ -76,8 +78,9 @@ jobs:
         run: |
           vsn=$(cat VERSION)
           make Docs.zip
-          shasum -a 1 "Docs-v${vsn}.zip" > Docs-v${vsn}.zip.sha1
-          shasum -a 256 "Docs-v${vsn}.zip" > Docs-v${vsn}.zip.sha256
+          mv "Docs-v${vsn}.zip" Docs.zip
+          shasum -a 1 "Docs.zip" > Docs.zip.sha1sum
+          shasum -a 256 "Docs.zip" > Docs.zip.sha256sum
       - name: Upload Docs
         if: ${{ matrix.build_docs }}
         env:
@@ -85,5 +88,5 @@ jobs:
         run: |
           vsn=$(cat VERSION)
           gh release upload --clobber "v${vsn}" \
-            Docs-v${vsn}.zip \
-            Docs-v${vsn}.zip.sha{1,256}
+            Docs.zip \
+            Docs.zip.sha{1,256}sum
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,7 +10,21 @@ env:
   LANG: C.UTF-8
 
 jobs:
+  create_draft_release:
+    runs-on: ubuntu-18.04
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - name: Create draft release
+        run: |
+          gh release create \
+            --repo ${{ github.repository }} \
+            --title ${{ github.ref_name }} \
+            --notes '' \
+            --draft \
+            ${{ github.ref_name }}
   build_zips:
+    needs: create_draft_release
     strategy:
       fail-fast: true
       matrix:
@@ -64,7 +78,7 @@ jobs:
           name: Docs.zip
           path: Docs.zip
         if: ${{ matrix.build_docs }}
-  create_draft_release:
+  upload_assets:
     needs: build_zips
     runs-on: ubuntu-18.04
     steps:
@@ -96,16 +110,12 @@ jobs:
           do
             shasum -a ${algo} elixir-${vsn}-otp-22.zip elixir-${vsn}-otp-23.zip elixir-${vsn}-otp-24.zip Docs.zip > "sha${algo}sums.txt"
           done
-      - name: Create a release and upload assets
+      - name: Upload assets
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           vsn=$(cat VERSION)
-          gh release create \
-            --title ${{ github.ref_name }} \
-            --notes '' \
-            --draft \
-            v${vsn} \
+          gh release upload --clobber v${vsn} \
             elixir-${vsn}-otp-22.zip \
             elixir-${vsn}-otp-23.zip \
             elixir-${vsn}-otp-24.zip \
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -95,6 +95,8 @@ jobs:
           for algo in 1 256
           do
             shasum -a ${algo} elixir-${vsn}-otp-22.zip elixir-${vsn}-otp-23.zip elixir-${vsn}-otp-24.zip Docs.zip > "sha${algo}sums.txt"
+            cat "sha${algo}sums.txt" | sed 's/^/    /' | sed "1s/^/### SHA${algo}||/" | tr '|' '\n' >> notes.md
+            printf '\n\n' >> notes.md
           done
       - name: Create a release and upload assets
         env:
@@ -103,7 +105,7 @@ jobs:
           vsn=$(cat VERSION)
           gh release create \
             --title ${{ github.ref_name }} \
-            --notes '' \
+            --notes-file notes.md \
             --draft \
             v${vsn} \
             elixir-${vsn}-otp-22.zip \
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -64,6 +64,7 @@ jobs:
       - uses: actions/checkout@v2
         with:
           repository: elixir-lang/ex_doc
+          ref: v0.26.0
           path: ex_doc
         if: ${{ matrix.build_docs }}
       - name: Build ex_doc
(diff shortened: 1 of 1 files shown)


wingyplus:
A current example from this workflow https://github.com/wingyplus/elixir/releases/tag/v1.14.0-test.4.

.github/workflows/release.yml
```diff
@@ -0,0 +1,40 @@
...
+          cd $RUNNER_TEMP
+          wget -O otp.tar.gz https://repo.hex.pm/builds/otp/ubuntu-18.04/${{ matrix.otp_release }}.tar.gz
+          mkdir -p otp
+          tar zxf otp.tar.gz -C otp --strip-components=1
+          otp/Install -minimal $(pwd)/otp
+          echo "$(pwd)/otp/bin" >> $GITHUB_PATH
+      - name: Build Elixir Release
+        run: |
+          make Precompiled.zip
+          mv Precompiled-v1.14.0-dev.zip elixir-${{ github.ref_name }}-${{ matrix.otp_release }}.zip
```
wingyplus, line 34: Still need to keep to `v1.14.0-dev` for testing purpose. I think we can renaming it to `github.ref_name` later.

.github/workflows/release.yml
```diff
@@ -0,0 +1,40 @@
...
+          ls -la
```
wingyplus, line 35: Remove this

.github/workflows/release.yml
wojtekmach: you can also read the `VERSION` file:

```bash
$ vsn=${cat VERSION}
$ mv Precompiled-v${vsn}.zip elixir-${vsn}-${{ matrix.otp_release }}.zip
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,40 @@
...
+        with:
+          fetch-depth: 50
+      - name: Install Erlang/OTP
+        run: |
+          cd $RUNNER_TEMP
+          wget -O otp.tar.gz https://repo.hex.pm/builds/otp/ubuntu-18.04/${{ matrix.otp_release }}.tar.gz
+          mkdir -p otp
+          tar zxf otp.tar.gz -C otp --strip-components=1
+          otp/Install -minimal $(pwd)/otp
+          echo "$(pwd)/otp/bin" >> $GITHUB_PATH
```
wojtekmach, line 30: I think it is totally fine to manually download and install OTP however another option worth mentioning is to simply:

```yaml
      - uses: erlef/setup-beam@v1
        with:
          otp-version: ${{ matrix.otp_release }}
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,40 @@
...
+      - name: Build Elixir Release
+        run: |
+          make Precompiled.zip
+          mv Precompiled-v1.14.0-dev.zip elixir-${{ github.ref_name }}-${{ matrix.otp_release }}.zip
+          ls -la
+      - name: Upload Pre-built
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh release upload ${{ github.ref_name }} elixir-${{ github.ref_name }}-${{ matrix.otp_release }}.zip
```
wojtekmach, line 40: double-checking, do we need to first create a release off of a tag before we can upload assets to it?

.github/workflows/release.yml
wingyplus: It could be simpler if we can use `setup-beam`.  :)

.github/workflows/release.yml
wingyplus: Yes. Currently, I assume that the maintainer should create a release first. After created, github will publish create tag event to the CI.

wingyplus: Another question, do we need to rely on `VERSION` file instead of ref_name?

wojtekmach: i believe either is fine

josevalim:
Hi @wingyplus, thank you so much, to answer your questions:

1. We need the docs indeed. But the docs can be computed only once (in the last OTP version).
2. We need both sha1 and sha256

.github/workflows/release.yml
wojtekmach: I'll let maintainers decide whether they would rather push a tag or a release. :)

Regardless how we get there, I think it would be preferred if the release starts as a _draft_ (e.g.: `$ gh release create --draft`) and it would only be _published_ (either manually or automatically) after all per-OTP builds have been uploaded. This way we'd only fire off release notifications when everything is ready.

wingyplus: Thanks! I would try testing with this approach first.

.github/workflows/release.yml
```diff
@@ -0,0 +1,36 @@
...
+      matrix:
+        otp_release: ["24", "23", "22"]
+    runs-on: ubuntu-18.04
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 50
+      - uses: erlef/setup-beam@v1
+        with:
+          otp-version: ${{ matrix.otp_release }}
```
wojtekmach, line 25: Remainder that setup-beam does version resolution, eg `24` would today be resolved as `24.2`. Is this what we want, to always pick latest? I think it is, but perhaps we should leave a comment in case it is not immediately obvious?

wingyplus:
> Hi @wingyplus, thank you so much, to answer your questions:
> 
>     1. We need the docs indeed. But the docs can be computed only once (in the last OTP version).
> 
>     2. We need both sha1 and sha256

Thank you to answer my question. I've another question for `2.`, what's file pattern should we do for both `sha1` and `sha256`? Currently, I thought about `elixir-${vsn}-otp-${{ matrix.otp_release }}.zip.sha{1,256}`. But other suggestions are welcome. :)

.github/workflows/release.yml
josevalim: > Regardless how we get there, I think it would be preferred if the release starts as a draft (e.g.: $ gh release create --draft) and it would only be published (either manually or automatically) after all per-OTP builds have been uploaded.

Exactly! I think some manual steps may still be required, unfortunately, but this will reduce them greatly! ❤️

wojtekmach:
I don't think we include sha _files_ instead we _append_ the following to the release notes, same as today:

```markdown
## Checksums

Precompiled.zip SHA1: 40762ffbef86cbbfd0c79a94e057fb987dc882ff
Precompiled.zip SHA512: bb02ead0d4ccf499ff0473fbbb17fd12de4ba476b463c0452138bc1ef9004547166e75ffe7c6f96b9497adc9fecbda4ad57bfe0f17f1ba95d8339a0e98c29b03
Docs.zip SHA1: 066586bc79319f6248b26143124b76520e995c19
Docs.zip SHA512: 7c33c6208250a88646ffc2a2f74a944e7f1a9c69473791e9faaf054bd476876a1dd80d8c792a5c47b5bf350a77c0ba4ef191f3b4509b9e962d2514ef88bff9a3
```

wingyplus:
> I don't think we include sha _files_ instead we _append_ the following to the release notes, same as today:
> 
> ```gfm
> ## Checksums
> 
> Precompiled.zip SHA1: 40762ffbef86cbbfd0c79a94e057fb987dc882ff
> Precompiled.zip SHA512: bb02ead0d4ccf499ff0473fbbb17fd12de4ba476b463c0452138bc1ef9004547166e75ffe7c6f96b9497adc9fecbda4ad57bfe0f17f1ba95d8339a0e98c29b03
> Docs.zip SHA1: 066586bc79319f6248b26143124b76520e995c19
> Docs.zip SHA512: 7c33c6208250a88646ffc2a2f74a944e7f1a9c69473791e9faaf054bd476876a1dd80d8c792a5c47b5bf350a77c0ba4ef191f3b4509b9e962d2514ef88bff9a3
> ```

The hard part is we now have 3 versions of `Precompiled.zip` due to the OTP version matrix which's produce a different hash of sha{1,256}. If we need to do it in that way, I think we need to find a way to know the OTP version in makefile and build the Elixir compiler from makefile itself instead of changing the name of `Precompiled.zip` in CI. Thought?

wingyplus:
This is just a prototype for reference https://github.com/wingyplus/elixir/releases/tag/v1.14.0-dev. :)

josevalim:
Yes, given there are multiple releases, including the sha files seem fine by. Do we have other options?

wojtekmach:
Regarding continuous builds of the main branch, all we need to have is a release we can attach the builds to. By convention, main branch would have a version with `-dev` suffix, e.g. `v1.14.0-dev`, so my proposal would be to have a GitHub job that would be constantly creating and uploading (and perhaps even constantly updating release notes with updated checksums) builds to a release associated with that tag. The logic is actually very similar between uploading to a stable release and to that -dev release, so I think we could have a single github workflow with:

```yaml
on:
  push:
    tags:
      - v*
    branches:
      - main
    # # OR, for nightly:
    # schedule:
    #   - cron: "0 0 * * *"
```

and inside the workflow we figure out what kind of release we're dealing with. If that sounds like a good idea, it could be a separate PR. Or maybe we actually start with _that_ given we'd be testing it as soon as possible as opposed to when there is a new tag. :)

wingyplus:
> Regarding continuous builds of the main branch, all we need to have is a release we can attach the builds to. By convention, main branch would have a version with `-dev` suffix, e.g. `v1.14.0-dev`, so my proposal would be to have a GitHub job that would be constantly creating and uploading (and perhaps even constantly updating release notes with updated checksums) builds to a release associated with that tag. The logic is actually very similar between uploading to a stable release and to that -dev release, so I think we could have a single github workflow with:
> 
> ```yaml
> on:
>   push:
>     tags:
>       - v*
>     branches:
>       - main
>     # # OR, for nightly:
>     # schedule:
>     #   - cron: "0 0 * * *"
> ```
> 
> and inside the workflow we figure out what kind of release we're dealing with. If that sounds like a good idea, it could be a separate PR. Or maybe we actually start with _that_ given we'd be testing it as soon as possible as opposed to when there is a new tag. :)

It’s sound interesting if we have a nightly build. :D

wojtekmach:
> If we need to do it in that way, I think we need to find a way to know the OTP version in makefile and build the Elixir compiler from makefile itself instead of changing the name of Precompiled.zip in CI

Sorry I don’t see any problem with this. We know the OTP version, it’s the build matrix param. Each job in matrix would build Elixir from scratch against the given OTP version.

if we go with appending checksums to release notes, the only slight complication is to avoid race conditions when appending since matrix jobs run in parallel. We could have a separate _dependant_ job in the same workflow which is the one that pushes the release update.

.github/workflows/release.yml
wingyplus: > I'll let maintainers decide whether they would rather push a tag or a release. :)
> 
> Regardless how we get there, I think it would be preferred if the release starts as a _draft_ (e.g.: `$ gh release create --draft`) and it would only be _published_ (either manually or automatically) after all per-OTP builds have been uploaded. This way we'd only fire off release notifications when everything is ready.

As I do some research, the draft release will not create a tag, so it will not trigger a tag event and per [release event documentation](https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#release), it doesn't trigger if it's draft release. :(

.github/workflows/release.yml
```diff
@@ -0,0 +1,69 @@
...
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 50
+      - uses: erlef/setup-beam@v1
+        with:
+          otp-version: 24
+      - name: Build Elixir
+        run: |
+          make compile
```
wojtekmach, line 56: it's a bit wasteful to compile Elixir _again_ just to build docs. What do you think about re-using the builds from the previous job? The matrix step could look like this:

```yaml
      matrix:
        include:
          - otp: 22
          - otp: 23
          - otp: 24
            build_docs: build_docs

      # ...

      - run: |
          # ...
        if: ${{ matrix.build_docs }}

```

wingyplus: This look better than my current setup. I’ll try it tomorrow since this is near midnight in my timezone. 😴

.github/workflows/release.yml
wojtekmach: Good find. So yeah, I believe our trigger needs to be a tag push event from maintainers executing `$ git push origin <tag>` and similar. And then they'd either manually create a release draft or more likely we'd do it for them in the job. (so that by the time the job executes, there would be a release to attach files to.)

wingyplus: Thanks for your idea! I try by create a job that create a draft  release first. After draft release is created, it'll upload the asset to the draft release. The draft release will looks like:

Screen Shot 2564-12-28 at 18 02 25

Here a link to draft release https://github.com/wingyplus/elixir/releases/tag/untagged-358a0dfc87f7628dd7c8. I'm not sure anyone can see the draft or not. So I attach screen for demo above ☝️ 

Here is the pipeline https://github.com/wingyplus/elixir/actions/runs/1630211055: 

Screen Shot 2564-12-28 at 18 04 52

.github/workflows/release.yml
```diff
@@ -0,0 +1,68 @@
...
+    runs-on: ubuntu-18.04
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 50
+      - name: Create draft release
+        run: |
+          gh release create --title ${{ github.ref_name }} --notes '' --draft ${{ github.ref_name }}
```
wingyplus, line 23: I think we may use repo name from `${{ github.repository }}` to push directly to the target repository instead of cloning a repo.

.github/workflows/release.yml
wojtekmach: I am not able to see the draft so thank you for posting the screenshot. Looks fantastic!

wingyplus:
Hi all, 

The workflow now works. It can build Elixir against 3 OTP releases and the doc file. I update the PR description for example including a link to my fork repo. I think the workflow can be improved but this is midnight in my timezone and feeling asleep. So I leave it to you all to reviewing it and I'll take a look and try improving it tomorrow. 🙏

.github/workflows/release.yml
wingyplus: Added. :)

.github/workflows/release.yml
```diff
@@ -0,0 +1,89 @@
...
+            elixir-${vsn}-otp-${{ matrix.otp_release }}.zip.sha{1,256}
+      - uses: actions/checkout@v2
+        with:
+          repository: elixir-lang/ex_doc
+          path: ex_doc
+        if: ${{ matrix.build_docs }}
+      - name: Build ex_doc
+        if: ${{ matrix.build_docs }}
+        run: |
+          mv ex_doc ../ex_doc
```
wingyplus, line 70: I found that checkout action doesn't allow to clone outside elixir path. So I workaround by moving the ex_doc to outside elixir manually.

wingyplus:
> > If we need to do it in that way, I think we need to find a way to know the OTP version in makefile and build the Elixir compiler from makefile itself instead of changing the name of Precompiled.zip in CI
> 
> Sorry I don’t see any problem with this. We know the OTP version, it’s the build matrix param. Each job in matrix would build Elixir from scratch against the given OTP version.
> 
> if we go with appending checksums to release notes, the only slight complication is to avoid race conditions when appending since matrix jobs run in parallel. We could have a separate _dependant_ job in the same workflow which is the one that pushes the release update.

Oh, it is possible. We may use caching or artifacts to share zip files between jobs. After that, we create a job that uploads zip files and generates the checksum. So we may need to decide which way is better between posting a checksum to the release note or creating a new file. :)

wingyplus:
Improvement for today:

* Use upload/download artifact to share zip files between jobs
* Create a new job to generate checksum and upload those zip files to the releases.
  * checksum file will generate to 2 files, `sha1.txt` and `sha256.txt`, each file containing all checksum of zip files.

.github/workflows/release.yml
```diff
@@ -0,0 +1,123 @@
...
+      - name: Upload assets
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          vsn=$(cat VERSION)
+          gh release upload --clobber v${vsn} \
+            elixir-${vsn}-otp-22.zip \
+            elixir-${vsn}-otp-23.zip \
+            elixir-${vsn}-otp-24.zip \
+            Docs-${vsn}.zip \
```
wingyplus, line 121: We may rename a `Docs-v${vsn}.zip` to `Docs.zip` to be the same name on the previous release.

.github/workflows/release.yml
```diff
@@ -0,0 +1,123 @@
...
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - name: Create draft release
+        run: |
+          gh release create \
+            --repo ${{ github.repository }} \
+            --title ${{ github.ref_name }} \
+            --notes '' \
+            --draft \
+            ${{ github.ref_name }}
```
wingyplus, line 25: If we want to do a nightly build, we should change it to create a release (or pre-release) not a draft?

.github/workflows/release.yml
wojtekmach: Agreed

.github/workflows/release.yml
```diff
@@ -0,0 +1,123 @@
...
+          mv elixir-otp-22.zip elixir-${vsn}-otp-22.zip
+          mv elixir-otp-23.zip elixir-${vsn}-otp-23.zip
+          mv elixir-otp-24.zip elixir-${vsn}-otp-24.zip
+          mv Docs.zip Docs-${vsn}.zip
+      - name: Generate checksum
+        run: |
+          vsn=$(cat VERSION)
+          for algo in 1 256
+          do
+            shasum -a ${algo} elixir-${vsn}-otp-22.zip elixir-${vsn}-otp-23.zip elixir-${vsn}-otp-24.zip Docs-${vsn}.zip > "sha${algo}.txt"
```
wingyplus, line 110: If we do this in the workflow, the checksum in Makefile is not necessary anymore?

.github/workflows/release.yml
wojtekmach: I think it is fine to _start_ with a draft, to have somewhere to upload files to, but we need to end up with a published release. My thinking was publishing a release fires off GitHub notifications (emails etc) so it would be ideal if we only do that after we have all the files.

for nightly build, we need to create the release only _once_ (per minor version) given we’d be mostly re-uploading new files to it every day.

wojtekmach: And yes, definitely _pre-release_ for nightly. It would be nice to do pre-release for RC tags too.

.github/workflows/release.yml
wingyplus: Changed. :)

wojtekmach (approved):

.github/workflows/release.yml
wingyplus: I see.

wingyplus: > I think it is fine to start with a draft, to have somewhere to upload files to, but we need to end up with a published release. My thinking was publishing a release fires off GitHub notifications (emails etc) so it would be ideal if we only do that after we have all the files.

But before publish a release. We need to put a changelog to the release which's do it manually at the moment right? It would be nice if we can publish a draft with posting a changelog. :D

wojtekmach: Yeah, I think doing the release notes and actually publishing would be manual steps for maintainers. Release notes are currently a cut-down version of the changelog, which I think is a good thing (especially for _patch_ releases), so I would not try to automate that part after all.

.github/workflows/release.yml
wojtekmach: I think we should pin the specific version after all, that's what Bob did for Elixir builds. cc @ericmj

wingyplus: Can we use the same version as described in ci.yml:

- 24.0
- 23.3
- 22.3

wojtekmach: Bob currently has 22.3, 23.3, 24.1, maybe let's use those?

btw, setup-beam allows us to do:

```yaml
otp-version: ...
version-type: strict
```

to pin, so let's do that too!

wingyplus: Already push a fix. The CI shows that it uses the strict version https://github.com/wingyplus/elixir/actions/runs/1633938089.

.github/workflows/release.yml
```diff
@@ -0,0 +1,124 @@
...
+  ERLC_OPTS: "warnings_as_errors"
+  LANG: C.UTF-8
+
+jobs:
+  create_draft_release:
+    runs-on: ubuntu-18.04
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - name: Create draft release
```
michallepicki, line 18: I don't know if amount of jobs makes a difference, but since the artifacts are uploaded at the end in a separate `upload_assets` job, the `build_zips` currently doesn't need to depend on `create_draft_release`. Would it make sense to remove `create_draft_release`, and move this step to the existing `upload_assets`? (And maybe rename accordingly).

It would also allow to automatically include the SHA sums in release notes, not only as attached files. (I don't see an easy way to append notes in [gh cli reference](https://cli.github.com/manual/gh_release)

michallepicki (commented):
Will `Precompiled.zip` no longer be included in releases, or should that be automated as well?

wingyplus:
> Will `Precompiled.zip` no longer be included in releases, or should that be automated as well?

The `Precompiled.zip` renamed to `elixir-{version}-otp-{otp}.zip` in the release. :)

.github/workflows/release.yml
wingyplus: > I don't know if amount of jobs makes a difference, but since the artifacts are uploaded at the end in a separate `upload_assets` job, the `build_zips` currently doesn't need to depend on `create_draft_release`. Would it make sense to remove `create_draft_release`, and move this step to the existing `upload_assets`? (And maybe rename accordingly).

Great suggestion! I'll move it to the upload assets step.

> It would also allow to automatically include the SHA sums in release notes, not only as attached files. (I don't see an easy way to append notes in [gh cli reference](https://cli.github.com/manual/gh_release)

It may need to construct a text and put it when creating a release, I thought.

wojtekmach:
Maybe we should leave Precompiled.zip (eg a copy of the build with oldest OTP version) for backwards compatibility with existing tooling? We can always dobór _later_ if it breaks anyone though.

josevalim:
> Maybe we should leave Precompiled.zip (eg a copy of the build with oldest OTP version) for backwards compatibility with existing tooling? We can always do that later if it breaks anyone though.

I wouldn't worry about it. The only tooling I know uses the `Precompiled.zip` is the elixir.csv file, which has a full address to the zip file. Plus, depending on the more specific version will lead to better optimized code, so we want to push people towards this direction anyway.

wingyplus:
Hi all, 

I've changed the workflow to:

Screen Shot 2564-12-29 at 20 07 17

The draft release will be created after precompiled was built. About checksum in the release note, I'm not sure which one is better between putting it to the release note or uploading it to the assets (or both). If we want to go putting to the release note way, I think we can put the result from `shasum` to under the code block directly, no need to strip `-` like Makefile does.

.github/workflows/release.yml
```diff
@@ -0,0 +1,114 @@
...
+  ERLC_OPTS: "warnings_as_errors"
+  LANG: C.UTF-8
+
+jobs:
+  build_zips:
+    strategy:
+      fail-fast: true
+      matrix:
+        include:
+          - otp: 22
```
wingyplus, line 18: I introduce this field to uses for naming zip file.

.github/workflows/release.yml
```diff
@@ -0,0 +1,114 @@
...
+          fetch-depth: 50
+      - uses: actions/download-artifact@v2
+        with:
+          name: elixir-otp-22.zip
+      - uses: actions/download-artifact@v2
+        with:
+          name: elixir-otp-23.zip
+      - uses: actions/download-artifact@v2
+        with:
+          name: elixir-otp-24.zip
```
josevalim, line 82: Can we download those from the matrix? This way when adding/removing OTP versions we only need to change one place in the file. :)

.github/workflows/release.yml
```diff
@@ -0,0 +1,114 @@
...
+      - uses: actions/download-artifact@v2
+        with:
+          name: Docs.zip
+      - name: Rename files
+        run: |
+          vsn=$(cat VERSION)
+          mv elixir-otp-22.zip elixir-${vsn}-otp-22.zip
+          mv elixir-otp-23.zip elixir-${vsn}-otp-23.zip
+          mv elixir-otp-24.zip elixir-${vsn}-otp-24.zip
```
josevalim, line 91:
```suggestion
```

I don't think we need this step. The Elixir version will already be in the Release tag. That's enough, no?

josevalim:
@wingyplus oh, now that we build the checksums only at the end, can we keep them as a single file? Or perhaps, even send them as part of the notes? Maybe we can do something like this:

```bash
$ cat shasums.txt | sed 's/^/    /' | sed "1s/^/### Checksums||/" | tr '|' '\n'
```

Here is an example of output:

```bash
$ shasum README.md CHANGELOG.md | sed 's/^/    /' | sed "1s/^/### Checksums||/" | tr '|' '\n'
### Checksums

    67b8c712cddd5129c4b24d07980b6b3c8ef37824  README.md
    c4304cb2d71313fb02835f24a1e145457608d7cf  CHANGELOG.md
```

.github/workflows/release.yml
josevalim: To be clear, what I mean is that it would be nice if the second part of the snippet downloaded the files and generated the checksums based on the OTP versions of the matrix. :)

If it not possible, consider doing `shasum -a 256 *.zip`, so there is less repetition. Potentially downloading them to a common subdirectory to avoid unwarranted `.zip` files from showing up?

wingyplus:
> 

Combine 2 shasum files into the single one is possible. Try with:

```shell
$ cat shasums.txt
c57e55824f2d17735b8d48b64b3ab5d284fad809  elixir-1.14.0-dev-otp-22.zip
d76f809fa76056c229a9cb3da709c5271458b228  elixir-1.14.0-dev-otp-23.zip
41e8584719252b9756b20c75c7e6fc9a2c577159  elixir-1.14.0-dev-otp-24.zip
4f22a2a27b3fd02ee11247dcb9fb45bc848c9095  Docs.zip
ac48a453005cc19c00d6deef148be1fa340607b6f0238b4fe63cbe3be764924b  elixir-1.14.0-dev-otp-22.zip
e468cbc3ba7e146f3b3533bff34bb86fa09cd440fcd9aa92d49577da6cbe647e  elixir-1.14.0-dev-otp-23.zip
e172675ff99aa8d98d4a010b1ab44bf4a7a71fd7075a737bbf9622c462aea8c3  elixir-1.14.0-dev-otp-24.zip
5a9a9a8f3f498fbbcb275c12679e28b535e1eb6557229738eed9f25cd2c96cca  Docs.zip

$ shasum -c ./shasums.txt
elixir-1.14.0-dev-otp-22.zip: OK
elixir-1.14.0-dev-otp-23.zip: OK
elixir-1.14.0-dev-otp-24.zip: OK
Docs.zip: OK
elixir-1.14.0-dev-otp-22.zip: OK
elixir-1.14.0-dev-otp-23.zip: OK
elixir-1.14.0-dev-otp-24.zip: OK
Docs.zip: OK
```

Putting it to the release note is also possible. As I understand, we need to separate a section between sha1 and sha256, why don't want to separate a shasum files too?

.github/workflows/release.yml
```diff
@@ -0,0 +1,114 @@
...
+
+env:
+  ELIXIR_OPTS: "--warnings-as-errors"
+  ERLC_OPTS: "warnings_as_errors"
+  LANG: C.UTF-8
+
+jobs:
+  build_zips:
+    strategy:
+      fail-fast: true
```
ericmj, line 15: What's the purpose of `fail-fast`? Don't we want to build as many zips as possible?

.github/workflows/release.yml
```diff
@@ -0,0 +1,114 @@
+name: Release
+on:
+  push:
+    tags:
+      - v*
+
+env:
+  ELIXIR_OPTS: "--warnings-as-errors"
+  ERLC_OPTS: "warnings_as_errors"
```
ericmj, line 9: Why should warnings fail building the zips?

.github/workflows/release.yml
```diff
@@ -0,0 +1,114 @@
...
+          - otp: 23
+            otp_version: 23.3
+          - otp: 24
+            otp_version: 24.1
+            build_docs: build_docs
+    runs-on: ubuntu-18.04
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 50
```
ericmj, line 29: What's the reason for this?

.github/workflows/release.yml
```diff
@@ -0,0 +1,114 @@
...
+      - name: Build Elixir Release
+        run: |
+          vsn=$(cat VERSION)
+          make Precompiled.zip
+          mv Precompiled-v${vsn}.zip elixir-otp-${{ matrix.otp }}.zip
+          echo "$PWD/bin" >> $GITHUB_PATH
+      - uses: actions/upload-artifact@v2
+        with:
+          name: elixir-otp-${{ matrix.otp }}.zip
+          path: elixir-otp-${{ matrix.otp }}.zip
```
ericmj, line 43: We should upload a zip named `elixir.zip` build against the oldest OTP version.

.github/workflows/release.yml
```diff
@@ -0,0 +1,114 @@
...
+      - uses: actions/checkout@v2
+        with:
+          repository: elixir-lang/ex_doc
+          path: ex_doc
```
ericmj, line 47: Is it possible to do `../ex_doc` here so you don't have to move later?

ericmj, line 47: I think we should checkout latest stable ex_doc instead of using the development version.

ericmj (commented):
I left a few comments and questions. Thanks for working on this!

.github/workflows/release.yml
wingyplus: I have been try but checkout action doesn't allowed. :(

.github/workflows/release.yml
wingyplus: We can rename the `path` but not for the `name` since it's used in the download artifact below.

.github/workflows/release.yml
wingyplus: I just copied from the `ci.yml`. So no reason specific.

wingyplus:
@josevalim does it look ok?

8FCC5984-EF15-4759-977C-F4156D07196B

link https://github.com/wingyplus/elixir/releases/tag/v1.14.0-dev

.github/workflows/release.yml
wingyplus: I thought that we need to re run the workflow when it failure. I’m not sure that the upload artifact can keep the success artifact. If it can keep, the fail fast may not need.

.github/workflows/release.yml
wingyplus: Noted. I will try checkout the stable version instead. :)

.github/workflows/release.yml
wingyplus: Oh sorry. I misread your comment. Why do we need to use elixir.zip for oldest build?

.github/workflows/release.yml
josevalim: It is probably best to not ship with a release with warnings? 🤔

.github/workflows/release.yml
josevalim: IIRC to reduce the amount of data downloaded.

.github/workflows/release.yml
josevalim: We discussed this and decided not to. Choosing your current OTP version should be preferred whenever possible.

.github/workflows/release.yml
wingyplus: My intention is to rely on VERSION file instead of github.ref_name. But i can use the ref_name instead for this case. What do you think?

josevalim:
@wingyplus it looks great! But I didn’t know about the `-c` flag and now I am thinking the file is better! :D sorry about that. I propose we revert, wdyt?

wojtekmach:
I initially thought about putting checksums into release notes, same as today, however the more I think about it, a checksum file per build file is probably a better option. For tooling that downloads and verifies the build it will be more convenient than grepping. Grabbing release notes via api would be even harder.

.github/workflows/release.yml
wingyplus: Not sure if upload artifact can upload to the same destination. Let me play around with this a little bit. :)

.github/workflows/release.yml
ericmj: Good point. Let's keep it.

.github/workflows/release.yml
ericmj: That should already fail the main CI workflow.

josevalim: Right, but those workflows are mostly unrelated right? And a failed build should then result in no artifacts as well, right?

.github/workflows/release.yml
josevalim: Right now it doesn’t matter because they will always be the same but I think in the future the ref name is better when we do nightly builds because we may have a short period of time where the version in the main branch is v1.14.0 while there is also a v1.14.0 tag.

josevalim: But the main point is that we don’t need those in the archive names I think. :)

wingyplus:
@josevalim @wojtekmach don't mind. :D I think we should revert to the first commit (https://github.com/elixir-lang/elixir/pull/11522/commits/56b70c83d18a5c82380a8e3d9b361c794b8c8af4). So the workflow could be simpler and the comment https://github.com/elixir-lang/elixir/pull/11522#discussion_r776331494 and https://github.com/elixir-lang/elixir/pull/11522#discussion_r776331149 will be not an issue anymore. 

What do you think?

josevalim:
Yeah, reverting to that commit sounds good to me.

wojtekmach:
Yes, I think so. Sorry for pushing towards wrong path.

wingyplus:
@josevalim @wojtekmach @ericmj The workflow now reverts to the first commit, pin the OTP version, and use the stable ex_doc version. Please help review again. 🙏

josevalim (approved):

ericmj (approved):

josevalim:

josevalim:
Thank you so much @wingyplus!

Btw, are you interested in sending a PR to enable the nightly build of the main branch too? I guess then we would have a branch in permanent draft state? Is that what other repos do?

wojtekmach:
> I guess then we would have a **branch** in permanent draft state

(emphasis mine)

to be more specific, we'd have a _tag_ and an associated _published_ release, a pre-release to be more specific. It cannot be a draft since other people won't see it.

josevalim:
I don't think we need a tag because release notes do not need to be associated to a tag. But you are right it cannot be draft.

wingyplus:
@josevalim yeah, i’m interesting. But maybe after new year holiday. 😊

josevalim:
Definitely, take your time and happy new year! 🍾

wingyplus:
Happy new year too. 🎉🎉

wingyplus:
> I don't think we need a tag because release notes do not need to be associated to a tag. But you are right it cannot be draft.

@josevalim It may need a workflow that re-create a pre-release every day to make sure the the nightly release tag the current commit on the main branch and publish the assets.

PR #11526, opened 2021-12-27
Fix typespec for Base.decode16!/2

prihandi:
Recently I got warning from Dialyzer when calling `Base.decode16!(some_text, case: :mixed)`. I found that the typespec is using `encode_case` type instead of `decode_case` type.

### DIFF
--- a/lib/elixir/lib/base.ex
+++ b/lib/elixir/lib/base.ex
@@ -340,7 +340,7 @@ defmodule Base do
       "foobar"
 
   """
-  @spec decode16!(binary, case: encode_case) :: binary
+  @spec decode16!(binary, case: decode_case) :: binary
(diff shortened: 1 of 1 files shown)


whatyouhide (approved):
Nice catch @prihandi! 💟

josevalim:

weisslj:
@josevalim this bugfix fd135405a from @prihandi is for a regression that was introduced from v1.13.0 to v1.13.1, in cbba61a38.

I saw that it was only fixed in the main branch, could it be backported to v1.13.4? That would be great!

Sorry for not noticing it sooner, before v1.13.3 a few days ago :disappointed:. This prevents upgrading to v1.13.3 for us, as dialyzer fais for our project ...

PR #11528, opened 2021-12-29
Define exception field as true instead of term when expanding struct type

michallepicki:
replaces #11527

### DIFF
--- a/lib/elixir/lib/kernel/typespec.ex
+++ b/lib/elixir/lib/kernel/typespec.ex
@@ -568,7 +568,7 @@ defmodule Kernel.Typespec do
     types =
       :lists.map(
         fn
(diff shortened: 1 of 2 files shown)


lib/elixir/lib/kernel/typespec.ex
```diff
@@ -567,7 +567,10 @@ defmodule Kernel.Typespec do
 
     types =
       :lists.map(
-        fn {field, _} -> {field, Keyword.get(fields, field, quote(do: term()))} end,
+        fn
+          {:__exception__ = field, true} -> {field, Keyword.get(fields, field, quote(do: true))}
```
michallepicki, line 571:
```suggestion
          {:__exception__ = field, true} -> {field, Keyword.get(fields, field, true)}
```

josevalim:

PR #11529, opened 2021-12-30
Fix hidden escape chars in File.stream!/3 docs

aptinio:
image

### DIFF
--- a/lib/elixir/lib/file.ex
+++ b/lib/elixir/lib/file.ex
@@ -1604,7 +1604,7 @@ defmodule File do
     :file.close(io_device)
   end
 
-  @doc """
+  @doc ~S"""
   Returns a `File.Stream` for the given `path` with the given `modes`.
 
   The stream implements both `Enumerable` and `Collectable` protocols,


lib/elixir/lib/file.ex
```diff
@@ -1612,8 +1612,8 @@ defmodule File do
 
   The `line_or_bytes` argument configures how the file is read when
   streaming, by `:line` (default) or by a given number of bytes. When
-  using the `:line` option, CRLF line breaks (`"\r\n"`) are normalized
-  to LF (`"\n"`).
+  using the `:line` option, CRLF line breaks (`"\\r\\n"`) are normalized
+  to LF (`"\\n"`).
```
whatyouhide, line 1616: Instead of this, to avoid others falling into this "gotcha" in the future, you can change this particular docstring to use the `~S` sigil. You can do that by changing line 1607 (a few lines above) from `@doc """` to `@doc ~S"""`. 🙃

aptinio: Great idea! Thanks @whatyouhide!

josevalim:

PR #11531, opened 2021-12-31
Improve Map and Keyword deprecations

eksperimental:
- Adds guards for fuctions
- Optmize Map.filter/2, Map.reject/2

### DIFF
--- a/lib/elixir/lib/map.ex
+++ b/lib/elixir/lib/map.ex
@@ -1038,19 +1038,31 @@ defmodule Map do
 
   @doc false
   @deprecated "Use Enum.filter/2 instead"
-  def filter(map, fun) when is_map(map) do
-    Enum.filter(map, fun) |> :maps.from_list()
+  def filter(map, fun) when is_map(map) and is_function(fun, 1) do
+    Enum.reduce(map, %{}, fn {key, value} = element, acc ->
+      if fun.(element) do
+        Map.put(acc, key, value)
+      else
+        acc
+      end
+    end)
   end
 
   @doc false
   @deprecated "Use Enum.reject/2 instead"
-  def reject(map, fun) when is_map(map) do
-    Enum.reject(map, fun) |> :maps.from_list()
+  def reject(map, fun) when is_map(map) and is_function(fun, 1) do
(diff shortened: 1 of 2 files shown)


lib/elixir/lib/map.ex
```diff
@@ -1038,19 +1038,31 @@ defmodule Map do
...
-  def filter(map, fun) when is_map(map) do
-    Enum.filter(map, fun) |> :maps.from_list()
+  def filter(map, fun) when is_map(map) and is_function(fun, 1) do
+    Enum.reduce(map, %{}, fn {key, value} = element, acc ->
+      if fun.(element) do
+        Map.put(acc, key, value)
+      else
+        acc
+      end
+    end)
```
michallepicki, line 1048: I am not sure it is worth to optimize deprecated functions? (Additionally I am not sure this will be even faster, optimizations should be backed by some benchmarks IMO)

josevalim:
The functions are deprecated. There is no need to optimize and the guards are optional too as they would crash anyway. Thanks!

lib/elixir/lib/map.ex
eksperimental: I agree with the optimization of deprecated functions, but in this particular case the optimization was for a replacement of the function that was being deprecated. The rationale was that we were iterating two times more than needed, one reversing the results from Enum.filter, and another one converting from a list to a map.
I ran a benchmark and you were right, it was less performant, so i tweaked it and I have achieved an optimization of 6-7% with this.

```elixir
  def filter(map, fun) when is_map(map) do
    Enum.reduce(map, [], fn element, acc ->
      if fun.(element) do
        [element | acc]
      else
        acc
      end
    end)
    |> :maps.from_list()
  end
```

So pretty much we save one `:lists.reverse/1`

josevalim: These versions existed undeprecated for two tiny patch versions only. We should keep them simple because nobody will actually be using them in practice.

eksperimental: 1.13.0 and 1.13.1 which is the current stable release.
Is this correct?

josevalim: Yes, they will be deprecated on 1.13.2.

PR #11532, opened 2022-01-01
Fix the nested `uniq:` acc name clash by preserving a scope

am-kantox:
Current implementation of `:elixir_erl_for.build_reduce/7` discards a scope, resulting in an accumulator name clash for nested `for/1`s when `into:` option is passed to more than one of them:

```elixir
for x <- [1, 1, 2], uniq: true do
  for y <- [3, 3], uniq: true do
    x * y
  end
end
#⇒ ** (MatchError) no match of right hand side value: {[3], %{3 => true}}
```

Expected behaviour:
```elixir
#⇒ [[3], [6]]
```

This PR provides a fix by preserving a scope in calls to `build_reduce/7`.


josevalim:

PR #11534, opened 2022-01-03
Add assignment as generator documentation at for comprehension

calvin-kargo:
based on https://elixirforum.com/t/for-comprehension-theres-no-assignment-only-filters/44930/10

Adding mention of variable assignment as filters, and provide alternative for variable assignment based on generators.

### DIFF
--- a/lib/elixir/lib/kernel/special_forms.ex
+++ b/lib/elixir/lib/kernel/special_forms.ex
@@ -1388,6 +1388,10 @@ defmodule Kernel.SpecialForms do
       iex> for n <- [1, 2, 3, 4, 5, 6], rem(n, 2) == 0, do: n
       [2, 4, 6]
 
+  Filters must evaluate to truthy values (everything but `nil`
+  and `false`). If a filter is falsy, then the current value is
+  discarded.
+
   Generators can also be used to filter as it removes any value
   that doesn't match the pattern on the left side of `<-`:
 
@@ -1408,19 +1412,20 @@ defmodule Kernel.SpecialForms do
   filters or inside the block, are not reflected outside of the
   comprehension.
 
-  Variable assignment based on `=` is treated as filters, and 
-  if the value of the expression is falsey (`nil` and `false`), it will be filtered out.
-  in case you don't want variable assignment to be treated as filters, 
-  you could move variable assignment into `do` block, or use a `<-` generator
-  and a single-element list literal on its right-hand side.
-      # A incorrect comprehension for getting language with grandparent.
-      # because `erlang` and `prolog` doesn't have grandparent, assigning `grandparent` is treated as false filter.
+  Variable assignments inside filters must still return a truthy value,
+  otherwise values are discarded. Let's see an example. Imagine you have
+  a keyword list where the key is a programming language and the value
+  is its direct parent. Then let's try to compute the grandparent of each
+  language. You could try this:
+
       iex> languages = [elixir: :erlang, erlang: :prolog, prolog: nil]
       iex> for {language, parent} <- languages, grandparent = languages[parent], do: {language, grandparent}
       [elixir: :prolog]
 
-      # A correct comprehension for getting language with grandparent.
-      # Here, assignment is moved into `do` block, and not treated as filter.
+  Given the grandparents of Erlang and Prolog were nil, those values were
+  filtered out. IF you don't want this behaviour, a simple option is to
+  move the filter inside the do-block:
+
       iex> languages = [elixir: :erlang, erlang: :prolog, prolog: nil]
       iex> for {language, parent} <- languages do
       ...>   grandparent = languages[parent]
@@ -1428,8 +1433,10 @@ defmodule Kernel.SpecialForms do
       ...> end
       [elixir: :prolog, erlang: nil, prolog: nil]
 
-      # Alternative correct comprehension for getting language with grandparent.
-      # Here, assignment changed as generator (<-) with one element list, useful in case you have other filters which depend on `grandparent` data.
+  However, such option is not always available, as you may have further
+  filters. An alternative is to convert the filter into a generator by
+  wrapping the right side of `=` in a list:
+
       iex> languages = [elixir: :erlang, erlang: :prolog, prolog: nil]
       iex> for {language, parent} <- languages, grandparent <- [languages[parent]], do: {language, grandparent}
       [elixir: :prolog, erlang: nil, prolog: nil]
--- a/lib/elixir/lib/kernel/special_forms.ex
+++ b/lib/elixir/lib/kernel/special_forms.ex
@@ -1423,8 +1423,8 @@ defmodule Kernel.SpecialForms do
       # A correct comprehension for getting language with grandparent.
       # Here, assignment is moved into `do` block, and not treated as filter.
       iex> for {language, parent} <- languages do
-            grandparent = languages[parent]
-            {language, grandparent}
+          grandparent = languages[parent]
+          {language, grandparent}
           end
       [elixir: :prolog, erlang: nil, prolog: nil]
 
--- a/lib/elixir/lib/kernel/special_forms.ex
+++ b/lib/elixir/lib/kernel/special_forms.ex
@@ -1413,15 +1413,15 @@ defmodule Kernel.SpecialForms do
   in case you don't want variable assignment to be treated as filters, 
   you could move variable assignment into `do` block, or use a `<-` generator
   and a single-element list literal on its right-hand side.
-      iex> languages = [elixir: :erlang, erlang: :prolog, prolog: nil]
-
       # A incorrect comprehension for getting language with grandparent.
       # because `erlang` and `prolog` doesn't have grandparent, assigning `grandparent` is treated as false filter.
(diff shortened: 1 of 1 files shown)


josevalim:
Thank you @calvin-kargo! It would be nice if we could come up with an actual example, instead of `a` and `b`. I tried but nothing came to mind. Maybe we can lift something from your actual use case?

In the scenarios in the PR, it felt I should just write `x[:b] == nil` instead. After all, why assign something if I am explicitly looking for it to be `nil`?

calvin-kargo:
>Maybe we can lift something from your actual use case?
In the use case when was realizing there's only filters, i am using assignment as intermediate value for something that is calculated. when that intermediate value return `nil`, even though it might be a valid value for next filter, the comprehension skip that value.

Paraphrasing the actual code, it looks something like below, the realization happen when:
1. some of the order hasn't had any primary_server_id or backup_server_id, returning nil
2. `matched_server?(nil, order)` return `false`
```
        for order <- orders, 
          server_id = order.primary_server_id || order.backup_server_id,
          current_server = Servers.get_server(server_id),
          !matched_server?(current_server, order),
          new_server = Servers.find_matching_server(order)
        do
          update(order, new_server.id)
        end
```

lib/elixir/lib/kernel/special_forms.ex
```diff
@@ -1408,6 +1408,19 @@ defmodule Kernel.SpecialForms do
   filters or inside the block, are not reflected outside of the
   comprehension.
 
+  Variable assignment based on `=` is treated as filters, and 
+  if value of the expression is falsey (`nil` and `false`), it would be filtered out.
```
whatyouhide, line 1412:
```suggestion
  if the value of the expression is falsey (`nil` and `false`), it will be filtered out.
```

josevalim:
Thanks! Yes, this is tough because the code has to become more complex but then it is harder to make the point. What if we go with a tree approach? We have a map of programming languages with their immediate parent:

```elixir
languages = [elixir: :erlang, erlang: :prolog, prolog: nil]
```

And then we want to return the language with their grandparents. This is wrong:

```elixir
iex(11)> for {language, parent} <- languages, grandparent = languages[parent], do: {language, grandparent}
[elixir: :prolog]
```

We can fix it with in two ways. Move the assignment to do the do-block:

```elixir
iex(11)> for {language, parent} <- languages do
...(11)>  grandparent = languages[parent]
...(11)>  {language, grandparent}
...(11)> end
[elixir: :prolog, erlang: nil, prolog: nil]
```

Or if not possible, convert the filter to a generator:

```elixir
iex(12)> for {language, parent} <- languages, grandparent <- [languages[parent]], do: {language, grandparent}
[elixir: :prolog, erlang: nil, prolog: nil]
```

WDYT?

lib/elixir/lib/kernel/special_forms.ex
```diff
@@ -1408,6 +1408,19 @@ defmodule Kernel.SpecialForms do
...
+  in case variable assignment don't want to be treated as filters, using assignment via generators would be preferred.
```
whatyouhide, line 1413:
```suggestion
  in case you don't want variable assignment to be treated as filters, use a `<-` generator
  and a single-element list literal on its right-hand side.
```

calvin-kargo:
>iex(12)> for {language, parent} <- languages, grandparent <- [languages[parent]], do: {language, grandparent}
[elixir: :prolog, erlang: nil, prolog: nil]

Looks good! though the case for having assignment in generator would look more pronounced if it is used in next filter, let's say we want to exclude all language which have grandparent `c` ?

That also bring up another topic, should there also included a general warning to do assignment if possible just in `for` comprehension body and not in filters/generator? 
```
>iex(12)> for {language, parent} <- languages, grandparent <- [languages[parent]], grandparent not in [:c], do: {language, grandparent}
[elixir: :prolog, erlang: nil, prolog: nil]
```

josevalim:
Yeah, I thought about the case of having another filter but that makes it more complex. So I would go with the example above and when we say "Or if not possible, because you have other filters, convert the filter to a generator:" or similar.

> That also bring up another topic, should there also included a general warning to do assignment if possible just in for comprehension body and not in filters/generator?

Hard to say if the assignment is not filtering unless we have a type system.

calvin-kargo:
Thanks for the feedback! PR updated

lib/elixir/lib/kernel/special_forms.ex
```diff
@@ -1408,6 +1408,32 @@ defmodule Kernel.SpecialForms do
   filters or inside the block, are not reflected outside of the
   comprehension.
 
+  Variable assignment based on `=` is treated as filters, and
```
palexanderm, line 1411:
```suggestion
  Variable assignment based on `=` is treated as a filter, and 
```

lib/elixir/lib/kernel/special_forms.ex
```diff
@@ -1408,6 +1408,32 @@ defmodule Kernel.SpecialForms do
   filters or inside the block, are not reflected outside of the
   comprehension.
 
+  Variable assignment based on `=` is treated as filters, and 
+  if the value of the expression is falsey (`nil` and `false`), it will be filtered out.
+  in case you don't want variable assignment to be treated as filters,
```
palexanderm, line 1413:
```suggestion
  In case you don't want variable assignments to be treated as filters, 
```

lib/elixir/lib/kernel/special_forms.ex
```diff
@@ -1408,6 +1408,32 @@ defmodule Kernel.SpecialForms do
   filters or inside the block, are not reflected outside of the
   comprehension.
 
+  Variable assignment based on `=` is treated as filters, and 
+  if the value of the expression is falsey (`nil` and `false`), it will be filtered out.
+  in case you don't want variable assignment to be treated as filters, 
+  you could move variable assignment into `do` block, or use a `<-` generator
```
palexanderm, line 1414:
```suggestion
  you could move the variable assignment into the `do` block, or use a `<-` generator
```

lib/elixir/lib/kernel/special_forms.ex
```diff
@@ -1408,6 +1408,32 @@ defmodule Kernel.SpecialForms do
...
+  and a single-element list literal on its right-hand side.
+      # A incorrect comprehension for getting language with grandparent.
```
palexanderm, line 1416:
```suggestion
      # An incorrect comprehension for getting language with grandparent.
```

lib/elixir/lib/kernel/special_forms.ex
```diff
@@ -1408,6 +1408,32 @@ defmodule Kernel.SpecialForms do
...
+      # because `erlang` and `prolog` doesn't have grandparent, assigning `grandparent` is treated as false filter.
```
palexanderm, line 1417:
```suggestion
      # Because `erlang` and `prolog` each  don't have a grandparent, assigning `grandparent` is treated as false filter.
```

lib/elixir/lib/kernel/special_forms.ex
```diff
@@ -1408,6 +1408,32 @@ defmodule Kernel.SpecialForms do
...
+      iex> languages = [elixir: :erlang, erlang: :prolog, prolog: nil]
+      iex> for {language, parent} <- languages, grandparent = languages[parent], do: {language, grandparent}
+      [elixir: :prolog]
+
+      # A correct comprehension for getting language with grandparent.
+      # Here, assignment is moved into `do` block, and not treated as filter.
```
palexanderm, line 1423:
```suggestion
      # Here, the assignment is moved into the `do` block, and not treated as a filter.
```

lib/elixir/lib/kernel/special_forms.ex
```diff
@@ -1408,6 +1408,32 @@ defmodule Kernel.SpecialForms do
...
+      iex> languages = [elixir: :erlang, erlang: :prolog, prolog: nil]
+      iex> for {language, parent} <- languages do
+      ...>   grandparent = languages[parent]
+      ...>   {language, grandparent}
+      ...> end
+      [elixir: :prolog, erlang: nil, prolog: nil]
+
+      # Alternative correct comprehension for getting language with grandparent.
+      # Here, assignment changed as generator (<-) with one element list, useful in case you have other filters which depend on `grandparent` data.
```
palexanderm, line 1432:
```suggestion
      # Here, the assignment is converted to a generator (<-) with a single-element list, which is useful in case you have other filters which depend on the `grandparent` data.
```

lib/elixir/lib/kernel/special_forms.ex
```diff
@@ -1408,6 +1408,32 @@ defmodule Kernel.SpecialForms do
   filters or inside the block, are not reflected outside of the
   comprehension.
 
+  Variable assignment based on `=` is treated as filters, and 
+  if the value of the expression is falsey (`nil` and `false`), it will be filtered out.
```
palexanderm, line 1412:
```suggestion
  if the value of the expression is falsey (i.e. either `nil` or `false`), it will be filtered out.
```

palexanderm (commented):
Just a few suggestions in regards to grammatical changes.

josevalim:

PR #11536, opened 2022-01-03
Add name lookup to pid/1 helper

ConnorRigby:
I frequently define my own helper for this. Thought it might be a useful add for others.


josevalim:

lib/iex/test/iex/helpers_test.exs
```diff
@@ -1347,6 +1347,7 @@ defmodule IEx.HelpersTest do
     test "builds a PID from string" do
       assert inspect(pid("0.32767.3276")) == "#PID<0.32767.3276>"
       assert inspect(pid("0.5.6")) == "#PID<0.5.6>"
+      assert inspect(pid(:init) == "#PID<0.0.0>")
```
sega-yarkin, line 1350: It looks like `inspect` should be around `pid/1` function, currently this test will never fail

josevalim: Good catch, please send a PR?

PR #11537, opened 2022-01-04
Use consistent argument name in DateTime doc

sabiwara:
Hi,

I noticed a couple of functions in the `DateTime` module had the default argument name `map` in the documentation, while the rest is consistently `datetime`.

`to_naive(map)` -> `to_naive(datetime)`

### DIFF
--- a/lib/elixir/lib/calendar/datetime.ex
+++ b/lib/elixir/lib/calendar/datetime.ex
@@ -856,6 +856,8 @@ defmodule DateTime do
 
   """
   @spec to_naive(Calendar.datetime()) :: NaiveDateTime.t()
+  def to_naive(datetime)
+
   def to_naive(%{
         calendar: calendar,
         year: year,
@@ -895,6 +897,8 @@ defmodule DateTime do
 
   """
   @spec to_date(Calendar.datetime()) :: Date.t()
+  def to_date(datetime)
+
   def to_date(%{
         year: year,
         month: month,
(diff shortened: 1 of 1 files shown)


whatyouhide (approved):
Great! Yes, `map` is the name generated by ex_doc when it sees a map literal. Another trick instead of adding a bodiless clause is to do

```elixir
def to_time(_datetime = %{...
```

so that ex_doc picks that name up instead. Saves a couple of lines 😉 I'm good either way.

sabiwara:
> Another trick instead of adding a bodiless clause is to do
> 
> ```elixir
> def to_time(_datetime = %{...
> ```
> 
> so that ex_doc picks that name up instead

TIL, thanks for the neat trick! :heart:
I've seen the bodiless clauses in the codebase including some recent examples like https://github.com/elixir-lang/elixir/commit/40b512031f9ea68b4fbcf57d323cbedf4568cdfe, so I figured this was the standard way to do it, but happy to go with the `_datetime` version too, it is more compact indeed.

whatyouhide:
@sabiwara yeah both approaches are fine. I always use bodiless clauses when I have more than one clause, but for single-clause functions, I usually go with the `_approach` above. In any case, the end result is the same 😉

wojtekmach:
Another idea is to infer argument name from typespec (if any) like this: we’d show it as `thing` for `Foo.thing()` or `Thing.t()`.

josevalim:
I would prefer to not mix typespecs for now. I think we don't use the `def foo(_foo = %{})` version in Elixir (nothing against it, the underscore version just came later), so the bodiless clause is the way to go IMO!

josevalim:

PR #11538, opened 2022-01-04
Update capture_log.ex

zolakeith:
add backticks

### DIFF
--- a/lib/ex_unit/lib/ex_unit/capture_log.ex
+++ b/lib/ex_unit/lib/ex_unit/capture_log.ex
@@ -57,7 +57,7 @@ defmodule ExUnit.CaptureLog do
 
   It is possible to configure the level to capture with `:level`,
   which will set the capturing level for the duration of the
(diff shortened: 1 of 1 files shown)


lib/ex_unit/lib/ex_unit/capture_log.ex
```diff
@@ -57,7 +57,7 @@ defmodule ExUnit.CaptureLog do
 
   It is possible to configure the level to capture with `:level`,
   which will set the capturing level for the duration of the
-  capture, for instance, if the log level is set to :error
+  capture, for instance, if the log level is set to `:error`
```
whatyouhide, line 60: Let's fix this too while we're at it:

```suggestion
  capture, for instance, if the log level is set to `:error`, then
```

whatyouhide (approved):

zolakeith:
Thanks @whatyouhide, I didn't spot that!

whatyouhide:
Thanks @zolakeith! ❤️

PR #11539, opened 2022-01-05
Fix a few issues in the mix test.coverage docs

doughsay:
Sorry for the incredibly trivial PR, but I was reading these docs and saw a few mistakes I thought could use fixing 😅


whatyouhide:
Not trivial at all, this is great! ❤️  Thanks @doughsay.

PR #11540, opened 2022-01-05
Add Node.spawn_link

lukaszsamson:
available since OTP 23
Related to https://github.com/elixir-lang/elixir/issues/9603

### DIFF
--- a/lib/elixir/lib/node.ex
+++ b/lib/elixir/lib/node.ex
@@ -261,8 +261,8 @@ defmodule Node do
   Spawns the given function on a node, monitors it and returns its PID
   and monitoring reference.
 
-  This functionality was added on Erlang/OTP 23. Using this functionality to communicate with
-  a node running on an earlier version will fail.
+  This functionality was added on Erlang/OTP 23. Using this function to
+  communicate with nodes running on earlier versions will fail.
 
   Inlined by the compiler.
   """
@@ -276,8 +276,8 @@ defmodule Node do
   Spawns the given module and function passing the given args on a node,
   monitors it and returns its PID and monitoring reference.
 
-  This functionality was added on Erlang/OTP 23. Using this functionality to communicate with
-  a node running on an earlier version will fail.
+  This functionality was added on Erlang/OTP 23. Using this function
+  to communicate with nodes running on earlier versions will fail.
 
   Inlined by the compiler.
   """
--- a/lib/elixir/lib/node.ex
+++ b/lib/elixir/lib/node.ex
@@ -258,6 +258,38 @@ defmodule Node do
   end
 
   @doc """
+  Returns the PID of a new process started by the application of `fun` on `node`,
+  and a reference for a monitor created to the new process. Otherwise works
+  like `spawn/1`
+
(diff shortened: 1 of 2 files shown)


lib/elixir/lib/node.ex
```diff
@@ -257,6 +257,38 @@ defmodule Node do
     :erlang.spawn_link(node, module, fun, args)
   end
 
+  @doc """
+  Returns the PID of a new process started by the application of `fun` on `node`,
+  and a reference for a monitor created to the new process. Otherwise works
+  like `spawn/1`
```
josevalim, line 263: We need to keep the first sentence short, as that is shown as a summary in ExDoc. Can you please take a look?

lib/elixir/lib/node.ex
```diff
@@ -257,6 +257,38 @@ defmodule Node do
...
+
+  If the node identified by `node` does not support distributed `spawn_monitor()`,
+  the call will fail with a `:notsup` exception.
```
josevalim, line 266: It doesn't show as `notsup` in Elixir. It will either be ErlangError or something else :)

lukaszsamson: Do you think we should leave it undocumented or indicate thet the call will raise `%ErlangError{original::notsup}`? as There is no specification for that one in `ErlangError.normalize`. Or should we normalize it to `%NotSupportedError{}`?

josevalim: My honest opinion: we would leave it undocumented. It only happens when using distribution across Erlang nodes running in different versions and, if you are at this point, then you are likely familiar with the disclaimers that may happen.

Or perhaps, we would document like this: "This functionality was added on Erlang/OTP 23. Using this functionality to communicate with a node running on an earlier version will fail.".

lib/elixir/lib/node.ex
```diff
@@ -257,6 +257,36 @@ defmodule Node do
     :erlang.spawn_link(node, module, fun, args)
   end
 
+  @doc """
+  Spawns the given function on a node, monitors it and returns its PID
+  and monitoring reference.
+
+  This functionality was added on Erlang/OTP 23. Using this functionality to communicate with
+  a node running on an earlier version will fail.
```
josevalim, line 265:
```suggestion
  This functionality was added on Erlang/OTP 23. Using this functionality to
  communicate with nodes running on earlier versions will fail.
```

lib/elixir/lib/node.ex
```diff
@@ -257,6 +257,36 @@ defmodule Node do
...
+  def spawn_monitor(node, fun) do
+    :erlang.spawn_monitor(node, fun)
+  end
+
+  @doc """
+  Spawns the given module and function passing the given args on a node,
+  monitors it and returns its PID and monitoring reference.
+
+  This functionality was added on Erlang/OTP 23. Using this functionality to communicate with
+  a node running on an earlier version will fail.
```
josevalim, line 280:
```suggestion
  This functionality was added on Erlang/OTP 23. Using this function
  to communicate with nodes running on earlier versions will fail.
```

josevalim:

PR #11541, opened 2022-01-05
Use :erlang.atom_to_binary/1

v0idpwn:
This function was added in OTP 23 and its the functional equivalent to
`:erlang.atom_to_binary/2` with `:utf8` as the second argument.

Part of #9603

### DIFF
--- a/lib/elixir/test/elixir/kernel/warning_test.exs
+++ b/lib/elixir/test/elixir/kernel/warning_test.exs
@@ -981,7 +981,7 @@ defmodule Kernel.WarningTest do
 
   # TODO: Simplify when we require Erlang/OTP 24
   if System.otp_release() >= "24" do
-    @argument_error_message "the call to :erlang.atom_to_binary/1"
+    @argument_error_message "the call to Atom.to_string/1"
     @arithmetic_error_message "the call to +/2"
   else
     @argument_error_message "this expression"
--- a/lib/elixir/test/elixir/kernel/warning_test.exs
+++ b/lib/elixir/test/elixir/kernel/warning_test.exs
@@ -981,7 +981,7 @@ defmodule Kernel.WarningTest do
 
   # TODO: Simplify when we require Erlang/OTP 24
   if System.otp_release() >= "24" do
-    @argument_error_message "the call to :erlang.atom_to_binary/2"
+    @argument_error_message "the call to Atom.to_string/1"
     @arithmetic_error_message "the call to +/2"
   else
     @argument_error_message "this expression"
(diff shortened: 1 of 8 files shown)


lib/elixir/test/elixir/kernel/warning_test.exs
```diff
@@ -981,7 +981,7 @@ defmodule Kernel.WarningTest do
 
   # TODO: Simplify when we require Erlang/OTP 24
   if System.otp_release() >= "24" do
-    @argument_error_message "the call to :erlang.atom_to_binary/2"
+    @argument_error_message "the call to Atom.to_string/1"
```
michallepicki, line 984: Not sure what prompted this change, is inlining working differently after the rewrite change?

v0idpwn: I'm also curious about this. Wondering if the change in elixir_rewrite.erl didn't work 🤔

michallepicki: Maybe it didn't work because arguments stay the same, and in that case the change in `elixir_rewrite.erl` should be changed from using a "rewrite" macro to "inline" macro. Just guessing though.

josevalim: @michallepicki correct. It can now be an inline rule.

v0idpwn: I did that, but it still doesn't work as I would expect. Am I missing something?

lib/elixir/test/elixir/kernel/warning_test.exs
```diff
@@ -981,7 +981,7 @@ defmodule Kernel.WarningTest do
 
   # TODO: Simplify when we require Erlang/OTP 24
   if System.otp_release() >= "24" do
-    @argument_error_message "the call to :erlang.atom_to_binary/2"
+    @argument_error_message "the call to :erlang.atom_to_binary/1"
```
josevalim, line 984: When you inline, we can now do a reverse mapping, so showing Atom.to_string/1 is the correct now! And that's good because it is most likely what the user wrote!

josevalim, line 984:
```suggestion
    @argument_error_message "the call to Atom.to_string/1"
```

v0idpwn: That's great! I was wondering if there was some mechanism like that in the back :)

josevalim:

PR #11545, opened 2022-01-05
Raise ArgumentError when defdelegate points to the delegating module without an as: alias

sashaafm:
The `defdelegate` macro is used in cases where we've got a higher-level module which may point to one of several implementations. For example, when we've may change cloud vendors, HTTP clients or when relying on `:mox` to mock and stub our dependencies.

Sometimes, when using a `Behaviour` based approach where we define a behaviour to be honoured by the implementing modules we use `defdelegate` with the target module being fetched from an application config, like so:

```elixir
defmodule Foo do
  defdelegate bar(), to: Application.fetch_env!(:my_app, :target_module)
end
```

However, if we erroneously set the target module as the delegating module itself, then calling the function will get stuck in an infinite loop. This may happen in cases where we follow a convention where the implementing module has the same name as the delegating module with an `.Impl` appended to it so it's an easy error to introduce when defining the application configuration if the user forgets to add the remainder of the implementing module's namespace. 

Besides, this is an error which can be overlooked during tests if these are written targeting only a mock module, thus the infinite loop is never triggered. Furthermore, this error may only be found when running a live production application since it's an error that's introduced at the configuration-level.

This PR aims to solve this by raising an `ArgumentError` with an explanatory message when, during compilation, a `defdelegate` is defined with the `:to` option pointing to the delegating module itself. AFAIK, there's no use-case to have the `defdelegate` macro point to its own module without an `:as` option (where it may alias to another function).

Any feedback on this issue or the PR itself would be appreciated 🙂


lib/elixir/lib/kernel.ex
```diff
@@ -5562,6 +5562,12 @@ defmodule Kernel do
       target =
         Keyword.get(opts, :to) || raise ArgumentError, "expected to: to be given as argument"
 
+      alias_ = Keyword.get(opts, :as)
+
+      if target == __MODULE__ and is_nil(alias_) do
+        raise ArgumentError, "to: cannot target the delegating module without an as: alias"
+      end
```
josevalim, line 5569:
```suggestion
      as = Keyword.get(opts, :as)

      if target == __MODULE__ and is_nil(as) do
        raise ArgumentError,
              "defdelegate function is calling itself, which will lead to an infinite loop. You should either change the value of the :to option or specify the :as option"
      end
```

lib/elixir/test/elixir/kernel/impl_test.exs
```diff
@@ -495,6 +495,16 @@ defmodule Kernel.ImplTest do
...
+  test "raises when :to targeting the delegating module is given without the :as option" do
+    assert_raise ArgumentError,
+                 ~r/to: cannot target the delegating module without an as: alias/,
+                 fn ->
+                   defmodule ImplAttributes do
+                     defdelegate foo(), to: __MODULE__
+                   end
+                 end
+  end
+
```
josevalim, line 507: This test should be moved to `kernel_test.exs`, as it is not related to the `@impl` attribute.

josevalim (changes requested):

sashaafm:
Provided the requested changes 🙂

josevalim:

PR #11549, opened 2022-01-06
Note performance as a use case for records in docs.

aegatlin:
This topic was discussed in [this ElixirForum thread](https://elixirforum.com/t/performance-impact-of-passing-complex-state-around-in-map-vs-a-lot-of-arguments/44919), and it was suggested in [this comment](https://elixirforum.com/t/performance-impact-of-passing-complex-state-around-in-map-vs-a-lot-of-arguments/44919/21?u=aegatlin) by @axelson that a PR to add performance as a use case for records could be a good idea. I thought I'd get that started here :D

### DIFF
--- a/lib/elixir/lib/record.ex
+++ b/lib/elixir/lib/record.ex
@@ -12,10 +12,11 @@ defmodule Record do
   manipulate the tuples, providing fast operations on top of
   the tuples' compact structure.
 
-  In Elixir, records are used mostly in two situations:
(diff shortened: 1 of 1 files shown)


josevalim:
I have dropped a comment in the forum saying that the use case in Postgrex and Inspect.Algebra is actually closer to the bullets listed here. Thank you and @axelson for keeping us in mind though and double checking. :)

Issue #11551, opened 2022-01-07
Nudge towards a common solution for a "do" expression SyntaxError


pdgonzalez872:
### Environment

Erlang/OTP 24 [erts-12.0.1] [source] [64-bit] [smp:12:12] [ds:12:12:10] [async-threads:1] [jit]

Interactive Elixir (1.13.0-dev) - press Ctrl+C to exit (type h() ENTER for help)

### Current behavior

Hi!

I'm reaching out about something that may add some value. I personally use
`with` a lot and sometimes I forget a trailing comma when moving functions
around in the `with` statement. I know that the error above means that I very likely 
forgot a trailing comma in a `with`, but the error doesn't nudge me to it despite
being completely correct in telling me why the error happened.

This is the current behavior:

```
~ :> iex
iex(1)> with true <- 1 == 1, do
** (SyntaxError) iex:1:22: unexpected reserved word: do. In case you wanted to write a "do" expression, you must either use do-blocks or separate the keyword argument with comma. For example, you should either write:

    if some_condition? do
      :this
    else
      :that
    end

or the equivalent construct:

    if(some_condition?, do: :this, else: :that)

where "some_condition?" is the first argument and the second argument is a keyword list

iex(1)>
```

### Expected behavior

A note/hint saying this SyntaxError could be a trailing comma before the `do`
in a `do` expression. That'd be great and think it would be a helpful nudge to
users (mostly newcomers) that may not haven't gotten used to/know what the
error could mean.

```
iex(1)> if true, do
** (SyntaxError) iex:1:10: unexpected reserved word: do. In case you wanted to write a "do" expression, you must either use do-blocks or separate the keyword argument with comma. For example, you should either write:

    if some_condition? do
      :this
    else
      :that
    end

or the equivalent construct:

    if(some_condition?, do: :this, else: :that)

where "some_condition?" is the first argument and the second argument is a keyword list.

** Note:

You may see this error if you forget a trailing comma before the "do" in a "do" expression.
This is common when using a `with` statement or a `for` comprehension spanning multiple lines.
If this is the case, please verify it is contructed correctly.

iex(1)>
```

As usual, feel free to discard the changes here if you want to, no hard
feelings! Thanks for your work! ❤️

whatyouhide:
Closing in favor of the PR #11552, we can keep the discussion going in there 👍 Thanks @pdgonzalez872! 💟

PR #11552, opened 2022-01-07
Add nudge to do expression error

pdgonzalez872:
Closes https://github.com/elixir-lang/elixir/issues/11551

### DIFF
--- a/lib/elixir/src/elixir_tokenizer.erl
+++ b/lib/elixir/src/elixir_tokenizer.erl
@@ -1479,7 +1479,11 @@ invalid_do_error(Prefix) ->
   "    end\n\n"
   "or the equivalent construct:\n\n"
   "    if(some_condition?, do: :this, else: :that)\n\n"
-  "where \"some_condition?\" is the first argument and the second argument is a keyword list"}.
+  "where \"some_condition?\" is the first argument and the second argument is a keyword list.\n\n"
+  "** Note:\n\n"
+  "You may see this error if you forget a trailing comma before the \"do\" in a \"do\" expression.\n"
+  "This is common when using a `with` statement or a `for` comprehension spanning multiple lines.\n"
+  "If this is the case, please verify it is contructed correctly."}.
 
 invalid_do_with_fn_error(Prefix) ->
   {Prefix, ". Anonymous functions are written as:\n\n"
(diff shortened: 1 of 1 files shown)


lib/elixir/src/elixir_tokenizer.erl
```diff
@@ -1479,7 +1479,11 @@ invalid_do_error(Prefix) ->
   "    end\n\n"
   "or the equivalent construct:\n\n"
   "    if(some_condition?, do: :this, else: :that)\n\n"
-  "where \"some_condition?\" is the first argument and the second argument is a keyword list"}.
+  "where \"some_condition?\" is the first argument and the second argument is a keyword list.\n\n"
+  "** Note:\n\n"
```
whatyouhide, line 1483: I love this message, but would not use the `** Note:` here. I'd just write the rest of the message 🙃

whatyouhide (approved):

lib/elixir/src/elixir_tokenizer.erl
josevalim, line 1483:
```suggestion
```

whatyouhide: Yes, should've added a suggestion, thanks José 😄

lib/elixir/src/elixir_tokenizer.erl
```diff
@@ -1479,7 +1479,11 @@ invalid_do_error(Prefix) ->
...
+  "You may see this error if you forget a trailing comma before the \"do\" in a \"do\" expression.\n"
+  "This is common when using a `with` statement or a `for` comprehension spanning multiple lines.\n"
+  "If this is the case, please verify it is contructed correctly."}.
```
josevalim, line 1486:
```suggestion
  "You may see this error if you forget a trailing comma before the \"do\" in a \"do\" expression"}.
```

josevalim:
Thank you, I did some changes and simplified it a bit. :) We need to be careful with specific wording in messages because it can be additionally confusing in different scenarios than the ones we originally consider.

lib/elixir/src/elixir_tokenizer.erl
pdgonzalez872: 👍 works for me!

pdgonzalez872:
No worries! Thanks @josevalim and @whatyouhide !

❤️

josevalim:

PR #11553, opened 2022-01-08
Fix syntax in notifications script

jonatanklosko:
Fun fact, this was discovered in a tree-sitter-elixir CI run :smile_cat:

PR #11554, opened 2022-01-08
remove extra parenthesis in notify.exs script

the-mikedavis:
:wave: super small fix here, just removing an unmatched parethesis

noticed it in the tree-sitter-elixir CI https://github.com/elixir-lang/tree-sitter-elixir/runs/4748062586?check_suite_focus=true


the-mikedavis:
oh shoot I didn't see #11553 :see_no_evil:

PR #11555, opened 2022-01-09
refactoring: File.write!/2  calls File.write/2

tamanugi:
`File.write/2` and `File.write!/2` has similar codes.
this PR change `File.write!/2` to call `File.write/2`, likes `File.rm!` , `File.cp!` ...

other motivation:

`File.write/2` and `File.write!/2` has bit different when passed character list  👀

```
iex(2)> File.write(["hoge.txt"], "content")
:ok
iex(3)> File.write!(["hoge.txt"], "content")
** (File.Error) could not write to file "hoge.txt": bad argument
    (elixir 1.12.2) lib/file.ex:1054: File.write!/3
```


josevalim:

PR #11556, opened 2022-01-09
Polish docs in the Path module

whatyouhide:


lib/elixir/lib/path.ex
```diff
@@ -3,21 +3,25 @@ defmodule Path do
   This module provides conveniences for manipulating or
   retrieving file system paths.
 
-  The functions in this module may receive a chardata as
-  argument (i.e. a string or a list of characters / string)
-  and will always return a string (encoded in UTF-8). If a binary
+  The functions in this module may receive a `t:IO.chardata/0` as
```
josevalim, line 6: Linking to the type is good but I am afraid it will make it less clearer. What if we keep the regular name and then say “see t:IO.chardata”?

josevalim (approved):

lib/elixir/lib/path.ex
whatyouhide: Sure, will do that and merge 👍

PR #11557, opened 2022-01-09
Raise ArgumentError with empty list in Path.join/1

whatyouhide:
We were raising a `FunctionClauseError` which can be pretty obscure
in some cases.

I think the `FunctionClauseError` makes perfect sense if you pass the wrong argument type (not a list in this case), but an empty list feels like a good candidate for a more explicit error.

### DIFF
--- a/lib/elixir/lib/path.ex
+++ b/lib/elixir/lib/path.ex
@@ -504,10 +504,14 @@ defmodule Path do
       iex> Path.join(["/", "foo", "bar/"])
       "/foo/bar"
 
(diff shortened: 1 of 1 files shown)


josevalim:
I would personally keep the function clause error. With clashing diff it is clear enough IMO :)

whatyouhide:
@josevalim it caught me off guard in the past, and clause diff is only available locally and not when reporting errors to services such as Appsignal, so I figured this would be nice. But no probs, will close this out!

